Issue #5668, opened 2024-08-01
Add more builtin comptime functions


asterite:
### Problem

Here's a list of comptime functions we should implement:

- [ ] `TraitImpl::add_method(self, method: Quoted)`
- [ ] A bunch of foreign functions, e.g. the other hash functions now that `poseidon2_permutation` is implemented

---

Missing variants that map to Expr:

```rust
pub enum ExpressionKind {
    Variable(Path),
    Quote(Tokens),
    Unquote(Box<Expression>),
    AsTraitPath(AsTraitPath),
    Error,
}

pub enum Literal {
    Str(String),
    RawStr(String, u8),
    FmtStr(String),
    Unit,
}

```

Missing variants that map to Type:

```rust
pub enum Type {
    FmtString(Box<Type>, Box<Type>),
    Alias(Shared<TypeAlias>, Vec<Type>),
    TypeVariable(TypeVariable, TypeVariableKind),
    TraitAsType(TraitId, Rc<String>, TraitGenerics),
    NamedGeneric(TypeVariable, Rc<String>, Kind),
    Function(
        Vec<Type>,
        /*return_type:*/ Box<Type>,
        /*environment:*/ Box<Type>,
        /*unconstrained*/ bool,
    ),
    Forall(GenericTypeVars, Box<Type>),
    Quoted(QuotedType),
    InfixExpr(Box<Type>, BinaryTypeOperator, Box<Type>),
    Error,
}
```

Missing variants that map to UnresolvedType:

```rust
pub enum UnresolvedTypeData {
    Array(UnresolvedTypeExpression, Box<UnresolvedType>), // [Field; 4] = Array(4, Field)
    Integer(Signedness, IntegerBitSize), // u32 = Integer(unsigned, ThirtyTwo)
    Expression(UnresolvedTypeExpression),
    String(UnresolvedTypeExpression),
    FormatString(UnresolvedTypeExpression, Box<UnresolvedType>),
    Parenthesized(Box<UnresolvedType>),
    Named(Path, Vec<UnresolvedType>, /*is_synthesized*/ bool),
    TraitAsType(Path, Vec<UnresolvedType>),
    Tuple(Vec<UnresolvedType>),
    Function(..),
    Quoted(..),
    AsTraitPath(Box<crate::ast::AsTraitPath>),
    Resolved(QuotedTypeId),
    Unspecified, // This is for when the user declares a variable without specifying it's type
    Error,
}
```

### Happy Case

[...]

PR #5680, opened 2024-08-05
feat: add `Type::as_struct`

asterite:
# Description

## Problem

Part of #5668

## Additional Context

None.

## Documentation

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
@@ -61,6 +61,7 @@ impl<'local, 'context> Interpreter<'local, 'context> {
             "type_as_constant" => type_as_constant(arguments, return_type, location),
             "type_as_integer" => type_as_integer(arguments, return_type, location),
             "type_as_slice" => type_as_slice(arguments, return_type, location),
+            "type_as_struct" => type_as_struct(arguments, return_type, location),
             "type_as_tuple" => type_as_tuple(arguments, return_type, location),
             "type_eq" => type_eq(arguments, location),
             "type_is_bool" => type_is_bool(arguments, location),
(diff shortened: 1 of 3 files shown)


test_programs/compile_success_empty/comptime_type/src/main.nr
```diff
@@ -66,5 +70,17 @@ fn main() {
...
         assert(bool_type.is_bool());
+
+        // Check Type::as_struct
+        assert(u8_type.as_struct().is_none());
+
+        let foo = Foo { x: 0 };
+        let foo_type = type_of(foo);
+        let (struct_definition, generics) = foo_type.as_struct().unwrap();
+        let fields = struct_definition.fields();
+        assert_eq(fields.len(), 1);
```
asterite, line 81: If I check `fields[0].1`, which is the type of the field, I get it's `T`. I wanted to assert that's it's the field type. Is the current behavior okay? I didn't know if `as_struct()` would return the uninstantiated struct type, or the instantiated one (it currently returns the uninstantiated one).

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -550,6 +551,27 @@ fn type_as_slice(
...
+    return_type: Type,
+    location: Location,
+) -> IResult<Value> {
+    type_as(arguments, return_type, location, |typ| {
+        if let Type::Struct(struct_type, generics) = typ {
+            Some(Value::Tuple(vec![
+                Value::StructDefinition(struct_type.borrow().id),
+                Value::Slice(
+                    generics.into_iter().map(Value::Type).collect(),
+                    Type::Quoted(QuotedType::Type),
```
jfecher, line 566: This should be the slice type, not the element type

test_programs/compile_success_empty/comptime_type/src/main.nr
jfecher: The only way to get the instantiated one is to use the generics to re-instantiate it. Ie. wrap it in a `Type::Struct` again.
We could add a method `get_struct_fields(self) -> Option<[(Quoted, Type)]>` on `Type` to get the instantiated struct field's.

jfecher (changes requested):
Need to update the slice type but otherwise LGTM

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
asterite: Done!

jfecher (approved):
:+1:

PR #5684, opened 2024-08-06
feat: add some `Module` comptime functions

asterite:
# Description

## Problem

Part of #5668

## Summary

Adds these comptime functions:
- `Quoted::as_module`
- `Module::name`
- `Module::is_contract`
- `Module::functions`
- `FunctionDefinition::name`

## Additional Context

Some notes:
- I made `Quoted::as_module` return `Option<ModuleId>`, while other `Quoted::as_...` methods currently return a non-Option instead. Eventually it would be nice to unify these but doing that in this PR would have made it much bigger and complex, mainly because those methods rely on methods that are used in many places in the compiler.
- For now I made `Module::name` and `FunctionDefinition::name` return `Quoted`. Once we have a `String` type we could change them to `String`, and the logic from going to Quoted -> String is very simple.

I was wondering... would a `String` type be what a Slice is for Arrays?

One more thought: I wonder if instead of returning `Option<ModuleId>` it wouldn't be better to return some kind of `Result`. That way the caller would know what the error is, if it happens. But I think we don't have `Result` yet in Noir, right?

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
@@ -16,7 +16,6 @@ use rustc_hash::FxHashMap as HashMap;
 
 use crate::{
     ast::IntegerBitSize,
-    elaborator,
     hir::comptime::{errors::IResult, value::add_token_spans, InterpreterError, Value},
     macros_api::{ModuleDefId, NodeInterner, Signedness},
     parser,
@@ -45,6 +44,7 @@ impl<'local, 'context> Interpreter<'local, 'context> {
             "function_def_return_type" => function_def_return_type(interner, arguments, location),
             "module_functions" => module_functions(self, arguments, location),
             "module_is_contract" => module_is_contract(self, arguments, location),
+            "module_name" => module_name(interner, arguments, location),
             "modulus_be_bits" => modulus_be_bits(interner, arguments, location),
             "modulus_be_bytes" => modulus_be_bytes(interner, arguments, location),
             "modulus_le_bits" => modulus_le_bits(interner, arguments, location),
@@ -756,6 +756,19 @@ fn module_is_contract(
     let self_argument = check_one_argument(arguments, location)?;
     let module_id = get_module(self_argument, location)?;
     Ok(Value::Bool(interpreter.elaborator.module_is_contract(module_id)))
+}
+
+// fn name(self) -> Quoted
+fn module_name(
+    interner: &NodeInterner,
+    arguments: Vec<(Value, Location)>,
+    location: Location,
+) -> IResult<Value> {
+    let self_argument = check_one_argument(arguments, location)?;
+    let module_id = get_module(self_argument, location)?;
+    let name = &interner.module_attributes(&module_id).name;
+    let tokens = Rc::new(vec![Token::Ident(name.clone())]);
+    Ok(Value::Quoted(tokens))
 }
 
 fn modulus_be_bits(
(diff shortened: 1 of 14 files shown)


compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -307,6 +312,31 @@ fn slice_insert(
...
+
+    let option_value = parser::path_no_turbofish()
+        .parse(quoted)
+        .ok()
+        .and_then(|path| {
+            interpreter.elaborate_item(interpreter.current_function, |elaborator| {
+                elaborator.resolve_module_by_path(path)
+            })
+        })
+        .map(Value::ModuleDefinition);
```
jfecher, line 335:
```suggestion
    let path = parser::path_no_turbofish().parse(quoted).ok();
    let option_value = path.and_then(|path| {
        let module = interpreter.elaborate_item(interpreter.current_function, |elaborator| {
            elaborator.resolve_module_by_path(path)
        });
        Value::ModuleDefinition(module)
    });
```
Trying to fix the formatting here a bit. Short lines like `.ok()` really bother me.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -652,6 +682,20 @@ fn zeroed(return_type: Type) -> IResult<Value> {
...
+fn function_def_name(
+    interner: &NodeInterner,
+    arguments: Vec<(Value, Location)>,
+    location: Location,
+) -> IResult<Value> {
+    let self_argument = check_one_argument(arguments, location)?;
+    let func_id = get_function_def(self_argument, location)?;
+    let func_meta = interner.function_meta(&func_id);
+    let name = &interner.definition(func_meta.name.id).name;
+    let tokens = Rc::new(vec![Token::Ident(name.clone())]);
```
jfecher, line 695:
```suggestion
    let name = interner.function_name(&func_id).clone();
    let tokens = Rc::new(vec![Token::Ident(name)]);
```

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -693,6 +737,55 @@ fn function_def_return_type(
...
+            if let ModuleDefId::FunctionId(func_id) = module_def_id {
+                Some(Value::FunctionDefinition(func_id))
+            } else {
+                None
+            }
+        })
+        .collect();
+
+    let slice_type = Type::Slice(Box::new(Type::Quoted(QuotedType::FunctionDefinition)));
+
```
jfecher, line 761:
```suggestion
```

compiler/noirc_frontend/src/parser/parser.rs
```diff
@@ -370,6 +370,10 @@ pub fn trait_bound() -> impl NoirParser<TraitBound> {
     traits::trait_bound()
 }
 
+pub fn path_no_turbofish() -> impl NoirParser<Path> {
+    path::path_no_turbofish()
+}
+
```
jfecher, line 376:
```suggestion
```
Can you make the `path` and `traits` module `pub(super)` so that we don't have to copy these functions to re-export them from here? Then we can remove this `path_no_turbofish` copy as well as the `trait_bound` copy.

asterite: Done! Let me know if I got this wrong.

jfecher (approved):
👍

PR #5685, opened 2024-08-06
feat: add mutating FunctionDefinition functions

asterite:
# Description

## Problem

Part of #5668

## Additional Context

I think we are missing a `FunctionDefinition::body`, but that could come later.

I had to get a `FuncMeta` many times in some functions because Rust wouldn't let me get two mutable borrows of interner.

Generics handling is missing... but I think right now generics can't appear in these functions, as attributes only run on top-level functions, not on impl or trait impl functions, but let me know if this is not the case.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
@@ -44,6 +44,8 @@ mod builtin;
 mod foreign;
 mod unquote;
 
+pub(crate) type ValueAndLocation = (Value, Location);
+
 #[allow(unused)]
 pub struct Interpreter<'local, 'interner> {
     /// To expand macros the Interpreter needs access to the Elaborator
@@ -76,7 +78,7 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
     pub(crate) fn call_function(
         &mut self,
         function: FuncId,
-        arguments: Vec<(Value, Location)>,
+        arguments: Vec<ValueAndLocation>,
         mut instantiation_bindings: TypeBindings,
         location: Location,
     ) -> IResult<Value> {
@@ -110,7 +112,7 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
     fn call_function_inner(
         &mut self,
         function: FuncId,
-        arguments: Vec<(Value, Location)>,
+        arguments: Vec<ValueAndLocation>,
         location: Location,
     ) -> IResult<Value> {
         let meta = self.elaborator.interner.function_meta(&function);
@@ -139,7 +141,7 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
     fn call_user_defined_function(
         &mut self,
         function: FuncId,
-        arguments: Vec<(Value, Location)>,
+        arguments: Vec<ValueAndLocation>,
         location: Location,
     ) -> IResult<Value> {
         let meta = self.elaborator.interner.function_meta(&function);
@@ -193,7 +195,7 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
     fn call_special(
         &mut self,
         function: FuncId,
-        arguments: Vec<(Value, Location)>,
+        arguments: Vec<ValueAndLocation>,
         return_type: Type,
         location: Location,
     ) -> IResult<Value> {
@@ -227,7 +229,7 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
         &mut self,
         closure: HirLambda,
         environment: Vec<Value>,
-        arguments: Vec<(Value, Location)>,
+        arguments: Vec<ValueAndLocation>,
         call_location: Location,
     ) -> IResult<Value> {
         let previous_state = self.enter_function();
@@ -1641,7 +1643,7 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
         self.evaluate_statement(statement)
     }
 
-    fn print_oracle(&self, arguments: Vec<(Value, Location)>) -> Result<Value, InterpreterError> {
+    fn print_oracle(&self, arguments: Vec<ValueAndLocation>) -> Result<Value, InterpreterError> {
         assert_eq!(arguments.len(), 2);
 
         let print_newline = arguments[0].0 == Value::Bool(true);
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
@@ -44,8 +44,6 @@ mod builtin;
 mod foreign;
 mod unquote;
 
-pub(crate) type ValueAndLocation = (Value, Location);
-
 #[allow(unused)]
 pub struct Interpreter<'local, 'interner> {
     /// To expand macros the Interpreter needs access to the Elaborator
@@ -78,7 +76,7 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
     pub(crate) fn call_function(
         &mut self,
         function: FuncId,
-        arguments: Vec<ValueAndLocation>,
+        arguments: Vec<(Value, Location)>,
         mut instantiation_bindings: TypeBindings,
         location: Location,
     ) -> IResult<Value> {
@@ -112,7 +110,7 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
     fn call_function_inner(
         &mut self,
         function: FuncId,
-        arguments: Vec<ValueAndLocation>,
+        arguments: Vec<(Value, Location)>,
         location: Location,
     ) -> IResult<Value> {
         let meta = self.elaborator.interner.function_meta(&function);
@@ -141,7 +139,7 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
     fn call_user_defined_function(
         &mut self,
         function: FuncId,
-        arguments: Vec<ValueAndLocation>,
+        arguments: Vec<(Value, Location)>,
         location: Location,
     ) -> IResult<Value> {
         let meta = self.elaborator.interner.function_meta(&function);
@@ -195,7 +193,7 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
     fn call_special(
         &mut self,
         function: FuncId,
-        arguments: Vec<ValueAndLocation>,
+        arguments: Vec<(Value, Location)>,
         return_type: Type,
         location: Location,
     ) -> IResult<Value> {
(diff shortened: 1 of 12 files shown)


jfecher:
> Generics handling is missing... but I think right now generics can't appear in these functions, as attributes only run on top-level functions, not on impl or trait impl functions, but let me know if this is not the case.

Top-level functions can still have generics:
```rs
fn length_times_two<T, let N: u32>(array: [T; N]) -> u32 {
    array.len() * 2
}
```

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -730,21 +756,206 @@ fn function_def_return_type(
...
+                InterpreterError::FailedToParseMacro {
+                    error,
+                    tokens: parameter_name_tokens,
+                    rule,
+                    file: location.file,
+                }
+            })?;
+
+        let hir_pattern = interpreter.elaborate_item(interpreter.current_function, |elaborator| {
+            elaborator.elaborate_pattern_and_store_ids(
```
jfecher, line 864: We don't want to resolve this parameter in the interpreter's current function - we want to resolve it in the target function.  So we should do `elaborate_item(Some(func_id), ...)`

This is probably the source of the "Non-comptime variable x referenced in comptime scope" error you mentioned over slack.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -730,21 +756,206 @@ fn function_def_return_type(
...
+        let parameter_pattern =
+            parser::pattern().parse(parameter_name_quoted).map_err(|mut errors| {
+                let error = errors.swap_remove(0);
+                let rule = "a pattern";
+                InterpreterError::FailedToParseMacro {
+                    error,
+                    tokens: parameter_name_tokens,
+                    rule,
+                    file: location.file,
+                }
```
jfecher, line 860:
```suggestion
                InterpreterError::FailedToParseMacro {
                    error: errors.swap_remove(0),
                    tokens: parameter_name_tokens,
                    rule: "a pattern",
                    file: location.file,
                }
```

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -730,21 +756,206 @@ fn function_def_return_type(
...
+        )?;
+        let parameter_type = get_type(tuple.pop().unwrap(), parameters_argument_location)?;
+        let parameter_name_tokens = get_quoted(tuple.pop().unwrap(), parameters_argument_location)?;
+        let parameter_name_quoted =
+            add_token_spans(parameter_name_tokens.clone(), parameters_argument_location.span);
+        let parameter_pattern =
+            parser::pattern().parse(parameter_name_quoted).map_err(|mut errors| {
+                let error = errors.swap_remove(0);
+                let rule = "a pattern";
+                InterpreterError::FailedToParseMacro {
```
jfecher, line 855: Looks like there's a lot of code in this loop. I wonder if we can clean it up any more?

E.g. I think:
1. Add token spans to quoted value
2. Parse token value with `[parser]`
3. Error with a `FailedToParseMacro`

Is a common pattern we can probably create a helper function for.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -730,21 +756,206 @@ fn function_def_return_type(
...
+        parameter_types,
+        Box::new(func_meta.return_type().clone()),
+        Box::new(Type::Unit),
+    );
+
+    // TODO: generics. In Elaborator::define_function_meta there's this code:
+    //
+    //     if !generics.is_empty() {
+    //         typ = Type::Forall(generics, Box::new(typ));
+    //     }
```
jfecher, line 888: Definitely need to handle generics! You should be able to take them from the existing function type. They may also be duplicated in `FuncMeta::all_generics` but I think those are in a different format.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -730,21 +756,206 @@ fn function_def_return_type(
...
+    let return_type = get_type(return_type_argument, return_type_argument_location)?;
+
+    let func_id = get_function_def(self_argument, self_argument_location)?;
+    let func_meta = interpreter.elaborator.interner.function_meta(&func_id);
+    match func_meta.function_body {
+        FunctionBody::Unresolved(_, _, _) => (),
+        FunctionBody::Resolving | FunctionBody::Resolved => {
+            return Err(InterpreterError::CannotMutateFunction { location })
+        }
+    }
```
jfecher, line 921: This chunk of code is also repeated in each of the new functions, so we can have a helper for it.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -730,21 +756,206 @@ fn function_def_return_type(
...
+
+    let func_meta = interpreter.elaborator.interner.function_meta(&func_id);
+    let function_type =
+        Type::Function(parameter_types, Box::new(return_type.clone()), Box::new(Type::Unit));
+
+    // TODO: generics. In Elaborator::define_function_meta there's this code:
+    //
+    //     if !generics.is_empty() {
+    //         typ = Type::Forall(generics, Box::new(typ));
+    //     }
```
jfecher, line 933: Probably easier to edit the existing function type

jfecher (changes requested):

compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs
```diff
@@ -31,10 +31,11 @@ fn poseidon2_permutation(
...
     location: Location,
 ) -> IResult<Value> {
-    let (input, state_length) = check_two_arguments(arguments, location)?;
+    let ((input, input_location), (state_length, state_length_location)) =
+        check_two_arguments(arguments, location)?;
 
-    let (input, typ) = get_array(interner, input, location)?;
-    let state_length = get_u32(state_length, location)?;
+    let (input, typ) = get_array(interner, input, input_location)?;
+    let state_length = get_u32(state_length, state_length_location)?;
```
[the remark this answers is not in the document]
asterite: Or maybe `get_u32`, `get_array`, etc., could take a `(Value, Location)` tuple instead of two separate arguments. I'll try that approach first.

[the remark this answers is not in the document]
asterite: Done. I think it looks okay, but let me know and I can change it back to `check_argument_count`. I mainly wanted to avoid having to pop arguments and get them in backwards order... but maybe it's not that confusing.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
asterite: In the end I got the generics from `FuncMeta.typ`

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
asterite: I tried to edit the existing function but for some reason it didn't work, I'm not sure why.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
asterite: Good idea, done!

asterite:
Regarding generics, I wanted to add a test to prove that it works with generics, something like:

```rust
#[mutate_generic]
fn identity<T>() {}

comptime fn mutate_generic(f: FunctionDefinition) {
  // I want to change f to be `fn identity<T>(x: T) -> T { x }`
  f.set_parameters(&[quote { x }, ???]); // what type can I pass here?
```

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
asterite: Oh, maybe it's because I also removed the `push_definition_type` call, but that should still be done. I'll try it again...

asterite: It was that! It's working now.

jfecher:
Is this PR ready for review again? It is hard for me to tell because of the force push but I see all the comments are resolved

asterite:
Oh, yes, it is, sorry! Next time I'll ask for a re-review after pushing several commits.

I'll fix the conflicts now.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -915,3 +987,73 @@ pub(crate) fn extract_option_generic_type(typ: Type) -> Type {
 
     generics.pop().expect("Expected Option to have a T generic type")
 }
+
+fn check_function_not_yet_resolved(
```
jfecher, line 991: Can you move these helpers to builtin_helpers.rs?

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
```diff
@@ -44,6 +44,8 @@ mod builtin;
 mod foreign;
 mod unquote;
 
+pub(crate) type ValueAndLocation = (Value, Location);
+
```
jfecher, line 48:
```suggestion
```
I don't think the name `ValueAndLocation` adds anything over `(Value, Location)`

jfecher (commented):
Can you avoid force pushes after a review has been done in the future? It makes it a bit more difficult to review since I can't see changes since the last review any more and need to start over.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -915,3 +988,59 @@ pub(crate) fn extract_option_generic_type(typ: Type) -> Type {
 
     generics.pop().expect("Expected Option to have a T generic type")
 }
+
+fn parse<T>(
```
jfecher, line 992: Can you move the other helper functions too? Here and below

asterite:
> Can you avoid force pushes after a review has been done in the future? It makes it a bit more difficult to review since I can't see changes since the last review any more and need to start over.

Sounds good!

I did a force push because we use squash-merge and this PR was on top of another PR. When the first PR merged, doing `git merge` gives conflicts. I found a simpler way to do that by using `git rebase --onto` to remove the extra commits. Is there another way to do this merge in a simpler way?

asterite:
(the conflicts are easy to solve, just always "keep mine", but it's a bit annoying to do it all the time for every stacked PR, so I wonder how you all deal with this)

jfecher:
@asterite usually when this happens I just go through the conflicts and "keep mine" or fix them in the case there are changes (e.g. changes requested from the previous PR).

jfecher (approved):
Thank you :+1:

Issue #5296, opened 2024-06-20
Dead Instruction Elimination pass removes out of bounds array sets


Thunkar:
### Aim

While moving some tests from constrained to `unconstrained` so they run faster (we're only checking function logic and assume circuit generation is fine), the code that used to work started failing with "Array index out of bounds" errors.

### Expected Behavior

Behavior should be consistent between constrained and unconstrained contexts.

### Bug

Out of bounds errors can be masked in certain scenarios, specifically when copying array positions from one to another. However, illegal acceses are always detected as expected in unconstrained functions.

### To Reproduce

```rust 
// Taken from noir-protocol-circuits
fn arr_copy_slice<T, N, M>(src: [T; N], mut dst: [T; M], offset: u32) -> [T; M] {
    for i in 0..dst.len() {
        // We are illegally accessing the src array here if dst.len() > src.len()
        dst[i] = src[i + offset];
    }
    dst
}

unconstrained fn boom<N, M>(src: [Field; N], dst: [Field; M]) -> [Field; M] {
    arr_copy_slice(src, dst, 0)
}


fn main() {
    // This shouldn't work
    let small_array_src = [1; 5];
    let mut big_array_dst = [0; 6];
    let mut result = arr_copy_slice(small_array_src, big_array_dst, 0);

    // Uncomment for fireworks. This *always* get caught in unconstrained land
    //let result = boom();

    // This is fine
    let first = result[0];
    dep::std::println(first);
    // This is also fine, since we are overwriting the "illegal" position.
    // BUT if we comment this line out, fireworks again (since we are now "using" the illegal position)
    result[5] = 1;

    dep::std::println(result[5]);
}
```

### Project Impact

Nice-to-have

### Workaround

None

### Installation Method

Compiled from source

### Would you like to submit a PR for this Issue?

None

jfecher:
This fails with:
```
error: Assertion failed: 'Array index out of bounds'
  ┌─ /Users/jakefecher/Code/Noir/noir/short/src/main.nr:6:18
  │
6 │         dst[i] = src[i + offset];
  │                  ---------------
  │
  = Call stack:
    1. /Users/jakefecher/Code/Noir/noir/short/src/main.nr:23:22
    2. /Users/jakefecher/Code/Noir/noir/short/src/main.nr:12:5
    3. /Users/jakefecher/Code/Noir/noir/short/src/main.nr:6:18
```
[...]

PR #5691, opened 2024-08-06
fix: Replace unused ArrayGet/Set with constrain if possibly out of bounds

asterite:
# Description

## Problem

For #5296

## Summary

In DIE, when we find an `ArrayGet/Set` that we could remove, if that instruction might be an index out of bounds we still remove it but leave a constrain check too.

## Additional Context

For now the code is a bit messy, I just wanted to get it working first.

Also, it's not working for slices yet because we can't insert out of bounds checks if we don't know the slice length. For that we'll need to add the checks like we do for `ArrayGet` for slice (this could be done in a separate PR, though).

## Documentation\*

Check one:

# PR Checklist\*


jfecher (changes requested):
How does this code compare with if we always inserted the asserts, but added an extra constant folding pass after DIE to de-duplicate them?

asterite:
> How does this code compare with if we always inserted the asserts, but added an extra constant folding pass after DIE to de-duplicate them?

The issue is that if we add asserts for `array_get`, they can't be removed (DIE doesn't remove asserts). Here we are only inserting asserts for unused instructions, and only if needed. Unless.. we change DIE to somehow detect that these asserts are related to unused instructions, and remove them... but I'm not sure how.

jfecher:
@asterite I'm suggesting we add a constant folding pass after DIE which will remove duplicate asserts. Specifically we'll want the `fold_constants_using_constraints` kind.

It would be a bit strange to run that pass once, run DIE, then run it again. I wonder if we'd still get the full benefit if we just waited until after DIE to run it.

asterite:
> I'm suggesting we add a constant folding pass after DIE which will remove duplicate asserts

I think that wouldn't be enough. Right now if you have something like this:

```rust
struct Foo {
  x: Field,
  y: Field,
}

fn main(array: [Foo; 3], index: Field) -> pub Field {
  let foo = array[index];
  foo.x
}
```

then this PR ends up not adding any checks at all:
- `array[index]` ends up doing two `array_get`
- one of them is used (the one for `x`)
- the other isn't used (the one for `y`)
- because at least one of them is used, we don't insert any checks (because a later pass, not sure which one, will check this)

If we inserted the checks early on, then added a pass that removed duplicate checks, we'd end up with one check here... unless that pass also somehow detects that none of those checks are needed, but that's what I wouldn't know how to do.

jfecher:
@asterite I was thinking we'd still try to insert the checks during DIE rather than early on for the constant folding approach

asterite:
So if I'm understand things correctly, instead of only inserting the necessary asserts, we'd insert all of the asserts, then do a pass to remove the extra ones? (or all of them if all were unused)

jfecher:
@asterite Right, the goal being to simplify the DIE pass. I'm hoping we can accomplish this by moving the existing fold_constants_with_predicates pass rather than adding another after DIE but am unsure how much this affects the optimization.

asterite:
> Right, the goal being to simplify the DIE pass

I'm not sure, but maybe adding the checks then trying to collapse them or remove them (given that a constraint might also involve some stuff before it, like comparing an index against a length) might be tricky.

Alternatively, if the goal is to simplify DIE, we could extract the "replace unused array_get/set with constrain checks" to a pass right before DIE. That pass would also have to traverse instructions backwards and see which ones are unused, to determine unused array_get/set, but it wouldn't have to keep track of some other stuff I'm seeing in DIE like `instructions_to_remove` or `rc_instructions`. What do you think?

vezenovm:
> Although I would like if there were a way to not have DIE care about this corner case.

We can move this logic out of DIE into a separate pass, but I think we are going to need to handle the case in a similar manner to the logic this PR placed inside of DIE. In the case of a dynamic array get on an array of structs, the constraint check is going to be `dynamic index < len`. We are not going to be able to remove those constraints by folding constants.

asterite:
@jfecher Now DIE remains the same as before (actually a bit shorter because I extracted some helper functions to a different file) and there's a new pass to insert bounds checks.

jfecher:
> In the case of a dynamic array get on an array of structs, the constraint check is going to be dynamic index < len. We are not going to be able to remove those constraints by folding constants.

@vezenovm I think you're missing that we also remove asserts with duplicate constraints already. That's what I'm saying would be useful here since we're worrying about inserting redundant asserts for the same array. Those should get removed by constant folding with predicates, which also performs dedeuplication.

jfecher:
A agree with @vezenovm that we shouldn't have a separate pass for this since it'd duplicate the work of DIE anyway by iterating in reverse and checking for unused array get/sets.

My suggestion earlier was to simplify DIE a bit by not having it keep track of whether a bounds check was already inserted or not and instead move the `fold_constants_with_predicates` pass after DIE to get rid of the extra checks.

vezenovm:
> My suggestion earlier was to simplify DIE a bit by not having it keep track of whether a bounds check was already inserted or not and instead move the `fold_constants_with_predicates` pass after DIE to get rid of the extra checks.

I tried this and it did not look to remove the extra checks. Maybe @asterite can also try to confirm.

vezenovm:
> I think you're missing that we also remove asserts with duplicate constraints already.

These are not exactly duplicate/redunant constraints though. We offset the dynamic index before the array get of a struct.

vezenovm:
Using the current state of this PR and commenting out the array group check for this program:
```rust
struct Foo {
  x: Field,
  y: Field,
  z: Field,
}

fn main(array: [Foo; 3], index: Field) -> pub Field {
  let foo = array[index];
  foo.x
}
```
We get the following SSA: 
```
After Constraint Folding:
acir(inline) fn main f0 {
  b0(v0: [Field, Field, Field; 3], v1: Field):
    v20 = cast v1 as u32
    v21 = mul v20, u32 3
    v22 = array_get v0, index v21
    v23 = add v21, u32 1
    v24 = array_get v0, index v23
    v25 = add v21, u32 2
    v26 = array_get v0, index v25
    return v22
}

After replacing unused array instructions with bounds checks:
acir(inline) fn main f0 {
  b0(v0: [Field, Field, Field; 3], v1: Field):
    v20 = cast v1 as u32
    v21 = mul v20, u32 3
    v22 = array_get v0, index v21
    v23 = add v21, u32 1
    v28 = lt v23, u32 9
    constrain v28 == u1 1 '"Index out of bounds"'
    v25 = add v21, u32 2
    v29 = lt v25, u32 9
    constrain v29 == u1 1 '"Index out of bounds"'
    return v22
}
```
Moving `fold_constants_using_constraints` after DIE does not remove these extra constraints and I don't expect it to do so. It is not clear to me how these constrains would be removed without the extra state @asterite has added in this PR.

jfecher:
Side note, I'm looking at
```
v23 = add v21, u32 1
v28 = lt v23, u32 9
```
and see an optimization chance we could add to get:
```
v23 = lt v21, u32 8
```
I'll make an issue for this.

jfecher:
On topic: I would have expected us to be performing checks on the length of the array in the array type, not the size of the array in fields. E.g. using your example I would have expected code like:
```
After replacing unused array instructions with bounds checks:
acir(inline) fn main f0 {
  b0(v0: [Field, Field, Field; 3], v1: Field):
    v20 = cast v1 as u32
    v21 = mul v20, u32 3
    v22 = array_get v0, index v21
    v28 = lt v20, u32 3
    constrain v28 == u1 1 '"Index out of bounds"'
    v29 = lt v20, u32 3
    constrain v29 == u1 1 '"Index out of bounds"'
    return v22
}
```
Hence why I expected deduplication of v28 and v29

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -166,6 +220,293 @@ impl Context {
...
+        .first();
+    let rhs = function
+        .dfg
+        .insert_instruction_and_results(
+            Instruction::binary(BinaryOp::Mul, rhs, casted_condition),
+            block_id,
+            None,
+            call_stack,
+        )
+        .first();
```
jfecher, line 508: Here's the formatting issue I mentioned in the call

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -166,6 +220,293 @@ impl Context {
...
+                // If we are here it means the index is dynamic, so let's add a check that it's less than length
+                let index = function
+                    .dfg
+                    .insert_instruction_and_results(
+                        Instruction::Cast(*index, Type::unsigned(SSA_WORD_SIZE)),
+                        block_id,
+                        None,
+                        call_stack.clone(),
+                    )
+                    .first();
```
jfecher, line 307: Some here as well

compiler/noirc_evaluator/src/ssa/opt/die.rs
asterite: Done! Let me know if it looks good now.

vezenovm (approved):
LGTM!

jfecher (approved):

PR #5696, opened 2024-08-08
fix: Do not duplicate redundant Brillig debug metadata

### LINKED ISSUE #5521: Calling the same unconstrained fn multiple times in the program multiplies debug metadata
### Aim

Have a program that uses brillig hints in a loop

### Expected Behavior

Artifact size and compile time should not explode.

### Bug

[In this project](https://github.com/iAmMichaelConnor/blob-lib/tree/domain-size-4096?tab=readme-ov-file) ram usage goes to 100gigs, postprocessing after ACIR generation can take 10 minutes and the compressed debug metadata in the artifact weights 1 gigabyte.

The problem is that we always associate debug metadata with acir callsites. We've succesfully deduplicated the brillig code by the separation unconstrained fn / brillig call in the Program struct, but brillig-related debug metadata still gets added over and over again on every callsite, just changing the ACIR index.

### To Reproduce

Minimal repro:
```rust

global ITERATION_COUNT = 100000;

unconstrained fn some_unconstrained_fn(x: Field, y: Field) -> Field {
    let x_u128 = U128::from_integer(x);
    let y_u128 = U128::from_integer(y);
    (x_u128 / y_u128).to_integer()
}

fn main(data: [Field; ITERATION_COUNT]) -> pub Field {
    let mut acc = data[0];

    for i in 1..ITERATION_COUNT {
        acc = some_unconstrained_fn(acc, data[i]);
    }

    acc
}

```

`some_unconstrained_fn` has 1558 locations. It gets called 100k times, and the final Program struct debug metadata `locations` has 155M entries.

### Project Impact

None

### Impact Context

_No response_

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Installation Method

None

### Nargo Version

_No response_

### NoirJS Version

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #5521 

## Summary\*

This PR adds a new `brillig_locations` map to our generated ACIR that maps a brillig function id to an opcode locations map `BTreeMap<OpcodeLocation, CallStack>`. If we have called that Brillig function id before we do not add its debug locations. 

For the same test as in the issue I ran `du -sh ./target` and got a size of 2.3M. This is a >400x reduction in size of the artifact. Compilation also takes a matter of a couple seconds now rather than 10 minutes.

## Additional Context

The call stacks are the same as expected. For example, under `execution_failure/fold_nested_brillig_assert_fail` we have the following call stack:
Screenshot 2024-08-08 at 10 33 28 AM

## Documentation\*

Check one:

# PR Checklist\*


acvm-repo/acvm/src/pwg/brillig.rs
```diff
@@ -29,6 +29,10 @@ pub enum BrilligSolverStatus<F> {
 pub struct BrilligSolver<'b, F, B: BlackBoxFunctionSolver<F>> {
     vm: VM<'b, F, B>,
     acir_index: usize,
+    /// This id references which Brillig function within the main ACIR program we are solving.
+    /// This is used for appropriately resolving errors as the ACIR program artifacts
+    /// set up their Brillig debug metadata by function id.
+    function_id: u32,
```
jfecher, line 35:
```suggestion
    function_id: BrilligFunctionId,
```
We should avoid raw integer types for IDs

jfecher (changes requested):

acvm-repo/acvm/src/pwg/brillig.rs
vezenovm: done, I also replaced other usage of u32s for brillig function ids

acvm-repo/acir/src/circuit/brillig.rs
```diff
@@ -27,3 +27,6 @@ pub enum BrilligOutputs {
 pub struct BrilligBytecode<F> {
     pub bytecode: Vec<BrilligOpcode<F>>,
 }
+
+/// Id for the function being called.
+pub type BrilligFunctionId = u32;
```
jfecher, line 32: This is better, but I still think it should be it's own struct type like we have for other Ids in the program.

vezenovm: I have updated to a struct type. We use raw integers types for ACIR calls as well. I think this can be left for a separate PR though so I have made an issue here https://github.com/noir-lang/noir/issues/5700

jfecher (approved):

vezenovm:
FYI @ggiraldez @anaPerezGhiglia this work is going to require debugger updates as the debug metadata format has been changed. I will look at just updating it myself, but if you know right away where the update is needed feel free to push.

Issue #1577, opened 2023-06-06
LSP: Support completions


phated:
### Problem

The LSP should support making completion suggestions following https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion

### Happy Case

A user should receive completion suggestions when editing code while using an LSP client, such as the vscode plugin.

### Would you like to submit a PR for this Issue?

No

PR #5704, opened 2024-08-09
feat: LSP autocompletion for use statement

asterite:
# Description

## Problem

Part of #1577

## Summary

Implements autocompletion for use statements, so typing anything after `use`.

https://github.com/user-attachments/assets/a9ecd738-30c2-41fc-8ea0-3ea93d5d99ab

Note: the gray hints are probably coming from copilot, I should have turned it off before recording 😅 

## Additional Context

To begin implementing autocompletion I thought starting with `use` would be the easiest thing, but also not having to look up the std (or your project) module hierarchy could be useful.

To implement this I had to allow parsing `use foo::`, that is, allowing trailing colons, because otherwise that wouldn't parse to a use statement and we wouldn't have any information to do the autocompletion.

There are still some cases that don't work. For example if you write `use ` and explicitly ask for autocompletion, you don't get anything. The reason is that `use ` gives a parse error so no AST is produced. We could tackle those in later PRs (I didn't want to continue making this PR bigger).

This PR also introduces a simple way to test autocompletions, so next autocompletion PRs should be much smaller.

Another thought: at first I was worried that caching `def_maps` in the LSP state would increase the memory usage by a lot, but in retrospective I think that's not a heavy data structure, or at least `NodeInterner`, which was already cached, is probably much larger in memory footprint. I tried this PR in a few projects and the memory usage was fine.

## Documentation

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir/def_map/module_data.rs
+++ b/compiler/noirc_frontend/src/hir/def_map/module_data.rs
@@ -137,8 +137,4 @@ impl ModuleData {
     pub fn value_definitions(&self) -> impl Iterator<Item = ModuleDefId> + '_ {
         self.definitions.values().values().flat_map(|a| a.values().map(|(id, _, _)| *id))
     }
-
-    pub fn value_idents(&self) -> impl Iterator<Item = &Ident> + '_ {
-        self.definitions.values().keys()
-    }
 }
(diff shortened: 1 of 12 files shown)


asterite:
Sorry for the double review request, I noticed it wasn't working well in some cases so I fixed it before requesting a review.

jfecher:
@asterite should this still be a draft?

asterite:
@jfecher Nope, forgot to undo that too. Thanks! It's ready for review :-)

jfecher (approved):
👍

PR #5708, opened 2024-08-09
feat: add `Quoted::as_expr` and `Expr::as_function_call`

asterite:
# Description

## Problem

Part of #5668

## Summary

What the title says.

## Additional Context

I didn't know how to test the exprs. I tried implementing `impl Eq for Expr` but even though two exprs looked the same (same tokens) they weren't equal... but I didn't dig deeper into why that is (I didn't know if `Eq` was desired in this case)

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
@@ -6,7 +6,7 @@ use std::{
 use acvm::{AcirField, FieldElement};
 use builtin_helpers::{
     check_argument_count, check_function_not_yet_resolved, check_one_argument,
-    check_three_arguments, check_two_arguments, get_function_def, get_module, get_quoted,
+    check_three_arguments, check_two_arguments, get_expr, get_function_def, get_module, get_quoted,
     get_slice, get_struct, get_trait_constraint, get_trait_def, get_tuple, get_type, get_u32,
     hir_pattern_to_tokens, mutate_func_meta_type, parse, parse_tokens,
     replace_func_meta_parameters, replace_func_meta_return_type,
@@ -17,8 +17,8 @@ use rustc_hash::FxHashMap as HashMap;
 
 use crate::{
     ast::{
-        FunctionKind, FunctionReturnType, IntegerBitSize, UnresolvedType, UnresolvedTypeData,
-        Visibility,
+        ExpressionKind, FunctionKind, FunctionReturnType, IntegerBitSize, UnresolvedType,
+        UnresolvedTypeData, Visibility,
     },
     hir::comptime::{errors::IResult, value::add_token_spans, InterpreterError, Value},
     hir_def::function::FunctionBody,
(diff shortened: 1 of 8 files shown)


compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -672,6 +688,41 @@ fn zeroed(return_type: Type) -> IResult<Value> {
...
+    arguments: Vec<(Value, Location)>,
+    return_type: Type,
+    location: Location,
+) -> IResult<Value> {
+    expr_as(arguments, return_type, location, |expr| {
+        if let ExpressionKind::Call(call_expression) = expr {
+            let function = Value::Expr(call_expression.func.kind);
+            let arguments = call_expression.arguments.into_iter();
+            let arguments = arguments.map(|argument| Value::Expr(argument.kind)).collect();
+            let arguments = Value::Slice(arguments, Type::Quoted(QuotedType::Expr));
```
jfecher, line 702: Type here should be a slice of quoted exprs

compiler/noirc_frontend/src/hir/comptime/value.rs
```diff
@@ -345,7 +348,8 @@ impl Value {
                 HirExpression::Literal(HirLiteral::Slice(HirArrayLiteral::Standard(elements)))
             }
             Value::Quoted(tokens) => HirExpression::Unquote(add_token_spans(tokens, location.span)),
-            Value::Pointer(..)
+            Value::Expr(..)
```
jfecher, line 351: We can convert an expression into a HIR expression by elaborating it.

I honestly need to review our use of `into_hir_expression` though on whether we should keep it at all. It feels wrong to have both it and `into_expression` and the `Value::Quoted` is different with the HIR version returning an `Unquote` expression instead of immediately parsing & elaborating the tokens into their actual hir expr variant.

noir_stdlib/src/meta/expr.nr
```diff
@@ -0,0 +1,7 @@
+use crate::cmp::Eq;
```
jfecher, line 1:
```suggestion
```
Don't need this yet

jfecher (commented):
> I didn't know how to test the exprs. I tried implementing impl Eq for Expr but even though two exprs looked the same (same tokens) they weren't equal... but I didn't dig deeper into why that is (I didn't know if Eq was desired in this case)

I imagine it's due to the Span fields on the nested Expressions. We'd have to manually implement equality to ignore those fields but I doubt it'd be worth it.

compiler/noirc_frontend/src/hir/comptime/value.rs
asterite: I looked at the code and it seems we don't have an `Elaborator` at this point. Do you want me to try to do that, or we first solve the thing that feels wrong? (not sure what changes I need to make here, if any)

Oh, and to clarify, I should have put a `todo!()` here.

jfecher: My bad, forgot we weren't in the interpreter. Let's leave it as-is for now with the error and we can revisit how into_hir_expression is used later.

jfecher (approved):

Issue #5707, opened 2024-08-09
Optimize arithmetic across multiple instructions


jfecher:
### Problem

Currently when simplifying instructions we only look at one instruction at a time for the most part.

For example, the program:
```rs
fn main(x: Field) {
    let x2 = x + 1 - 1;
    assert_eq(x, x2);
}
```
Does not optimize out x2:
```
After Array Set Optimizations:
acir(inline) fn main f0 {
  b0(v0: Field):
    v9 = add v0, Field 1
    v10 = sub v9, Field 1
    constrain v0 == v10
    return 
}
```

### Happy Case

There are many cases when multiple instructions could constant combine like this so I won't list them all. Another case as mentioned in https://github.com/noir-lang/noir/pull/5691 is with addition and less than which can show up in length checks of arrays in some cases:
```
v23 = add v21, u32 1
v28 = lt v23, u32 9
```
Which could be simplified to:
```
v23 = lt v21, u32 8
```

There are a lot of checks we could add but I think this could be a relatively easy optimization win still since I anticipate patterns like this will show up in user code either directly or as the result of previous optimizations as in the `lt` case above.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #5709, opened 2024-08-10
feat: add initial optimizations for multi-instruction arithmetic

TomAFrench:
# Description

## Problem\*

Resolves (partially) #5707

## Summary\*

This PR add some simple handling for simplifying arithmetic which spans multiple instructions. I've only handled a couple of simple cases and made some (imo reasonable) assumptions about the shape of the SSA instructions being optimised in order to keep the code simple. 

## Additional Context

One thing to consider with these forms of optimisations is that we can end up removing range checks which would have been applied to intermediate values, eg. 

```rust
fn main(x: u8) {
    let x2 = x + 255 - 255;
    assert_eq(x, x2);
}
```
This program would originally produce the ACIR

```
func 0
current witness index : 2
private parameters indices : [0]
return value indices : []
BLACKBOX::RANGE [(0)] [ ]
EXPR [ (1, _0) (-1, _1) 255 ]
BLACKBOX::RANGE [(1)] [ ]
EXPR [ (1, _1) (-1, _2) -255 ]
BLACKBOX::RANGE [(2)] [ ]
EXPR [ (1, _0) (-1, _2) 0 ]
```

which would obviously fail for any inputs other than `x=0` whereas now this program will always succeed. I think it's desirable to "rearrange the users arithmetic" in such a way to avoid overflows on intermediate values however.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Regressions are cause by the case where an intermediate value is being constrained and this allows another instruction to be resolved to a constant, e.g.

```
acir(inline) fn main f0 {
  b0(v0: Field):
    v17 = add v0, Field 1
    v18 = eq v17, Field 1
    constrain v17 == Field 1
    v19 = allocate
    v20 = add v17, Field 2
    v21 = eq v20, Field 3
    constrain v20 == Field 3
    return 
}

```

Now that we're rewriting `v20` in terms of `v0` we cannot remove the second constrain instruction.

TomAFrench:
This is starting to fall into the same issues as the "bubble up" SSA pass.

jfecher:
Hmm the optimization clash with constant folding with constraints certainly makes this less of a simple issue

compiler/noirc_evaluator/src/ssa/ir/instruction/constrain.rs
```diff
@@ -2,6 +2,44 @@ use acvm::{acir::AcirField, FieldElement};
 
 use super::{Binary, BinaryOp, ConstrainError, DataFlowGraph, Instruction, Type, Value, ValueId};
 
+/// Try to simplify this constrain instruction. This function will inspect the inputs to the constraint such that
+/// it acts on variables as early in the DFG as possible.
+pub(super) fn simplify_constrain(
```
jfecher, line 7: Was this to try to limit the scope of the clash with the constant folding w/ constraints issue you mentioned?

TomAFrench: Yes, that's correct.

TomAFrench:
This may need to be switched into a dedicated SSA pass rather than something to be done immediately on inserting the instruction.

PR #5712, opened 2024-08-12
feat: LSP path completion

asterite:
# Description

## Problem

Part of #1577

## Summary

LSP will now suggest completions anywhere there's a Path. Here are some scenarios:

### Autocomplete functions in full paths, with parameters

lsp-complete-function

### Autocomplete functions when they exist in a use statement

lsp-complete-function-after-use

### Built-in functions are also suggested

lsp-builtin-functions

### Autocomplete local variables

lsp-local-var

### Autocomplete types in several type positions

lsp-suggest-type

### Built-in types are also suggested

lsp-builtin-type

### Autocomplete type parameters

lsp-type-parameter

## Additional Context

Sorry for the large diff, the bulk of it is traversing the AST. I thought about using something like the Visitor pattern, but I'm not exactly sure how it would look in Rust and whether it would greatly reduce the amount of code. It could be a separate refactor, though! (and we have tons of tests now, so it should be safe to do too).

Another big chunk of code is tests... So the actual logic isn't that large. 

Some more things could be done on top of this, like suggesting `Self` if it's available, but I wanted to stop making this PR bigger 😊 

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/lsp/src/requests/completion.rs
+++ b/tooling/lsp/src/requests/completion.rs
@@ -116,6 +116,8 @@ struct NodeFinder<'a> {
     def_maps: &'a BTreeMap<CrateId, CrateDefMap>,
     dependencies: &'a Vec<Dependency>,
     interner: &'a NodeInterner,
+    /// Completion items we find along the way.
+    completion_items: Vec<CompletionItem>,
     /// Local variables in the current scope, mapped to their locations.
     /// As we traverse the AST, we collect local variables.
     local_variables: HashMap<String, Span>,
@@ -154,31 +156,31 @@ impl<'a> NodeFinder<'a> {
             def_maps,
             dependencies,
             interner,
+            completion_items: Vec::new(),
             local_variables: HashMap::new(),
             type_parameters: HashSet::new(),
         }
     }
 
     fn find(&mut self, parsed_module: &ParsedModule) -> Option<CompletionResponse> {
-        self.find_in_parsed_module(parsed_module)
-    }
+        self.find_in_parsed_module(parsed_module);
 
-    fn find_in_parsed_module(
-        &mut self,
-        parsed_module: &ParsedModule,
-    ) -> Option<CompletionResponse> {
-        for item in &parsed_module.items {
-            if let Some(response) = self.find_in_item(item) {
-                return Some(response);
-            }
+        if self.completion_items.is_empty() {
+            None
+        } else {
+            Some(CompletionResponse::Array(std::mem::take(&mut self.completion_items)))
         }
-
-        None
     }
 
-    fn find_in_item(&mut self, item: &Item) -> Option<CompletionResponse> {
+    fn find_in_parsed_module(&mut self, parsed_module: &ParsedModule) {
+        for item in &parsed_module.items {
+            self.find_in_item(item);
+        }
+    }
+
+    fn find_in_item(&mut self, item: &Item) {
         if !self.includes_span(item.span) {
-            return None;
+            return;
         }
 
         match &item.kind {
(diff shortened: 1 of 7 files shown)


tooling/lsp/src/requests/completion.rs
```diff
@@ -381,6 +1149,79 @@ impl<'a> NodeFinder<'a> {
...
+                "assert(…)",
+                CompletionItemKind::FUNCTION,
+                "assert(${1:predicate})",
+                Some("fn(T)".to_string()),
+            ));
+            self.completion_items.push(snippet_completion_item(
+                "assert(…)",
+                CompletionItemKind::FUNCTION,
+                "assert(${1:predicate}, ${2:message})",
+                Some("fn(T, str)".to_string()),
```
jfecher, line 1164: Maybe we should change this to `fn<let N: u32>(T, str<N>)` ? Along with assert_eq below.

It did strike me as a bit odd when I saw the completion message and saw two listings for `assert`. What do other languages do for optional arguments?

tooling/lsp/src/requests/completion.rs
```diff
@@ -381,6 +1149,79 @@ impl<'a> NodeFinder<'a> {
...
+            }
+        }
+    }
+
+    fn builtin_types_completion(&mut self, prefix: &str) {
+        for typ in [
+            "bool",
+            "i8",
+            "i16",
+            "i32",
```
jfecher, line 1199: This list looks like it'll be very easy to forget to update when we add more builtin types for comptime code.

jfecher (approved):
LGTM. I mostly only skimmed though. I think aztec people will be very excited for this new LSP feature.

tooling/lsp/src/requests/completion.rs
asterite: Good point. I was wondering how we'd dehardcode it... for example some token keywords are type names (FunctionDefinition, etc.) so we could iterate over all keywords and ask whether a keyword corresponds to a built-in type (and which one), so if we add a new keyword for a type we'd have to consider that new case.

We could maybe do the same thing with `Token::IntType`, though the bit sizes are `u32` and not an enum...

Let me know if this approach or a different one would be good and I can work on it.

I guess we'd do something similar with `assert` and `assert_eq` (they come from keywords)

jfecher: I think iterating through the tokens is more or less the only way I can think of to get an error if we miss any. There's a similar test we have for keywords in the tokens file.

tooling/lsp/src/requests/completion.rs
asterite: Ah, yes, I initially had only one overload (the one without a message). Then I realized two versions existed and I thought of suggesting both...

I just checked typescript and when it suggests a function name you just get the function name, not the parameters. Then when you type `(` you get hints for the parameters (we can do that, but it's a different LSP feature).

We could either show just one overload (the one without a message?) or not show these built-in functions... or something else? what's your preference?

tooling/lsp/src/requests/completion.rs
asterite: > There's a similar test we have for keywords in the tokens file.

Oh, nice! I was looking at how I can iterate over all enum values, and found that there's a crate called `strum`, but I wasn't sure about adding that dependency... but we already have it! So, great, I'll try to do it.

tooling/lsp/src/requests/completion.rs
asterite: For now I kept only one overload (the one without a message), mainly for simplicity. We could revisit this minor issue later if needed.

tooling/lsp/src/requests/completion.rs
asterite: Done! The only thing that's still hardcoded are the several integer types, and the `false` and `true` keywords. But maybe those won't get larger...

PR #5716, opened 2024-08-13
feat: add `Type::get_trait_impl`

asterite:
# Description

## Problem

Part of #5668

## Additional Context

I thought about adding some `TraitImpl` methods but given that this introduces a new type and a new value I wanted to get feedback on that approach first. There's also a TODO that needs to be addressed.

## Documentation

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -507,6 +510,34 @@ fn type_eq(arguments: Vec<(Value, Location)>, location: Location) -> IResult<Val
...
+    {
+        match trait_impl_kind {
+            crate::node_interner::TraitImplKind::Normal(trait_impl_id) => {
+                Some(Value::TraitImpl(trait_impl_id))
+            }
+            crate::node_interner::TraitImplKind::Assumed { .. } => None,
+        }
+    } else {
+        None
+    };
```
jfecher, line 536:
```suggestion
    let option_value = match interner.try_lookup_trait_implementation(&typ, trait_id, &generics) {
        Ok((TraitImplKind::Normal(trait_impl_id), _)) => Some(Value::TraitImpl(trait_impl_id)),
        _ => None,
    };
```

compiler/noirc_frontend/src/hir/comptime/value.rs
```diff
@@ -528,6 +532,12 @@ impl<'value, 'interner> Display for ValuePrinter<'value, 'interner> {
                 let trait_ = self.interner.get_trait(*trait_id);
                 write!(f, "{}", trait_.name)
             }
+            Value::TraitImpl(trait_impl_id) => {
+                let trait_impl = self.interner.get_trait_implementation(*trait_impl_id);
+                let trait_impl = trait_impl.borrow();
+                // TODO: what's a good display for a trait impl?
+                write!(f, "{}", trait_impl.ident)
```
jfecher, line 539: I think we should probably display the "signature" of the impl.
E.g. `impl<A, B> Foo for Bar<A, B> where A: Foo`.

Not sure if we store all of the generics so it may have to be close enough if not.

jfecher (changes requested):
Looks good, just the one formatting item and the impl printing. Blocking for the printing issue.

jfecher (approved):

PR #5719, opened 2024-08-14
fix: let LSP autocompletion work in more contexts

asterite:
# Description

## Problem

There were a couple of contexts where autocompletion wouldn't trigger. This PR fixes some of these.

### Autocompletion inside if condition without following body

Before:

lsp-autocomplete-if-before

After:

lsp-autocomplete-if-after

### Autocompletion in nested expression

Before:

lsp-autocomplete-nested-before

The reason is that is was autocompleting for the type of `foo.bar & foo` instead of just `foo`.

After:

lsp-autocomplete-nested-after

### Autocompletion in call chains

Before:

lsp-autocomplete-chain-before

After:

lsp-autocomplete-chain-after

### Autocompletion when assignment follows

Before:

lsp-autocomplete-assignment-follows-before

After:

lsp-autocomplete-assignment-follows-after

## Additional Context

I might send a follow-up PR to split `completion.rs` into several files because it's getting pretty big. For example tests could go in a separate file.

## Documentation

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/expressions.rs
+++ b/compiler/noirc_frontend/src/elaborator/expressions.rs
@@ -369,7 +369,7 @@ impl<'context> Elaborator<'context> {
                     function_args.push((typ, arg, span));
                 }
 
-                let location = Location::new(span, self.file);
+                let location = Location::new(object_span, self.file);
                 let method = method_call.method_name;
                 let turbofish_generics = generics.clone();
                 let is_macro_call = method_call.is_macro_call;
--- a/compiler/noirc_frontend/src/elaborator/expressions.rs
+++ b/compiler/noirc_frontend/src/elaborator/expressions.rs
@@ -369,7 +369,8 @@ impl<'context> Elaborator<'context> {
                     function_args.push((typ, arg, span));
                 }
 
-                let location = Location::new(method_name_span, self.file);
+                let call_span = Span::from(object_span.start()..method_name_span.end());
+                let location = Location::new(call_span, self.file);
                 let method = method_call.method_name;
                 let turbofish_generics = generics.clone();
                 let is_macro_call = method_call.is_macro_call;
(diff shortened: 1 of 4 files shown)


compiler/noirc_frontend/src/elaborator/expressions.rs
```diff
@@ -369,7 +369,7 @@ impl<'context> Elaborator<'context> {
                     function_args.push((typ, arg, span));
                 }
 
-                let location = Location::new(span, self.file);
+                let location = Location::new(method_name_span, self.file);
```
asterite, line 372: I changed this because...

Given a call like this:

```rust
foo.bar(1);
```

two ExprId are interned (their locations and types):
1. The method call function type (for example `fn(i32) -> Field` assuming `bar` returns `Field`)
2. The expression's type (`Field`)

Both used the same span so later when getting the type at that location we'd get two types and only one of them would win.

But I think it's more correct to only use the location of the method's name (bar) here if we are going to intern the function type, and not the return type. But let me know if this change is wrong!

vezenovm: This is technically the location for the entire method call though and not just the method name. Was this causing issues for the LSP?

asterite: Ah, good point. I changed it to be the span from the object to the name, like this:

```rust
foo.bar(...)
^^^^^^^
```

It was causing issues because in `id_to_location` and `id_to_type` we captured two expressions with the same location and different types (one for the function type, another for the function return type). LSP relies on going from a location to an expression to a type, and it was finding two expressions in that locations. The logic in that case is to use the first one, but they are unordered so we could get the correct one or the wrong one.

I'm not sure why the function type is captured in `id_to_location` and `id_to_type`, though, but maybe if the type i the function type maybe it makes more sense for its location to not include the arguments, etc.

jfecher (approved):

PR #5725, opened 2024-08-15
feat: LSP signature help

### LINKED ISSUE #1585: LSP: Support signature help
### Problem

The LSP should support signature help following https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_signatureHelp

### Happy Case

I'm not exactly sure what "Signature Help" is, so we need to do some more research on the outcome of supporting this from the user perspective. 

### Alternatives Considered

_No response_

### Additional Context

_No response_

No

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #1585

## Summary

While working on autocompletion I was always checking what Rust Analyzer did, or how it worked, and I noticed that in Rust Analyzer when it autocompletes a function a little popup with the parameter names and types would show up, which is really useful! But it turned out that's a separate feature: signature help.

This PR implements that. I think it makes autocompletion much better. 

Before:

lsp-signature-help-before

After:

lsp-signature-help-after

Note: in the latest version the popup text starts with "fn ", but I didn't want to re-record all the gifs 😅 .

You basically don't have to remember what were the parameter types.

It also works with methods:

lsp-signature-help-method

And if you have an `fn`:

lsp-signature-help-fn

And here it's working in an aztec project (I always check that it works outside of small toy programs 😊):

lsp-signature-help-aztec

## Additional Context

I promise this will be the last big LSP PR I send 🤞 . Now that we have most of the LSP features registered, next PRs should be much smaller, just adding on top of what there is.

These PRs are also relatively big because they traverse the AST and that code is kind of boilerplatey. For that reason here I decided to put the boilerplatey code in a `traversal.rs` file, so that it's clear there's no extra logic there other than traversing the AST. I might send a follow-up PR doing the same for autocompletion.

One more thing: you can trigger this functionality as a VSCode command (Trigger Parameter Hints). But then when offering autocompletion you can also tell VSCode to execute a command, and in this PR we tell it to execute exactly that command (Rust Analyzer does the same, though they have a custom command for it, not sure why).

UPDATE: I managed to implement the visitor pattern to reduce the boilerplate. [Here's a PR](https://github.com/noir-lang/noir/pull/5727) with that. I think it will greatly simplify things, and make it easier to keep the code updated as we add more nodes. It takes inspiration in [how it's done in the Crystal compiler](https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/syntax/visitor.cr), which in turn is inspired by how it was done in the Java editor support for Eclipse 😄 

## Documentation\*

Check one:

# PR Checklist\*


tooling/lsp/src/requests/signature_help/tests.rs
```diff
@@ -0,0 +1,197 @@
...
+    async fn get_signature_help(src: &str) -> SignatureHelp {
+        let (mut state, noir_text_document) = test_utils::init_lsp_server("document_symbol").await;
+
+        let (line, column) = src
+            .lines()
+            .enumerate()
+            .filter_map(|(line_index, line)| {
+                line.find(">|<").map(|char_index| (line_index, char_index))
+            })
+            .next()
```
jfecher, line 23:
```suggestion
            .find_map(|(line_index, line)| {
                line.find(">|<").map(|char_index| (line_index, char_index))
            })
```
`filter_map -> next` can be `find_map`. I think there is a clippy warning for this

jfecher (approved):

PR #5727, opened 2024-08-15
chore: introduce the Visitor pattern

asterite:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*


asterite:
To be honest, I'm a bit hesitant about this. It helps with not repeating how to traverse the AST, but on the other hand if a new AST node is added we can add it to the visitor but there's no requirement for all visitors to handle it, so we might miss handling new nodes 🤔

asterite:
I continued working on this a bit more, and... I'm less and less sure about this. It makes the code a bit harder to follow, and it's easy to forget to handle some cases. The resulting code is a bit shorter, but traversing the whole AST is about 300 lines of code, and it's simple code, so it's also not that bad to do it manually.

I'll close this, but the code is here in case we want to use it in the future.

jfecher (commented):
>  it's easy to forget to handle some cases.

I think something that would help this & the code size a bit would be to collapse some cases. E.g. Rust has pattern matching so we don't need to have a method for every expression or statement type. We can have one `visit_expression` function where the function can then just match on the expression cases it cares about since the trait should still handle the recursion, e.g:
```
fn visit_expression(&mut self, expr: &Expression) {
    match &expr.kind {
        ExpressionKind::Variable(..) => do_something(),
        ExpressionKind::Literal(..) => do_something_else(),
        _ => (), // don't care! Let the trait helpers recur for us behind the scenes.
    }
}
```
Similarly, another strategy would be that we don't have to go all out with this. The trait could just be for recurring over expressions. We also may not need a trait at all. It is more common in functional languages to have a mapping or folding function than a full visitor pattern. E.g. we can have a helper that recurs over expressions and calls a helper function on each node so that we no longer have to handle the recursion ourselves.

PR #5730, opened 2024-08-15
feat: suggest tuple fields in LSP completion

asterite:
# Description

## Problem

Part of #1577

(there are many more completions we could support and it's hard to list all of them upfront, so I'll just link to that issue from now on)

## Summary

Suggest tuple fields (0, 1) in autocompletion, which is useful if you don't know the type you are handling with (maybe it's a result of another expression that's not in a variable).

Also I noticed tuple methods showed up duplicate, so that's fixed too.

Before:

lsp-complete-tuple-before

After:

lsp-complete-tuple-after

## Additional Context

The two first commits are just refactors, so I'd suggest going commit by commit.

## Documentation

Check one:

# PR Checklist


jfecher (approved):
:+1:

PR #5734, opened 2024-08-15
feat: add `Expr::as_binary_op`

asterite:
# Description

## Problem

Part of #5668

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/noir_stdlib/src/meta/op.nr
+++ b/noir_stdlib/src/meta/op.nr
@@ -19,3 +19,74 @@ impl UnaryOp {
         self.op == 3
     }
 }
+
+struct BinaryOp {
+    op: Field
+}
+
+impl BinaryOp {
+    fn is_add(self) -> bool {
+        self.op == 0
+    }
+
+    fn is_subtract(self) -> bool {
+        self.op == 1
+    }
+
+    fn is_multiply(self) -> bool {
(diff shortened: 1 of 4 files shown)


noir_stdlib/src/meta/op.nr
```diff
@@ -19,3 +19,74 @@ impl UnaryOp {
...
     }
 }
+
+struct BinaryOp {
+    op: Field
+}
+
+impl BinaryOp {
+    fn is_add(self) -> bool {
+        self.op == 0
```
TomAFrench, line 29: Worth creating some globals for this for readability?

asterite: Hey Tom, welcome back! 👋 

I'm not sure. I think it's clear what the number means based on the enclosing method name. But if you think it's worth it, let me know and I'll do it.

jfecher (approved):
LGTM. I don't see us needing to use these constants elsewhere so I don't think we need globals for them.

noir_stdlib/src/meta/op.nr
TomAFrench: I don't have overly strong opinions on this. Based on the number of these that we have, I would add these as a quasi-enum but it's 2-1 on this point.

Issue #5544, opened 2024-07-17
Ambiguous associated type panics


michaeljklein:
### Aim

Attempted to compile:

```rust
trait MyTrait { type X; }

fn main() {
    // `X` is ambiguous
    let foo: MyTrait::X;
}
```

### Expected Behavior

Expected compilation to fail with a user error

### Bug

Compiler panics:

```bash
The application panicked (crashed).
Message:  index out of bounds: the len is 0 but the index is 18446744073709551615
Location: compiler/noirc_frontend/src/node_interner.rs:1156
```

### To Reproduce

1.
2.
3.
4.

### Project Impact

None

### Workaround

None

### Installation Method

Binary (`noirup` default)

### NoirJS Version

nargo version = 0.31.0 noirc version = 0.31.0+45e82a672b9ba7f7326e8d9f8800e2489013e2e8 (git version hash: 45e82a672b9ba7f7326e8d9f8800e2489013e2e8, is dirty: false)

### Would you like to submit a PR for this Issue?

None

asterite:
Reopened because the linked issue says "does not fix ..." but I think GitHub just sees the "fix" part.

Issue #3470, opened 2023-11-10
Associated types on traits


jfecher:
### Problem

We should have associated types on traits. These are partially implemented currently but are untested and require type checking changes, among others.

### Happy Case

Associated types such as the following work:
```rs
trait Foo {
    type Bar;
}

impl Foo for u32 {
    type Bar = u64;
}
```

### Would you like to submit a PR for this Issue?

No

PR #5739, opened 2024-08-16
feat: Explicit Associated Types & Constants

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/3470

(does not fix https://github.com/noir-lang/noir/issues/5544)

## Summary\*

Implements a basic form of associated types in Noir. 
- The main limitation of this work is that all associated types must be explicitly specified in trait constraints (e.g. `Foo: Iterator<Elem = i32>`) Instead of `Foo: Iterator`
  - Eliding these requires inserting additional implicit generics anywhere these constraints are used which I considered for this PR but opted to limit the size instead and leave it for later work.

## Additional Context

This is a large PR which has quite a few non-trivial changes so I'll try to summarize as best I can to help with reviews:
- Made the `Span` on `UnresolvedType` non-optional for better error messages.
- Added `GenericTypeArgs` struct to hold both "ordered" and "named" generics. Ordered generics are the ones we are used to where they are specified in order. Named generics are associated types where they are prefixed with `Name =` and may be given in any order in a generics list.
- Added a `Generic` trait to abstract over something that can be generic (struct, type alias, trait, etc) so that we can share the same method for resolving all generic things. This saved some code since we were duplicating checks on these before, and have much more now that we have to check named args as well.
- Added `TraitGenerics` as the resolved version of `GenericTypeArgs`. This is used in a couple places including the `impl Trait` type in the type system.
- Added associated types to the `lookup_trait_implementation` functions in the NodeInterner. These are just treated as normal generics more or less. When we allow these to be omitted in the future we should probably fill these with fresh type variables so that they are still present but always succeed unification.
- Stored associated types in trait impls in a separate map in the NodeInterner since they are needed to be referenced within the type signature of methods in the impl, which is before the impl as a whole finishes resolution.
- Resolved `Self::AssociatedType` paths. Not that https://github.com/noir-lang/noir/issues/5544 is not fixed, `Trait::AssociatedType` is still a panic.
  - Added `current_trait` and `current_trait_impl` fields to the elaborator to resolve these. Otherwise we don't know what `Self` refers to.
- Resolved `<Type as Trait>::Type` paths. These are called `AsTraitPath`s. Caveat: these are not allowed/resolved in arithmetic generics still. This prevents e.g. `let Size = <T as Serialize>::Size + <U as Serialize>::Size;`. This is more of an issue for when we allow implicit associated types. For now since they must be explicit, `<T as Serialize>::Size` must already be a named generic in scope so the example would look more like `let Size = TSize + USize;`, see the `serialize` test as an example.
- Additionally I had to make a couple more updates to arithmetic generics to get the `serialize` test working:
  - The existing rules were mostly good until monomorphization when we'd have to re-do the same trait constraints but now we'd know the actual end sizes (constants) for the constraints. This turned out to make it more difficult as we had to solve `4 = a + b` when serializing a pair where `a` and `b` are the serialized sizes of each item. In this case `1` and `3` were the correct answers, and since these refer to specific items in the pair, even an answer of `3` and `1` would be incorrect. I ended up solving `<constant> = a + b` to `<constant> - b = a` so that when we recur on the trait constraints for each item in the pair we'd get `T: Serialize<Size = 4 - b>` and `U: Serialize<Size = a>`. ~~I have a feeling the ordering could be wrong here. It type checks but requires type annotations on the result of `serialize`.~~. We could add more cases besides just subtraction but since this issue isn't directly related to associated generics though I'm considering this future work.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
Getting some test failures. I'll mark this as a draft and take it out when they're fixed.

jfecher:
Tests should be fixed now. I'd still like to add some frontend tests to this PR to test error cases but it should still be ready for review in the meantime.

compiler/noirc_frontend/src/ast/mod.rs
```diff
@@ -180,11 +216,38 @@ pub enum UnresolvedTypeExpression {
...
+        match self {
+            GenericTypeArg::Ordered(typ) => typ.fmt(f),
+            GenericTypeArg::Named(name, typ) => write!(f, "{name} = {typ}"),
+        }
+    }
+}
+
+impl std::fmt::Display for GenericTypeArgs {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if self.ordered_args.is_empty() && self.named_args.is_empty() {
```
asterite, line 239: Optional:

```suggestion
        if self.is_empty() {
```

compiler/noirc_frontend/src/elaborator/patterns.rs
```diff
@@ -643,10 +642,12 @@ impl<'context> Elaborator<'context> {
         if let ImplKind::TraitMethod(_, mut constraint, assumed) = ident.impl_kind {
             constraint.apply_bindings(&bindings);
             if assumed {
-                let trait_impl = TraitImplKind::Assumed {
-                    object_type: constraint.typ,
-                    trait_generics: constraint.trait_generics,
+                let trait_generics = TraitGenerics {
+                    ordered: constraint.trait_generics,
+                    named: constraint.associated_types,
```
asterite, line 647: Would it make sense for `TraitConstraint` to have a `TraitGenerics` field instead of these two? But I don't know if it's intentional that they are separate to have clearer names.

compiler/noirc_frontend/src/hir/type_check/generics.rs
```diff
@@ -0,0 +1,154 @@
...
+        interner.get_trait(*self).generics.clone()
+    }
+
+    fn accepts_named_type_args(&self) -> bool {
+        true
+    }
+
+    fn named_generics(&self, interner: &NodeInterner) -> Vec<ResolvedGeneric> {
+        interner.get_trait(*self).associated_types.clone()
+    }
```
asterite, line 51: Would it make sense for `named_generics` to return `Option`, with `None` meaning `accepts_named_type_args() == false`?

jfecher: I thought about it but I think this way is more self documenting. I can change it over if anyone feels strongly about it though.

test_programs/compile_success_empty/associated_types/src/main.nr
```diff
@@ -0,0 +1,28 @@
+trait Collection {
+    type Elem;
+
+    fn cget(self, index: u32) -> Option<Self::Elem>;
```
asterite, line 4: Two things I noticed:
- If you change `Self::Elem` to `Self::Elem2` you get an error saying "Could not resolve Self in path"
- If you change `Self::Elm` to `Elem`, it still compiles fine (but maybe it shouldn't, like in Rust?) That said, changing the one below does give an error

asterite (approved):
Looks great!

compiler/noirc_frontend/src/elaborator/patterns.rs
jfecher: Yes, I made the change and it lead me to find somewhere where associated types weren't being checked. Fixing that made type hints no longer required in serialize!

test_programs/compile_success_empty/associated_types/src/main.nr
jfecher: Yes, I think these are existing issues with how we handle `Self` in general. The second probably shouldn't compile... but seems okay to allow I think. It's a result of us treating traits as a pseudo-module so Elem is in scope to cget.

jfecher:
Not sure why the report gates diff keeps being cancelled.

TomAFrench:
Updating `rayon` doesn't fix this.

jfecher:
`rebuild.sh` works for me locally after limiting the maximum threads of parallel

jfecher:
Going to leave this PR open for another day and merge it late tomorrow unless anyone finds anything else that should be changed

TomAFrench:
I'm good with this being merged once we add something to `rebuild.sh` to limit parallelism so CI still works.

test_programs/compile_success_empty/serialize/src/main.nr
```diff
@@ -0,0 +1,59 @@
+trait Serialize {
+    let Size: u32;
+
+    // Note that Rust disallows referencing constants here!
+    fn serialize(self) -> [Field; Self::Size];
+}
+
+impl<A, B, let AS: u32, let BS: u32> Serialize for (A, B) where A: Serialize<Size = AS>, B: Serialize<Size = BS> {
+    // let Size = <A as Serialize>::Size + <B as Serialize>::Size;
```
vezenovm, line 9: Doesn't work?

test_programs/compile_success_empty/serialize/src/main.nr
```diff
@@ -0,0 +1,59 @@
...
+        }
+        for i in 0 .. b.len() {
+            array[i + a.len()] = b[i];
+        }
+        array
+    }
+}
+
+impl<T, let N: u32, let TS: u32> Serialize for [T; N] where T: Serialize<Size = TS> {
+    // let Size = <T as Serialize>::Size * N;
```
vezenovm, line 28: I see this commented out here as well? Should we leave a todo note?

vezenovm (approved):
Deferring to Ary and Tom's review as I did a quick review. Just a couple nits.

test_programs/compile_success_empty/serialize/src/main.nr
jfecher: All associated types must be explicit currently so this would have to be rewritten as
```rs
let Size = <A as Serialize<Size = AS>>::Size + <B as Serialize<Size = BS>>::Size;
```
which is somewhat pointless when you're already specifying AS and BS. This limitation is mentioned in the docs as well.

There's another issue in that the `<T as Trait>::Type` syntax isn't allowed in arithmetic generics currently. I plan to make an issue for this.

test_programs/compile_success_empty/serialize/src/main.nr
jfecher: I've made an issue for it. I'll probably split it into a separate test once it is implemented so that we can continue to check both options.

PR #5741, opened 2024-08-16
feat: LSP auto-import completion

asterite:
# Description

## Problem

Part of https://github.com/noir-lang/noir/issues/1577

## Summary

I said I wasn't going to send any more LSP features, but I had this one in the back of my head since I started working on autocompletion. I had the idea of how to do it in my head, but I wanted to first work on the easier stuff (regular autocompletion). It turns out auto-import wasn't that hard to implement after all!

lsp-autoimport

lsp-autoimport-nested

## Additional Context

In this PR every new imported item is added as a full path. Ideally, like in Rust Analyzer, we'd group common use prefixes, add to existing ones, etc. We can do that! But maybe it could be part of a follow-up PR, or maybe we could do it much later. Or maybe `nargo fmt` could handle this, so the auto-import doesn't do it, but once you save the file it's done. That's why I didn't spend time on that in this PR.

There's another thing I noticed while working on this PR: Rust Analyzer will offer matches that match **any** part of the name, not just the beginning. So, in Noir, if you type `merkle` it should probably offer `compute_merkle_root` as an auto-import completion, which is nice because maybe you are looking for stuff related to "merkle" but don't know the full name... but I didn't want to introduce that change in this PR (it also works for every other autocompletion). But, as always, this could be done later on.

## Documentation

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/mod.rs
+++ b/compiler/noirc_frontend/src/elaborator/mod.rs
@@ -1333,7 +1333,7 @@ impl<'context> Elaborator<'context> {
             .add_definition_location(ReferenceId::Global(global_id), Some(self.module_id()));
 
         if let Some(name) = name {
-            self.interner.register_name_for_autoimport(
+            self.interner.register_name_for_auto_import(
                 name,
                 ModuleDefId::GlobalId(global_id),
                 ItemVisibility::Public,
--- a/compiler/noirc_frontend/src/elaborator/mod.rs
+++ b/compiler/noirc_frontend/src/elaborator/mod.rs
@@ -1333,7 +1333,11 @@ impl<'context> Elaborator<'context> {
             .add_definition_location(ReferenceId::Global(global_id), Some(self.module_id()));
 
         if let Some(name) = name {
-            self.interner.register_name_for_autoimport(name, ModuleDefId::GlobalId(global_id));
+            self.interner.register_name_for_autoimport(
+                name,
+                ModuleDefId::GlobalId(global_id),
+                ItemVisibility::Public,
+            );
         }
 
         self.local_module = old_module;
--- a/compiler/noirc_frontend/src/elaborator/mod.rs
+++ b/compiler/noirc_frontend/src/elaborator/mod.rs
@@ -1297,7 +1297,7 @@ impl<'context> Elaborator<'context> {
         self.current_item = Some(DependencyId::Global(global_id));
         let let_stmt = global.stmt_def;
 
-        let name = if self.is_in_lsp_mode() {
+        let name = if self.interner.is_in_lsp_mode() {
             if let Pattern::Identifier(ident) = &let_stmt.pattern {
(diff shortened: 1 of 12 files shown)


compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1463,4 +1481,8 @@ impl<'context> Elaborator<'context> {
             _ => true,
         })
     }
+
+    fn is_in_lsp_mode(&self) -> bool {
+        self.interner.is_in_lsp_mode()
+    }
```
jfecher, line 1487:
```suggestion
```
I don't think there's a lot of need for this yet

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1297,6 +1297,16 @@ impl<'context> Elaborator<'context> {
...
 
+        let name = if self.is_in_lsp_mode() {
+            if let Pattern::Identifier(ident) = &let_stmt.pattern {
+                Some(ident.to_string())
+            } else {
+                None
+            }
+        } else {
+            None
+        };
```
jfecher, line 1308: Can you move this to a method on `Pattern`? You can probably reuse name_ident since we don't allow tuple globals.

compiler/noirc_frontend/src/hir/def_collector/dc_mod.rs
```diff
@@ -322,10 +333,18 @@ impl<'a> ModCollector<'a> {
...
+                context.def_interner.add_definition_location(
+                    ReferenceId::Struct(id),
+                    Some(ModuleId { krate, local_id: self.module_id }),
+                );
+
+                context.def_interner.register_name_for_auto_import(
+                    name.to_string(),
+                    ModuleDefId::TypeId(id),
+                    ItemVisibility::Public,
+                );
```
jfecher, line 346: Does add_definition_location imply register_name_for_auto_import? If so maybe there could be a helper.

asterite: I initially thought about doing that. But... we also do `add_definition_location` for impl methods and trait impl methods, struct members, and I think local variables too. So it's not always that they need to be done together.

That said, maybe they could be done together but the name for auto import could be an `Option` that's passed to `add_definition_location`?

jfecher: Sounds good to me, or optionally another wrapper that calls both functions that you only need to call in a couple cases.

asterite: I ended up introducing several wrappers, mainly because in one place a `ModuleDefId` is used and in another place a `ReferenceId` is used. That said, I think it looks much better than now (less code, and more readable).

jfecher (approved):

PR #5744, opened 2024-08-16
fix: Allow comptime code to use break without also being `unconstrained`

jfecher:
# Description

## Problem\*

Resolves 

## Summary\*

Allows `comptime` code to use `break` and `continue` without also marking the functions `unconstrained`. This is more important with the `unsafe { }` change since we don't want to have to wrap comptime code in unsafe blocks, nor do we want to mark unconstrained functions used in a comptime block as requiring `unsafe { }`.

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
👍

PR #5750, opened 2024-08-19
feat: add `Expr::as_array`, `Expr::as_repeated_element_array` and same for slice

asterite:
# Description

## Problem

Part of #5668

## Additional Context

I'm still not sure about the "repeated" name, and also about the similarity of the methods between array and slice (maybe we can distinguish array and slice with a boolean? not sure) but changing them should be easy, so let me know.

## Documentation

Check one:

# PR Checklist\*


test_programs/compile_success_empty/comptime_exp/src/main.nr
TomAFrench, line 1: Heads up that you could move these tests into the stdlib directly.

jfecher (approved):
I don't see another way of exposing this without having methods for both arrays and repeated element arrays. I think the separate versions for slices makes sense since they're separate types within Noir.

test_programs/compile_success_empty/comptime_exp/src/main.nr
asterite: Oh, nice! I'll do it, thanks.

asterite: How do I run the std tests? I tried compiling nargo, then running `nargo test` inside the stdlib directory but I'm getting a bunch of errors.

jfecher: They're only run when using `cargo t` iirc

TomAFrench: You can run `cargo test stdlib` in order to just run the tests in the stdlib.

asterite: I noticed that given that these are comptime tests, this comptime code runs whenever you compile any program... which is fine, but I guess if we start adding more and more comptime checks to the std it will make things slow (or at least slower).

Do you think it's worth moving these tests back to a test program? Also in a test program I can use `println` to debug things.

PR #5751, opened 2024-08-19
feat: user `super::` in LSP autocompletion if possible

asterite:
# Description

## Problem

Part of #1577

## Summary

We recently added `super::` to `use` so I thought it would be nice if the auto-importer used that if possible.

lsp-use-super

## Additional Context

This PR also includes a small refactor in the hover code because the code to format a module path exists there too (but the two are different enough that I didn't reuse them).

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/lsp/src/requests/completion.rs
+++ b/tooling/lsp/src/requests/completion.rs
@@ -1226,13 +1226,13 @@ mod completion_name_matches_tests {
 
     #[test]
     fn test_name_matches() {
-        // assert!(name_matches("foo", "foo"));
-        // assert!(name_matches("foo_bar", "bar"));
-        // assert!(name_matches("FooBar", "foo"));
-        // assert!(name_matches("FooBar", "bar"));
-        // assert!(name_matches("FooBar", "foo_bar"));
+        assert!(name_matches("foo", "foo"));
+        assert!(name_matches("foo_bar", "bar"));
+        assert!(name_matches("FooBar", "foo"));
+        assert!(name_matches("FooBar", "bar"));
+        assert!(name_matches("FooBar", "foo_bar"));
         assert!(name_matches("bar_baz", "bar_b"));
 
-        // assert!(!name_matches("foo_bar", "o_b"));
+        assert!(!name_matches("foo_bar", "o_b"));
     }
 }
--- a/tooling/lsp/src/requests/completion.rs
+++ b/tooling/lsp/src/requests/completion.rs
@@ -1185,18 +1185,14 @@ fn name_matches(name: &str, prefix: &str) -> bool {
 
     let mut last_index: i32 = -1;
     for prefix_part in prefix.split('_') {
-        if let Some(mut name_part_index) = name_parts
-            .iter()
-            .skip(
-                // Look past parts we already matched
-                if last_index >= 0 { last_index as usize + 1 } else { 0 },
-            )
-            .position(|name_part| name_part.starts_with(prefix_part))
+        // Look past parts we already matched
+        let offset = if last_index >= 0 { last_index as usize + 1 } else { 0 };
+
+        if let Some(mut name_part_index) =
+            name_parts.iter().skip(offset).position(|name_part| name_part.starts_with(prefix_part))
         {
-            if last_index >= 0 {
-                // Need to adjust the index if we skipped some segments
-                name_part_index += last_index as usize + 1;
-            }
+            // Need to adjust the index if we skipped some segments
+            name_part_index += offset;
 
             if last_index >= name_part_index as i32 {
                 return false;
--- a/tooling/lsp/src/requests/completion.rs
+++ b/tooling/lsp/src/requests/completion.rs
@@ -1185,9 +1185,19 @@ fn name_matches(name: &str, prefix: &str) -> bool {
 
     let mut last_index: i32 = -1;
     for prefix_part in prefix.split('_') {
-        if let Some(name_part_index) =
-            name_parts.iter().position(|name_part| name_part.starts_with(prefix_part))
+        if let Some(mut name_part_index) = name_parts
+            .iter()
+            .skip(
+                // Look past parts we already matched
+                if last_index >= 0 { last_index as usize + 1 } else { 0 },
+            )
+            .position(|name_part| name_part.starts_with(prefix_part))
         {
+            if last_index >= 0 {
+                // Need to adjust the index if we skipped some segments
+                name_part_index += last_index as usize + 1;
+            }
+
             if last_index >= name_part_index as i32 {
                 return false;
             }
@@ -1220,12 +1230,13 @@ mod completion_name_matches_tests {
 
     #[test]
     fn test_name_matches() {
-        assert!(name_matches("foo", "foo"));
-        assert!(name_matches("foo_bar", "bar"));
-        assert!(name_matches("FooBar", "foo"));
-        assert!(name_matches("FooBar", "bar"));
-        assert!(name_matches("FooBar", "foo_bar"));
+        // assert!(name_matches("foo", "foo"));
+        // assert!(name_matches("foo_bar", "bar"));
+        // assert!(name_matches("FooBar", "foo"));
+        // assert!(name_matches("FooBar", "bar"));
+        // assert!(name_matches("FooBar", "foo_bar"));
+        assert!(name_matches("bar_baz", "bar_b"));
 
-        assert!(!name_matches("foo_bar", "o_b"));
+        // assert!(!name_matches("foo_bar", "o_b"));
     }
 }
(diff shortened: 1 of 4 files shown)


tooling/lsp/src/requests/completion/auto_import.rs
```diff
@@ -111,9 +118,26 @@ impl<'a> NodeFinder<'a> {
     }
 }
 
-fn get_parent_module(interner: &NodeInterner, module_def_id: ModuleDefId) -> Option<&ModuleId> {
-    let reference_id = module_def_id_to_reference_id(module_def_id);
-    interner.reference_module(reference_id)
+fn get_parent_module(
```
TomAFrench, line 121: Can you add a comment here explaining why just using `reference_module` doesn't give what we want when passing a module's `ModuleDefId`? This is because `reference_modules` will point to the module containing the line `mod foo` rather than `foo.nr`, right?

tooling/lsp/src/requests/completion/auto_import.rs
```diff
@@ -130,66 +154,69 @@ fn module_def_id_to_reference_id(module_def_id: ModuleDefId) -> ReferenceId {
...
+    let crate_name = if is_relative {
+        None
+    } else {
+        match crate_id {
+            CrateId::Root(_) => {
+                if Some(module_id) == current_module_parent_id {
+                    Some("super".to_string())
+                } else {
+                    Some("crate".to_string())
+                }
```
TomAFrench, line 204: Is this ever going to trigger? If `Some(module_id) == current_module_parent_id` then we're going to have `is_relative == true` surely?

tooling/lsp/src/requests/completion/tests.rs
```diff
@@ -1573,4 +1573,29 @@ mod completion_tests {
...
+        assert_eq!(items.len(), 1);
+
+        let item = &items[0];
+        assert_eq!(item.label, "barbaz()");
+        assert_eq!(
+            item.label_details,
+            Some(CompletionItemLabelDetails {
+                detail: Some("(use super::barbaz)".to_string()),
+                description: Some("fn()".to_string())
+            })
```
TomAFrench, line 1598:
```suggestion
            pub fn bar_baz() {}

            mod tests {
                fn foo() {
                    bar_b>|<
                }
            }
        "#;
        let items = get_completions(src).await;
        assert_eq!(items.len(), 1);

        let item = &items[0];
        assert_eq!(item.label, "bar_baz()");
        assert_eq!(
            item.label_details,
            Some(CompletionItemLabelDetails {
                detail: Some("(use super::bar_baz)".to_string()),
                description: Some("fn()".to_string())
            })
```

Something to make spellcheck happy.

tooling/lsp/src/requests/completion/auto_import.rs
```diff
@@ -130,66 +154,69 @@ fn module_def_id_to_reference_id(module_def_id: ModuleDefId) -> ReferenceId {
...
+                .find(|dep| dep.crate_id == crate_id)
+                .map(|dep| format!("{}", dep.name)),
+            CrateId::Stdlib(_) => Some("std".to_string()),
+            CrateId::Dummy => None,
+        }
+    };
+
+    if let Some(crate_name) = &crate_name {
+        segments.push(crate_name);
+    };
```
TomAFrench, line 218: Can probably just merge this into the above if-statement.

TomAFrench (commented):
Looks good, just a couple nits

tooling/lsp/src/requests/completion/auto_import.rs
asterite: I tried it, but it says crate_name doesn't live long enough

TomAFrench: Ah ok.

tooling/lsp/src/requests/completion/auto_import.rs
asterite: Oh, it's because `reference_modules` doesn't track parents of modules, because that's in def maps. I added a small comment about it.

Maybe we can start tracking that in reference_modules too? But then the data would be duplicated.

tooling/lsp/src/requests/completion/auto_import.rs
asterite: Oh, it confused me too, but it can happen. For example if we have:

```rust
// We are in module foo
mod tests {
  // We want to import something that's in module foo
}
```

then from `tests` we get the parent, which is `foo`, and we want to get the full path of something in `foo` from `foo`. That can be `super`. So if there's `fn bar()` inside `foo` we end up using `super::bar`.

Maybe there's a way to make this function clearer?

tooling/lsp/src/requests/completion/auto_import.rs
asterite: Maybe the function should be called `parent_module_id_path`?

TomAFrench: Ok, I was a little confused as the `get_parent_module` was reading from there previously so I'm guessing we just had a bug where `get_parent_module` would return the module rather than it's parent when passed a `ModuleDefId::ModuleId(_)`?

asterite: Oh, I just realized that case I was handling can't happen, because `ModuleDefId::ModuleId` is handled differently.

tooling/lsp/src/requests/completion/auto_import.rs
```diff
@@ -127,69 +142,73 @@ fn module_def_id_to_reference_id(module_def_id: ModuleDefId) -> ReferenceId {
     }
 }
 
-/// Computes the path of `module_id` relative to `current_module_id`.
-/// If it's not relative, the full path is returned.
+/// Returns the path to reach an item inside `module_id` from inside `current_module_id`.
+/// Returns a relative path if possible.
 fn module_id_path(
-    module_id: &ModuleId,
+    module_id: ModuleId,
```
TomAFrench, line 148:
```suggestion
    target_module_id: ModuleId,
```

Can we rename this to make it more distinct from `current_module_id`?

asterite: Makes sense, done!

tooling/lsp/src/requests/completion.rs
```diff
@@ -1220,12 +1226,13 @@ mod completion_name_matches_tests {
...
-
-        assert!(!name_matches("foo_bar", "o_b"));
+        // assert!(name_matches("foo", "foo"));
+        // assert!(name_matches("foo_bar", "bar"));
+        // assert!(name_matches("FooBar", "foo"));
+        // assert!(name_matches("FooBar", "bar"));
+        // assert!(name_matches("FooBar", "foo_bar"));
+        assert!(name_matches("bar_baz", "bar_b"));
+
+        // assert!(!name_matches("foo_bar", "o_b"));
```
TomAFrench, line 1236: Looks like these need to be reenabled.

tooling/lsp/src/requests/completion/auto_import.rs
```diff
@@ -127,69 +142,69 @@ fn module_def_id_to_reference_id(module_def_id: ModuleDefId) -> ReferenceId {
...
+        None
+    } else {
+        match crate_id {
+            CrateId::Root(_) => Some("crate".to_string()),
+            CrateId::Stdlib(_) => Some("std".to_string()),
+            CrateId::Crate(_) => dependencies
+                .iter()
+                .find(|dep| dep.crate_id == crate_id)
+                .map(|dep| format!("{}", dep.name)),
+            CrateId::Dummy => None,
```
TomAFrench, line 200:
```suggestion
            CrateId::Dummy => unreachable!("ICE: A dummy CrateId should not be accessible"),
```

This branch should be unreachable afaik.

tooling/lsp/src/requests/completion/auto_import.rs
```diff
@@ -127,69 +142,69 @@ fn module_def_id_to_reference_id(module_def_id: ModuleDefId) -> ReferenceId {
...
+    let crate_name = if is_relative {
+        None
+    } else {
+        match crate_id {
+            CrateId::Root(_) => Some("crate".to_string()),
+            CrateId::Stdlib(_) => Some("std".to_string()),
+            CrateId::Crate(_) => dependencies
+                .iter()
+                .find(|dep| dep.crate_id == crate_id)
+                .map(|dep| format!("{}", dep.name)),
```
TomAFrench, line 199:
```suggestion
                .map(|dep| dep.name.to_string()),
```

TomAFrench (approved):

tooling/lsp/src/requests/completion/auto_import.rs
TomAFrench: Yeah, this is nice now.

tooling/lsp/src/requests/completion.rs
asterite: Oops, done!

PR #5753, opened 2024-08-19
feat: fault-tolerant parsing of `fn` and `impl`

asterite:
# Description

## Problem

Part of https://github.com/noir-lang/noir/issues/1577

## Summary

Autocompletion wasn't triggering in function parameter types, or in impl names. The reason is that when you start writing them the body will still be missing, and that discarded the entire fn or impl. In this PR we make that parse well, but produce an error (and create a fake empty body).

Now autocompletion works fine in these scenarios:

lsp-recover-fn

lsp-recover-impl

## Additional Context

I'm not sure about the many new "parser error" introduced, maybe there should be a generic "expected these tokens"? I think chumsky has one of those errors but it's kind of tied to the parsing logic. On the other hand adding new errors is simple, so maybe it's fine.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/parser/parser/traits.rs
+++ b/compiler/noirc_frontend/src/parser/parser/traits.rs
@@ -23,21 +23,12 @@ use crate::{
 use super::{generic_type_args, parse_type, primitives::ident};
 
 pub(super) fn trait_definition() -> impl NoirParser<TopLevelStatement> {
-    attributes()
-        .then_ignore(keyword(Keyword::Trait))
-        .then(ident())
-        .then(function::generics())
-        .then(where_clause())
-        .then(
-            just(Token::LeftBrace)
-                .ignore_then(trait_body())
-                .then_ignore(just(Token::RightBrace))
-                .or_not(),
-        )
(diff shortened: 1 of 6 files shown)


compiler/noirc_frontend/src/parser/parser/traits.rs
```diff
@@ -28,11 +28,25 @@ pub(super) fn trait_definition() -> impl NoirParser<TopLevelStatement> {
...
+
+            let items = if let Some(items) = items {
+                items
+            } else {
+                emit(ParserError::with_reason(
+                    ParserErrorReason::ExpectedLeftBracketOrWhereOrLeftBraceOrArrowAfterTraitName,
+                    span,
+                ));
+                vec![]
+            };
```
TomAFrench, line 48: Seems like we could move this validation up to the `trait_body().or_not`? This would keep the validation closer to what we're validating. We could then break this out this parser into a `let trait_body_or_error =` (feel free to use a better name) above the main chain of parsers.

asterite: Good idea, done!

jfecher (approved):

PR #5757, opened 2024-08-19
feat(perf): Additional function state in mem2reg to optimize across blocks

vezenovm:
# Description

## Problem\*

Resolves 

## Summary\*

Currently a draft. Now that we have a brillig gates diff I wanted to pull out individual changes each on their own to see their impact across all of our brillig tests. I have a feeling we may not see much with this change specifically but the new github action helps confirm that.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
@@ -66,6 +66,8 @@ mod block;
 
 use std::collections::{BTreeMap, BTreeSet};
 
+use fxhash::FxHashMap as HashMap;
+
 use crate::ssa::{
     ir::{
         basic_block::BasicBlockId,
@@ -111,6 +113,10 @@ struct PerFunctionContext<'f> {
     /// We avoid removing individual instructions as we go since removing elements
     /// from the middle of Vecs many times will be slower than a single call to `retain`.
     instructions_to_remove: BTreeSet<InstructionId>,
+
+    /// Track across all blocks the last load of a value.
+    /// If that value is not used in anymore loads we can remove the last store to that value.
+    last_loads: HashMap<ValueId, InstructionId>,
 }
 
 impl<'f> PerFunctionContext<'f> {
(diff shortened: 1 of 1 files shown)


TomAFrench:
image


:tada:

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -140,6 +147,14 @@ impl<'f> PerFunctionContext<'f> {
...
             self.analyze_block(block, references);
         }
+
+        for (_, block) in self.blocks.iter() {
+            for (value, store_instruction) in block.last_stores.iter() {
+                if self.last_loads.get(value).is_none() {
+                    self.instructions_to_remove.insert(*store_instruction);
+                }
+            }
+        }
```
jfecher, line 157: So if we never load from an address within a function we can remove all stores to it. This isn't true for reference parameters but we can't have those currently since we always inline all brillig functions into each other. Do we want to rely on this behavior into the future? May need a check for that case if not.

vezenovm: sounds good. I'll add a check.

vezenovm: done

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -140,6 +147,19 @@ impl<'f> PerFunctionContext<'f> {
...
             self.analyze_block(block, references);
         }
+
+        // If we never load from an address within a function we can remove all stores to that address.
+        // This rule does not apply to reference parameters, which we must also check for before removing these stores.
+        for (block_id, block) in self.blocks.iter() {
+            let block_params = self.inserter.function.dfg.block_parameters(*block_id);
+            for (value, store_instruction) in block.last_stores.iter() {
+                let is_reference_param = block_params.contains(value)
+                    && self.inserter.function.dfg.value_is_reference(*value);
```
jfecher, line 157:
```suggestion
```
You can get rid of this line since if we're loading from it it should be a reference

jfecher (approved):

PR #5760, opened 2024-08-19
fix(sha256): Perform compression per block and utilize ROM instead of RAM when setting up the message block

### LINKED ISSUE #5761: sha256 with variable length disagree with sha256 with constant length of the same length
Consider the program

```rust
fn main(input: [u8; 2], toggle: bool) {
    let size: Field = 1 + toggle as Field;
    assert(!toggle);

    let variable_sha = std::sha256::sha256_var(input, size as u64);
    let constant_sha = std::sha256::sha256_var(input, 1);

    assert_eq(variable_sha, constant_sha);

} 
```

We should expect that this program passes for the inputs below:
```
input = [0, 0]
toggle = false
```

This is as `size == 1` in this case and so the lengths being passed to the `sha256_var` calls are the same (and so the hashes are the same).

However this program fails as the two hashes produced are different.

vezenovm:
# Description

## Problem\*

Resolves #5761 

Resolution to performance blow-up found with sha256_var.

### Issue

The crux of the blow-up was the result of calling `sha256_compression` inside of the same loop where we build the message block. In the current `sha256_var` algorithm we are looping over the entire message and conditionally checking a msg byte pointer (the pointer into the msg block) to determine whether we have filled up a msg block and should run the sha compression. However, in a circuit this leads to us calling the compression opcode `N` times where `N` is the size of the message. 

We also were utilize RAM to build our message block when we do not have to do so. We can instead construct our block outside of the circuit and verify that the block has been constructed as we expect with assertion that just require ROM.

### Improvements

This PR produces a ~16x improvement in ACIR opcodes a >13x improvement in backend constraints for the following circuit:
```rust
fn main(foo: [u8; 95], toggle: bool) {
    let size: Field = 93 + toggle as Field * 2;
    let hash = std::sha256::sha256_var(foo, size as u64);
    println(f"{hash}");
}  
```
#### master
nargo info:
```
+---------+----------------------------+----------------------+--------------+-----------------+
| Package | Function                   | Expression Width     | ACIR Opcodes | Brillig Opcodes |
+---------+----------------------------+----------------------+--------------+-----------------+
| sha256  | main                       | Bounded { width: 4 } | 125852       | 243             |
+---------+----------------------------+----------------------+--------------+-----------------+
| sha256  | print_unconstrained        | N/A                  | N/A          | 230             |
+---------+----------------------------+----------------------+--------------+-----------------+
| sha256  | directive_integer_quotient | N/A                  | N/A          | 6               |
+---------+----------------------------+----------------------+--------------+-----------------+
| sha256  | directive_invert           | N/A                  | N/A          | 7               |
+---------+----------------------------+----------------------+--------------+-----------------+
```
bb gates:
```
{"functions": [
  {
        "acir_opcodes": 125852,
        "circuit_size": 597646,
```

#### This PR

Output of nargo info:
```
+----------------------------+----------------------------+----------------------+--------------+-----------------+
| Package                    | Function                   | Expression Width     | ACIR Opcodes | Brillig Opcodes |
+----------------------------+----------------------------+----------------------+--------------+-----------------+
| sha256_var_size_regression | main                       | Bounded { width: 4 } | 7768         | 1041            |
+----------------------------+----------------------------+----------------------+--------------+-----------------+
| sha256_var_size_regression | build_msg_block_iter       | N/A                  | N/A          | 299             |
+----------------------------+----------------------------+----------------------+--------------+-----------------+
| sha256_var_size_regression | pad_msg_block              | N/A                  | N/A          | 201             |
+----------------------------+----------------------------+----------------------+--------------+-----------------+
| sha256_var_size_regression | attach_len_to_msg_block    | N/A                  | N/A          | 298             |
+----------------------------+----------------------------+----------------------+--------------+-----------------+
| sha256_var_size_regression | print_unconstrained        | N/A                  | N/A          | 230             |
+----------------------------+----------------------------+----------------------+--------------+-----------------+
| sha256_var_size_regression | directive_integer_quotient | N/A                  | N/A          | 6               |
+----------------------------+----------------------------+----------------------+--------------+-----------------+
| sha256_var_size_regression | directive_invert           | N/A                  | N/A          | 7               |
+----------------------------+----------------------------+----------------------+--------------+-----------------+
```
bb gates output:
```
{"functions": [
  {
        "acir_opcodes": 7768,
        "circuit_size": 44663,
```

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
This solves #5761, can you add this as a test program as a regression test?

vezenovm:
> This solves #5761, can you add this as a test program as a regression test?

Done

vezenovm:
I guess we could remove `test_programs/execution_success/sha256_var_size_regression` if the test in #5761 is included. However, the larger size does show the blow-up more prominently.

noir_stdlib/src/sha256.nr
```diff
@@ -40,21 +40,21 @@ fn hash_final_block(msg_block: [u8; 64], mut state: [u32; 8]) -> [u8; 32] {
 }
 
 // Variable size SHA-256 hash
+#[no_predicates]
 pub fn sha256_var<let N: u32>(msg: [u8; N], message_size: u64) -> [u8; 32] {
     let mut msg_block: [u8; 64] = [0; 64];
     let mut h: [u32; 8] = [1779033703, 3144134277, 1013904242, 2773480762, 1359893119, 2600822924, 528734635, 1541459225]; // Intermediate hash, starting with the canonical initial value
     let mut i: u64 = 0; // Message byte pointer
+    assert(message_size <= N as u64);
```
TomAFrench, line 48: This breaks the rules on `no_predicates` functions right?

vezenovm: Ah yes you are right. I was going to put it in a wrapper but I am realizing do we even need the assert? `message_size` is only used at the end for being attached to the final block.

guipublic:
Do you know why this version increases the circuit size?
I would imagine that using ROM instead of RAM would be better?

This PR does not fix the underlying issue, it just hides it, so if it degrades the performance, I don't think we should merge it.

vezenovm:
> Do you know why this version increases the circuit size? I would imagine that using ROM instead of RAM would be better?
> 
> This PR does not fix the underlying issue, it just hides it, so if it degrades the performance, I don't think we should merge it.

For the smaller hashes it does look we are still getting an increase and that is why I have left this a draft. Looking at a larger hash though, as per the PR description we have a >15x ACIR opcode improvement and a >10x backend opcode improvement.

vezenovm:
> Do you know why this version increases the circuit size?

I am also checking the entire msg block after an unconstrained function. In some cases like a constant message size, we can optimize away certain RAM opcodes if we know they will never be hit. Now they have been replaced with assertions that always happen. This leads to an increase for small hashes but not for large ones. I think this is a reasonable trade-off as most meaningful use cases will require larger hashes not hashes that are smaller than a single message block.

guipublic:
The description gives numbers but no comparaison with the previous version, and the benchmarks do not look good.
15x improvement is definitively very good!

vezenovm:
> The description gives numbers but no comparaison with the previous version, and the benchmarks do not look good.

I have updated the description to include the exact numbers on master and provide comparisons.

TomAFrench (approved):

TomAFrench:
Holding off from merging as another review from @guipublic was requested.

noir_stdlib/src/hash/sha256.nr
```diff
@@ -17,82 +17,207 @@ pub fn digest<let N: u32>(msg: [u8; N]) -> [u8; 32] {
...
+    msg32
+}
+
+unconstrained fn build_msg_block_iter<let N: u32>(
+    msg: [u8; N],
+    message_size: u64,
+    mut msg_block: [u8; 64]
+) -> ([u8; 64], u64) {
+    let mut msg_byte_ptr: u64 = 0; // Message byte pointer
+    for k in 0..N {
```
guipublic, line 41: This function looks wrong to me:
if you need several blocks, you are going to overwrite the msg_block and only get the latest one?

noir_stdlib/src/hash/sha256.nr
```diff
@@ -17,82 +17,207 @@ pub fn digest<let N: u32>(msg: [u8; N]) -> [u8; 32] {
...
+    msg_byte_ptr = msg_byte_ptr + 1;
+    unsafe {
+        let (new_msg_block, new_msg_byte_ptr)= pad_msg_block(msg_block, msg_byte_ptr);
+        msg_block = new_msg_block;
+        if crate::runtime::is_unconstrained() {
+            msg_byte_ptr = new_msg_byte_ptr;
+        }
+    }
+
+    if !crate::runtime::is_unconstrained() {
```
guipublic, line 124: I wonder if this is safe?
You only check bytes 57 to 64 but you modify potentially the full block in the unsafe block.

noir_stdlib/src/hash/sha256.nr
```diff
@@ -17,82 +17,207 @@ pub fn digest<let N: u32>(msg: [u8; N]) -> [u8; 32] {
...
-            msg_block[i] = msg[k];
-            i = i + 1;
-            if i == 64 {
-                // Enough to hash block
-                h = crate::hash::sha256_compression(msg_u8_to_u32(msg_block), h);
+    let mut msg_byte_ptr = 0;
 
-                i = 0;
-            }
+    if num_blocks == 0 {
```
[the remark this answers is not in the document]
vezenovm: We will not enter the loop at all if `num_blocks` is 0.

noir_stdlib/src/hash/sha256.nr
vezenovm: Yeah we are going to rewrite it on each iteration, but I think I am doing it incorrectly here. Will push an update.

noir_stdlib/src/hash/sha256.nr
vezenovm: I could add a check of the whole block here, like I do after `attach_len_to_msg_block`.

noir_stdlib/src/hash/sha256.nr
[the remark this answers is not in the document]
guipublic: and then if  msg_byte_ptr <57 there is no call to sha256_compression() ?

noir_stdlib/src/hash/sha256.nr
vezenovm: The msg index now starts at `BLOCK_SIZE * i` where `i` is the current block we are iterating over.

Issue #5027, opened 2024-05-13
Conditional RAM writes cost many constraints


sirasistant:
### Aim

Using conditional ram writes to build a bounded vec:
```rust
global SIZE = 500;

fn main(fields: [Field; SIZE], to_keep: [bool; SIZE]) -> pub [Field; SIZE] {
    let mut bounded_vec: BoundedVec<Field, SIZE> = BoundedVec::new();

    for i in 0..SIZE {
        if to_keep[i] {
            bounded_vec.push(fields[i]);
        }
    }

    assert_eq(bounded_vec.len(), 0);
    bounded_vec.storage
}
```

### Expected Behavior

I expected conditional RAM writes to be in the same order of magnitude as regular ram writes:
```rust
global SIZE = 500;

fn main(mut fields: [Field; SIZE], indices: [u64; SIZE]) -> pub [Field; SIZE] {
    for i in 0..SIZE {
        fields[indices[i]] = i as Field;
    }

    fields
}
```

### Bug

Conditional ram writes like the first example cost around 2300 constraints/write. Unconditional ones like in the second one cost around 25, which is a huge difference.

### To Reproduce

1.
2.
3.
4.

### Project Impact

None

### Workaround

None

### Installation Method

None

### Would you like to submit a PR for this Issue?

None

TomAFrench:
https://github.com/noir-lang/noir/blob/9c6de4b25d318c6d211361dd62a112a9d2432c56/noir_stdlib/src/collections/bounded_vec.nr#L24

We should be able to just remove this assertion (or throw the error in an unconstrained block if we want to maintain the error message). The attempt to write past the end of the array would trigger the ACVM to fail.

TomAFrench:
This ends up being just a small contribution (~4000) constraints.

guipublic:
Did you really compare the number of constraints or just the number of acir opcodes?
Because 25 constraints for writing 500 times into an array of 500 elements does not seem right. I would expect ~2000 instead.

jfecher:
> One thing that jumps out is the check that we're not pushing past the capacity of the BoundedVec. We have enough information here that we should be able to rule that out at compile time.

[...]

Issue #5501, opened 2024-07-12
Push `ArrayGet` instructions backwards through `IfElse` instructions to avoid expensive array merges


TomAFrench:
Consider the program

```rust
fn main(array_a: [Field; 1000], array_b: [Field; 1000], is_b: bool, whatever_index: u32) {
    let mut array = array_a;
    if is_b {
    array = array_b;
    } 

    assert(array[whatever_index] == 0);
}

```

After flattening the control flow graph we end up with the instructions

```
// Here v0 and v1 are `array_a` and `array_b`
v7 = if v2 then v1 else if v5 then v0
v8 = array_get v7, index v3
```

This results in us merging two 1000 element arrays (performing 2000 array reads, merging these values into a new 1000 element array) and then performing a single read at `whatever_index`.

Ideally users would refactor their code to do single reads from the two arrays and then merge that single index, e.g.

```rust
let value = if is_b { array_b[whatever_index] } else { array_a[whatever_index] };
assert(value == 0);
```

We can perform this optimisation automatically however, we can recognise that we're performing an array read from the output of an `IfElse` instruction and pull out reads from the original arrays and generate a new `IfElse` for that single array element. We then get the SSA

```
v9 = array_get v0, index v3
v10 = array_get v1, index v3
v11 = if v2 then v9 else if v5 then v10
v7 = if v2 then v1 else if v5 then v0 // preserve original merged array in case used elsewhere.
```

Constant folding will prevent any extra gates being produced should `v7` survive DIE.

jfecher:
This strikes me as vaguely similar to what we do in `try_merge_only_changed_indices` here https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/flatten_cfg/value_merger.rs#L147. Only that instead of finding changed indices to shortcut the array merge we're finding only the indices that are actually checked later on and removing the merger of the rest of the array entirely.

PR #5762, opened 2024-08-19
feat(optimization): Perform array gets as early as possible

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/5027

(https://github.com/noir-lang/noir/issues/5501 is a separate issue)

## Summary\*

Due to other optimizations out of the user's control, programs can often end up with sequences resembling the following:
```
    v90208 = array_set v90198, index v90207, value u8 0
    v90210 = add v90203, u64 1
    v90211 = mul v90098, v90204
    v90212 = array_get v90208, index v90207
    v90213 = array_get v90198, index v90207
```
Where we have an `array_set` that cannot be made mutable because there is still an `array_get` later to the unmutated array.

To fix this, I've added another pass to move array gets as early as possible in the program. The earliest they can be moved is always just after their last ValueId dependency was inserted. This usually frees up any array_sets to be made mutable afterward.

## Additional Context

The program
```
fn main(foo: [u8; 95], toggle: bool) {
    let size: Field = 93 + toggle as Field * 2;
    let hash = sha256::sha256_var(foo, size as u64);
    println(f"{hash}");
}   
```
as well as the new test in `execution_success/array_get_optimization` now have 0 non-mutable array sets.

Not much change from most tests. I think this pattern was mostly a result of the array merge optimization to only merge changed indices since this is the only optimization that inserts array sets.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/move_array_gets.rs
```diff
@@ -0,0 +1,181 @@
...
+//!   - The array itself, the index to set, and the enable_side_effects variable.
+//!
+//! This optimization has two passes:
+//! 1. Traverse the function to find each array_get and remember its dependencies,
+//!    draining each instruction as we go.
+//! 2. Traverse the function re-inserting each instruction. When all dependencies
+//!    of an array_get have been inserted we can insert the corresponding array_get itself.
+//!    Since it can be expensive checking this for every instruction, this pass makes
+//!    the assumption that the last dependency will always have the highest ValueId.
+use fxhash::FxHashMap;
```
vezenovm, line 21:
```suggestion
use fxhash::FxHashMap as HashMap;
```
nit: just for consistency with what we do throughout the codebase

vezenovm (approved):

compiler/noirc_evaluator/src/ssa/opt/move_array_gets.rs
```diff
@@ -0,0 +1,181 @@
...
+                move_array_gets(state, &mut func.dfg, block, instructions);
+            }
+        }
+        self
+    }
+}
+
+#[derive(Default)]
+struct State {
+    array_gets: FxHashMap<ValueId, Vec<ArrayGet>>,
```
vezenovm, line 48:
```suggestion
    array_gets: HashMap<ValueId, Vec<ArrayGet>>,
```

vezenovm, line 48: And could we also add a comment here just explaining this map and that the key is the last dependency?

jfecher: Done

TomAFrench (changes requested):
Requesting changes just for some discussion. 

I'm a little concerned about this PR as it's similar to my "bubble up" optimisation which does the same operation but for constrain instructions. This resulted in situations where we ended up getting tests failing due to the noir code not being executed in the order which a user would expect.

I'm unsure how long we can maintain this guarantee however so we may need to open discussions on breaking this.

jfecher:
@TomAFrench I'm going to try changing the array merge optimization directly for now to delay the array sets there instead of moving the array gets later. This is the first approach I tried and thought it didn't work initially but I think now that the noir program was just getting cached from a previous version of the compiler. So I'll try again and hopefully we'll see similar improvements.

I'll also note that the conditional_1 test failure did not occur when I used a smaller constant value for that array merge optimization, like the previous 2. Not sure why that'd be causing the failure.

jfecher:
Running into some difficulties experimenting with this PR. Most notably is a difference in semantics between SSA Pre and Post-Flattening.

When optimizing array sets we optimize them if the array and index are known, but we never check the current enable side effects value to see if the array set may be disabled. This is fine Pre-flattening since control flow prevents referring to these directly if it is not what is wanted. Post flattening however it should be valid to optimize an array merge of the following:
```
b0(v1: [Field; 2], v2: u1):
  enable_side_effects v2
  v3 = array_set v1, index 0, value 1
  v4 = not v2
  enable_side_effects v4
  v5 = array_set v1, index 1, value 2
  enable_side_effects u1 1
  v6 = if v2 then v3 else if v4 then v5
```
to:
```
b0(v1: [Field; 2], v2: u1):
  enable_side_effects v2
  v3 = array_set v1, index 0, value 1
  v4 = not v2
  enable_side_effects v4
  v5 = array_set v3, index 1, value 2  // changed v1 to v3
  enable_side_effects u1 1
  // v6 = v5
```
Since the then branch should be disabled in the else case, we should expect `v3 = v1`, and can set to that directly instead. Likewise, since the else branch is disabled in the then case, we don't have to worry about `v5` changing the result of the merge in that case. This also lets us write mutably in both branches. But I believe this doesn't currently work when `v1` is constant because the else branch will be optimized to a constant `[_, 2]` even if it was not meant to be executed.

jfecher:
Going to pause work on this PR until https://github.com/noir-lang/noir/issues/5782 is fixed. I'm also running into it with the current approach in this PR which is fairly similar to what we currently use. The nested dynamic arrays test is failing with this panic currently.

jfecher:
Hmm I'm surprised changing the array merge optimization to use `[array1, array2][condition as u32]` is so much less efficient than merging every index of the arrays and creating a new array.

jfecher:
I'm closing this for now since the results are currently negative. We can retry similar optimizations in another PR later if wanted.

Issue #5767, opened 2024-08-20
Push `array_get`s through `array_set`s if they act on different constant indices.


TomAFrench:
Looking at the `conditional_1` program failure on #5762 I noticed that we're generating SSA with snippets of the form. 

```
 v11060 = array_set mut [v8829, v8910, v8988, v9062, v9142, v9220, v9300, v9376, v9456, v9530, v9610, v9686, v9766, v9842, v9920, v10000, v10076, v10158, v10236, v10310, v10390, v10466, v10544, v10622, v10700, v10778, v10860, v10934, v11014, v11009, v11006, v11003], index u32 2⁵, value v11000
v11061 = array_get v11060, index u32 0
v11063 = array_get v11060, index u32 1
v11065 = array_get v11060, index u32 2
v11067 = array_get v11060, index u32 3
v11093 = array_get v11060, index u32 2⁵
v11099 = array_get v11060, index u32 29
v11104 = array_get v11060, index u32 30
v11109 = array_get v11060, index u32 31
```

It's clear that `v11061 == v8829`, etc. as the `array_set` only affects the index 32. We could then optimize out all of these `array_gets` using this knowledge.

PR #5772, opened 2024-08-20
feat(optimization): Follow past `array_set`s when optimizing `array_get`s

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/5767

## Summary\*

Instead of checking if an array_get is from a constant array, we can follow array_sets to find a previous constant array. If the index was ever unknown we fail to optimize. If the index is known but changed by an array set we fail as well instead of optimizing to the set value. This is in case the side effect variables are different for the get & set.

## Additional Context

I set an arbitrary maximum of 5 array_sets to check before giving up and failing to find a constant array.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/instruction.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/instruction.rs
@@ -10,6 +10,7 @@ use acvm::{
     },
     FieldElement,
 };
+use fxhash::FxHashSet as HashSet;
 use fxhash::FxHasher;
 use iter_extended::vecmap;
 use noirc_frontend::hir_def::types::Type as HirType;
@@ -608,16 +609,11 @@ impl Instruction {
                 }
             }
             Instruction::ArrayGet { array, index } => {
-                let array = dfg.get_array_constant(*array);
-                let index = dfg.get_numeric_constant(*index);
-                if let (Some((array, _)), Some(index)) = (array, index) {
-                    let index =
-                        index.try_to_u32().expect("Expected array index to fit in u32") as usize;
-                    if index < array.len() {
-                        return SimplifiedTo(array[index]);
-                    }
+                if let Some(index) = dfg.get_numeric_constant(*index) {
+                    try_optimize_array_get_from_previous_set(dfg, *array, index)
+                } else {
+                    None
                 }
-                None
             }
             Instruction::ArraySet { array, index, value, .. } => {
                 let array = dfg.get_array_constant(*array);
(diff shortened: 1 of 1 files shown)


jfecher:
I added the optimization to use the value from an array_set if the index is known and it does not seem to break anything

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -744,6 +739,67 @@ impl Instruction {
...
+            Value::Array { array, typ: _ } => {
+                elements = Some(array.clone());
+                break;
+            }
+            _ => return SimplifyResult::None,
+        }
+    }
+
+    if let Some(array) = elements {
+        if let Some(index) = target_index.try_to_u64() {
```
vezenovm, line 793:
```suggestion
    if let (Some(array), Some(index)) = (elements, target_index.try_to_u64()) {
```
nit: looks like we could reduce the nesting here

vezenovm (approved):

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -744,6 +739,65 @@ impl Instruction {
...
+/// through several array sets. For each array set:
+/// - If the index is non-constant we fail the optimization since any index may be changed
+/// - If the index is constant and is our target index, we conservatively fail the optimization
+///   in case the array_set is disabled from a previous `enable_side_effects_if` and the array get
+///   was not.
+/// - Otherwise, we check the array value of the array set.
+///   - If the array value is constant, we use that array.
+///   - If the array value is from a previous array-set, we recur.
+fn try_optimize_array_get_from_previous_set(
+    dfg: &mut DataFlowGraph,
```
TomAFrench, line 758:
```suggestion
    dfg: &DataFlowGraph,
```

Loks like we don't mutate this.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -744,6 +739,65 @@ impl Instruction {
...
+                match &dfg[*instruction] {
+                    Instruction::ArraySet { array, index, value, .. } => {
+                        if let Some(constant) = dfg.get_numeric_constant(*index) {
+                            if constant == target_index {
+                                return SimplifyResult::SimplifiedTo(*value);
+                            }
+
+                            array_id = *array; // recur
+                        } else {
+                            return SimplifyResult::None;
```
TomAFrench, line 778: Haven't thought about whether this is desirable but we could also simplify this so that we perform the read on the earliest suitable array, i.e.

```
v1 = array_set [10, 11, 12], index v0, value: 5
v2 = array_set v1, index 2, value: 6
v3 = array_set v2, index 2, value: 7
v4 = array_get v3, index 1

// goes to

v1 = array_set [10, 11, 12], index v0, value: 5
v2 = array_set v1, index 2, value: 6
v3 = array_set v2, index 2, value: 7
v4 = array_get v1, index 1
```

TomAFrench: This may end up being counterproductive in some situations as can prevent mutation of `v1`.

jfecher: It could possibly be useful in combination with instruction de-duplication.

I think I'll merge this PR then make a separate PR to test this comment.

TomAFrench: Yep, agreed on tackling this separately.

PR #5775, opened 2024-08-20
fix: (LSP) only add cached files relevant to workspace

### LINKED ISSUE #5774: LSP hover information is sometimes incorrect
### Aim

LSP hover (and other features) sometimes works in a wrong way when you navigate to a different file and come back.

For example:

lsp-files-bug

### Expected Behavior

Hovering over the same thing should always give you the same result.

### Bug

Hover misbehaves.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #5774

## Summary

LSP caches open files so that features work relative to open, potentially unsaved, file buffers. Before analyzing a crate or processing a package, we add these open files to a FileManager. The problem was that we were adding all open files, regardless of whether they belonged to a package. Because we don't always call `check_crate` on LSP features (for example when you hover we collect a packages files but don't call `check_crate`) a packages file FileIDs would change if we added new files that didn't exist before in that package.

It's... kind of hard to explain. But it also makes sense to not add files that are not relevant to the package. I debugged it for hours until I found the reason. With this fix LSP should work really well now (it wasn't working that well in Aztec-Packages).

## Additional Context

None.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/lsp/src/lib.rs
+++ b/tooling/lsp/src/lib.rs
@@ -387,11 +387,19 @@ pub fn insert_all_files_for_workspace_into_file_manager(
     workspace: &Workspace,
     file_manager: &mut FileManager,
 ) {
+    let root_dir: String = workspace.root_dir.as_os_str().to_string_lossy().into();
+
     // First add files we cached: these have the source code of files that are modified
     // but not saved to disk yet, and we want to make sure all LSP features work well
     // according to these unsaved buffers, not what's saved on disk.
     for (path, source) in &state.input_files {
         let path = path.strip_prefix("file://").unwrap();
+
+        // Only add files relevant to the workspace
+        if !path.starts_with(&root_dir) {
+            continue;
+        }
+
         file_manager.add_file_with_source_canonical_path(Path::new(path), source.clone());
     }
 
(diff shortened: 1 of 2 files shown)


asterite:
Ah, nevermind, it seems this doesn't fix it.

asterite:
Oops, I thought it wasn't working because I commented the to record the bad behavior, then forgot I did that and noticed it was misbehaving again... but trying it with the fix I can't reproduce the bug anymore, so I'm sure it fixes it.

TomAFrench (approved):

TomAFrench:
Actually, how does this interact with path dependencies which aren't in the workspace?

asterite:
> Actually, how does this interact with path dependencies which aren't in the workspace?

Ah, it's a good point, it doesn't.

I'll think of another way to fix this...

asterite:
Okay, I found a way.

The main issue with how things were before this PR is that cached files were added to a FileManager **before** adding all of a workspace files. That meant that if new cached files showed up, they would shift the FileID numbers mapped to existing paths.

So, now we always add all of the files in the same order, except that we can pass a HashMap of source code overrides for some paths. That way FileIDs will be consistent no matter what cached files we have.

PR #5779, opened 2024-08-21
fix: do not use predicate for index in array operation, when the index is safe

### LINKED ISSUE #5761: sha256 with variable length disagree with sha256 with constant length of the same length

guipublic:
# Description

## Problem\*

Resolves #5761

## Summary\*

For array operation, we keep the index when it is safe, instead of using the predicate.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/acir_gen/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/acir_gen/mod.rs
@@ -1165,11 +1165,15 @@ impl<'a> Context<'a> {
         let index_var = self.convert_numeric_value(index, dfg)?;
         let index_var = self.get_flattened_index(&array_typ, array_id, index_var, dfg)?;
 
-        // predicate_index = index*predicate + (1-predicate)*offset
-        let offset = self.acir_context.add_constant(offset);
-        let sub = self.acir_context.sub_var(index_var, offset)?;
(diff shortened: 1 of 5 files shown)


jfecher (commented):
Do we know why this was breaking the example? I'm having trouble thinking of why this optimization could be bad.

guipublic:
> Do we know why this was breaking the example? I'm having trouble thinking of why this optimization could be bad.

Yes, I did not see why as well. I have investigated and the optimisation was replacing an array get with another one under predicate, leading to a different behaviour. Since the index is safe, wee do not need to handle the predicate in acir-gen, so finally we can keep the same behaviour and the optimisation as well.

TomAFrench (approved):

vezenovm (approved):

PR #5784, opened 2024-08-21
feat: Add `Expr::as_block` and `Expr::has_semicolon`

asterite:
# Description

## Problem

Part of #5668

## Summary

Both ExpressionKind and StatementKind will be exposed as `Expr` (as discussed with Jake). So `Expr::as_block` will return the block statements as a list of expressions. There's also the thing that some of these statements might be `Semi` (well, all except maybe the last one) and having to explicitly unwrap that with `as_semi` might be tedious or bug-prone, so instead that's unwrapped automatically when you call any of the `as_...` methods, but then there's an `has_semicolon` on `Expr` to know if it has a semicolon.

## Additional Context

There's just those two methods in this PR because I wanted to validate this approach. Next I'll add more `as_...` methods but for StatementKind.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
@@ -1084,24 +1084,8 @@ fn expr_as_unary_op(
 // fn as_has_semicolon(self) -> bool
 fn expr_has_semicolon(arguments: Vec<(Value, Location)>, location: Location) -> IResult<Value> {
     let self_argument = check_one_argument(arguments, location)?;
-    let mut expr_value = get_expr(self_argument)?;
-
-    loop {
-        match expr_value {
-            ExprValue::Expression(ExpressionKind::Parenthesized(expression)) => {
-                expr_value = ExprValue::Expression(expression.kind);
-            }
-            ExprValue::Statement(StatementKind::Expression(expression)) => {
-                expr_value = ExprValue::Expression(expression.kind);
-            }
-            ExprValue::Statement(StatementKind::Semi(..)) => {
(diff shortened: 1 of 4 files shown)


compiler/noirc_frontend/src/hir/comptime/value.rs
```diff
@@ -230,8 +245,9 @@ impl Value {
                     }
                 };
             }
-            Value::Expr(expr) => expr,
-            Value::Pointer(..)
+            Value::Expr(ExprValue::Expression(expr)) => expr,
+            Value::Expr(_)
```
jfecher, line 249: We can convert statements into expressions by wrapping them in a block

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -1031,14 +1073,37 @@ fn expr_as_unary_op(
...
+    let mut expr_value = get_expr(self_argument)?;
+
+    loop {
+        match expr_value {
+            ExprValue::Expression(ExpressionKind::Parenthesized(expression)) => {
+                expr_value = ExprValue::Expression(expression.kind);
+            }
+            ExprValue::Statement(StatementKind::Expression(expression)) => {
+                expr_value = ExprValue::Expression(expression.kind);
+            }
```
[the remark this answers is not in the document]
asterite: Done!

compiler/noirc_frontend/src/hir/comptime/value.rs
asterite: Done!

jfecher (approved):

PR #5785, opened 2024-08-21
feat(optimization): Redirect array gets to earlier arrays if possible

jfecher:
# Description

## Problem\*

Resolves comment https://github.com/noir-lang/noir/pull/5772#discussion_r1723641049

## Summary\*

This PR redirects `array_get`s to work on earlier arrays if they can (currently only block parameters.
This can potentially optimize code by allowing more array gets to be deduplicated but may also prevent optimizing `array_set`s to be mutable if older arrays are now being used more.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Is this ready for review?

jfecher:
Maybe. I'm still not sure if this optimization is worth it. Since it'd only have an effect on small programs I think due to the constant limit of how many instructions it can look past to. It also may lead to worse performance if that block parameter had a `array_set_mut` to it previously but can't anymore since now more array gets to it are inserted.

jfecher:
Closing this PR, it should be redundant now that we have instruction deduplication in general

PR #5788, opened 2024-08-21
fix: Handle multiple entry points for Brillig call stack resolution after metadata deduplication

vezenovm:
# Description

## Problem\*

Resolves 

Issue created by https://github.com/noir-lang/noir/pull/5696.

## Summary\*

After the creation of the brillig locations map in https://github.com/noir-lang/noir/pull/5696. we do the following to attach Brillig locations to our generated acir:
```rust
        if inserted_func_before {
            return;
        }

        for (brillig_index, call_stack) in generated_brillig.locations.iter() {
            self.brillig_locations.entry(brillig_function_index).or_default().insert(
                OpcodeLocation::Brillig {
                    acir_index: self.opcodes.len() - 1,
                    brillig_index: *brillig_index,
                },
                call_stack.clone(),
            );
        }
```
This inserts the ACIR index at the location the brillig call was pushed. However, this is only valid for the first Brillig call. Any later Brillig calls will also use this acir index to resolve against the debug metadata. 

We just need to avoid trying to match against this acir index as the Brillig call stacks do not require the acir index in order to be resolved against the debug metadata.

I also fixed the flamegraph in this PR.

For the following snippet where x == 0 causing a failure:
```rust
fn main(x: Field) {
    unsafe {
        assert(1 == conditional_wrapper(!x as bool));
        assert(1 == conditional_wrapper(x as bool));
    }
}
unconstrained fn conditional_wrapper(x: bool) -> Field {
    conditional(x)
}
unconstrained fn conditional(x: bool) -> Field {
    assert(x);
    1
}
```

Before:
Screenshot 2024-08-21 at 11 45 45 PM
After:
Screenshot 2024-08-21 at 11 47 17 PM

The flame graph also is operating as expected now. Here some example output from one of the Brillig entry point of the `brillig_calls` test:
Screenshot 2024-08-21 at 11 48 01 PM

## Additional Context

The assert messages still use the `Brillig` variant of `OpcodeLocation`. This is a breaking serialization change so I have opted for not changing that here and making a separate struct type for Brillig opcode locations.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/acvm-repo/acir/src/circuit/mod.rs
+++ b/acvm-repo/acir/src/circuit/mod.rs
@@ -161,7 +161,7 @@ pub enum OpcodeLocation {
 }
 
 impl OpcodeLocation {
-    pub fn reset_brillig_acir_index(self) -> Self {
+    pub fn reset_acir_index(self) -> Self {
         match self {
             OpcodeLocation::Brillig { brillig_index, .. } => {
                 OpcodeLocation::Brillig { acir_index: None, brillig_index }
--- a/acvm-repo/acir/src/circuit/mod.rs
+++ b/acvm-repo/acir/src/circuit/mod.rs
@@ -157,10 +157,16 @@ pub enum OpcodeLocation {
     // Resolved locations are generally specified when returning an error during execution.
     // The acir index is not actually needed to determine a Brillig location
     // as it can be resolved separately.
+    //
+    // We can not get rid of the acir index entirely just yet as this format is still
+    // used for resolving assert messages.
     Brillig { acir_index: Option<usize>, brillig_index: usize },
 }
 
 impl OpcodeLocation {
+    // Utility method to allow easily comparing a resolved Brillig location and unresolved Brillig location.
+    // This method is useful when fetching Brillig debug locations as this does not need an ACIR index,
+    // and just need the Brillig index.
     pub fn reset_acir_index(self) -> Self {
         match self {
             OpcodeLocation::Brillig { brillig_index, .. } => {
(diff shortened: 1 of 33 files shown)


michaeljklein (approved):
LGTM

acvm-repo/acir/src/circuit/mod.rs
```diff
@@ -153,9 +153,28 @@ pub struct ResolvedOpcodeLocation {
 /// map opcodes to debug information related to their context.
 pub enum OpcodeLocation {
     Acir(usize),
+    // We can not get rid of this enum field entirely just yet as this format is still
```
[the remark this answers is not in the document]
vezenovm: I made an issue [here](https://github.com/noir-lang/noir/issues/5792) and linked it in the code. In general we have been using `OpcodeLocation` in other places such as for resolving errors, so it used in other locations as well, but those are not blocking serialization changes. The thing actually blocking `acir_index` being removed from the Brillig variant are the assertion messages.

compiler/noirc_errors/src/debug_info.rs
```diff
@@ -99,7 +100,8 @@ pub struct DebugInfo {
     #[serde_as(as = "BTreeMap<DisplayFromStr, _>")]
     pub locations: BTreeMap<OpcodeLocation, Vec<Location>>,
     #[serde_as(as = "BTreeMap<_, BTreeMap<DisplayFromStr, _>>")]
```
TomAFrench, line 102: This should be no longer needed afaik.

TomAFrench:
LGTM, I think we have quite a bit of tech debt surrounding debug info so it would probably make sense to sit down, take stock of it and see if we can simplify how we handle it.

compiler/noirc_errors/src/debug_info.rs
vezenovm: Yeah I believe you are correct. I have removed it.

PR #5800, opened 2024-08-22
chore: Add comptime docs

jfecher:
# Description

## Summary\*

Most comptime features checked the "Documentation to be submitted in a separate PR" box. This is that PR.

## Additional Context

Draft because I haven't documented each function in `std::meta` yet and want to see the docs preview.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Looks good. Can we put the examples (especially the larger ones) into snippets to avoid them becoming stale?

jfecher:
@TomAFrench do we have a way to wrap tests like that? Since tests must be a single function currently.

I guess maybe they can be in their own private module in the stdlib?

michaeljklein:
@jfecher I would prefer to avoid adding them in the stdlib to avoid making `nargo` bigger. Why does each test need to be a single function?

docs/docs/noir/concepts/comptime.md
```diff
@@ -0,0 +1,358 @@
...
+- `comptime let` to define a variable whose value is evaluated at compile-time.
+- `comptime for` to run a for loop at compile-time. Syntax sugar for `comptime { for .. }`.
+
+## Scoping
+
+Note that while in a `comptime` context, any runtime variables _local to the current function_ are never visible.
+
+## Evaluating
+
+Evaluation rules of `comptime` follows the normal evaluation rules for other Noir code. There are a few things to note though:
```
michaeljklein, line 41:
```suggestion
Evaluation rules of `comptime` follows the normal unconstrained evaluation rules for other Noir code. There are a few things to note though:
```

docs/docs/noir/concepts/comptime.md
```diff
@@ -0,0 +1,358 @@
...
+
+- Certain built-in functions may not be available, although more may be added over time.
+- Evaluation order of global items is currently unspecified. For example, given the following two functions we can't guarantee
+which `println` will execute first. The ordering of the two printouts will be arbitrary, but should be stable across multiple compilations as long as the program is also unchanged.
```
michaeljklein, line 45:
```suggestion
which `println` will execute first. The ordering of the two printouts will be arbitrary, but should be stable across multiple compilations with the same `nargo` version as long as the program is also unchanged.
```
?

docs/docs/noir/concepts/comptime.md
```diff
@@ -0,0 +1,358 @@
...
+}
+
+fn two() {
+    comptime { println("two"); }
+}
+```
+
+- Since evaluation order is unspecified, care should be taken when using mutable globals so that they do not rely on a particular ordering.
+For example, using globals to generate unique ids should be fine but relying on certain ids always being produced (especially after edits to the program) should be avoided.
+- Although most ordering of globals is unspecified, two are:
```
michaeljklein, line 59: Seems to contradict `Evaluation order of global items is currently unspecified.` above

docs/docs/noir/concepts/comptime.md
```diff
@@ -0,0 +1,358 @@
...
+  - Dependencies of a crate will always be evaluated before the dependent crate.
+  - `comptime` items are evaluated before runtime items, however `comptime` functions can call runtime functions in which case
+    the runtime functions will be executed at compile-time.
+
+## Lowering
+
+When a `comptime` value is used in runtime code it must be lowered into a runtime value. This often means replacing the expression with the literal that it evaluated to. For example, the code:
```
michaeljklein, line 66: `This often means replacing the expression with the literal that it evaluated to.`
What else does it mean?

jfecher:
> Why does each test need to be a single function?

Because all `#[test]`s refer to a single function, while many of these examples have additional functions, structs, or traits. We could expose them all but then the test name would be exposed too. I wish we could have snippets in the docs be checked automatically without placing them in the stdlib as well.

docs/docs/noir/concepts/comptime.md
```diff
@@ -0,0 +1,358 @@
...
+# Comptime API
+
+Although `comptime`, `quote`, and unquoting provide a flexible base for writing macros,
+Noir's true metaprogramming ability comes from being able to interact with the compiler through a compile-time API.
+This API can be accessed through built-in functions in `std::meta` as well as on methods of several `comptime` types.
+
+The following is an incomplete list of some `comptime` types along with some useful methods on them.
+
+- `Quoted`: A token stream
+- `Type`: The type of a type in Noir
```
michaeljklein, line 258:
```suggestion
- `Type`: The type of a Noir type
```
?

`type of a type` sounds like a type kind to me

michaeljklein (changes requested):

docs/docs/noir/concepts/comptime.md
jfecher: Quoted code will be lowered into each token then parsed as an expression for example. Although I think I'm going to move to calling that macro insertion or something since you can also lower quoted code down to an actual `quote { ... }` expression.

docs/docs/noir/concepts/comptime.md
jfecher: Reworded

docs/docs/noir/concepts/comptime.md
jfecher: Reworded/clarified this and the next section once we talk about Quoted values

TomAFrench:
> do we have a way to wrap tests like that? Since tests must be a single function currently.

I'm not sure I follow. A snippet can contain arbitrary content so we just need to place the examples into a program which will be checked that it compiles successfully and then we can pull out the section of code we want to display. An example of this is the `// docs:start:new_example` snippet in `noir_test_success/bounded_vec`

michaeljklein (approved):

jfecher:
Ok, I've moved as many snippets as I can through docs. The ones I did not move either:
- Used `println` internally (execution ordering example)
- Had text around the examples explaining specific items in the examples (`x` and `y`) which could implicitly break if the example were changed elsewhere without updating the docs (unquote example).

PR #5811, opened 2024-08-23
chore(perf): Simplify poseidon2 algorithm 

vezenovm:
# Description

## Problem\*

Resolves 

Optimizations found while looking exploring other Brillig opts.

## Summary\*

There are a couple optimizations here:

1. I noticed that we loop over the cache and do some resetting inside of `squeeze` of `Poseidon2`. However, we our `Hasher` always creates a fresh Poseidon2 object so it seems unnecessary to reset the cache in this way. In Brillig this leads to an extra loop that is essentially unused and blows up the code size of any programs using poseidon in an unconstrained environment.
2. We were writing into a `result` array and returning it from `perform_duplex`. This result was unused inside of `absorb` and we can directly access `self.state` inside of `squeeze`. I no longer return anything from `perform_duplex`.

## Documentation\*

Check one:

# PR Checklist\*


michaeljklein (approved):
LGTM

I wondered whether this could cause issues if someone manually cloned the `Poseidon2Hasher` struct to provide both the hash of a prefix and the hash of the whole collection efficiently, i.e. `hash([a, b, c])` and `hash([a, b, c, d, ..])`, but the following test both executes w/o error and prints the same results as before:

```rust
use std::hash::poseidon2;

fn main() {
    let mut hasher_1 = poseidon2::Poseidon2Hasher::default();
    for _ in 0..128 {
        hasher_1.write(42);
    }
    hasher_1.write(1);

    let mut hasher_2 = poseidon2::Poseidon2Hasher { _state: hasher_1._state };
    for _ in 0..128 {
        hasher_2.write(42);
    }
    let result_1 = hasher_1.finish();
    hasher_2.write(2);
    let result_2 = hasher_2.finish();

    let mut hasher_3 = poseidon2::Poseidon2Hasher::default();
    for _ in 0..128 {
        hasher_3.write(42);
    }
    hasher_3.write(1);

    for _ in 0..128 {
        hasher_3.write(42);
    }
    hasher_3.write(2);

    let result_3 = hasher_3.finish();

    assert(result_2 == result_3);
    println(result_1);
    println(result_2);
}
```

noir_stdlib/src/hash/poseidon2.nr
```diff
@@ -38,61 +38,30 @@ impl Poseidon2 {
...
             self.cache[0] = input;
             self.cache_size = 1;
-        } else if (!self.squeeze_mode) & (self.cache_size != RATE) {
+        } else {
             // If we're absorbing, and the cache is not full, add the input into the cache
             self.cache[self.cache_size] = input;
             self.cache_size += 1;
-        } else if self.squeeze_mode {
-            // If we're in squeeze mode, switch to absorb mode and add the input into the cache.
-            // N.B. I don't think this code path can be reached?!
```
vezenovm, line 61: This note I believe is correct. I don't think this code path can be reached, and we should remove it as it adds to the code size. I have kept an assert that we are in squeeze mode for the absorb function. I think we could even have a version of this hash impl with an "unsafe absorb" and "unsafe squeeze" where we do not have this assert and the developer is aware of how they should absorb and squeeze. I didn't do it in this PR as the cost of this check is quite minimal.

guipublic (approved):
LGTM
I just wonder why the code was like this, as if it was done to work also in other 'mode'

vezenovm:
> as if it was done to work also in other 'mode'

Looks like it. But we never use the other "mode" so I figured we should delete it. If we want it back, we can bring it back.

Issue #5806, opened 2024-08-23
Comptime blocks in 


TomAFrench:
> I noticed that given that these are comptime tests, this comptime code runs whenever you compile any program... which is fine, but I guess if we start adding more and more comptime checks to the std it will make things slow (or at least slower).
>
> Do you think it's worth moving these tests back to a test program? Also in a test program I can use `println` to debug things.

@asterite noted in #5730 that placing comptime blocks in the stdlib for testing results in us having to interpret these comptime blocks on every compilation. Considering these are test functions this is completely wasted effort as they should be ignored if we're not actively running tests on the stdlib. This will affect any library which uses comptime blocks in tests.

One solution to this would be to have some kind of `#[cfg(test)]` block similar to rust which would allow us to ignore the tests entirely when not actively running them.

Performing analysis on which comptime blocks are reachable and only interpreting those doesn't seem easily doable as we could use multiple comptime blocks to stitch together another call to a comptime function so it doesn't seem like an option in the short term.

PR #5822, opened 2024-08-26
feat: add Expr::as_method_call

asterite:
# Description

## Problem

Part of #5668

## Summary

Also moves the tests to a test program to avoid evaluating that `comptime` code on each compilation (see #5806)

## Additional Context

The rest of the comptime methods involve `Path` or `Pattern` and I'd like to handle those (or one of those) in a separate PR.

## Documentation

Check one:

# PR Checklist


test_programs/compile_success_empty/comptime_expr/src/main.nr
```diff
@@ -0,0 +1,313 @@
+mod tests {
+    use std::meta::op::UnaryOp;
+    use std::meta::op::BinaryOp;
+
+    #[test]
+    fn test_expr_as_array() {
```
jfecher, line 6: These are all tests so maybe `noir_test_success` would be a better place for them?
All of them are run at comptime regardless though so it doesn't actually matter much.

jfecher (approved):

PR #5825, opened 2024-08-26
feat: add `FunctionDef::body`

asterite:
# Description

## Problem

Part of #5668

## Summary

Also allows `quote { ... }.as_expr()` to work with statements and L-values. Then adds an example that injects a `_context: Context` parameter to functions annotated with `#[aztec]`, and to calls inside those functions (any call for now). Also adds an incomplete `Expr::map` to help doing this. 

Oh, and allows unquoting `Value::Expr`, which wasn't possible before.

## Additional Context

A lot is remaining, but I thought this was a good cutting point. Next I might work on #5828 because debugging comptime code without it is tricky.

I didn't add docs for `Expr::map` yet as it's still not fully functional. I'll add them once it works for in all cases.

## Documentation

Check one:

# PR Checklist


noir_stdlib/src/meta/expr.nr
```diff
@@ -110,4 +110,51 @@ impl Expr {
...
+
+fn map_block<Env>(expr: Expr, f: fn[Env](Expr) -> Option<Expr>) -> Option<Expr> {
+    let exprs = expr.as_block();
+    if exprs.is_some() {
+        let exprs = exprs.unwrap();
+        let exprs = exprs.map(|expr: Expr| expr.map(f));
+        Option::some(new_block(exprs))
+    } else {
+        Option::none()
+    }
```
jfecher, line 132:
```suggestion
    expr.as_block().map(|exprs| {
        let exprs = exprs.map(|expr: Expr| expr.map(f));
        new_block(exprs)
    })
```
Can do something similar for other methods.

You can't do the following for `if` since it has an optional else, but for many other variants you could feed them to one function `map_helper(exprs: [Expr], f) -> [Expr]` which takes a slice so you can just bundle up differing number of exprs and map them all with that. You'd still have to unpack the returned slice manually though.

noir_stdlib/src/meta/expr.nr
```diff
@@ -110,4 +110,51 @@ impl Expr {
...
+        let alternative = alternative.and_then(f);
+        Option::some(new_if(condition, consequence, alternative))
+    } else {
+        Option::none()
+    }
+}
+
+fn new_block(exprs: [Expr]) -> Expr {
+    let exprs = exprs.map(|expr| quote { $expr }).join(quote { });
+    quote { { $exprs }}.as_expr().unwrap()
```
jfecher, line 150: Hmm, I wonder if mapping these to another Expr could remove the semicolons since we implicitly remove those when matching on Exprs.

test_programs/compile_success_empty/aztec_macro_add_context/Nargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "aztec_macro_add_context"
```
jfecher, line 2: Lets rename this test to `inject_context_attribute`

noir_stdlib/src/meta/expr.nr
```diff
@@ -110,4 +110,51 @@ impl Expr {
     // docs:start:is_continue
     fn is_continue(self) -> bool {}
     // docs:end:is_continue
+
+    fn map<Env>(self, f: fn[Env](Expr) -> Option<Expr>) -> Expr {
+        let result = f(self);
+        let result = result.or_else(|| map_if(self, f));
+        let result = result.or_else(|| map_block(self, f));
+        // TODO: check the remaining possibilities...
```
michaeljklein, line 118: Remaining TODO

Also, please add docs for these new functions. I'd guess that `map` here is something like "attempt to run the function on the whole expression and then each option for sub-expression in [??] order" so perhaps `try_map_subexpression` or a similar name would avoid confusion with collection-style map functions?

Also, the ordering of the sub-expressions seems a bit error prone to me:
- As defined, adding support for different expressions in `f` could change which sub-expression `f` gets executed on
  + I.e., the ordering of which case is tried first seems both significant and unpredictable.
- My favorite alternative is to run each case, pick the first successful case as you're doing here, and then assert that exactly one case succeeded.

test_programs/compile_success_empty/aztec_macro_add_context/src/main.nr
```diff
@@ -0,0 +1,63 @@
+struct Context {
+    value: Field,
+}
+
+#[aztec]
```
jfecher, line 5: And this macro can be `#[inject_context]`

test_programs/compile_success_empty/aztec_macro_add_context/src/main.nr
```diff
@@ -0,0 +1,63 @@
...
+    f.set_body(body);
+}
+
+fn mapping_function(expr: Expr) -> Option<Expr> {
+    let func_call = expr.as_function_call();
+    if func_call.is_some() {
+        // TODO: only do this if the function is annotated with #[aztec]
+        let (name, arguments) = func_call.unwrap();
+        let arguments = arguments.map(|arg: Expr| arg.map(mapping_function));
+        let arguments = arguments.push_back(quote { _context }.as_expr().unwrap());
```
jfecher, line 51:
```suggestion
        let arguments = arguments.push_front(quote { _context }.as_expr().unwrap());
```

test_programs/compile_success_empty/aztec_macro_add_context/src/main.nr
```diff
@@ -0,0 +1,63 @@
...
+#[aztec]
+fn qux(x: Field) -> Field {
+    // 21 * 2 => 42
+    x * 2
+}
+
+fn aztec(f: FunctionDefinition) {
+    // Add a `_context: Context` parameter to the function
+    let parameters = f.parameters();
+    let parameters = parameters.push_back((quote { _context }, quote { Context }.as_type()));
```
jfecher, line 30:
```suggestion
    let parameters = parameters.push_front((quote { _context }, quote { Context }.as_type()));
```
Nitpick: let's arbitrarily have the context first so it always has position 0 in every call

jfecher (changes requested):

noir_stdlib/src/meta/expr.nr
```diff
@@ -110,4 +110,51 @@ impl Expr {
...
+        Option::none()
+    }
+}
+
+fn new_block(exprs: [Expr]) -> Expr {
+    let exprs = exprs.map(|expr| quote { $expr }).join(quote { });
+    quote { { $exprs }}.as_expr().unwrap()
+}
+
+fn new_if(condition: Expr, consequence: Expr, alternative: Option<Expr>) -> Expr {
```
michaeljklein, line 153: Please add tests for the new stdlib functions

michaeljklein (changes requested):

noir_stdlib/src/meta/expr.nr
asterite: I _think_ there's no issue in how it's currently done:
- First we call `f(self)`. If that returns `Some` we'll return that (all of the other `or_else` won't call their lambda)
- Otherwise it's none so we check which expression we are. Now here only one of the next `or_else` lambdas will be called, because the expression can only be one of `if`, `array`, `block`, etc. And for the one that it returns `Some`, that'll be the returned value.

So... I think the order in which we evaluate these doesn't matter?

asterite: I added some docs... I'm still not sure about the docs for `map`. Maybe @jfecher could come up with a better description? (he had the idea of this method and its signature)

compiler/noirc_frontend/src/ast/statement.rs
```diff
@@ -97,6 +101,9 @@ impl StatementKind {
             // A semicolon on a for loop is optional and does nothing
             StatementKind::For(_) => self,
 
+            // No semicolon needed for a resolved statement
+            StatementKind::Interned(_) => self,
```
jfecher, line 105: Is this true? What if we have an interned `StatementKind::Expression`?

compiler/noirc_frontend/src/ast/statement.rs
```diff
@@ -612,17 +620,53 @@ impl LValue {
...
+                    crate::ast::UnaryOp::Dereference { implicitly_added: false }
+                ) {
+                    LValue::Dereference(Box::new(LValue::from_expression(prefix.rhs)), span)
+                } else {
+                    panic!("Called LValue::from_expression with an invalid prefix operator")
+                }
+            }
+            ExpressionKind::Interned(id) => LValue::Interned(id, span),
+            _ => {
+                panic!("Called LValue::from_expression with an invalid expression")
```
jfecher, line 658: I imagine this can happen as a result of macros inserting into an lvalue position, e.g. `quote { $foo = bar; }`. So I think it should be a user-facing error

compiler/noirc_frontend/src/hir/comptime/value.rs
```diff
@@ -609,3 +637,204 @@ fn display_trait_constraint(interner: &NodeInterner, trait_constraint: &TraitCon
...
     format!("{}: {}{}", trait_constraint.typ, trait_.name, trait_constraint.trait_generics)
 }
+
+// Returns a new Expression where all Interned and Resolved expressions have been turned into non-interned ExpressionKind.
+fn resolve_expression(interner: &NodeInterner, expr: Expression) -> Expression {
+    Expression { kind: resolve_expression_kind(interner, expr.kind), span: expr.span }
+}
+
+// Returns a new ExpressionKind where all Interned and Resolved expressions have been turned into non-interned ExpressionKind.
+fn resolve_expression_kind(interner: &NodeInterner, expr: ExpressionKind) -> ExpressionKind {
```
jfecher, line 647: These are only used for display, right? If so, can we use `to_display_ast` instead? Or does that not cover some cases?

noir_stdlib/src/meta/expr.nr
```diff
@@ -110,4 +110,305 @@ impl Expr {
     // docs:start:is_continue
     fn is_continue(self) -> bool {}
     // docs:end:is_continue
+
+    // docs:start:map
+    fn map<Env>(self, f: fn[Env](Expr) -> Option<Expr>) -> Expr {
+        // docs:end:map
+        let result = f(self);
+        let result = result.or_else(|| map_array(self, f));
+        let result = result.or_else(|| map_assign(self, f));
```
jfecher: I thought of an issue with our current design of this function.
The use case is to provide an easy way for users to traverse the Ast and modify the nodes they want to. For example, adding an argument to each function call.

However, since we don't apply the function recursively after we already return `Some` on a node, we can run into an issue where `foo(bar(baz))` would turn into `foo(ctx, bar(baz))` instead of `foo(ctx, bar(ctx, baz))`. I think the easiest way to fix this issue is to apply `f` _after_ we recur on the subexpressions instead of before - and to apply it unconditionally.

E.g.
```rs
let result = map_array(self, f);
let result = result.or_else(|| map_assign(self, f));
...
let result = result.or_else(|| map_unsafe(self, f));

if result.is_some() {
    let modified = f(result.unwrap_unchecked());
    if modified.is_some() {
        modified
    } else {
        result
    }
} else {
    self
}
// could also be:
// result.and_then(|expr| f(expr).unwrap_or(expr)).unwrap_or(self)
```

noir_stdlib/src/meta/expr.nr
```diff
@@ -110,4 +110,305 @@ impl Expr {
...
+    quote { [$exprs]}.as_expr().unwrap()
+}
+
+fn new_assign(lhs: Expr, rhs: Expr) -> Expr {
+    quote { $lhs = $rhs }.as_expr().unwrap()
+}
+
+fn new_binary_op(lhs: Expr, op: BinaryOp, rhs: Expr) -> Expr {
+    let op = op.quoted();
+    quote { $lhs $op $rhs }.as_expr().unwrap()
```
jfecher, line 333: Probably want to parenthesize this. Otherwise these could combine in odd ways with precedence.

noir_stdlib/src/meta/expr.nr
```diff
@@ -110,4 +110,305 @@ impl Expr {
...
+}
+
+fn new_block(exprs: [Expr]) -> Expr {
+    let exprs = join_expressions(exprs, quote { ; });
+    quote { { $exprs }}.as_expr().unwrap()
+}
+
+fn new_cast(expr: Expr, typ: UnresolvedType) -> Expr {
+    quote { $expr as $typ }.as_expr().unwrap()
```
jfecher, line 342: Parenthesize

noir_stdlib/src/meta/expr.nr
```diff
@@ -110,4 +110,305 @@ impl Expr {
...
+}
+
+fn new_tuple(exprs: [Expr]) -> Expr {
+    let exprs = join_expressions(exprs, quote { , });
+    quote { ($exprs) }.as_expr().unwrap()
+}
+
+fn new_unary_op(op: UnaryOp, rhs: Expr) -> Expr {
+    let op = op.quoted();
+    quote { $op $rhs }.as_expr().unwrap()
```
jfecher, line 404: Parenthesize

noir_stdlib/src/meta/op.nr
```diff
@@ -26,6 +26,20 @@ impl UnaryOp {
...
+    pub fn quoted(self) -> Quoted {
+        // docs:end:unary_quoted
+        if self.is_minus() {
+            quote { - }
+        } else if self.is_not() {
+            quote { ! }
+        } else if self.is_mutable_reference() {
+            quote { &mut }
+        } else {
+            quote { * }
```
jfecher, line 40:
```suggestion
        } else if self.is_dereference() {
            quote { * }
        } else {
            assert(false, "unhandled unary operator");
        }
```
This makes it more future-proof if we ever add unary operators

noir_stdlib/src/meta/op.nr
```diff
@@ -128,5 +142,43 @@ impl BinaryOp {
...
+        } else if self.is_or() {
+            quote { | }
+        } else if self.is_xor() {
+            quote { ^ }
+        } else if self.is_shift_right() {
+            quote { >> }
+        } else if self.is_shift_left() {
+            quote { << }
+        } else {
+            quote { % }
```
jfecher, line 180: Let's add an `assert(false, _)` default case here too

test_programs/compile_success_empty/inject_context_attribute/src/main.nr
```diff
@@ -0,0 +1,55 @@
...
+
+fn inject_context(f: FunctionDefinition) {
+    // Add a `_context: Context` parameter to the function
+    let parameters = f.parameters();
+    let parameters = parameters.push_front((quote { _context }, quote { Context }.as_type()));
+    f.set_parameters(parameters);
+
+    // Create a new body where every function call has `_context` added to the list of arguments.
+    let body = f.body().map(mapping_function).quoted();
+    f.set_body(body);
```
jfecher, line 35: Hmm, maybe it'd make more sense for `set_body` to accept a `Expr` directly

test_programs/compile_success_empty/inject_context_attribute/src/main.nr
```diff
@@ -0,0 +1,55 @@
...
+    let func_call = expr.as_function_call();
+    if func_call.is_some() {
+        let (name, arguments) = func_call.unwrap();
+        let arguments = arguments.map(|arg: Expr| arg.map(mapping_function));
+        let arguments = arguments.push_front(quote { _context }.as_expr().unwrap());
+        let arguments = arguments.map(|arg: Expr| arg.quoted()).join(quote { , });
+        Option::some(quote { $name($arguments) }.as_expr().unwrap())
+    } else {
+        Option::none()
+    }
```
jfecher, line 48:
```suggestion
    expr.as_function_call().map(|expr| {
        let (name, arguments) = expr;
        let arguments = arguments.push_front(quote { _context }.as_expr().unwrap());
        let arguments = arguments.map(|arg: Expr| arg.quoted()).join(quote { , });
        quote { $name($arguments) }.as_expr().unwrap()
    })
```
Shouldn't have to recursively apply the mapping function here. It was only necessary from a design mistake on my part for the `map` function. Once we address the previous comment on that it should be unnecessary to explicitly recur here.

test_programs/noir_test_success/comptime_expr/src/main.nr
```diff
@@ -308,6 +554,16 @@ mod tests {
...
     }
+
+    comptime fn times_two(expr: Expr) -> Option<Expr> {
+        expr.as_integer().and_then(
+            |integer: (Field, bool)| {
+          let (value, _) = integer;
+            let value = value * 2;
+            quote { $value }.as_expr()
+        }
+        )
```
jfecher, line 565: Thank you `nargo fmt` :sweat_smile:

docs/docs/noir/standard_library/meta/expr.md
```diff
@@ -161,3 +161,21 @@ comptime {
...
+
+#include_code map noir_stdlib/src/meta/expr.nr rust
+
+Applies a mapping function to this expression or to all of its sub-expressions.
+
+If `f` returns `Option::some`, that's the return value of this method. Otherwise,
+`f` will be recursively called on each of this expression sub-expressions.
+
+For example, calling `map` on `[1, 2, [3, 4]]` with an `f` that returns `Option::some`
+for expressions that are integers, doubling them, would return `[2, 4, [6, 8]]`.
```
jfecher:
```suggestion
Applies a mapping function to this expression and to all of its sub-expressions.

`f` will be applied to each sub-expression first, then applied to the expression itself.
This happens recursively for every expression within `self`.

For example, calling `map` on `(&[1], &[2, 3])` with an `f` that returns `Option::some`
for expressions that are integers, doubling them, would return `(&[2], &[4, 6])`.
```
- Slight wording change to fit the new suggestion I made for the method
- Changed the example a bit to be a valid value

jfecher:
> [map -> mutate](https://github.com/noir-lang/noir/pull/5825/commits/3cdbedaa7c9171ebc757370fe1475c976c53ebb7)

Can you rename it `modify` instead of `mutate` since we're not mutating the expression?

compiler/noirc_frontend/src/ast/statement.rs
asterite: The way it currently is, this method (`LValue::from_expression_kind`) is only called by `NodeInterner::get_lvalue`, and that only returns things that were put by `NodeInterner::push_lvalue`. And when pushing an LValue it's converted to an expression, so going backwards this should never fail.

compiler/noirc_frontend/src/ast/statement.rs
asterite: I guess we can wrap it in `Semi` but never produce an error, right?

asterite: Ah, only if it's not the last statement in the block.

jfecher: Sounds good to me, agreed we should avoid errors here.

asterite: Ah, the problem is that `Semi` wraps an expression, and here we only have a statement... I don't know how to solve this.

What's the issue if there are two statements in a row inside a block but the first one isn't semi? I just tried a program like that and it seems to work fine.

I'm also not sure how to handle this if we wanted to because at this point we don't know what the interned expression expands to.

jfecher: The only issue with an expression in a block lacking a semicolon is a potential type warning that it expected a Unit type. E.g. `3; 4` should issue this warning IIRC.

I'm more concerned about the case where `4;` is the last expression in a block and we can't add the semicolon because `4` is interned. If we can't fix it though then we can't fix it. We can address it later, I don't think it should block this PR.

compiler/noirc_frontend/src/hir/comptime/value.rs
asterite: The scenario this is handling is if you have an interned expression and you put it inside something else, for example an array literal:

```rust
let expr = quote { 1 }.as_expr().unwrap();
let array = quote { [ $expr, $expr ] }.as_expr().unwrap();
println(array);
```

Before doing all of that `resolve_expression_kind` that printed this:

```
[?Interned, ?Interned, ?Interned]
```

Even if I checked if the expression was an interned and resolved that, it would print that. At first I was confused, but the thing is that the expression was an `ArrayLiteral` that had interned expressions inside it. So I had two options:
1. Implement `fmt` again for all of the expression kinds, turning any inner interned ones into non-interned ones.
2. Only convert an `ExpressoinKind` into another one by mapping interned sub-expressions into non-interned ones,  then calling `fmt` on that.

I chose 2 to avoid repeating the printing logic.

asterite: (we could also just leave it like `[?Interned, ?Interned, ?Interned]` but while coding the "inject" example, and the "mutate" method, I used print-debugging a lot and it was very hard to understand what was going on without unwrapping these expressions.

noir_stdlib/src/meta/expr.nr
asterite: > we can run into an issue where foo(bar(baz)) would turn into foo(ctx, bar(baz)) instead of foo(ctx, bar(ctx, baz))

Right, I noticed that too. In the "inject" example what I ended up doing is, if `Expr` is a function call, add the `context` argument but also call `mutate` on each of the arguments to continue injecting `context`. I think doing it that way is a bit more cumbersome but it also gives some more flexibility. For example if I only wanted to add a `context` argument to top-level calls but not calls nested in those, I couldn't do it if sub-expressions were always mutated (or maybe I could if I kept some external state to know where I am, but maybe it would be more confusing).

Let me know if you'd still prefer to always recur.

noir_stdlib/src/meta/op.nr
asterite: Great! I forgot we had a way to kind of `panic!` at compile-time.

asterite: Oh, it doesn't seem to work:

```
error: Expected type Quoted, found type ()
   ┌─ std/meta/op.nr:41:16
   │
41 │           } else {
   │ ╭────────────────'
42 │ │             assert(false, "unhandled unary operator")
43 │ │         }
   │ ╰─────────'
```

I guess assert's type is `()`? Do we have a bottom type (`!`) like in Rust?

test_programs/compile_success_empty/inject_context_attribute/src/main.nr
asterite: I was wondering the same thing. I'll change it. I guess it's better for it to accept `Expr` because it means you'll always have to do `as_expr().unwrap()` so you can never set something that's not an `Expr` (plus no need to parse it inside `set_body`).

asterite: Well, almost: `Expr` can also be an `LValue`... though I can we can always turn that into an Expression, so I'll do that.

noir_stdlib/src/meta/expr.nr
asterite: Actually, I'll change it to always recur. We can always re-evaluate this as we implement more use cases. Maybe we never bump into one that doesn't need to always recur.

compiler/noirc_frontend/src/hir/comptime/value.rs
jfecher: Ah, ok. I misunderstood the method and thought we could replace it entirely with to_display_ast because of the Interned case.

Can we rename this method to remove the `resolve_` suffix since that is used by name resolution? We can call it something alluding to removing interned expressions, following interned expressions or similar.

jfecher (approved):
:+1:

jfecher:
@michaeljklein

michaeljklein (approved):

PR #5830, opened 2024-08-26
feat: Improve "type annotations needed" errors

jfecher:
# Description

## Problem\*

Resolves a common complaint that the "type annotations needed" error is applied too broadly.

## Summary\*

I've completely removed the old error and split it into several cases:
- Cannot call a method (please add a type annotation for the object type)
- Cannot access a struct field (please add a type annotation for the object type)
- Couldn't determine type of generic arg (please add a type annotation)

Errors that were added but shouldn't occur in practice:
- Cannot find a trait impl (multiple matching impls)
- Cyclic type constructed
- Couldn't evaluate array length to concrete size

## Documentation\*

Check one:

# PR Checklist\*


michaeljklein (approved):
LGTM

PR #5840, opened 2024-08-27
fix!: Check unused generics are bound

jfecher:
# Description

## Summary\*

Ran into this while working on arithmetic generics. It's possible to have cases like:

```rs
fn foo<let N: u32>() {}

fn main() {
    foo();
}
```

For which `N` is unused but we don't issue a "type annotations needed" error for it currently because it isn't present in the instantiated function type of `foo`. This PR fixes this and issues the error.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/test_programs/compile_failure/unspecified_generic/src/main.nr
+++ b/test_programs/compile_failure/unspecified_generic/src/main.nr
@@ -1,0 +1,5 @@
+fn foo<T>() {}
+
+fn main() {
+    foo();
+}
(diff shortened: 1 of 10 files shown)


test_programs/compile_failure/unspecified_generic/src/main.nr
```diff
@@ -0,0 +1,5 @@
+fn foo<T>() {}
+
+fn main() {
+    foo();
```
vezenovm, line 4: I think it is better to move error cases under `noirc_frontend/tests.rs` when we can. This is non-blocking though, just taking note.

vezenovm (approved):

test_programs/compile_failure/unspecified_generic/src/main.nr
jfecher: Can't move it to the frontend tests unfortunately since this error is during monomorphization.

jfecher:
Turns out this is a breaking change. We're now (correctly) erroring on test code we weren't before. E.g. all `HashMap` methods are in an impl which is generic over `H` despite most of the methods not referring to `H` at all. I'll fix the tests.

PR #5842, opened 2024-08-27
feat: show backtrace on comptime assertion failures

### LINKED ISSUE #5828: Comptime assertion failures should show the full backtrace
### Problem

When an assertion fails during comptime code evaluation, the error only shows the line that failed the assertion. That sometimes makes it impossible to understand the real cause of the assertion failure. In particular, `Option::unwrap` always fails the same way:

```
error: Assertion failed
   ┌─ std/option.nr:33:16
   │
33 │         assert(self._is_some);
   │                -------------
```

### Happy Case

The full trace is shown.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #5828

## Summary

`CustomDiagnostic` didn't have a `call_stack` property because so far this wasn't needed. I only added it for `FailingConstraint`, which is, _I think_, the only case where a backtrace might be needed. In all other cases (for example "wrong number of arguments" it might be clear why it's failing without needing a backtrace).

Here's an example output:

```
error: Assertion failed
   ┌─ std/option.nr:33:16
   │
33 │         assert(self._is_some);
   │                -------------
   │
   ┌─ src/main.nr:8:9
   │
 8 │         foo();
   │         -----
   │
   ┌─ src/other.nr:2:5
   │
 2 │     bar();
   │     -----
   ·
 7 │     let _ = expr.as_integer().unwrap();
   │             --------------------------
   │

```

## Additional Context

None.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/comptime.rs
+++ b/compiler/noirc_frontend/src/elaborator/comptime.rs
@@ -139,7 +139,7 @@ impl<'context> Elaborator<'context> {
             return Err((ResolverError::NonFunctionInAnnotation { span }.into(), self.file));
         };
 
-        let mut interpreter = self.setup_interpreter(location);
+        let mut interpreter = self.setup_interpreter();
         let mut arguments =
             Self::handle_attribute_arguments(&mut interpreter, function, arguments, location)
                 .map_err(|error| {
@@ -345,15 +345,12 @@ impl<'context> Elaborator<'context> {
         }
     }
 
-    pub fn setup_interpreter<'local>(
-        &'local mut self,
-        location: Location,
-    ) -> Interpreter<'local, 'context> {
+    pub fn setup_interpreter<'local>(&'local mut self) -> Interpreter<'local, 'context> {
         let current_function = match self.current_item {
             Some(DependencyId::Function(function)) => Some(function),
             _ => None,
         };
-        Interpreter::new(self, self.crate_id, current_function, location)
+        Interpreter::new(self, self.crate_id, current_function)
     }
 
     pub(super) fn debug_comptime<T: Display, F: FnMut(&mut NodeInterner) -> T>(
--- a/compiler/noirc_frontend/src/elaborator/comptime.rs
+++ b/compiler/noirc_frontend/src/elaborator/comptime.rs
@@ -139,7 +139,7 @@ impl<'context> Elaborator<'context> {
             return Err((ResolverError::NonFunctionInAnnotation { span }.into(), self.file));
         };
 
-        let mut interpreter = self.setup_interpreter();
+        let mut interpreter = self.setup_interpreter(location);
         let mut arguments =
             Self::handle_attribute_arguments(&mut interpreter, function, arguments, location)
                 .map_err(|error| {
@@ -345,12 +345,15 @@ impl<'context> Elaborator<'context> {
         }
     }
 
-    pub fn setup_interpreter<'local>(&'local mut self) -> Interpreter<'local, 'context> {
+    pub fn setup_interpreter<'local>(
+        &'local mut self,
+        location: Location,
+    ) -> Interpreter<'local, 'context> {
         let current_function = match self.current_item {
             Some(DependencyId::Function(function)) => Some(function),
             _ => None,
         };
-        Interpreter::new(self, self.crate_id, current_function)
+        Interpreter::new(self, self.crate_id, current_function, location)
     }
 
     pub(super) fn debug_comptime<T: Display, F: FnMut(&mut NodeInterner) -> T>(
--- a/compiler/noirc_frontend/src/elaborator/comptime.rs
+++ b/compiler/noirc_frontend/src/elaborator/comptime.rs
@@ -46,7 +46,7 @@ impl<'context> Elaborator<'context> {
             self.crate_id,
             self.debug_comptime_in_file,
             self.enable_arithmetic_generics,
-            self.interpreter_call_stack,
+            self.interpreter_call_stack.clone(),
         );
 
         elaborator.function_context.push(FunctionContext::default());
--- a/compiler/noirc_frontend/src/elaborator/comptime.rs
+++ b/compiler/noirc_frontend/src/elaborator/comptime.rs
@@ -46,6 +46,7 @@ impl<'context> Elaborator<'context> {
             self.crate_id,
             self.debug_comptime_in_file,
             self.enable_arithmetic_generics,
+            self.interpreter_call_stack,
         );
 
         elaborator.function_context.push(FunctionContext::default());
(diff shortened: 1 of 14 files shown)


compiler/noirc_frontend/src/hir/comptime/interpreter.rs
```diff
@@ -70,7 +72,9 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
         current_function: Option<FuncId>,
     ) -> Self {
         let bound_generics = Vec::new();
-        Self { elaborator, crate_id, current_function, bound_generics, in_loop: false }
+        let in_loop = false;
+        let call_stack = Vec::new();
```
jfecher, line 76: I think this will result in the call stack being reset if we ever elaborate within the interpreter, than that elaborator also runs interpreted code.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -146,7 +146,8 @@ impl<'local, 'context> Interpreter<'local, 'context> {
 }
 
 fn failing_constraint<T>(message: impl Into<String>, location: Location) -> IResult<T> {
-    Err(InterpreterError::FailingConstraint { message: Some(message.into()), location })
+    let call_stack = Vec::new();
+    Err(InterpreterError::FailingConstraint { message: Some(message.into()), location, call_stack })
```
jfecher, line 150: I think we'd want a call stack for these builtin functions too.

jfecher (commented):
I'm unsure about adding the call stack to custom diagnostics directly. We also handle call stacks differently here than we do in SSA where we report them as each file name and location upon a failed assert.

asterite:
> I'm unsure about adding the call stack to custom diagnostics directly

How else could we do it?

> We also handle call stacks differently here than we do in SSA where we report them as each file name and location upon a failed assert.

Right, I noticed that. When working with comptime code I find it clearer if the error mentions exactly where the error might be... otherwise on every `unwrap()` the only thing you'd see is that `assert(self._is_some)` and then you'll have to click on each of the frames to see where the error is... I don't know if in SSA or even when executing code it would be helpful to see the stacktrace like that too.

In any case, I can change it to just show the list of files and line numbers.

jfecher:
Oh, another issue is if we're 1000 stack frames deep in the interpreter we wouldn't want to display all of those in the error message, we should probably cut out some in the middle.

jfecher:
> How else could we do it?

They're added as secondaries so maybe we should allow specifying the file id in each secondary as well? There's a few other errors which would benefit from this. E.g. "duplicate definition of foo" then in a separate error we display "previously defined here"

asterite:
> Oh, another issue is if we're 1000 stack frames deep in the interpreter we wouldn't want to display all of those in the error message, we should probably cut out some in the middle.

What do you think we should do for now? Maybe cut it at 2 or 3 frames? I think in most cases assertion failures will be on `unwrap()` and there with two frames the error is already clear.

jfecher:
I'd think 3-5 frames? Just in case it's in library code as well.

Could be neat if we filtered out some frames from the same file in case we have something like:
```
user_function1  foo.rs
user_function2  foo.rs
library_function1 bar.rs
library_function2 bar.rs
unwrap  option.rs
```
Maybe ideally we'd display
```
user_function2
library_function 2
unwrap
```
I could also see this be confusing why `library_function1` was omitted. "But I called library_function1, not library_function2!" etc.

asterite:
I changed it to only show the last 3 locations. I tried it first with 5 but it was too much output. We could always adjust it later (or even make it configurable with a command-line argument if needed).

compiler/noirc_frontend/src/elaborator/comptime.rs
```diff
@@ -345,12 +345,15 @@ impl<'context> Elaborator<'context> {
         }
     }
 
-    pub fn setup_interpreter<'local>(&'local mut self) -> Interpreter<'local, 'context> {
+    pub fn setup_interpreter<'local>(
+        &'local mut self,
+        location: Location,
+    ) -> Interpreter<'local, 'context> {
```
jfecher, line 351: I was thinking we'd store the full call stack in the elaborator so that when we go `interpreter -> elaborator -> interpreter` it isn't lost. I expect this case to be rare though so I won't block on it.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -145,8 +145,16 @@ impl<'local, 'context> Interpreter<'local, 'context> {
...
-    Err(InterpreterError::FailingConstraint { message: Some(message.into()), location })
+fn failing_constraint<T>(
+    message: impl Into<String>,
+    location: Location,
+    call_stack: &[Location],
+) -> IResult<T> {
+    Err(InterpreterError::FailingConstraint {
+        message: Some(message.into()),
+        location,
+        call_stack: call_stack.to_vec(),
```
jfecher, line 156: Can you store a `im::Vector` instead of a `Vec` for the failing constraint here? That way we can clone much more cheaply.

asterite: Done! I didn't know if you wanted `im::Vector` for the call stack that's now on the Elaborator.

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -168,6 +168,8 @@ pub struct Elaborator<'context> {
 
     /// Temporary flag to enable the experimental arithmetic generics feature
     enable_arithmetic_generics: bool,
+
+    pub(crate) interpreter_call_stack: Vec<Location>,
```
jfecher, line 172:
```suggestion
    pub(crate) interpreter_call_stack: im::Vector<Location>,
```
We should use `im::Vector` here for call stacks like we do in SSA

jfecher (approved):

jfecher (commented):
Oh, one more thing, can you update the function in `elaborator::comptime` so that the new elaborator created by the interpreter has the interpreter's callstack?

We only have an elaborator parameter so we need to pass in the call stack from the interpreter since the one from the previous elaborator won't have the interpreter's call frames.

asterite:
@jfecher Done! I had to change the call stack to be a mutable reference now to avoid cloning on every Elaborator instantiation. Let me know if that's okay.

jfecher:
@asterite we shouldn't need a reference, `im::Vector`s are made to be cloned. From `im::Vector::clone`:

> Time: O(1), or O(n) with a very small, bounded n for an inline vector.

PR #5847, opened 2024-08-28
feat: warn on unused imports

### LINKED ISSUE #4762: Add a diagnostic for unused imports in nargo language server
### Problem

When refactoring Noir code, developers face the challenge of manually adjusting imports, as there is currently no automated way to generate required imports or remove unused ones. This manual process results in tedious and error-prone import management.

[Noir Docs](https://noir-lang.org/docs/)
[Example Noir Code with Modules](https://github.com/vlayer-xyz/noir-trie-proofs/tree/vlayer-main/lib/src)

### Happy Case

The Noir VSCode extension has the potential to alleviate this problem by communicating with the Nargo language server. This communication enables the extension to provide import-related diagnostics and actions, streamlining the import management process for developers.

### Project Impact

While not critical, integrating automated import management into the Noir VSCode extension would greatly enhance the development experience by reducing manual effort and improving code cleanliness.

### Impact Context

Without automated import management, developers must spend significant time and effort managing imports manually during refactoring, which hampers productivity and introduces the risk of errors.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

Noir already supports unused variables diagnostic and it's very useful

### Existing Functionality

Noir already has a language server implemented, residing in the repository under tooling/lsp. This server provides some functionality, such as reporting unused variables, which are highlighted in the IDE by the Noir extension. However, it currently does not report unused imports. Adding support for this functionality is the focus of this issue.

### Proposal for Resolution

Integrate automated import management features into the Noir VSCode extension, leveraging the capabilities of the existing Noir language server. This includes adding support for reporting and addressing unused imports, similar to the existing functionality for unused variables.

Maybe

### Support Needs

While the implementation of this feature is feasible, it may require collaboration with the community and feedback from project maintainers to ensure alignment with project goals and standards.

asterite:
# Description

## Problem

Resolves #4762

## Summary

The compiler now produces a warning on unused imports.

## Additional Context

I don't know if this is the correct approach. I added a new HashSet to track imported names, then those are removed as we import things. I thought about tracking this in `ItemScope` but it's tricky because the way things are they include imported and self things.

There's another thing: eventually it would be nice to warn on unused types, like in Rust. For that maybe it would make sense to track this in `ItemScope`... but given that we currently don't have visibility modifiers for types, it can't be done right now. So maybe doing it just for imports with specific code is fine.

I had to refactor a bit `resolve_path` because some paths were looked up not using that method and so some things weren't marked as used... but now `StandardPathResolver::new` is used in exactly one place in the compiler.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_driver/src/lib.rs
+++ b/compiler/noirc_driver/src/lib.rs
@@ -283,12 +283,14 @@ pub fn check_crate(
     let macros: &[&dyn MacroProcessor] =
         if options.disable_macros { &[] } else { &[&aztec_macros::AztecMacro] };
 
+    let error_on_unused_imports = true;
     let mut errors = vec![];
     let diagnostics = CrateDefMap::collect_defs(
         crate_id,
         context,
         options.debug_comptime_in_file.as_deref(),
         options.arithmetic_generics,
+        error_on_unused_imports,
         macros,
     );
     errors.extend(diagnostics.into_iter().map(|(error, file_id)| {
--- a/compiler/noirc_driver/src/lib.rs
+++ b/compiler/noirc_driver/src/lib.rs
@@ -279,11 +279,11 @@ pub fn check_crate(
     context: &mut Context,
     crate_id: CrateId,
     options: &CompileOptions,
+    error_on_unused_imports: bool,
 ) -> CompilationResult<()> {
     let macros: &[&dyn MacroProcessor] =
         if options.disable_macros { &[] } else { &[&aztec_macros::AztecMacro] };
 
-    let error_on_unused_imports = true;
     let mut errors = vec![];
     let diagnostics = CrateDefMap::collect_defs(
         crate_id,
@@ -324,7 +324,8 @@ pub fn compile_main(
     options: &CompileOptions,
     cached_program: Option<CompiledProgram>,
 ) -> CompilationResult<CompiledProgram> {
-    let (_, mut warnings) = check_crate(context, crate_id, options)?;
+    let error_on_unused_imports = true;
+    let (_, mut warnings) = check_crate(context, crate_id, options, error_on_unused_imports)?;
 
     let main = context.get_main_function(&crate_id).ok_or_else(|| {
         // TODO(#2155): This error might be a better to exist in Nargo
@@ -359,7 +360,8 @@ pub fn compile_contract(
     crate_id: CrateId,
     options: &CompileOptions,
 ) -> CompilationResult<CompiledContract> {
-    let (_, warnings) = check_crate(context, crate_id, options)?;
+    let error_on_unused_imports = true;
+    let (_, warnings) = check_crate(context, crate_id, options, error_on_unused_imports)?;
 
     // TODO: We probably want to error if contracts is empty
     let contracts = context.get_all_contracts(&crate_id);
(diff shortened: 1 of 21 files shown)


jfecher (approved):
Looks good but I'm curious why `error_on_unused_imports` is configurable. When is it false?

compiler/noirc_frontend/src/hir/def_map/mod.rs
```diff
@@ -298,6 +305,23 @@ impl CrateDefMap {
...
+    ) {
+        for module in context.def_maps[&crate_id].modules() {
+            for ident in module.unused_imports() {
+                errors.push((
+                    CompilationError::ResolverError(ResolverError::UnusedImport {
+                        ident: ident.clone(),
+                    }),
+                    module.location.file,
+                ));
+            }
```
TomAFrench, line 322: Can replace this with a `.map` and a `.extend` to avoid repeated pushing.

asterite:
> Looks good but I'm curious why error_on_unused_imports is configurable. When is it false?

The way it's done, it will only error for the root crate, not for dependencies. Does that make sense? If there are unused imports in the standard library or a dependency, you wouldn't be able to immediately fix them.

compiler/noirc_frontend/src/hir/def_map/mod.rs
asterite: Done!

compiler/noirc_frontend/src/elaborator/scope.rs
```diff
@@ -42,15 +43,36 @@ impl<'context> Elaborator<'context> {
...
+        Ok(path_resolution.module_def_id)
+    }
+
+    pub(super) fn resolve_path(&mut self, path: Path) -> PathResolutionResult {
         let mut module_id = self.module_id();
         let mut path = path;
 
+        if path.kind == PathKind::Plain {
+            self.def_maps.get_mut(&self.crate_id).unwrap().modules[module_id.local_id.0]
+                .use_import(&path.segments[0].ident);
```
TomAFrench, line 65:
```suggestion
            let module_data =
                &mut self.def_maps.get_mut(&self.crate_id).unwrap().modules[module_id.local_id.0];
            module_data.use_import(&path.segments[0].ident);
```

Small nit but I'd split this line up

TomAFrench (commented):
I can see this being an issue in libraries where I want to rexport certain items in order to create my external interface. I'm not using these imports these directly myself but they are still desirable but this PR would create warnings for them.

asterite:
> I can see this being an issue in libraries where I want to rexport certain items in order to create my external interface. I'm not using these imports these directly myself but they are still desirable but this PR would create warnings for them.

Ah, good point. I guess then we'd also need a `pub use` feature together with this, right?

TomAFrench:
I think for now we can restrict this feature to `bin` or `contract` crates, once we have `pub use` then we can roll it out to libraries.

compiler/noirc_driver/src/lib.rs
```diff
@@ -131,6 +131,56 @@ pub struct CompileOptions {
     pub skip_underconstrained_check: bool,
 }
 
+/// Similar to CompileOptions but with some extra fields specific for `check_crate`.
+#[derive(Clone, Debug, Default)]
+pub struct CheckOptions {
```
jfecher, line 136: If we want this to be a superset, let's just have a CompileOptions field internally

asterite: Sounds good. I thought the same thing while having lunch, I pushed this change right before lunch 😀

PR #5848, opened 2024-08-28
chore: underconstrained check in parallel

guipublic:
# Description

## Problem\*

underconstrained check can add significant time to the compilation process

## Summary\*
perform the underconstrained check in parallel.
The PR parallelises the check per function. In the example I tried, it gives no benefit as all the work is done in one function.
Then the PR also parallelise the 'merge sets' by doing it over chunks. In the example I tried, it reduces the duration from 77s down to 1s.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Will take a look today.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -1,34 +1,35 @@
...
     pub(crate) fn check_for_underconstrained_values(&mut self) -> Vec<SsaReport> {
-        let mut warnings: Vec<SsaReport> = Vec::new();
-        for function in self.functions.values() {
-            match function.runtime() {
-                RuntimeType::Acir { .. } => {
-                    warnings.extend(check_for_underconstrained_values_within_function(
-                        function,
+        let functions_id = self.functions.values().map(|f| f.id().to_usize()).collect::<Vec<_>>();
+        functions_id
+            .into_par_iter()
```
TomAFrench, line 21:
```suggestion
        let functions_id = self.functions.values().map(|f| f.id().to_usize());
        functions_id
            .par_bridge()
```

We can avoid the allocation of the vec here by using `par_bridge` to make the iterator parallel directly.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -327,7 +326,27 @@ impl Context {
...
+        while len > 1000 && len < prev_len {
+            let chunks: Vec<_> = sets.par_chunks(1000).collect();
+            sets = chunks.par_iter().flat_map(|sets| Self::merge_sets(sets)).collect();
+
+            prev_len = len;
+            len = sets.len();
+        }
+        // TODO: if prev_len >= len, this means we cannot effectively merge the sets anymore
+        // we should instead partition the sets into disjoint chunks an work on those chunks
+        // for now we backup to the non-parallel implementation
```
TomAFrench, line 348:
```suggestion
        // TODO: if prev_len >= len, this means we cannot effectively merge the sets anymore.
        // We should instead partition the sets into disjoint chunks and work on those chunks,
        // but for now we backup to the non-parallel implementation
```

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -327,7 +326,27 @@ impl Context {
...
+
+    /// Parallel version of merge_sets
+    /// The sets are merged by chunks, and then the chunks are merged together
+    fn merge_sets_par(sets: &[HashSet<ValueId>]) -> Vec<HashSet<ValueId>> {
+        let mut sets = sets.to_owned();
+        let mut len = sets.len();
+        let mut prev_len = len + 1;
+
+        while len > 1000 && len < prev_len {
+            let chunks: Vec<_> = sets.par_chunks(1000).collect();
```
TomAFrench, line 340:
```suggestion
        while len > 1000 && len < prev_len {
            let chunks: Vec<_> = sets.par_chunks(1000).collect();
```

Do we need the `len > 1000` condition? `par_chunks` should work with fewer chunks afaik. Do we end up slower with the parallelisation overhead without this?

TomAFrench (commented):
Looks good, I'd like a little more clarity on the termination conditions for the chunk merging but otherwise good to go.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
guipublic: Yes I think we need the condition. The point is that we need to do a (sequential) merge_sets at the end, because the chunks produce sets that are properly merged inside the chunk, but they may intersect with sets from other chunks. Testing the len is a simple way to know if we have chunks or not.

TomAFrench (approved):

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
TomAFrench: Sounds good.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -327,7 +327,27 @@ impl Context {
...
+            sets = chunks.par_iter().flat_map(|sets| Self::merge_sets(sets)).collect();
```
jfecher, line 342: Do we need to `.collect()` then immediately `.par_iter()`? Or can we iterate from the `par_chunks` directly?
Rayon's docs mention `par_chunks` returns a "Parallel iterator over immutable non-overlapping chunks of a slice" so I think we should be able to call `.par_chunks(1000).flat_map(...)` without collecting

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -327,7 +327,26 @@ impl Context {
...
+
+    /// Parallel version of merge_sets
+    /// The sets are merged by chunks, and then the chunks are merged together
+    fn merge_sets_par(sets: &[HashSet<ValueId>]) -> Vec<HashSet<ValueId>> {
+        let mut sets = sets.to_owned();
+        let mut len = sets.len();
+        let mut prev_len = len + 1;
+
+        while len > 1000 && len < prev_len {
+            sets = sets.par_chunks(1000).flat_map(|sets| Self::merge_sets(sets)).collect();
```
jfecher, line 341:
```suggestion
            sets = sets.par_chunks(1000).flat_map(Self::merge_sets).collect();
```

PR #5790, opened 2024-08-22
feat: Sync from aztec-packages

AztecBot:
Automated pull of Noir development from [aztec-packages](https://github.com/AztecProtocol/aztec-packages).
BEGIN_COMMIT_OVERRIDE
feat: Added indirect const instruction (https://github.com/AztecProtocol/aztec-packages/pull/8065)
feat: optimize to_radix (https://github.com/AztecProtocol/aztec-packages/pull/8073)
feat: automate verify_honk_proof input generation (https://github.com/AztecProtocol/aztec-packages/pull/8092)
feat: Sync from noir (https://github.com/AztecProtocol/aztec-packages/pull/8093)
feat: Poseidon2 gates for Ultra arithmetisation (https://github.com/AztecProtocol/aztec-packages/pull/7494)
feat: Sync from noir (https://github.com/AztecProtocol/aztec-packages/pull/8008)
feat: Add reusable procedures to brillig generation (https://github.com/AztecProtocol/aztec-packages/pull/7981)
feat: Small optimization in toradix (https://github.com/AztecProtocol/aztec-packages/pull/8040)
feat: unify all acir recursion constraints based on RecursionConstraint and proof_type (https://github.com/AztecProtocol/aztec-packages/pull/7993)
feat: Sync from noir (https://github.com/AztecProtocol/aztec-packages/pull/7958)
END_COMMIT_OVERRIDE

PR #5851, opened 2024-08-28
feat!: return arrays instead of slices from `to_be_radix` functions

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR experiments with removing the `limb_count` argument to the `to_be_radix` methods on `Field` and replacing it with a numeric generic. This allows us to return an array rather than a slice which should allow better codegen.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/acir_gen/mod.rs
```diff
@@ -2177,19 +2178,42 @@ impl<'a> Context<'a> {
...
             Intrinsic::ToBits(endian) => {
                 let field = self.convert_value(arguments[0], dfg).into_var()?;
-                let bit_size = self.convert_value(arguments[1], dfg).into_var()?;
 
-                let result_type = Self::array_element_type(dfg, result_ids[1]);
+                let Type::Array(result_type, array_length) = dfg.type_of_value(result_ids[0])
+                else {
+                    println!("{:?}", result_ids);
+                    println!("{:?}", dfg.type_of_value(result_ids[0]));
+                    println!("{:?}", dfg.type_of_value(result_ids[1]));
```
TomAFrench, line 2204:
```suggestion
```

TomAFrench:
This is just failing on Zac's code now (which is expected as this is a breaking change.)

I haven't made the necessary changes to documentation yet but I'd like to do a first round of review before I write that.

TomAFrench:
cc @vezenovm as this is a bytecode size motivated change.

compiler/noirc_evaluator/src/brillig/brillig_ir/codegen_intrinsic.rs
```diff
@@ -66,48 +66,45 @@ impl<F: AcirField + DebugToString> BrilligContext<F> {
...
+            output: target_array.to_heap_array(),
         });
 
-        let limb_field = SingleAddrVariable::new(self.allocate_register(), F::max_num_bits());
+        let limb_field = SingleAddrVariable::new_field(self.allocate_register());
         let limb_casted = SingleAddrVariable::new(self.allocate_register(), limb_bit_size);
 
         if limb_bit_size != F::max_num_bits() {
-            self.codegen_loop(target_vector.size, |ctx, iterator_register| {
+            self.codegen_loop(size.address, |ctx, iterator_register| {
```
vezenovm, line 91:
```suggestion
            self.codegen_loop(target_array.size, |ctx, iterator_register| {
```

compiler/noirc_evaluator/src/brillig/brillig_ir/codegen_intrinsic.rs
```diff
@@ -66,48 +66,45 @@ impl<F: AcirField + DebugToString> BrilligContext<F> {
...
-                    target_vector.pointer,
-                    iterator_register,
-                    limb_casted.address,
-                );
+                ctx.codegen_array_set(target_array.pointer, iterator_register, limb_casted.address);
             });
         }
 
         // Deallocate our temporary registers
+        self.deallocate_single_addr(size);
```
vezenovm, line 102:
```suggestion
```

compiler/noirc_evaluator/src/brillig/brillig_ir/codegen_intrinsic.rs
```diff
@@ -66,48 +66,45 @@ impl<F: AcirField + DebugToString> BrilligContext<F> {
...
     ) {
         assert!(source_field.bit_size == F::max_num_bits());
 
-        self.usize_const_instruction(target_vector.size, limb_count.into());
-        self.usize_const_instruction(target_vector.rc, 1_usize.into());
-        self.codegen_allocate_array(target_vector.pointer, target_vector.size);
+        let size = SingleAddrVariable::new_usize(self.allocate_register());
+        self.usize_const_instruction(size.address, target_array.size.into());
+        self.usize_const_instruction(target_array.rc, 1_usize.into());
+        self.codegen_allocate_array(target_array.pointer, size.address);
```
vezenovm, line 79:
```suggestion
self.codegen_allocate_fixed_length_array(target_array.pointer, target_array.size);
self.usize_const_instruction(target_array.rc, 1_usize.into());
```
I think you can replace with this, then you don't need to keep `size` around. `usize_const_instruction` allocates a register internally.

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg/capacity_tracker.rs
```diff
@@ -72,7 +72,6 @@ impl<'a> SliceCapacityTracker<'a> {
                         // where v7 is the slice length and v8 is the popped slice itself.
                         Intrinsic::SlicePopFront => (Some(1), results.len() - 1),
                         // The slice capacity of these intrinsics is not determined by the arguments of the function.
```
vezenovm, line 74:
```suggestion
```
This comment can technically be deleted. We can also remove the option wrapper around `argument_index`. This is a nit though and we can do it in a follow-up

noir_stdlib/src/ec/tecurve.nr
```diff
@@ -354,11 +354,9 @@ mod curvegroup {
 
         // Scalar multiplication (p + ... + p n times)
         pub fn mul(self, n: Field, p: Point) -> Point {
-            let N_BITS = crate::field::modulus_num_bits();
```
vezenovm, line 357: Just taking note this is another place we are ingraining the modulus size

vezenovm (commented):
Overall changes look good. Just concerned that bn254 continues to be further ingrained into the compiler and stdlib. We can mitigate this in follow-ups though as that is already the way it is...

compiler/noirc_evaluator/src/brillig/brillig_ir/codegen_intrinsic.rs
TomAFrench: This needs a `MemoryAddress` but `target_array.size` is a `u32`.

compiler/noirc_evaluator/src/brillig/brillig_ir/codegen_intrinsic.rs
TomAFrench: We'd need to initialise another constant register in order to hold the array size for the for-loop so it's best to have this explicit so we can avoid the extra `CONST` opcode imo.

noir_stdlib/src/ec/tecurve.nr
TomAFrench: Note that we're harcoding 254 already anyway on the next line so :shrug:. We need to get nargo building with a different field to tackle this properly.

vezenovm: Didn't notice that next line (although I do think it can be removed based on the comment). Just made the comment as while going through the PR I noticed these extra places we were assuming 254 I didn't realize we were.

compiler/noirc_evaluator/src/brillig/brillig_ir/codegen_intrinsic.rs
vezenovm: Can we not just do `MemoryAddress:from`?

vezenovm: Either way I just realized this code path looks different after the sync from aztec

vezenovm:
We're going to have to update this again after the [sync](https://github.com/noir-lang/noir/pull/5790) which includes this PR (https://github.com/AztecProtocol/aztec-packages/pull/8073).

noir_stdlib/src/field/bn254.nr
```diff
@@ -28,8 +28,8 @@ unconstrained pub(crate) fn decompose_hint(x: Field) -> (Field, Field) {
 }
 
 fn compute_lt(x: Field, y: Field, num_bytes: u32) -> bool {
-    let x_bytes = x.to_le_radix(256, num_bytes);
-    let y_bytes = y.to_le_radix(256, num_bytes);
+    let x_bytes: [u8; 254] = x.to_le_bytes();
+    let y_bytes: [u8; 254] = y.to_le_bytes();
```
TomAFrench, line 32:
```suggestion
    let x_bytes: [u8; 32] = x.to_le_bytes();
    let y_bytes: [u8; 32] = y.to_le_bytes();
```

asterite (approved):
LGTM

PR #5854, opened 2024-08-28
feat(meta): Comptime keccak

vezenovm:
# Description

## Problem\*

Resolves 

Just under 1/3 of a public transfer in Aztec is due to a keccak hash of the selector. We should ideally be simplifying this through our normal compiler passes as the inputs are all known at compile time. This is difficult at the moment due to having the same compilation pipeline for SSA and ACIR and would require changing how we inline. For an easy win we can enable comptime hashing of keccak.

This also just more generally provides another one of our foreign functions in the comptime environment which will be needed in the future.

## Summary\*

I added functions in the interpreter for `keccakf1600` and `to_le_radix`. I also included a test under `compile_succes_empty` that calls keccak in a comptime env.

The `assert_constant` builtin is also used by our byte decomposition functions. For this we can just return true as we already check that no non-comptime variables are referenced in comptime code.

## Documentation\*

Check one:
- [] No documentation needed.

# PR Checklist\*


compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -52,6 +53,9 @@ impl<'local, 'context> Interpreter<'local, 'context> {
         match name {
             "array_as_str_unchecked" => array_as_str_unchecked(interner, arguments, location),
             "array_len" => array_len(interner, arguments, location),
+            // We do not allow anything from the non-comptime context into the comptime context.
+            // Thus, we can just return true for `assert_constant`.
```
jfecher, line 57:
```suggestion
```
I don't think this needs additional justification

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -110,6 +114,7 @@ impl<'local, 'context> Interpreter<'local, 'context> {
             "struct_def_as_type" => struct_def_as_type(interner, arguments, location),
             "struct_def_fields" => struct_def_fields(interner, arguments, location),
             "struct_def_generics" => struct_def_generics(interner, arguments, location),
+            "to_le_radix" => to_le_radix(arguments, location), // TODO
```
jfecher, line 117: What is left to do?

compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs
```diff
@@ -18,6 +21,7 @@ pub(super) fn call_foreign(
 ) -> IResult<Value> {
     match name {
         "poseidon2_permutation" => poseidon2_permutation(interner, arguments, location),
+        "keccakf1600" => comptime_keccakf1600(interner, arguments, location),
```
jfecher, line 24: Is this different from the normal keccakf1600? If not we should name them the same like we do for poseidon2_permutation. I'd prefer to qualify `acvm::blackbox_solver::keccakf1600` if needed

jfecher (commented):
Looks good. Are there a lot of instances of calls to keccak that could be comptime?

jfecher (approved):

compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs
vezenovm: > I'd prefer to qualify acvm::blackbox_solver::keccakf1600 if needed

No it isn't, it was just for naming. I'll just do this suggestion.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
vezenovm: leftover from debugging, deleted

vezenovm:
> Looks good. Are there a lot of instances of calls to keccak that could be comptime?

Function selectors in aztec are computed using a constant string representing the function signature. As per the PR description this can be almost 1/3rd of the total Brillig bytecode size.

jfecher:
> Function selectors in Aztec are computed using a constant string representing the function signature. As per the PR description this can be almost 1/3rd of the total Brillig bytecode size.

Sounds good. Just as a quick note for them if they see this PR though, comptime functions won't actually be called at comptime currently unless they're in a `comptime` block. So they'd have to change their macro to insert a `comptime` block around the call.

Issue #4515, opened 2024-03-08
Epic: Add visibility system to hide implementation details internal to modules


TomAFrench:
### Problem

Currently all types/values defined in Noir are public, this causes an issue where libraries cannot provide a restricted interface as users can reach in an import internal implementation details.

### Happy Case

Noir should have a visibility system similar to Rust where by default functions, globals, traits, etc. are private to the modules they're defined in.

```[tasklist]
### Tasks
- [ ] https://github.com/noir-lang/noir/issues/4775
- [ ] https://github.com/noir-lang/noir/pull/4491
- [ ] https://github.com/noir-lang/noir/issues/4837
- [ ] Switch warnings into errors
```

Savio-Sou:
Good to close?

TomAFrench:
The only remaining task is #6934 which is already part of the 1.0 milestone so we can close this.

PR #5856, opened 2024-08-29
feat: use visibility

asterite:
# Description

## Problem

Part of #4515

## Summary

We recently added a warning for unused imports... but only for bin and contract packages. We didn't enable it for lib packages because a `use` could also be used as `pub use`, something we don't have yet. I thought it would be really nice if we had `pub use`, and warned on unused imports in libs too. I checked the code and we already track visibility for any item, in general, it's just that for things that don't allow a visibility modifier we just consider it's public. So I tried to see how difficult it would be to implement it, and it turned out it wasn't that hard or time-consuming.

That said, visibility for `use` involves some more logic, particularly for autocompletion, because now `pub use` should be suggested, but the "parent" module of that item isn't the actual parent (it's the module where the `pub use` is defined) but that was relatively striaght-forward to implement too.

## Additional Context

If we decide to go forward with this, any existing `use` that was used as `pub use` will likely start producing a warning for libs (a lot of them in Aztec-Packages), but now that can be silenced by changing them to `pub use`.

Where should this new feature be documented? I'm not sure if it should go in `dependencies.md` or `modules.md`.

## Documentation\*

Check one:

# PR Checklist


TomAFrench:
Haven't checked to see if a test already exists but we should make sure that we can't turn a private item `pub` with a `pub use`.

asterite:
Ah, I forgot to add tests. I'll try, but right now running `cargo test` on the frontend project gives a linking error (just on Mac, I think?) so it's kind of hard to test things.

TomAFrench:
> What's a code to do that?

```rust
// library/lib.nr

use foo::private_func;

mod foo {
    fn private_func() -> Field {
        42
    }
}

// binary/main.nr

fn main() -> Field {
    library::private_func()
}
```

This should work.

asterite:
Ah, I see.

Right now that code gives a warning on library:

```
warning: private_func is private and not visible from the current module
  ┌─ deps/foo/src/lib.nr:1:14
  │
1 │ pub use foo::private_func;
  │              ------------ private_func is private
```

but no error on the user of the library.

What should we do in this case? If a `pub use` tries to export a private item we should consider it as just `use`? Or is it fine if it just warns on the library side?

TomAFrench:
We could go further and make `private_func` be `pub(crate)` which should mean that no warnings/errors are thrown either on the library or user's side.

```rust
// library/lib.nr

pub use foo::private_func;

mod foo {
    pub(crate) fn private_func() -> Field {
        42
    }
}

// binary/main.nr

fn main() -> Field {
    library::private_func()
}
```

TomAFrench:
Realistically we should be returning a hard error for these visibility issues rather than just warnings. We should also be erroring/warning in the case in which a `use` statement is expanding the visibility of the item it's importing.

asterite:
Done! Now you get a warning if you try to re-export an item with less visibility than the use statement, plus the use statement's visibility changes to the lowest of visibilities (so you get a warning too on the user of the `pub use`).

As you said, eventually all of these should be errors, but at least now we get a warning on both sides.

compiler/noirc_driver/src/lib.rs
```diff
@@ -290,20 +278,19 @@ pub fn add_dep(
...
 
     let mut errors = vec![];
+    let error_on_unused_imports = true;
     let diagnostics = CrateDefMap::collect_defs(
         crate_id,
         context,
         options.debug_comptime_in_file.as_deref(),
         options.arithmetic_generics,
-        check_options.error_on_unused_imports,
+        error_on_unused_imports,
```
TomAFrench, line 293: left in from debugging the stdlib?

asterite: Ah, no. It's that way so it only warns on the root crate and not on dependencies.

compiler/noirc_frontend/src/hir/def_map/module_data.rs
```diff
@@ -123,15 +123,16 @@ impl ModuleData {
     pub fn import(
         &mut self,
         name: Ident,
+        visibility: ItemVisibility,
         id: ModuleDefId,
         is_prelude: bool,
     ) -> Result<(), (Ident, Ident)> {
         // Empty spans could come from implicitly injected imports, and we don't want to track those
-        if name.span().start() < name.span().end() {
+        if visibility == ItemVisibility::Private && name.span().start() < name.span().end() {
```
TomAFrench, line 131:
```suggestion
        if visibility != ItemVisibility::Public && name.span().start() < name.span().end() {
```

We should be able to find unused `pub(crate)` imports as well.

compiler/noirc_frontend/src/hir/resolution/import.rs
```diff
@@ -298,13 +307,17 @@ fn resolve_name_in_module(
...
+            // something that's visible from the current module.
+            if (plain && index == 0)
+                || can_reference_module_id(
+                    def_maps,
+                    importing_crate,
+                    starting_mod,
+                    current_mod_id,
+                    visibility,
+                )
+            {
```
TomAFrench, line 320: Does `can_reference_module_id` return the wrong value here? I'm wondering why we're adding an extra argument.

TomAFrench (commented):
Gonna do some mucking about with test programs to see if I get any counter-intuitive behaviour but looks good.

compiler/noirc_frontend/src/hir/def_map/module_data.rs
asterite: I was thinking about this today. I think we could even detect unused functions (or unused anything) in a similar way, but I'll leave that for a separate PR (I'll push the pub(crate) use change, though!)

compiler/noirc_frontend/src/hir/resolution/import.rs
asterite: I think the issue is that when you do this:

```rust
use foo::bar
```

then `bar` is added as a private item to the module. Then, say `bar` is a module and you do `bar::baz()` it would error saying that `bar` is private.

The error didn't happen before because all imports were considered public.

So the rule here is: if it's the first segment and we find something, it can't be non-visible (it's either defined in the same module, or imported into it but if it's imported it's visible)

TomAFrench: Ah yeah, I ran into a similar issue when adding module visibility previously. This makes sense now.

compiler/noirc_frontend/src/tests.rs
```diff
@@ -3231,3 +3231,78 @@ fn errors_on_unused_import() {
...
+    mod foo {
+        mod bar {
+            pub(crate) fn baz() {}
+        }
+
+        pub use bar::baz;
+    }
+
+    fn main() {
+    }
```
TomAFrench, line 3330:
```suggestion
    mod foo {
        mod bar {
            pub(crate) fn baz() {}
        }

        pub use bar::baz;
    }

    fn main() {
        foo::baz();
    }
```

I'm getting unused import errors even if I use `foo::baz()` in `main`.

asterite:
I pushed a commit to solve the remaining issue, but I had to change a "def maps" parameter to be mutable and I'm not sure that's correct. I'm thinking maybe tracking unused stuff should be outside of `ModuleData`, as a separate structure, and that's something you would pass to `resolve_path`. I'll try to do that after working on other things.

asterite:
Now usage/unused tracking is held in `NodeInterner` with a new type `UsageTracker`. I made it very explicit that this is about tracking imports, but that type could later be used to track any unused items (we'll have to rename some methods, of course). It _could_ be done right now, but it's probably better to do it in a separate PR.

I also added docs (I copied some of that from the [Rust docs](https://doc.rust-lang.org/reference/items/use-declarations.html#use-visibility)).

TomAFrench (approved):

PR #5858, opened 2024-08-29
chore: ec addition for non-zero points

guipublic:
# Description

## Problem\*

This PR performs 'safe' ec addition in Noir, assuming the EC ADD opcode is unsafe, i.e it does not handle point at infinity.

an ec addition for the embedded curve dedicated to non-zero inputs.

## Summary\*
embedded_curve_add in stdlib now handles corner cases for ec addition
Add 2 methods:
embedded_curve_add_not_nul which assumes the inputs are not null
embedded_curve_add_unsafe which assumes inputs are not null and distinct (or identical)

Enable the constant inputs for ec_add.

## Additional Context
This PR should cause an overhead, because corner cases are not checked in Noir and in BB. However, we want to change BB so that EC ADD is not safe anymore which will improve the performance, so we need to add the safe version in Noir first.

I am not sure whether we want to document this or not?

## Documentation\*

Check one:

# PR Checklist\*


guipublic:
We should hold this PR until we see if it is still relevant after switching to the alternative ec_add.

noir_stdlib/src/embedded_curve_ops.nr
```diff
@@ -125,20 +125,64 @@ pub fn fixed_base_scalar_mul(scalar: EmbeddedCurveScalar) -> EmbeddedCurvePoint
...
+/// Except for points on the curve, the other assumptions are checked by the function. It will cause assertion failure if they are not respected.
+pub fn embedded_curve_add_not_nul(point1: EmbeddedCurvePoint, point2: EmbeddedCurvePoint) -> EmbeddedCurvePoint {
+    assert(point1.x - point2.x != 0);
+    assert(point1.is_infinite == false);
+    assert(point2.is_infinite == false);
+    embedded_curve_add_unsafe(point1, point2)
+}
+
+/// Unsafe ec addition
+/// It assumes (but does not check) that the points have not the same absissa, or that they are the same variable
```
TomAFrench, line 177:
```suggestion
/// It assumes (but does not check) that the points' x-coordinates are not equal.
```

I think it'll be clearer to use `x-coordinate` rather than absissa

noir_stdlib/src/embedded_curve_ops.nr
```diff
@@ -125,20 +125,64 @@ pub fn fixed_base_scalar_mul(scalar: EmbeddedCurveScalar) -> EmbeddedCurvePoint
...
+/// This function assumes that:
+/// The points are on the curve, and
+/// The points have not the same abscissa, and
+/// Neither point is the infinity point.
+/// If it is used with correct input, the function ensures the correct non-zero result is returned.
+/// Except for points on the curve, the other assumptions are checked by the function. It will cause assertion failure if they are not respected.
+pub fn embedded_curve_add_not_nul(point1: EmbeddedCurvePoint, point2: EmbeddedCurvePoint) -> EmbeddedCurvePoint {
+    assert(point1.x - point2.x != 0);
+    assert(point1.is_infinite == false);
+    assert(point2.is_infinite == false);
```
TomAFrench, line 172:
```suggestion
    assert(point1.x != point2.x);
    assert(!point1.is_infinite);
    assert(!point2.is_infinite);
```

noir_stdlib/src/embedded_curve_ops.nr
```diff
@@ -125,20 +125,64 @@ pub fn fixed_base_scalar_mul(scalar: EmbeddedCurveScalar) -> EmbeddedCurvePoint
...
+    result.is_infinite = result_is_infinity | (point1.is_infinite & point2.is_infinite);
+    result
+}
+
+#[foreign(embedded_curve_add)]
+fn embedded_curve_add_array_return(_point1: EmbeddedCurvePoint, _point2: EmbeddedCurvePoint) -> [Field; 3] {}
+
+/// This function assumes that:
+/// The points are on the curve, and
+/// The points have not the same abscissa, and
```
TomAFrench, line 165:
```suggestion
/// The points don't share an x-coordinate, and
```

TomAFrench (approved):

TomAFrench:
I'm surprised that this isn't resulting in an increase in the number of gates.

noir_stdlib/src/embedded_curve_ops.nr
guipublic: We need to keep 'or that they are the same variable'. I rephrased the comment hoping to be more clear.

noir_stdlib/src/embedded_curve_ops.nr
guipublic: Done

noir_stdlib/src/embedded_curve_ops.nr
guipublic: Done

guipublic:
> I'm surprised that this isn't resulting in an increase in the number of gates.

I guess it is due to the activation of constant folding for the ec add opcode.

PR #5861, opened 2024-08-29
fix(sha256): Add extra checks against message size when constructing msg blocks

vezenovm:
# Description

## Problem\*

Resolves 

Issue found in zk passport https://github.com/ocelots-app/passport-verifier/blob/47e9464e7e782b07b6d791bf1d13257fce2f486b/crates/lib/data-check/integrity/src/lib.nr#L118 when performing sha on a message with a large padding.

## Summary\*

The current sha algorithm accounts for message padding, but only where ithe padding is still contained in the block we are compressing. For the case where we have a padding that extends multiple blocks past the message size we end up with a correctness error. We need to add more checks against the message size to make sure we are comrpessing the correct msg block.

An increase in gate count is expected from these changes. 

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
> Program	ACIR opcodes (+/-)	%	Circuit size (+/-)	%
> **sha256_regression**	+32,838 ❌	**+444.78%**	+114,735 ❌	**+121.86%**
> **regression_4449**	+3,850 ❌	**+18.39%**	+5,793 ❌	**+1.96%**

Two big new additional regressions were included under `sha256_regression` in this PR. If you look at the full report `sha256`, `sha2_byte`, and `sha256_var_size_regression` are largely untouched in circuit size by these changes.

test_programs/execution_success/sha256_regression/src/main.nr
```diff
@@ -1,3 +1,4 @@
+// A bunch of different test tastes for sha256_var in the stdlib
```
michaeljklein, line 1:
```suggestion
// A bunch of different test cases for sha256_var in the stdlib
```

michaeljklein (approved):
Small typo in comment, otherwise LGTM

Issue #5793, opened 2024-08-22
Check for correct bit-size in ACVM solver


michaeljklein:
### Problem

Several of the blackbox functions fail in a variety of ways on inputs with too many bits. See https://github.com/noir-lang/noir/pull/5614/ for examples.

NOTE: this is currently causing non-deterministic failures, see [Maxim's comment](https://github.com/noir-lang/noir/issues/5793#issuecomment-2313294089) below

### Happy Case

As per [Maxim's comments](https://github.com/noir-lang/noir/pull/5614/#discussion_r1710410959) on https://github.com/noir-lang/noir/pull/5614:

- We should have a clearer ban on the bit size for these functions so this isn't possible at all.
- Let's make sure we are setting up the input bits correctly here and update the issue accordingly.
- At a minimum we should panic if the incorrect bit size is provided (e.g. 254 instead of 8 for keccak256) so that we know codegen is incorrect.
- We could probably just return an `OpcodeResolutionError<F>::BlackBoxFunctionFailed`

### Workaround

None

### Project Impact

Nice-to-have

### Blocker Context

It's similar to overflow checks for blackbox function _inputs_

### Would you like to submit a PR for this Issue?

None

vezenovm:
This is causing non-deterministic failures  https://github.com/noir-lang/noir/actions/runs/10582178394/job/29321514166?pr=5838.

PR #5864, opened 2024-08-29
feat: add assertions for ACVM `FunctionInput` `bit_size`

michaeljklein:
# Description

- [x] Convert `FunctionInput::constant/witness` assertions to solver errors

## Problem\*

- Resolves https://github.com/noir-lang/noir/issues/5793

## Summary\*

- Adds assertions 

## Documentation\*

Check one:

# PR Checklist\*


test_programs/noir_test_success/regression_4080/src/main.nr
```diff
@@ -1,7 +1,8 @@
 // This test checks that `var^var` is assigned the correct type.
 // https://github.com/noir-lang/noir/issues/4080
 
-#[test(should_fail_with = "attempt to add with overflow")]
+// NOTE: expected error is the following, but the 'failed_assertion' is None in check_expected_failure_message
+#[test(should_fail_with = "Cannot solve opcode: FunctionInput value has too many bits")]
```
vezenovm, line 5: I still would like `attempt to add with overflow` to be the error we receive here rather than an internal solver error. Were we not getting any failure output on this test?

michaeljklein: > Were we not getting any failure output on this test?

Yes, the ACVM error was blocking any other errors

acvm-repo/acvm/src/compiler/optimizers/redundant_range.rs
```diff
@@ -107,10 +110,15 @@ impl<F: AcirField> RangeOptimizer<F> {
         let mut optimized_opcodes = Vec::with_capacity(self.circuit.opcodes.len());
         for (idx, opcode) in self.circuit.opcodes.into_iter().enumerate() {
             let (witness, num_bits) = match opcode {
-                Opcode::BlackBoxFuncCall(BlackBoxFuncCall::RANGE {
-                    input:
-                        FunctionInput { input: ConstantOrWitnessEnum::Witness(w), num_bits: bits },
-                }) => (w, bits),
+                Opcode::BlackBoxFuncCall(BlackBoxFuncCall::RANGE { input })
+                    if matches!(input.input_ref(), ConstantOrWitnessEnum::Witness(..)) =>
```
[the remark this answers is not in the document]
michaeljklein: Because those fields are now private to ensure the bit size is checked

tooling/fuzzer/src/dictionary/mod.rs
```diff
@@ -83,18 +83,20 @@ fn build_dictionary_from_circuit<F: AcirField>(circuit: &Circuit<F>) -> HashSet<
...
 
-            Opcode::BlackBoxFuncCall(BlackBoxFuncCall::RANGE {
-                input: FunctionInput { input: ConstantOrWitnessEnum::Constant(c), num_bits },
-            }) => {
-                let field = 1u128.wrapping_shl(*num_bits);
-                constants.insert(F::from(field));
-                constants.insert(F::from(field - 1));
-                constants.insert(*c);
+            Opcode::BlackBoxFuncCall(BlackBoxFuncCall::RANGE { input })
+                if matches!(input.input(), ConstantOrWitnessEnum::Constant(..)) =>
```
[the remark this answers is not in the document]
michaeljklein: Because those fields are now private to ensure that the bit size is checked

test_programs/noir_test_success/regression_4080/src/main.nr
vezenovm: We should avoid this regression it is going to be very confusing for a user to see `Cannot solve opcode: FunctionInput value has too many bits`.

compiler/noirc_evaluator/src/errors.rs
```diff
@@ -55,7 +55,7 @@ pub enum RuntimeError {
     #[error("Could not resolve some references to the array. All references must be resolved at compile time")]
     UnknownReference { call_stack: CallStack },
     #[error("{invalid_input_bit_size}")]
-    InvalidInputBitSize { invalid_input_bit_size: InvalidInputBitSize, call_stack: CallStack },
```
vezenovm: Apologies when I first reviewed I didn't notice we were using `acir:: InvalidInputBitSize` inside these evaluator errors. I don't think we should do that to avoid propagating the field through `RuntimeError`. I was looking at OpcodeResolutionError where we already have a generic field as part of its definition. In the evaluator we can have an error similar to `UnsupportedIntegerSize`.

acvm-repo/acvm/src/compiler/optimizers/redundant_range.rs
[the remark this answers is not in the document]
vezenovm: Let's just do a normal match on `input.input_ref()` then to avoid the `if matches!(..)` followed by an `if let` with an unreachable panic.

test_programs/noir_test_success/regression_4080/src/main.nr
michaeljklein: Ok, I've reverted the test and disabled the ACVM bitsize checks for the associated opcodes: https://github.com/noir-lang/noir/issues/5985

acvm-repo/acvm/src/pwg/blackbox/logic.rs
```diff
@@ -51,8 +51,11 @@ fn solve_logic_opcode<F: AcirField>(
     result: Witness,
     logic_op: impl Fn(F, F) -> F,
 ) -> Result<(), OpcodeResolutionError<F>> {
-    let w_l_value = input_to_value(initial_witness, *a)?;
-    let w_r_value = input_to_value(initial_witness, *b)?;
+    // TODO(https://github.com/noir-lang/noir/issues/5985): re-enable these once we figure out how to combine these with existing
+    // noirc_frontend/noirc_evaluator overflow error messages
+    let skip_bitsize_checks = true;
```
vezenovm, line 56:
```suggestion
    let skip_bit_size_checks = true;
```
nit: let's do this or add bitsize to cspell

compiler/noirc_evaluator/src/errors.rs
```diff
@@ -7,7 +7,7 @@
 //! An Error of the former is a user Error
 //!
 //! An Error of the latter is an error in the implementation of the compiler
-use acvm::FieldElement;
+use acvm::{acir::InvalidInputBitSize, FieldElement};
```
vezenovm, line 10: I still would like for us to not leak this ACVM solving error into the noirc_evaluator errors. We should have an error that looks similar to `UnsupportedIntegerSize` with two `u32` values representing the bit sizes, and a name like `InvalidBlackBoxInputBitSize`.

vezenovm (commented):
Two small nits otherwise lgtm

acvm-repo/acvm/src/pwg/blackbox/logic.rs
michaeljklein: added to cspell

compiler/noirc_evaluator/src/errors.rs
```diff
@@ -32,8 +32,13 @@ pub enum RuntimeError {
     TypeConversion { from: String, into: String, call_stack: CallStack },
     #[error("{name:?} is not initialized")]
     UnInitialized { name: String, call_stack: CallStack },
-    #[error("Integer sized {num_bits:?} is over the max supported size of {max_num_bits:?}")]
-    UnsupportedIntegerSize { num_bits: u32, max_num_bits: u32, call_stack: CallStack },
+    #[error("Integer{} sized {num_bits:?} is over the max supported size of {max_num_bits:?}", opt_value.as_ref().map(|x| format!(" ({})", x)).unwrap_or("".to_string()))]
```
vezenovm, line 35: I think it would be cleaner to not have this `opt_value` and just a separate error variant. As the error is now, for cases of `None` we are always going to have a comma trailing the error message even if there is not `opt_value`. We can have a separate error variant `InvalidBlackBoxInputBitSize` that says the wrong sized value was used for black box function input.

acvm-repo/acvm/src/pwg/mod.rs
```diff
@@ -633,12 +656,31 @@ pub fn witness_to_value<F>(
...
-    match input.input {
-        ConstantOrWitnessEnum::Witness(witness) => Ok(*witness_to_value(initial_witness, witness)?),
+    match input.input() {
+        ConstantOrWitnessEnum::Witness(witness) => {
+            let initial_value = *witness_to_value(initial_witness, witness)?;
+            if skip_bitsize_checks || initial_value.num_bits() <= input.num_bits() {
+                Ok(initial_value)
+            } else {
+                let value_num_bits = initial_value.num_bits();
+                let value = format!("{}", initial_value);
```
TomAFrench, line 673:
```suggestion
                let value = initial_value.to_string();
```

TomAFrench (approved):

Issue #4535, opened 2024-03-12
SSA optimization is ineffective when an unconstrained function has loops


sirasistant:
### Problem

This code:
```rust
struct EnumEmulation {
    a: Option<Field>,
    b: Option<Field>,
    c: Option<Field>,
}

unconstrained fn main() -> pub Field {
    let mut emulated_enum = EnumEmulation { a: Option::some(1), b: Option::none(), c: Option::none() };

    for _ in 0..1 {
        assert_eq(emulated_enum.a.unwrap(), 1);
    }

    emulated_enum.a = Option::some(2);
    emulated_enum.a.unwrap()
}

```

is functionally the same as
```rust
struct EnumEmulation {
    a: Option<Field>,
    b: Option<Field>,
    c: Option<Field>,
}

unconstrained fn main() -> pub Field {
    let mut emulated_enum = EnumEmulation { a: Option::some(1), b: Option::none(), c: Option::none() };

    assert_eq(emulated_enum.a.unwrap(), 1);

    emulated_enum.a = Option::some(2);
    emulated_enum.a.unwrap()
}

```

However, their optimized SSAs are completely different
```
After Dead Instruction Elimination:
brillig fn main f0 {
  b0():
    v111 = allocate
    store u1 1 at v111
    v112 = allocate
    store Field 1 at v112
    v113 = allocate
    store u1 0 at v113
    v114 = allocate
    store Field 0 at v114
    v115 = allocate
    store u1 0 at v115
    v116 = allocate
    store Field 0 at v116
    jmp b1(u64 0)
  b1(v13: u64):
    v117 = eq v13, u64 0
    jmpif v117 then: b2, else: b3
  b2():
    v124 = load v111
    v125 = load v112
    constrain v124 == u1 1
    constrain v125 == Field 1
    v131 = add v13, u64 1
    jmp b1(v131)
  b3():
    v120 = load v113
    v121 = load v114
    v122 = load v115
    v123 = load v116
    store u1 1 at v111
    store Field 2 at v112
    store v120 at v113
    store v121 at v114
    store v122 at v115
    store v123 at v116
    return Field 2
}
```
vs
```
After Dead Instruction Elimination:
brillig fn main f0 {
  b0():
    return Field 2
}
```

I think mem2reg is failing to promote to reg values that are used across multiple blocks

### Happy Case

Ideally, the two programs should generate the same optimized SSA

### Project Impact

Nice to have

### Impact Context

[...]

PR #5865, opened 2024-08-29
feat(mem2reg): Remove trivial stores 

vezenovm:
# Description

## Problem\*

Partially resolves https://github.com/noir-lang/noir/issues/4535. We can still do some more cleanup but will handle that in a follow-up. References the PR discussion comments for more context.

## Summary\*

Just marking the result of a load known was causing some failures, specifically for arrays. So for now I just look for trivial stores that are immediately storing the same value that was just loaded.

Also testing on CI as I keep getting rayon errors with cargo test for some reason. Making this check mark more values known was also difficult to due to this issue with testing so I am just pushing the more trivial case in this draft.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/mem2reg/block.rs
```diff
@@ -164,6 +164,9 @@ impl Block {
                 // known references if this reference is ever stored to.
             }
         }
+        // else {
+        //     self.references.insert(result, ReferenceValue::Known(address));
+        // }
```
jfecher, line 169: I think we don't handle these `Expression::Dereference` cases well in mem2reg in general.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -274,6 +273,13 @@ impl<'f> PerFunctionContext<'f> {
                     self.instructions_to_remove.insert(*last_store);
                 }
 
+                if let Some(last_load) = self.last_loads.get(&address) {
+                    let load_result = self.inserter.function.dfg.instruction_results(*last_load)[0];
+                    if load_result == value {
+                        self.instructions_to_remove.insert(instruction);
+                    }
+                }
```
jfecher, line 281: I'm unsure on the correctness of this (and last_loads as a whole) in the presence of aliasing. If we had e.g.
```rs
let foo = &mut 0;
let alias = foo;

let foo_load1 = *foo;
*alias = 7;
*foo = foo_load1; // This looks like it'd be optimized out

assert_eq(*foo, 7);
```

jfecher (changes requested):
Some potential aliasing issues

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
vezenovm: We expect this assert to fail right? This still fails with these changes. 

Either way this approach does look to have aliasing issues, specifically with arrays so I'm going to have to update this approach.

vezenovm:
I'm kind of surprised by the brillig_cow_assign failure looking at the SSA:
```
brillig fn main f0 {
  b0():
    inc_rc [Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0]
    v2 = allocate
    store [Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0] at v2
    v3 = load v2
    inc_rc v3
    v4 = load v2
    inc_rc v4
    store v4 at v2
    v5 = allocate
    store v3 at v5
    jmp b1(u32 0)
  b1(v6: u32):
    v9 = lt v6, u32 10
    jmpif v9 then: b2, else: b3
  b2():
    v21 = eq v6, u32 5
    jmpif v21 then: b4, else: b5
  b4():
    v27 = load v2
    inc_rc v27
    v28 = load v2
    inc_rc v28
    store v28 at v2
    store v27 at v5
    jmp b5()
  b5():
    v22 = load v2
    v23 = array_set v22, index v6, value Field 27
    v25 = add v6, u32 1
    store v23 at v2
    v26 = add v6, u32 1
    jmp b1(v26)
  b3():
    v10 = load v2
    v12 = array_get v10, index u32 6
    v14 = eq v12, Field 27
    constrain v12 == Field 27
    v15 = load v5
    v16 = array_get v15, index u32 6
    v17 = eq v16, Field 27
    v18 = not v17
    constrain v17 == u1 0
    return 
}
```
```
After Mem2Reg:
brillig fn main f0 {
  b0():
    inc_rc [Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0]
    v30 = allocate
    inc_rc [Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0]
    inc_rc [Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0]
    store [Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0] at v30
    v33 = allocate
    store [Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0, Field 0] at v33
    jmp b1(u32 0)
  b1(v6: u32):
    v34 = lt v6, u32 10
    jmpif v34 then: b2, else: b3
  b2():
    v42 = eq v6, u32 5
    jmpif v42 then: b4, else: b5
  b4():
    v43 = load v30
    inc_rc v43
    v44 = load v30
    inc_rc v44
    store v43 at v33
    jmp b5()
  b5():
    v45 = load v30
    v46 = array_set v45, index v6, value Field 27
    v47 = add v6, u32 1
    store v46 at v30
    v48 = add v6, u32 1
    jmp b1(v48)
  b3():
    v35 = load v30
    v36 = array_get v35, index u32 6
    v37 = eq v36, Field 27
    constrain v36 == Field 27
    v38 = load v33
    v39 = array_get v38, index u32 6
    v40 = eq v39, Field 27
    v41 = not v40
    constrain v40 == u1 0
    return 
}
```
When comparing to the SSA without this optimization the only difference is there is an additional `store v44 at v30` before `store v43 at v33` in `b4`. I assume this is due Brillig's CoW optimization and how `inc_rc` is processed, although the SSA looks like it should execute correctly. 

Edit: Found this hack (https://github.com/noir-lang/noir/blob/5c4f19f097dd3704522996330c961bf0a2db8d99/compiler/noirc_evaluator/src/ssa/function_builder/mod.rs#L466). This PR is removing that re-store.

vezenovm:
Accounting the extra re-load and [re-store for inc_rc instructions](https://github.com/noir-lang/noir/blob/5c4f19f097dd3704522996330c961bf0a2db8d99/compiler/noirc_evaluator/src/ssa/function_builder/mod.rs#L476) fixes the `brillig_cow_assign` bug. 

Interestingly, `uhashmap` is the last failure then and it only occurs with the [additional mem2reg and DIE passes](https://github.com/noir-lang/noir/blob/82bc2ea0126dce2e7bc7e2a07605e7dec63d3da1/compiler/noirc_evaluator/src/ssa.rs#L123). Those passes were added to cleanup leftover unused stores after the loads referencing that store are removed during DEI.
e.g. without those two passes `brillig_loop_size_regression` gives the following SSA:
```
brillig fn main f0 {
  b0():
    v50 = allocate
    store u1 1 at v50, from_rc false
    v51 = allocate
    store Field 1 at v51, from_rc false
    v52 = allocate
    store u1 0 at v52, from_rc false
    v53 = allocate
    store Field 0 at v53, from_rc false
    v54 = allocate
    store u1 0 at v54, from_rc false
    v55 = allocate
    store Field 0 at v55, from_rc false
    jmp b1(u32 0)
  b1(v10: u32):
    v56 = eq v10, u32 0
    jmpif v56 then: b2, else: b3
  b2():
    v63 = load v50
    v64 = load v51
    constrain v63 == u1 1
    constrain v64 == Field 1
    v70 = add v10, u32 1
    jmp b1(v70)
  b3():
    store u1 1 at v50, from_rc false
    store Field 2 at v51, from_rc false
    return Field 2
}
```
Then with those passes we get the following:
```
brillig fn main f0 {
  b0():
    v71 = allocate
    store u1 1 at v71, from_rc false
    v72 = allocate
    store Field 1 at v72, from_rc false
    jmp b1(u32 0)
  b1(v10: u32):
    v77 = eq v10, u32 0
    jmpif v77 then: b2, else: b3
  b2():
    v78 = load v71
    v79 = load v72
    constrain v78 == u1 1
    constrain v79 == Field 1
    v80 = add v10, u32 1
    jmp b1(v80)
  b3():
    store u1 1 at v71, from_rc false
    store Field 2 at v72, from_rc false
    return Field 2
}
```

jfecher:
> Interestingly, uhashmap is the last failure then and it only occurs with the [additional mem2reg and DIE passes](https://github.com/noir-lang/noir/blob/82bc2ea0126dce2e7bc7e2a07605e7dec63d3da1/compiler/noirc_evaluator/src/ssa.rs#L123). Those passes were added to cleanup leftover unused stores after the loads referencing that store are removed during DEI.

Ideally we get rid of those extra passes anyway. It's a fairly unsatisfactory tradeoff with compilation time IMO to add two extra passes to get rid of a couple extra instructions in some brillig functions. More concerning is why running these again (although the culprit is presumably mem2reg) produces an issue in the first place.

vezenovm:
> Ideally we get rid of those extra passes anyway. It's a fairly unsatisfactory tradeoff with compilation time IMO to add two extra passes to get rid of a couple extra instructions in some brillig functions.

Yeah agreed. I'm going to mark this PR ready for review without those extra passes as it still shows a good improvement. 

> More concerning is why running these again (although the culprit is presumably mem2reg) produces an issue in the first place.

Then in a follow-up we can investigate the cause of `uhashmap` failing.

compiler/noirc_evaluator/src/ssa/ir/printer.rs
```diff
@@ -173,8 +173,8 @@ fn display_instruction_inner(
         }
         Instruction::Allocate => writeln!(f, "allocate"),
         Instruction::Load { address } => writeln!(f, "load {}", show(*address)),
-        Instruction::Store { address, value } => {
-            writeln!(f, "store {} at {}", show(*value), show(*address))
+        Instruction::Store { address, value, from_rc } => {
+            writeln!(f, "store {} at {}, from_rc {}", show(*value), show(*address), *from_rc)
```
jfecher, line 177: I'm not a big fan of including `from_rc` on every store instruction for a mem2reg-specific issue.

Can we add tracking to mem2reg specifically instead? E.g. if we see a load -> dec-rc -> store we mark that we can't remove the store?

vezenovm: I'll look at switching to that

vezenovm: I switched to tracking the current rc reload per instruction with an `Option<(ValueId, bool)>`.

vezenovm: Ah something looks to be failing. I had it passing before but I guess I made a bad change while cleaning up.

vezenovm: I had moved when I was calling my method to track the rc reload state, but this led to inadvertently calling the method on the wrong instruction id. This is now fixed and the PR is ready for review again.

vezenovm: Looks to still be failing on the debugger actually as the debugger inserts foreign calls which are breaking up the expected block of this form:
```
    v72 = load v57
    inc_rc v72
    v73 = load v57
    inc_rc v73
    store v73 at v57
    store v72 at v60
```
In the debugger we see the following after mem2reg:
```
    v80 = load v65
    inc_rc v80
    call v44(Field 1, v80)
    inc_rc v80
    store v80 at v67
```

jfecher:
@vezenovm if it helps for this PR I was talking with @sirasistant who mentioned changing the design of how inc/dec-rc works in brillig. A side-effect of that work would be that we no longer would have to `store` after we `load` and `dec-rc` arrays at the end of each function. Should make things easier here I imagine?

vezenovm:
> A side-effect of that work would be that we no longer would have to `store` after we `load` and `dec-rc` arrays at the end of each function. Should make things easier here I imagine?

Yeah it should. The main issue here is I have to differentiate which stores I can actually remove if the known value of a store equals the address it is storing into. Without inc_rc/dec_rc, I can safely remove any of these stores. It does look to work with the solution on this branch. Although it is a bit hacked around as I have `uhashmap` failures on https://github.com/noir-lang/noir/pull/5905 which I believe are happening to due these stores being inadvertently deleted.

jfecher:
We can pause this work for a week or so then while the brillig changes are being worked on. Hopefully with the removal of the special tracking uhashmap won't fail anymore.

vezenovm:
> We can pause this work for a week or so then while the brillig changes are being worked on. Hopefully with the removal of the special tracking uhashmap won't fail anymore.

I haven't nailed down the exact cause but I think `uhashmap` might be failing due to a separate reason. I want to nail down exactly why it is failing. If it is due to inc_rc/dec_rc I'll wait for the brillig changes, otherwise I will try to resolve it.

vezenovm:
> If it is due to inc_rc/dec_rc I'll wait for the brillig changes, otherwise I will try to resolve it.

Ok I have nailed down the cause of the failure and it is in fact unrelated to inc_rc/dec_rc. 

When cleaing up stores we check a couple things:
1. That there is not a load of that store. 
2. That we do not have a reference param.

Now that I am cleaning up loads as well (https://github.com/noir-lang/noir/pull/5905) we can remove some stores if we know all loads to that store have been removed. However, I was not checking whether the address of the store we want to remove is possibly used as a reference directly such as in the parameter of a call. `uhashmap` is now passing for me locally on #5905 when checking whether the last store we want to remove is used in a call.

vezenovm:
@jfecher I am going to pull out the changes from #5905 as I had built it off of this PR and the `uhashmap` failure is unrelated to the inc_rc/dec_rc edge case. I will pause working on this PR (but leave it open) until the design of inc_rc/dec_rc is updated in brillig.

vezenovm:
@jfecher I was thinking we could replace this PR with (https://github.com/noir-lang/noir/pull/5935) which handles inc_rc / dec_rc in the safest manner by just always assuming when we have an inc_rc/dec_rc before a store we cannot remove that store. In this PR I was attempting to check for specifically [this case](https://github.com/noir-lang/noir/blob/5c4f19f097dd3704522996330c961bf0a2db8d99/compiler/noirc_evaluator/src/ssa/function_builder/mod.rs#L476) but that was leading to issues. PR #5935 just accepts a smaller improvement (21% on this PR -> 7% on the new one) for safety. And then when inc_rc/dec_rc are redesigned we can remove the small check.

vezenovm:
Closing in favor of https://github.com/noir-lang/noir/pull/5935

Issue #1579, opened 2023-06-06
LSP: Support code actions


phated:
### Problem

The LSP should support various code actions following https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction

Code actions can perform many behaviors, but a common one would be to import missing identifiers.

### Happy Case

A user should be presented with useful, contextual code actions while using an LSP client, such as the vscode plugin.

### Would you like to submit a PR for this Issue?

No

PR #5876, opened 2024-08-30
feat: LSP code actions to import or qualify unresolved paths

asterite:
# Description

## Problem

Part of #1579

## Summary

There are many code actions we could support, but the most useful ones are importing unknown identifiers or fully-qualifying them. This is what this PR is about.

lsp-code-actions

Also thanks to VS Code that it notices that if you request code action on an error, it actually requests code actions for that error span... so it seems like the two are connected, but they are not! And it works just fine :-)

Also fixes a bug I found in "go to definition" when there was no definition to resolve to (precisely the case this code action works on).

## Additional Context

Note that in Rust Analyzer it shows as "Qualify [name]" or "Import [name]" and if there are multiple possibilities, a popup shows up letting you choose which one. I wanted to do the same, but looking at the server trace for Rust Analyzer I see each code action has a "group" property, and that's how they are grouped, but... I couldn't find that property in the lsp-types crate, or even in the LSP specification 😮 . So for now each possibility will show up (no intermediate popup). But I think that's better than not having this functionality at all, and we could always improve it later if we find out how.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/lsp/src/modules.rs
+++ b/tooling/lsp/src/modules.rs
@@ -1,0 +1,131 @@
+use std::collections::BTreeMap;
+
+use noirc_frontend::{
+    ast::ItemVisibility,
+    graph::CrateId,
+    hir::def_map::{CrateDefMap, ModuleId},
+    macros_api::{ModuleDefId, NodeInterner},
+    node_interner::ReferenceId,
+};
+
+use crate::visibility::is_visible;
+
+pub(crate) fn get_parent_module(
+    interner: &NodeInterner,
+    module_def_id: ModuleDefId,
+) -> Option<ModuleId> {
+    let reference_id = module_def_id_to_reference_id(module_def_id);
+    interner.reference_module(reference_id).copied()
+}
+
+pub(crate) fn get_parent_module_id(
+    def_maps: &BTreeMap<CrateId, CrateDefMap>,
+    module_id: ModuleId,
+) -> Option<ModuleId> {
+    let crate_def_map = &def_maps[&module_id.krate];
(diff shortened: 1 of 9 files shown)


tooling/lsp/src/modules.rs
```diff
@@ -0,0 +1,131 @@
...
+                break;
+            };
+
+            segments.push(&parent_attributes.name);
+            current_attributes = parent_attributes;
+        }
+    }
+
+    if !is_relative {
+        // We don't record module attriubtes for the root module,
```
TomAFrench, line 122:
```suggestion
        // We don't record module attributes for the root module,
```

tooling/lsp/src/visibility.rs
```diff
@@ -0,0 +1,13 @@
+use noirc_frontend::{ast::ItemVisibility, hir::def_map::ModuleId};
+
+pub(super) fn is_visible(
+    visibility: ItemVisibility,
+    current_module: ModuleId,
+    target_module: ModuleId,
+) -> bool {
+    match visibility {
+        ItemVisibility::Public => true,
+        ItemVisibility::Private => false,
```
TomAFrench, line 10: This isn't always the case as you can import a private item from a submodule of where it's defined.

Looks like we'd have to make larger changes in order to suggest private items for auto-import so we can just create an issue for this.

TomAFrench (approved):

tooling/lsp/src/visibility.rs
asterite: Makes sense. Captured in #5879

PR #5888, opened 2024-09-02
feat: module attributes

### LINKED ISSUE #5495: Support module attributes
### Problem

We currently support running comptime attribute functions when a attribute is placed on a struct, trait, or function.

### Happy Case

We should support module-level attributes as well as attributes that apply to submodule declarations, e.g:

```
#![module_level_attribute]

...

#[submodule_attribute]
module foo { ... }

#[submodule_attribute]
contract bar { ... }
```

### Project Impact

None

### Impact Context

_No response_

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #5495

## Summary

Pending:
- [ ] Decide whether to keep attributes as String or SecondaryAttribute in ModuleData
- [ ] Parsing of module attributes is not ideal (it errors on non-secondary attributes, but I think not all secondary attributes are valid for modules... but maybe it's fine because struct attributes work the same way)

## Documentation\*

Check one:

# PR Checklist\*


asterite:
I asked for a review but I'm sure there are still many things to do here.

For example, I couldn't make a function attached to a module attribute to create new functions (but this seems to also happen with function attributes 🤔).

compiler/noirc_frontend/src/elaborator/comptime.rs
```diff
@@ -399,9 +402,55 @@ impl<'context> Elaborator<'context> {
...
         self.run_attributes_on_functions(functions, &mut generated_items);
+
         generated_items
     }
 
+    fn run_attributes_on_modules(&mut self, generated_items: &mut CollectedItems) {
+        if self.ran_module_attributes {
+            return;
+        }
+        self.ran_module_attributes = true;
```
jfecher, line 413: Instead of storing this bool on the elaborator and finding module attributes through the def_maps, I think we should try to add module attributes as another field on `CollectedItems` instead. We don't store anything else for modules there but they can be the first. Since we'll also want to support module-level attributes like `#![foo]` in the future.

compiler/noirc_frontend/src/parser/mod.rs
```diff
@@ -291,6 +293,7 @@ impl std::fmt::Display for SortedModule {
 #[derive(Clone, Debug, Default)]
 pub struct ParsedModule {
     pub items: Vec<Item>,
+    pub inner_attributes: Vec<SecondaryAttribute>,
```
jfecher, line 296: Hmm I was going to object to these not being in the `items` vec so that they'd keep their order when we run `nargo fmt` but I suppose we probably want to move them all to the top of the program anyway.

compiler/noirc_frontend/src/parser/parser.rs
```diff
@@ -175,13 +176,20 @@ fn program() -> impl NoirParser<ParsedModule> {
...
-            .to(ParsedModule::default())
-            .then(spanned(top_level_statement(module_parser)).repeated())
-            .foldl(|mut program, (statement, span)| {
-                if let Some(kind) = statement.into_item_kind() {
-                    program.items.push(Item { kind, span });
-                }
+        inner_attributes()
+            .then(
+                empty()
+                    .to(ParsedModule::default())
```
jfecher, line 182: I think we can be more flexible than rust here and allow parsing these as a normal top level statement. Notably this could allow other attributes to output an inner attribute.

test_programs/compile_success_empty/comptime_module/src/main.nr
```diff
@@ -1,10 +1,45 @@
+#[outer_attribute]
 mod foo {
+    #![some_attribute]
```
jfecher, line 3: I wish we had a way of differentiating attributes which are just annotations from ones that we should run as functions. I spent a bit too long searching for `out_attribute` and `some_attribute` functions.

compiler/noirc_frontend/src/parser/parser.rs
asterite: Sounds good, I also hesitated about this approach or the other. I saw Rust only allows them at the top, though in reality it also allows them syntactically anywhere but it gives an error... I'll change it to a TopLevelItem, there's also something I forgot to mention which is that these inner attributes sometimes disappear when the file gets formatted (only if the module is empty) so moving them to a TopLevelItem might fix that too 🤞

compiler/noirc_frontend/src/elaborator/comptime.rs
asterite: Done! The code didn't end up as nice as I wanted, though, gathering the attributes happens inside some nested calls instead of in a single place like `collect_functions`. But let me know if there's a way to improve this.

Also, now `ModuleData` ends up having `inner_attributes` and `outer_attributes` but that distinction is no longer needed.... should we just flatten that? For understanding what was going on it helped me to always pass both in functions to know I wasn't forgetting one of them.

compiler/noirc_frontend/src/elaborator/comptime.rs
```diff
@@ -374,12 +386,15 @@ impl<'context> Elaborator<'context> {
...
     pub(super) fn run_attributes(
         &mut self,
+        module_attributes: &Vec<ModuleAttribute>,
         traits: &BTreeMap<TraitId, UnresolvedTrait>,
         types: &BTreeMap<StructId, UnresolvedStruct>,
         functions: &[UnresolvedFunctions],
     ) -> CollectedItems {
         let mut generated_items = CollectedItems::default();
 
+        self.run_attributes_on_modules(module_attributes, &mut generated_items);
```
jfecher, line 396: Request from Grego: Can we run module attributes _after_ each other attribute type instead? So after functions in this case

Also: since the ordering we run them in matches the ordering of parameters in this function, lets update that too.

compiler/noirc_frontend/src/elaborator/comptime.rs
```diff
@@ -399,9 +414,29 @@ impl<'context> Elaborator<'context> {
         }
 
         self.run_attributes_on_functions(functions, &mut generated_items);
+
         generated_items
     }
 
+    fn run_attributes_on_modules(
+        &mut self,
+        module_attributes: &Vec<ModuleAttribute>,
```
jfecher, line 423:
```suggestion
        module_attributes: &[ModuleAttribute],
```
Surprised clippy doesn't give a lint here

compiler/noirc_frontend/src/elaborator/comptime.rs
```diff
@@ -374,12 +386,15 @@ impl<'context> Elaborator<'context> {
     /// Returns any new items generated by attributes.
     pub(super) fn run_attributes(
         &mut self,
+        module_attributes: &Vec<ModuleAttribute>,
```
jfecher, line 389:
```suggestion
        module_attributes: &[ModuleAttribute],
```

jfecher (approved):

PR #5889, opened 2024-09-02
feat: remove blocks which consist of only a jump to another block

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR removes blocks which are empty and just perform an unconditional jump to another block. We instead just update any references to this block with the block which it will eventually jump to.

I initially tried applying this optimization universally but ran into troubles where ACIR functions would run into early returns due before branches reached a common block again and so the cfg flattening pass would fail. This optimisation then only applies to brillig functions (although it wouldn't really affect ACIR anyway).

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Seems like this should be gated to only affect brillig functions.

compiler/noirc_evaluator/src/ssa/opt/simplify_cfg.rs
```diff
@@ -102,6 +108,79 @@ fn check_for_constant_jmpif(
...
+
+    if !function.dfg[block].instructions().is_empty()
+        || !function.dfg[block].parameters().is_empty()
+    {
+        return;
+    }
+
+    if let Some(TerminatorInstruction::Jmp { destination: final_destination, arguments, .. }) =
+        function.dfg[block].terminator()
+    {
```
jfecher, line 126: Can you do another early return here? `let Some(..) = ... else { return; }`

jfecher (approved):

PR #5890, opened 2024-09-03
chore: error on false constraint

guipublic:
# Description

## Problem\*

Report a bug during compilation when a constraint is false

## Summary\*

It is done in acir-gen which ensures that all side effects are resolved.

## Additional Context

The reported bug does not fail the compilation, because it will mess around compile/run-time failures and also for the Noir tests. We want a consistent handling of assert, whether it fails at compile-time, run-time, or test-time.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/aztec_macros/src/utils/ast_utils.rs
+++ b/aztec_macros/src/utils/ast_utils.rs
@@ -187,8 +187,8 @@ pub fn check_trait_method_implemented(trait_impl: &NoirTraitImpl, method_name: &
 
 /// Checks if an attribute is a custom attribute with a specific name
 pub fn is_custom_attribute(attr: &SecondaryAttribute, attribute_name: &str) -> bool {
-    if let SecondaryAttribute::Custom(custom_attr) = attr {
-        custom_attr.as_str() == attribute_name
+    if let SecondaryAttribute::Custom(custom_attribute) = attr {
+        custom_attribute.contents.as_str() == attribute_name
     } else {
         false
     }
--- a/aztec_macros/src/utils/parse_utils.rs
+++ b/aztec_macros/src/utils/parse_utils.rs
@@ -53,6 +53,7 @@ fn empty_item(item: &mut Item) {
         ItemKind::Import(use_tree, _) => empty_use_tree(use_tree),
         ItemKind::Struct(noir_struct) => empty_noir_struct(noir_struct),
         ItemKind::TypeAlias(noir_type_alias) => empty_noir_type_alias(noir_type_alias),
+        ItemKind::InnerAttribute(_) => (),
     }
 }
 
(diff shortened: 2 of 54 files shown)


jfecher (commented):
We removed this previously because it could cause asserts to seemingly fail out of order. If the first assert is not constant but a later one is it could look to users like the first succeeded when it did not since they only see the failure for the later one.

guipublic:
> We could maybe consider a warning instead like "assert is always false" although we'd have to have a full callstack since the same assert could be in the program multiple times due to inlining.

I used an `InternalBug` instead, as discussed in Scrum.

compiler/noirc_evaluator/src/errors.rs
```diff
@@ -111,6 +112,8 @@ pub enum InternalWarning {
 pub enum InternalBug {
     #[error("Input to brillig function is in a separate subgraph to output")]
     IndependentSubgraph { call_stack: CallStack },
+    #[error("Assertion is false")]
+    AssertFailed { call_stack: CallStack },
```
jfecher, line 116: Can we change this to something that hints that we're showing this at compile-time so that users may not be as confused why other possibly-false assertions may be skipped?

E.g.
- "This assertion is always false"
- "Code after always-false assertion is unreachable"

guipublic: Done, I have added a secondary message to explain better.

compiler/noirc_evaluator/src/errors.rs
```diff
@@ -88,6 +88,7 @@ impl From<SsaReport> for FileDiagnostic {
                     InternalBug::IndependentSubgraph { call_stack } => {
                         ("There is no path from the output of this brillig call to either return values or inputs of the circuit, which creates an independent subgraph. This is quite likely a soundness vulnerability".to_string(),call_stack)
                     }
+                    InternalBug::AssertFailed { call_stack } => ("As a result, the compiled circuit is ensured to fail. Other assertion might fail also during execution".to_string(), call_stack)
```
jfecher, line 91:
```suggestion
                    InternalBug::AssertFailed { call_stack } => ("As a result, the compiled circuit is ensured to fail. Other assertions may also fail also during execution".to_string(), call_stack)
```

jfecher (approved):
LGTM

compiler/noirc_evaluator/src/errors.rs
```diff
@@ -88,6 +88,7 @@ impl From<SsaReport> for FileDiagnostic {
                     InternalBug::IndependentSubgraph { call_stack } => {
                         ("There is no path from the output of this brillig call to either return values or inputs of the circuit, which creates an independent subgraph. This is quite likely a soundness vulnerability".to_string(),call_stack)
                     }
+                    InternalBug::AssertFailed { call_stack } => ("As a result, the compiled circuit is ensured to fail. Other assertions may also fail also during execution".to_string(), call_stack)
```
jfecher, line 91:
```suggestion
                    InternalBug::AssertFailed { call_stack } => ("As a result, the compiled circuit is ensured to fail. Other assertions may also fail during execution".to_string(), call_stack)
```

PR #5891, opened 2024-09-03
feat: simplify `jmpif`s by reversing branches if condition is negated

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

Currently if an `if-else` block has a negated condition we generate an extra instruction to calculate the condition rather than just swapping the "if" and "else" branches. I've added a check for this in the `simplify_cfg` pass.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
This branch changes the order of the jmpif as expected in the simplify_cfg pass however the issue comes up in flattening as we end up merging the values incorrectly.

Master
```
acir(inline) fn main f0 {
  b0(v0: Field):
    v1 = allocate
    v2 = allocate
    store Field 0 at v2
    v5 = eq v0, Field 10
    v6 = not v5
    enable_side_effects v6
    v7 = load v2
    store Field 2 at v2
    store v7 at v2
    enable_side_effects u1 1
    v10 = cast v6 as Field
    v11 = cast v5 as Field
    v12 = mul v10, Field 2 // NOT(v5) * 2
    v13 = mul v11, v7 // v5 * 0
    v14 = add v12, v13
    store v14 at v2
    v15 = load v2
    v16 = eq v15, Field 2
    constrain v15 == Field 2
    return 
}
```

This branch
```
acir(inline) fn main f0 {
  b0(v0: Field):
    v1 = allocate
    v2 = allocate
    store Field 0 at v2
    v5 = eq v0, Field 10
    v6 = not v5
    enable_side_effects v5
    v7 = load v2
    store Field 2 at v2
    v9 = not v5 
    store v7 at v2
    enable_side_effects u1 1
    v11 = cast v5 as Field
    v12 = cast v9 as Field
    v13 = mul v11, Field 2 // v5 * 2
    v14 = mul v12, v7 // NOT(v5) * 0
    v15 = add v13, v14
    store v15 at v2
    v16 = load v2
    v17 = eq v16, Field 2
    constrain v16 == Field 2
    return 
}
```

vezenovm:
> This branch changes the order of the jmpif as expected in the simplify_cfg pass however the issue comes up in flattening as we end up merging the values incorrectly.

If we do not see an simple or obvious way to have this optimization work w/ flattening we could make this optimization just work on Brillig for now.

jfecher:
I'd like if we investigated why flattening isn't adapting to the new changes. Could be related to the bug I've been looking at.

TomAFrench:
Agreed, we should at least understand the root cause of this first.

michaeljklein:
I was able to narrow down one of the failing tests to the following:

This fails the final assertion:
```noir
// x == 5
fn main(x: Field) {
    let mut z = false;
    if x != 20 {
        z = true;
    }
    assert(x != 20);

    // fails here
    assert(z);
}
```

But this succeeds:
```noir
// x == 5
fn main(x: Field) {
    let mut z = false;
    if x != 20 {
        z = true;
    } else {
        z = false;
    }
    assert(x != 20);

    assert(z);
}
```

So there appears to be a bug around single-block `if` statements and a negated dynamic condition.

TomAFrench:
I've done some investigating on this and the error is due to the fact that we've moved the `return` terminator to the `then` branch rather than the `else` branch. We then end up going down a different branch in `handle_terminator` and miss the codepath which handles switching to the other branch gracefully.

https://github.com/noir-lang/noir/blob/10a9f8104e1ebb8fad044927ff130a0e2ce9131b/compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs#L369-L390

Essentially we've created an early return in the program.

TomAFrench:
The two options for solving this then are:

- Prevent a return statement from existing within the `else` branch, that is we need to turn every `if` statement into and `if-else` with an empty `else` block. This avoids a return statement in the `else` block so it cannot be switched into being in the `then` block by this optimisation.
- Prevent this optimisation from running in ACIR functions and just have it work on brillig (which can tolerate early returns).

Considering the lack of any change in circuit sizes, I'm inclined to go for the latter as it avoids making changes across the codebase (and relying on someone not undoing this change and accidentally causing a breakage).

TomAFrench:
> Considering the lack of any change in circuit sizes

On second thoughts, this is due to the fact that at least one program has failed in each commit of this PR so we haven't measured this. Another solution would be to query the CFG to make sure that we don't merge in the `else` block before applying this optimization.

PR #5892, opened 2024-09-03
feat: warn on unused functions

asterite:
# Description

## Problem

Now that we warn on unused imports, doing that for functions too is relatively straight-forward.

## Summary

Now unused private or `pub(crate)` functions will be reported as unused.

## Additional Context

I'd like to try this on some Aztec-Packages projects, but for that it would be nice to merge #5895 first so I could get all warnings/errors for a package inside VS Code.

We can eventually do the same thing with globals, traits, etc., once we track their visibility.

I also think we could warn on unused `pub` functions in a "bin" or "contract" package, not sure... but if we decide to do that, it could be a separate PR.

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/hir/def_collector/dc_crate.rs
```diff
@@ -388,20 +389,14 @@ impl DefCollector {
                         let result = current_def_map.modules[resolved_import.module_scope.0]
                             .import(name.clone(), visibility, module_def_id, is_prelude);
 
-                        // Empty spans could come from implicitly injected imports, and we don't want to track those
-                        if visibility != ItemVisibility::Public
-                            && name.span().start() < name.span().end()
-                        {
```
jfecher, line 394: Why was this check removed?

compiler/noirc_frontend/src/hir/def_collector/dc_crate.rs
```diff
@@ -476,26 +471,34 @@ impl DefCollector {
...
+    fn check_unused_items(
+        context: &mut Context,
         crate_id: CrateId,
         errors: &mut Vec<(CompilationError, FileId)>,
     ) {
-        let unused_imports = context.def_interner.usage_tracker.unused_imports().iter();
+        // Assume `main` is used so we don't warn on that
+        let root = ModuleId { krate: crate_id, local_id: context.def_maps[&crate_id].root };
+        let main = Ident::new("main".to_string(), Span::default());
+        context.def_interner.usage_tracker.mark_as_used(root, &main);
```
jfecher, line 489: Have you tested this with e.g. contract crates to ensure the entry functions there don't warn?

compiler/noirc_frontend/src/hir/resolution/errors.rs
```diff
@@ -20,8 +20,8 @@ pub enum ResolverError {
     DuplicateDefinition { name: String, first_span: Span, second_span: Span },
     #[error("Unused variable")]
     UnusedVariable { ident: Ident },
-    #[error("Unused import")]
-    UnusedImport { ident: Ident },
+    #[error("Unused {item_type}")]
+    UnusedItem { ident: Ident, item_type: String },
```
jfecher, line 24: Nit: can `item_type` be a `&'static str`?

compiler/noirc_frontend/src/hir/def_collector/dc_crate.rs
asterite: It was moved to `UsageTracker::add_unused_item`

compiler/noirc_frontend/src/hir/def_collector/dc_crate.rs
asterite: Oh, I didn't know abot that. What are contract entry functions?

asterite: Ah, I should have searched a bit more, I guess they are `fn entrypoint` 😅

asterite: I just checked and even though they are private, they don't error. They are annotated with `aztec`, and looking a bit into the macro code it seems they are changed to `pub fn` 😮‍💨

jfecher: Can you assure this is still the case for contract functions not associated with `aztec`?
IIRC entry point functions in contracts need to be annotated with `#[entry_point]` or something similar:
```rs
contract foo {
    #[entry_point]
    fn bar() {}
}
```

asterite:
I pushed and reverted a commit that removed a couple of unused functions in the standard library: I don't know if users use them despite the warning that they are private so that's why I left them. They could be removed in a separate PR, though.

compiler/noirc_frontend/src/hir/def_collector/dc_crate.rs
asterite: Okay, it seems all contract methods are public entrypoints, so for those we shouldn't produce an unused warning. I'll push that change.

asterite: Done! Also made sure that case is covered in a test. I also checked again in Aztec-Packages and the functions that were giving a warning no longer give one.

jfecher (approved):

PR #1676, opened 2023-06-13
feat(lsp): Publish diagnostics on file save

phated:
# Description

## Problem\*

Resolves #1584 

## Summary\*

This adds diagnostic publishing to the LSP. We decided to only run the diagnostic check on file save and then we perform the "push" diagnostic pattern. This is due to the compiler not being built to operate in "pull" diagnostic mode since the client can request them at any time with source code provided; however, the compiler expects a filesystem graph and builds it from scratch. We also decided to only run this on save because "file change" notifications can come in quicker than we can compile them.

## Documentation

- [ ] This PR requires documentation updates when merged.

  

  - [ ] I will submit a noir-lang/docs PR.

  

  - [ ] I will request for and support Dev Rel's help in documenting this PR.

  
  

## Additional Context

I need to create some follow up issues related to TODOs before I move this out of draft.

# PR Checklist\*

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #5895, opened 2024-09-03
feat: LSP diagnostics for all package files

asterite:
# Description

## Problem

I noticed that currently LSP will only produce diagnostics for the file you save. That's not very convenient, as if for example you make a change to a function signature that breaks someone using it, you won't find out unless you go to that other file and save it (but how do you know which file is it?) My guess is that you'd use `nargo` on the command line to find out.

## Summary

Now when a package is type-checked, all errors are reported, regardless of whether they match the file that was saved. I believe this will speed up developing with Nargo! (it was also slowing me down when trying out new warnings or errors)

lsp-error-on-all-files

## Additional Context

I thought implementing this was going to be straight-forward (just start reporting errors in all files) but there's a catch: if a file had errors but doesn't have errors anymore, we still have to notify that the file now **doesn't** have errors. This requires us tracking which files currently have errors and making a diff with the new errors.

That said: maybe there was a reason errors were only reported in the file that was saved?

Final comment: ideally when you save a file in a package, all packages that depend on this package would also get type-checked... but that's currently slow (or noticeably slow), but I hope we could eventually make it work (like in Rust Analyzer).

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/lsp/src/lib.rs
+++ b/tooling/lsp/src/lib.rs
@@ -4,7 +4,7 @@
 #![cfg_attr(not(test), warn(unused_crate_dependencies, unused_extern_crates))]
 
 use std::{
-    collections::{BTreeMap, HashMap},
+    collections::{BTreeMap, HashMap, HashSet},
     future::Future,
     ops::{self, ControlFlow},
     path::{Path, PathBuf},
@@ -95,6 +95,9 @@ pub struct LspState {
     cached_parsed_files: HashMap<PathBuf, (usize, (ParsedModule, Vec<ParserError>))>,
     cached_def_maps: HashMap<String, BTreeMap<CrateId, CrateDefMap>>,
     options: LspInitializationOptions,
+
+    // Tracks files that currently have errors, by package root.
+    files_with_errors: HashMap<String, HashSet<Url>>,
 }
 
 impl LspState {
@@ -113,6 +116,8 @@ impl LspState {
             cached_parsed_files: HashMap::new(),
             cached_def_maps: HashMap::new(),
             options: Default::default(),
+
+            files_with_errors: HashMap::new(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


TomAFrench:
I'm fairly sure that we restricted the updating of diagnostics for performance reasons.

asterite:
> I'm fairly sure that we restricted the updating of diagnostics for performance reasons.

I see. But the way it works (even before this PR) is that we type-check the entire package, but then don't report errors that aren't for the saved files. So it's not like we are type-checking more stuff now, and I checked with some Aztec-Packages projects and it's still fast (or, well, the same as before).

asterite:
I found the PR that introduced diagnostics: https://github.com/noir-lang/noir/pull/1676

It says:

> We decided to only run the diagnostic check on file save and then we perform the "push" diagnostic pattern. This is due to the compiler not being built to operate in "pull" diagnostic mode since the client can request them at any time with source code provided; however, the compiler expects a filesystem graph and builds it from scratch.

But I don't understand that explanation.

tooling/lsp/src/lib.rs
```diff
@@ -95,6 +95,9 @@ pub struct LspState {
     cached_parsed_files: HashMap<PathBuf, (usize, (ParsedModule, Vec<ParserError>))>,
     cached_def_maps: HashMap<String, BTreeMap<CrateId, CrateDefMap>>,
     options: LspInitializationOptions,
+
+    // Tracks files that currently have errors, by package root.
+    files_with_errors: HashMap<String, HashSet<Url>>,
```
jfecher, line 100: Looks like the roots are originally `PathBuf`s, why not keep them as PathBufs instead of going to as_os_str -> to_string_lossy?

asterite: Good idea, done! I also made that changed for a couple more HashMaps.

asterite:
> We decided to only run the diagnostic check on file save and then we perform the "push" diagnostic pattern.

Ah, I understand this comment better now. It seems diagnostics can be pushed by the server to the client, or the client can ask them for a specific file (pull model), which is a newer thing (since 2023). But that seems to be unrelated to a server pushing diagnostics for all files where there are errors. Also: Rust Analyzer works this way too, so it's _probably_ okay.

jfecher (approved):

PR #5896, opened 2024-09-03
feat: better println for Quoted

### LINKED ISSUE #5899: Fix formatting of `Quoted` values
### Problem

Unquoting inside string literals doesn't work

### Happy Case

This should work:

```rust
fn main() {
    comptime {
        let x = 3;
        let q = quote { x is $x };
        assert(false, f"q = {q}");
    }
}
```

Right now it gives this error:

```
error: q = quote { x is (UnquoteMarker) }
  ┌─ src/main.nr:5:16
  │
5 │         assert(false, f"q = {q}");
  │                ----- Assertion failed
  │
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #5899

For debugging purposes, when you `println` a quoted value, some token values aren't expanded, making it a bit harder to understand what's going on.

## Summary

Now interned expressions are shown in `Quoted` values.

This program:

```rust
fn main() {
    comptime
    {
        let n = quote { [1, 2, 3] }.as_expr().unwrap();
        let q = quote { $n };
        println(q);
    }
}
```

Used to print:

```
quote { (expr) }
```

Now it prints: 

```
quote { [1, 2, 3] }
```

## Additional Context

None.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/value.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/value.rs
@@ -567,37 +567,33 @@ impl<'value, 'interner> Display for ValuePrinter<'value, 'interner> {
             Value::Quoted(tokens) => {
                 write!(f, "quote {{")?;
                 for token in tokens.iter() {
+                    write!(f, " ")?;
+
                     match token {
                         Token::QuotedType(id) => {
-                            write!(f, " {}", self.interner.get_quoted_type(*id))?;
+                            write!(f, "{}", self.interner.get_quoted_type(*id))?;
                         }
                         Token::InternedExpr(id) => {
-                            let expr = self.interner.get_expression_kind(*id);
-                            let expr =
-                                remove_interned_in_expression_kind(self.interner, expr.clone());
-                            write!(f, " {}", expr)?;
+                            let value = Value::expression(ExpressionKind::Interned(*id));
+                            value.display(self.interner).fmt(f)?;
                         }
                         Token::InternedStatement(id) => {
-                            let stmt = self.interner.get_statement_kind(*id);
-                            let stmt =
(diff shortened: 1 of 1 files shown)


compiler/noirc_frontend/src/hir/comptime/value.rs
```diff
@@ -571,6 +571,32 @@ impl<'value, 'interner> Display for ValuePrinter<'value, 'interner> {
...
+                        }
+                        Token::InternedUnresolvedTypeData(id) => {
+                            let typ = self.interner.get_unresolved_type_data(*id);
+                            write!(f, " {}", typ)?;
+                        }
+                        Token::UnquoteMarker(id) => {
+                            let hir_expr = self.interner.expression(id);
+                            let expr = hir_expr.to_display_ast(self.interner, Span::default());
+                            write!(f, " {}", expr.kind)?;
+                        }
```
jfecher, line 599: Instead of handling these cases here we should create `Value` variants like `Value::TypedExpr(id)` for this case then recursively call `fmt` on those. Currently we'd print a quoted expr containing an unquoted typed expr well, but the base case for typed exprs below on line 661 is unchanged so normal TypedExpr values would be printed differently.

If we create values for these cases and format those recursively instead this inner match can hopefully be a bit smaller and we can share more code.

jfecher (changes requested):

compiler/noirc_frontend/src/hir/comptime/value.rs
asterite: Great idea, I just pushed a commit for it.

jfecher (approved):
Thanks!

Issue #5897, opened 2024-09-03
SSA: Extra mem2reg pass causing failures for UHashMap


vezenovm:
### Aim

While working on https://github.com/noir-lang/noir/pull/5865 our `brillig_loop_size_regression` program is left with a bunch of extra stores and allocates following DIE which removes any loads that reference those stores. As a quick solution to test getting to a minimal bytecode I placed an extra mem2reg and DIE pass following the previously final DIE pass. The two extra passes are not an ideal solution for #5865 due to efficiency reasons, but it is still worrisome that `uhashmap` would fail given an extra mem2reg and DIE pass so I am taking note of that in this issue.

### Expected Behavior

I should be able to insert mem2reg anywhere in the compilation flow and not cause test failures.

### Bug

Running mem2reg and DIE after the already existing DIE causes uhashmap to fail with a type mismatch:
```
error: Assertion failed: 'Bit size for lhs 0 does not match op bit size 32'
    ┌─ std/collections/umap.nr:358:12
    │
358 │   if self.len() + 1 >= self.capacity() / 2 {
```

### To Reproduce

1. You can reference this commit https://github.com/noir-lang/noir/pull/5865/commits/a595349ddbf531ce63011a35a434f4c63fef7552 for more context about failures.
2. The easiest way to reproduce is to add an extra `mem2reg` and `die` pass after our existing `die` pass. 
3. For both strategies recompile and execute `execution_success/uhashmap`

### Workaround

Yes

### Workaround Description

Do not run mem2reg after DIE

### Project Impact

Blocker

### Blocker Context

Not a current blocker, but definitely has the potential to block in the future

### Would you like to submit a PR for this Issue?

None

vezenovm:
Closing this as the cause this failure was due to inappropriately accounting for reference params https://github.com/noir-lang/noir/pull/5865#issuecomment-2329735593 and https://github.com/AztecProtocol/aztec-packages/pull/8448 removed the need for a re-load and store when laying down inc/dec rc instruction

PR #5905, opened 2024-09-03
feat(perf): Remove unused last loads in mem2reg

vezenovm:
# Description

## Problem\*

Resolves 

## Summary\*

Follow-up to https://github.com/noir-lang/noir/pull/5865

Make the last loads removal smarter by checking whether that last loads is actually used anywhere. If it is not, remove the loads and allow stores to that loads address to be removed as well.

## Documentation\*

Check one:

# PR Checklist\*


test_programs/execution_success/brillig_loop_size_regression/src/main.nr
```diff
@@ -7,10 +7,8 @@ struct EnumEmulation {
 unconstrained fn main() -> pub Field {
     let mut emulated_enum = EnumEmulation { a: Option::some(1), b: Option::none(), c: Option::none() };
 
-    // Do a copy to optimize out loads in the loop
-    let copy_enum = emulated_enum;
```
vezenovm, line 11: After this PR:
Even after removing this copy optimization we get a Brillig bytecode size improvement from 55 -> 40.

Keeping the copy optimizations we get the following final SSA which produces 24 brillig opcodes:
```
brillig fn main f0 {
  b0():
    jmp b1(u32 0)
  b1(v6: u32):
    v27 = eq v6, u32 0
    jmpif v27 then: b2, else: b3
  b2():
    v28 = add v6, u32 1
    jmp b1(v28)
  b3():
    return Field 2
}
```

vezenovm: Ok this PR now produces the optimal SSA specified above without the copy optimization. Just getting a failure on `uhashmap` that is the same as https://github.com/noir-lang/noir/issues/5897.

vezenovm:
Closing in favor of https://github.com/noir-lang/noir/pull/5925

PR #5907, opened 2024-09-03
feat: Add `String` to stdlib

jfecher:
# Description

## Summary\*

Adds an unsized `String` type to Noir's stdlib, implemented using slices.

## Additional Context

There are several issues with this PR:

1. While I was testing this I think I found a bug with slices/builtins which can be seen in the strings test. The program fails with:

```
Failed to solve program: 'Cannot solve opcode: Attempted to load uninitialized memory block'
```

Pinging @vezenovm for visibility

2. Also, I noticed we can call `String::from("foo")` but not `"let _: String = foo".into()`. Seems to be an issue with impl search.

3. Another concern I have for this PR is that users will want to store Strings in arrays but this would be forbidden since we do not allow storing slices in arrays. Perhaps Strings should be comptime-only?

4. Since `String` is a struct, printing it just prints out the inner bytes field rather than an actual string value.

## Documentation\*

Check one:

# PR Checklist\*


Savio-Sou:
Is this meant to supersede/replace the native string type?

jfecher:
@Savio-Sou No, it was meant to provide an alternate unsized string type but it seems there are issues with it currently.

jfecher:
Closing this since we have CtString now

Issue #5771, opened 2024-08-20
Content of a struct gets "destroyed"


benesjan:
### Aim

I was just working on an arbitrary task in aztec-nr and my code started to fail in an e2e test because a value in a struct was suddently zero.

### Expected Behavior

Struct should not get reset to "zero".

### Bug

A content of a struct gets reset to zero. A bug will be obvious when running the reproduction 

### To Reproduce (updated on 30.9.2024)

1. git clone https://github.com/AztecProtocol/aztec-packages.git
2. cd aztec-packages
3. git fetch origin janb/context-bug-reproduction && git checkout janb/context-bug-reproduction
4. cd noir-projects/noir-contracts/contracts/test_contract
5. nargo execute --package test_contract --silence-warnings

Once you run these commands you should get the following error:

<img width="922" alt="image" src="https://github.com/user-attachments/assets/0c748583-2ff4-4610-965f-70d3cecbac9f">

### To Reproduce (original occurrence)
1. git clone https://github.com/AztecProtocol/aztec-packages.git
2. cd aztec-packages
3. git fetch origin dereferencing-bug && git checkout dereferencing-bug
4. cd noir-projects/noir-contracts/contracts/token_with_refunds_contract
5. nargo execute --package token_with_refunds_contract --silence-warnings

Once you run these commands you should get the following error:
<img width="910" alt="image" src="https://github.com/user-attachments/assets/4f4de673-cca9-4f1b-adbd-980cea82ac03">

### Workaround

None

### Project Impact

Blocker

### Blocker Context

It prevents merging of [this PR](https://github.com/AztecProtocol/aztec-packages/pull/7912)

### Nargo Version

nargo version = 0.33.0 noirc version = 0.33.0+26b87408b9360b55a59ed1589a0275dfc5ef9a02 (git version hash: 26b87408b9360b55a59ed1589a0275dfc5ef9a02, is dirty: false)

### Would you like to submit a PR for this Issue?

None

TomAFrench:
This branch is very out of date with the current master branch of aztec-packages so migrating it across to this repository is then difficult. Can you update this so that it pulls aztec-nr from a fixed commit from https://github.com/AztecProtocol/aztec-nr?

benesjan:
@TomAFrench pointing it to a specific version in aztec-nr repo is not possible because I've modified aztec-nr to not do oracle calls. But I've just merged the branch with master. Is that enough?

jfecher:
[...]

PR #5909, opened 2024-09-03
chore: Add pass to normalize Ids in SSA

jfecher:
# Description

## Problem\*

SSA can be difficult to debug manually, particularly for very large files. Even when comparing two sources side by side it can be hard to find exactly where they differ since one optimization difference earlier on can affect where ValueIds start in every function afterward.

## Summary\*

This PR adds a pass to normalize ids in an SSA program - restarting from v0 after every SSA pass instead of continuing from the previous end. The goal of this is to be able to take two SSA programs and easily diff them to find out where they start differing.

E.g. using this on two files containing the final SSA from https://github.com/noir-lang/noir/issues/5771 in both failing and passing versions, it is clear that they differ in exactly one ValueId in one instruction.

## Additional Context

This new pass is only run when `--show-ssa` is specified, and is run before each printout.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
@@ -397,6 +397,10 @@ impl<'f> PerFunctionContext<'f> {
             TerminatorInstruction::JmpIf { .. } => (), // Nothing to do
             TerminatorInstruction::Jmp { destination, arguments, .. } => {
                 let destination_parameters = self.inserter.function.dfg[*destination].parameters();
+                if destination_parameters.len() != arguments.len() {
+                    eprintln!("{}", self.inserter.function);
+                    eprintln!("Current block {block}, dest {destination}({destination_parameters:?}), in function {}", self.inserter.function.name());
+                }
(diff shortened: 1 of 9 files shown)


compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -397,6 +397,10 @@ impl<'f> PerFunctionContext<'f> {
             TerminatorInstruction::JmpIf { .. } => (), // Nothing to do
             TerminatorInstruction::Jmp { destination, arguments, .. } => {
                 let destination_parameters = self.inserter.function.dfg[*destination].parameters();
+                if destination_parameters.len() != arguments.len() {
+                    eprintln!("{}", self.inserter.function);
+                    eprintln!("Current block {block}, dest {destination}({destination_parameters:?}), in function {}", self.inserter.function.name());
+                }
```
TomAFrench, line 403:
```suggestion
```

TomAFrench (approved):
Looks good except for what looks like some leftover debugging code.

PR #5910, opened 2024-09-03
feat(perf): Remove last store in return block if last load is before that store

vezenovm:
# Description

## Problem\*

Working towards #4535 

## Summary\*

Small optimization found while working in mem2reg. Now that we have a small per function state we can see which stores/loads that are leftover can be removed. 

A place we are potentially missing stores that can be removed is in return blocks. If we have a store inside of a return block that is not loaded from inside that block, we can safely remove that store (given it is not a reference param). 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
@@ -171,10 +171,10 @@ impl<'f> PerFunctionContext<'f> {
                         .get(store_address)
                         .map(|(_, last_load_block)| *last_load_block != *block_id)
                         .unwrap_or(true);
-                    !is_return_value && last_load_not_in_return;
+                    !is_return_value && last_load_not_in_return
                 } else {
                     self.last_loads.get(store_address).is_none()
-                }
+                };
 
                 if remove_load && !is_reference_param {
                     self.instructions_to_remove.insert(*store_instruction);
--- a/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
@@ -116,7 +116,7 @@ struct PerFunctionContext<'f> {
 
     /// Track a value's last load across all blocks.
     /// If a value is not used in anymore loads we can remove the last store to that value.
-    last_loads: HashMap<ValueId, InstructionId>,
+    last_loads: HashMap<ValueId, (InstructionId, BasicBlockId)>,
 }
 
 impl<'f> PerFunctionContext<'f> {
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -152,9 +152,31 @@ impl<'f> PerFunctionContext<'f> {
...
+
+                    // If the last load of a store is not part of the block with a return terminator,
+                    // we can safely remove this store.
+                    let last_load_not_in_return = self
+                        .last_loads
+                        .get(store_address)
+                        .map(|(_, last_load_block)| *last_load_block != *block_id)
+                        .unwrap_or(true);
+                    remove_load = !is_return_value && last_load_not_in_return;
+                }
```
TomAFrench, line 177:
```suggestion
                let terminator = self.inserter.function.dfg[*block_id].unwrap_terminator();

                let is_return = matches!(terminator, TerminatorInstruction::Return { .. });
                let remove_load = if is_return {
                    // Determine whether the last store is used in the return value
                    let mut is_return_value = false;
                    terminator.for_each_value(|return_value| {
                        is_return_value = return_value == *store_address || is_return_value;
                    });

                    // If the last load of a store is not part of the block with a return terminator,
                    // we can safely remove this store.
                    let last_load_not_in_return = self
                        .last_loads
                        .get(store_address)
                        .map(|(_, last_load_block)| *last_load_block != *block_id)
                        .unwrap_or(true);
                    !is_return_value && last_load_not_in_return;
                } else {
                    self.last_loads.get(store_address).is_none()
                }
```

TomAFrench: Small reordering to avoid `remove_load` being mutable.

TomAFrench:
Looking at `brillig_loop_size_regression` we end up with a bunch of unused stores which are never read from now so we can look at this in a followup.

```
brillig fn main f0 {
  b0():
    v50 = allocate
    store u1 1 at v50
    v51 = allocate
    store Field 1 at v51
    v52 = allocate
    store u1 0 at v52
    v53 = allocate
    store Field 0 at v53
    v54 = allocate
    store u1 0 at v54
    v55 = allocate
    store Field 0 at v55
    jmp b1(u32 0)
  b1(v10: u32):
    v56 = eq v10, u32 0
    jmpif v56 then: b2, else: b3
  b2():
    v63 = load v50
    v64 = load v51
    constrain v63 == u1 1
    constrain v64 == Field 1
    v70 = add v10, u32 1
    jmp b1(v70)
  b3():
    return Field 2
}
```

TomAFrench (approved):

vezenovm:
> Looking at `brillig_loop_size_regression` we end up with a bunch of unused stores which are never read from (or only stored and read from once) now so we can look at this in a followup.

Yup that is the plan! Currently trying to get mem2reg to do it during its pass. We only end up with these empty stores after they are removed by DIE. So we need to remove the unused loads during mem2reg and then the stores.

PR #5918, opened 2024-09-04
fix: prevent comptime println from crashing LSP

### LINKED ISSUE #5904: LSP should communicate via socket, not standard input/otuput
### Problem

If you `println` at comptime, that gets mixed in the stdin/stdout stream of the LSP server and it ends up crashing it.

### Happy Case

We should change our LSP code to communicate via sockets. The `nargo lsp` command should get the port to use. Then the extension should start communicating via a socket. We should still keep the old mode until users upgrade to the newest version of the extension.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #5904

## Summary

I found out that writing to the error stream when connected to the LSP client actually sends that output to the LSP output window ("Noir Language Server") in our case so I chose to do that for comptime println/print. Advantages:
- No more crashing of the LSP server on comptime println
- If you know it goes to that output you can actually see the output without having to go to a terminal to run that code, so debugging comptime will be pretty fast now
- (if you don't know the output goes there, it's fine too: at least the LSP server doesn't crash 😄)

lsp-comptime-println

## Additional Context

Though the original issue was about using sockets to communicate with the LSP client, I think we won't need that after all if comptime println was the only issue (and I think most LSP servers use stdin/stdout for communication because it's simpler).

Also: it would be better if this output showed up in a dedicated output window, say "Noir Language comptime output" or something like that, but only the LSP client can create those windows and write to them... so implementing that would mean somehow communicating this data from the server to the client (maybe a socket?), which is much harder to do... so for now the output will have to show up like regular logs.

And a question: when the LSP server starts we have this code:

```rust
eprintln!("LSP starting...");
```

that shows up on the output window... should we remove it? It feel like it was used to debug the server to know that it started, but I don't know how useful it is now.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
@@ -1653,10 +1653,20 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
         assert_eq!(arguments.len(), 2);
 
         let print_newline = arguments[0].0 == Value::Bool(true);
-        if print_newline {
-            println!("{}", arguments[1].0.display(self.elaborator.interner));
+        let contents = arguments[1].0.display(self.elaborator.interner);
+        if self.elaborator.interner.is_in_lsp_mode() {
+            // If we `println!` in LSP it gets mixed with the protocol stream and leads to crashing
+            // the connection. If we use `eprintln!` not only it doesn't crash, but the output
+            // appears in the "Noir Language Server" output window in case you want to see it.
+            if print_newline {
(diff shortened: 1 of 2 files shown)


jfecher (approved):
> eprintln!("LSP starting...");

I'm fine with removing it since rust-analyzer doesn't have it

TomAFrench (approved):

PR #5921, opened 2024-09-04
feat: check argument count and types on attribute function callback

### LINKED ISSUE #5903: Error instead of panic when attribute function doesn't have the right arguments
### Aim

The compiler shouldn't panic if an attribute function doesn't have the expected arguments.

### Expected Behavior

This code:

```rust
#[attr]
fn foo() {}

fn attr() -> Quoted {
}

fn main() {
    bar();
}
```

should give an error saying that `attr` should have a `FunctionDefinition` argument.

Similarly, if there's a first argument but it's not a `FunctionDefinition`, it should also error.

### Bug

No error for the above case.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #5903

## Summary

Now the compiler will check that attribute function callbacks have at least one argument, that that argument's type matches the corresponding type, and that remaining arguments also match the types given.

Also previously errors on these callbacks were shown on the function that had the attribute, instead of on the attribute, likely because attributes didn't have a Span attached to them: this PR adds that too.

## Additional Context

The error message is still a bit strange because if you have code like this:

```rust
#[attr]
fn foo() {}

fn attr() {}

fn main() {}
```

You get this:

```
error: Expected 0 arguments, but 1 was provided
  ┌─ src/main.nr:1:3
  │
1 │ #[attr]
  │   ---- Too many arguments
```

which kind of makes sense, because 1 implicit argument was provided but 0 are expected in the callback, but maybe the error should point out that the callback actually needs one argument. Let me know if you think we should improve the error message here... but at least it doesn't error anymore.

Oh, I remember why I didn't improve that error message: the error should likely be on the callback function, but it should point out that the error happens because of a given attribute, so we need two different locations for the error, which I think we currently doesn't support.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/aztec_macros/src/utils/ast_utils.rs
+++ b/aztec_macros/src/utils/ast_utils.rs
@@ -187,7 +187,7 @@ pub fn check_trait_method_implemented(trait_impl: &NoirTraitImpl, method_name: &
 
 /// Checks if an attribute is a custom attribute with a specific name
 pub fn is_custom_attribute(attr: &SecondaryAttribute, attribute_name: &str) -> bool {
-    if let SecondaryAttribute::Custom(custom_attr) = attr {
+    if let SecondaryAttribute::Custom(custom_attr, _) = attr {
         custom_attr.as_str() == attribute_name
     } else {
         false
--- a/aztec_macros/src/utils/ast_utils.rs
+++ b/aztec_macros/src/utils/ast_utils.rs
@@ -187,8 +187,8 @@ pub fn check_trait_method_implemented(trait_impl: &NoirTraitImpl, method_name: &
 
 /// Checks if an attribute is a custom attribute with a specific name
 pub fn is_custom_attribute(attr: &SecondaryAttribute, attribute_name: &str) -> bool {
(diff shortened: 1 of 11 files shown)


asterite:
Now that I'm thinking about it... maybe if you have this:

```rust
#[attr]
fn foo() {}

fn attr() {}

fn main() {}
```

then it should compile and run fine? the `attr` function has no arguments so we could invoke it by not passing anything to it (though I don't know if that would be useful 🤔)

jfecher:
> then it should compile and run fine? the attr function has no arguments so we could invoke it by not passing anything to it (though I don't know if that would be useful 🤔)

I think requiring these to have a `FunctionDefinition` parameter is fine, even if they go unused. It helps to classify what type of attribute that function is for if we ever need to do so.

jfecher (approved):
:+1:

PR #5924, opened 2024-09-04
feat: unquote some value as tokens, not as unquote markers

### LINKED ISSUE #5916: Unquoting computed values in comptime fails in certain positions
### Aim

```noir
#[derive_via(derive_serialize)]
trait Serialize<let N: u32> {
    fn serialize(self) -> [Field; N];
}

fn flatten_to_fields(name: Quoted, typ: Type) -> ([Quoted], [Quoted]) {
    // ...not important, aux function that given name/type
    // returns a slice of flattened Fields, for example: self.a, self.b self.nested.a, self.an_array[i] 
    // It also returns aux vars needed to generate said array
    // for example in case strings exist in the struct
    // and we need let field_as_str = str.to_bytes();
    (fields, aux_vars)
}

comptime fn derive_serialize(s: StructDefinition) -> Quoted {
    let typ = s.as_type();
    let (fields, aux_vars) = flatten_to_fields(quote { self }, typ);
    let aux_vars_quote = if aux_vars.len() > 0 {
        let joint = aux_vars.join(quote {;});
        quote { $joint; }
    } else {
        quote {}
    };

    let field_serializations = fields.join(quote {,});
    // Here comes trouble
    let serialized_len = fields.len(); // Setting this to a constant doesn't work either
    quote {
        impl Serialize<$serialized_len> for $typ {
            fn serialize(self) -> [Field; $serialized_len] {
                $aux_vars_quote
                [ $field_serializations ]
            }
        }
    }
}
```
In the above example, serialized_len cannot be unquoted into the trait impl generic, compiler fails with 

```
error: Failed to parse macro's token stream into an expression
   ┌─ src/main.nr:75:1
   │    
75 │ ╭ ╭ #[derive(Serialize)]
76 │ │ │ struct Smol {
77 │ │ │     a: Field,
78 │ │ │     b: Field,
79 │ │ │ }
   │ ╰─│─'
   │   ╰─' expected <, where or { after impl type
   │    
   = The resulting token stream was: impl Serialize < (UnquoteMarker) > for (type) { fn serialize ( self ) -> [ Field ; (UnquoteMarker) ] { [ self . a as Field , self . b as Field ] } }
   = To avoid this error in the future, try adding input validation to your macro. Erroring out early with an `assert` can be a good way to provide a user-friendly error message
```

### Expected Behavior

Example above should work

### Bug

From a slack convo with @jfecher 

> Ahh, ok. So when we unquote expressions we actually usually keep them as expressions to avoid re-typechecking them and stuff. We leave little “UnquoteMarker” tokens where this happens, and these are later picked up by the parser as expressions. What is happening here is that the parser sees an UnquoteMarker in a type position and says “No, I wanted a Type, not an expression :triumph:.” Should be a quick fix.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #5916

## Additional Context

I think there might be other `Value`s that we could convert into tokens, like `String`, but I didn't want to handle those in this PR (should it be `Str` or `RawStr`? etc.). And it's possible that we only need this special logic for integers anyway.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/value.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/value.rs
@@ -429,6 +429,9 @@ impl Value {
         location: Location,
     ) -> IResult<Vec<Token>> {
         let token = match self {
+            Value::Unit => {
+                return Ok(vec![Token::LeftParen, Token::RightParen]);
+            }
             Value::Quoted(tokens) => return Ok(unwrap_rc(tokens)),
             Value::Type(typ) => Token::QuotedType(interner.push_quoted_type(typ)),
             Value::Expr(ExprValue::Expression(expr)) => {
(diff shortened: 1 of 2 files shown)


jfecher (approved):
LGTM. There may be some other tokens we can do but I don't think strings are ever used outside of expressions currently. Getting the integer tokens is most important.

PR #5925, opened 2024-09-04
feat(perf): Remove unused loads in mem2reg and check stores per function

vezenovm:
# Description

Attempt number 2. This copies over the changes from (https://github.com/noir-lang/noir/pull/5905) but without being branched off of https://github.com/noir-lang/noir/pull/5865

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
@@ -116,7 +116,45 @@ struct PerFunctionContext<'f> {
 
     /// Track a value's last load across all blocks.
     /// If a value is not used in anymore loads we can remove the last store to that value.
-    last_loads: HashMap<ValueId, (InstructionId, BasicBlockId)>,
+    last_loads: HashMap<ValueId, (InstructionId, BasicBlockId, u32)>,
+
+    /// Track whether a load result was used across all blocks.
+    load_results: HashMap<ValueId, PerFuncLoadResultContext>,
+
+    /// Track whether a reference was passed into another entry point
+    stores_used_in_calls: HashMap<ValueId, Vec<(InstructionId, BasicBlockId)>>,
+
+    /// Flag for tracking whether we had to perform a re-load as part of the Brillig CoW optimization.
+    /// Stores made as part of this optimization should not be removed.
+    /// We want to catch stores of this nature:
+    /// ```text
+    /// v3 = load v1
+    //  inc_rc v3
+    //  v4 = load v1
+    //  inc_rc v4
+    //  store v4 at v1
+    //  store v3 at v2
+    /// ```
+    ///
+    /// We keep track of an optional boolean flag as we go through instructions.
+    /// If the flag exists it means we have hit a load instruction.
+    /// If the flag is false it means we have processed a single load, while if the flag is true
+    /// it means we have performed a re-load.
+    /// The field is reset to `None` on every instruction that is not a load, inc_rc, dec_rc, or function call.
+    inside_rc_reload: Option<bool>,
+}
+
+#[derive(Debug, Clone)]
+struct PerFuncLoadResultContext {
+    load_counter: u32,
+    load_instruction: InstructionId,
+    instructions_using_result: Vec<(InstructionId, BasicBlockId)>,
+}
+
+impl PerFuncLoadResultContext {
+    fn new(load_instruction: InstructionId) -> Self {
+        Self { load_counter: 0, load_instruction, instructions_using_result: vec![] }
+    }
 }
 
 impl<'f> PerFunctionContext<'f> {
@@ -131,6 +169,9 @@ impl<'f> PerFunctionContext<'f> {
             blocks: BTreeMap::new(),
             instructions_to_remove: BTreeSet::new(),
             last_loads: HashMap::default(),
+            load_results: HashMap::default(),
+            inside_rc_reload: None,
+            stores_used_in_calls: HashMap::default(),
         }
     }
 
@@ -148,6 +189,27 @@ impl<'f> PerFunctionContext<'f> {
             self.analyze_block(block, references);
         }
 
+        let mut loads_removed = HashMap::default();
+        for (_, PerFuncLoadResultContext { load_counter, load_instruction, .. }) in
+            self.load_results.iter()
+        {
+            let Instruction::Load { address } = self.inserter.function.dfg[*load_instruction]
+            else {
+                panic!("Should only have a load instruction here");
+            };
+
+            if *load_counter == 0 {
+                if let Some(counter) = loads_removed.get_mut(&address) {
+                    *counter += 1;
+                } else {
+                    loads_removed.insert(address, 1);
+                }
+
+                self.instructions_to_remove.insert(*load_instruction);
+            }
+        }
+
+        let mut not_removed_stores: HashMap<ValueId, (InstructionId, u32)> = HashMap::default();
         // If we never load from an address within a function we can remove all stores to that address.
         // This rule does not apply to reference parameters, which we must also check for before removing these stores.
         for (block_id, block) in self.blocks.iter() {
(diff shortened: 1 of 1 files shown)


TomAFrench:
currently reviewing this.

TomAFrench (approved):
LGTM, some very minor nits

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -116,13 +116,36 @@ struct PerFunctionContext<'f> {
 
     /// Track a value's last load across all blocks.
     /// If a value is not used in anymore loads we can remove the last store to that value.
-    last_loads: HashMap<ValueId, (InstructionId, BasicBlockId)>,
+    last_loads: HashMap<ValueId, (InstructionId, BasicBlockId, u32)>,
```
jfecher, line 119: Can you add a comment explaining what the `u32` is for

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -116,13 +116,36 @@ struct PerFunctionContext<'f> {
...
+
+    /// Track whether a load result was used across all blocks.
+    load_results: HashMap<ValueId, PerFuncLoadResultContext>,
+
+    /// Track whether a reference was passed into another entry point
+    /// This is needed to determine whether we can remove a store.
+    calls_reference_input: HashSet<ValueId>,
```
jfecher, line 126: Shouldn't passing a reference to a function already mark the reference's value as unknown? What is the second hashset needed for?

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -116,13 +116,36 @@ struct PerFunctionContext<'f> {
...
 
     /// Track whether the last instruction is an inc_rc/dec_rc instruction.
     /// If it is we should not remove any known store values.
     inside_rc_reload: bool,
 }
 
+#[derive(Debug, Clone)]
+struct PerFuncLoadResultContext {
+    /// Reference counter that keeps track of how many times a load was used in other instructions
+    result_counter: u32,
```
jfecher, line 136:
```suggestion
    uses: u32,
```
Seems to make more sense with the comment above

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -290,7 +290,13 @@ impl<'f> PerFunctionContext<'f> {
...
                     references.set_known_value(result, address);
 
-                    self.last_loads.insert(address, (instruction, block_id));
+                    self.load_results.insert(result, PerFuncLoadResultContext::new(instruction));
+
+                    let load_counter = self
+                        .last_loads
+                        .get(&address)
+                        .map_or(1, |(_, _, load_counter)| *load_counter + 1);
+                    self.last_loads.insert(address, (instruction, block_id, load_counter));
```
jfecher, line 299:
```suggestion
                    let last_load = self
                        .last_loads
                        .entry(address)
                        .or_insert((instruction, block_id, 0));
                    last_load.load_counter += 1;
```
Could probably fix the formatting with an intermediate variable if this version is too long.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -364,7 +373,14 @@ impl<'f> PerFunctionContext<'f> {
                     references.aliases.insert(expression, aliases);
                 }
             }
-            Instruction::Call { arguments, .. } => self.mark_all_unknown(arguments, references),
+            Instruction::Call { arguments, .. } => {
+                for arg in arguments {
+                    if self.inserter.function.dfg.value_is_reference(*arg) {
+                        self.calls_reference_input.insert(*arg);
+                    }
```
jfecher, line 380: Looks like this doesn't track aliases of `arg`

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -444,7 +460,15 @@ impl<'f> PerFunctionContext<'f> {
     fn handle_terminator(&mut self, block: BasicBlockId, references: &mut Block) {
         self.inserter.map_terminator_in_place(block);
 
-        match self.inserter.function.dfg[block].unwrap_terminator() {
+        let terminator = self.inserter.function.dfg[block].unwrap_terminator();
+
+        terminator.for_each_value(|value| {
+            if let Some(context) = self.load_results.get_mut(&value) {
+                context.result_counter += 1;
+            }
```
jfecher, line 468: Makes me wonder if these checks all need to handle aliases as well. If we have multiple aliases this looks like it'd only increase the result counter for one of them.

jfecher (changes requested):

jfecher:
Still uncertain about how this PR handles aliases. Makes me think we need much more aliasing tests for mem2reg

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
vezenovm: I did this locally as to take advantage of rename symbol.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
vezenovm: Nice yeah it did improve the formatting

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
jfecher: last stores don't exist in SSA generally so I'm assuming you mean the allocations themselves. Those should already be handled by existing checks

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
vezenovm: I made a context object `PerFuncLastLoadContext` and explain the fields in the struct def.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
vezenovm: Ah so I switched this to `PerFuncLastLoadContext::new` and we have the multi-line formatting again. I kept this logic either way though.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
vezenovm: Yeah so the existing check does not remove these stores. I added this map as when I am doing a function cleanup we may spot this lone store without any loads and remove it as we never load from this address. However, we cannot remove it due to this reference arg case.

jfecher: What if the called function mutates the address to alias it with another?
Would probably require double references like `let mut x = &mut 0` or `let y = &mut &mut 0`.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
vezenovm: I actually realized that this breaks the last loads functionality as we always want to reinsert the last load that we have, so I switched back to the map_or.

jfecher: Why do we want to reinsert? I don't see the difference between the map_or version and the or_insert version above

vezenovm: We use the block_id of the last load to determine whether the last load is in the return block. I could do `last_load.block_id = block_id` with the counter increment, but I found the map_or cleaner for this purpose.

jfecher: Oh, I see we always mutate the instruction and block id. Was missing that thinking we only did it when inserting a new item.

vezenovm:
> Still uncertain about how this PR handles aliases. Makes me think we need much more aliasing tests for mem2reg

I'm going to add some extra aliasing tests then we can check them against this PR and update accordingly. Looks like #5935 is causing some aliasing issues in https://github.com/AztecProtocol/aztec-packages/pull/8378 so I am going to reproduce a small case from those failures.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
vezenovm: For now, I just check whether a reference has been aliased, and other I do not remove that store in the per function cleanup. In the case of just removing load results, it should be safe to remove a load if its result is not used anywhere even if an alias is being loaded. This counter is only being used to remove unused loads.

vezenovm:
> **references**	295 (+8)	**+2.79%**

Pretty surprised by this as this PR should only be adding additional instructions for removal. Perhaps the mapping of instructions during function cleanup is adding back some instructions that are not supposed to be?

test_programs/execution_success/references_testing/src/main.nr
jfecher, line 1: I'm fine with keeping this test but we should rename it

jfecher (commented):
Feel like I don't understand mem2reg anymore.
Can you update the doc comments for this pass to keep them up to date with how it works now?

vezenovm:
> Can you update the doc comments for this pass to keep them up to date with how it works now?

I added some doc comments. Let me know if anything is unclear.

test_programs/execution_success/references_testing/src/main.nr
vezenovm: Renamed to `reference_only_used_as_alias`

jfecher (approved):

PR #5928, opened 2024-09-04
feat: add `fmtstr::contents`

### LINKED ISSUE #5899: Fix formatting of `Quoted` values

### LINKED ISSUE #5914: It's not possible to unquote code to be part of an Ident
### Aim

Using the unquote operator as part of identifiers

### Expected Behavior

Maybe using an escape character or a dedicated method to make this possible?

### Bug

```noir
let name: Quoted = func.name();
let struct_name_based_on_fn_name = quote { $name_suffix }; // this doesn't work, since the var name is now name_suffix
let struct_name_based_on_fn_name = name.append(quote {_suffix}); // this is parsed as separate tokens
```

Those examples should be allowed somehow

### To Reproduce

_No response_

### Workaround

Yes

### Workaround Description

As a hacky workaround, I implemented the following builtin for the `Quoted` struct:

```rust
// fn join_tokens(another: Quoted) -> Quoted
fn quoted_join_tokens(arguments: Vec<(Value, Location)>, location: Location) -> IResult<Value> {
    let (self_value, other_value) = check_two_arguments(arguments, location)?;

    let self_quoted = get_quoted(self_value)?;
    let other_quoted = get_quoted(other_value)?;

    let result = [&self_quoted[..], &other_quoted[..]]
        .concat()
        .iter()
        .map(|token| token.to_string())
        .collect::<Vec<_>>()
        .join("");

    let (tokens, _) = Lexer::lex(result.as_str());

    let tokens: Vec<_> = tokens.0.iter().map(|token| token.clone().into_token()).collect();

    Ok(Value::Quoted(Rc::new(tokens[0..tokens.len() - 1].to_vec())))
}
```

### Additional Context

_No response_

### Project Impact

Blocker

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #5899
Resolves #5914

## Summary

Two things here:
1. When interpolating quotes values inside a format string, we do that without producing the `quote {` and `}` parts, which is likely what a user would expect (similar to unquoting those values).
2. In order to create identifiers (or any piece of code in general) by joining severa quoted values you can use format strings together with the new `fmtstr::contents` method, which returns a `Quoted` value with the string contents (that is, without the leading and trailing double quotes).

## Additional Context

I originally thought about a method like `fmtstr::as_identifier` that would try to parse the string contents as an identifier (maybe an `Ident`, or maybe a `Path`), returning `Option<Quoted>` and `None` in case it couldn't be parsed to that. But I think in general it could be more useful to just get the string contents as a `Quoted` value. After all, if it isn't an identifier you'll learn it later on once the value is unquoted or interpolated.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/docs/docs/noir/standard_library/meta/fmtstr.md
+++ b/docs/docs/noir/standard_library/meta/fmtstr.md
@@ -1,0 +1,13 @@
+---
+title: fmtstr
+---
+
+`std::meta::format_string` contains comptime methods on the `fmtstr` type for format strings.
+
+## Methods
+
+### contents
+
+#include_code quoted noir_stdlib/src/meta/format_string.nr rust
+
+Returns the format string contents (that is, without the leading and trailing double quotes) as a `Quoted` value.
--- a/docs/docs/noir/standard_library/meta/fmtstr.md
+++ b/docs/docs/noir/standard_library/meta/fmtstr.md
@@ -8,6 +8,6 @@ title: fmtstr
 
 ### contents
 
-#include_code quoted noir_stdlib/src/meta/format_string.nr rust
+#include_code contents noir_stdlib/src/meta/format_string.nr rust
 
 Returns the format string contents (that is, without the leading and trailing double quotes) as a `Quoted` value.
--- a/test_programs/compile_success_empty/unquote_struct/src/main.nr
+++ b/test_programs/compile_success_empty/unquote_struct/src/main.nr
@@ -10,11 +10,13 @@ fn foo(x: Field, y: u32) -> u32 {
 
 // Given a function, wrap its parameters in a struct definition
 comptime fn output_struct(f: FunctionDefinition) -> Quoted {
-    let fields = f.parameters().map(|param: (Quoted, Type)| {
+    let fields = f.parameters().map(
+        |param: (Quoted, Type)| {
         let name = param.0;
         let typ = param.1;
         quote { $name: $typ, }
-    }).join(quote {});
+    }
+    ).join(quote {});
 
     quote {
         struct Foo { $fields }
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
@@ -586,7 +586,19 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
                     consuming = false;
 
                     if let Some(value) = values.pop_front() {
-                        result.push_str(&value.display(self.elaborator.interner).to_string());
+                        // When interpolating a quoted value inside a format string, we don't include the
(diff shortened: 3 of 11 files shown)


michaeljklein (approved):

docs/docs/noir/standard_library/meta/fmtstr.md
```diff
@@ -0,0 +1,13 @@
+---
+title: fmtstr
+---
+
+`std::meta::format_string` contains comptime methods on the `fmtstr` type for format strings.
```
jfecher, line 5: Can you move this file to be a general docs for `fmtstr`? Users shouldn't know/care that these methods are in `std::meta::format_string` since they just want to  see all the methods on `fmtstr` in one place.

docs/docs/noir/standard_library/meta/fmtstr.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Methods
+
+### contents
+
+#include_code contents noir_stdlib/src/meta/format_string.nr rust
+
+Returns the format string contents (that is, without the leading and trailing double quotes) as a `Quoted` value.
```
jfecher, line 13: Can you rename this function to `quoted_contents()`? I think that makes it clearer it returns a quoted value, and will help steer away users that are not doing comptime things.

test_programs/compile_success_empty/unquote_struct/src/main.nr
```diff
@@ -10,11 +10,13 @@ fn foo(x: Field, y: u32) -> u32 {
 
 // Given a function, wrap its parameters in a struct definition
 comptime fn output_struct(f: FunctionDefinition) -> Quoted {
-    let fields = f.parameters().map(|param: (Quoted, Type)| {
+    let fields = f.parameters().map(
+        |param: (Quoted, Type)| {
```
jfecher, line 14: Hmm wonder how this got merged in my PR without nargo fmt rejecting it.

Also interesting seeing all the blue now that Noir is a recognized language on github.

PR #5939, opened 2024-09-05
fix: use element_size() instead of computing it with division

### LINKED ISSUE #5936: Array get on empty array panics at compile time
### Aim

Attempted to compile a program where an empty array is accessed within a non-executed `if` statement:

```noir
fn main(x: pub bool) {
    let empty_array: [Field; 0] = [];

    if x {
        let _ = empty_array[0];
    }

    // the same error occurs when the `if` block is elided:
    // let _ = empty_array[0];

    // and when the index is dynamic (x: pub u32)
    // if x == 0 {
    //     let _ = empty_array[x];
    // }

    // and when the array is dynamic (xs: pub [Field; 0])
    // let empty_array: [Field; 0] = xs;
}
```

### Expected Behavior

Expected the program to compile successfully

### Bug

`nargo compile` panics during the DIE pass:

```bash
Message:  attempt to divide by zero
Location: compiler/noirc_evaluator/src/ssa/opt/die.rs:386

```

### To Reproduce

1.
2.
3.
4.

### Workaround

Yes

### Workaround Description

- Locate empty array
- Replace with the singleton array `[std::mem::zeroed()]`
- Add a new parameter to ignore the `std::mem::zeroed()` when needed

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 0.33.0 noirc version = 0.33.0+712468a08dd95eaed42345320e8c3e6751e88c47 (git version hash: 712468a08dd95eaed42345320e8c3e6751e88c47, is dirty: false)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #5936

## Additional Context

I was computing an array's element size in a way that could lead to dividing by zero, but there was already a method to get the element size.

## Documentation

Check one:

# PR Checklist


test_programs/compile_success_empty/empty_composite_array_get/Nargo.toml
TomAFrench, line 1: this should be moved to `execution_failure` as we emit a single opcode which will always fail.

test_programs/execution_failure/empty_composite_array_get/src/main.nr
```diff
@@ -0,0 +1,4 @@
+fn main() {
+    let empty_array: [(Field, Field); 0] = [];
+    let _ = empty_array[0];
```
jfecher, line 3: Can you add an empty array parameter to `main`? I'd like to ensure that doesn't panic in acir-gen as well

test_programs/execution_failure/empty_composite_array_get/src/main.nr
```diff
@@ -1,4 +1,6 @@
-fn main() {
+fn main(empty_input: [(Field, Field); 0]) {
     let empty_array: [(Field, Field); 0] = [];
     let _ = empty_array[0];
+
+    let _ = empty_input[0];
```
jfecher, line 5:
```suggestion
    let _ = empty_input[0];
    let _ = empty_array[0];
```
Reordering this so the empty input fails first to see if we get a panic

jfecher (approved):
LGTM

Issue #5780, opened 2024-08-21
`sort_via` triggers a constraint failure on arrays with any equal elements


jfecher:
### Aim

Sorting any array with 2 identical elements in `sort_via` with the `<` comparison function

### Expected Behavior

The resulting array to be sorted

### Bug

The ordering assertion afterward causes a constraint failure:
```rs
        // Ensure the array is sorted
        for i in 0..N - 1 {
            assert(ordering(result[i], result[i + 1]));
        }
```
If the `ordering` function that was provided returns false for equal elements, such as `<` or `>`.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

michaeljklein:
This will be fixed once the stdlib version of the following PR is merged: https://github.com/noir-lang/noir_sort/pull/21

PR #5940, opened 2024-09-05
feat: use Zac's quicksort algorithm in stdlib sorting

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This pulls in Zac's quicksort algorithm as a replacement for the existing sorting algorithm in the stdlib.

## Documentation\*

Check one:

# PR Checklist\*


noir_stdlib/src/array/mod.nr
```diff
@@ -114,25 +66,39 @@ impl<T, let N: u32> [T; N] {
...
+        unsafe {
+            // Safety: `sorted` array is checked to be:
+            //   a. a permutation of `input`'s elements
+            //   b. satisfying the predicate `ordering`
+            let sorted = quicksort::quicksort(self, ordering);
+
+            if !is_unconstrained() {
+                for i in 0..N - 1 {
+                    assert(ordering(sorted[i], sorted[i + 1]));
+                }
```
jfecher, line 87: Looks like we'd still be failing if arrays contained any equal elements: https://github.com/noir-lang/noir/issues/5780

Before it was just a sanity check but now if we're relying on this to constrain the sorted outputs then maybe we need documentation that the sort function must be a `<=` instead of a `<`?

noir_stdlib/src/array/mod.nr
```diff
@@ -114,25 +66,44 @@ impl<T, let N: u32> [T; N] {
...
+    pub fn sort(self) -> Self {
+        self.sort_via(|a: T, b: T| a <= b)
+    }
+}
+
+impl<T, let N: u32> [T; N] where T: Eq {
+
+    /// Sorts the array using a custom predicate function `ordering`.
+    ///
+    /// # Safety
```
TomAFrench, line 79:
```suggestion
```

noir_stdlib/src/array/mod.nr
TomAFrench: I've added some documentation for this issue with duplicate elements.

docs/docs/noir/concepts/data_types/arrays.md
```diff
@@ -128,7 +128,7 @@ fn main() {
 
 ### sort_via
 
-Sorts the array with a custom comparison function
+Sorts the array with a custom comparison function. The ordering function must return true if the first argument should be sorted to be before the second argument, otherwise it should return false.
```
jfecher, line 131:
```suggestion
Sorts the array with a custom comparison function. The ordering function must return true if the first argument should be sorted to be before the second argument or is equal to the second argument.

Using this method with an operator like `<` that does not return `true` for equal values will result in an assertion failure for arrays with equal elements.
```

jfecher (changes requested):
We should fix the sort_via example as well which still uses `<` and `>` (can't suggest changes there).

TomAFrench:
One thing that occurs to me is that this new implementation is not safe in the situation where the `Eq` implementation does not assert a direct equality over every constituent field of the types being sorted. In this situation a malicious prover can mutate any fields not constrained in the `Eq` implementation.

This may seem niche but have been a couple of instances where an `Eq` like this has been used in Aztec code in the past.

TomAFrench:
It's also not specific to this change but core to the "unconstrained helper and then check equality" workflow so shouldn't be a blocker.

jfecher:
I wonder if we can fix that issue generally somehow?
E.g. a second Eq trait with the semantics of always constraining each field? `ConstrainEq`?

TomAFrench:
@jfecher I've updated the documentation to use a non-strict inequality.

jfecher (approved):
👍

PR #5942, opened 2024-09-05
feat: Add `Quoted::tokens`

Thunkar:
# Description

## Problem\*

While `fmtstr.quoted_contents` is great for 99% of use cases, sometimes the raw str (including double quotes) is needed. Consider this case (fn signature generation in `aztec.nr`)

```noir 
pub(crate) comptime fn compute_fn_selector(f: FunctionDefinition) -> Field {
    let fn_name = f.name();
    let args_signatures = f.parameters().map(
        | (_, typ): (Quoted, Type) | {
        signature_of_type(typ)
    }
    ).join(quote {,});
    let signature = f"{fn_name}({args_signatures})".quoted_contents();
    // from_signature takes a str, quoted_contents is no good here!
    let computation_quote = quote {
        protocol_types::abis::function_selector::FunctionSelector::from_signature($signature).to_field()
    };
    unquote!(computation_quote)
}
```

We have a bunch of `Quoted` values obtained during comptime that we have to string (ha) together. The result is then needed to compute a hash. 

## Summary\*
~~This just adds a builtin that "resolves" the `fmtstr` and then turns it into a `Token::Str`~~ 

It's possible to create a string from quoted values by doing:

```noir
   let signature = f"{name}({args})";
   let signature_as_str = unquote!(quote { $signature });
```

But this includes whitespaces around the tokens. To get rid of them, ended up going with the approach suggested by @jfecher, which certainly complicates user code, but avoids the "eating whitespaces" problem:

> Going forward I think we'll need to instead control the formatting of Quoted values more. The compiler itself can never guess what exact formatting users will want of them since spaces are already not present. So I'm leaning towards a Quoted::tokens(self) -> [Quoted] method to return a slice of each individual token in the Quoted value. Then users could iterate over it and apply (or not apply) spacing in between each token as they see fit.

Also, the returned values are `str<_>`, since the compiler has no idea of the returned length after formatting. This means values can only be used by unquoting them and not directly.

## Additional Context

My first intuition was to do

```noir
let signature = f"\"{quotedvalues}\""
```

But it doesn't work =/

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
@@ -97,6 +97,7 @@ impl<'local, 'context> Interpreter<'local, 'context> {
             "expr_resolve" => expr_resolve(self, arguments, location),
             "is_unconstrained" => Ok(Value::Bool(true)),
             "fmtstr_quoted_contents" => fmtstr_quoted_contents(interner, arguments, location),
+            "fmtstr_quoted" => fmtstr_quoted(interner, arguments, location),
             "function_def_body" => function_def_body(interner, arguments, location),
             "function_def_has_named_attribute" => {
                 function_def_has_named_attribute(interner, arguments, location)
@@ -1593,6 +1594,28 @@ fn fmtstr_quoted_contents(
     }
 
     Ok(Value::Quoted(Rc::new(tokens)))
+}
+
+// fn quoted(self) -> Quoted
+fn fmtstr_quoted(
+    interner: &NodeInterner,
+    arguments: Vec<(Value, Location)>,
+    location: Location,
+) -> IResult<Value> {
+    let self_argument = check_one_argument(arguments, location)?;
+    let (string, _) = get_format_string(interner, self_argument)?;
+    let (tokens, _) = Lexer::lex(&string);
+    let str = tokens
+        .0
+        .into_iter()
+        .filter(|token| !matches!(token.clone().into_token(), Token::EOF))
+        .map(|token| token.into_token().display(interner).to_string())
+        .collect::<Vec<_>>()
+        .join("");
+
+    let token = Token::Str(str);
+
+    Ok(Value::Quoted(Rc::new(vec![token])))
 }
 
 // fn body(self) -> Expr
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
@@ -536,7 +536,7 @@ fn quoted_as_type(
     Ok(Value::Type(typ))
 }
 
-// fn as_expr(quoted: Quoted) -> Option<Expr>
+// fn tokens(quoted: Quoted) -> [Quoted]
 fn quoted_tokens(arguments: Vec<(Value, Location)>, location: Location) -> IResult<Value> {
     let argument = check_one_argument(arguments, location)?;
     let value = get_quoted(argument)?;
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
@@ -33,7 +33,7 @@ use crate::{
     },
     hir_def::function::FunctionBody,
     lexer::Lexer,
-    macros_api::{HirExpression, HirLiteral, ModuleDefId, NodeInterner, Signedness},
+    macros_api::{HirExpression, HirLiteral, Ident, ModuleDefId, NodeInterner, Signedness},
     node_interner::{DefinitionKind, TraitImplKind},
     parser::{self},
     token::Token,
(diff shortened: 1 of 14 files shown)


jfecher (changes requested):
This is a string formatting issue, not a quoting format strings issue.

As discussed on slack the issue is that formatting quoted values into a format string will add spaces between each token which breaks your use case. `fmtstr_quoted` here will lex the entire format string, then append each token together as a string with no separator. This would remove all the spaces in the format string, even for the static string contents which haven't been formatted. E.g. `f"    ".fmtstr_quoted()` would return an empty string. Two identifiers: `f"foo bar".fmtstr_quoted()` would also be collapsed to just one with this method: `"foobar"

Going forward I think we'll need to instead control the formatting of Quoted values more. The compiler itself can never guess what exact formatting users will want of them since spaces are already not present. So I'm leaning towards a `Quoted::tokens(self) -> [Quoted]` method to return a slice of each individual token in the Quoted value. Then users could iterate over it and apply (or not apply) spacing in between each token as they see fit.

asterite:
Now that I finished writing the above, I realized this can be done like this:

```rust
use std::meta::unquote;

fn main() {
comptime
    {
        let name = "hello";
        let fn_name_as_str_quoted = f"{name}";
        let fn_name_len: u32 = unquote!(quote { "$fn_name_as_str_quoted".as_bytes().len()});
    }
}
```

/cc @Thunkar

jfecher:
@asterite You don't need the `as_bytes` trick, you can just quote and unquote the format string:
```rs
fn main() {
    comptime {
        let name = quote { add };
        let args = quote { 1, 2 };
        let signature = f"{name}({args})";
        print_str(std::meta::unquote!(quote { $signature }));
    }
}

fn print_str<let N: u32>(s: str<N>) {
    println(s);
}
```
When you unquote a format string, it's unquoted as a normal string literal since all the formatting is already done.

jfecher:
Looks like this is a corner case of lowering versus unquoting where the two differ.
- Unquoting: When we use `$foo` in a quote to splice in a value into the token stream
- Lowering: When we have a comptime value and lower it into "runtime" code. Lowering also applies to higher-comptimed items. E.g. the `unquote!` macro calls when we're already in a comptime context still perform lowering since that is how macros are inserted.

I'll make an edit to the docs to include this case.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -97,6 +97,7 @@ impl<'local, 'context> Interpreter<'local, 'context> {
             "expr_resolve" => expr_resolve(self, arguments, location),
             "is_unconstrained" => Ok(Value::Bool(true)),
             "fmtstr_quoted_contents" => fmtstr_quoted_contents(interner, arguments, location),
+            "fmtstr_quoted" => fmtstr_quoted(interner, arguments, location),
```
jfecher, line 100:
```suggestion
```

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -1654,6 +1667,28 @@ fn fmtstr_quoted_contents(
...
+        .filter(|token| !matches!(token.clone().into_token(), Token::EOF))
+        .map(|token| token.into_token().display(interner).to_string())
+        .collect::<Vec<_>>()
+        .join("");
+
+    let token = Token::Str(str);
+
+    Ok(Value::Quoted(Rc::new(vec![token])))
+}
+
```
jfecher, line 1691:
```suggestion
```

asterite:
I think there's still no way to generate strings in `quote` 🤔 

For example, how to do this?

```noir
fn main() {
    comptime
    {
        let values = [quote { a }, quote { b }];

        // Generate this body from the above values:
        quote {
            println("a, b");
        }
    }
}
```

Or at least I can't find a way to generate a string literal with content from quoted values.

There is a way to geneate a quoted value from other quoted values, using `f"..."`, but there's no way to turn that into a string literal.

I think the first commit in this PR did that: allow turning a quoted value into a string token with.

As a side note, I found out that `f"..."` inside `quote` doesn't work well:

```noir
fn main() {
    comptime
    {
        println(quote {
            f"Hello"
        })
    }
}
```

The above prints `quote { fHello }`.

/cc @jfecher

jfecher:
@asterite The first example can be done by moving the format string:
```rs
comptime fn foo() -> Quoted {
    let values = [quote { a }, quote { b }];

    let a = values[0];
    let b = values[1];
    let s = f"{a}, {b}";
    quote {
        println($s);
    }
}
```
Although it seems like there is a bug preventing me from calling `foo` from main:
```
fn main() {
    // error: Comptime-only type `Quoted` used in runtime code
    foo!();
}
```
Which is odd since we use a macro call to unquote it. I'd think this would be caught in one of the existing macro tests so I'm unsure what this is.

Formatting with an arbitrary number of slice elements is more difficult. IIRC Grego had a rather hacky solution to it. We probably should add a compile-time String type.

The format string printing issue is a result of our change to print strings without their surrounding quotes. Not sure what to do there, maybe we just print them with quotes when we're within a quoted value.

asterite:
> Formatting with an arbitrary number of slice elements is more difficult. IIRC Grego had a rather hacky solution to it. We probably should add a compile-time String type.

Right, a compile-time String might solve this.

Until them, I'm a bit blocked. The context is that I need to generate a function signature from its parameters, and I get these extra spaces. I guess for now I'll go with the extra spaces and we can fix that once we have compile-time strings.

PR #5947, opened 2024-09-05
feat: `Module::add_item`

### LINKED ISSUE #5875: Add `Module::add_item`
### Problem

(From a slack request)

Given just a `Module` metaprogramming object, there is no way to add additional functions or types to the module.

### Happy Case

We add the function:
```
impl Module {
    fn add_item(self, item: Quoted) {}
}
```
To add a top-level item to the module.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #5875

## Documentation

Check one:

# PR Checklist


asterite:
@jfecher I think I'm going to need some help on this one. Here's the code I'm trying it with:

```noir
#[run]
mod other {}

struct Foo {

}

fn run(m: Module) {
    m.add_item(
        quote {pub fn get_foo() -> Foo {
            Foo {}
        }}
    );
}

fn main() {
    let foo = other::get_foo();
    println(foo);
}
```

The issue is that `Foo` isn't found. I'm trying to elaborate the function as if it's in `run`, and I'm trying to add it to `other`. I tried something but it doesn't seem to be working (it's not meant as the definitive code, I was just trying some things).

docs/docs/noir/standard_library/meta/module.md
```diff
@@ -8,6 +8,12 @@ declarations in the source program.
 
 ## Methods
 
+### add_item
+
+#include_code add_item noir_stdlib/src/meta/module.nr rust
+
+Adds a top-level item to the module.
```
jfecher, line 15: Can you give some examples here for users that may not know what a top-level item is? Functions, structs, trait impls

compiler/noirc_frontend/src/elaborator/comptime.rs
```diff
@@ -89,6 +89,39 @@ impl<'context> Elaborator<'context> {
...
     }
 
+    pub fn elaborate_item_in_module<'a, T>(
+        &'a mut self,
+        module: ModuleId,
+        f: impl FnOnce(&mut Elaborator<'a>) -> T,
+    ) -> T {
+        // Create a fresh elaborator to ensure no state is changed from
+        // this elaborator
+        let mut elaborator = Elaborator::new(
```
jfecher, line 99: Can you refactor this so that this and `elaborate_item_from_comptime` share a common helper method? Probably with an optional function and module? Although I guess it doesn't make sense to have both. Maybe a lambda for setting up the fresh elaborator then?

jfecher (commented):
Not sure why you're getting name resolution issues. I'll take another look at it tomorrow but I'd think since you ran the dc_mod functions in the module's scope then the function in main should at least see the function in the module.

jfecher (approved):
:+1:

PR #5949, opened 2024-09-05
fix(frontend): Ban type vars bound to a reference from passing the unconstrained boundary 

vezenovm:
# Description

## Problem\*

Resolves 

Quick fix found while searching for other bugs

## Summary\*

I have provided an example test which would previously panic during ACIR gen about all references not being resolved. However, we should have been allowing the code in the first place as it was attempting to pass a mutable reference from the constrained boundary to the unconstrained boundary. 

## Documentation\*

Check one:

# PR Checklist\*


jfecher (approved):
LGTM

PR #5956, opened 2024-09-06
feat: add `FunctionDefinition::module` and `StructDefinition::module`

### LINKED ISSUE #5953: Access current module data in `FunctionDefinition` and `StructDefinition`
### Problem

It's sometimes necessary to learn about module data inside a func/struct annotation. This is required for initializer checks in `aztec.nr`: Inside `#[private]` annotation we have to check whether any other function in the contract is an initializer.

### Happy Case

The `Module` object in which the current fn/struct lives is available via annotations:

```noir
#[my_struct_annotation]
struct foo {}

#[my_fn_annotation]
fn bar() {}

comptime fn my_fn_annotation(f: FunctionDefinition) {
  let m: Module = f.get_module();
}

comptime fn my_struct_annotation(s: StructDefinition) {
  let m: Module = s.get_module();
}
```

### Workaround

Yes

### Workaround Description

It's probably possible to modify all functions in a top-level annotation in the contract, but it is a lot uglier since we would have a giant single annotation and would lose debugability.

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #5953

## Documentation

Check one:

# PR Checklist


test_programs/compile_success_empty/comptime_struct/Nargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "comptime_struct"
```
jfecher, line 2: We already have a test for structs in comptime_type_definition, from before the type was renamed. Maybe we should move it to `comptime_struct_definition`?

jfecher (approved):
LGTM, one note

test_programs/compile_success_empty/comptime_struct/Nargo.toml
asterite: Good point. I searched for an existing one but couldn't find it. I'll merge both.

PR #5964, opened 2024-09-06
feat: add `Expr::as_let`

asterite:
# Description

## Problem

Part of #5668

## Summary

Adds `Expr::as_let`, but also introduces a new `ExprValue` for patterns, their inlining code, etc.

## Additional Context

I wonder if `Pattern` should be an `Expr` or something else, because it can't be converted to an expression like a statement or like an LValue 🤔 

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -1789,6 +1825,9 @@ fn expr_resolve(
                 let (expr_id, _) = elaborator.elaborate_expression(expr);
                 Value::TypedExpr(TypedExpr::ExprId(expr_id))
             }
+            ExprValue::Pattern(_) => {
+                todo!("Can't resolve a pattern");
```
asterite, line 1829: Should we give a compile-error here? What should it say? Or maybe `Pattern` should be a separate type?

jfecher: If we can't use them in other places where we expect to use expressions then I suppose we can just make them a separate type.

It's a bit unfortunate since I think the only variant that can't be represented as an expression would be the `mut` pattern. We could potentially have a `try_as_expression` I suppose then error if it can't be converted.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -2013,6 +2055,9 @@ fn function_def_set_body(
         }),
         ExprValue::Statement(statement_kind) => statement_kind,
         ExprValue::LValue(lvalue) => StatementKind::Expression(lvalue.as_expression()),
+        ExprValue::Pattern(_) => {
+            todo!("Can't set function body to a pattern");
```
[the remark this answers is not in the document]
jfecher: Same case applies here. Either:
- Pattern is a new type: always error
- We have `try_as_expression`: we try to convert and error if we can't

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -1807,30 +1807,32 @@ fn expr_resolve(
...
+            Ok(Value::TypedExpr(TypedExpr::StmtId(stmt_id)))
+        }
+        ExprValue::LValue(lvalue) => {
+            let expr = lvalue.as_expression();
+            let (expr_id, _) = elaborator.elaborate_expression(expr);
+            Ok(Value::TypedExpr(TypedExpr::ExprId(expr_id)))
+        }
+        ExprValue::Pattern(pattern) => {
+            if let Some(expression) = pattern.try_as_expression(elaborator.interner) {
+                Ok(Value::expression(expression.kind))
```
jfecher, line 1834: This should probably resolve to a typed expr still (HirPattern) but we can save that for another PR.

jfecher (approved):

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
asterite: Ah, good catch! I pushed another commit in this PR.

asterite: Apparently the commit didn't go out 😅

PR #5965, opened 2024-09-06
fix: let `derive(Eq)` work for empty structs

asterite:
# Description

## Problem

While playing around with the derive attribute I noticed `derive(Eq)` fails to compile for an empty struct.

## Summary

Fixes the above.

## Documentation

Check one:

# PR Checklist


jfecher (approved):
Good catch

PR #5966, opened 2024-09-06
feat: add a `panic` method to the stdlib

### LINKED ISSUE #5849: Add `panic` method to stdlib
### Problem

We should add a `panic` method to stdlib as a way to halt computation and return a value of any type, since we know it can't be used. `assert(false)` alone is insufficient since it returns a unit type.

### Happy Case

We add the function:
```rs
fn panic<T, U, let N: u32>(message: fmtstr<N, T>) -> U {
    assert(false, message);
    std::meta::zeroed()
}
```
The prelude would be a good place for it since users will be used to `panic!()` being available in rust without import.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #5849

## Summary

Adds a `panic` method exactly as proposed in #5849 , and uses it in a couple of places where `panic` would have been useful to have.

## Additional Context

Is it okay in `std::panic::panic`? Where should we document this? (maybe in a new `panic.md` file, or somewhere else?)

## Documentation\*

Check one:

# PR Checklist

### DIFF
--- a/noir_stdlib/src/meta/op.nr
+++ b/noir_stdlib/src/meta/op.nr
@@ -40,7 +40,7 @@ impl UnaryOp {
             quote { * }
         } else {
             let op = self;
-            panic(f"Unexpected unary operator in UnaryOp::quoted: {op}")
+            std::panic::panic(f"Unexpected unary operator in UnaryOp::quoted: {op}")
         }
     }
 }
@@ -183,7 +183,7 @@ impl BinaryOp {
             quote { % }
         } else {
             let op = self;
-            panic(f"Unexpected binary operator in BinaryOp::quoted: {op}")
+            std::panic::panic(f"Unexpected binary operator in BinaryOp::quoted: {op}")
         }
     }
 }
--- a/noir_stdlib/src/meta/op.nr
+++ b/noir_stdlib/src/meta/op.nr
@@ -40,7 +40,7 @@ impl UnaryOp {
             quote { * }
         } else {
             let op = self;
-            std::panic::panic(f"Unexpected unary operator in UnaryOp::quoted: {op}")
+            crate::panic::panic(f"Unexpected unary operator in UnaryOp::quoted: {op}")
         }
     }
 }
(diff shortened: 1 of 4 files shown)


noir_stdlib/src/meta/op.nr
```diff
@@ -39,7 +39,8 @@ impl UnaryOp {
         } else if self.is_dereference() {
             quote { * }
         } else {
-            crate::mem::zeroed()
+            let op = self;
+            std::panic::panic(f"Unexpected unary operator in UnaryOp::quoted: {op}")
```
jfecher, line 43: The errors you're getting are because we have to use `crate::` for the current crate rather than std
```suggestion
            crate::panic::panic(f"Unexpected unary operator in UnaryOp::quoted: {op}")
```

noir_stdlib/src/meta/op.nr
```diff
@@ -181,7 +182,8 @@ impl BinaryOp {
         } else if self.is_modulo() {
             quote { % }
         } else {
-            crate::mem::zeroed()
+            let op = self;
+            std::panic::panic(f"Unexpected binary operator in BinaryOp::quoted: {op}")
```
jfecher, line 186:
```suggestion
            crate::panic::panic(f"Unexpected binary operator in BinaryOp::quoted: {op}")
```

noir_stdlib/src/panic.nr
```diff
@@ -0,0 +1,4 @@
+pub fn panic<T, U, let N: u32>(message: fmtstr<N, T>) -> U {
+    assert(false, message);
+    std::mem::zeroed()
```
jfecher, line 3:
```suggestion
    crate::mem::zeroed()
```

noir_stdlib/src/meta/op.nr
```diff
@@ -39,7 +39,8 @@ impl UnaryOp {
         } else if self.is_dereference() {
             quote { * }
         } else {
-            crate::mem::zeroed()
+            let op = self;
+            crate::panic::panic(f"Unexpected unary operator in UnaryOp::quoted: {op}")
```
jfecher, line 43: Does interpolating `self` not work? Can you create an issue for that if so?

jfecher (approved):

noir_stdlib/src/meta/op.nr
asterite: Ah, I initially wrote `{self.op}` 🤦 . I think `{self}` will work fine.

PR #5968, opened 2024-09-06
feat: show doc comments in LSP

### LINKED ISSUE #5797: Reorganize `std` docs so that they follow the same tree structure that the `std` crate does
### Problem

Organization of functions & other items in the standard library docs for Noir is arbitrary.

### Happy Case

The organization should mimic the actual modules in the std crate itself. This will  also give a better idea to users on how the stdlib is structured.

For bonus points these docs could be autogenerated from doc comments on stdlib items.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Part of #5797

## Summary

lsp-doc-comments

This is just an experiment! 

We now parse doc comments, like we parse attributes, and attach them to AST nodes. From there we store them in `NodeInterner` (for now only if in LSP mode). Then we retrieve those comments for hover and autocompletion.

## Additional Context

I decided to always store the doc comments in the NodeInterner.

I've been thinking and maybe the doc comments shouldn't be parsed as `String`, but instead be a `Span`. Then we wouldn't need to allocate a string for them, and only read from the source file when needed. But... that's an optimization that can be done later.

After merging this, the next steps could be:
1. Copy the doc comments found in the many `.md` files into the standard library (they will be duplicated, but just for a wihle)
2. Create a script or something that generates the `.md` files from the doc comments

Note: I removed the lalrpop code because I got some errors and I thought maybe it's not very useful to keep that code around if we have to make changes to it to maintain it, but it's not working (we could always revive it from git's history). But let me know if this is wrong.

## Documentation

Check one:

# PR Checklist


TomAFrench (approved):
Very excited by the prospect of getting automatic docs generation!

PR #5969, opened 2024-09-09
feat: let `nargo` and LSP work well in the stdlib

asterite:
# Description

## Problem

`nargo test` doesn't work in the standard library, and LSP gives lots of (incorrect) errors when opening the standard library.

## Summary

The main issue is that when you run `nargo` against the standard library, a copy of it is also loaded as the standard library. Then we get duplicate definitions, etc.

I tried several things to solve this and one that worked and ended up making a bit of sense is having a crate be the root and the standard library at the same time.

Now with this PR we can run `nargo test`, LSP works well, and you can even run tests directly from the editor, which could speed up the stdlib development.

## Additional Context

1. 7 warnings still remain in the standard library (all unused functions) but I'd like to solve that in a separate PR.
2. If this gets merged we could eventually change CI to compile nargo, then run `nargo test`, instead of running stdlib tests as part of one crate's tests.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_driver/src/lib.rs
+++ b/compiler/noirc_driver/src/lib.rs
@@ -212,17 +212,19 @@ fn add_debug_source_to_file_manager(file_manager: &mut FileManager) {
 
 /// Adds the file from the file system at `Path` to the crate graph as a root file
 ///
-/// Note: This methods adds the stdlib as a dependency to the crate.
-/// This assumes that the stdlib has already been added to the file manager.
+/// Note: If the stdlib dependency has not been added yet, it's added. Otherwise
+/// this method assumes the root crate is the stdlib (useful for running tests
+/// in the stdlib, getting LSP stuff for the stdlib, etc.).
 pub fn prepare_crate(context: &mut Context, file_name: &Path) -> CrateId {
     let path_to_std_lib_file = Path::new(STD_CRATE_NAME).join("lib.nr");
-    let std_file_id = context
-        .file_manager
-        .name_to_id(path_to_std_lib_file)
-        .expect("stdlib file id is expected to be present");
-    let std_crate_id = context.crate_graph.add_stdlib(std_file_id);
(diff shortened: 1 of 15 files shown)


tooling/lsp/src/requests/hover.rs
```diff
@@ -353,7 +352,7 @@ fn format_parent_module_from_module_id(
 
     // We don't record module attriubtes for the root module,
```
TomAFrench, line 353:
```suggestion
    // We don't record module attributes for the root module,
```

compiler/noirc_driver/src/lib.rs
```diff
@@ -212,18 +212,20 @@ fn add_debug_source_to_file_manager(file_manager: &mut FileManager) {
...
-        .expect("stdlib file id is expected to be present");
-    let std_crate_id = context.crate_graph.add_stdlib(std_file_id);
+    let std_file_id = context.file_manager.name_to_id(path_to_std_lib_file);
+    let std_crate_id = std_file_id.map(|std_file_id| context.crate_graph.add_stdlib(std_file_id));
 
     let root_file_id = context.file_manager.name_to_id(file_name.to_path_buf()).unwrap_or_else(|| panic!("files are expected to be added to the FileManager before reaching the compiler file_path: {file_name:?}"));
 
+    let Some(std_crate_id) = std_crate_id else {
+        return context.crate_graph.add_crate_root_and_stdlib(root_file_id);
+    };
```
TomAFrench, line 227: We can do an if-else here to avoid the early return.

TomAFrench (approved):

jfecher (approved):
👍

asterite:
Another nice thing I forgot to mention is that now when working on the stdlib in VS Code, autocompletions will never suggest `std` as that doesn't exist, and they'll always suggest `crate::` instead :-)

PR #5974, opened 2024-09-09
chore: document BoundedVec

asterite:
# Description

## Problem

Part of #5797

## Summary

This PR copies the docs from the `.md` file to the `.nr` file.

## Additional Context

I'll be sending one PR per `.md` file. It looks like they are 36 in total so it will take some time to port all the docs, but not that much. And this way reviewing these PRs will be more manageable.

For now examples are copied inline into the `.nr` file. Eventually we could maybe have `nargo test` run or at least type-check these example

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/noir_stdlib/src/collections/bounded_vec.nr
+++ b/noir_stdlib/src/collections/bounded_vec.nr
@@ -50,14 +50,14 @@ impl<T, let MaxLen: u32> BoundedVec<T, MaxLen> {
     ///     let v1: BoundedVec<Field, 3> = BoundedVec::new();
     ///     let v2 = BoundedVec::new();
     ///
-    ///     // Ok! MaxLen is known from the type of foo's return value
+    ///     // Ok! MaxLen is known from the type of `good`'s return value
     ///     v2
     /// }
     ///
     /// fn bad() {
+    ///     // Error: Type annotation needed
+    ///     // The compiller can't infer `MaxLen` from the following code:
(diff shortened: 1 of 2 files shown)


noir_stdlib/src/collections/bounded_vec.nr
```diff
@@ -1,66 +1,271 @@
...
+    ///     // Ok! MaxLen is known from the type of foo's return value
+    ///     v2
+    /// }
+    ///
+    /// fn bad() {
+    ///     let mut v3 = BoundedVec::new();
+    ///
+    ///     // Not Ok! We don't know if v3's MaxLen is at least 1, and the compiler often infers 0 by default.
+    ///     v3.push(5);
+    /// }
```
jfecher, line 62: Oh, looks like we should update the docs here. The compiler should now error if a generic is not specified and can't be inferred.

jfecher (approved):

noir_stdlib/src/collections/bounded_vec.nr
asterite: Thanks! I updated the `.md` too.

For now I'm leaving code that won't compile in the example code, uncommented. If we eventually can type-check or run this code, it will fail... but I guess then we can have a special directive to say "this shouldn't compile".

jfecher: I made an issue for now to update the outdated docs

PR #5978, opened 2024-09-09
fix: Error when `quote` is used in runtime code

jfecher:
# Description

## Problem\*

Resolves 

## Summary\*

Fixes a few small issues:
- Error when `quote` is used in runtime code
  - Had to update some stdlib code for this
  - This could maybe be an issue in the future for something like `impl Append for Quoted` which relies on this. These impl methods have to be comptime now which we currently allow even though the parent trait says nothing about it. At least with https://github.com/noir-lang/noir/pull/5976 we'd get an error if one of these accidentally sneaks into runtime code.
- `self.in_comptime_context()` wasn't being tracked accurately in the case of attributes and globals where the function context would only be of length 2 due to there being no outer function. I changed the check to make it more direct instead of relying on indirect changes.
- Comptime globals weren't evaluated in their own function context before so we waited to resolve traits until the end of the program. This could lead to trait errors when interpreting them.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/comptime.rs
+++ b/compiler/noirc_frontend/src/elaborator/comptime.rs
@@ -159,18 +159,18 @@ impl<'context> Elaborator<'context> {
         generated_items: &mut CollectedItems,
     ) {
         if let SecondaryAttribute::Custom(attribute) = attribute {
-            self.function_context.push(FunctionContext::default());
-            if let Err(error) = self.run_comptime_attribute_name_on_item(
-                &attribute.contents,
-                item.clone(),
-                span,
-                attribute.contents_span,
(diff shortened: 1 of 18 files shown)


asterite (approved):
Looks good!

asterite:
One thought about this: should functions with comptime-types not marked as `comptime` produce an error? For example this:

```noir
// Error: fn must be comptime because FunctionDefinition is comptime
fn foo(x: FunctionDefinition) {}
```

jfecher:
> One thought about this: should functions with comptime-types not marked as comptime produce an error? For example this:

Yep! That's the issue I plan on working on next

PR #5980, opened 2024-09-09
feat: add `Expr::as_constructor`

asterite:
# Description

## Problem

Part of #5668

## Summary

Also add doc comments for `expr.nr`.

## Additional Context

I wasn't sure how to represent a `Path`... so I did it with a `Quoted` value. I first thought about using a `Variable`, but then that wouldn't be parsed right when parsing a constructor, so it had to be some kind of Path.

## Documentation

Check one:

# PR Checklist


noir_stdlib/src/meta/expr.nr
```diff
@@ -1,131 +1,209 @@
...
     #[builtin(expr_as_comptime)]
     // docs:start:as_comptime
     fn as_comptime(self) -> Option<[Expr]> {}
     // docs:end:as_comptime
 
+    /// If this expression is a constructor `Type { field1: expr1, ..., fieldN: exprN }`,
+    /// return the type and the fields.
+    #[builtin(expr_as_constructor)]
+    // docs:start:as_constructor
+    fn as_constructor(self) -> Option<(Quoted, [(Quoted, Expr)])> {}
```
jfecher, line 72: I think we can return the Path as a UnresolvedType in this case.

Would probably have to add some support for re-parsing those when we're working on parsing a constructor but I think it would be clearer from a user's perspective than just a Quoted value.

jfecher: > And then, if we unquote it, when we parse it as a constructor we can't know if that UnresolvedType is actually a Named one or not.

This sounds fine to me, we could check during name resolution assuming it is interned.

I'm debating changing the constructor parser itself to match as well. Maybe it should always parse a type then later on we just confirm it is a struct type.

asterite: Done!

compiler/noirc_frontend/src/ast/mod.rs
```diff
@@ -344,6 +344,19 @@ impl UnresolvedType {
     pub(crate) fn is_type_expression(&self) -> bool {
         matches!(&self.typ, UnresolvedTypeData::Expression(_))
     }
+
+    pub fn from_path(mut path: Path) -> Self {
+        let span = path.span;
+        let last_segment = path.segments.last_mut().unwrap();
+        let generics = std::mem::take(&mut last_segment.generics);
```
jfecher, line 351:
```suggestion
        let generics = last_segment.generics.take();
```

compiler/noirc_frontend/src/elaborator/patterns.rs
```diff
@@ -188,7 +188,10 @@ impl<'context> Elaborator<'context> {
             Some(Type::Struct(struct_type, generics)) => (struct_type, generics),
             None => return error_identifier(self),
             Some(typ) => {
-                self.push_err(ResolverError::NonStructUsedInConstructor { typ, span });
+                self.push_err(ResolverError::NonStructUsedInConstructor {
+                    typ: typ.to_string(),
+                    span,
+                });
```
jfecher, line 194:
```suggestion
                let typ = typ.to_string();
                self.push_err(ResolverError::NonStructUsedInConstructor { typ, span });
```

compiler/noirc_frontend/src/hir/comptime/errors.rs
```diff
@@ -394,8 +394,7 @@ impl<'a> From<&'a InterpreterError> for CustomDiagnostic {
                 let mut diagnostic =
                     CustomDiagnostic::simple_error(primary, secondary, location.span);
 
-                // Only take at most 3 frames starting from the top of the stack to avoid producing too much output
-                for frame in call_stack.iter().rev().take(3) {
+                for frame in call_stack.iter().rev() {
```
jfecher, line 397: Why was the `.take(3)` limit removed?

jfecher (commented):
LGTM, just a question on the call stack before I approve

compiler/noirc_frontend/src/hir/comptime/errors.rs
asterite: Oh, I forgot to mention this. With just 3 sometimes I couldn't understand the main source of the error. I changed it to 5 first and that worked, but I feel that it depends on each case. Maybe for now we can leave it unlimited until we find it's a issue.

jfecher: I think we should still have some bound on it. I don't want cases where the user is 20+ functions deep and getting such an error. Maybe just bump it up to 5 for now?

jfecher (approved):

TomAFrench:
Remove from merge queue so I can make a release.

PR #5988, opened 2024-09-10
feat: stdlib docs generator

asterite:
# Description

## Problem

Part of #5797

## Summary

This is just an experiment to see if we can already generate stdlib docs from our code.

Pending:
- [ ] Before being able to merge this, we need to document all the standard library (copy doc comments from `.md` files into `.nr` files)
- [ ] Let the generator handle globals
- [ ] Let the generator handle traits
- [ ] Let the generator handle trait impls
- [ ] Let the generator handle type alises

## Additional Context

The current doc generator currently has just one goal in mind: be able to generate stdlib docs from the `.nr` files, without us having to manually write `.md` files with doc comments. This means that the doc generator can't be used for other crates (or maybe it can, but the output is not ideal). The reason is that to generate docs we traverse the AST so we don't have type information. Ideally a doc generator would put links between types, etc., but doing that is much more complex and more time consuming. It can be done later, but the current goal is just to simplify the docs writing process.

## Documentation

Check one:

# PR Checklist


asterite:
Closing for now. We can reopen or revive once we have more stdlib docs in the `.nr` files.

PR #5992, opened 2024-09-10
feat: add `TypedExpr::get_type`

asterite:
# Description

## Problem

Part of #5668

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/docs/docs/noir/standard_library/meta/typed_expr.md
+++ b/docs/docs/noir/standard_library/meta/typed_expr.md
@@ -16,4 +16,4 @@ If this expression refers to a function definitions, returns it. Otherwise retur
 
 #include_code get_type noir_stdlib/src/meta/typed_expr.nr rust
 
-Returns the type of the expression, if the expression could be resolved without errors.
+Returns the type of the expression, or `Option::none()` if there were errors when the expression was previously resolved.
--- a/docs/docs/noir/standard_library/meta/typed_expr.md
+++ b/docs/docs/noir/standard_library/meta/typed_expr.md
@@ -6,8 +6,14 @@ title: TypedExpr
 
 ## Methods
 
-### as_function_definition
+### get_type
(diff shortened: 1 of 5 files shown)


docs/docs/noir/standard_library/meta/typed_expr.md
```diff
@@ -6,8 +6,14 @@ title: TypedExpr
...
 
-If this expression refers to a function definitions, returns it. Otherwise returns `Option::none()`.
\ No newline at end of file
+If this expression refers to a function definitions, returns it. Otherwise returns `Option::none()`.
+
+### get_type
+
+#include_code get_type noir_stdlib/src/meta/typed_expr.nr rust
+
+Returns the type of the expression, if the expression could be resolved without errors.
```
jfecher, line 19:
```suggestion
Returns the type of the expression, or `Option::none()` if there were errors when the expression was previously resolved.
```
Trying to be more explicit that the expression is already resolved. If not, users would have to worry about the scope `get_type` is called in.

test_programs/compile_success_empty/comptime_typed_expr/src/main.nr
```diff
@@ -0,0 +1,7 @@
+fn main() {
+    comptime
+    {
+        let expr = quote { [1, 3] }.as_expr().unwrap().resolve(Option::none());
```
jfecher, line 4: In another PR we should add

```rs
fn as_typed_expr(q: Quoted, scope: Option<FunctionDefinition>) -> Option<TypedExpr> {
    q.as_expr().map(|e: Expr| e.resolve(scope))
} 
```
One issue is that passing None would resolve within `as_typed_expr` instead of `main` though. You wouldn't be able to reference local variables even if you passed main explicitly I think.

jfecher (approved):

PR #5994, opened 2024-09-10
feat: add more `Type` and `UnresolvedType` methods

asterite:
# Description

## Problem

Part of #5668

## Summary

Work in progress.... (I'll add a bunch of methods in one go)

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -185,14 +188,23 @@ impl<'local, 'context> Interpreter<'local, 'context> {
             "type_implements" => type_implements(interner, arguments, location),
             "type_is_bool" => type_is_bool(arguments, location),
             "type_is_field" => type_is_field(arguments, location),
+            "type_is_unit" => type_is_unit(arguments, location),
```
michaeljklein, line 191: These `is_T` functions can be stdlib functions without builtins by using the existing `Eq` for `Type` implementation

michaeljklein (changes requested):
LGTM besides the note r.e. `is_T` functions

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
asterite: Ah, you mean comparing against `quote { Field}.as_type()} `? I think that could work but it would be slower, not sure if that's a concern.

asterite: The main issue right now is that there's no way to produce a `Field` unresolved type... or, well, we could get it from a `let` statement, or we'd have to add `quote { ... }.as_unresolved_type()` (which makes sense). I guess I'll do the latter, then use that.

asterite: Actually, I'm not sure. I think the code to implement these `is_`? methods is relatively simple, and it would be more complex to implement them in Noir.

michaeljklein: Sure, it would be slightly more complex to implement a polymorphic version in Noir and then have each `is_T` call out to it, but this approach results in a constant number of builtins (in terms of the number of variants in our `Type` enum).

asterite: @michaeljklein I need `Type::is_unit` for the dispatch macro. Would it be okay to merge this PR as-is? I'm still not sure that not using built-ins for this will be better.

asterite: Actually I was able to workaround this. I defined a function like this:

```noir
comptime fn get_type<T>() -> Type {
    let t: T = std::mem::zeroed();
    std::meta::type_of(t)
}
```

Then I can do:

```noir
let unit = get_type::<()>();
```

and compare that to another type to check if it's unit or not. I guess the same could be done with other primitive types.

TomAFrench: I think that considering we already have a number of `is_T` functions we can add this one and reassess whether we want to remove them in one go.

michaeljklein (approved):

noir_stdlib/src/meta/typ.nr
```diff
@@ -1,69 +1,195 @@
...
 // docs:start:fresh_type_variable
 pub comptime fn fresh_type_variable() -> Type {}
 // docs:end:fresh_type_variable
 
 impl Type {
+    /// If this type is an array, return a pair of (element type, size type).
+    /// 
+    /// Example:
+    /// 
+    /// ```rust
```
vezenovm, line 66:
```suggestion
    /// ```noir
```
nit

Issue #6006, opened 2024-09-11
Expected identical type with arithmetic generics


jfecher:
### Aim

```rs
/// A trait for types that can be decoded from a calldata field array
trait FromCallData<let N: u32, let M: u32> {
    /// Decodes Self by consuming it from the head of calldata, returning the remaining calldata
    fn from_calldata(calldata: [Field; N]) -> (Self, [Field; M]);
}

struct Point { x: Field, y: Field }

impl <let N: u32> FromCallData<N, N - 1> for Field {
    fn from_calldata(calldata: [Field; N]) -> (Self, [Field; (N - 1)]) {
        let slice = calldata.as_slice();
        let (value, slice) = slice.pop_front();
        (value, slice.as_array())
    }
}

impl <let N: u32> FromCallData<N, N - 2> for Point {
    fn from_calldata(calldata: [Field; N]) -> (Self, [Field; (N - 2)]) {
        let (x, calldata) = FromCallData::from_calldata(calldata);
        let (y, calldata) = (1, foo(calldata));//FromCallData::from_calldata(calldata);
        (Self { x, y }, calldata)
    }
}

fn foo<let N: u32>(_a: [Field; N]) -> [Field; N - 1] {
    std::mem::zeroed()
}

fn main() {
    let calldata = [1, 2];
    let _: (Point, _) = FromCallData::from_calldata(calldata);
}
```

### Expected Behavior

The program to type check correctly

### Bug

```
error: expected type (Point, [Field; (N - 2)]), found type (Point, [Field; (N - 2)])
   ┌─ src/main.nr:18:47
   │
18 │     fn from_calldata(calldata: [Field; N]) -> (Self, [Field; (N - 2)]) {
   │                                               ------------------------ expected (Point, [Field; (N - 2)]) because of return type
   ·
21 │         (Self { x, y }, calldata)
   │         ------------------------- (Point, [Field; (N - 2)]) returned here
   │
```
(The error changes if the `foo` call is replaced with another `from_call_data` call)

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #6008, opened 2024-09-11
fix: Try to move constant terms to one side for arithmetic generics

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6006

## Summary\*

Previously we were failing for constraints like `a + 3 = b + 1` when we could instead move the constant terms to one side: `a + 2 = b` then solve with `b := a + 2`.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
+++ b/compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
@@ -1,6 +1,6 @@
 use std::collections::BTreeSet;
 
-use crate::{BinaryTypeOperator, Type};
+use crate::{BinaryTypeOperator, Type, TypeBindings, UnificationError};
 
 impl Type {
     /// Try to canonicalize the representation of this type.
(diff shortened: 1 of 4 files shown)


compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
```diff
@@ -212,4 +212,44 @@ impl Type {
...
+    /// Try to unify equations like `(..) + 3 = (..) + 1`
+    /// by transforming them to `(..) + 2 =  (..)`
+    pub(super) fn try_unify_by_moving_constant_terms(
+        &self,
+        other: &Type,
+        bindings: &mut TypeBindings,
+    ) -> Result<(), UnificationError> {
+        if let Type::InfixExpr(lhs_a, op_a, rhs_a) = self {
+            if let Some(inverse) = op_a.inverse() {
+                if let Some(rhs_a) = rhs_a.evaluate_to_u32() {
```
michaeljklein, line 225: Could it be useful to attempt this in both directions? I.e. attempt to evaluate `lhs` and continue with `try_unify`?

PR #6018, opened 2024-09-12
feat: add `fmtstr::quoted`

asterite:
# Description

## Problem

Reviving the first commit in #5942

## Summary

This adds a way to turn a `fmtstr` into a `Quoted` value that holds a string token. Otherwise there's no way to create string tokens by interpolating dymamic (well, at compile-time) values.

This PR also fixes an issue with how string tokens are displayed. Previously they wouldn't include the double quotes, and special characters weren't escaped.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/docs/docs/noir/standard_library/fmtstr.md
+++ b/docs/docs/noir/standard_library/fmtstr.md
@@ -6,6 +6,12 @@ title: fmtstr
 
 ## Methods
 
+### quoted
+
+#include_code quoted noir_stdlib/src/meta/format_string.nr rust
+
+Returns the string literal as a `Quoted` value (this includes the leading and trailing double quoted).
+
 ### quoted_contents
 
 #include_code quoted_contents noir_stdlib/src/meta/format_string.nr rust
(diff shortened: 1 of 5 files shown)


docs/docs/noir/standard_library/fmtstr.md
```diff
@@ -6,6 +6,12 @@ title: fmtstr
 
 ## Methods
 
+### quoted
+
+#include_code quoted noir_stdlib/src/meta/format_string.nr rust
+
+Returns the string literal as a `Quoted` value (this includes the leading and trailing double quoted).
```
asterite, line 13:
```suggestion
Returns the string literal as a `Quoted` value (this includes the leading and trailing double quotes).
```

asterite:
> We could possibly instead add a helper method

Would that be a method of `fmtstr`?

asterite:
I'll split the "[Correctly print string tokens](https://github.com/noir-lang/noir/pull/6018/commits/c8f293a03e9d070530087544c03f30bb2d949748)" commit into a separate PR in the meantime...

PR #6026, opened 2024-09-12
feat: Add a `comptime` string type for string handling at compile-time

jfecher:
# Description

## Problem\*

Resolves 

## Summary\*

Adds CtString, a dynamically sized string type available at compile-time only.

## Additional Context

- Named `CtString` because I didn't want to use up the more general `String` in case we wanted one later. `CtString` also avoids users accidentally using this in runtime code before they know what it is.
- Had to make a small change to allow `comptime` on trait functions so that I could have one return `CtString` which is only allowed in comptime code.

## Documentation\*

Check one:

# PR Checklist\*


noir_stdlib/src/meta/ctstring.nr
```diff
@@ -0,0 +1,98 @@
...
+    comptime fn append_fmtstr<let N: u32, T>(self, s: fmtstr<N, T>) -> Self {
+        // docs:end:append_fmtstr
+        f"{self}{s}".as_ctstring()
+    }
+
+    /// CtString cannot directly return a str since the size would not be known.
+    /// To get around this, we return a quoted str and the underlying str can
+    /// be accessed using macro insertion `foo.as_quoted_str!()`.
+    // docs:start:as_quoted_str
+    comptime fn as_quoted_str(self) -> Quoted {
```
michaeljklein, line 27: To use this, would a user also need the length of the `CtString`? Please provide an example and/or test that uses this method

asterite (approved):
Looks good!

noir_stdlib/src/meta/ctstring.nr
jfecher: There's already an example, `as_quoted_str_example` and it is in the docs as well.

PR #6027, opened 2024-09-13
fix: disable side-effects for no_predicates functions

guipublic:
# Description

## Problem\*

Brillig output is nullified if the call is with a predicate, however, functions with no_predicates attribute expect valid values from their brillig calls because when they constrain the values , it will not have a predicate.

Resolves e-2-e test failing on PR https://github.com/AztecProtocol/aztec-packages/pull/4571

## Summary\*
Disable side-effects when calling a function with no_predicates

## Additional Context
The dedicated inlining pass for no_predicates functions can be removed after this fix.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -307,8 +313,23 @@ impl<'f> Context<'f> {
...
+            if let Value::Function(fid) = self.inserter.function.dfg[func] {
+                result = *no_predicates.get(&fid).unwrap_or(&false);
+            }
+        }
+        result
+    }
+
     // Inline all instructions from the given block into the entry block, and track slice capacities
-    fn inline_block(&mut self, block: BasicBlockId) {
+    fn inline_block(&mut self, block: BasicBlockId, no_predicates: &HashMap<FunctionId, bool>) {
```
vezenovm, line 332: Nit: I think it would be a bit cleaner to have this map be part of the `Context`, but this is non-blocking as we don't have to thread the map through many functions.

vezenovm (approved):

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
guipublic: That's what I did initially but they were borrow issues so passing the map is what seems the simplest to me.

Issue #6036, opened 2024-09-13
Quoted `Exprs` from blocks fail to parse back into expressions


Thunkar:
### Aim

Quote an expression from a block expression, add it to a quote, unquote the whole thing as a new expression.

This is because the expression is an `InternedStatement` internally

### Expected Behavior

The above should work

### Bug

Compiler fails to parse the Expr:

Parsed expr Err(FailedToParseMacro { 
    error: ParserError { expected_tokens: {}, expected_labels: {Expression}, found: InternedStatement(InternedStatementKind(Index(0)))

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

Blocker

### Would you like to submit a PR for this Issue?

None

PR #6040, opened 2024-09-13
fix: Parse a statement as an expression

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6036

## Summary\*

Grego was having an issue from some macros where some code was taken from a block and later put in the rhs of a `let` statement: `let _ = $former_block_statement`. Since this expression originated from a block statement though, it is a `Token::InternedStatement` and not a valid expression.

I've updated the parser to accept interned statements in an expression position, and we just later validate that they're either `StatementKind::Expression` or `StatementKind::Semi` nodes.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/test_programs/compile_success_empty/comptime_parse_statement_as_expression/'
+++ b/test_programs/compile_success_empty/comptime_parse_statement_as_expression/'
@@ -1,0 +1,4 @@
+struct ArrData<let N: u32> {
+    a: [Field; N],
+    b: [Field; N + N - 1],
+}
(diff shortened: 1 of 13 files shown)


test_programs/compile_success_empty/comptime_parse_statement_as_expression/'
```diff
@@ -0,0 +1,4 @@
+struct ArrData<let N: u32> {
+    a: [Field; N],
+    b: [Field; N + N - 1],
+}
```
asterite, line 4: Extra file

asterite (approved):
Looks good!

PR #6050, opened 2024-09-16
feat(metaprogramming): Add `#[use_callers_scope]`

jfecher:
# Description

## Problem\*

We have many methods now such as `Quoted::as_type` or `Quoted::as_trait_constraint` which resolve items in the current scope. It is useful to change the scope these are resolved in but somewhat impractical to do what we did with `Expression::resolve` and have them all take module parameters.

## Summary\*

I've added `#[use_callers_scope]` which can be added to a comptime function to change the scope any builtins resolve in to that of the caller. This also works in helper functions and closures.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/comptime.rs
+++ b/compiler/noirc_frontend/src/elaborator/comptime.rs
@@ -319,7 +319,7 @@ impl<'context> Elaborator<'context> {
         // If the function is varargs, push the type of the last slice element N times
         // to account for N extra arguments.
         let modifiers = interpreter.elaborator.interner.function_modifiers(&function);
-        let is_varargs = modifiers.attributes.is_varargs();
+        let is_varargs = modifiers.attributes.has_varargs();
         let varargs_type = if is_varargs { parameters.pop() } else { None };
 
         let varargs_elem_type = varargs_type.as_ref().and_then(|t| t.slice_element_type());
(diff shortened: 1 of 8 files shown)


asterite:
> but somewhat impractical to do what we did with Expression::resolve and have them all take module parameters

Why, is it because of verbosity?

I wonder how this will interact with other functions. For example we have `FunctionDefinition::set_parameters`. In that method we evaluate the types of parameters relative to the function, and adding `#[use_callers_scope]` won't change that. Or it won't change how `Expr::resolve` works. So I don't know if `#[use_callers_scope]` should apply to these (and actually all) these solving methods, or we'll just have some methods work in one way, some methods work in another way.

jfecher:
> Why, is it because of verbosity?

Yes, I think it's a bit impractical to always have users pass around module ids everywhere when most of the time they just want to work as if the attribute is in the scope of the item it's being used on.

> So I don't know if `#[use_callers_scope]` should apply to these (and actually all) these solving methods, or we'll just have some methods work in one way, some methods work in another way.

`#[use_callers_scope]` definitely shouldn't apply everywhere. I just have it changing the "default" currently but it is still overridden by anything that explicitly sets the function or module. So `resolve` with `None` for the  module argument would use the caller as the "current" scope instead. Some things like `Module::add_item` should also always be within that module since it is the destination of the item and doesn't make sense to be anywhere else.

I suppose we could also discuss whether the module argument should be removed from `Expression::resolve` but I saw adding this attribute as a quick way of changing the scope for all of these resolution methods that are currently fixed, and doing it in a way that is consistent for users across each function (although does have a learning curve as they'd need to be aware of the attribute unlike a parameter which is obvious).

asterite (approved):

PR #6051, opened 2024-09-16
fix: let LSP suggest fields and methods in LValue chains

asterite:
# Description

## Problem

LSP autocompletion didn't trigger in some cases.

## Summary

Makes LSP autocompletion work when typing a dot but in the next line an l-value assignment follows. In that case the current line and the line that follows are parsed as a single assignment to an l-value and we must complete in the middle of that entire expression.

## Additional Context

Also adds a missing `follow_bindings()` that caused some of these cases to not work.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/lsp/src/requests/completion.rs
+++ b/tooling/lsp/src/requests/completion.rs
@@ -17,7 +17,7 @@ use noirc_frontend::{
     ast::{
         AsTraitPath, AttributeTarget, BlockExpression, CallExpression, ConstructorExpression,
         Expression, ExpressionKind, ForLoopStatement, GenericTypeArgs, Ident, IfExpression,
-        ItemVisibility, Lambda, LetStatement, MemberAccessExpression, MethodCallExpression,
+        ItemVisibility, LValue, Lambda, LetStatement, MemberAccessExpression, MethodCallExpression,
         NoirFunction, NoirStruct, NoirTraitImpl, Path, PathKind, Pattern, Statement,
         TraitImplItemKind, TypeImpl, UnresolvedGeneric, UnresolvedGenerics, UnresolvedType,
         UnresolvedTypeData, UseTree, UseTreeKind, Visitor,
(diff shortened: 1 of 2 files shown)


tooling/lsp/src/requests/completion.rs
```diff
@@ -1443,6 +1538,34 @@ impl<'a> Visitor for NodeFinder<'a> {
...
+    }
+}
+
+fn get_array_element_type(typ: Type) -> Option<Type> {
+    match typ {
+        Type::Array(_, typ) | Type::Slice(typ) => Some(*typ),
+        Type::Alias(alias_type, generics) => {
+            let typ = alias_type.borrow().get_type(&generics);
+            get_array_element_type(typ)
+        }
```
jfecher, line 1564: May want to mathc/follow type variables here too

jfecher (approved):

tooling/lsp/src/requests/completion.rs
```diff
@@ -1559,7 +1559,7 @@ fn get_array_element_type(typ: Type) -> Option<Type> {
     match typ {
         Type::Array(_, typ) | Type::Slice(typ) => Some(*typ),
         Type::Alias(alias_type, generics) => {
-            let typ = alias_type.borrow().get_type(&generics);
+            let typ = alias_type.borrow().get_type(&generics).follow_bindings();
```
jfecher, line 1562: Not here, I meant in the match itself we should recur in the type variable case

asterite: Ah, I see. Let me know if I got it right this time.

tooling/lsp/src/requests/completion.rs
```diff
@@ -1556,10 +1556,10 @@ fn get_field_type(typ: &Type, name: &str) -> Option<Type> {
 }
 
 fn get_array_element_type(typ: Type) -> Option<Type> {
-    match typ {
+    match typ.follow_bindings() {
```
jfecher, line 1559: Instead of matching on `typ.follow_bindings()`, since we're already matching on the type we should just match on the `Type::TypeVariable` and `Type::NamedGeneric` variants explicitly. It is some more code but should be a bit faster at least.

asterite: The third's a charm 🤞

Issue #4775, opened 2024-04-10
Support "pub" for structs and traits


michaeljklein:
### Problem

Compiling the following snippet fails with a parsing error:

```rust
pub struct Foo {}
```

```rust
error: Unexpected struct, expected one of fn, open
  ┌─ /Users/michaelklein/Coding/noir/name_shadowing_template/library2/src/lib.nr:7:5
  │
7 │ pub struct Foo {}
  │     ------
  │
```

Likewise, compiling a similar snippet for traits fails with the same error:

```rust
pub trait Foo { ..
```

```bash
error: Unexpected trait, expected one of fn, open
   ┌─ /Users/michaelklein/Coding/noir/name_shadowing_template/library2/src/lib.nr:15:5
   │
15 │ pub trait Foo {
   │     -----
   │
```

### Happy Case

Either support `pub` for `struct`'s or fail with a more informative error message, e.g. `the 'pub' keyword is only supported for functions at this time`.

### Project Impact

Nice-to-have

### Workaround

None

### Would you like to submit a PR for this Issue?

None

PR #6054, opened 2024-09-17
feat: allow visibility modifiers in struct definitions

asterite:
# Description

## Problem

Part of #4515
Part of #4775

## Summary

Allows marking structs as pub or pub(crate). Also will now report unused structs.

## Additional Context

Let me know if some of the structs I marked as `pub` shouldn't actually be `pub`. I didn't mark some as pub, I think only the ones used for tests.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/ast/structure.rs
+++ b/compiler/noirc_frontend/src/ast/structure.rs
@@ -6,13 +6,14 @@ use crate::token::SecondaryAttribute;
 use iter_extended::vecmap;
 use noirc_errors::Span;
 
-use super::Documented;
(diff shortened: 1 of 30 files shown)


TomAFrench (approved):
LGTM, not looking forward to this sync :laughing:

asterite:
Actually, the unused warning isn't super correct right now. If you have `struct Foo` then `impl Foo` it will consider `Foo` as being used. That said, the warning will never give a false positive, and this PR's intention is to enable visiblity. We can improve the unused warning later on.

PR #6057, opened 2024-09-17
feat: check unconstrained trait impl method matches

### LINKED ISSUE #5717: Allow traits to declare unconstrained functions
### Problem

Currently all trait functions must be constrained: adding the `unconstrained` keyword results in a parser error. It'd be good to also be able to make these functions unconstrained, as there are many scenarios in which we require this property from an interface.

An example is Aztec's note abstraction, which includes the ability to compute a note nullifier. These nullifiers include a secret that is injected by an oracle and then constrained by an external component. More concretely, we take a mutable reference to an array of validation requests that are constrained to later be fulfilled - this deferring exists because said constraining must be done by a different circuit.

However, we also need to have an unconstrained way of computing these notes, in which we essentially skip all of the validation. Due to these complicated interactions with the deferred validation requests, we've not been able to come up with a better abstraction than to simply have a second function that does not perform any constraining - the obvious problem being that we cannot mark this function as `unconstrained`, even though it is. 

This may not be a huge deal today, but with #4442 there'll be an expectation of being unable to accidentally call unconstrained functions from constrained contexts, and lacking the ability to properly tag trait functions as such prevents us from leveraging this feature.

### Happy Case

```rust
trait Foo {
  fn bar() -> Field;
  unconstrained fn bar_unconstrained() -> Field;
}
```

And then force implementations of the trait to make `bar_unconstrained` be indeed `unconstrained`.
  

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Fixes #5717

## Summary

In a previous PR we started allowing `unconstrained` and `comptime` in trait impls. But two things remained:
1. Defining an `unconstrained` trait impl method would produce a warning.
2. We should error if a trait method is unconstrained but the trait impl method is not, or the other way around.

This PR will also include "unconstrained" in trait imp method stubs in LSP.

## Additional Context

I initially also errored if there's a mismatch in `comptime`, but that's probably not good as we have some implementation of `Append` be comptime and some not. This made me wonder what's the purpose of `comptime` in functions, if non-comptime functions can also be interpreted.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/type_check/errors.rs
+++ b/compiler/noirc_frontend/src/hir/type_check/errors.rs
@@ -268,7 +268,7 @@ impl<'a> From<&'a TypeCheckError> for Diagnostic {
             }
             TypeCheckError::UnconstrainedMismatch { item, expected, span } => {
                 let msg = if *expected {
-                    format!("{item} expected to be unconstrained")
+                    format!("{item} is expected to be unconstrained")
                 } else {
                     format!("{item} is not expected to be unconstrained")
                 };
--- a/compiler/noirc_frontend/src/hir/type_check/errors.rs
+++ b/compiler/noirc_frontend/src/hir/type_check/errors.rs
@@ -61,6 +61,8 @@ pub enum TypeCheckError {
     ParameterCountMismatch { expected: usize, found: usize, span: Span },
     #[error("{item} expects {expected} generics but {found} were given")]
     GenericCountMismatch { item: String, expected: usize, found: usize, span: Span },
+    #[error("{item} has incompatible `unconstrained`")]
+    UnconstrainedMismatch { item: String, expected: bool, span: Span },
     #[error("Only integer and Field types may be casted to")]
     UnsupportedCast { span: Span },
     #[error("Index {index} is out of bounds for this tuple {lhs_type} of length {length}")]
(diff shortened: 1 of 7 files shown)


michaeljklein (approved):
LGTM

jfecher:
> This made me wonder what's the purpose of comptime in functions, if non-comptime functions can also be interpreted.

comptime functions can _only_ be run at compile-time

compiler/noirc_frontend/src/hir/type_check/errors.rs
```diff
@@ -264,6 +266,14 @@ impl<'a> From<&'a TypeCheckError> for Diagnostic {
                 let msg = format!("{item} expects {expected} generic{empty_or_s} but {found} {was_or_were} given");
                 Diagnostic::simple_error(msg, String::new(), *span)
             }
+            TypeCheckError::UnconstrainedMismatch { item, expected, span } => {
+                let msg = if *expected {
+                    format!("{item} expected to be unconstrained")
```
jfecher, line 271:
```suggestion
                    format!("{item} is expected to be unconstrained")
```

jfecher (approved):

PR #6058, opened 2024-09-17
feat: visibility for type aliases

asterite:
# Description

## Problem

Part of #4515

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/ast/type_alias.rs
+++ b/compiler/noirc_frontend/src/ast/type_alias.rs
@@ -1,4 +1,4 @@
-use super::{Ident, UnresolvedGenerics, UnresolvedType};
+use super::{Ident, ItemVisibility, UnresolvedGenerics, UnresolvedType};
 use iter_extended::vecmap;
 use noirc_errors::Span;
 use std::fmt::Display;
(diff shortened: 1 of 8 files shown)


jfecher (approved):

TomAFrench (changes requested):
checking something.

TomAFrench:
```
mod foo {
    struct Foo {
        inner: Field
    }
    pub type Bar = Foo;
}

fn main() {
    // The `foo::Foo` raises a warning because we can't construct from a type alias currently
    // if we could then this would be entirely legal.
    let _: foo::Bar = foo::Foo { inner: 0 };
}
```

This PR currently allows us to expand the visibility of types by defining a type alias.

jfecher:
> This PR currently allows us to expand the visibility of types by defining a type alias.

I kinda like this feature :sweat_smile:

asterite:
It's interesting because the above code in Rust gives an error:

```rust
mod foo {
    struct Foo {
        pub inner: i32,
    }
    pub type Bar = Foo;
}
```

The error is:

> Type `Foo` is more private than item `Bar`

However, if I make `Foo` be `pub(crate)`:

```rust
mod foo {
    pub(crate) struct Foo {
        pub inner: i32,
    }
    pub type Bar = Foo;
}
```

that error no longer happens 🤔 

I don't understand why Rust doesn't give an error in this case. `Foo` is still more private than `Bar`.

For now I'll make both cases give an error in Noir, but let me know if it should be like in Rust instead.

TomAFrench (approved):

Issue #6059, opened 2024-09-17
`use $typ` statements can be unquoted in ilegal positions


Thunkar:
### Expected Behavior

```noir 
struct MyType {
    a: u64
}


comptime fn annotation(m: Module) -> Quoted {
    let typ = quote { MyType }.as_type();
    quote {
        use $typ;
    }
}
```

should return an error

### Bug

> It looks like this is the old "we don't error if we fail to parse everything" issue. So it just parses up to that use . For my example since the use is the first/only thing it just returns an empty list of statements

@jfecher 

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

Issue #5803, opened 2024-08-22
`Quoted::as_...` doesn't consume all tokens


asterite:
### Aim

When turning Quoted into something else, it seems once tokens were consumed to form the target type, the rest of the tokens are discarded. Maybe in that case it should error with "expecting EOF" or something like that.

For example:

```rust
fn main() {
    comptime
    {
        let expr = quote { x = hello world ([{}]) }.as_expr().unwrap();
        println(expr);
    }
}
```

Output:

```
x
``` 

### Expected Behavior

The above program errors because the entire token stream doesn't parse to an expression.

### Bug

Remaining tokens are silently discarded.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

asterite:
I'm also thinking, given that `as_expr()` can actually wrap a statement, an LValue, eventually a Pattern... maybe `as_expr()` would try to parse each of those things in turn, right? So you could do `quote { a = 1; }.as_expr()` instead of having to wrap that in a block expression.

jfecher:
Yeah. I do wonder if that could be problematic though. If there's ever when you'd parse a variable but you wanted an lvalue variable or a pattern variable or something. Not sure but if that's ever the case we'd have to split Expr into the actual 2-4 types it wraps.

PR #6064, opened 2024-09-17
fix: Always parse all tokens from quoted token streams

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/5803
Resolves https://github.com/noir-lang/noir/issues/6059

## Summary\*

If we ever had a token stream with a subset of valid input we'd parse just that subset instead of erroring that the full input doesn't parse.

## Additional Context

We don't have EOF tokens in token streams so I used the `end()` primitive instead.

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
Looks great!

PR #6076, opened 2024-09-18
fix: don't crash on untyped global used as array length

### LINKED ISSUE #6046: Using an untyped global as array length crashes the compiler
### Aim

I want to be able to derive globals from other globals, and if I accidentially include a typo, I want a error message indicating the line with the faulty global definition.

### Expected Behavior

When I derive a global from a non-existing global, I want the compiler to mark that line as an error (IntelliSense should underline it red as well)

### Bug

The Noir LSP simply panics:

Message:  index out of bounds: the len is 4918 but the index is 18446744073709551615
Location: compiler/noirc_frontend/src/node_interner.rs:1145

### To Reproduce

Did not reproduce it from scratch, but can be easily reproduced in any working project:

global SERIALIZED_OBSTACLES_ARRAY_SIZE = 3;
global SERIALIZED_ENEMY_OBSTACLES_ARRAY_SIZE = SERIALIED_OBSTACLES_ARRAY_SIZE + 1;

After a few seconds Intellisense should stop working and a popup should complain about the LSP having crashed 5 times and will not be restarted anymore. In case you are not using Intellisense:

nargo compile

This will cause Noir to panic as well and show the same error message

### Workaround

None

### Workaround Description

_No response_

### Additional Context

Very annoying, as it is really hard to guess what caused the panic. The missing Z in the above example was really hard to spot. I only noticed it after checking the node_interner.rs file in the Noir github. Only thanks to the compiler source code I was able to find out that I should be looking for an invalid global definition. An error, at least indicating that a global definition caused this would help a lot!

### Project Impact

Nice-to-have

### Blocker Context

_No response_

### Nargo Version

noirc version = 0.34.0+359caafac5e489901d9ff02b08d1a688178d9b0a

### NoirJS Version

0.32.0

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #6046

## Summary

Another case of an unhandled`DefinitionId::dummy()`.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/types.rs
+++ b/compiler/noirc_frontend/src/elaborator/types.rs
@@ -655,18 +655,21 @@ impl<'context> Elaborator<'context> {
                 int.try_into_u128().ok_or(Some(ResolverError::IntegerTooLarge { span }))
             }
             HirExpression::Ident(ident, _) => {
-                let definition = self.interner.definition(ident.id);
-                match definition.kind {
-                    DefinitionKind::Global(global_id) => {
-                        let let_statement = self.interner.get_global_let_statement(global_id);
-                        if let Some(let_statement) = let_statement {
-                            let expression = let_statement.expression;
-                            self.try_eval_array_length_id_with_fuel(expression, span, fuel - 1)
-                        } else {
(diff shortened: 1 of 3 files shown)


jfecher (approved):
LGTM

Issue #6077, opened 2024-09-18
Unquoting a numeric type fails to parse


Thunkar:
### Aim

Use array types in quoted values

```noir
global A_GLOBAL = 3;

struct WeirdStruct<T, U> {
    a: T,
    b: U,
}

#[mangle_fn]
fn my_fn() -> [u8; A_GLOBAL] {
    [0; A_GLOBAL]
}

comptime fn mangle_fn(f: FunctionDefinition) {
    let return_type = f.return_type();

    // I have to construct this as a quoted value for my use case, cannot use the type directly
    let generics = f"Field,{return_type}".quoted_contents();
    // This prints the faulty type <Field,[u8; (3: numeric Field)]>
    println(generics);
    let new_return_type = quote { WeirdStruct<$generics>}.as_type();

    let new_body = quote {
        {
            WeirdStruct { a: 1, b: [0;3] }
        }
    }.as_expr().unwrap();

    f.set_return_type(new_return_type);
    f.set_body(new_body);
}
```

### Expected Behavior

The above example should compile

### Bug

```
error: Failed to parse macro's token stream into a type
   ┌─ src/main.nr:19:27
   │
19 │     let new_return_type = quote { WeirdStruct<$generics>}.as_type();
   │                           -------------------------------
   │                           │
   │                           Unexpected UnquoteMarker(ExprId(Index(44777))), expected one of bool, crate, CtString, dep, Expr, Field, fn, fmtstr, FunctionDefinition, impl, Module, Quoted, str, StructDefinition, super, TopLevelItem, TraitConstraint, TraitDefinition, TraitImpl, TypedExpr, Type, unconstrained, UnresolvedType, <, &, (, [, integer type, type expression, Ident, QuotedType, InternedUnresolvedTypeData 
```

### To Reproduce
 Use the example

### Workaround

Yes

### Workaround Description

```noir
    let fn_return_type_quoted = if fn_return_type.as_array().is_some() {
        let (element_type, array_len) = fn_return_type.as_array().unwrap();
        let array_len = array_len.as_constant().unwrap();
        quote { [$element_type; $array_len] }
    } else {
        quote { $fn_return_type }
    };
```

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #6079, opened 2024-09-18
fix: Fix comptime type formatting

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6077

## Summary\*

The change made to `impl Debug for Type` to display kinds differently was breaking the use case of formatting types into a format string then using `.quoted_contents()` to quote the contents of the string.

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
Looks good!

I don't know if it's captured, but when there's a "failed to parse macro stream" I think we aren't "resolving" tokens so it's hard to understand what the actual stream was.

jfecher:
@asterite I think showing the tokens as "UnquoteMarker" or "Interned" are more useful to compiler developers though at least. Since expanding the tokens would be a lie since that is not what the parser actually sees.

PR #6080, opened 2024-09-18
feat!: always print Type::Constant kinds

michaeljklein:
# Description

## Problem\*

Instead of printing:
```bash
Expected type [RootParityInput; 4)], found type [RootParityInput; 4)]
```

We print:
```bash
Expected type [RootParityInput; (4: u32)], found type [RootParityInput; (4: Field)]
```

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir_def/types.rs
+++ b/compiler/noirc_frontend/src/hir_def/types.rs
@@ -163,7 +163,7 @@ impl std::fmt::Display for Kind {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Kind::Normal => write!(f, "normal"),
-            Kind::Numeric(typ) => write!(f, "numeric {}", typ),
+            Kind::Numeric(typ) => write!(f, "{}", typ),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


TomAFrench (approved):

jfecher (changes requested):
We should almost never print constant kinds, this is the same formatting which broke https://github.com/noir-lang/noir/issues/6077.

We need to either only print these when they differ or use a different error type entirely for this case. We should see what Rust does here.

jfecher:
For:
```rust
struct Foo<const N: usize> {
    x: [u8; N],
}

fn main() {}

fn bar<const M: u32>() -> Foo<M> {
    todo!()
}
```
Rust gives:
```
error: the constant `M` is not of type `usize`
 --> t.rs:8:27
  |
8 | fn bar<const M: u32>() -> Foo<M> {
  |                           ^^^^^^ expected `usize`, found `u32`
  |
note: required by a bound in `Foo`
 --> t.rs:2:12
  |
2 | struct Foo<const N: usize> {
  |            ^^^^^^^^^^^^^^ required by this bound in `Foo`
```

michaeljklein:
The original issue appears to be resolved by https://github.com/noir-lang/noir/pull/6083

Issue #6082, opened 2024-09-18
Cannot specify a non-u32 numeric generic for a struct


jfecher:
### Aim

```
struct S<let N: u8> {}

fn main() {
    let _: S<3> = S {};
}
```

### Expected Behavior

The code to work

### Bug

```
error: Expected kind numeric u8, found kind numeric u32
  ┌─ src/main.nr:4:14
  │
9 │     let _: S<3> = S {};
  │              -
  │
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

Issue #6081, opened 2024-09-18
global type is not inferred from usage when used in a type


jfecher:
### Aim

```noir
global FOO = 3;

fn main() {
    let x: [u32; FOO] = [1, 2, 3];
    println(x);
}
```

### Expected Behavior

`FOO` should be inferred to be a `u32` from being used in an array

### Bug

```
error: Expected type [u32; 3], found type [Field; 3]
  ┌─ src/main.nr:4:25
  │
4 │     let x: [u32; FOO] = [1, 2, 3];
  │                         ---------
  │
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #6083, opened 2024-09-18
fix!: Infer globals to be u32 when used in a type

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6081
Resolves https://github.com/noir-lang/noir/issues/6082

## Summary\*

Now that we check kinds explicitly we were erroring if globals weren't already `u32` yet leading to a lot of existing code needing to be changed.

If we change the exact equality to a unification check we can instead infer the globals need to be a `u32` without requiring user code to be changed.

This PR is still breaking since there were causes we didn't catch before of invalid kinds being used in array lengths, e.g. `fn foo<let N: u8>(a: [Field; N])` in one of our unit tests.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir_def/types.rs
+++ b/compiler/noirc_frontend/src/hir_def/types.rs
@@ -151,11 +151,35 @@ impl Kind {
     }
 
     pub(crate) fn matches_opt(&self, other: Option<Self>) -> bool {
-        other.as_ref().map_or(true, |other_kind| self == other_kind)
+        other.as_ref().map_or(true, |other_kind| self.unifies(other_kind))
     }
 
     pub(crate) fn u32() -> Self {
         Self::Numeric(Box::new(Type::Integer(Signedness::Unsigned, IntegerBitSize::ThirtyTwo)))
+    }
+
+    /// Unifies this kind with the other. Returns true on success
+    pub(crate) fn unifies(&self, other: &Kind) -> bool {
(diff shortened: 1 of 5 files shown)


vezenovm (approved):

jfecher:
Changed the function to take an expected kind so that we can also fix https://github.com/noir-lang/noir/issues/6082

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -1465,13 +1489,13 @@ impl Type {
                 }
             }
 
-            (NamedGeneric(binding_a, name_a, _), NamedGeneric(binding_b, name_b, _)) => {
```
michaeljklein, line 1468: Should this also be applied to the above
```rust
(NamedGeneric(binding, _, _), other) | (other, NamedGeneric(binding, _, _))
```
case?

michaeljklein (changes requested):

michaeljklein:
Also, in `evaluate_to_u32` we're currently implicitly downcasting values with kinds larger than `u32`

jfecher:
> Also, in evaluate_to_u32 we're currently implicitly downcasting values with kinds larger than u32

Yeah :/
We also don't check that type constants fit into their kind to begin with

michaeljklein (approved):

PR #6084, opened 2024-09-18
fix: ensure to_bytes returns the canonical decomposition

### LINKED ISSUE #5846: Update `Field.to_be_bytes` methods to enforce that the byte decomposition does not result in overflows
We flag up in https://github.com/noir-lang/noir/pull/5843 that providing large values for `byte_size` in `Field.to_be_bytes` results in the prover being able to provide multiple different byte representations for certain inputs as providing the byte decomposition for `[p, 2^(8*byte_size))` will satisfy the constraints for inputs `[0, 2^(8*byte_size)-p)`.

We currently recommend ensuring that `byte_size` is chosen such that this situation cannot occur, however this results in a small range of `Field` values not being able to be decomposed using these methods. We should add additional logic to make it safe to decompose the full range of `Field` values without needing to worry about malicious provers.

guipublic:
# Description

## Problem\*

Resolves #5846 

## Summary\*
Enforce canonical decomposition for to_bytes

## Additional Context
I also added from_bytes methods since they are often requested

## Documentation\*
I updated the comments, I assume the doc is auto generated but I am not sure. Let me know if I need to modify it manually.

Check one:

# PR Checklist\*


noir_stdlib/src/field/mod.nr
```diff
@@ -49,39 +50,69 @@ impl Field {
...
-    /// wrap around due to overflow when verifying the decomposition.
+    /// The result is ensured to be the canonical decomposition of the field element
     // docs:start:to_le_bytes
     pub fn to_le_bytes<let N: u32>(self: Self) -> [u8; N] {
-        self.to_le_radix(256)
+        // docs:end:to_le_bytes
+        // Compute the byte decomposition
+        let bytes = unsafe {
+            compute_byte_decomposition(self)
+        };
```
TomAFrench, line 68: Can you revert this change  so that we still use the directive?

noir_stdlib/src/field/mod.nr
```diff
@@ -49,39 +50,69 @@ impl Field {
...
+
+        if !is_unconstrained() {
+            // Verify that the byte decomposition is correct
+            let mut y = 0;
+            let mut r = 1;
+            for i in 0..N {
+                y += (bytes[i] as Field)*r;
+                r *= 256;
+            }
+            assert_eq(self, y);
```
TomAFrench, line 78: Can then remove this.

noir_stdlib/src/field/mod.nr
```diff
@@ -49,39 +50,69 @@ impl Field {
...
+            let mut ok = bytes.len() != p.len();
+            for i in 0..N {
+                if !ok {
+                    if (bytes[N - 1 - i] != p[N - 1 - i]) {
+                        assert(bytes[N - 1 - i] < p[N - 1 - i]);
+                        ok = true;
+                    }
+                }
+            }
+            assert(ok);
```
TomAFrench, line 92: This can all go inside of a check that `N` is greater than or equal to the field modulus length in bytes.

noir_stdlib/src/field/mod.nr
```diff
@@ -49,39 +50,69 @@ impl Field {
...
     // docs:start:to_be_bytes
     pub fn to_be_bytes<let N: u32>(self: Self) -> [u8; N] {
-        self.to_be_radix(256)
+        // docs:end:to_be_bytes
+        let le_bytes: [u8; N] = self.to_le_bytes();
+        let mut be_bytes = [0; N];
+        for i in 0..N {
+            be_bytes[i] = le_bytes[N - 1 - i];
+        }
+        be_bytes
```
TomAFrench, line 114: Before we push the swapping of endianness into Noir rather than relying on separate builtins then we need to know that this doesn't result in unnecessary bytecode being emitted in brillig. Can you revert this and then we can tackle in another PR?

TomAFrench (changes requested):

noir_stdlib/src/field/mod.nr
guipublic: This is already the case, the check is done by 
```
         assert(bytes.len() <= p.len());
         let mut ok = bytes.len() != p.len();
```
or do you mean to refactor this with an explicit check N< p.len() ?

TomAFrench: I mean that if `bytes.len() < p.len()` then it the bytes array doesn't have enough elements to overflow and we can just skip the for-loop entirely.

TomAFrench: We then want to ensure that we don't emit unnecessary bytecode in that situation.

guipublic: if `bytes.len() < p.len()`, then ok is true so the following loop is empty and will not generate any byte code.

noir_stdlib/src/field/mod.nr
guipublic: Done

noir_stdlib/src/field/mod.nr
guipublic: Done

noir_stdlib/src/field/mod.nr
guipublic: This does not generate additional byte code because we access the bytes with constant index, just after 'fully' retrieving them.

guipublic: n.b. I don't understand the brillig report. The most significant programs in the report have 0 brillig opcode when I compile them.

TomAFrench: The brillig report compiles a number of ACIR programs to brillig. If you add the `--force-brillig` flag then you should get the output at in the report.

guipublic: In that case (to_be_bytes() used inside a brillig function) the swapping is indeed responsible for the brillig overhead, I will revert it.

noir_stdlib/src/field/mod.nr
```diff
@@ -132,6 +165,30 @@ impl Field {
     }
 }
 
+/// Convert a big endian byte array to a field element by modding
```
TomAFrench, line 168:
```suggestion
/// Convert a little endian byte array to a field element.
/// If the provided byte array overflows the field modulus then the Field will silently wrap around.
```

noir_stdlib/src/field/mod.nr
```diff
@@ -132,6 +165,30 @@ impl Field {
...
+
+    for i in 0..N {
+        result += (bytes[i] as Field) * v;
+        v = v * 256;
+    }
+    result
+}
+
+/// Convert a little endian byte array to a field element by modding
+pub fn from_be_bytes<let N: u32>(bytes: [u8; N]) -> Field {
```
TomAFrench, line 181: These should be class methods on Field rather than independent functions.

noir_stdlib/src/field/mod.nr
guipublic: Done

noir_stdlib/src/field/mod.nr
guipublic: Done

TomAFrench (approved):
It's unfortunate how the compiler doesn't infer the value of `N` when constructing the field from bytes.

PR #6088, opened 2024-09-18
feat(perf): Track last loads per block in mem2reg and remove them if possible

vezenovm:
# Description

## Problem\*

Resolves 

Part of general effort to improve mem2reg. 

## Summary\*

We sometimes have situations such as the following:
```
            b3():
              v9 = load v0
              v10 = eq v9, Field 2
              constrain v9 == Field 2
              v11 = load v2
              v12 = load v2
              v13 = eq v12, Field 2
              constrain v11 == Field 2
```
`v2` does not have a known value, thus we do not remove the load. The mem2reg pass is acting as expected here. However, without a store or call to the reference between `v11 = load v2` and `v12 = load v2` we should be able to safely remove `v12 = load v2` and map `v12 -> v11`. 

This PR adds this logic as part of the initial mem2reg pass. We have a new `last_loads` map as part of a `Block`. This is currently cleared after analyzing block and is meant to only be _per block_. Unifying these last loads across blocks and the accurate predecessors can come in a follow-up. This is an initial proof of concept to show the optimizations validity. 

Given an instruction we act as following:
#### Load 
 - Check if we have a last load from the current load address. If we do, remove the current current and map its result to the previous load result.
 - Add a last load for the address.
#### Store
- Remove the last load for the address
#### Call
- Remove the last load for any reference arguments

I have also added two unit tests to `mem2reg.rs`

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
> reference_only_used_as_alias | +7 ❌ | +2.81%

Hmm this is surprising.

vezenovm:
> > reference_only_used_as_alias | +7 ❌ | +2.81%
> 
> Hmm this is surprising. I'm guessing that removing some of these loads is perhaps reducing the amount of trivial stores we can remove, but I'm not sure.

The main difference between the SSA on master and this PR looks to be the `inc_rc` instructions remaining in place. Before `mem2reg` we have this pattern:
```
    v61 = load v52
    inc_rc v61
    inc_rc [u8 0, u8 0, u8 0, u8 0, u8 0]
    inc_rc [u8 0, u8 0, u8 0, u8 0, u8 0]
    v64 = load v51
    inc_rc v64
    v65 = load v52
    inc_rc v65
    v66 = load v52
    inc_rc v66
    v67 = load v52
    v68 = load v53
    v70 = lt v68, u32 4
    constrain v70 == u1 1 '"push out of bounds"'
    v72 = load v52
    v73 = load v53
    v74 = load v52
    v75 = load v53
    v76 = mul v75, u32 4
    v77 = array_set v72, index v76, value Field 0
```
The repeat loads are removed in this PR, but those follow-up inc_rc instructions remain. I think this can be handled in a follow-up though so I am marking this PR ready for review again.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -1706,4 +1748,230 @@ mod tests {
...
+        builder.insert_constrain(v9, two, None);
+        let v11 = builder.insert_load(v2, v2_type.clone());
+        let v12 = builder.insert_load(v2, Type::field());
+        let _ = builder.insert_binary(v12, BinaryOp::Eq, two);
+
+        builder.insert_constrain(v11, two, None);
+        builder.terminate_with_return(vec![]);
+
+        let ssa = builder.finish();
+        println!("{}", ssa);
```
michaeljklein, line 1831: Leftover debug `println`

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -1706,4 +1748,230 @@ mod tests {
...
+
+        builder.insert_constrain(v11, two, None);
+        builder.terminate_with_return(vec![]);
+
+        builder.new_function("foo".into(), foo_id, InlineType::default());
+        let _foo_v0 = builder.add_parameter(Type::Reference(Arc::new(Type::field())));
+        builder.terminate_with_return(vec![]);
+
+        let ssa = builder.finish();
+        println!("{}", ssa);
```
michaeljklein, line 1950: leftover debug `println`

compiler/noirc_evaluator/src/ssa/opt/mem2reg/block.rs
```diff
@@ -113,6 +116,7 @@ impl Block {
     fn invalidate_all_references(&mut self) {
         self.references.clear();
         self.last_stores.clear();
+        // self.last_loads.clear();
```
michaeljklein, line 119: Is this needed or leftover?

michaeljklein (changes requested):
A couple cleanup notes, but otherwise LGTM

compiler/noirc_evaluator/src/ssa/opt/mem2reg/block.rs
vezenovm, line 119:
```suggestion
```

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -1706,4 +1748,229 @@ mod tests {
...
+        builder.insert_constrain(v9, two, None);
+        let v11 = builder.insert_load(v2, v2_type.clone());
+        let v12 = builder.insert_load(v2, Type::field());
+        let _ = builder.insert_binary(v12, BinaryOp::Eq, two);
+
+        builder.insert_constrain(v11, two, None);
+        builder.terminate_with_return(vec![]);
+
+        let ssa = builder.finish();
+        println!("{}", ssa);
```
vezenovm, line 1831:
```suggestion
```

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -277,6 +287,10 @@ impl<'f> PerFunctionContext<'f> {
             self.remove_stores_that_do_not_alias_parameters(&references);
         }
 
+        // Last loads are truly "per block". During unification we are creating a new block from the current one,
+        // so we must clear the last loads of the current block before return the new block.
```
vezenovm, line 291:
```suggestion
        // so we must clear the last loads of the current block before we return the new block.
```

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -390,6 +404,28 @@ impl<'f> PerFunctionContext<'f> {
                             .insert((address, block_id), instruction_index);
                     }
                 }
+
+                // Check whether the block has a repeat load from the same address (w/ no calls or stores in between the loads).
+                // If we do have a repeat load, we can remove the current load and map its result to the previous loads result.
+                if let Some(last_load) = references.last_loads.get(&address) {
```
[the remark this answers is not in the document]
vezenovm: > If v1 aliases v0 then it looks like we'd set v3 back to the previous value

In `keep_last_load_for` we remove the last load for the store address and all of its aliases.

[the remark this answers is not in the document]
vezenovm: I have added an additional unit test for this case.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -679,6 +678,7 @@ impl<'f> PerFunctionContext<'f> {
         let stores_were_removed =
             self.remove_remaining_last_stores(&removed_loads, &remaining_last_stores);
 
+        println!("{}", self.inserter.function);
```
TomAFrench:
```suggestion
```

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -1708,4 +1749,384 @@ mod tests {
...
+
+        let _ = builder.insert_binary(v15, BinaryOp::Eq, two);
+        builder.insert_constrain(v17, two, None);
+
+        builder.terminate_with_return(vec![]);
+
+        let ssa = builder.finish();
+
+        // Expected result:
+        // acir(inline) fn main f0 {
```
vezenovm, line 2085:
```suggestion
        // Expected result:
        //
        // acir(inline) fn main f0 {
```

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -1972,6 +1975,159 @@ mod tests {
...
+        //       v14 = load v13
+        //       constrain v14 == Field 2
+        //       jmp b4()
+        //     b4():
+        //       v15 = load v11
+        //       store Field 2 at v13
+        //       v16 = load v11
+        //       v17 = eq v15, Field 2
+        //       constrain v16 == Field 2
+        //       return
```
jfecher, line 2016: This doesn't quite cover the case in my last comment since from looking at this one block we can find out that v15 and v16 are aliases. We should have block arguments instead that are both aliased to the same value

jfecher: Also, why is the test so long? Do you need more than 2 blocks?

jfecher:
I tried this with the aliasing test in https://github.com/noir-lang/noir/issues/6120 and compared to master there's another regression. The output with this PR is:
```
acir(inline) fn main f0 {
  b0(v3: u1):
    jmpif v3 then: b1, else: b2
  b1():
    v11 = allocate
    jmp b3(v3, v3)
  b3(v6: &mut Field, v7: &mut Field):
    v12 = load v6
    store Field 2 at v7
    constrain v12 == Field 0
    constrain v12 == Field 2
    return 
  b2():
    v10 = allocate
    jmp b3(v3, v3)
}
```
The results of both constrains are replaced with `v12` even though they should be different values due to the `store Field 2 at v7` where `v7` aliases `v6`. Originally the `== FIeld 2` was done on a load that came after the store, which should then also hold 2.

jfecher:
Original example from #6120 had a typo. After fixing the typo it works on master but is still seeing a regression in this PR. Here's the final output:
```
acir(inline) fn main f0 {
  b0(v3: u1):
    jmpif v3 then: b1, else: b2
  b1():
    v11 = allocate
    store Field 0 at v11
    jmp b3(v11, v11)
  b3(v6: &mut Field, v7: &mut Field):
    v12 = load v6
    store Field 2 at v7
    constrain v12 == Field 0
    constrain v12 == Field 2
    return 
  b2():
    v10 = allocate
    store Field 1 at v10
    jmp b3(v10, v10)
}
```
I've fixed the typo in the issue so that this can be reproduced

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -599,10 +647,32 @@ impl<'f> PerFunctionContext<'f> {
...
+                                }
                             }
                         }
                     }
                 }
+
+                // Set the aliases of the parameters
+                for (_, aliased_params) in arg_set {
+                    for param in aliased_params.iter() {
+                        self.set_aliases(
```
vezenovm, line 669: We were adding parameters as aliases to the arguments, but we were missing actually whether the parameters alias each other if they have been passed the same argument.

vezenovm:
> I tried this with the aliasing test in #6120 and compared to master there's another regression

I have switched `keep_repeat_loads_with_alias_store` to the test in the issue with an extra parameter to also check that we are accounting for parameters with more just one other alias.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -589,6 +630,13 @@ impl<'f> PerFunctionContext<'f> {
             TerminatorInstruction::Jmp { destination, arguments, .. } => {
                 let destination_parameters = self.inserter.function.dfg[*destination].parameters();
                 assert_eq!(destination_parameters.len(), arguments.len());
+                // let parameters_set = destination_parameters.iter().collect::<HashSet<_>>();
```
vezenovm, line 633:
```suggestion
```

vezenovm:
> **brillig_rc_regression_6123**	+19 ❌	**+10.16%**
> **fold_2_to_17**	+42 ❌	**+6.89%**
> **bench_2_to_17**	+21 ❌	**+6.60%**
> **fold_numeric_generic_poseidon**	+51 ❌	**+6.57%**
> **no_predicates_numeric_generic_poseidon**	+51 ❌	**+6.57%**
> **poseidon2**	+21 ❌	**+6.54%**

Looks like we are getting various regressions now after [the RC correctness fix](https://github.com/noir-lang/noir/pull/6134). Going to table this PR for now, and look at further optimizing RC instruction removals .

vezenovm:
Following some regressions from PR #6505 (https://github.com/noir-lang/noir/pull/6505#issuecomment-2482557984) I decided to update this PR with master. It looks like we have some benefits from the optimizations aside for `brillig_rc_regression_6123`. However, it looks to be a minor regression so I am marking this PR ready for review again.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -1044,4 +1114,322 @@ mod tests {
         let main = ssa.main();
         assert_eq!(count_loads(main.entry_block(), &main.dfg), 1);
     }
+
+    #[test]
+    fn remove_repeat_loads() {
```
TomAFrench, line 1119: We now have a shiny new SSA parser thanks to Ary. Can you update these tests to make use of it?

vezenovm: Done. Reduced the diff almost 40% :)

TomAFrench (commented):
still need to do a proper review but a couple of nits

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -583,10 +630,32 @@ impl<'f> PerFunctionContext<'f> {
...
+                                // Check if we have seen the same argument
+                                if let Some(seen_parameters) = arg_set.get_mut(&argument) {
+                                    // Add the current parameter to the parameter we have seen for this argument.
+                                    // The previous parameters and the current one alias one another.
+                                    seen_parameters.insert(*parameter);
+                                } else {
+                                    let mut new_param_set = HashSet::default();
+                                    new_param_set.insert(*parameter);
+                                    arg_set.insert(argument, new_param_set);
+                                }
```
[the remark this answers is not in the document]
vezenovm: yeah it is, updated

jfecher (approved):
LGTM. I'm still not 100% on the interactions with aliases here but I haven't been able to trick it with e.g. nested mutable references to create aliases, passing one of those to a function, mutating it, then loading again trying to get that load removed since only the alias was passed.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -314,6 +324,10 @@ impl<'f> PerFunctionContext<'f> {
             self.remove_stores_that_do_not_alias_parameters(&references);
         }
 
+        // Last loads are truly "per block". During unification we are creating a new block from the current one,
+        // so we must clear the last loads of the current block before we return the new block.
+        references.last_loads.clear();
```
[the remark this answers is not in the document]
vezenovm: I moved the clear to where last_stores is cleared

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -583,10 +630,27 @@ impl<'f> PerFunctionContext<'f> {
...
                     }
                 }
+
+                // Set the aliases of the parameters
+                for (_, aliased_params) in arg_set {
+                    for param in aliased_params.iter() {
+                        self.set_aliases(
+                            references,
+                            *param,
+                            AliasSet::known_multiple(aliased_params.iter().copied()),
```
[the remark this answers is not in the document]
vezenovm: I cannot move out of `aliased_params` as it is re-used in the loop

compiler/noirc_evaluator/src/ssa/opt/mem2reg/alias_set.rs
```diff
@@ -24,6 +24,10 @@ impl AliasSet {
         Self { aliases: Some(aliases) }
     }
 
+    pub(super) fn known_multiple(values: impl IntoIterator<Item = ValueId>) -> AliasSet {
+        Self { aliases: Some(BTreeSet::from_iter(values)) }
+    }
```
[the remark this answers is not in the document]
vezenovm: done

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -574,6 +615,12 @@ impl<'f> PerFunctionContext<'f> {
                 let destination_parameters = self.inserter.function.dfg[*destination].parameters();
                 assert_eq!(destination_parameters.len(), arguments.len());
 
+                // If we have multiple parameters that alias that same argument value,
+                // then those parameters also alias each other.
+                // We save parameters with repeat arguments to later mark those
+                // parameters as aliasing one another.
+                let mut arg_set: HashMap<ValueId, HashSet<ValueId>> = HashMap::default();
```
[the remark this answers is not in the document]
vezenovm: done

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -410,6 +424,28 @@ impl<'f> PerFunctionContext<'f> {
 
                     self.last_loads.insert(address, (instruction, block_id));
                 }
+
+                // Check whether the block has a repeat load from the same address (w/ no calls or stores in between the loads).
+                // If we do have a repeat load, we can remove the current load and map its result to the previous loads result.
+                if let Some(last_load) = references.last_loads.get(&address) {
```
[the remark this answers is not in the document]
vezenovm: It should only work because it is per block. When there is a store we remove the last load for every alias of that store address.

TomAFrench (approved):

PR #6089, opened 2024-09-18
feat: Implement solver for mov_registers_to_registers

sirasistant:
# Description

## Problem\*

mov_registers_to_registers was a suboptimal codegen.

## Summary\*

Moving a set of registers to another set a registers with possible overlap among them is an operation that is used in the codegen for calling conventions. In order to solve that, in the past, we just made a backup of every source register. That is inefficient and generates more than the optimal number of register movements. In order to achieve the optimal number of movements, this PR implements a solver that finds the dependency chains among registers. We then move following the reverse of the dependency chain, and, in the case that the dependency chain is a loop (like MOV 2=>3 MOV 3=>2, but of arbitrary size) we backup only one value to a temporary register.

## Documentation\*

Check one:

# PR Checklist\*


sirasistant:
The savings in the e2e examples are small, but this optimization will be very important when we don't inline the whole program in brillig fns.

sirasistant:
Oh, there are cases where the same register could be moved to more than one destination addresses in mov_registers_to_registers. I need to take that into account, closing it for now

sirasistant:
Added support for that too

compiler/noirc_evaluator/src/brillig/brillig_ir/codegen_stack.rs
```diff
@@ -1,26 +1,290 @@
...
+    pub(crate) fn create_context() -> BrilligContext<FieldElement, Stack> {
+        let mut context = BrilligContext::new(true);
+        context.enter_context(Label::function(FunctionId::test_new(0)));
+        context
+    }
+
+    #[test]
+    #[should_panic(expected = "Multiple moves to the same register found")]
+    fn test_mov_registers_to_registers_overwrite() {
+        let movements = vec![(10, 11), (12, 11), (10, 13)];
```
vezenovm, line 208: What would happen if we have two loops where the first source is the same register? We would also hit this panic right?

vezenovm (approved):

compiler/noirc_evaluator/src/brillig/brillig_ir/codegen_stack.rs
sirasistant: We'd hit that panic because in order to be a loop all the members of the loop need to be written within the loop, so two loops cannot share members because then that member would be written twice

sirasistant: However an offset written to two different targets is a completely reasonable ask, for example if you have



fn foo () -> (Field, Field) {
 (42, 42)
}

PR #6090, opened 2024-09-18
feat: let LSP suggest macro calls too

asterite:
# Description

## Problem

There was no way to get completions for macro calls, so the user had to complete, go back and add the `!`. 

## Summary

Now if something can be a macro call it's suggested too.

A special case is done for `unquote!`: it's only suggested as a macro call, never as a function call.

lsp-suggest-macros

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/lsp/src/requests/completion/completion_items.rs
+++ b/tooling/lsp/src/requests/completion/completion_items.rs
@@ -217,19 +217,24 @@ impl<'a> NodeFinder<'a> {
             )
         };
 
-        // It's unlikely users will define a function named `unquote` that does something different than std's unquote.
-        if name == "unquote" {
-            vec![make_completion_item(true)]
-        } else {
-            let modifiers = self.interner.function_modifiers(&func_id);
+        // When suggestion functions in attributes, never suggest a macro call
+        if attribute_first_type.is_some() {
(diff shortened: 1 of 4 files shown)


tooling/lsp/src/requests/completion/completion_items.rs
```diff
@@ -186,23 +186,71 @@ impl<'a> NodeFinder<'a> {
...
+                func_self_type,
+                function_completion_kind,
+                function_kind,
+                attribute_first_type,
+                self_prefix,
+                is_macro_call,
+            )
+        };
+
+        // When suggestion functions in attributes, never suggest a macro call
```
jfecher, line 220:
```suggestion
        // When suggesting functions in attributes, never suggest a macro call
```

jfecher (approved):
I like that these are only issued for functions returning `Quoted` :+1:

asterite:
I wonder if the return type shown for macro calls should be `-> Quoted`... when I was typing it it was a bit confusing, because the actual type ends up being something different.

What do you think if it's shown as `-> _`?

asterite:
In any case that can be changed later on. I'll merge this for now :-)

Issue #6078, opened 2024-09-18
Types cannot be unquoted to call static methods


Thunkar:
### Aim

Given

```noir
mod types {
    use std::meta::type_of;

    struct Foo {
        x: Field,
    }

    impl Foo {
        fn static() -> Field {
            3
        }
    }

    comptime fn add_static_call(m: Module) -> Quoted {
        // If I want this to be resolved in AMod, I need to use the type, since the struct name is just Foo and I need types::Foo
        let type_of_foo =quote { Foo }.as_type();
        quote {
            fn call_static() -> Field{
                $type_of_foo::static()
            }
        }
    }
}

use types::add_static_call;

#[add_static_call]
mod AMod {}

```

### Expected Behavior

The code above should compile, so that `Foo` can be used in `AMod` even if it was resolved in its own mod.

### Bug

The compiler fails to parse the macro token stream

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

Blocker

### Would you like to submit a PR for this Issue?

None

asterite:
This can be solved with a trick Jake told me today (I needed to do something similar and I didn't want to use a fully qualified name):

```noir
mod types {
    struct Foo {
        x: Field,
    }

    impl Foo {
        fn static() -> Field {
            3
        }
    }

    comptime pub fn add_static_call(m: Module) -> Quoted {
        let static_call = Foo::static;
        quote {
            fn call_static() -> Field {
                $static_call()
            }
        }
    }
}

use types::add_static_call;

#[add_static_call]
mod AMod {}
```

asterite:
I also tried doing this but it didn't work:

```noir
let foo = std::meta::type_of(Foo { x: 1 });
quote {
  fn call_static() -> Field {
    $foo::static()
  }
}
```

The reason is that `$foo` in that parsing position must be a `Path`, not a `Type`.

Unrelated, but I'm thinking it would be nice to be able to somehow convert a type to a `Type`. We can do that with `std::meta::type_of`, but we have to pass an actual instance.

But, we can also do this:

```noir
struct Foo {
    x: Field
}

comptime fn get_type<T>() -> Type {
    let exp: T = std::mem::zeroed();
    std::meta::type_of(exp)
}

fn main() {
    comptime
    {
        let foo = get_type::<Foo>();
        println(foo);
    }
}
```

jfecher:
[...]

PR #6093, opened 2024-09-18
feat: Implement type paths

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6078

## Summary\*

Adds the ability to do `Type::method` where `Type` is a primitive type or interned type. Other named types should already be supported by existing paths.

## Additional Context

I also made a small update to `lookup_method` to support looking up methods without a `self` parameter to allow for us to possibly have something like `Field::static_method()` in the future, although we have no functions like this in the stdlib currently.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/test_programs/compile_success_empty/type_path/src/main.nr
+++ b/test_programs/compile_success_empty/type_path/src/main.nr
@@ -12,6 +12,5 @@ struct Foo {}
 
 impl Foo {
     fn static() {
-        println("Foo::static called");
     }
 }
--- a/test_programs/compile_success_empty/type_path/src/main.nr
+++ b/test_programs/compile_success_empty/type_path/src/main.nr
@@ -1,0 +1,16 @@
+fn main() {
+    comptime {
+        let foo = quote { Foo }.as_type();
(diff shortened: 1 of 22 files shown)


asterite:
>  although we have no functions like this in the stdlib currently

When implementing `Expr::modify` I introduced some `new_let`, `new_if`, etc., functions that I originally wrote as static methods of `Expr`, but then I changed them to regular functions because that didn't work (my idea was to have these `new_` functions actually exposed to users in case they wanted to programmatically build expressions, though in the end maybe that's not very useful).

test_programs/compile_success_empty/type_path/src/main.nr
```diff
@@ -0,0 +1,17 @@
...
+            $foo::static()
+        }
+    }
+}
+
+struct Foo {}
+
+impl Foo {
+    fn static() {
+        println("Foo::static called");
```
jfecher, line 15:
```suggestion
```

Issue #6067, opened 2024-09-17
Synchronize `TypeVariableKind` and `Kind`


michaeljklein:
### Problem

`Kind` and `TypeVariableKind` currently have different behavior for the same name (`Normal`) and because of it, `Type::kind() -> Option<Kind>` currently ignores `TypeVariableKind` info.

### Happy Case

- Match kinds between `TypeVariableKind` and `Kind`
  + Make `TypeVariableKind::Normal` correspond to `Kind::Normal` or else rename to avoid confusion
    * This may require adding some sort of `Kind::Any`/`Kind::Unresolved` or propagating a lot more `Kind` information, e.g. in the generics in `Type::Forall`
    * This will allow removing `None` as an "any" kind in `Type::kind`
- Use the matched kinds to allow `Type::kind()` on type variables to account for the `TypeVariableKind`
- Add tests to ensure moving between types that use `Kind` (e.g. `Type::Constant`) and type variables preserves the kind

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #6094, opened 2024-09-18
feat!: Syncing TypeVariableKind with Kind

michaeljklein:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6067

## Summary\*

- Combines `TypeVariableKind` and `Kind` (now only `Kind`)
- Adds `Kind` arguments where needed
- Adds more `Kind` checks

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -665,6 +694,13 @@ impl std::fmt::Display for Type {
                     write!(f, "{}", binding.borrow())
                 }
             }
+            Type::TypeVariable(binding, TypeVariableKind::Numeric(typ)) => {
+                if let TypeBinding::Unbound(_) = &*binding.borrow() {
+                    write!(f, "{}", typ)
+                } else {
+                    write!(f, "{}", binding.borrow())
+                }
+            }
```
jfecher, line 703: Hmm so if we try to display an array of a certain size we'd get `[Field; 4]` but if the size was unknown we'd display `[Field; u32]`.

Seems a bit confusing so maybe we should just stick to the previous formatting which uses the underscore notation other type variables use `[Field; _]`

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -1351,7 +1382,12 @@ impl Type {
         };
 
         let this = self.substitute(bindings).follow_bindings();
-        if let Some(binding) = this.get_inner_type_variable() {
+        if let Some((binding, kind)) = this.get_inner_type_variable() {
+            // TypeVariableKind::Normal can't bind to numeric Kind's
+            if kind.is_numeric() {
+                return Err(UnificationError)
+            }
```
jfecher, line 1389: Looks like try_bind_to_polymorphic_int is missing this check

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -1594,7 +1640,12 @@ impl Type {
                 // We may have already "bound" this type variable in this call to
                 // try_unify, so check those bindings as well.
                 match bindings.get(id) {
-                    Some((_, binding)) => binding.clone().try_unify(self, bindings),
+                    Some((_, kind, binding)) => {
+                        if !kind.unifies(&binding.kind()) {
+                            return Err(UnificationError);
+                        }
+                        binding.clone().try_unify(self, bindings)
+                    }
```
jfecher, line 1648: I was thinking of if we could still perform this check if we removed kind from TypeBindings and I think we can since it should be the same kind as the type variable that was passed in, so we'd have to add a parameter.

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -2532,13 +2615,19 @@ impl std::hash::Hash for Type {
 
 impl PartialEq for Type {
     fn eq(&self, other: &Self) -> bool {
-        if let Some(variable) = self.get_inner_type_variable() {
+        if let Some((variable, kind)) = self.get_inner_type_variable() {
+            if !kind.unifies(&other.kind()) {
+                return false;
+            }
```
jfecher, line 2621: I think we should avoid unification in the Eq impl

jfecher (commented):
I think we can probably get rid of Kind in TypeBindings/bind and rely on checks in unify instead since we can't issue errors from bind anyway.

michaeljklein:
Note: this is expected to fix https://github.com/noir-lang/noir/issues/6125

compiler/noirc_frontend/src/node_interner.rs
```diff
@@ -1717,7 +1717,7 @@ impl NodeInterner {
         // Replace each generic with a fresh type variable
         let substitutions = impl_generics
             .into_iter()
-            .map(|typevar| (typevar.id(), (typevar, self.next_type_variable())))
+            .map(|typevar| (typevar.id(), (typevar, Kind::Normal, self.next_type_variable())))
```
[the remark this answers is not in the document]
michaeljklein: I'm moving `TypeVariableKind` to `TypeBinding::Unbound` since
- When we have a `TypeBinding::Bound`, we can easily derive the `Kind`
- This ensure we'll always have the kind, i.e. allows `TypeVariable::kind(&self) -> Kind`

compiler/noirc_frontend/src/elaborator/types.rs
```diff
@@ -1565,8 +1572,14 @@ impl<'context> Elaborator<'context> {
...
+                // TODO cleanup
+                // if let Some(error) = NoMatchingImplFoundError::new(self.interner, constraints, span)
+                if let Some(error) =
+                    NoMatchingImplFoundError::new(self.interner, constraints.clone(), span)
                 {
+                    // TODO cleanup
+                    dbg!("ImplSearchErrorKind::Nested", &constraints);
+
                     self.push_err(TypeCheckError::NoMatchingImplFound(error));
                 }
```
TomAFrench, line 1584: unresolved TODO

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -150,18 +167,35 @@ impl Kind {
...
+            // Kind::Any unifies with everything
+            (Kind::Any, _)
+            | (_, Kind::Any)
+
+            // Kind::Normal unifies with Kind::Integer and Kind::IntegerOrField
+            | (Kind::Normal, Kind::Integer | Kind::IntegerOrField)
+            | (Kind::Integer | Kind::IntegerOrField, Kind::Normal)
+
+            // Kind::Integer unifies with Kind::IntegerOrField
+            | (Kind::Integer | Kind::IntegerOrField, Kind::Integer | Kind::IntegerOrField) => true,
```
TomAFrench, line 196:
```suggestion
            // Kind::Any unifies with everything
            (Kind::Any, _) | (_, Kind::Any) => true

            // Kind::Normal unifies with Kind::Integer and Kind::IntegerOrField
            (Kind::Normal, Kind::Integer | Kind::IntegerOrField)
            | (Kind::Integer | Kind::IntegerOrField, Kind::Normal) => true

            // Kind::Integer unifies with Kind::IntegerOrField
            (Kind::Integer | Kind::IntegerOrField, Kind::Integer | Kind::IntegerOrField) => true,
```

Nit but I think this is clearer where we just have separate branches for these cases. If clippy complains then we can silence it.

compiler/noirc_frontend/src/monomorphization/mod.rs
```diff
@@ -916,20 +923,31 @@ impl<'interner> Monomorphizer<'interner> {
                     ast::Expression::Ident(ident)
                 }
             },
-            DefinitionKind::GenericType(type_variable) => {
+            DefinitionKind::GenericType(type_variable, numeric_typ) => {
```
TomAFrench, line 926: It seems like we should rename this to `DefinitionKind::NumericGeneric` as it's not really a type and they're just carrying the value of the numeric generic. Shouldn't be done in this PR however.

TomAFrench (commented):
Looks good to me. There's a few small bits to clean up before we merge.

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -134,7 +134,24 @@ pub enum Type {
 /// is expected (such as in an array length position) are expected to be of kind `Kind::Numeric`.
 #[derive(PartialEq, Eq, Clone, Hash, Debug, PartialOrd, Ord)]
 pub enum Kind {
+    /// Can bind to any type
+    Any,
```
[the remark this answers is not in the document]
michaeljklein: I believe that this or a roughly equivalent `Kind::Unresolved` is required for `_` types, e.g.
```noir
let x: Field = 3;
let z: _ = x;
```

And also unspecified types, i.e. a `UnresolvedTypeData::Unspecified` that's filled in later.

(In the `no implicit numeric generics` PR, I used `Option<Kind>` where the `None` case was used for these unspecified cases.)

However, I'm not sure how a `Kind::Unresolved` would be different. Perhaps we could throw an error if a `Kind::Any`/`Kind::Unresolved` is reached during monomorphization? Though that might be able to occur in valid code if a `_`-typed variable is completely unused.

Regardless, I'll comment and make an issue to follow up, since we'll likely want the metaprogramming builtin for fresh type variables to take an explicit `Kind` (or at least offer a variant that does).

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -2244,7 +2349,8 @@ fn convert_array_expression_to_slice(
 
 impl BinaryTypeOperator {
     /// Perform the actual rust numeric operation associated with this operator
-    pub fn function(self, a: u32, b: u32) -> Option<u32> {
+    // TODO: the Kind is included since it'll be needed for size checks
```
vezenovm, line 2352: unresolved todo, issue?

compiler/noirc_frontend/src/tests.rs
```diff
@@ -1724,11 +1721,6 @@ fn numeric_generic_as_param_type() {
         errors[1].0,
         CompilationError::TypeError(TypeCheckError::TypeKindMismatch { .. }),
     ));
-    // Error from the return type
```
vezenovm, line 1727: We do not error for the return type anymore? I would still expect this error.

compiler/noirc_frontend/src/hir/def_collector/errors.rs
```diff
@@ -100,6 +100,19 @@ impl DefCollectorErrorKind {
     }
 }
 
+impl<'a> From<&'a UnsupportedNumericGenericType> for Diagnostic {
+    fn from(error: &'a UnsupportedNumericGenericType) -> Diagnostic {
+        let name = &error.ident.0.contents;
+        let typ = &error.typ;
+
+        Diagnostic::simple_error(
+            format!("{name} has a type of {typ}. The only supported types of numeric generics are integers, fields, and booleans."),
```
vezenovm, line 109:
```suggestion
            format!("{name} has a type of {typ}. The only supported numeric generic types are `u1`, `u8`, `u16`, and `u32`."),
```
I see this was copied from the error that existed before. Let's just update it to match the error in the parser. I guess this hasn't been hit as the parser has been catching it, but good to be aligned here.

vezenovm (commented):
Couple nits, otherwise looks good.

compiler/noirc_frontend/src/hir_def/types.rs
michaeljklein: Ah yeah, need to link [here](https://github.com/noir-lang/noir/pull/6137)

compiler/noirc_frontend/src/monomorphization/mod.rs
michaeljklein: https://github.com/noir-lang/noir/issues/6181

compiler/noirc_frontend/src/hir_def/types.rs
michaeljklein: The `Kind` in `TypeBindings` can be more specific than the other `Kind`, e.g. without storing it in `TypeBindings`, we can lose the fact that we've specialized a `Kind::IntegerOrField` to a `Kind::Integer` during unification, and thus lose the ability to revert the type binding.

compiler/noirc_frontend/src/tests.rs
michaeljklein: I found and fixed two bugs from this:
- It was actually the error from the parameter type (added tests to distinguish the cases)
- The parameter types were being defaulted to `Kind::Any`, but only `Kind::Normal` should be allowed

compiler/noirc_frontend/src/tests.rs
```diff
@@ -1726,9 +1723,71 @@ fn numeric_generic_as_param_type() {
     ));
     // Error from the return type
     assert!(matches!(
-        errors[2].0,
+        errors[1].0,
```
vezenovm, line 1726:
```suggestion
        errors[2].0,
```
This should still be 2 no?

compiler/noirc_frontend/src/elaborator/types.rs
```diff
@@ -1564,7 +1571,10 @@ impl<'context> Elaborator<'context> {
                 self.push_err(TypeCheckError::TypeAnnotationsNeededForMethodCall { span });
             }
             ImplSearchErrorKind::Nested(constraints) => {
-                if let Some(error) = NoMatchingImplFoundError::new(self.interner, constraints, span)
+                // TODO cleanup
+                // if let Some(error) = NoMatchingImplFoundError::new(self.interner, constraints, span)
+                if let Some(error) =
+                    NoMatchingImplFoundError::new(self.interner, constraints.clone(), span)
```
vezenovm, line 1577:
```suggestion
                if let Some(error) = NoMatchingImplFoundError::new(self.interner, constraints, span)
```

vezenovm (approved):

PR #6095, opened 2024-09-19
fix: decode databus return values

### LINKED ISSUE #5715: NoirJS doesn't report return values when doing witness gen on a function with databus (return_data)
### Aim

Get the return values along with the witness when calling a function with return_data, to pass them to the next circuit in the IVC chain.

### Expected Behavior

As the prover entity, I should be able to get the databus return_data to pass it to the next circuit for recursive verification.

### Bug

They are not due to this line https://github.com/noir-lang/noir/blob/master/tooling/noirc_abi/src/lib.rs#L351

### To Reproduce

1. Create a noir circuit with return_data
2. Call it via noirjs

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

TomAFrench:
# Description

## Problem\*

Resolves #5715

## Summary\*

This PR picks up from @guipublic's work on #6074 to enable decoding databus return values in `noirc_abi`. I've added a test to show this working

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/acir_gen/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/acir_gen/mod.rs
@@ -450,7 +450,12 @@ impl<'a> Context<'a> {
         warnings.extend(self.acir_context.warnings.clone());
 
         // Add the warnings from the alter Ssa passes
-        Ok(self.acir_context.finish(input_witness, return_witnesses, warnings))
+        Ok(self.acir_context.finish(
+            input_witness,
+            // Don't embed databus return witnesses into the circuit.
+            if self.data_bus.return_data.is_some() { Vec::new() } else { return_witnesses },
+            warnings,
+        ))
     }
 
     fn initialize_databus(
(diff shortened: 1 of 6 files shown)


TomAFrench:
@guipublic Can you take a look at this?

guipublic (changes requested):
Either we write the databus return values like it is done in this PR, and then I think we need to re-work noirc_abi/lib.rs:
- has_public_inputs() should only test for public parameter
- encode() may need (I'm not completely sure) to specify return_data or public for the return value, or at least change the comment about the verifier.
And we should also change the prover so that the 'databus' return values are not added to the proof as public inputs.

Or we do not, and the prover can use the witness index of the return values in the ABI to recover them from the witness map.

TomAFrench:
Thanks, I've updated `has_public_inputs`.

Although I'm not clear on your second point. We shouldn't need to update the `encode` function afaik and the databus return values are not added to the public inputs (the ABI doesn't specify this but it's embedded in the circuit.)

guipublic:
I just tried and I think I am getting public inputs in the proof, but we can say that it's a prover issue, as long as it can get the required info, which is the ABI, so it's fine.

guipublic (approved):

PR #6103, opened 2024-09-19
feat: represent assertions more similarly to function calls

### LINKED ISSUE #6102: Improve how `assert` and `assert_eq` are parsed and represented
### Problem

Right now `assert` and `assert_eq` have specific parse code to make sure we parse the correct number of arguments. This has a few issues:
1. Not passing the correct number of arguments leads to a parsing error, which might be more confusing than "wrong number of arguments"
2. If the number of arguments is wrong, and because it's a parser failure, LSP features like autocompletion stop working

### Happy Case

When passing a wrong number of arguments to `assert` and `assert_eq` I should get a "wrong number of arguments" errors, not a parsing error. And LSP features like autocompletion should still work.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #6102

## Summary

See the relevant issue, but in addition to LSP working better, and getting a better error message when giving an incorrect number of arguments to `assert` and `assert_eq`, I think the code ends up being slightly simpler.

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/elaborator/statements.rs
```diff
@@ -110,12 +113,60 @@ impl<'context> Elaborator<'context> {
...
+        let min_args_count = match stmt.kind {
+            ConstrainKind::Assert | ConstrainKind::Constrain => 1,
+            ConstrainKind::AssertEq => 2,
+        };
+        let max_args_count = min_args_count + 1;
+        let actual_args_count = stmt.arguments.len();
+
+        let (message, expr) = if actual_args_count < min_args_count
+            || actual_args_count > max_args_count
+        {
```
jfecher, line 130:
```suggestion
        let (message, expr) = if !(min_args_count ..= max_args_count).contains(actual_args_count) {
```
Could be cleaner?

compiler/noirc_frontend/src/elaborator/statements.rs
```diff
@@ -110,12 +113,60 @@ impl<'context> Elaborator<'context> {
...
+            self.push_err(TypeCheckError::AssertionParameterCountMismatch {
+                expected1: min_args_count,
+                expected2: max_args_count,
+                found: actual_args_count,
+                span,
+            });
```
jfecher, line 136: `expected1` and `expected2` sound like vague names for this field. Maybe just only pass the minimum argument count since the max is always that + 1? Or rename the fields to min/max arg count.

compiler/noirc_frontend/src/elaborator/statements.rs
```diff
@@ -110,12 +113,60 @@ impl<'context> Elaborator<'context> {
...
+            let message = None;
+            let kind = ExpressionKind::Literal(crate::ast::Literal::Bool(true));
+            let expr = Expression { kind, span };
+            (message, expr)
+        } else {
+            let message = if actual_args_count == min_args_count {
+                None
+            } else {
+                Some(stmt.arguments.pop().unwrap())
+            };
```
jfecher, line 149:
```suggestion
            let message = (actual_args_count != min_args_count).then(|| {
                stmt.arguments.pop().unwrap()
            });
```
More nitpicks

compiler/noirc_frontend/src/hir/type_check/errors.rs
```diff
@@ -260,6 +262,11 @@ impl<'a> From<&'a TypeCheckError> for Diagnostic {
                 let msg = format!("Function expects {expected} parameter{empty_or_s} but {found} {was_or_were} given");
                 Diagnostic::simple_error(msg, String::new(), *span)
             }
+            TypeCheckError::AssertionParameterCountMismatch { expected1, expected2, found, span } => {
+                let was_or_were = if *found == 1 { "was" } else { "were" };
+                let msg = format!("Function expects {expected1} or {expected2} parameters but {found} {was_or_were} given");
```
jfecher, line 267: I wonder if we should explicitly say assert or assert_eq here. Since they aren't really functions since they're not first class like the others are. I suppose if we passed the ConstrainKind we also wouldn't need to pass the expected argument count anymore either. It could be a method on ConstrainKind.

asterite: Done! I hesitated between that method returning somehow the min and max values, but ended up just returning the minimum arguments count and assuming the caller would know the maximum is that plus one (maybe it's clearer).

jfecher (approved):
:+1:

PR #6105, opened 2024-09-19
fix: Allow macros to change types on each iteration of a comptime loop

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/pull/6086#issuecomment-2361085425

## Summary\*

Allows macros to have differing result types on different loop iterations again after it was banned by the unification in #6086.

This is a difficult feature to use since users will still have to work around the type system only type checking loops once. The recommended technique found for this PR is to delay type checks by quoting the relevant function calls that need to be re-typechecked so that when they are unquoted by the interpreter they are forced to be typechecked every loop iteration.

The downside of this is that it requires some knowledge of when to do this. If we go back to the naive approach of `comptime_change_type_each_loop_iteration` before the quoting:

```noir
fn main() {
    comptime {
        for i in 9..11 {
            // Lengths are different on each iteration: "foo9", "foo10"
            let name = f"foo{i}".as_ctstring().as_quoted_str!();
            let hash = from_signature(name);
            assert(hash > 3);
        }
    }
}

fn from_signature<let N: u32>(_signature: str<N>) -> u32 {
    N
}
```

We get a confusing error now that the unification is gone: "Non-integer array length `_`" pointing to the body of `from_signature`. The fix for this is to quote & unquote the `from_signature` call:

```noir
let hash = unquote!(quote { from_signature($name) });
```

But there's nothing telling users this. Ideally we could warn users when a type changes like this and perhaps have them opt-in to it with a "I know what I'm doing" of some kind.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
As another consequence of this, the `macro_result_type` test added in the previous PR has also been moved to compile_failure

asterite (approved):
Looks great! I also tried on the code I have in Aztec-Packages and it works fine.

PR #6107, opened 2024-09-19
fix(mem2reg): Remove possibility of underflow

vezenovm:
# Description

## Problem\*

Resolves ICE ran into by @asterite 

## Summary\*

After #5925 we have a couple counters that both increment and decrement. In theory they are never supposed to underflow. @asterite has discovered a bug with the remaining last stores counter. I would like to implement the remaining last stores removal in a better way that removes the need for this counter to be decremented, but for now I just block if by check we are not decrementing zero.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -343,6 +347,7 @@ impl<'f> PerFunctionContext<'f> {
             self.increase_load_ref_counts(value);
         }
 
+        // let mut values_to_reduce_counts = Vec::new();
```
vezenovm, line 350:
```suggestion
```

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -617,6 +613,14 @@ impl<'f> PerFunctionContext<'f> {
         }
     }
 
+    fn reduce_load_result_count(&mut self, value: ValueId) {
+        if let Some(context) = self.load_results.get_mut(&value) {
+            if context.uses != 0 {
```
vezenovm, line 618: I added this to the load result decrement, even though it is not the ICE that @asterite was able to trigger. This should theoretically never underflow though if we are appropriately incrementing the counter, so perhaps we should remove the check here so that we can be notified about the load result usage counter being incremented or decremented incorrectly.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -741,7 +745,9 @@ impl<'f> PerFunctionContext<'f> {
                 if all_loads_removed && !store_alias_used {
                     self.instructions_to_remove.insert(*store_instruction);
                     if let Some((_, counter)) = remaining_last_stores.get_mut(store_address) {
-                        *counter -= 1;
+                        if *counter != 0 {
```
vezenovm, line 748: >  I would like to implement the remaining last stores removal in a better way that removes the need for this counter to be decremented, but for now I just block if by check we are not decrementing zero.

From PR description

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -617,6 +613,14 @@ impl<'f> PerFunctionContext<'f> {
         }
     }
 
+    fn reduce_load_result_count(&mut self, value: ValueId) {
+        if let Some(context) = self.load_results.get_mut(&value) {
+            if context.uses != 0 {
+                context.uses -= 1;
+            }
```
jfecher, line 620:
```suggestion
            context.uses = context.uses.saturating_sub(1);
```

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -741,7 +745,9 @@ impl<'f> PerFunctionContext<'f> {
                 if all_loads_removed && !store_alias_used {
                     self.instructions_to_remove.insert(*store_instruction);
                     if let Some((_, counter)) = remaining_last_stores.get_mut(store_address) {
-                        *counter -= 1;
+                        if *counter != 0 {
+                            *counter -= 1;
+                        }
```
jfecher, line 750:
```suggestion
                        *counter = counter.saturating_sub(1);
```

jfecher (approved):

PR #6112, opened 2024-09-20
fix: disambiguate field or int static trait method call

### LINKED ISSUE #6106: Methods on primitive integers get mixed
### Aim

The following code:

```noir
trait Read {
    fn read(data: [Field; 1]) -> Self;
}

impl Read for Field {
    fn read(data: [Field; 1]) -> Self {
        data[0]
    }
}

impl Read for u32 {
    fn read(data: [Field; 1]) -> Self {
        data[0] as u32
    }
}

fn main() {
    let data = [1];
    let _: u32 = u32::read(data);
}
```

doesn't compile because the compiler thinks `u32::read` returns a `Field`.

### Expected Behavior

The above code compiles without error.

### Bug

The code above errors. I think it's because we store all integer-and-field methods in a single bucket (not sure!)

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #6106

## Additional Context

Let me know if this isn't the correct way to solve this.

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/node_interner.rs
```diff
@@ -2283,20 +2304,28 @@ impl Methods {
...
-                        return Some(method);
+                        candidate = Some(method);
+
+                        // If we recorded the concrete type this trait impl method belongs to,
+                        // and it matches typ, it's an exact match and we return that.
+                        if let Some(method_type) = method_type {
+                            if typ == method_type {
+                                break;
+                            }
+                        }
```
jfecher, line 2320: Always avoid raw type equality! You should prefer `unify` or `try_unify` since only those will handle type variables and arithmetic generics correctly. In this case we should do the same we do on line 2305 where we check if try_unify is successful before applying the bindings and returning.

Also - why change this to use `candidate` instead of an early return? Looks like it requires more unnecessary code to me.

test_programs/compile_success_empty/field_or_integer_static_trait_method/src/main.nr
```diff
@@ -0,0 +1,21 @@
...
+impl Read for u32 {
+    fn read(data: [Field; 1]) -> Self {
+        data[0] as u32
+    }
+}
+
+fn main() {
+    let data = [1];
+    let _: u32 = u32::read(data);
+    let _: Field = Field::read(data);
```
jfecher, line 20: Can you add asserts on the return values so that we can ensure the correct version is called?

E.g. each impl could add a different number to the result.

jfecher (changes requested):
This is a good approach I didn't think of, but we should avoid raw type equality.

compiler/noirc_frontend/src/node_interner.rs
jfecher: This check is only done for primitive types but I think there may still be room for type variables, particularly ones bound to primitive types, to sneak in. It's also good practice to just avoid `==` for types in general IMO. It's just an unfortunate necessity to allow types in hashmaps in certain cases, although that does also mean you have to ensure there are no type variables in those types or be alright with some duplication.

asterite: > Always avoid raw type equality!

I was thinking that if there's no self param we'll always have a concrete type... but I guess that if you have a `T` in some generic code and you do `T::method` that `T` could be a `FieldOrInteger`... not sure! In any case I'll change it.

> Also - why change this to use candidate instead of an early return?

I guess that makes sense. I was thinking that we don't always record the actual self type so it might not be correct to stop at the first match, but now I'm thinking that for a give type all of the methods will have a "self type" recorded, or all of them won't, so stopping at the first match is correct. I'll change it too.

asterite: Any reason why `try_unify` is called on line 2305 instead of `unify`?

jfecher: > Any reason why try_unify is called on line 2305 instead of unify?

Hmm looks like it can be replaced. It was either a mistake or old enough from when `unify` unified type variables as it went even before it knew if the result succeeded. That should've been before we had `try_unify` at all though.

jfecher (approved):
LGTM

You can change the try_unify to unify if you want

test_programs/execution_success/field_or_integer_static_trait_method/src/main.nr
```diff
@@ -0,0 +1,25 @@
+trait Read {
```
vezenovm, line 1: Can this test be moved to compile_success_empty?

asterite: It has a couple of `assert_eq`. I could change the code to run at comptime instead. Would that be okay?

jfecher: assert_eq should still be optimized out as long as it doesn't use program inputs

asterite: Nice, I didn't know that

PR #6115, opened 2024-09-20
feat: add crsPath param to BackendOptions

saleel:
# Description

Add crsPath param to `BackendOptions` so they can be used in  `UltraHonkBackend`, `BarretenbergBackend`, `BarretenbergVerifier` and `UltraHonkVerifier` to override the default CRS download path.

## Problem\*

The default path of `$HOME/.bb-crs` might not be always available in all deployments - for example inside a serverless function (NextJS API, AWS Lambda) only `/tmp` is available for writing files.

## Summary\*

`Crs.new()` already expose a parameter to override the download path, and in this PR we are just adding it to backend and verifier classes.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench (changes requested):
This conflicts with https://github.com/AztecProtocol/aztec-packages/pull/8677.

TomAFrench:
My general thinking on this package is that the direction of travel is that we should be upstreaming as much of this as possible. We could then just have the constructor take one of the verifier instances produced by bb.js rather than constructing one internally

saleel:
Closed this and opened https://github.com/AztecProtocol/aztec-packages/pull/8775

PR #6116, opened 2024-09-20
feat(perf): Expand RC optimization pass to search return block for inc_rcs

vezenovm:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/pull/6088#issuecomment-2359249383  (at least attempts to do so)

Part of general effort to lower Brillig bytecode sizes. 

## Summary\*

inc/dec rc instructions take 3 brillig opcodes. We should look to remove unnecessary ones. The current RC opt pass is very conservative. This PR is an initial effort to expand the scope of the RC opt pass. 

We currently only search for inc_rc instructions issued in the entry block. This greatly simplifies the pass as we do not have to account for inc/dec rc in loops. However, a simple low-hanging fruit would be to also search for inc_rc instructions in an exit block with a return terminator as we know that we are no longer in a loop after the exit block is executed. 

I also updated the array sets scan to not just keep all inc rcs of the same type. Now we keep all rcs with an array set of the same type where rc instruction has the same value or the rc value comes from a function of block param.

I also moved the RC opt pass to happen after simplifying the cfg. After simplification we may have blocks that can be simplified into the return block for which we would not be catching these possible optimizations.

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
Hmm it does feel like marginal benefits for running this pass again. Although in general we can expand the scope of the RC opt pass, perhaps just not with this PR. I'll leave it up for comments though.

jfecher (commented):
Benefits seem a bit marginal to me to add another SSA pass for it.
It'd be nice if we had timing metrics for the compiler itself so we could see exactly how much new compiler passes impact it.

vezenovm:
> Benefits seem a bit marginal to me to add another SSA pass for it.

Yeah I agree. I'm going to look at how I can continue to make the pass less conservative where it may justify running the pass again.

> It'd be nice if we had timing metrics for the compiler itself so we could see exactly how much new compiler passes impact it.

Agreed, that would definitely helpful info as we continue to update the SSA pass.

vezenovm:
Closing as we solved the poseidon case this PR was looking to solve by searching for these useless inc_rc/dec_rc pairing per block rather than per function https://github.com/noir-lang/noir/pull/6160. #6160 also applied optimizations to more tests than this PR.

PR #6117, opened 2024-09-20
feat: LSP autocompletion for `TypePath`

asterite:
# Description

## Problem

Autocompletion for `TypePath` didn't exist because it's a recently added feature.

## Summary

Offers suggestions when you type, for example, `u32::`. For this I had to allow parsing a `TypePath` without any ident following it (like we do for paths), still producing a parsing error.

lsp-complete-type-path

Also fixes the issue that built-in types weren't suggest in any position (only where types are expected, like function argument types, but now that you can call static methods on these types it makes sense to always offer them):

lsp-suggest-type

## Documentation

Check one:

# PR Checklist


jfecher (approved):
:+1:

PR #6135, opened 2024-09-23
fix: let token pretty printer handle `+=` and similar token sequences

asterite:
# Description

## Problem

Just a small thing: we have a token pretty printer and it was printing a sequence of tokens like `+ =` like "+ =" instead of not putting a space between them.

## Summary

Fixes the above.

## Additional Context

I'd add a test, but I get a linker error when trying to run tests on the frontend project since a few weeks/months (I think it's because of chumsky, not sure!). In any case this behavior is not critical, it's just helpful for outputting quoted values at compile-time.

## Documentation

Check one:

# PR Checklist


michaeljklein (changes requested):
Please add tests for this case:
- Quote/unquote
- `nargo fmt` test

Otherwise LGTM

michaeljklein (approved):
LGTM (tested locally)

PR #6137, opened 2024-09-23
feat!: kind size checks

michaeljklein:
# Description

## Problem\*

* Need to check that type constants fit into their `Kind`'s
* The sizes of results from `op.function` and `evaluate_to_u32` are unchecked

## Summary\*

Split out from https://github.com/noir-lang/noir/pull/6094
- Some parts only work with its additional kind information
- Several follow up issues, including:
  * https://github.com/noir-lang/noir/issues/6247
  * https://github.com/noir-lang/noir/issues/6245
  * https://github.com/noir-lang/noir/issues/6238

## Additional Context

TODO:
- [x] Add this test and/or similar execution tests unless we already have a similar one (sanity-check that global `Field` arithmetic works past `u32::MAX`)

```noir
// 2^32 + 1
global A: Field = 4294967297;
global B: Field = 4294967297;
global C: Field = A + B;

fn main() {
    // 2 * (2^32 + 1)
    assert(C == 8589934594);

    let mut leading_zeroes = 0;
    let mut stop = false;
    let bits: [u1; 64] = C.to_be_bits();
    for i in 0..64 {
        if (bits[i] == 0) & !stop {
            leading_zeroes += 1;
        } else {
            stop = true;
        }
    }
    let size = 64 - leading_zeroes;

    // 8589934594 has 34 bits
    assert(size == 34);
}
```

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/hir/type_check/errors.rs
```diff
@@ -227,6 +232,15 @@ impl<'a> From<&'a TypeCheckError> for Diagnostic {
                     *expr_span,
                 )
             }
+            // TODO(https://github.com/noir-lang/noir/issues/6238): implement
+            // handling for larger types
+            TypeCheckError::EvaluatedGlobalPartialSizeChecks { expected_kind, expr_kind, expr_span } => {
+                Diagnostic::simple_warning(
+                    format!("Expected type {expected_kind:?} when evaluating globals: found {expr_kind:?}, size checks are being implemented for this case"),
```
jfecher, line 239:
```suggestion
                    format!("Expected type `{expected_kind}` when evaluating global, found `{expr_kind}`"),
```

compiler/noirc_frontend/src/hir/type_check/errors.rs
```diff
@@ -48,12 +48,17 @@ pub enum TypeCheckError {
...
+    #[error("Expected type {expected_kind:?} when evaluating globals: found {expr_kind:?}, size checks are being implemented for this case")]
+    EvaluatedGlobalPartialSizeChecks { expected_kind: String, expr_kind: String, expr_span: Span },
     #[error("Expected {expected:?} found {found:?}")]
     ArityMisMatch { expected: usize, found: usize, span: Span },
     #[error("Return type in a function cannot be public")]
     PublicReturnType { typ: Type, span: Span },
     #[error("Cannot cast type {from}, 'as' is only for primitive field or integer types")]
-    InvalidCast { from: Type, span: Span },
+    InvalidCast { from: Type, span: Span, reason: String },
+    #[error("Casting value with {from} to a smaller type")]
```
jfecher, line 60: "Casting value with i32 to a smaller type" reads a bit awkwardly

compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
```diff
@@ -215,8 +223,12 @@ impl Type {
                 if l_op == Division {
                     op = op.inverse()?;
                 }
+
+                let divides_evenly = !lhs.infix_kind(rhs).is_field_element()
+                    && l_const.to_i128() % r_const.to_i128() != 0;
```
jfecher, line 228: I think this is backwards. If it divides evenly the remainder should be 0.

We'd also need to check for divide by zero before this

compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
```diff
@@ -237,8 +249,9 @@ impl Type {
     ) -> Result<(), UnificationError> {
         if let Type::InfixExpr(lhs_a, op_a, rhs_a) = self {
             if let Some(inverse) = op_a.inverse() {
-                if let Some(rhs_a_u32) = rhs_a.evaluate_to_u32() {
-                    let rhs_a = Box::new(Type::Constant(rhs_a_u32, lhs_a.infix_kind(rhs_a)));
+                let kind = lhs_a.infix_kind(rhs_a);
+                if let Some(rhs_a_value) = rhs_a.evaluate_to_field_element(&kind) {
+                    let rhs_a = Box::new(Type::Constant(rhs_a_value, lhs_a.infix_kind(rhs_a)));
```
jfecher, line 254:
```suggestion
                    let rhs_a = Box::new(Type::Constant(rhs_a_value, kind));
```

compiler/noirc_frontend/src/parser/parser/type_expression.rs
```diff
@@ -444,7 +427,8 @@ mod tests {
         let UnresolvedTypeExpression::Constant(n, _) = expr else {
             panic!("Expected constant");
         };
-        assert_eq!(n, 42);
+        let fourty_two = 42_u32.into();
+        assert_eq!(n, fourty_two);
```
jfecher, line 431:
```suggestion
        assert_eq!(n, 42_u32.into());
```

compiler/noirc_frontend/src/parser/parser/type_expression.rs
```diff
@@ -496,7 +480,8 @@ mod tests {
...
```
jfecher, line 484:
```suggestion
        assert_eq!(n, 42_u32.into());
```

compiler/noirc_frontend/src/tests.rs
```diff
@@ -3185,17 +3323,17 @@ fn trait_unconstrained_methods_typechecked_correctly() {
...
-            unconstrained fn foo(self) -> u64;
+            unconstrained fn foo(self) -> Field;
         }
 
-        impl Foo for Field {
-            unconstrained fn foo(self) -> u64 {
-                self as u64
+        impl Foo for u64 {
+            unconstrained fn foo(self) -> Field {
+                self as Field
```
jfecher, line 3331: Why was this test changed?

jfecher (changes requested):
Going to finish reviewing hir_def/types.rs and interpreter.rs after lunch

compiler/noirc_frontend/src/elaborator/types.rs
```diff
@@ -417,7 +417,17 @@ impl<'context> Elaborator<'context> {
...
 
-                Some(Type::Constant(self.eval_global_as_array_length(id, path), kind))
+                // TODO(https://github.com/noir-lang/noir/issues/6238):
+                // support non-u32 generics here
+                if !kind.unifies(&Kind::u32()) {
+                    let error = TypeCheckError::EvaluatedGlobalPartialSizeChecks {
+                        expected_kind: Kind::u32().to_string(),
+                        expr_kind: kind.to_string(),
+                        expr_span: path.span(),
+                    };
```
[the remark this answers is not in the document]
michaeljklein: This is a stopgap warning until https://github.com/noir-lang/noir/issues/6238 is addressed. I'd be happy to update the error, but mostly copied from `KindSizeMismatch` and updated the message

compiler/noirc_frontend/src/elaborator/types.rs
```diff
@@ -629,7 +639,9 @@ impl<'context> Elaborator<'context> {
 
         match result.map(|length| length.try_into()) {
             Ok(Ok(length_value)) => return length_value,
-            Ok(Err(_cast_err)) => self.push_err(ResolverError::IntegerTooLarge { span }),
+            Ok(Err(_cast_err)) => {
+                self.push_err(ResolverError::IntegerTooLarge { span });
+            }
```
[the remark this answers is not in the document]
michaeljklein: Ok, I had something similar before and it was rejected by `cargo clippy`, but will attempt to revert

compiler/noirc_frontend/src/tests.rs
michaeljklein: Thanks for catching this: I had changed it to debug `check_cast` and need to change it back

michaeljklein: (Reverted)

compiler/noirc_frontend/src/hir/type_check/errors.rs
michaeljklein: Updated to `Casting value with type {from} to a smaller type ({to})`: ok?

compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
michaeljklein: Updated negation and checking division by zero using `checked_rem`

compiler/noirc_frontend/src/tests.rs
```diff
@@ -1733,7 +1733,7 @@ fn numeric_generic_as_param_type() {
     ));
     // Error from the return type
     assert!(matches!(
-        errors[2].0,
+        errors[1].0,
```
TomAFrench, line 1736: We've got 2 assertions on the same error now.

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -229,6 +241,30 @@ impl Kind {
...
+            None => Some(value),
+            Some(maximum_size) => {
+                if <&T as Into<&FieldElement>>::into(&value) <= &maximum_size {
+                    Some(value)
+                } else {
+                    None
+                }
+            }
+        }
+    }
```
jfecher, line 267:
```suggestion
    fn ensure_value_fits(&self, value: FieldElement) -> Option<FieldElement> {
        match self.integral_maximum_size() {
            None => Some(value),
            Some(maximum_size) => (value <= maximum_size).then_some(value),
        }
    }
```

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -956,6 +1003,15 @@ impl Type {
         matches!(self.follow_bindings(), Type::Integer(_, _))
     }
 
+    fn is_field_element(&self) -> bool {
+        match self.follow_bindings() {
+            Type::FieldElement => true,
+            Type::TypeVariable(var) => var.is_field_element(),
+            Type::Constant(_, kind) => kind.is_field_element(),
```
jfecher, line 1010: It seems odd to return true for both Type::FieldElement and type-level constants of type Field. Do we need both?

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -1838,19 +1893,36 @@ impl Type {
...
+            .and_then(|field_element| field_element.try_to_u32())
+    }
+
+    pub(crate) fn evaluate_to_field_element(&self, kind: &Kind) -> Option<acvm::FieldElement> {
+        if let Some((binding, binding_kind)) = self.get_inner_type_variable() {
             if let TypeBinding::Bound(binding) = &*binding.borrow() {
-                return binding.evaluate_to_u32();
+                if kind.unifies(&binding_kind) {
+                    return binding.evaluate_to_field_element(&binding_kind);
+                }
```
jfecher, line 1905: Hopefully we can find out where we're missing kind checks elsewhere so we can remove these kind checks during evaluation eventually. That's outside this PR though.

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -1838,19 +1893,36 @@ impl Type {
     /// If this type is a Type::Constant (used in array lengths), or is bound
     /// to a Type::Constant, return the constant as a u32.
     pub fn evaluate_to_u32(&self) -> Option<u32> {
-        if let Some((binding, _kind)) = self.get_inner_type_variable() {
+        self.evaluate_to_field_element(&Kind::u32())
+            .and_then(|field_element| field_element.try_to_u32())
```
jfecher, line 1897: Do we have any tests which show that we still error for overflowing in intermediate u32 kind computations?
E.g. the equivalent of `U32_MAX + 1 - X` where X > 0

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -2352,6 +2423,51 @@ impl Type {
...
+
+    pub(crate) fn integral_maximum_size(&self) -> Option<FieldElement> {
+        match self {
+            Type::FieldElement => None,
+            Type::Integer(sign, num_bits) => {
+                let mut max_bit_size = num_bits.bit_size();
+                if sign == &Signedness::Signed {
+                    max_bit_size >>= 1;
+                }
+                Some(((1u128 << max_bit_size) - 1).into())
```
jfecher, line 2435: Shouldn't the signed case be `max_bit_size -= 1`?
We want to divide the total end value by 2 but this is just the bit count

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -2352,6 +2423,51 @@ impl Type {
...
+                let binding = &var.1;
+                match &*binding.borrow() {
+                    TypeBinding::Unbound(_, type_var_kind) => match type_var_kind {
+                        Kind::Any | Kind::Normal | Kind::Integer | Kind::IntegerOrField => None,
+                        Kind::Numeric(typ) => typ.integral_maximum_size(),
+                    },
+                    TypeBinding::Bound(typ) => typ.integral_maximum_size(),
+                }
+            }
+            Type::Alias(alias, _args) => alias.borrow().typ.integral_maximum_size(),
```
jfecher, line 2448: Need to use `.get_type` here to apply generic args

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -2389,15 +2505,36 @@ fn convert_array_expression_to_slice(
...
+                BinaryTypeOperator::Addition => Some(a + b),
+                BinaryTypeOperator::Subtraction => Some(a - b),
+                BinaryTypeOperator::Multiplication => Some(a * b),
+                BinaryTypeOperator::Division => {
+                    if b == FieldElement::zero() {
+                        None
+                    } else {
+                        Some(a / b)
+                    }
+                }
```
jfecher, line 2520:
```suggestion
                BinaryTypeOperator::Division => (b != FieldElement::zero()).then(|| a / b),
```
Nitpick

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -2389,15 +2505,36 @@ fn convert_array_expression_to_slice(
...
+                let maximum_size = maximum_size.to_i128();
+                let a = a.to_i128();
+                let b = b.to_i128();
+
+                let result = match self {
+                    BinaryTypeOperator::Addition => (a + b) % maximum_size,
+                    BinaryTypeOperator::Subtraction => (a - b) % maximum_size,
+                    BinaryTypeOperator::Multiplication => (a * b) % maximum_size,
+                    BinaryTypeOperator::Division => (a.checked_div(b)?) % maximum_size,
+                    BinaryTypeOperator::Modulo => a.checked_rem(b)?,
```
jfecher, line 2533: Looks like we wrap now instead of error on overflow - I think we should keep the error

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
```diff
@@ -895,71 +897,155 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
...
+            result: Option<T>,
+        ) -> IResult<T> {
+            let (lhs, rhs, location) = lhs_rhs_location;
+            result.ok_or(make_error_value(lhs, rhs, location, operator))
+        }
+
+        let location = self.elaborator.interner.expr_location(&id);
+        let error_ctx = (&lhs_value, &rhs_value, location);
+        let make_error = |lhs: Value, rhs: Value, operator| -> IResult<Value> {
+            Err(make_error_value(&lhs, &rhs, location, operator))
```
jfecher, line 931: We can replace this setup along with `make_error_value` with
```rs
        let lhs_type = lhs_value.get_type().into_owned();
        let rhs_type = rhs_value.get_type().into_owned();
        let location = self.elaborator.interner.expr_location(&id);

        let error = |operator| {
            let lhs = lhs_type;
            let rhs = rhs_type;
            InterpreterError::InvalidValuesForBinary { lhs, rhs, location, operator }
        };
```
Then each match case becomes
```rs
                (Value::I8(lhs), Value::I8(rhs)) => {
                    Ok(Value::I8(lhs.checked_add(rhs).ok_or(error("+"))?))
                }
```
and the catch-all make_error cases are now:
```rs
                _ => Err(error("+")),
```

compiler/noirc_frontend/src/tests.rs
michaeljklein: Fixed: it's still passing locally, thankfully

compiler/noirc_frontend/src/hir_def/types.rs
michaeljklein: neat!

compiler/noirc_frontend/src/hir/type_check/errors.rs
jfecher: I don't think we need to - we can add future errors or warnings for this case still

compiler/noirc_frontend/src/hir/type_check/errors.rs
jfecher: How about "Casting value of type {from} to a smaller type ({to})"

"casting value with type" sometimes reads to me like you're casting to the from type.

compiler/noirc_frontend/src/hir_def/types.rs
michaeljklein: Both are currently used, since checking whether the `Type::Constant`'s `Kind` is a field element uses the check for `Type::FieldElement`. I'll split the two versions apart to be more clear.

compiler/noirc_frontend/src/hir_def/types.rs
michaeljklein: Added a follow up issue and TODO for it: https://github.com/noir-lang/noir/issues/6260

compiler/noirc_frontend/src/hir_def/types.rs
michaeljklein: No, only tests that show that we're currently erroring in cases where arithmetic generics are `> u32::MAX`. It would be part of https://github.com/noir-lang/noir/issues/6238 so I've added a requirement to test this there.

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
michaeljklein: That worked with one small change: 
```
            let lhs = lhs_type.clone();
            let rhs = rhs_type.clone();
```
otherwise the double-usage when `try_into()` is called throws an error

compiler/noirc_frontend/src/hir/type_check/errors.rs
jfecher: The colon still reads oddly to me. Also - do we not have a Display impl for kinds?

compiler/noirc_frontend/src/elaborator/types.rs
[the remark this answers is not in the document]
michaeljklein: Updated to `EvaluatedGlobalIsntU32`

jfecher (approved):
👍

compiler/noirc_frontend/src/hir/type_check/errors.rs
michaeljklein: Updated to `"Expected type {expected_kind} when evaluating globals, but found {expr_kind} (this warning may become an error in the future)"`

PR #6149, opened 2024-09-24
feat: allow silencing an unused variable defined via `let`

### LINKED ISSUE #6148: Allow silencing "unused variable" warnings
### Problem

Given some code like this:

```noir
fn main() {
  #[allow(unused_variables)]
  let x = 1;
}
```

the compiler should not produce a warning for that unused variable.

This is especially useful for variables defined via macros where the macro writer can't know if such variable is going to be used or not.

### Happy Case

The compiler doesn't produce a warning.

### Workaround

None

### Workaround Description

The macro could define the variable as `_x` but documenting that variable and using it becomes less ideal.

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #6148

## Summary

lsp-allow-unused

## Additional Context

This is only for `let`, which is what we currently need.

Should this be documented? I guess so, but I couldn't find where in our docs we explain `let`. (on the other hand we have `deprecated` but it seems to also be undocumented)

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/mod.rs
+++ b/compiler/noirc_frontend/src/elaborator/mod.rs
@@ -362,7 +362,12 @@ impl<'context> Elaborator<'context> {
             if let Kind::Numeric(typ) = &generic.kind {
                 let definition = DefinitionKind::GenericType(generic.type_var.clone());
                 let ident = Ident::new(generic.name.to_string(), generic.span);
-                let hir_ident = self.add_variable_decl(ident, false, false, false, definition);
+                let hir_ident = self.add_variable_decl(
+                    ident, false, // mutable
+                    false, // allow_shadowing
+                    false, // warn_if_unused
+                    definition,
+                );
                 self.interner.push_definition_type(hir_ident.id, *typ.clone());
             }
         }
@@ -759,7 +764,7 @@ impl<'context> Elaborator<'context> {
                 typ.clone(),
                 DefinitionKind::Local(None),
                 &mut parameter_idents,
-                true,
+                true, // warn_if_unused
                 None,
             );
 
--- a/compiler/noirc_frontend/src/elaborator/mod.rs
+++ b/compiler/noirc_frontend/src/elaborator/mod.rs
@@ -362,8 +362,7 @@ impl<'context> Elaborator<'context> {
             if let Kind::Numeric(typ) = &generic.kind {
                 let definition = DefinitionKind::GenericType(generic.type_var.clone());
                 let ident = Ident::new(generic.name.to_string(), generic.span);
-                let hir_ident =
-                    self.add_variable_decl_inner(ident, false, false, false, definition);
+                let hir_ident = self.add_variable_decl(ident, false, false, false, definition);
                 self.interner.push_definition_type(hir_ident.id, *typ.clone());
             }
         }
@@ -760,6 +759,7 @@ impl<'context> Elaborator<'context> {
                 typ.clone(),
                 DefinitionKind::Local(None),
                 &mut parameter_idents,
+                true,
                 None,
             );
 
--- a/compiler/noirc_frontend/src/elaborator/patterns.rs
+++ b/compiler/noirc_frontend/src/elaborator/patterns.rs
@@ -26,6 +26,7 @@ impl<'context> Elaborator<'context> {
         pattern: Pattern,
         expected_type: Type,
         definition_kind: DefinitionKind,
+        warn_if_unused: bool,
     ) -> HirPattern {
         self.elaborate_pattern_mut(
             pattern,
@@ -33,6 +34,7 @@ impl<'context> Elaborator<'context> {
             definition_kind,
             None,
             &mut Vec::new(),
+            warn_if_unused,
             None,
         )
     }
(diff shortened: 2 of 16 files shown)


compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -760,6 +759,7 @@ impl<'context> Elaborator<'context> {
                 typ.clone(),
                 DefinitionKind::Local(None),
                 &mut parameter_idents,
+                true,
```
michaeljklein, line 762: Please label this boolean so that it's easier to read

compiler/noirc_frontend/src/elaborator/patterns.rs
```diff
@@ -263,6 +278,7 @@ impl<'context> Elaborator<'context> {
                 definition.clone(),
                 mutable,
                 new_definitions,
+                true,
```
michaeljklein, line 281: Is this the same boolean?

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -2338,6 +2338,7 @@ fn function_def_set_parameters(
                 parameter_type.clone(),
                 DefinitionKind::Local(None),
                 &mut parameter_idents,
+                true,
```
michaeljklein, line 2341: Here as well

compiler/noirc_frontend/src/lexer/token.rs
```diff
@@ -948,6 +952,14 @@ impl SecondaryAttribute {
             SecondaryAttribute::Abi(_) => Some("abi".to_string()),
             SecondaryAttribute::Varargs => Some("varargs".to_string()),
             SecondaryAttribute::UseCallersScope => Some("use_callers_scope".to_string()),
+            SecondaryAttribute::Allow(_) => Some("allow".to_string()),
+        }
+    }
+
+    pub(crate) fn is_allow_unused_variables(&self) -> bool {
+        match self {
+            SecondaryAttribute::Allow(string) => string == "unused_variables",
```
michaeljklein, line 961: Do we also need to prevent someone making a function that they would expect to be usable as a `SecondaryAttribute` named `unused_variables`? (I.e. with a clear error)

compiler/noirc_frontend/src/parser/parser/traits.rs
```diff
@@ -203,7 +203,7 @@ fn trait_implementation_body() -> impl NoirParser<Vec<Documented<TraitImplItem>>
         .map(|(name, alias)| TraitImplItemKind::Type { name, alias });
 
     let let_statement = let_statement(expression()).then_ignore(just(Token::Semicolon)).try_map(
-        |((pattern, typ), expr), span| match pattern {
+        |(_attributes, pattern, typ, expr), span| match pattern {
```
michaeljklein, line 206: Please throw an error on unexpected attributes here

compiler/noirc_frontend/src/tests/unused_items.rs
```diff
@@ -157,3 +160,14 @@ fn errors_on_unused_trait() {
     assert_eq!(ident.to_string(), "Foo");
     assert_eq!(*item_type, "trait");
 }
+
+#[test]
```
michaeljklein, line 164: Please add tests for the struct/trait cases

michaeljklein (changes requested):

compiler/noirc_frontend/src/lexer/token.rs
asterite: Maybe, but so far this isn't done in any other of the other attributes. Or... I guess you could define a function named like that, it's just that you won't be able to use it to run macro code.

compiler/noirc_frontend/src/tests/unused_items.rs
asterite: What should be these cases? This PR only handles local variables.

compiler/noirc_frontend/src/parser/parser/traits.rs
asterite: My thinking was that we'll probably eventually want to support attributes here. Or, put another way, adding attributes here is fine, it's just that this particular item doesn't support any of the existing attributes.

compiler/noirc_frontend/src/lexer/token.rs
asterite: Done! Now attributes are only parsed in a let declaration in statements, never in a trait.

compiler/noirc_frontend/src/parser/parser/traits.rs
michaeljklein: My concern is that any attributes added here are ignored without warning. Since metaprogramming is only partially documented, a user could add an attribute here that silently fails.

compiler/noirc_frontend/src/tests/unused_items.rs
michaeljklein: Checking that attributes on structs/traits throw errors/warnings

asterite: Oh, you mean, if you use `allow(unused_variables)` on a struct or a trait then it should give an error?

Right now we don't error on these cases. For example you can say `#[deprecated] struct Foo {}` and it's completely ignored.

I'm not against adding those checks, I'm just not sure we should do that in this PR.

PR #6154, opened 2024-09-25
feat: expose `derived_generators` and `pedersen_commitment_with_separator` from the stdlib

nventuro:
These are functions that we need to import directly in `aztec-nr`, from discussion with @vezenovm there doesn't seem to be a good reason to not export them.

### DIFF
--- a/noir_stdlib/src/hash/mod.nr
+++ b/noir_stdlib/src/hash/mod.nr
@@ -105,7 +105,7 @@ fn __derive_generators<let N: u32, let M: u32>(
     // Same as from_field but:
     // does not assert the limbs are 128 bits
     // does not assert the decomposition does not overflow the EmbeddedCurveScalar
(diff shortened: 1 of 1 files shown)


noir_stdlib/src/hash/mod.nr
```diff
@@ -105,7 +105,7 @@ fn __derive_generators<let N: u32, let M: u32>(
     // Same as from_field but:
     // does not assert the limbs are 128 bits
     // does not assert the decomposition does not overflow the EmbeddedCurveScalar
-    fn from_field_unsafe(scalar: Field) -> EmbeddedCurveScalar {
+    pub fn from_field_unsafe(scalar: Field) -> EmbeddedCurveScalar {
```
TomAFrench, line 108:
```suggestion
fn from_field_unsafe(scalar: Field) -> EmbeddedCurveScalar {
```

This function is not intended to be public. However I think the other two are fine to be exposed.

TomAFrench (approved):

PR #6158, opened 2024-09-26
feat: Hoist constant allocation outside of loops

sirasistant:
# Description

## Problem\*

Avoids the runtime cost of reinitializing constants inside loops.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/brillig/brillig_gen/constant_allocation.rs
+++ b/compiler/noirc_evaluator/src/brillig/brillig_gen/constant_allocation.rs
@@ -1,7 +1,7 @@
 //! This module analyzes the usage of constants in a given function and decides an allocation point for them.
 //! The allocation point will be the common dominator of all the places where the constant is used.
 //! By allocating in the common dominator, we can cache the constants for all subsequent uses.
-use fxhash::FxHashMap as HashMap;
+use fxhash::{FxHashMap as HashMap, FxHashSet as HashSet};
 
 use crate::ssa::ir::{
     basic_block::BasicBlockId,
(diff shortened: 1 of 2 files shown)


compiler/noirc_evaluator/src/brillig/brillig_gen/constant_allocation.rs
```diff
@@ -114,29 +117,106 @@ impl ConstantAllocation {
...
+            }
+        }
+    }
+
+    blocks_in_loops
+}
+
+/// Return each block that is in a loop starting in the given header block.
+/// Expects back_edge_start -> header to be the back edge of the loop.
+fn find_blocks_in_loop(
```
sirasistant, line 195: This was (mostly) taken from unrolling.rs. Should I extract this to a common module? If so, which?

vezenovm:
Could we not do this during SSA and with more than just constants? We can have a loop invariant code motion pass that moves code that does the same thing to the pre-header of a loop.

sirasistant:
That'd be compatible with the change in this pass, this is for the case where a value is a constant, we can hoist the initialization of the constant. The pass that you're referring to would hoist instructions, however those have a harder requirement since all the dependencies of the instruction need to be available when hoisting it up. SSA currently doesn't have any instruction for initializing a constant, only brillig does so this is why it's done outside of SSA before brillig generation

sirasistant:
That will change for arrays with this one https://github.com/noir-lang/noir/pull/6071

compiler/noirc_evaluator/src/brillig/brillig_gen/constant_allocation.rs
vezenovm: We could probably make this a utility inside of `unrolling.rs` that is re-used in there and in here. Perhaps in the future as we add more loop optimizations we can have a general `loops` module which would also include utility functions like this one.

vezenovm (approved):
Not going to block on making `find_blocks_in_loop` as this can be done in a separate PR. LGTM

TomAFrench:
Agree on not blocking.

PR #6160, opened 2024-09-26
feat(perf): Remove useless paired RC instructions within a block during DIE

vezenovm:
# Description

## Problem\*

Resolves 

Part of general effort to reduce Brillig bytecode sizes.

## Summary\*

We currently have a pass for removing paired `inc_rc` and `dec_rc` instructions as long as there is no array set to an array of the same type between them. However, this is pass only checks the first block for inc_rc instructions and the return block for dec_rc instructions. This is because during SSA gen we add inc_rc instructions on array function parameters and dec_rc parameters when we exit the function scope, so this pass still allows us to get rid of a lot of instructions. However, function scope entry and exit are not the only places we add inc_rc instructions. And once we inline functions and go through other SSA opts, it is common to have inc_rc and dec_rc instructions through a block. 

e.g. This is the SSA of the return block of `execution_success/poseidon2`:
```
  b4():
    v17 = load v10
    v19 = call poseidon2_permutation(v17, u32 4)
    inc_rc v19
    dec_rc v19
    inc_rc v19
    v20 = array_get v19, index u32 0
    dec_rc v19
    constrain v20 == v1
    return 
```
All of those inc_rc/dec_rc pairs are useless and we should be able to remove them. 

This PR removes paired inc_rc/dec_rc instructions per block when there is not an array set to an array of the same type between them.
The return block of `execution_success/poseidon2` after this PR is the following:
```
  b4():
    v17 = load v10
    v19 = call poseidon2_permutation(v17, u32 4)
    v20 = array_get v19, index u32 0
    constrain v20 == v1
    return 
```

## Additional Context

Right now I just focused on removing these inc_rc/dec_rc instructions per block for simplicity. Removing them across blocks is more complex and that work can come in a follow-up.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/die.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/die.rs
@@ -1,7 +1,6 @@
 //! Dead Instruction Elimination (DIE) pass: Removes any instruction without side-effects for
 //! which the results are unused.
-use std::collections::HashSet;
-
+use fxhash::{FxHashMap as HashMap, FxHashSet as HashSet};
 use im::Vector;
 use noirc_errors::Location;
 
@@ -17,6 +16,8 @@ use crate::ssa::{
     },
     ssa_gen::{Ssa, SSA_WORD_SIZE},
 };
+
+use super::rc::{pop_rc_for, IncRc};
 
 impl Ssa {
     /// Performs Dead Instruction Elimination (DIE) to remove any instructions with
@@ -76,6 +77,12 @@ struct Context {
     rc_instructions: Vec<(InstructionId, BasicBlockId)>,
 }
 
+// struct IncRc {
+//     id: InstructionId,
+//     array: ValueId,
+//     possibly_mutated: bool,
+// }
+
 impl Context {
     /// Steps backwards through the instruction of the given block, amassing a set of used values
     /// as it goes, and at the same time marking instructions for removal if they haven't appeared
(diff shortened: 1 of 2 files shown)


compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -76,6 +77,12 @@ struct Context {
     rc_instructions: Vec<(InstructionId, BasicBlockId)>,
 }
 
+// struct IncRc {
+//     id: InstructionId,
+//     array: ValueId,
+//     possibly_mutated: bool,
+// }
+
```
vezenovm, line 85:
```suggestion
```

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -105,6 +106,16 @@ impl Context {
 
         let instructions_len = block.instructions().len();
 
+        // We can track IncrementRc instructions per block to determine whether they are useless.
+        // IncrementRc and DecrementRc instructions are normally side effectual instructions, but we remove
+        // them if their value is not used anywhere in the function. However, even when their value is used, their existence
+        // is extra pointless logic if there is no array set between the increment and the decrement of the reference counter.
```
vezenovm, line 112:
```suggestion
        // is pointless logic if there is no array set between the increment and the decrement of the reference counter.
```

TomAFrench (approved):

PR #6161, opened 2024-09-26
feat: visibility for globals

asterite:
# Description

## Problem

Part of #4515

## Additional Context

I left all of the stdlib globals as private except one which needed to be `pub(crate)` to avoid a warning. I don't know if some of these globals should actually be public.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/tests/unused_items.rs
+++ b/compiler/noirc_frontend/src/tests/unused_items.rs
@@ -239,3 +239,26 @@ fn errors_if_type_alias_aliases_more_private_type_in_generic() {
     assert_eq!(typ, "Foo");
     assert_eq!(item, "Bar");
 }
+
+#[test]
+fn errors_on_unused_global() {
+    let src = r#"
+    global foo = 1;
+    global bar = 1;
+
+    fn main() {
+        let _ = bar;
+    }
+    "#;
+
+    let errors = get_program_errors(src);
+    assert_eq!(errors.len(), 1);
+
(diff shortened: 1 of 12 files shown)


compiler/noirc_frontend/src/tests/unused_items.rs
```diff
@@ -239,3 +239,26 @@ fn errors_if_type_alias_aliases_more_private_type_in_generic() {
...
+    assert_eq!(errors.len(), 1);
+
+    let CompilationError::ResolverError(ResolverError::UnusedItem { ident, item }) = &errors[0].0
+    else {
+        panic!("Expected an unused item error");
+    };
+
+    assert_eq!(ident.to_string(), "foo");
+    assert_eq!(item.item_type(), "global");
+}
```
TomAFrench, line 264:
```suggestion
fn warns_on_unused_global() {
    let src = r#"
    global foo = 1;
    global bar = 1;

    fn main() {
        let _ = bar;
    }
    "#;

    let errors = get_program_errors(src);
    assert_eq!(errors.len(), 1);

    let CompilationError::ResolverError(ResolverError::UnusedItem { ident, item }) = &errors[0].0
    else {
        panic!("Expected an unused item warning");
    };

    assert_eq!(ident.to_string(), "foo");
    assert_eq!(item.item_type(), "global");
}
```

TomAFrench (approved):
Yeah, I think there's not much reason to have stdlib globals to be public.

PR #6165, opened 2024-09-27
feat: visibility for modules

asterite:
# Description

## Problem

Part of #4515

## Additional Context

I made most of the stdlib modules public. These are all except the test modules and some modules that I thought were just implementation details of some other modules.

I'm adding very short documentation for each of these. I'm thinking that once we get all visibility done it should be documented in its own section, with examples.

## Documentation

Check one:

# PR Checklist


TomAFrench:
```noir
fn main(x: Field) {
    let y = foo::bar::baz();
    assert_eq(x, y);
}

mod foo {
    mod bar {
        pub fn baz() -> Field {
            2
        }
    }
}

```

This program currently throws two warnings rather than one.


```
warning: bar is private and not visible from the current module
  ┌─ temp/src/main.nr:2:18
  │
2 │     let y = foo::bar::baz();
  │                  --- bar is private
  │

warning: bar is private and not visible from the current module
  ┌─ temp/src/main.nr:2:18
  │
2 │     let y = foo::bar::baz();
  │                  --- bar is private
  │


```

noir_stdlib/src/ops/mod.nr
```diff
@@ -1,5 +1,5 @@
-mod arith;
-mod bit;
+pub mod arith;
+pub mod bit;
```
TomAFrench, line 2:
```suggestion
mod arith;
mod bit;
```

We reexport anything of interest from these modules so we can keep these private.

TomAFrench (commented):
Looks good, we should add a bunch of frontend tests for this to show that behaviour matches that of rust

asterite:
@TomAFrench it seems that duplicate error also happens with private functions... which means that this error already existed a long time before this PR. Do you think I should try to fix it in this PR?

TomAFrench:
If it's a preexisting error then there's no reason to block this PR over it. We can sort it in a separate PR.

asterite:
> Looks good, we should add a bunch of frontend tests for this to show that behaviour matches that of rust

Do you know what things we could test? To be honest I don't know all the details of Rust's visibility... but if module visibility behaves the same as functions, structs, etc., then it probably works fine as the visibility check is the same for any item.

TomAFrench:
I'd just want to test that you can't reach through private modules to access public contents (except for direct submodules). That's the main case I'd want to test (how I noticed that we threw multiple warnings)

TomAFrench (approved):
Nico's monday is gonna be ruined but LGTM

PR #6166, opened 2024-09-27
feat(ssa): Simplify signed casts

vezenovm:
# Description

## Problem\*

Resolves 

Small optimization I noticed while working on other tasks.

## Summary\*

We currently simplify when casting to a field or an unsigned type. We can also simplify when casting to a signed type. This is an initial version that is quite conservative in the simplification. If we we have a constant that is below `2^(bit_size-1)` when casting to a signed type we return the constant with the new signed type. I also added the case of signed -> field as this can be the same as going unsigned -> field. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/instruction/cast.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/instruction/cast.rs
@@ -27,10 +27,10 @@ pub(super) fn simplify_cast(
                 SimplifiedTo(value)
             }
             (
-                Type::Numeric(NumericType::Unsigned { .. }),
+                Type::Numeric(NumericType::Unsigned { .. } | NumericType::Signed { .. }),
                 Type::Numeric(NumericType::NativeField),
             ) => {
-                // Unsigned -> Field: redefine same constant as Field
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/ir/instruction/cast.rs
```diff
@@ -27,10 +27,10 @@ pub(super) fn simplify_cast(
                 SimplifiedTo(value)
             }
             (
-                Type::Numeric(NumericType::Unsigned { .. }),
+                Type::Numeric(NumericType::Unsigned { .. } | NumericType::Signed { .. }),
                 Type::Numeric(NumericType::NativeField),
             ) => {
-                // Unsigned -> Field: redefine same constant as Field
```
TomAFrench, line 33: This got me thinking that it's a bit strange we convert `-1` to `2**N - 1` rather than to `p-1`. It's preexisting behaviour and would need discussion so it's not a blocker for this PR.

TomAFrench (approved):

PR #6168, opened 2024-09-27
feat(perf): Remove inc_rc instructions for arrays which are never mutably borrowed

vezenovm:
# Description

## Problem\*

## Summary\*

An array can be borrowed without being mutated. We code gen inc_rc instructions quite pessimistically on every borrow or reassignment of an array. If we know that the array is never mutated, it should be safe to remove these inc_rc instructions. 

During DIE we now track all inc_rc instructions in a block and if that array is never used in an array set or a store, we mark its inc_rc instruction for removal.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/die.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/die.rs
@@ -926,7 +926,6 @@ mod test {
         //       return v2
         //   }
         let ssa = ssa.dead_instruction_elimination();
-        println!("{}", ssa);
         let main = ssa.main();
 
         let instructions = main.dfg[main.entry_block()].instructions();
--- a/compiler/noirc_evaluator/src/ssa/opt/die.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/die.rs
@@ -113,8 +113,11 @@ impl Context {
         // We track per block whether an IncrementRc instruction has a paired DecrementRc instruction
         // with the same value but no array set in between.
         // If we see an inc/dec RC pair within a block we can safely remove both instructions.
-        let mut inc_rcs: HashMap<Type, Vec<RcInstruction>> = HashMap::default();
-        let mut inc_rcs_to_remove = HashSet::default();
+        let mut rcs_with_possible_pairs: HashMap<Type, Vec<RcInstruction>> = HashMap::default();
+        let mut rc_pairs_to_remove = HashSet::default();
+
+        let mut inc_rcs: HashMap<ValueId, HashSet<InstructionId>> = HashMap::default();
+        let mut borrowed_arrays: HashSet<ValueId> = HashSet::default();
 
         // Indexes of instructions that might be out of bounds.
         // We'll remove those, but before that we'll insert bounds checks for them.
@@ -146,14 +149,20 @@ impl Context {
             self.track_inc_rcs_to_remove(
                 *instruction_id,
                 function,
+                &mut rcs_with_possible_pairs,
+                &mut rc_pairs_to_remove,
                 &mut inc_rcs,
-                &mut inc_rcs_to_remove,
+                &mut borrowed_arrays,
             );
         }
 
-        for id in inc_rcs_to_remove {
-            self.instructions_to_remove.insert(id);
+        for inc_rc_value in inc_rcs.keys() {
+            if !borrowed_arrays.contains(inc_rc_value) {
+                self.instructions_to_remove.extend(&inc_rcs[inc_rc_value]);
+            }
         }
+
+        self.instructions_to_remove.extend(rc_pairs_to_remove);
 
         // If there are some instructions that might trigger an out of bounds error,
         // first add constrain checks. Then run the DIE pass again, which will remove those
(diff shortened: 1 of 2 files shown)


compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -779,4 +806,132 @@ mod test {
...
+        // Compiling main
+        let mut builder = FunctionBuilder::new("main".into(), main_id);
+        let v0 = builder.add_parameter(Type::Array(Arc::new(vec![Type::field()]), 2));
+        builder.increment_array_reference_count(v0);
+        builder.increment_array_reference_count(v0);
+        builder.increment_array_reference_count(v0);
+
+        let zero = builder.numeric_constant(0u128, Type::unsigned(32));
+        let v1 = builder.insert_array_get(v0, zero, Type::field());
+        // builder.decrement_array_reference_count(v0);
```
vezenovm, line 912:
```suggestion
```

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -779,4 +806,131 @@ mod test {
...
+
+        // Expected output:
+        //
+        // acir(inline) fn main f0 {
+        //     b0(v0: [Field; 2]):
+        //       v2 = array_get v0, index u32 0
+        //       return v2
+        //   }
+        let ssa = ssa.dead_instruction_elimination();
+        println!("{}", ssa);
```
vezenovm, line 929:
```suggestion
```

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -779,4 +806,130 @@ mod test {
...
+        builder.increment_array_reference_count(v0);
+        builder.increment_array_reference_count(v0);
+        builder.increment_array_reference_count(v0);
+
+        let zero = builder.numeric_constant(0u128, Type::unsigned(32));
+        let v1 = builder.insert_array_get(v0, zero, Type::field());
+        builder.terminate_with_return(vec![v1]);
+
+        let ssa = builder.finish();
+        println!("{}", ssa);
```
vezenovm, line 915:
```suggestion
```

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -113,8 +113,13 @@ impl Context {
         // We track per block whether an IncrementRc instruction has a paired DecrementRc instruction
         // with the same value but no array set in between.
         // If we see an inc/dec RC pair within a block we can safely remove both instructions.
-        let mut inc_rcs: HashMap<Type, Vec<RcInstruction>> = HashMap::default();
-        let mut inc_rcs_to_remove = HashSet::default();
+        let mut rcs_with_possible_pairs: HashMap<Type, Vec<RcInstruction>> = HashMap::default();
+        let mut rc_pairs_to_remove = HashSet::default();
+        // We also separately track all IncrementRc instructions and all arrays which have been mutably borrowed.
+        // This is done to determine whether the array has ever been borrowed.
```
vezenovm, line 119:
```suggestion
```

sirasistant (commented):
Is this safe in the presence of calls to other functions? I guess yes since we increase the rc in the called fn?

vezenovm:
> Is this safe in the presence of calls to other functions? I guess yes since we increase the rc in the called fn?

Yes I believe it should be safe due to the fact that if the called function performs a mutation we will keep the rc on the array parameter inside the called fn.

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -146,15 +150,21 @@ impl Context {
...
             );
         }
 
-        for id in inc_rcs_to_remove {
-            self.instructions_to_remove.insert(id);
+        for inc_rc_value in inc_rcs.keys() {
+            if !borrowed_arrays.contains(inc_rc_value) {
+                self.instructions_to_remove.extend(&inc_rcs[inc_rc_value]);
+            }
         }
```
TomAFrench, line 164:
```suggestion
        let unborrowed_arrays = inc_rcs.keys().filter(|id| !borrowed_arrays.contains(id));
        self.instructions_to_remove.extend(unborrowed_arrays);
```

nit

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -779,4 +805,129 @@ mod test {
...
+        // We expect the output to be unchanged
+        let ssa = ssa.dead_instruction_elimination();
+        let main = ssa.main();
+
+        assert_eq!(main.dfg[main.entry_block()].instructions().len(), 3);
+        assert_eq!(main.dfg[b1].instructions().len(), 2);
+    }
+
+    #[test]
+    fn keep_inc_rc_on_borrowed_array_set() {
```
TomAFrench, line 858: Something these tests are missing is a test on how we handle `inc_rc` _after_ when a array is borrowed, e.g.

```
        // acir(inline) fn main f0 {
        //     b0(v0: [u32; 2]):
        //       inc_rc v0
        //       v3 = array_set v0, index u32 0, value u32 1
        //       inc_rc v0
        //       inc_rc v0
        //       inc_rc v0
        //       v4 = array_get v0, index u32 0
        //       return v4
        //   }
```

This would give a little more security against us accidentally removing more instructions than we expect in future changes.

TomAFrench (approved):
LGTM, couple of nits

compiler/noirc_evaluator/src/ssa/opt/die.rs
vezenovm: done

compiler/noirc_evaluator/src/ssa/opt/die.rs
vezenovm: `inc_rcs[inc_rc_value]` actually returns `HashSet` of instructions so I had to change this suggestion to be a `filter_map` with `flatten`. It ends up being longer than the for-loop version, but it is more functional.

PR #6170, opened 2024-09-28
Add globals associated with integers

### LINKED ISSUE #4572: Add limits of each integer type to stdlib
### Problem

Rust has `MIN`, `MAX`, and `BITS` on each integer type: https://doc.rust-lang.org/std/primitive.u32.html#associatedconstant.MAX, but Noir does not. Users that need these limits need to hard code them themselves.

### Happy Case

Noir adds MIN, MAX, and BITS constants for each integer type to its stdlib.

### Project Impact

None

### Impact Context

_No response_

### Workaround

Yes

### Workaround Description

Users can define the constants themselves.

### Additional Context

_No response_

None

### Support Needs

_No response_

c410-f3r:
# Description

## Problem

Fix #4572

## Summary

Adds several global/constant variables associated with integers into the new `num` module. Feel free to point out if such a module is not desirable.

AFAICT, the lack of an explicit type declaration allows maximum flexibility  as long as the number literals can be inferred at the calling site.

```rust
// Allowed!
let x: u16 = std::num::U8_MAX;
```

## Documentation

Check one:

# PR Checklist


TomAFrench (changes requested):
I think rather than creating a new module for this it would be preferable to just add these as associated constants on the primitive types themselves.

```noir
fn main() {
    let x: u8 = u8::MAX;
    let y: u8 = u8::MIN;
    let z = x + y;
    assert (z == u8::MAX);
}
```

c410-f3r:
Can you confirm? The issue explicitly says that the constants should be in the standard library.

vezenovm:
> Can you confirm? The issue explicitly says that the constants should be in the standard library.

Yes they should be in the standard library, but we allow for impls to be written on primitive types in the stdlib. Here are some examples: https://github.com/noir-lang/noir/blob/619c5451b152d62e01d3c4c1da7e13ff6502f915/noir_stdlib/src/cmp.nr#L31 and https://github.com/noir-lang/noir/blob/619c5451b152d62e01d3c4c1da7e13ff6502f915/noir_stdlib/src/field/mod.nr#L5

c410-f3r:
> but we allow for impls to be written on primitive types

OK, then it is about creating a trait and implement it for integers. Should be done now.

AFAICT it is not possible to use types or associated types so `Field` is the only solution at the current time. `Field` also implies numbers up to `u32`.

```
Only unsigned integer constants up to `u32`, globals, generics, +, -, *, /, and % may be used in this context.
```

`run_stdlib_tests` is passing but despite my efforts, I couldn't create unit tests.

1. `<T as IntConstans>::MAX`:  not yet implemented: Implement AsTraitPath.
2. `u8::MAX`: noirc thinks that `MAX` is a function.

I kept the `num` module, feel free to say if this is undesirable or anything else.

TomAFrench:
I think that it would be best for us to just wait on this feature until we can have an associated constant applied directly to the integer type impls here rather than bringing in a trait and dealing with the issues associated with needing to case these `Fields` (and the necessary breaking change to fix later on).

c410-f3r:
Sure, I will wait.

PR #6178, opened 2024-09-30
feat: let `Module::functions` and `Module::structs` return them in definition order

asterite:
# Description

## Problem

The above functions would return items in different order depending on the run.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
@@ -2458,10 +2458,12 @@ fn module_functions(
     let module_id = get_module(self_argument)?;
     let module_data = interpreter.elaborator.get_module(module_id);
     let func_ids = module_data
-        .value_definitions()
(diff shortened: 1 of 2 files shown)


compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -2482,10 +2484,12 @@ fn module_structs(
     let module_id = get_module(self_argument)?;
     let module_data = interpreter.elaborator.get_module(module_id);
     let struct_ids = module_data
-        .type_definitions()
+        .definitions()
+        .definitions()
```
TomAFrench, line 2488: :thinking:

asterite: I know... but it's the only way I found to get the definitions as a `Vec`. (I think I don't understand how `ItemScope` yet)

TomAFrench (approved):

Issue #4469, opened 2024-03-04
Cannot use type aliases of structs in paths


jfecher:
### Aim

```rs
use std::{collections::map::HashMap, hash::{BuildHasherDefault, poseidon::PoseidonHasher}};

type Alias = HashMap<Field, Field, 10, BuildHasherDefault<PoseidonHasher>>;

fn main() {
    let map1 = HashMap::default();

    // panic here
    let map2 = Alias::default();
}
```

### Expected Behavior

To be able to use `Alias` anywhere `HashMap` is expected, including in paths

### Bug

```
The application panicked (crashed).
Message:  type aliases cannot be used in type namespace
Location: compiler/noirc_frontend/src/hir/resolution/import.rs:159

This is a bug. We may have already fixed this in newer versions of Nargo so try searching for similar issues at https://github.com/noir-lang/noir/issues/.
If there isn't an open issue for this bug, consider opening one at https://github.com/noir-lang/noir/issues/new?labels=bug&template=bug_report.yml
exit 101
```

### To Reproduce

1.
2.
3.
4.

### Project Impact

None

### Workaround

Yes

### Workaround Description

The original type (`HashMap` in this case) must be used instead of the alias

### Installation Method

None

### Would you like to submit a PR for this Issue?

None

asterite:
To be able to work on this we first need to implement #6000.

The reason is that we can solve the alias but it solves to a type that has concrete generics, and then we are calling a method on that... but we don't have code yet to support that, so we first need to implement that.

PR #6179, opened 2024-09-30
feat: visibility for impl functions

asterite:
# Description

## Problem

Part of #4515

## Summary

After searching for a while I found that to make this work we only needed to change one line 😊 

## Additional Context

I didn't include tests for now, but I'll add them later because we first need to figure out which `impl` functions should be marked as `pub` or `pub(crate)` in the stdlib. I'm not familiar with the stdlib though I probably can't do it. That said, if you open the standard library with a nargo compiled from this PR you already get some warnings, so those functions should at least not be private.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/mod.rs
+++ b/compiler/noirc_frontend/src/elaborator/mod.rs
@@ -1134,7 +1134,7 @@ impl<'context> Elaborator<'context> {
                 // object types in each method overlap or not. If they do, we issue an error.
                 // If not, that is specialization which is allowed.
                 let name = method.name_ident().clone();
-                if module.declare_function(name, ItemVisibility::Public, *method_id).is_err() {
+                if module.declare_function(name, method.def.visibility, *method_id).is_err() {
                     let existing = module.find_func_with_name(method.name_ident()).expect(
                         "declare_function should only error if there is an existing function",
                     );


TomAFrench:
`turbofish_numeric_generic_nested_call` is failing so it seems like private impl functions cannot be called from the same module (i.e. the functions are private to within the impl rather than to the module in which the impl sits)

asterite:
Ah, right. I just bumped into this while trying to add some visibilities. It seems that also `U128::method` fails if you are inside `U128`, but it doesn't fail if you do `Self::method` 🤔 I'll take a look 😄

asterite:
Done! All tests pass now. I think there are still some `fn`s that could be made `pub`.

asterite:
I think I got it now! 😄

asterite:
I noticed that visibility isn't checked for the new `TypePath`, which is something like `Field::from_le_bytes` where what's before `::` isn't a Path but a Type. That requires different code to do it so I'd prefer to do it in a separate PR.

TomAFrench:
Yeah, there's a good amount of stuff broken with type aliases due to this, a lot of issues related to type aliases stem from this https://github.com/noir-lang/noir/issues/4469

asterite:
I see.

Well, this is ready for review in case you want to take a look at it.

noir_stdlib/src/embedded_curve_ops.nr
```diff
@@ -67,14 +67,14 @@ impl EmbeddedCurveScalar {
...
-    fn from_field(scalar: Field) -> EmbeddedCurveScalar {
+    pub fn from_field(scalar: Field) -> EmbeddedCurveScalar {
         let (a,b) = crate::field::bn254::decompose(scalar);
         EmbeddedCurveScalar { lo: a, hi: b }
     }
 
     //Bytes to scalar: take the first (after the specified offset) 16 bytes of the input as the lo value, and the next 16 bytes as the hi value
     #[field(bn254)]
-    fn from_bytes(bytes: [u8; 64], offset: u32) -> EmbeddedCurveScalar {
+    pub fn from_bytes(bytes: [u8; 64], offset: u32) -> EmbeddedCurveScalar {
```
TomAFrench, line 77: Is this used outside of the stdlib? It's interface doesn't look very user facing.

TomAFrench (approved):
One concern about a function we're making public but otherwise looks fine.

noir_stdlib/src/embedded_curve_ops.nr
asterite: Apparently not. I'll undo this.

Issue #853, opened 2023-02-15
Remove chumsky fork


jfecher:
# Problem

We have been using a fork of chumsky for some time for the added parser `.recover_via` for custom recovery strategies. Chumsky 0.9.0 has now been released with the parser `skip_parser` which will do the same thing.

# Solution

We should abandon the fork, change our chumsky version to 0.9.0, and test with the new version. We should also test #791 again with the new version to see if there have been any performance improvements.

pczarn:
I started working on this issue.

kevaundray:
@jfecher wants the status on this issue? Would be good to get this closed once and for all (given the problem is tractable)

jfecher:
@kevaundray we're still using the fork. I suppose we can try to revisit using chumsky again, but last time there was an issue with the parser recovery from it.

jfecher:
According to https://github.com/zesterer/chumsky/releases it looks like the first alpha version for the zero-copy update of chumsky released March 2nd. Instead of migrating over to 0.9.0 now, I think we should wait until 1.0.0 is eventually finalized and released since it sounds like it'll come with many new features, performance improvements, and changes.

Issue #6150, opened 2024-09-25
Expression inside `lt` triggers Stack Overflow Error


MarkCampbell90:
### Aim

I want to executing following (already minimized) program:

~~~
pub fn main(input : Field) -> pub bool {
    input.lt((((((((((((((((((1))))))))))))))))))
}
~~~

If the expression is moved to a variable everything is fine:

~~~
pub fn main(input : Field) -> pub bool {
    let var : Field = (((((((((((((((((1)))))))))))))))));
    nput.lt(var)
}
~~~

I would be interested to know:
  * Is this a known limitation of the `lt` function?
  * Are there other limitations like this?

### Expected Behavior

I expected expressions to behave similar everywhere in code, no matter if present in an assignment or as an argument to `lt( ... )`. Additionally,  the expression depth limit seems very small in this particular case.

### Bug

~~~
thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow
Aborted
~~~

### To Reproduce

1. create a `noir` / `nargo` project with the given content
2. `nargo execute`
3.
4.

### Workaround

None

### Workaround Description

Assigning the expression to a variable before passing it to the `lt` function as argument.

### Project Impact

None

### Nargo Version

nargo version = 0.34.0 noirc version = 0.34.0+e3cdebe515e4dc4ee6e16e01bd8af25135939798 (git version hash: e3cdebe515e4dc4ee6e16e01bd8af25135939798, is dirty: false)

### Would you like to submit a PR for this Issue?

None

Issue #5092, opened 2024-05-23
Stack overflow in Chumsky


Rumata888:
### Aim

I was compiling a program with a lot of recursive loops to test how loop unrolling would be handled in SSA.

### Expected Behavior

No failure in frontend

### Bug
Stack overflow
![image](https://github.com/noir-lang/noir/assets/4798775/b726c388-b555-484e-baa0-6505a7c4dda8)

### To Reproduce

Compile the following file:
[main.txt](https://github.com/noir-lang/noir/files/15420958/main.txt)

### Project Impact

Nice-to-have

### Workaround

None

### Installation Method

None

### Nargo Version

nargo version = 0.29.0 noirc version = 0.29.0+e5ab24d6a4154d11b3c8ae08f4431b7b93c76f23 (git version hash: e5ab24d6a4154d11b3c8ae08f4431b7b93c76f23, is dirty: true)

### Would you like to submit a PR for this Issue?

None

Issue #4763, opened 2024-04-09
Parse block test fails with uninformative message


michaeljklein:
### Aim

The `parse_all_failing` tests in [`parse_block`](https://github.com/noir-lang/noir/blob/master/compiler/noirc_frontend/src/parser/parser.rs#L1403) include a test case with a message where the expected and found tokens are both empty, defaulting to `EOF`.

By modifying the following, it's possible to see that it
fails with an opaque error. (i.e. the 'expected' and 'found' are both empty)

```rust
// src/parser/errors.rs
impl chumsky::Error<Token> for ParserError {
    fn expected_input_found<Iter>(span: Self::Span, expected: Iter, found: Option<Token>) -> Self
```

### Expected Behavior

Return an error that explains what went wrong

### Bug

Error is appx. `expected token EOF while found EOF`. Inspecting the expected/found tokens shows both are empty. Likely due to 

### To Reproduce

Run the mentioned test and modify `expected_input_found` to panic when both `expected` and `found` are empty.

### Project Impact

Nice-to-have

### Workaround

None

### Installation Method

None

### Would you like to submit a PR for this Issue?

None

asterite:
Do you know what specific code triggers that error message? (that is, a full program that triggers that error message)

michaeljklein:
@asterite IIRC approximately:

```rust
let expected: Vec<_> = expected.iter().collect();
assert!(!(expected.empty() && found.is_none()));
let expected = expected.iter();
```

michaeljklein:
Ah, my link went out of date. These: https://github.com/noir-lang/noir/blob/bd861f282144056ecb52954fa9f6fd8db918e093/compiler/noirc_frontend/src/parser/parser.rs#L1438

asterite:
Ah, I see. I'll try to reproduce it. The problem is that parsing a block in isolation never happens in a real program, it always happens in the context of something else and so the error mentioned above might not happen in a real program.

asterite:
This can be closed as the old parser doesn't exist anymore :-)

PR #6180, opened 2024-09-30
feat: handwritten parser

asterite:
# Description

Resolves https://github.com/noir-lang/noir/issues/853

## Problem

There are some issues with the current parser:
- it leads to "stack overflow" with small programs that should probably compile (like a program with a chanin of 17 `if-else-if`)
- it leads to some of us not being able to run the noric_frontend tests because of linker errors, and sometimes making changes to the parser will lead to linker errors that we have to workaround
- it's (very) slow

## Summary

This PR implements a hand-written parser. It parses any program with just one look-ahead token. It has a decent error-recovery (though this can probably be improved).

I tested the parser's performance by copying the contents of `noir-contracts/contracts/avm_test_contract/src/main.nr` in Aztec-Packages 100 times to a single file. That ends up with a file of about 57K lines. The times:
- chumsky parser: 1.52 **seconds**
- handwritten parser: 52.97 **milliseconds**

Some other benefits:
- The linker errors are gone!
- Compiling noirc_frontend is slightly faster
- Macro code also becomes faster (`quote { ... }.as_expr()`, etc, invoke the parser). For example `test_programs/noir_test_success/comptime_expr/src/main.nr` takes around one second with chumsky and 140ms with the handwritten parser (to do `nargo compile`)
- Even though the parser is handwritten, I think the parsing code is relatively simple. It's just "check if we get this token, then do this" or sometimes "check if we get this token followed by this one (or not followed by this one). Also also the `impl`s and `traits` that we needed for chumsky (and the lifetimes, and passing parsers around, and cloning them, and calling `boxed()`, etc.) are gone, which I believe make the code much simpler.

I tested this parser by running `./boostrap.sh` on the Aztec-Packages contracts and they compile file (of course they didn't compile right away, I had to fix some bugs to get there).

Pending:
- [ ] Thoroughly test this new parser
- [ ] Improve error recovery (at least cover the cases that exist in the old parser)
- [ ] Decide if we really want a handwritten parser (I hope so! Most mainstream programming languages have handwritten parsers because of their simplicity and efficiency)

## Documentation

Check one:

# PR Checklist


asterite:
Some other things I noticed...

Here's the CI times for PRs other than this one:

image

That's between 4~6 minutes to run the tests.

For this PR:

image

It went down to 1 minute! I honestly can't explain why (I can't imagine the parser alone just taking 3 more minutes) but the time is definitely lower (and I checked that in previous CI runs of this PR that's also the case).

Then `nargo fmt`. For PRs other than this one:

image

For this PR:

image

So an additional benefit of the handwritten parser seems to be faster development cycles. Hopefully running CI on Aztec-Packages will be faster, but I don't know if significantly.

TomAFrench (changes requested):
These are some amazing speed improvements, very excited for this.

Blocking for now due to failures on aztec-packages

asterite:
@TomAFrench  Oh, I forgot that I had to make one change to Aztec-Packages to make it compile:

https://github.com/AztecProtocol/aztec-packages/blob/e0d5e06d8fc30cbdda7e4102dbf3412808382377/noir-projects/noir-protocol-circuits/crates/types/src/address/aztec_address.nr#L1-L2

The line is:

```noir
use crate::{
    crate::address::{partial_address::PartialAddress, public_keys_hash::PublicKeysHash},
```

Note how it does `use crate::{crate::address...}`. For some reason that works in Noir but it's forbidden in Rust. My guess is that it should be forbidden in Noir too.

Is it okay if I send a PR to remove that extra `crate::`? I can also make the parser work on that case... but maybe it should error?

TomAFrench:
Oh yeah, that's very weird and shouldn't have compiled. Let's make a quick PR in aztec-packages to fix that.

asterite:
CI is green! ✅ 🎉 

Well, it's still failing on `AztecProtocol/aztec-nr` but I think that's expected because it's missing the fix I sent to Aztec-Pacakges. But we can wait until tomorrow 😄

asterite:
One potential issue about writing a new parser is losing the existing parsing recovery.

I'm happy to say that the new parser has better error recovery than the old parser.

Here's the old parser:

recovery-old

Here's the new one:

recovery-new

Note that in the old one sometimes the type hints are gone, meaning that we couldn't recover and keep parsing from an erroneous token (though in some cases the old parser succeeds to continue).

asterite:
I pushed a few more commits to make the code a bit easier to read and follow, and also informally/formally documented the grammar of the language on top of methods for two reasons:
1. To help understand the code
2. For external contributors wanting to understand the grammar without needing to fully dive into the code (someone was asking for a grammar recently in Discord)

aakoshh:
Just for future reference about the linker error on Mac: one workaround we found was to install the LLMV linker and run the tests with some custom flags:

```shell
brew install lld
RUSTFLAGS="-C link-arg=-fuse-ld=lld" cargo test -p noirc_frontend
```

asterite:
More stats...

Compiling an empty program:
- before: 650ms
- after: 140ms

Compiling all contracts in Aztec-Packages (`./bootstrap.sh` but just the nargo compile call)
- before: 50s
- after: 43s

asterite:
I checked a few existing issues:


### #4763

I couldn't understand how to reproduce it

### #4322

The error message in the handwritten parser is better. For this code:

```noir
use foo::{bar baz};
```

the chumsky parser says:

```
error: Unexpected Keyword(Mod), expected one of comptime, fn
  ┌─ std/lib.nr:1:5
  │
1 │ pub mod hash;
  │     ---
  │

error: Expected a ; but found ::
  ┌─ src/main.nr:1:8
  │
1 │ use foo::{bar baz};
  │        --
  │

error: Expected a Ident but found {
  ┌─ src/main.nr:1:10
  │
1 │ use foo::{bar baz};
  │          -
```

The new parser says:

```
error: Expected a `,` separating these two use trees
  ┌─ src/main.nr:1:15
  │
1 │ use foo::{bar baz};
  │               ---
```

Actually, the full error message I get is this:

```
error: Could not resolve 'foo' in path
  ┌─ src/main.nr:1:5
  │
1 │ use foo::{bar baz};
  │     ---
  │

error: Could not resolve 'foo' in path
  ┌─ src/main.nr:1:5
  │
1 │ use foo::{bar baz};
  │     ---
  │

error: Expected a `,` separating these two use trees
  ┌─ src/main.nr:1:15
  │
1 │ use foo::{bar baz};
  │               ---
```

That's because the parser noticed there's a missing comma but continued parsing it as if there was one, producing subsequent errors if any (I think this is better, and that's what Rust does too)

### #1153

The error message is still poor in the new parser, but it could be easily improved.

### #6150

Stack overflow in the old parser, works fine in the new parser.

### #5092

Stack overflow in both parsers, though the program is huge and a bit unreal (340 nested `for` loops). That said Rust parses it just fine, despite them also having a handwritten recursive descent parser. I might try to make this work but not sure.

### #3525

Doesn't error in neither parser.

### #2355

Doesn't error in neither parser.

---

So I guess the conclusion is that the new parser still can't handle really big nested loops, but otherwise it performs better than the old parser in every other scenario.

asterite:
I edited the PR description with some other benefits, which I'll include here for more visibility :-)

- Tests run faster:
  - Running noirc_frontend tests:
    - before: 1:03 minute
    - after: 6 seconds
  - Running lsp tests: 
    - before: 55 seconds
    - after: 6 seconds
  - Running nargo_cli tests:
    - before: 2:47 minutes
    - after: 38 seconds
- CI runs faster (for example each of the four partitions take 1 minute instead of 4 minutes
- Building the compiler is faster:
  - before: 1:29 minutes
  - after: 1:19 minutes (so building noirc_frontend is 10 seconds faster because that's the only thing changed in this PR)

The above means that our developer experience (compiler developers) improves a lot: we can start running tests locally and not have to wait a minute or two, we get faster CI cycles, and we can test changes quicker (the frontend compiles faster). So please take those benefits into account when reviewing this PR.

tooling/nargo_fmt/tests/input/databus.nr
```diff
@@ -1,2 +1,2 @@
-fn main(x: pub u8, y: call_data u8) -> return_data u32 {}
+fn main(x: pub u8, y: call_data(0) u8) -> return_data u32 {}
```
michaeljklein, line 1: Can this be a new tests or do we not parse `y: call_data u8` any more?

Looking at this example from https://noir-lang.org/docs/noir/concepts/data_bus

asterite: Ah, right, I forgot to mention about this. I had to change it because that's invalid syntax, this is the error I get:

```
error: Expected a ( but found u8
  ┌─ src/main.nr:1:33
  │
1 │ fn main(x: pub u8, y: call_data u8) -> return_data u32 {}
  │                                 --
```

I also couldn't find that syntax in the chumsky parser.

The test was passing because it failed to format the file, and the expected output is the same as the input... but because nothing changes, it passes.

asterite: Or, put another way, it's currently invalid syntax. Maybe the intention was for it to be valid? In that case we could update the parser to support it (but maybe in a follow-up PR, given that the current parser doesn't handle it either).

TomAFrench: This should be invalid syntax, can you add some perturbation to the input file so that it needs to modify it in order to match the output?

asterite: Done!

compiler/noirc_frontend/src/parser/parser/generics.rs
```diff
@@ -0,0 +1,295 @@
...
+        None
+    }
+
+    /// VariableGeneric = identifier
+    fn parse_variable_generic(&mut self) -> Option<UnresolvedGeneric> {
+        if let Some(ident) = self.eat_ident() {
+            return Some(UnresolvedGeneric::Variable(ident));
+        }
+
+        None
```
jfecher, line 65:
```suggestion
        self.eat_ident().map(UnresolvedGeneric::Variable)
```

compiler/noirc_frontend/src/parser/parser/generics.rs
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    /// NumericGeneric = 'let' identifier ':' Type
+    fn parse_numeric_generic(&mut self) -> Option<UnresolvedGeneric> {
+        if !self.eat_keyword(Keyword::Let) {
+            return None;
+        }
```
jfecher, line 72: if we return an Option from eat_keyword then we could `?` here

compiler/noirc_frontend/src/parser/parser/generics.rs
```diff
@@ -0,0 +1,295 @@
...
+
+        let Some(ident) = self.eat_ident() else {
+            return None;
+        };
```
jfecher, line 76:
```suggestion
        let ident = self.eat_ident()?;
```

compiler/noirc_frontend/src/parser/parser/generics.rs
```diff
@@ -0,0 +1,295 @@
...
+        let typ = self.parse_type_or_error();
+        if let UnresolvedTypeData::Integer(signedness, bit_size) = &typ.typ {
+            if matches!(signedness, Signedness::Signed)
+                || matches!(bit_size, IntegerBitSize::SixtyFour)
+            {
+                self.push_error(ParserErrorReason::ForbiddenNumericGenericType, typ.span);
+            }
+        }
+
+        Some(UnresolvedGeneric::Numeric { ident, typ })
```
jfecher, line 95: We only allow integer types here (not sure if we also allow field or not) so we should probably error if this is not an integer? Although we already do this during name resolution IIRC.

compiler/noirc_frontend/src/parser/parser/generics.rs
```diff
@@ -0,0 +1,295 @@
...
+        if let Some(token) = self.eat_kind(TokenKind::QuotedType) {
+            match token.into_token() {
+                Token::QuotedType(id) => {
+                    return Some(UnresolvedGeneric::Resolved(id, self.previous_token_span));
+                }
+                _ => unreachable!(),
+            }
+        }
+
+        None
```
jfecher, line 109:
```suggestion
        let token = self.eat_kind(TokenKind::QuotedType)?;
        match token.into_token() {
            Token::QuotedType(id) => {
                Some(UnresolvedGeneric::Resolved(id, self.previous_token_span))
            }
            _ => unreachable!(),
        }
```

compiler/noirc_frontend/src/parser/parser/generics.rs
```diff
@@ -0,0 +1,295 @@
...
+    ///     | OrderedTypeArg
+    ///
+    /// NamedTypeArg = identifier '=' Type
+    ///
+    /// OrderedTypeArg = TypeOrTypeExpression
+    pub(super) fn parse_generic_type_args(&mut self) -> GenericTypeArgs {
+        let mut generic_type_args = GenericTypeArgs::default();
+        if !self.eat_less() {
+            return generic_type_args;
+        }
```
jfecher, line 127: More diet shaming 😒

compiler/noirc_frontend/src/parser/parser/generics.rs
```diff
@@ -0,0 +1,295 @@
...
+            let typ = self.parse_type_or_error();
+            return Some(GenericTypeArg::Named(ident, typ));
+        }
+
+        // Otherwise
+        let typ = self.parse_type_or_type_expression();
+        let Some(typ) = typ else {
+            self.expected_label(ParsingRuleLabel::TypeOrTypeExpression);
+            return None;
+        };
```
jfecher, line 164:
```suggestion
        let Some(typ) = self.parse_type_or_type_expression() else {
            self.expected_label(ParsingRuleLabel::TypeOrTypeExpression);
            return None;
        };
```

compiler/noirc_frontend/src/parser/parser/generics.rs
```diff
@@ -0,0 +1,295 @@
...
+        Some(GenericTypeArg::Ordered(typ))
+    }
+}
+
+fn type_u32() -> UnresolvedType {
+    UnresolvedType {
+        typ: UnresolvedTypeData::Integer(Signedness::Unsigned, IntegerBitSize::ThirtyTwo),
+        span: Span::default(),
+    }
+}
```
jfecher, line 175: We should avoid a default span here - a zero-width span if possible would be preferable. Maybe we could take the position as an input or make this require self.

compiler/noirc_frontend/src/parser/parser/modifiers.rs
```diff
@@ -0,0 +1,38 @@
...
+pub(crate) struct Modifiers {
+    pub(crate) visibility: ItemVisibility,
+    pub(crate) visibility_span: Span,
+    pub(crate) unconstrained: Option<Span>,
+    pub(crate) comptime: Option<Span>,
+    pub(crate) mutable: Option<Span>,
+}
+
+impl<'a> Parser<'a> {
+    /// Modifiers = 'unconstrained'? ItemVisibility 'comptime'? 'mut'?
```
jfecher, line 17: Sidenote: I still think `unconstrained` should be after `pub` but this would be a large breaking change.

compiler/noirc_frontend/src/parser/parser/modifiers.rs
```diff
@@ -0,0 +1,38 @@
...
+    pub(crate) fn parse_modifiers(&mut self, allow_mutable: bool) -> Modifiers {
+        let unconstrained = self.parse_modifier(Keyword::Unconstrained);
+
+        let start_span = self.current_token_span;
+        let visibility = self.parse_item_visibility();
+        let visibility_span = self.span_since(start_span);
+
+        let comptime = self.parse_modifier(Keyword::Comptime);
+        let mutable = if allow_mutable { self.parse_modifier(Keyword::Mut) } else { None };
```
jfecher, line 26: When do we allow `mut` in the same position we allow `unconstrained`, `pub`, or `comptime`?

compiler/noirc_frontend/src/parser/parser/modifiers.rs
```diff
@@ -0,0 +1,38 @@
...
+        Modifiers { visibility, visibility_span, unconstrained, comptime, mutable }
+    }
+
+    fn parse_modifier(&mut self, keyword: Keyword) -> Option<Span> {
+        if self.eat_keyword(keyword) {
+            Some(self.previous_token_span)
+        } else {
+            None
+        }
+    }
```
jfecher, line 37: Seems like this should be `accept_modifier` or similar? Other `parse_*` methods may error if the thing is not present so we should be a bit more consistent at least for base cases like this.

compiler/noirc_frontend/src/parser/parser/parse_many.rs
```diff
@@ -0,0 +1,108 @@
+use crate::token::Token;
+
+use super::Parser;
+
+impl<'a> Parser<'a> {
+    /// Parses a list of items separated by a token, optionally ending when another token is found.
+    /// The given function `f` must parse an item. If no item is parsed, `f` must report an error
+    /// and return `None`.
```
jfecher, line 8: "`f` must parse an item" reads to me like this expects to parse 1 or more items instead of 0 or more which I believe is the intention

compiler/noirc_frontend/src/parser/parser/expression.rs
```diff
@@ -0,0 +1,1708 @@
...
+                panic!("Expected infix for {operator}");
+            };
+            assert_eq!(infix_expr.lhs.to_string(), "1");
+            assert_eq!(infix_expr.operator.contents, operator);
+            assert_eq!(infix_expr.rhs.to_string(), "2");
+        }
+    }
+
+    #[test]
+    fn parses_operator_precedence() {
```
TomAFrench, line 1556: This is the only test that (directly) checks how the parser handles operator precedence that we have currently. Can you improve coverage here to ensure that we handle mixes of arithmetic, bitwise, equality, bitshift, modulo operators correctly?

compiler/noirc_frontend/src/parser/parser.rs
```diff
@@ -91,1764 +44,504 @@ use traits::where_clause;
...
-                ));
-                // into_ident panics on tuple or struct patterns but should be fine to call here
-                // since the `self` parser can only parse `self`, `mut self` or `&mut self`.
-                (other.into_ident(), param.typ)
-            }
+impl<'a> TokenStream<'a> {
+    fn next(&mut self) -> Option<SpannedTokenResult> {
+        match self {
+            TokenStream::Lexer(lexer) => lexer.next(),
+            TokenStream::Tokens(tokens) => tokens.0.pop().map(Ok),
```
jfecher, line 83: Wouldn't this pop in reverse order?

Edit: Ah, I see we `.reverse()` the tokens in `for_tokens`. Can you add a comment here that that is an invariant?

compiler/noirc_frontend/src/parser/parser.rs
```diff
@@ -91,1764 +44,504 @@ use traits::where_clause;
...
-                .delimited_by(just(Token::LeftBrace), just(Token::RightBrace))
-                .map(UseTreeKind::List);
-
-            prefix.then(tree).map(|(prefix, kind)| UseTree { prefix, kind })
+    fn new(tokens: TokenStream<'a>) -> Self {
+        let mut parser = Self {
+            errors: Vec::new(),
+            tokens,
+            token: SpannedToken::default(),
+            next_token: SpannedToken::default(),
```
jfecher, line 120: If this is the only case we needed SpannedToken::default and Token::default for, I'd rather we just manually created the filler tokens here to prevent the rest of the codebase from using default on these types.

compiler/noirc_frontend/src/parser/parser.rs
```diff
@@ -91,1764 +44,504 @@ use traits::where_clause;
...
-        rhs,
-        |lhs, rhs, span| match rhs {
-            UnaryRhs::Call((is_macro, args)) => {
-                Expression::call(lhs, is_macro.is_some(), args, span)
-            }
-            UnaryRhs::ArrayIndex(index) => Expression::index(lhs, index, span),
-            UnaryRhs::Cast(r#type) => Expression::cast(lhs, r#type, span),
-            UnaryRhs::MemberAccess(field) => {
-                Expression::member_access_or_method_call(lhs, field, span)
+    fn eat_self(&mut self) -> bool {
```
jfecher, line 228: 😲 🫢

jfecher: I do find it odd some `eat_` functions return a bool while other return Options. If it makes the call sites cleaner I think returning `Option<()>` would be nice even if IIRC we have to ignore a clippy warning for it.

compiler/noirc_frontend/src/parser/parser.rs
```diff
@@ -91,1764 +44,504 @@ use traits::where_clause;
...
-        for expr in parse_all(array_expr(expression()), valid) {
-            match expr_to_array(expr) {
-                ArrayLiteral::Standard(elements) => assert_eq!(elements.len(), 5),
-                ArrayLiteral::Repeated { length, .. } => {
-                    assert_eq!(length.kind, ExpressionKind::integer(5i128.into()));
-                }
+    fn eat_quote(&mut self) -> Option<Tokens> {
+        if matches!(self.token.token(), Token::Quote(..)) {
+            let token = std::mem::take(&mut self.token);
+            self.next_token();
```
jfecher, line 321: If `self.next_token()` returned the previous token we wouldn't need to `take`. Could be a bit confusing though that it could look like it was returning the next token. Maybe it'd be solved with a rename to `self.advance()` or similar but I'm just nitpicking now.

compiler/noirc_frontend/src/parser/parser/expression.rs
asterite: Sounds good! Do you know if there's an existing test for the chumsky parser?

compiler/noirc_frontend/src/parser/parser/expression.rs
```diff
@@ -0,0 +1,1708 @@
...
+    ///     | MemberAccessOrMethodCallExpression
+    ///     | CastExpression
+    ///     | IndexExpression
+    fn parse_unary_right(&mut self, mut atom: Expression, start_span: Span) -> (Expression, bool) {
+        let mut parsed;
+
+        (atom, parsed) = self.parse_call(atom, start_span);
+        if parsed {
+            return (atom, parsed);
+        }
```
jfecher, line 124: Why don't we return a `Option<Expression>` here?

compiler/noirc_frontend/src/parser/parser/expression.rs
asterite: Nevermind, I think I know how to make a simple and good test for this.

jfecher (commented):
Didn't get through everything but generally looks good. My main worry is the same as Tom's comment I saw where I'm not 100% sure if we're parsing expression precedence the same. That being said I'd hope our existing unit & integration tests would be enough to check for that but it's possible there are cases missing with some of the more rare operators.

I see we are manually creating each parsing rule for each separate precedence level in parser/infix.rs. I think something like shunting yard could be simpler but it's not a blocker. Noir should also be fairly stable now with relatively few new operators being added.

compiler/noirc_frontend/src/parser/parser/modifiers.rs
TomAFrench: I agree with this change. I was quite surprised when I tried to make a `pub unconstrained` fn a while back and saw this.

We shouldn't do this in this PR but we could allow both but format into the pub-first format. We can restrict it again in a raft of 1.0 breaking changes.

compiler/noirc_frontend/src/parser/parser/expression.rs
asterite: Done! Let me know if the test I added is good.

compiler/noirc_frontend/src/parser/parser/modifiers.rs
TomAFrench: `pub mut comptime global Foo = 0;` is the only location I think.

asterite:
> I see we are manually creating each parsing rule for each separate precedence level in parser/infix.rs. I think something like shunting yard could be simpler but it's not a blocker. Noir should also be fairly stable now with relatively few new operators being added.

Right... Initially it even had a lot of code duplication and then I reduced that a bit with the `parse_infix` helper.

I read about [pratt parsers](https://journal.stuffwithstuff.com/2011/03/19/pratt-parsers-expression-parsing-made-easy/) but I feel that it would lead to slower code... and then the final code isn't that big (less than 200 lines) and we'll likely won't add new operators, or if we do it should be rare, so we won't be modifying that file at all or a lot (like you said).

compiler/noirc_frontend/src/elaborator/comptime.rs
```diff
@@ -261,9 +260,14 @@ impl<'context> Elaborator<'context> {
...
-            .parse(tokens)
-            .map_err(|mut errors| (errors.swap_remove(0).into(), location.file))?;
+        let mut parser = Parser::for_tokens(tokens);
+        let expression = parser.parse_expression();
+        if !parser.errors.is_empty() {
+            return Ok(None);
+        }
+        let Some(expression) = expression else {
+            return Ok(None);
+        };
```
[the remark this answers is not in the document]
asterite: I initially tried using `Parser::parse_result` for this but it didn't work and I couldn't understand why. I'll take a look again.

compiler/noirc_frontend/src/lexer/token.rs
```diff
@@ -126,7 +126,7 @@ pub enum BorrowedToken<'input> {
     Invalid(char),
 }
 
-#[derive(PartialEq, Eq, Hash, Debug, Clone, PartialOrd, Ord)]
+#[derive(PartialEq, Eq, Hash, Debug, Clone, PartialOrd, Ord, Default)]
```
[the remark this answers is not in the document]
asterite: Ah, good point. I added this in the very beginning but ended up not needing it. I'll remove it 👍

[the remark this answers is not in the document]
asterite: Oh, it's because I use `std::mem::take` in one place with a `SpannedToken`... but I'll just create the "default" by hand, it'll be a bit more explicit and it's just in a few places.

compiler/noirc_frontend/src/parser/parser/arguments.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+impl<'a> Parser<'a> {
+    /// Arguments = '(' ArgumentsList? ')'
+    ///
+    /// ArgumentsList = Expression ( ',' Expression )? ','?
+    pub(crate) fn parse_arguments(&mut self) -> Option<Vec<Expression>> {
+        if !self.eat_left_paren() {
+            return None;
+        }
```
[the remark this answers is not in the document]
asterite: Good point. That's actually how I had it in the beginning (most `eat` methods would return `Option`) but in cases where something other than `None` was returned I had to do `self.eat_left_parent().is_none()` which is a bit more verbose... so in the end I decided to stick with bools.

compiler/noirc_frontend/src/parser/parser/arguments.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    /// CallArguments = '!'? Arguments
+    pub(super) fn parse_call_arguments(&mut self) -> Option<CallArguments> {
+        let is_macro_call = self.tokens_follow(Token::Bang, Token::LeftParen);
+
+        if is_macro_call {
+            // Given that we expected '!' '(', it's safe to skip the '!' because the next
+            // `self.parse_arguments()` will always return `Some`.
+            self.next_token();
+        }
```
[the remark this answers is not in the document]
asterite: Oh, the issue is that what if we find `!` but not `(`. Then we shouldn't parse that as a call (I don't know what we'd parse it into, it's probably a syntax error). That's why I check if both `!` and `(` come next and just then consume the macro call.

compiler/noirc_frontend/src/parser/parser/arguments.rs
```diff
@@ -0,0 +1,40 @@
...
+            separated_by_comma_until_right_paren(),
+            Self::parse_expression_in_list,
+        );
+
+        Some(arguments)
+    }
+
+    /// CallArguments = '!'? Arguments
+    pub(super) fn parse_call_arguments(&mut self) -> Option<CallArguments> {
+        let is_macro_call = self.tokens_follow(Token::Bang, Token::LeftParen);
```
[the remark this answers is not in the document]
asterite: I added `tokens_follow` to make the code a bit more shorter and easier to read. But... if it's making it less clear it's probably better to change it to how it was before. I think at that time I didn't have the `at` and `next_at` methods so doing `if self.at(...) && self.next_at(...)` will be a bit more clear, and not terribly longer.

compiler/noirc_frontend/src/parser/parser/generics.rs
asterite: I agree that in many cases we return `None` when a keyword isn't there. I'll start changing these to return `Option`.

compiler/noirc_frontend/src/parser/parser/doc_comments.rs
```diff
@@ -1,36 +1,66 @@
...
+use crate::token::{DocStyle, Token, TokenKind};
 
-pub(super) fn outer_doc_comments() -> impl NoirParser<Vec<String>> {
-    outer_doc_comment().repeated()
-}
+use super::Parser;
+
+impl<'a> Parser<'a> {
+    /// InnerDocComments = inner_doc_comment*
+    pub(super) fn parse_inner_doc_comments(&mut self) -> Vec<String> {
```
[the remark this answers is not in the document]
asterite: In the end there weren't that many uses of `many0` (I think) but I'll still try to crate a helper function for this. I wanted to avoid as many `loop` and `while` in the parser to avoid accidentally looping forever.

compiler/noirc_frontend/src/parser/parser/arguments.rs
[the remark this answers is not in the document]
asterite: Removed :-)

compiler/noirc_frontend/src/lexer/token.rs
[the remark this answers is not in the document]
asterite: Done! I also found a way to do things a bit better, so now `std::mem::take` is never used, and `std::mem::replace` is used instead (I think that should be slightly faster, though I didn't notice any improvement in the big file I'm trying it with).

compiler/noirc_frontend/src/parser/parser/arguments.rs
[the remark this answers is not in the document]
jfecher: Don't we already check for `(` in parse_arguments though? Seems redundant unless that only parses arguments and a closing parenthesis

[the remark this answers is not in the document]
asterite: I removed `tokens_follow`

compiler/noirc_frontend/src/parser/parser/doc_comments.rs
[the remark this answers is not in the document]
jfecher: After looking at `parse_many` it looks like you could add a helper for that with no separator

compiler/noirc_frontend/src/parser/parser/generics.rs
asterite: I'm applying these one by one because for a batch GitHub tells me to go to the files tab, and it's kind of hard to find these comments in that big diff 😅

compiler/noirc_frontend/src/parser/parser/modifiers.rs
asterite: I agree too! I was going to ask about it last week in our channel, but I also didn't want to change things in this PR.

We could probably allow it `unconstrained` before and after `pub` to begin with, erroring if it's duplicate, then eventually only allow it after `pub`.

compiler/noirc_frontend/src/parser/parser.rs
asterite: Agreed, removed :-)

compiler/noirc_frontend/src/parser/parser/expression.rs
asterite: Because if we couldn't parse a unary we have to keep the old expression. I thought of some other ways to do this, like returning a struct... I don't like the end result but I couldn't find a better way to do it.

compiler/noirc_frontend/src/parser/parser/generics.rs
asterite: I started making this change but the places where the return value is used as a bool are much more than those where we could early return with `?`, so I'm not sure it's worth it (it would make the code also read a bit worse, with `if self.eat_something().is_some()` instead of `if self.eat_something()`

compiler/noirc_frontend/src/parser/parser/arguments.rs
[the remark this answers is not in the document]
asterite: As mentioned in the other comment, this will simplify a few locations but increase the verbosity and reduce clarify of many other places.

compiler/noirc_frontend/src/parser/parser/generics.rs
asterite: Right, I thought so too. I initially errored on other types but some test failed so I removed that. It's also the same in the chumsky parser: https://github.com/noir-lang/noir/blob/8236cbdff60c1aaf41fc53142b6f0f9ea2fc2fa8/compiler/noirc_frontend/src/parser/parser/function.rs#L129-L143

compiler/noirc_frontend/src/parser/parser/generics.rs
asterite: Done!

compiler/noirc_frontend/src/parser/parser.rs
asterite: Like in the other comments, I'm not sure about this.

compiler/noirc_frontend/src/parser/parser/attributes.rs
```diff
@@ -1,77 +1,84 @@
...
-pub(super) fn validate_attributes(
-    attributes: Vec<Attribute>,
-    span: Span,
-    emit: &mut dyn FnMut(ParserError),
-) -> Attributes {
-    let mut primary = None;
-    let mut secondary = Vec::new();
+    /// Attributes = attribute*
+    pub(super) fn parse_attributes(&mut self) -> Vec<(Attribute, Span)> {
+        let mut attributes = Vec::new();
```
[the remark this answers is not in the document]
asterite: I just remembered that `parse_many` can be used without a separator, so I pushed a commit to replace these `while` loops with that. Thanks!

compiler/noirc_frontend/src/elaborator/comptime.rs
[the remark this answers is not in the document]
asterite: This reminded me that there's a lot of repetition in tests, so I'll try to reduce that too.

compiler/noirc_frontend/src/ast/expression.rs
```diff
@@ -308,7 +308,7 @@ impl Expression {
 
 pub type BinaryOp = Spanned<BinaryOpKind>;
 
-#[derive(PartialEq, PartialOrd, Eq, Ord, Hash, Debug, Copy, Clone)]
+#[derive(PartialEq, PartialOrd, Eq, Ord, Hash, Debug, Copy, Clone, strum_macros::EnumIter)]
```
TomAFrench, line 311:
```suggestion
#[derive(PartialEq, PartialOrd, Eq, Ord, Hash, Debug, Copy, Clone)]
#[cfg_attr(test, derive(strum_macros::EnumIter))]
```

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -243,7 +243,7 @@ impl std::fmt::Display for Kind {
     }
 }
 
-#[derive(Debug, PartialEq, Eq, Copy, Clone, Hash, PartialOrd, Ord)]
+#[derive(Debug, PartialEq, Eq, Copy, Clone, Hash, PartialOrd, Ord, strum_macros::EnumIter)]
```
TomAFrench, line 246:
```suggestion
#[derive(Debug, PartialEq, Eq, Copy, Clone, Hash, PartialOrd, Ord)]
#[cfg_attr(test, derive(strum_macros::EnumIter))]
```

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -675,15 +672,24 @@ fn quoted_as_expr(
...
+    if let Ok(expr) = result {
+        return option(return_type, Some(Value::expression(expr.kind)));
+    }
 
-    let expr = parse(interner, argument, parser, "an expression").ok();
+    let result =
+        parse(interner, argument.clone(), Parser::parse_statement_or_error, "an expression");
+    if let Ok(stmt) = result {
+        return option(return_type, Some(Value::statement(stmt.kind)));
+    }
```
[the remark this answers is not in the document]
asterite: This reminded me that we could have a `parse_statement_or_expression_or_lvalue` that returns either of those in one go. I think it's doable. That would mean no need to `argument.clone()` and it'll be slightly faster, which is important for macro code. But maybe this could be for a follow-up PR... for now I'll try to reduce the repetition.

compiler/noirc_frontend/src/elaborator/comptime.rs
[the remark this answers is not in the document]
asterite: I pushed a commit that simplifies this. Let me know if it's not okay.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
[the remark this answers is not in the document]
asterite: I couldn't find a quick way to reduce this repetition, but... it's the only repetition, and it's a small one. Once we have `parse_statement_or_expression_or_lvalue` it'll be gone 🤞

compiler/noirc_frontend/src/parser/parser/modifiers.rs
asterite: To solve this I inlined `parse_modifier` into callsites. The code ends up being 1 line longer, so that helper function wasn't brining much value.

TomAFrench (approved):

asterite:
Approved! 🤩 

Thank you!

I'll merge this... 🤞 

Worse case scenario we can revert it, or fix things if we find them 😄

Aristotelis2002:
Hello, we have noticed a significant speedup in running the complete test suite from the root of the project (with `cargo test -q`) after this change. This is great!

We were just wondering: could it be related to this PR? It's so fast that we are actually suspicious whether the tests are running or not :sweat_smile: ... but they seem to be.

asterite:
Hi! Indeed, it's related to this PR. I think `cargo test` on the root runs some of the tests, for example the ones on the nargo_cli project, and the times I got for this were:
- before this PR: 2:47 minutes
- after this PR: 38 seconds

PR #6197, opened 2024-10-01
fix(ssa): Check if result of array set is used in value of another array set

vezenovm:
# Description

## Problem\*

Resolves 

## Summary\*

Fix for failure in the `test_barycentric` of the `blob` lib in `noir-protocol-circuits`. The test now passes when Brillig's `MAX_STACK_SIZE` is increased to 32k. 

I included an additional map to track whether an array set result was used as a value in another array set. We can not allow for inner nested array sets to be mutable across blocks as the full nested array may be used in another block. We could add additional logic to check whether the parent array set will be marked mutable, but this seems overly complex for little benefit at the moment as this fix showed a quite small size regression.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/brillig/brillig_ir/entry_point.rs
+++ b/compiler/noirc_evaluator/src/brillig/brillig_ir/entry_point.rs
@@ -9,7 +9,7 @@ use super::{
 };
 use acvm::acir::{brillig::MemoryAddress, AcirField};
 
-pub(crate) const MAX_STACK_SIZE: usize = 32768;
+pub(crate) const MAX_STACK_SIZE: usize = 2048;
 pub(crate) const MAX_SCRATCH_SPACE: usize = 64;
 
 impl<F: AcirField + DebugToString> BrilligContext<F, Stack> {
--- a/compiler/noirc_evaluator/src/brillig/brillig_ir/entry_point.rs
+++ b/compiler/noirc_evaluator/src/brillig/brillig_ir/entry_point.rs
@@ -9,7 +9,7 @@ use super::{
 };
 use acvm::acir::{brillig::MemoryAddress, AcirField};
 
-pub(crate) const MAX_STACK_SIZE: usize = 2048;
+pub(crate) const MAX_STACK_SIZE: usize = 32768;
 pub(crate) const MAX_SCRATCH_SPACE: usize = 64;
 
 impl<F: AcirField + DebugToString> BrilligContext<F, Stack> {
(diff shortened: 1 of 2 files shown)


compiler/noirc_evaluator/src/brillig/brillig_ir/entry_point.rs
```diff
@@ -9,7 +9,7 @@ use super::{
 };
 use acvm::acir::{brillig::MemoryAddress, AcirField};
 
-pub(crate) const MAX_STACK_SIZE: usize = 2048;
+pub(crate) const MAX_STACK_SIZE: usize = 32768;
```
vezenovm, line 12:
```suggestion
pub(crate) const MAX_STACK_SIZE: usize = 2048;
```

compiler/noirc_evaluator/src/ssa/opt/array_set.rs
```diff
@@ -219,42 +239,47 @@ mod tests {
...
-        let v14 = builder.insert_binary(v5, BinaryOp::Add, one);
-        builder.terminate_with_jmp(b1, vec![v14]);
+        let v17 = builder.insert_binary(v5, BinaryOp::Add, one);
+        builder.terminate_with_jmp(b1, vec![v17]);
 
         builder.switch_to_block(b2);
         builder.terminate_with_return(vec![]);
 
         let ssa = builder.finish();
+        println!("{}", ssa);
```
vezenovm, line 279:
```suggestion
```

compiler/noirc_evaluator/src/ssa/opt/array_set.rs
```diff
@@ -219,42 +239,46 @@ mod tests {
...
+        let v17 = builder.insert_binary(v5, BinaryOp::Add, one);
+        builder.terminate_with_jmp(b1, vec![v17]);
 
         builder.switch_to_block(b2);
         builder.terminate_with_return(vec![]);
 
         let ssa = builder.finish();
         // We expect the same result as above
         let ssa = ssa.array_set_optimization();
+        println!("{}", ssa);
```
vezenovm, line 281:
```suggestion
```

compiler/noirc_evaluator/src/ssa/opt/array_set.rs
```diff
@@ -55,9 +58,11 @@ impl Function {
 fn analyze_last_uses(
     dfg: &DataFlowGraph,
     block_id: BasicBlockId,
+    is_brillig_func: bool,
     array_to_last_use: &mut HashMap<ValueId, InstructionId>,
     instructions_that_can_be_made_mutable: &mut HashSet<InstructionId>,
     arrays_from_load: &mut HashSet<ValueId>,
+    inner_nested_arrays: &mut HashMap<ValueId, InstructionId>,
```
vezenovm, line 65: This method is getting quite long. I am going to make a context struct in a follow-up refactor.

TomAFrench (approved):
LGTM

PR #6198, opened 2024-10-01
feat: remove orphaned blocks from cfg to improve `simplify_cfg` pass.

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

Inspired by https://github.com/noir-lang/noir/pull/6184, I noticed that after simplifying a constant jmpif we wouldn't inline the block following the inlined block into the first block, as an example consider the SSA

```
After Dead Instruction Elimination:
brillig fn main f0 {
  b0(v0: Field, v1: Field):
    constrain v0 == Field 1
    constrain v1 == Field 0
    jmpif u1 0 then: b2, else: b1
  b2():
    v8 = add v0, Field 1
    jmp b3(v0, v8)
  b3(v2: Field, v3: Field):
    v9 = add v0, Field 1
    constrain v2 == v9
    constrain v3 == v0
    constrain v0 == Field 1
    constrain v1 == Field 0
    return 
  b1():
    v7 = add v0, Field 1
    jmp b3(v7, v0)
}
```
Here we can see that we never enter `b2`, instead going `b0 -> b1 -> b3`. However when the simplify_cfg pass looks at the predecessors of `b3` it sees both `b1` **and `b2`** despite `b2` not being reachable anymore. It then doesn't inline `b3` into its sole predecessor.

On master we'll simplify the CFG for the SSA to give the program:
```
After Simplifying:
brillig fn main f0 {
  b0(v0: Field, v1: Field):
    constrain v0 == Field 1
    constrain v1 == Field 0
    v6 = add v0, Field 1
    jmp b1(v6, v0)
  b1(v2: Field, v3: Field):
    v7 = add v0, Field 1
    constrain v2 == v7
    constrain v3 == v0
    constrain v0 == Field 1
    constrain v1 == Field 0
    return 
}
```

This PR checks the previous successors of any block which has it's CFG recomputed to check if they've been left orphaned, if so we remove it from the CFG entirely.

We now instead fully simplify the CFG to give the program:
```
After Simplifying:
brillig fn main f0 {
  b0(v0: Field, v1: Field):
    constrain v0 == Field 1
    constrain v1 == Field 0
    v4 = add v0, Field 1
    v5 = add v0, Field 1
    constrain v4 == v5
    constrain v0 == Field 1
    constrain v1 == Field 0
    return 
}

```

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/simplify_cfg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/simplify_cfg.rs
@@ -106,7 +106,7 @@ fn check_for_constant_jmpif(
             let call_stack = call_stack.clone();
             let jmp = TerminatorInstruction::Jmp { destination, arguments, call_stack };
             function.dfg[block].set_terminator(jmp);
-            cfg.recompute_block(function, block);
+            cfg.recompute_block(function, block, true);
         }
     }
 }
@@ -171,9 +171,9 @@ fn check_for_double_jmp(function: &mut Function, block: BasicBlockId, cfg: &mut 
         };
 
         function.dfg[predecessor_block].set_terminator(redirected_terminator_instruction);
-        cfg.recompute_block(function, predecessor_block);
+        cfg.recompute_block(function, predecessor_block, false);
     }
-    cfg.recompute_block(function, block);
+    cfg.recompute_block(function, block, false);
 }
 
 /// If the given block has block parameters, replace them with the jump arguments from the predecessor.
@@ -221,8 +221,8 @@ fn try_inline_into_predecessor(
         drop(successors);
         function.dfg.inline_block(block, predecessor);
 
-        cfg.recompute_block(function, block);
-        cfg.recompute_block(function, predecessor);
+        cfg.recompute_block(function, block, false);
+        cfg.recompute_block(function, predecessor, false);
         true
     } else {
         false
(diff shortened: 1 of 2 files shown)


TomAFrench:
We can maybe pull this extra logic out of `recompute_cfg` as we only need to run it when simplifying constant jmpifs rather than every time we recompute the CFG for a block. I just left it here as it maintained the same interface.

TomAFrench:
Seems like the above is required. Marking draft so I can do some cleanup tomorrow.

compiler/noirc_evaluator/src/ssa/opt/simplify_cfg.rs
```diff
@@ -99,14 +98,23 @@ fn check_for_constant_jmpif(
...
+                (*then_destination, *else_destination)
+            };
 
             let arguments = Vec::new();
             let call_stack = call_stack.clone();
             let jmp = TerminatorInstruction::Jmp { destination, arguments, call_stack };
             function.dfg[block].set_terminator(jmp);
             cfg.recompute_block(function, block);
+
+            // If `block` was the only predecessor to `unchose_destination` then it's no long reachable through the CFG,
```
TomAFrench, line 113:
```suggestion
            // If `block` was the only predecessor to `unchosen_destination` then it's no long reachable through the CFG,
```

vezenovm (approved):

PR #6200, opened 2024-10-02
feat: Inclusive for loop

### LINKED ISSUE #6171: Inclusive loops (e.g. `0..=length`)
### Problem

Inclusive ranges are especially useful if you want to iterate over every possible value in a range. Currently the iteration in Noir is like in Rust 1.0 through 1.25 versions. If you want inclusion of the last element, you need to add one to your end value like:
```
fn main() {
    for i in 0..(26 + 1) {
        println!("{}", i);
    }
}
```

### Happy Case

In Rust >=1.26 versions, you can now create an inclusive range and it will be very beneficial to add this syntactic sugar to Noir as well.
Example:
```
fn main() {
    for i in 0..=26 {
        println!("{}", i);
    }
}
```

### Project Impact

Nice-to-have

None

aakoshh:
# Description

## Problem\*

Resolves #6171

## Summary\*

Adds support for inclusive for syntax, e.g. `for i in 0..=10`.

## Additional Context

- Added a `ForBounds` struct to handle the desugaring of `start..=end` into `start..end+1`.
- `ForBounds` has an `inclusive` flag; I thought about adding a `ForRange::RangeInclusive` similar to `std::ops::RangeInclusive` in Rust, but then I thought `for` loops will not have the other variants like `..end`, `start..` and `..`, which wouldn't have two separate expressions, and the flag is simpler. 
- Transforming an inclusive range to Hir and back makes it non-inclusive, but I think that's fine because the block created by `ForRange::into_for` to iterate over the array is not turned back either.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/ast/statement.rs
```diff
@@ -770,13 +771,60 @@ impl LValue {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct ForBounds(
+    /*start:*/ pub Expression,
+    /*end:*/ pub Expression,
+    /*inclusive:*/ pub bool,
```
asterite, line 778: What do you think about making these named fields?

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -1514,7 +1514,8 @@ fn expr_as_for_range(
 ) -> IResult<Value> {
     expr_as(interner, arguments, return_type, location, |expr| {
         if let ExprValue::Statement(StatementKind::For(for_statement)) = expr {
-            if let ForRange::Range(from, to) = for_statement.range {
+            if let ForRange::Range(bounds) = for_statement.range {
+                let (from, to) = bounds.into_half_open();
```
asterite, line 1518: We might have to change this built-in macro method to also return a `bool` for whether the range is inclusive or not, but we can do it in a separate PR.

compiler/noirc_frontend/src/lexer/lexer.rs
```diff
@@ -704,7 +720,7 @@ mod tests {
 
     #[test]
     fn test_single_double_char() {
-        let input = "! != + ( ) { } [ ] | , ; : :: < <= > >= & - -> . .. % / * = == << >>";
+        let input = "! != + ( ) { } [ ] | , ; : :: < <= > >= & - -> . .. ..= % / * = == << >>";
```
asterite, line 723: The name of this test now slightly doesn't match because of this, but it's not a big deal (feel free to choose a different name if you want, or leave it as is)

compiler/noirc_frontend/src/ast/statement.rs
aakoshh: I think now that we have constructor methods for them it would be a good change; where they are used in pattern matching the syntax would be very similar, just `{ start, end, inclusive }` instead of `(from, to, inclusive)` 👍

compiler/noirc_frontend/src/hir/comptime/tests.rs
```diff
@@ -160,6 +160,19 @@ fn for_loop() {
     assert_eq!(result, Value::U8(15));
 }
 
+#[test]
+fn for_loop_inclusive() {
```
asterite, line 164: Could you also add a test in test_programs/execution_success that proves that the new loop bounds work? It can be like this program but with an `assert_eq(...)` at the end. Mainly because the interpreter and the actual execution have different code flows (I'm sure the test will pass, but just to prevent regressions).

asterite (commented):
Looks great! I left a couple of comments.

compiler/noirc_frontend/src/hir/comptime/tests.rs
aakoshh: This is great, thanks for pointing me at these tests 👀 

I added a test to `brillig_loop` in 6eee91b3f8a39cbce19e4a880ba83dd7998b3476
Couldn't find any other test about looping specifically that isn't using the `unconstrainted` keyword, which I assume has something to do with Brillig. 
Do you think this is enough, or should I create a separate test case?

compiler/noirc_frontend/src/ast/statement.rs
aakoshh: Renamed to `test_single_multi_char`.

compiler/noirc_frontend/src/hir/comptime/tests.rs
aakoshh: Added another to be sure.

asterite: I think putting it in a separate test is good. Modifying the existing brilling one might be fine too but we'd have to check with the rest of the team: we have some code that checks the number of opcodes produced by program, and by modifying that one it increased (as expected) but I don't know if those programs should not be modified. You could ask Tom or Maxim about this.

aakoshh: I would assume that the reports are there to show any unintended increase due to changes in the compiler. I just wasn't sure whether Brillig does things differently in a way that leaves the normal constrained code paths untested. Having both can't hurt :)

asterite (approved):
Looks great!

PR #6206, opened 2024-10-02
fix: ignore compression of blocks after msg.len in sha256_var

### LINKED ISSUE #6163: sha256_var appears to produce different hash for each block
### Aim

Tried to use sha256_var to produce a hash on a message padded by more than 64 bytes

### Expected Behavior

std::hash::sha256::sha256_var() should produce the same hash for a (message: [u8; N], length: u64) regardless of N

### Bug

each new sha256 block changes the hash value. it does not appear that the values set after the length have any effect on the outputted hash, just the max size of the inputted message array

### To Reproduce

https://gist.github.com/jp4g/d5953faae9eadb2909357474f7901e58

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Blocker

### Blocker Context

_No response_

### Nargo Version

0.34.0+359caafac5e489901d9ff02b08d1a688178d9b0a

### NoirJS Version

_No response_

bb 0.55.0 (not used though)

None

### Support Needs

_No response_

saleel:
# Description

Fix an issue where `sha256_var` produces wrong results for messages with larger paddings

## Problem\*

Resolves #6163 

## Summary\*

`h = sha256_compression(msg_u8_to_u32(msg_block), h)` was run for bytes after length as well. This block is moved to `if msg_start < message_size` block

## Documentation\*

Check one:

# PR Checklist\*


noir_stdlib/src/hash/sha256.nr
```diff
@@ -110,8 +110,10 @@ pub fn sha256_var<let N: u32>(msg: [u8; N], message_size: u64) -> [u8; 32] {
 
         // If the block is filled, compress it.
         // An un-filled block is handled after this loop.
-        if msg_byte_ptr == BLOCK_SIZE {
-            h = sha256_compression(msg_u8_to_u32(msg_block), h);
+        if msg_start < message_size {
+            if msg_byte_ptr == BLOCK_SIZE {
```
vezenovm, line 114:
```suggestion
        if msg_start < message_size & msg_byte_ptr == BLOCK_SIZE {
```
Should be able to collapse this if statement

saleel: Ah yea, had error with this, but just had to add brackets for each condition
Updated now

vezenovm (approved):
Can you also add the test from https://gist.github.com/jp4g/d5953faae9eadb2909357474f7901e58 to our stdlib? If that passes then LGTM.

vezenovm:
> Nice, thank you. I ran those tests with my local Noir build and works well. Let me add it the repo.

I see you added it to `execution_success`. Let's also replicate it in the stdlib Noir tests in the sha256 module where the inputs will be all compile time.

PR #6208, opened 2024-10-02
feat(perf): Follow array sets backwards in array set from get optimization

vezenovm:
# Description

## Problem\*

Part of general effort reduce Brillig bytecode sizes

## Summary\*

Follow-up to https://github.com/noir-lang/noir/pull/6207

From comments inside the PR:
```
/// If we have an array set whose value is from an array get on the same array at the same index,
/// we can simplify that array set to the array we were looking to perform an array set upon.
///
/// Simple case:
/// v3 = array_get v1, index v2
/// v5 = array_set v1, index v2, value v3
///
/// If we could not immediately simplify the array set from its value, we can try to follow
/// the array set backwards in the case we have constant indices:
///
/// v3 = array_get v1, index 1
/// v5 = array_set v1, index 2, value [Field 100, Field 101, Field 102]
/// v7 = array_set mut v5, index 1, value v3
///
/// We want to optimize `v7` to `v5`. We see that `v3` comes from an array get to `v1`. We follow `v5` backwards and see an array set
/// to `v1` and see that the previous array set occurs to a different constant index.
///
/// For each array set:
/// - If the index is non-constant we fail the optimization since any index may be changed.
/// - If the index is constant and is our target index, we conservatively fail the optimization.
/// - Otherwise, we check the array value of the `array_set``. We will refer to this array as array'.
///   In the case above, array' is `v1` from `v5 = array set ...`
///   - If the original `array_set` value comes from an `array_get`, check the array in that `array_get` against array'. 
///   - If the two values are equal we can simplify.
///     - Continuing the example above, as we have `v3 = array_get v1, index 1`, `v1` is
///       what we want to check against array'. We now know we can simplify `v7` to `v5` as it is unchanged.
///   - If they are not equal, recur marking the current `array_set` array as the new array id to use in the checks
```

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/brillig/brillig_ir/entry_point.rs
```diff
@@ -9,7 +9,7 @@ use super::{
 };
 use acvm::acir::{brillig::MemoryAddress, AcirField};
 
-pub(crate) const MAX_STACK_SIZE: usize = 32768;
+pub(crate) const MAX_STACK_SIZE: usize = 2048;
```
vezenovm, line 12: Accidentally committed this stack size increase when testing against the blob lib.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -818,25 +818,69 @@ fn try_optimize_array_get_from_previous_set(
...
+
+// If we could not immediately simplify the array set from its value, we can try to follow
+// the array set backwards:
+
+// v3 = array_get v1, index v2
+// v5 = array_set v1, index v4, value [Field 100, Field 101, Field 102]
+// v7 = array_set mut v5, index v2, value v3
+
+// We want to optimize `v7` to `v5`. We see that `v3` comes from an array get to `v1`. We follow `v5` backwards and see an array set
+// to `v1` and that array set occurs to a different index. We now know we can simplify `v7` to `v5` as it is unchanged.
```
TomAFrench, line 836: We can't make this optimization safely as we don't have a guarantee that `v2` is different from `v4`. If those are both user inputs then we could have `v2 == v4` and so the final result would be `v1` and returning `v5` would be incorrect.

vezenovm: Ah yes 🤦. I will change this optimization to work with just constants and see what we get.

TomAFrench:
Note that `run-external-checks` only runs `nargo check` on aztec packages so it'll only find issues which appear in the frontend. It won't pick up any SSA errors so you'll need to check that manually (or we can compile the contracts)

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -818,25 +818,87 @@ fn try_optimize_array_get_from_previous_set(
...
+                        }
+
+                        if *array == array_from_get {
+                            return SimplifyResult::SimplifiedTo(original_array_id);
+                        }
+
+                        array_id = *array; // recur
+                    } else {
+                        return SimplifyResult::None;
+                    }
```
TomAFrench, line 893:
```suggestion
                    let Some(index) = dfg.get_numeric_constant(*index) else  {
                        return SimplifyResult::None;
                    }    
                    
                    if index == target_index {
                        return SimplifyResult::None;
                    }

                    if *array == array_from_get {
                        return SimplifyResult::SimplifiedTo(original_array_id);
                    }

                    array_id = *array; // recur
```

We could also remove a couple of layers from the match statements in a similar fashion tbh.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -818,25 +818,87 @@ fn try_optimize_array_get_from_previous_set(
...
+                    return SimplifyResult::None;
                 }
             }
-            _ => SimplifyResult::None,
+            _ => return SimplifyResult::None,
         },
-        _ => SimplifyResult::None,
+        _ => return SimplifyResult::None,
+    };
+
```
TomAFrench, line 869:
```suggestion
        // At this point we have determined that the value we are writing in the `array_set` instruction
        // comes from an `array_get` from the same index at which we want to write it at.
        // It's possible that we're acting on the same array where other indices have been mutated in between
        // the `array_get` and `array_set` (resulting in the `array_id` not matching).
        //
        // We then inspect the set of `array_set`s which which led to the current array the `array_set` is acting on.
        // If we can work back to the array on which the `array_get` was reading from without having another `array_set`
        // act on the same index then we can be sure that the new `array_set` can be removed without affecting the final result.
```

TomAFrench (approved):
I think it would be good to have some more documentation within the function about our reasoning about what we're doing at particular steps.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -818,25 +818,87 @@ fn try_optimize_array_get_from_previous_set(
...
+///
+/// We want to optimize `v7` to `v5`. We see that `v3` comes from an array get to `v1`. We follow `v5` backwards and see an array set
+/// to `v1` and see that the previous array set occurs to a different constant index.
+///
+/// For each array set:
+/// - If the index is non-constant we fail the optimization since any index may be changed.
+/// - If the index is constant and is our target index, we conservatively fail the optimization.
+/// - Otherwise, we check the array value of the array set. We will refer to this array as array'.
+///   In the case above, array' is `v1` from `v5 = array set ...`
+///   - Check the original array set value's array get against array'. If they are equal we can simplify.
```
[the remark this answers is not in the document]
vezenovm: Yeah I felt this sentence was a bit of a mess 😅  Let me know if that is more clear.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -818,25 +818,96 @@ fn try_optimize_array_get_from_previous_set(
...
+/// For each array set:
+/// - If the index is non-constant we fail the optimization since any index may be changed.
+/// - If the index is constant and is our target index, we conservatively fail the optimization.
+/// - Otherwise, we check the array value of the `array_set``. We will refer to this array as array'.
+///   In the case above, array' is `v1` from `v5 = array set ...`
+///   - If the original `array_set` value comes from an `array_get`, check the array in that `array_get` against array'.
+///   - If the two values are equal we can simplify.
+///     - Continuing the example above, as we have `v3 = array_get v1, index 1`, `v1` is
+///       what we want to check against array'. We now know we can simplify `v7` to `v5` as it is unchanged.
+///   - If they are not equal, recur marking the current `array_set` array as the new array id to use in the checks
```
TomAFrench, line 847:
```suggestion
/// For each array_set:
/// - If the index is non-constant we fail the optimization since any index may be changed.
/// - If the index is constant and is our target index, we conservatively fail the optimization.
/// - Otherwise, we check the array value of the `array_set`. We will refer to this array as array'.
///   In the case above, array' is `v1` from `v5 = array set ...`
///   - If the original `array_set` value comes from an `array_get`, check the array in that `array_get` against array'.
///   - If the two values are equal we can simplify.
///     - Continuing the example above, as we have `v3 = array_get v1, index 1`, `v1` is
///       what we want to check against array'. We now know we can simplify `v7` to `v5` as it is unchanged.
///   - If they are not equal, recur marking the current `array_set` array as the new array id to use in the checks
```

Issue #1258, opened 2023-04-30
Thin down stdlib to the minimum necessary and move code into external libraries.


TomAFrench:
### Problem

With various grant projects we've set the final step as "acceptance into the Noir standard library", this is problematic for a couple of reasons:

- It signals a transfer of "ownership" over the code from the grantee to the Noir language team.
  - These Noir developers are the best placed people to support, maintain and extend these libraries so we don't want to give the impression that we're "taking away" their code. They wrote it so they should be able to keep their name on it.
  - Similarly we can't / don't want to force a grantee to maintain a library if they don't want to, but a new maintainer's barrier to entry when forking an external library is much lower than making PRs to the stdlib.
- It prevents any breaking changes to these libraries without a breaking change to the standard library (and then to Noir itself) @kevaundray shared [this discussion thread](https://users.rust-lang.org/t/rust-should-have-a-big-standard-library-and-heres-why/37449/31) on a fat vs thin rust standard library.

### Proposed solution

We should restrict the usage of the stdlib to only be the most common of functions and functions which are required to be in the stdlib (e.g. black box function definitions). Everything else should be spun out into libraries which have their own separate versioning.

- `merkle.nr` can be spun out into `noir-lang/merkle` (I've pushed an initial version of this to https://github.com/TomAFrench/noir-merkle which can be transferred to the org)
- `sha256.nr`, `sha512.nr`, `hash/poseidon.nr` should be broken out into a `noir-lang/hashes` monorepo similar to https://github.com/RustCrypto/hashes.
- `ec.nr` can spun out into `noir-lang/elliptic-curves`.

Pushing out these functions from the stdlib forces us to deal with remote dependencies much more than when everything is in the stdlib. This is a good opportunity for dogfooding and allows us to strengthen our integration tests.

### Alternatives considered

We can retain these functions in the standard library and have it grow over time (continuously?). This will likely be a backwards compatibility nightmare in future.

### Submission Checklist

- [X] Once I hit submit, I will assign this issue to the Project Board with the appropriate tags.

kevaundray:
>This is complicated a little by some of the hash implementations in the stdlib being alternatives for black box functions, e.g. sha256. These fallback implementations would have to stay inside the standard library imo.

I think we should figure out a better way to do this because:

- We have fallback mechanisms in noir and in acvm (Not the end of the world)
- The fallback mechanism makes it such that we cannot compile a Noir program without a backend (since we need to know what exactly should have fallbacks and replce them if we can)
[...]

PR #6214, opened 2024-10-03
feat: expose sha256 helper functions

saleel:
# Description

Expose sha256 internal methods

## Problem\*

Expose these functions so they could be used for [partial-sha256](https://github.com/zkemail/zkemail.nr/blob/main/lib/src/partial_hash.nr) implementation (by Mach34/zkemail.nr). Currently these methods are copy/pasted.

## Summary\*

Make below functions public:
- verify_msg_block
- build_msg_block_iter
- msg_u8_to_u32
- pad_msg_block
- attach_len_to_msg_block
- hash_final_block

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/noir_stdlib/src/hash/sha256.nr
+++ b/noir_stdlib/src/hash/sha256.nr
@@ -21,7 +21,7 @@ pub fn digest<let N: u32>(msg: [u8; N]) -> [u8; 32] {
 }
 
 // Convert 64-byte array to array of 16 u32s
-fn msg_u8_to_u32(msg: [u8; 64]) -> [u32; 16] {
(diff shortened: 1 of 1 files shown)


TomAFrench (changes requested):
Internal methods are internal on purpose for reasons set out in https://github.com/noir-lang/noir/issues/1258

The more stuff we expose through the stdlib, the harder it is to make non breaking changes. Copy-pasting this code is fine.

saleel:
> The more stuff we expose through the stdlib, the harder it is to make non breaking changes.

True. Closing.

PR #6220, opened 2024-10-03
feat(perf): Simplify array get when most recent array set writes to the same index

vezenovm:
# Description

## Problem\*

## Summary\*

This takes inspiration from https://github.com/noir-lang/noir/pull/6207 that the value is still known at a previous array set when the index is dynamic. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/instruction.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/instruction.rs
@@ -761,7 +761,7 @@ impl Instruction {
 /// Simple case:
 /// v4 = array_set v1, index v2, value v3
 /// v5 = array_get v4, index v2
-/// 
+///
 /// If we could not immediately simplify the array get, we can try to follow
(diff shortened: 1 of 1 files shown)


michaeljklein (approved):

TomAFrench (changes requested):
I don't think this takes into account `enable_side_effects`. Can you add a test for this?

```
fn main(v1: [Field; 3], v2: u32) {
   enable_side_effects u1 0
   v4 = array_set v1, index v2, value v3
   enable_side_effects u1 1
   v5 = array_get v4, index v2
   return v5
}
```

vezenovm:
> I don't think this takes into account `enable_side_effects`

Ah yes good catch. I'm not sure if we can have this optimization here then as we need more information than just the array instructions. Going to convert this back to a draft for now.

vezenovm:
Closing in favor of https://github.com/noir-lang/noir/pull/6684 which takes side effects into account

PR #6221, opened 2024-10-03
feat: visibility for struct fields

### LINKED ISSUE #4837: Implement visibility for fields
### Problem

The standard library includes several types such as `BoundedVec` and `HashMap` which are only meant to be interacted with via their APIs. Allowing users to access and mutate fields directly can lead to them being in an invalid state.

### Happy Case

We should add visibility for fields (private, public, pub(crate)) to control access to them.

A private field would also make the constructor of a type automatically private.

### Project Impact

None

### Impact Context

_No response_

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #4837

Part of #4515

## Summary

This checks struct visibility in member access, constructors and patterns.

## Additional Context

Private fields are still suggested in LSP. I plan to hide those fields in a follow-up PR.

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/hir/resolution/import.rs
```diff
@@ -437,12 +437,12 @@ pub fn can_reference_module_id(
     // Note that if the target module is in a different crate from the current module then we will either
     // return true as the target module is public or return false as it is private without looking at the `CrateDefMap` in either case.
     let same_crate = target_module.krate == importing_crate;
-    let target_crate_def_map = &def_maps[&target_module.krate];
 
     match visibility {
         ItemVisibility::Public => true,
         ItemVisibility::PublicCrate => same_crate,
         ItemVisibility::Private => {
+            let target_crate_def_map = &def_maps[&target_module.krate];
```
asterite, line 445: Moving this is unrelated to this PR, but accessing this hashmap is only needed here so I moved it now before I forget (I don't know if Rust optimizes this out, maybe not because it can't know if that call has side-effects, so less hash accesses, better compiler performance).

compiler/noirc_frontend/src/elaborator/expressions.rs
```diff
@@ -618,11 +619,21 @@ impl<'context> Elaborator<'context> {
                 });
             }
 
-            if let Some(expected_index) = expected_index {
+            if let Some((index, visibility)) = expected_index_and_visibility {
+                let struct_type = struct_type.borrow();
+                let field_span = field_name.span();
+                let field_name = &field_name.0.contents;
+                self.check_struct_field_visibiilty(
```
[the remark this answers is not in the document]
asterite: Thanks! I turned off the spellcheck warnings while I was developing the new parser and forgot to turn it on. I don't know who made spellcheck infos to go into the editor, but it's great!

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -556,7 +557,11 @@ fn struct_def_set_fields(
 
                 match name_tokens.first() {
                     Some(Token::Ident(name)) if name_tokens.len() == 1 => {
-                        Ok((Ident::new(name.clone(), field_location.span), typ))
+                        Ok(hir_def::types::StructField {
+                            visibility: ItemVisibility::Private,
```
michaeljklein, line 561: Why not default to public members or else accept visibility parameters for each field, e.g. `[(Quoted, Type)] -> [(Quoted, Type, ItemVisibility)]` or similar?

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -487,9 +488,9 @@ fn struct_def_fields(
 
     let mut fields = im::Vector::new();
```
michaeljklein, line 490: Should `struct_def_fields` / `StructDefinition::fields` also return the visibilities?

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -388,28 +399,46 @@ impl StructType {
...
             },
         )
     }
 
     /// Returns all the fields of this type, after being applied to the given generic arguments.
+    pub fn get_fields_with_visibility(
+        &self,
+        generic_args: &[Type],
+    ) -> Vec<(String, ItemVisibility, Type)> {
+        let substitutions = self.get_fields_subtitutions(generic_args);
```
michaeljklein, line 412:
```suggestion
        let substitutions = self.get_fields_substitutions(generic_args);
```
And a couple other places

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -388,28 +399,46 @@ impl StructType {
...
+
+        vecmap(&self.fields, |field| {
+            let name = field.name.0.contents.clone();
+            (name, field.visibility, field.typ.substitute(&substitutions))
+        })
+    }
+
     pub fn get_fields(&self, generic_args: &[Type]) -> Vec<(String, Type)> {
+        let substitutions = self.get_fields_subtitutions(generic_args);
```
michaeljklein, line 421:
```suggestion
        let substitutions = self.get_fields_substitutions(generic_args);
```

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -388,28 +399,46 @@ impl StructType {
...
+
+        vecmap(&self.fields, |field| {
+            let name = field.name.0.contents.clone();
+            (name, field.typ.substitute(&substitutions))
+        })
+    }
+
+    fn get_fields_subtitutions(
```
michaeljklein, line 429:
```suggestion
    fn get_fields_substitutions(
```

compiler/noirc_frontend/src/tests/visibility.rs
```diff
@@ -146,3 +146,110 @@ fn does_not_error_if_calling_private_struct_function_from_same_module() {
...
+    )) = &errors[0].0
+    else {
+        panic!("Expected a private error");
+    };
+
+    assert_eq!(ident.to_string(), "x");
+}
+
+#[test]
+fn error_when_using_private_struct_field_in_struct_pattern() {
```
michaeljklein, line 227: Please add a test for the case(s) where the field is `pub x`

docs/docs/noir/concepts/data_types/structs.md
```diff
@@ -79,4 +81,16 @@ pub struct Animal {
...
+```
+
+The same applies to struct fields: by default they are private to the module they exist in,
+but they can be made `pub` or `pub(crate)`:
+
+```rust
+// This struct is now public
+pub struct Animal {
+    hands: Field,           // private to its module
+    pub(crate) legs: Field, // accessible from the entire crate
```
michaeljklein, line 93: Please also add a test or two for the `pub(crate)` case(s). (It may need to be a `test_programs` test.)

michaeljklein (changes requested):

compiler/noirc_frontend/src/tests/visibility.rs
asterite: Sounds good. That said, it's kind of tested already in the standard library because if that doesn't work the stdlib produces warnings. But I'll add a test 👍

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
asterite: Ah, regarding these built-in macro methods, I didn't want to change them because it would be a breaking change and it'll break macros on the Aztec side. But maybe that's fine? We'd have to think of an API to define the visibility, and possibly also return the visibility when asking for struct fields. But maybe it's better to do that change in a separate PR? What do you think?

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
asterite: Same as above: I didn't want to change the macros API in this PR.

docs/docs/noir/concepts/data_types/structs.md
asterite: Actually... it can't be done because these are warnings. Or we'd need a new category like `compile_success_no_warnings`. Should I do that?

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
asterite: I guess for now we could default to public to avoid this macro usage from introducing warnings. I'll do that :-)

michaeljklein: >  it would be a breaking change and it'll break macros on the Aztec side.

To avoid breakages, we could define a separate function `set_fields_with_visibility`
(And `StructDefinition::fields_with_visibility`)

> But maybe it's better to do that change in a separate PR? What do you think?

Sure, please add an issue for that

docs/docs/noir/concepts/data_types/structs.md
michaeljklein: Will the warnings become errors in a later PR? If so, then please make a note to follow up (e.g. in the issue to make them errors). Otherwise, this seems important enough to add `compile_success_no_warnings`, but let's save that for another PR.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
asterite: Done: #6229

compiler/noirc_frontend/src/tests/visibility.rs
```diff
@@ -146,3 +146,146 @@ fn does_not_error_if_calling_private_struct_function_from_same_module() {
     "#;
     assert_no_errors(src);
 }
+
+#[test]
+fn error_when_accessing_private_struct_field() {
```
michaeljklein, line 151: Are we covering the cases where the field is more public than the struct itself, i.e.
```noir
struct Foo {
    pub x: Field,
}
```
or
```noir
pub(crate) struct Foo {
    pub x: Field,
}
```

michaeljklein (changes requested):

compiler/noirc_frontend/src/tests/visibility.rs
asterite: Ah, good point. I didn't include those checks in this PR because we are missing those in other places. For example for this code:

```rust
struct Foo {}

pub fn bar() -> Foo {
    Foo {}
}
```

Rust says that `Foo` is more private than `bar`.

Would it be okay to capture an issue for all of these and fix them in a separate PR?

michaeljklein: > Would it be okay to capture an issue for all of these and fix them in a separate PR?

Yes, definitely 👍

asterite: Done! https://github.com/noir-lang/noir/issues/6248

compiler/noirc_frontend/src/tests/visibility.rs
```diff
@@ -146,3 +146,146 @@ fn does_not_error_if_calling_private_struct_function_from_same_module() {
...
+
+    let CompilationError::ResolverError(ResolverError::PathResolutionError(
+        PathResolutionError::Private(ident),
+    )) = &errors[0].0
+    else {
+        panic!("Expected a private error");
+    };
+
+    assert_eq!(ident.to_string(), "x");
+}
```
TomAFrench, line 244: I think the warning for this case is a bit suboptimal as we don't flag up that using a constructor for a struct with private fields is disallowed entirely. I think this can have an issue created and we address this later.

compiler/noirc_frontend/src/tests/visibility.rs
```diff
@@ -146,3 +146,146 @@ fn does_not_error_if_calling_private_struct_function_from_same_module() {
...
+        }
+    }
+
+    fn foo(foo: moo::Foo) -> Field {
+        let moo::Foo { x } = foo;
+        x
+    }
+
+    fn main() {
+    }
```
TomAFrench, line 261: Ditto here, it would be nice to have an error on the full destructuring pattern saying that this is not valid in this location due to private fields.

TomAFrench (approved):

michaeljklein (approved):

compiler/noirc_frontend/src/tests/visibility.rs
asterite: Sounds good, I'll capture an issue. Note that at least in Rust the error is on `x` saying it's private, though we could of course improve the error message in our case.

asterite: Captured in #6255

PR #6233, opened 2024-10-07
feat(test): Fuzz test stdlib hash functions

### LINKED ISSUE #6141: Add fuzz tests to show equivalence between stdlib cryptographic functions and rust implementations
We should add some fuzz tests which generates random inputs and feeds them into the stdlib implementations of `keccak256`/`sha256`/etc. alongside an off-the-shelf rust implementation to give greater assurances about correctness.

Note that this won't work with just a regular fuzz test but will need a rust harness.

aakoshh:
# Description

## Problem\*

Resolves #6141

## Summary\*

Adds property based testing for the following hash function in the standard library:
* Keccak256
* Sha256
* Sha512

In a followup I will carry on with Poseidon and Poseidon 2, and Schnorr signatures. I also wanted to run the code with the [interpreter](https://github.com/noir-lang/noir/blob/master/compiler/noirc_frontend/src/hir/comptime/tests.rs) but felt like it would complicate things in this PR.

## Additional Context

```console
❯ cargo test -p nargo_cli --test stdlib-props
    Finished test [optimized + debuginfo] target(s) in 0.23s
     Running tests/stdlib-props.rs (target/debug/deps/stdlib_props-cbab5917b839ebf4)

running 4 tests
test test_basic ... ok
test test_sha256 ... ok
test test_keccak256 ... ok
test test_sha512 ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 20.25s
```

## Documentation\*

Check one:

# PR Checklist\*


tooling/nargo_cli/Cargo.toml
```diff
@@ -75,9 +81,13 @@ fm.workspace = true
 criterion.workspace = true
 pprof.workspace = true
 paste = "1.0.14"
+proptest.workspace = true
+sha2 = "0.10"
+sha3 = "0.10"
```
TomAFrench, line 86: We depend on these crates in the crate `blackbox_solver` so it would be good to make these workspace deps.

tooling/nargo_cli/Cargo.toml
```diff
@@ -67,6 +72,7 @@ tokio-util = { version = "0.7.8", features = ["compat"] }
 
 [dev-dependencies]
 tempfile.workspace = true
+digest = "0.10"
```
TomAFrench, line 75: Is this used? We can import this trait through the sha2/sha3 crate.

tooling/nargo_cli/tests/stdlib-props.rs
```diff
@@ -0,0 +1,263 @@
...
+                    std::hash::keccak256(input, message_size)
+                }}"
+            )
+        },
+        |data| sha3::Keccak256::digest(data).try_into().unwrap(),
+    );
+}
+
+#[test]
+fn test_sha256() {
```
TomAFrench, line 226: I think this test naming can be improved, can we rename these to `fuzz_sha256_equivalence`, etc.?

TomAFrench (approved):
Looks good, some small nits.

TomAFrench:
I've pushed a small change which increases the number of lengths we run the sha256 test for to cover a recent issue we ran into.

tooling/nargo_cli/Cargo.toml
aakoshh: Okay, I wasn't sure if it was deliberately hidden away in the blackbox solver. Moved into the workspace.

tooling/nargo_cli/tests/stdlib-props.rs
aakoshh: Renamed

Issue #6143, opened 2024-09-24
`derive`d traits are not found by `get_trait_impl`


Thunkar:
### Aim

Given: 

```noir
mod traits; // Serialize, Deserialize and associated derive_via fns
mod meta; // Where "storage" macro lives

mod find_serialize {
    use std::meta::derive;
    use crate::traits::{Serialize, Deserialize};
    use crate::meta::storage;

    #[derive(Serialize, Deserialize, Eq)]
    struct State {
        token0: Field,
        token1: Field,
        liquidity_token: Field,
        reserve0: u64,
        reserve1: u64,
    }

    impl State {
        fn create() -> State {
            State { token0: 0, token1: 1, liquidity_token: 2, reserve0: 3, reserve1: 4 }
        }
    }

    #[storage]
    struct Storage {
        state: State,
    }

    fn main() {
        let state = State::create();
        let serialized = state.serialize();
        let deserialized = State::deserialize(serialized);

        assert(state == deserialized);
    }
}

```

If I have a storage annotation like this:

```noir
use std::meta::typ::fresh_type_variable;

pub comptime fn storage(s: StructDefinition) -> Quoted {
    println("STORAGE");
    let (field_name, stored_type) = s.fields()[0];

    let any = fresh_type_variable();
    let maybe_serialize_impl = stored_type.get_trait_impl(quote { crate::traits::Serialize<$any> }.as_trait_constraint());
    assert(
        maybe_serialize_impl.is_some(), f"Attempted to fetch serialization length, but {stored_type} does not implement the Serialize trait"
    );
    let serialize_impl = maybe_serialize_impl.unwrap();
    let size = serialize_impl.trait_generic_args()[0].as_constant().unwrap();
    // Hoping for 5
    println(f"SERIALIZED_SIZE {size}");
    let name = s.name();
    quote {
        impl $name {
            fn init() -> Self {
                Self {
                    $field_name: $stored_type::create()
                }
            }
        }
    }
}

```

It should be possible to find the derived Serialize implementation

### Expected Behavior

The above should compile

### Bug

Fails with 

> error: Attempted to fetch serialization length, but State does not implement the Serialize trait
   ┌─ src/meta/mod.nr:10:9
   │
10 │         maybe_serialize_impl.is_some(), f"Attempted to fetch serialization length, but {stored_type} does not implement the Serialize trait"
   │         ------------------------------ Assertion failed
   │
   = Call stack:
     1. src/main.nr:24:7

### To Reproduce

[...]

PR #6234, opened 2024-10-07
fix!: Elaborate generated items as we go

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6143

## Summary\*

Elaborate items generated at comptime as they're generated instead of all at once later on. This is a breaking change since any code e.g. outputting functions from attributes which rely on types that are outputted later will now error and the ordering will need to be changes.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/comptime.rs
+++ b/compiler/noirc_frontend/src/elaborator/comptime.rs
@@ -138,16 +138,9 @@ impl<'context> Elaborator<'context> {
         item: Value,
         span: Span,
         attribute_context: AttributeContext,
-        generated_items: &mut CollectedItems,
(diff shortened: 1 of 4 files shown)


jfecher:
```
 error: No matching impl found for `MyOtherOtherStruct<B>: Default`
   ┌─ execution_success/derive/src/main.nr:26:3
   │
26 │ #[derive(Eq, Default, Hash, Ord)]
   │   ------------------------------ No impl for `MyOtherOtherStruct<B>: Default`
   │
```
This may be a no-go as it breaks `derive` when the wrapper type is derived before all the field types are

PR #6235, opened 2024-10-07
fix!: Change tag attributes to require a ' prefix

jfecher:
# Description

## Summary\*

Changes tag attributes (previously Custom attributes) to require a `'` prefix. This distinguishes them from the more common Meta attributes which are expected to call a metaprogramming function.

The program:

```rs
#[not_in_scope]
fn main() {}
```
Now errors with:
```
error: Attribute function `not_in_scope` is not in scope
  ┌─ src/main.nr:1:3
  │
1 │ #[not_in_scope]
  │   ------------
  │
```

## Additional Context

- It turns out the `#[deprecated = ...]` attribute was not being handled previously. I'm not sure if this was expected but it's now easier to see now that we get an error if it does not get a `'` prefix.
- This also pointed out that `derive` was not working within the stdlib in the `meta/ops.nr` file. `derive` was never imported there. Moreover, after being used we get errors since derive tries to insert `std::cmp::Ordering` paths which are invalid within the stdlib where they should be `crate::cmp::Ordering` instead. So I replaced these calls with explicit impls.

## Documentation\*

Check one:
  - We may possibly need to document Tag attributes, but I'm unsure who the user would be.

# PR Checklist\*


michaeljklein (approved):

asterite:
> It turns out the #[deprecated = ...] attribute was not being handled previously. I'm not sure if this was expected but it's now easier to see now that we get an error if it does not get a ' prefix.

I bumped into this with the new parser (well, it let me know of a bug I had back then). I think that `deprecated = ...` syntax is not valid because it doesn't trigger any deprecation warning. If you do `#[deprecated]` or `#[deprecated("...")]` then it does trigger the warning (though is no case it uses the string you pass to it: it seems it's completely ignored).

I wonder if we should change those `deprecated = ...` attributes to their correct form.

jfecher:
I changed each `#['deprecated = ...]` usage to `#[deprecated]`. With `#[deprecated(...)]` I was still getting a "deprecated function not found" error message so I assume those weren't being counted as deprecated either and just removed the message.

jfecher:
Ok, all attributes in tests should be fixed now.

Pinging @vezenovm for visibility on `#[inline(never)]` not doing anything in the inline_never test

PR #6237, opened 2024-10-07
Optimize `Quoted::as_expr` by parsing just once

asterite:
# Description

## Problem

Follow-up to the parser PR to avoid calling the parser three times in `Quoted::as_expr()`

## Summary

This PR replaces `parse_lvalue_or_error`, which was only used in `Quoted::as_expr()`, with `parse_statement_or_expression_or_lvalue`, which is what `Quoted::as_expr()` needs. This avoids cloning the token stream, and also calls the parser just once.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
@@ -32,10 +32,12 @@ use crate::{
         def_collector::dc_crate::CollectedItems,
         def_map::ModuleDefId,
     },
-    hir_def::expr::{HirExpression, HirLiteral},
-    hir_def::function::FunctionBody,
+    hir_def::{
+        expr::{HirExpression, HirLiteral},
+        function::FunctionBody,
+    },
     node_interner::{DefinitionKind, NodeInterner, TraitImplKind},
-    parser::Parser,
+    parser::{Parser, StatementOrExpressionOrLValue},
     token::{Attribute, SecondaryAttribute, Token},
     Kind, QuotedType, ResolvedGeneric, Shared, Type, TypeVariable,
 };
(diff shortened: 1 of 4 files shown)


compiler/noirc_frontend/src/parser/parser/statement_or_expression_or_lvalue.rs
```diff
@@ -0,0 +1,55 @@
...
+    /// Parses either a statement, an expression or an LValue. Returns `StatementKind::Error`
+    /// if none can be parsed, recording an error if so.
+    ///
+    /// This method is only used in `Quoted::as_expr`.
+    pub(crate) fn parse_statement_or_expression_or_lvalue(
+        &mut self,
+    ) -> StatementOrExpressionOrLValue {
+        let start_span = self.current_token_span;
+
+        // First check if it's an interned LVAlue
```
vezenovm, line 25:
```suggestion
        // First check if it's an interned LValue
```
nit

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -684,21 +686,20 @@ fn quoted_as_expr(
...
-    }
-
-    let result =
-        parse(interner, argument.clone(), Parser::parse_statement_or_error, "an expression");
-    if let Ok(stmt) = result {
-        return option(return_type, Some(Value::statement(stmt.kind)));
-    }
+    let result = parse(
+        interner,
+        argument.clone(),
```
vezenovm, line 691:
```suggestion
        argument,
```
doesn't look like you need the clone here

vezenovm (approved):

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -684,21 +686,20 @@ fn quoted_as_expr(
...
-    let result = parse(interner, argument, Parser::parse_lvalue_or_error, "an expression");
-    if let Ok(lvalue) = result {
-        return option(return_type, Some(Value::lvalue(lvalue)));
+    if let Ok(statement_or_expression_or_lvalue) = result {
+        let value = match statement_or_expression_or_lvalue {
+            StatementOrExpressionOrLValue::Expression(expr) => Value::expression(expr.kind),
+            StatementOrExpressionOrLValue::Statement(statement) => Value::statement(statement.kind),
+            StatementOrExpressionOrLValue::LValue(lvalue) => Value::lvalue(lvalue),
+        };
+        return option(return_type, Some(value));
```
jfecher, line 702: Could use `result.ok().map(...)` so that we don't need to early-return and only need the final call to `option`

asterite: Good idea, done!

jfecher (approved):

PR #6239, opened 2024-10-08
fix(frontend): Do not warn when a nested struct is provided as input to main

vezenovm:
# Description

## Problem\*

No issue as found while working on other tasks.

Compile `nested_array_dynamic` and you will see a warning the `Bar` is never constructed. However, `Bar` is an internal field to the `Foo` struct which is provided as inputs and is thus constructed.

## Summary\*

When marking a parameter type as used we also go over all of the struct fields to account for any nested structs.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/mod.rs
+++ b/compiler/noirc_frontend/src/elaborator/mod.rs
@@ -846,6 +846,9 @@ impl<'context> Elaborator<'context> {
                 for generic in generics {
                     self.mark_parameter_type_as_used(generic);
                 }
+                for (_, typ) in struct_type.borrow().get_fields(generics) {
+                    self.mark_parameter_type_as_used(&typ);
+                }
             }
             Type::Alias(alias_type, generics) => {
                 self.mark_parameter_type_as_used(&alias_type.borrow().get_type(generics));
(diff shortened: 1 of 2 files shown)


compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -872,15 +875,6 @@ impl<'context> Elaborator<'context> {
             | Type::Forall(..)
             | Type::Error => (),
         }
-
-        if let Type::Alias(alias_type, generics) = typ {
-            self.mark_parameter_type_as_used(&alias_type.borrow().get_type(generics));
-            return;
-        }
-
-        if let Type::Struct(struct_type, _generics) = typ {
```
vezenovm, line 881: Deleted both of these `if let` statements as they are redundant to the match statement above.

compiler/noirc_frontend/src/tests/unused_items.rs
```diff
@@ -210,3 +210,25 @@ fn warns_on_unused_global() {
...
+        bar: Bar,
+    }
+
+    fn main(foos: [Foo; 1]) {
+        assert_eq(foos[0].a, 10);
+    }
+    "#;
+
+    let errors = get_program_errors(src);
+    println!("{errors:?}");
```
vezenovm, line 232:
```suggestion
```

jfecher (approved):
👍

Issue #6242, opened 2024-10-08
Arithmetic generics assumes integer multiplication cancels integer division


jfecher:
### Aim

```noir
fn main() {
    oops::<6, 4>([1, 2, 3, 4, 5, 6]);

    // Expected type [Field; 6], found type [Field; 4]
    oops::<6, 4>([1, 2, 3, 4]);
}

// This gets erroneously simplified to `_x: [Field; N]` internally
fn oops<let N: u32, let M: u32>(_x: [Field; N / M * M]) {}
```

### Expected Behavior

An error on the first call to `oops` and no error on the second.

### Bug

We get no error on the first call, and instead get an error on the second call despite `6 / 4 * 4 = 4` in integer arithmetic.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #6243, opened 2024-10-08
fix!: Integer division is not the inverse of integer multiplication

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6242

## Summary\*

Making this PR for visibility to show that fixing this issue breaks our serialize test.

Notably this fix allows rounding arithmetic generics using the `/ N * N` pattern which was simplified away previously.

## Additional Context

I initially removed the cancellation of / and * entirely but found that this breaks even more tests. So I added `approx_inverse` in a few cases that only involve constants that I thought may be correct but we still fail `serialize` even with these cases.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
```diff
@@ -125,7 +125,7 @@ impl Type {
 
         // Note that this is exact, syntactic equality, not unification.
         // `rhs` is expected to already be in canonical form.
-        if l_op.inverse() != Some(op) || l_rhs.canonicalize() != *rhs {
+        if l_op.approx_inverse() != Some(op) || l_rhs.canonicalize() != *rhs {
```
michaeljklein, line 128: Should we emit a warning in the cases where `approx_inverse` is used?

michaeljklein (changes requested):
Question about warning r.e. `approx_inverse`, otherwise LGTM

jfecher:
> Question about warning r.e. approx_inverse, otherwise LGTM

Even with `approx_inverse` we're still failing to compile the serialize test. It seems that we rely on this simplification that this PR removes so I'm not sure on a path forward.

michaeljklein (approved):
LGTM

compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
jfecher: No, I don't think a warning is needed. We should limit uses of approx_inverse to cases where we know it is correct, e.g. when we're checking if something divides evenly

PR #6246, opened 2024-10-08
feat: allow `unconstrained` after visibility

asterite:
# Description

## Problem

We currently parse `unconstrained pub fn ...` but ideally we'd want it to be `pub unconstrained fn ...`

See this comment too: https://github.com/noir-lang/noir/pull/6180#discussion_r1790221742

## Summary

This PR implements the above, but actually all of these:
- We now allow `unconstrained pub fn ...` and `pub unconstrained fn` (the former for backwards compatibility)
- The formatter will change `unconstrained pub fn` to `pub unconstrained fn` (so after some time we can remove the backwards-compatibility case)
- The formatter will now format a function "header" (I didn't know what to call it): that's the fn outer comments, attributes, and modifiers. For example before this PR if you wrote `pub   fn   foo() {}` it was left untouched, but now it's changed to `pub fn foo() {}`.

## Additional Context

None.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/nargo_fmt/src/visitor/item.rs
+++ b/tooling/nargo_fmt/src/visitor/item.rs
@@ -184,7 +184,7 @@ impl super::FmtVisitor<'_> {
         result
     }
 
-    fn append_comments_if_any<'a>(&self, parser: &mut Parser<'a>, result: &mut String) {
+    fn append_comments_if_any(&self, parser: &mut Parser, result: &mut String) {
         while let Some(token) = parser.eat_kind(TokenKind::Comment) {
             match token.token() {
                 Token::LineComment(..) => {
--- a/tooling/nargo_fmt/src/visitor/item.rs
+++ b/tooling/nargo_fmt/src/visitor/item.rs
@@ -136,9 +136,12 @@ impl super::FmtVisitor<'_> {
 
         self.append_comments_if_any(&mut parser, &mut result);
 
+        let mut is_unconstrained = false;
+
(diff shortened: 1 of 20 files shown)


tooling/nargo_fmt/src/visitor/item.rs
```diff
@@ -109,6 +112,94 @@ impl super::FmtVisitor<'_> {
...
+
+        // First there might be outer doc comments
+        while let Some(token) = parser.eat_kind(TokenKind::OuterDocComment) {
+            result.push_str(&token.to_string());
+            result.push('\n');
+            result.push_str(&self.indent.to_string());
+        }
+
+        // Then, optionally, attributes
+        while let Some((attribute, _span)) = parser.parse_attribute() {
```
jfecher, line 129: Why are we using the parser in the formatter here, can we not achieve the same with the lexer alone?
What happens if the formatter receives invalid input here, does it remove it?

asterite: I tried to use the lexer in the past (a couple of weeks ago). It's slightly trickier because if we get `pub` we have to check for `(`... but I think I bumped into other issues when I tried it. I'll try again now :-)

If there's invalid input then the formatter doesn't run.

asterite: I managed to do it :-)

jfecher (approved):

PR #6252, opened 2024-10-08
feat: trait inheritance

### LINKED ISSUE #3949: Support Trait Inheritance
### Problem

We'd like to be able to define a trait as a combination of other traits.

### Happy Case

The following syntax should compile:
```rust
trait Backend: SmartContract + ProofSystemCompiler + Default + Debug {}
```

### Alternatives Considered

It can be worked around by creating a blanket implementation for the trait `Backend` for all types that implement `SmartContract + ProofSystemCompiler + Default + Debug ` but it's not ideal.

Another workaround is to have each function with a `T: Backend` bound replaced with the inherited bounds instead: `T: SmartContract + ProofSystemCompiler + Default + Debug`

### Additional Context

_No response_

No

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #3949

## Summary

Still pending documentation.

## Additional Context

I copied part of the documentation from [here](https://practice.course.rs/generics-traits/advanced-traits.html#supertraits), I hope it's fine.

## Documentation\*

Check one:

# PR Checklist\*


asterite:
@jfecher In case you want to take a look. I feel like I'm almost there but it's failing on `verify_trait_constraint`... I'll take a look tomorrow, but if you see this PR and see a way to make it work, please go ahead 🙏

compiler/noirc_frontend/src/elaborator/types.rs
```diff
@@ -1808,6 +1819,16 @@ impl<'context> Elaborator<'context> {
     }
 }
 
+fn lookup_method_in_trait(
+    the_trait: &Trait,
+    method_name: &str,
+    trait_bound: &ResolvedTraitBound,
+) -> Option<HirMethodReference> {
+    let trait_method = the_trait.find_method(method_name)?;
+    let generics = trait_bound.trait_generics.clone();
```
jfecher, line 1828: The generics here could be wrong in the case of an inherited trait bound with generics. E.g. for the trait

```noir
trait Foo<A>: Bar<A> { ... }

trait Bar<T> { fn bar(self); }
```
When looking up `bar` (in `lookup_method_in_trait_constraints`) from a trait bound of `Foo<B>` we'd use the trait definition `Bar<T>` and the generics list would be `T` instead of the desired `B`

tooling/nargo_fmt/tests/input/trait.nr
```diff
@@ -0,0 +1,2 @@
+trait Foo: Bar + Baz {}
```
jfecher, line 1: Assuming this is a TODO but I'll comment anyway that the input file is the same as the expected file

compiler/noirc_frontend/src/elaborator/traits.rs
```diff
@@ -53,6 +59,21 @@ impl<'context> Elaborator<'context> {
...
+    ) -> Vec<ResolvedTraitBound> {
+        let mut resolved_trait_bounds = Vec::new();
+
+        for trait_bound in &unresolved_trait.trait_def.bounds {
+            if let Some(resolved_trait_bound) = self.resolve_trait_bound(trait_bound) {
+                resolved_trait_bounds.push(resolved_trait_bound);
+            }
+        }
+
+        resolved_trait_bounds
```
jfecher, line 74:
```suggestion
        unresolved_trait.trait_def.bounds.iter()
            .filter_map(|bound| self.resolve_trait_bound(bound))
            .collect()
```

jfecher (commented):
We'll also need a check that when you implement a super trait that the inherited traits are also implemented.

tooling/nargo_fmt/tests/input/trait.nr
asterite: Oh, it seems traits are formatted "as-is" (without changes), so not really formatted. I wrote the test before realizing this, but I thought it was still good to have it.

compiler/noirc_frontend/src/elaborator/types.rs
asterite: I see... yes, to be honest, I get a bit lost with generics here. But I'll try to figure it out. I guess it'll get trickier if a trait has a parent that also has a parent, etc.

compiler/noirc_frontend/src/elaborator/traits.rs
asterite: Done, thanks! (I didn't apply this because rustmt formats it differently)

compiler/noirc_frontend/src/elaborator/types.rs
asterite: I think I figured it out :-)

compiler/noirc_frontend/src/hir/resolution/errors.rs
```diff
@@ -150,6 +150,8 @@ pub enum ResolverError {
     AttributeFunctionIsNotAPath { function: String, span: Span },
     #[error("Attribute function `{name}` is not in scope")]
     AttributeFunctionNotInScope { name: String, span: Span },
+    #[error("The trait `{the_trait}` is not implemented for `{typ}")]
+    TraitNotImplemented { the_trait: String, typ: String, span: Span },
```
jfecher, line 154: This is only issued for child trait requirements, right? Perhaps we should be more explicit why the trait needs to be implemented. E.g. rust says:
```
error[E0277]: the trait bound `S: Foo` is not satisfied
  --> foo.rs:10:14
   |
10 | impl Bar for S {}
   |              ^ the trait `Foo` is not implemented for `S`
   |
   = help: the trait `Foo` is implemented for `()`
note: required by a bound in `Bar`
  --> foo.rs:3:13
   |
3  | trait Bar : Foo {}
   |             ^^^ required by this bound in `Bar`
```

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -987,6 +1020,55 @@ impl<'context> Elaborator<'context> {
...
+                    &parent_trait_bound.trait_generics.named,
+                )
+                .is_err()
+            {
+                let mut the_trait = parent_trait.name.to_string();
+                if !parent_trait_bound.trait_generics.is_empty() {
+                    the_trait.push('<');
+                    the_trait.push_str(&parent_trait_bound.trait_generics.to_string());
+                    the_trait.push('>');
+                }
```
jfecher, line 1061: trait_generics will already do this formatting when converting it to a string, no need for the extra is_empty or <>

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -987,6 +1020,55 @@ impl<'context> Elaborator<'context> {
...
+
+            if self
+                .interner
+                .try_lookup_trait_implementation(
+                    object_type,
+                    parent_trait_bound.trait_id,
+                    &parent_trait_bound.trait_generics.ordered,
+                    &parent_trait_bound.trait_generics.named,
+                )
+                .is_err()
```
jfecher, line 1054: Looks like we're using the original generics from the supertype here. Does this still work if they're changed in the program? E.g.

```rs
trait Foo<A> {}

trait Bar<B>: Foo<B> {}

impl Foo<i32> for () {}

impl Bar<i32> for () {}
```
Trying this I get `error: The trait Foo<<B>> is not implemented for ()`

asterite: Ah, you are right.

Where do I get the type bindings from the `Trait`? I see there's `associated_types` and `generics` on `Trait`, but I'm not sure how to do it...

asterite: I pushed a commit that solves this case... is there an example with associated types? (I don't fully know how that works in Rust nor in Noir yet)

jfecher: You can see if the following passes for associated types:
```rs
trait Foo { type A; }

trait Bar<B>: Foo<A = B> {}

impl Foo for () { type A = i32; }

impl Bar<i32> for () {}
```

asterite: Apparently yes. I added a test for it.

compiler/noirc_frontend/src/hir/resolution/errors.rs
asterite: It seems we don't have a way for an error to have two locations? I see `Diagnostic::simple_error` has "secondary span" but that seems to be the primary span (or at least I can't find a primary span)

asterite: Hm, maybe there's a way to do it without those convenience methods... I'll check.

asterite: I think I got it! Here's what it looks like now:

```
error: The trait bound `Struct: Foo` is not satisfied
  ┌─ src/main.nr:3:16
  │
3 │ pub trait Bar: Foo {}
  │                --- required by this bound in `Bar
  ·
7 │ impl Bar for Struct {}
  │              ------ The trait `Foo` is not implemented for `Struct
```

docs/docs/noir/concepts/traits.md
```diff
@@ -464,6 +464,32 @@ Since we have an impl for our own type, the behavior of this code will not chang
 to provide its own `impl Default for Foo`. The downside of this pattern is that it requires extra wrapping and
 unwrapping of values when converting to and from the `Wrapper` and `Foo` types.
 
+### Supertraits
+
+Sometimes, you might need one trait to use another trait’s functionality( like the "inheritance" in other languages ). In this case, you need to rely on the dependent trait also being implemented. The trait you rely on is a supertrait of the trait you’re implementing.
```
jfecher, line 469:
```suggestion
Sometimes, you might need one trait to use another trait’s functionality (like "inheritance" in some other languages). In this case, you can specify this relationship by listing any child traits after the parent trait's name and a colon. Now, whenever the parent trait is implemented it will require the child traits to be implemented as well. A parent trait is also called a "super trait."
```

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1033,6 +1033,25 @@ impl<'context> Elaborator<'context> {
...
+        for (param, arg) in the_trait.generics.iter().zip(trait_impl.resolved_trait_generics.iter())
+        {
+            // Avoid binding t = t
+            if !arg.occurs(param.type_var.id()) {
+                bindings.insert(
+                    param.type_var.id(),
+                    (param.type_var.clone(), param.kind(), arg.clone()),
+                );
+            }
+        }
```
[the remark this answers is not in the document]
asterite: I see. I'll try to think of some code that would fail without adding the associated types (without it I can't be sure I implemented it correctly).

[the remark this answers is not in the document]
asterite: I pushed a bunch of refactors so the "same length assertion" will happen now.

To check associated types now we'd need to call `bind_named_generics` but I don't know where to get `args: &[NamedType]` from.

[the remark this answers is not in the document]
jfecher: There's `the_trait.associated_types` for the generics and `trait_impl.associated_types` for the bindings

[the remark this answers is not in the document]
jfecher: Also - instead of writing these special methods, I wonder if we can make use of `resolve_type_args` which should accept any generic thing with some `GenericTypeArgs`. One such thing for the `item: impl Generic` parameter is a TraitId.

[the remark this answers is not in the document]
asterite: > trait_impl.associated_types

It seems those are UnresovledTypes...

I tried to find some code that doesn't work with associated types and couldn't. I think it's because if you have something like this:

```noir
impl Bar<Field, C = i32> for () {}
```

The associated types specified in `Bar` are checked against `Bar`, but they don't show up in parent traits. So for checking a parent trait we don't need to analyze the associated types. But not sure!

I'll try to use `resolve_type_args`.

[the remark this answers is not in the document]
jfecher: I tested this and couldn't break it either, I'll just resolve this comment 👍

jfecher (approved):
LGTM

PR #6253, opened 2024-10-09
feat(perf): Flamegraphs for test program execution benchmarks

### LINKED ISSUE #6186: Add ACVM witness generation benchmarks for a set of test programs
It's very important for us to have performant ACVM execution as this is required to prevent witness generation from being a bottleneck for proof generation.

We currently have some minor benchmarking for ACVM execution in [this file](https://github.com/noir-lang/noir/blob/master/tooling/nargo_cli/benches/criterion.rs). This boils down to use essentially timing the CLI command `nargo execute --force` for 3 of our test functions.

This has a couple problems:
- `nargo execute --force` is recompiling the circuit from scratch each time so we're benchmarking much more than just execution speed
- We're running nargo in a subprocess so we don't get any useful data out of the flamegraph which is generated.

Ideally we would have benchmarks which would:
- Compile the circuit once and loads it into memory
- Runs the circuit using the [ProgramExecutor](https://github.com/noir-lang/noir/blob/be9dcfe56d808b1bd5ef552d41274705b2df7062/tooling/nargo/src/ops/execute.rs#L62) as exposed by the `nargo` crate rather than relying on the CLI.
- Applies https://doc.rust-lang.org/stable/std/hint/fn.black_box.html to the circuit inputs.
- Generates a useful flamegraph which can be used to inform future ACVM optimisations.

We can do this for a single Noir program to begin with and then roll it out to a decently sized suite in future.

aakoshh:
# Description

## Problem\*

Resolves #6186 

## Summary\*

* Breaks up the `nargo_cli/benches/criterion.rs` into two steps: 1) a compilation step which happens only once and 2) the execution of the already compiled program.
* Limits the code in the benchmark loop to be only the circuit execution, excluding the parsing of the program and its inputs and the saving of results as the execute command would do.

These are the timings of just the circuit execution:
```text
eddsa_execute           time:   [111.13 ms 111.19 ms 111.25 ms]
regression_execute      time:   [151.27 µs 151.84 µs 152.66 µs]
struct_execute          time:   [602.39 ns 606.05 ns 609.77 ns]
```

The flame graphs look as follows (you can right click and use the "Save Linked File As..." to download it and open it in a browser to make it interactive):
* `eddsa`: eddsa
* `regression`: regression
* `struct`: struct

To generate them run the following commands:

```shell
./scripts/benchmark_start.sh
cargo bench -p nargo_cli --bench criterion -- --profile-time=30
./scripts/benchmark_stop.sh
```

## Additional Context

The problem with the current `nargo_cli/benches/criterion.rs` was that it executed the `nargo execute --program-dir ... --force` command as a sub-process, and the profiler that creates the Flamegraph only sampled the criterion executor itself, not what the subprocess was doing. 

My first idea to get the flame graphs to include the actual execution was to turn `nargo_cli` into a library _and_ a binary (https://github.com/noir-lang/noir/commit/af19dfc56d583e4b9b36bae042b861e5f715f90c), so that we can import the commands into the benchmark and run them in the same process. This resulted in the following flame graphs:

* `eddsa`: eddsa
* `regression`: regression
* `struct`: struct

These include the entire `ExecuteCommand::run` command, which unfortunately results in the flame graph dominated by parsing logic, reading inputs and writing outputs to files. In fact in all but the `eddsa` example the `execute_circuit` was so fast I can't even find it on the flamegraph. 

These are the timings for the command execution per test program:
```text
eddsa_execute           time:   [984.12 ms 988.74 ms 993.95 ms]
regression_execute      time:   [71.240 ms 71.625 ms 71.957 ms]
struct_execute          time:   [68.447 ms 69.414 ms 70.438 ms]
```

For this reason I rolled back the library+binary change and broke up the execution further by parsing the program into a `CompiledProgram` once and calling `execute_program` in the benchmark loop without saving the results. 

I copied two utility functions to read the program artefact and the input map. Not sure these are worth moving around, even the errors they raise a CLI specific.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
> For this reason I'll undo the library+binary change and try to break up the execution further by parsing the program into a CompiledProgram once and calling execute_program in the benchmark loop without saving the results.

Agreed, the usage of rayon is making these pretty difficult to gain insights from.

tooling/nargo_cli/benches/criterion.rs
```diff
@@ -1,33 +1,147 @@
...
+fn read_compiled_program_with_input(dir: &Path) -> (CompiledProgram, WitnessMap<FieldElement>) {
+    let toml_path = nargo_toml::get_package_manifest(dir).expect("failed to read manifest");
+    let workspace = nargo_toml::resolve_workspace_from_toml(
+        &toml_path,
+        nargo_toml::PackageSelection::All,
+        Some(noirc_driver::NOIR_ARTIFACT_VERSION_STRING.to_string()),
+    )
+    .expect("failed to resolve workspace");
+
+    // Assuming there is only one binary package in these tests.
```
michaeljklein, line 41: Please add a sanity check for this assumption: in case someone tries to benchmark a larger package by just adding it to `selected_tests`

michaeljklein (changes requested):
Tiny change, otherwise LGTM

By the way, I see some explicit bench Noir programs:
- `test_programs/benchmarks/*`
- `execution_success/bench_2_to_17`
- `test_programs/execution_success/bench_ecdsa_secp256k1`

Would it make sense to add these to the `selected_tests`?

tooling/nargo_cli/benches/criterion.rs
aakoshh: Good point, I changed it to return a `Vec<_>` instead, to behave like `ExecuteCommand` and run all of them.

aakoshh:
> By the way, I see some explicit bench Noir programs:
> Would it make sense to add these to the selected_tests?

Thanks, I haven't noticed these. Added all programs from the `benchmarks` directory to the selection. The `execution_success/bench_*` ones seem very similar to those.

tooling/nargo_cli/benches/criterion.rs
TomAFrench: I would say that assuming that we're only going to have a single package in the workspace is a good assumption (our workspace tests tend to be very simple as we're just testing how we handle dependencies, etc.). It's not a blocker now that it's in though.

aakoshh: That’s what I thought, but perhaps if we add test programs in bulk like with the benchmarks directory then this is a more defensive way that won’t stop working because of an outlier.

michaeljklein (approved):
LGTM

PR #6257, opened 2024-10-09
feat: don't crash LSP when there are errors resolving the workspace

asterite:
# Description

## Problem

Before this PR, LSP would crash for assumed workspaces (when we can't find a Nargo.toml) or when there were errors in the Nargo.toml.

## Summary

Now LSP doesn't crash in the above cases:
- for assumed workspaces we don't crash (we used to not crash but I made some changes in the past that broke this)
- for errors in Nargo.toml we now output to STDERR, so you can see the error in the output. This is what Rust Analyzer does, except that they also have a "rust-analyzer" thing at the bottom that becomes yellow when there's an error... I didn't implement this (it probably requires some code on the extension side) but I think if something doesn't work you'd check the logs or ask, and it's worse if it crashes.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/lsp/src/lib.rs
+++ b/tooling/lsp/src/lib.rs
@@ -267,12 +267,16 @@ fn byte_span_to_range<'a, F: files::Files<'a> + ?Sized>(
 
 pub(crate) fn resolve_workspace_for_source_path(file_path: &Path) -> Result<Workspace, LspError> {
     if let Some(toml_path) = find_file_manifest(file_path) {
-        return resolve_workspace_from_toml(
+        match resolve_workspace_from_toml(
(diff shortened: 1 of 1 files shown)


michaeljklein (approved):
LGTM

Perhaps make a follow-up issue r.e. the following, in case someone doesn't see the logs in the `STDERR` and looks for an existing issue?
> except that they also have a "rust-analyzer" thing at the bottom that becomes yellow when there's an error... I didn't implement this (it probably requires some code on the extension side) but I think if something doesn't work you'd check the logs or ask, and it's worse if it crashes.

asterite:
Done! #6259

PR #6261, opened 2024-10-09
fix: visibility for impl methods

asterite:
# Description

## Problem

Part of #4515

Apparently in #6179 I only made it work for calls like `foo::Bar::baz()` but no checks were done for method calls (where `self` is involved).

## Summary

This PR now warns when calling private or `pub(crate)` methods, either on structs or primitive types, when they are not accessible from the current module.

We also now don't suggest non-accessible methods from LSP.

## Additional Context

I also removed an unused global that was introduced some time ago.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir_def/types.rs
+++ b/compiler/noirc_frontend/src/hir_def/types.rs
@@ -983,6 +983,34 @@ impl Type {
         }
     }
 
+    pub fn is_primitive(&self) -> bool {
+        match self.follow_bindings() {
+            Type::FieldElement
+            | Type::Array(_, _)
+            | Type::Slice(_)
+            | Type::Integer(..)
+            | Type::Bool
+            | Type::String(_)
+            | Type::FmtString(_, _)
+            | Type::Unit
+            | Type::Function(..)
+            | Type::Tuple(..) => true,
+            Type::Alias(alias_type, generics) => {
+                alias_type.borrow().get_type(&generics).is_primitive()
+            }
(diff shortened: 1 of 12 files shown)


compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -983,6 +983,34 @@ impl Type {
...
+            | Type::Function(..)
+            | Type::Tuple(..) => true,
+            Type::Alias(alias_type, generics) => {
+                alias_type.borrow().get_type(&generics).is_primitive()
+            }
+            Type::MutableReference(typ) => typ.is_primitive(),
+            Type::Struct(..)
+            | Type::TypeVariable(..)
+            | Type::TraitAsType(..)
+            | Type::NamedGeneric(..)
```
jfecher, line 1005: type variables & named generics can be primitive if they're bound to one

jfecher (commented):
One thing, otherwise LGTM

compiler/noirc_frontend/src/hir_def/types.rs
asterite: I see. I was about to fix this but I noticed there's a `follow_bindings()` call at the beginning. Would that take care of this?

jfecher: Ah, I missed that as well. It's probably fine then

jfecher (approved):

PR #6262, opened 2024-10-09
feat: Add `checked_transmute`

jfecher:
# Description

## Summary\*

Adds a way for users to get around limitations in the type checker, particularly for arithmetic generics, by using a new `checked_transmute` builtin. This will cast any type to another but will assert during monomorphization that the types unify.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/monomorphization/mod.rs
```diff
@@ -1406,11 +1408,35 @@ impl<'interner> Monomorphizer<'interner> {
...
+        &mut self,
+        expr_id: node_interner::ExprId,
+        arguments: &[node_interner::ExprId],
+        argument_values: &[ast::Expression],
+    ) -> Result<ast::Expression, MonomorphizationError> {
+        let location = self.interner.expr_location(&expr_id);
+        let actual = self.interner.id_type(arguments[0]).follow_bindings();
+        let expected = self.interner.id_type(expr_id).follow_bindings();
+
+        if actual.try_unify(&expected, &mut TypeBindings::new()).is_err() {
```
michaeljklein, line 1431: Don't we need to apply any type bindings, or else assert that there are none from `try_unify`?

docs/docs/noir/standard_library/mem.md
```diff
@@ -0,0 +1,52 @@
...
+
+# `std::mem::checked_transmute`
+
+```rust
+fn checked_transmute<T, U>(value: T) -> U
+```
+
+Transmutes a value of one type into the same value but with a new type `U`.
+
+This function is safe to use since both types are asserted to be equal later during compilation.
```
michaeljklein, line 44: `later during compilation` -> `when all type-level variables are known`?

docs/docs/noir/standard_library/mem.md
```diff
@@ -0,0 +1,52 @@
...
+This function is useful for cases where the compiler may fails a type check that is expected to pass where
+a user knows the two types to be equal. For example, when using arithmetic generics there are cases the compiler
+does not see as equal, such as `[Field; N*(A + B)]` and `[Field; N*A + N*B]`, which users may know to be equal.
+In these cases, `checked_transmute` can be used to cast the value to the desired type while also preserving safety
+by checking this equality later on.
```
michaeljklein, line 49:
```suggestion
by checking this equality once `N`, `A`, `B` are fully resolved.
```
?

Not sure how to describe monomorphization without mentioning `monomorphization`

noir_stdlib/src/mem.nr
```diff
@@ -4,3 +4,14 @@
 #[builtin(zeroed)]
 pub fn zeroed<T>() -> T {}
 
+/// Transmutes a value of type T to a value of type U.
+///
+/// This operation is checked in that in a later stage of compilation
```
michaeljklein, line 9: `This operation is checked in that in a later stage`

Awkward phrasing, perhaps something like:

`This operation is checked in a stage of compilation after normal type checking, where`

?

noir_stdlib/src/mem.nr
```diff
@@ -4,3 +4,14 @@
...
+/// both types are asserted to be equal. If not, a compile error is issued.
```
michaeljklein, line 10:
```suggestion
/// both types are asserted to be equal. If not, a compilation error is issued.
```
?

noir_stdlib/src/mem.nr
```diff
@@ -4,3 +4,14 @@
...
+///
+/// This function is useful for types using arithmetic generics for cases
```
michaeljklein, line 12:
```suggestion
/// This function is useful for types using arithmetic generics in cases
```

noir_stdlib/src/mem.nr
```diff
@@ -4,3 +4,14 @@
...
+/// which the compiler otherwise cannot prove as equal during type checking.
```
michaeljklein, line 13:
```suggestion
/// which the compiler otherwise cannot prove equal during type checking.
```

michaeljklein (changes requested):

compiler/noirc_frontend/src/monomorphization/mod.rs
jfecher: I was thinking all type variables should be gone by monomorphization but I might as well change it to unify

docs/docs/noir/standard_library/mem.md
jfecher: Changed it to "... after the concrete values for generic types become known"

noir_stdlib/src/mem.nr
jfecher: Reworded this a bit

michaeljklein (approved):
LGTM

PR #6265, opened 2024-10-10
feat(test): Fuzz test poseidon2 hash equivalence

aakoshh:
# Description

## Problem\*

Related to #6141 

## Summary\*

Adds a fuzz test for `poseidon2.nr` comparing the results to [bn254_blackbox_solver::poseidon_hash](https://github.com/noir-lang/noir/blob/70cbeb4322a0b11c1c167ab27bf0408d04fe7b7d/acvm-repo/bn254_blackbox_solver/src/poseidon2.rs#L547), which says it's `"Performs a poseidon hash with a sponge construction equivalent to the one in poseidon2.nr"`

To pass the test the Rust implementation was given a new `is_variable_length` parameter to inform it whether it needs to append an extra 1 like Noir and Berratenberg do. 

## Additional Context

The test initially failed:

```console
❯ cargo test -p nargo_cli --test stdlib-props poseidon
  
test fuzz_poseidon2_equivalence ... FAILED

---- fuzz_poseidon2_equivalence stdout ----
Test failed: assertion failed: `(left == right)` 
  left: `Field(1187863985434533916290764679013201786939267142671550539990974992402592744116)`, 
 right: `Field(11250791130336988991462250958918728798886439319225016858543557054782819955502)`: max_len = 1 at tooling/nargo_cli/tests/stdlib-props.rs:106.
minimal failing input: io = SnippetInputOutput {
    description: "max_len = 1",
    inputs: {
        "input": Vec(
            [
                Field(
                    0,
                ),
            ],
        ),
        "message_size": Field(
            0,
        ),
    },
    expected_output: Field(
        11250791130336988991462250958918728798886439319225016858543557054782819955502,
    ),
}
```

So we pass in `input = [0; 1]` with `message_size=0`. It fails because the Noir code treats the case where the `message_size` is different from the maximum length differently by [appending](https://github.com/noir-lang/noir/blob/70cbeb4322a0b11c1c167ab27bf0408d04fe7b7d/noir_stdlib/src/hash/poseidon2.nr#L75-L80) an extra 1, to keep variable and fixed length hashes distinct. The Rust implementation doesn't do this, nor did the other hashes tested so far. I'm not sure if it's worth noting that the hash will not depend on how much shorter the message is than the maximum, just that it's shorter.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/acvm-repo/bn254_blackbox_solver/src/poseidon2.rs
+++ b/acvm-repo/bn254_blackbox_solver/src/poseidon2.rs
@@ -544,12 +544,22 @@ impl<'a> Poseidon2<'a> {
 }
 
 /// Performs a poseidon hash with a sponge construction equivalent to the one in poseidon2.nr
-pub fn poseidon_hash(inputs: &[FieldElement]) -> Result<FieldElement, BlackBoxResolutionError> {
+///
+/// The `is_variable_length` parameter is there to so we can produce an equivalent hash with
+/// the Barretenberg implementation which distinguishes between variable and fixed length inputs.
+/// Set it to true if the input length matches the static size expected by the Noir function.
+pub fn poseidon_hash(
+    inputs: &[FieldElement],
+    is_variable_length: bool,
+) -> Result<FieldElement, BlackBoxResolutionError> {
     let two_pow_64 = 18446744073709551616_u128.into();
     let iv = FieldElement::from(inputs.len()) * two_pow_64;
     let mut sponge = Poseidon2Sponge::new(iv, 3);
     for input in inputs.iter() {
         sponge.absorb(*input)?;
+    }
+    if is_variable_length {
+        sponge.absorb(FieldElement::from(1u32))?;
     }
     sponge.squeeze()
 }
@@ -640,7 +650,7 @@ mod test {
             FieldElement::from(3u128),
             FieldElement::from(4u128),
         ];
-        let result = super::poseidon_hash(&fields).expect("should hash successfully");
+        let result = super::poseidon_hash(&fields, false).expect("should hash successfully");
         assert_eq!(
             result,
             field_from_hex("130bf204a32cac1f0ace56c78b731aa3809f06df2731ebcf6b3464a15788b1b9"),
--- a/tooling/nargo_cli/tests/stdlib-props.rs
+++ b/tooling/nargo_cli/tests/stdlib-props.rs
@@ -259,14 +259,13 @@ fn fuzz_sha512_equivalence() {
 fn fuzz_poseidon2_equivalence() {
     use bn254_blackbox_solver::poseidon_hash;
 
-    for max_len in [0, 1, 4, 511, 512] {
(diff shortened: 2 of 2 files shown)


TomAFrench:
This appending of 1 matches with [the barretenberg implementation of poseidon2](https://github.com/AztecProtocol/aztec-packages/blob/662b61e4c20a2d4217980922d4578f4dfeacae6b/barretenberg/cpp/src/barretenberg/stdlib/hash/poseidon2/sponge/sponge.hpp#L148-L153) so looks like the issue is that we're not telling the rust impl to do this also. We could modify `poseidon_hash` but also could modify the fuzz test so that we just append 1 to the input there.

aakoshh:
Okay, so as I understand the Rust implementation should be able to perform the same hash, but it needs a hint because it takes a slice and cannot tell if it's dealing with variable length data.

> We could ... also could modify the fuzz test so that we just append 1 to the input there.

Unfortunately that would result in a change in the IV, which depends on the input length. 

I'll add some other function to the module and document this gotcha. 

I was trying to find external implementations of Poseidon2 to test with. For example there is [this one](https://github.com/HorizenLabs/poseidon2/tree/main/plain_implementations/src/poseidon2) which we might be able to configure for bn254, but it's not obvious for me what the configuration are as it takes more parameters than ours. Do you think it's worth pursuing this, or if the Noir parts on top of the foreign permutation agree with the Rust parts then we're okay?

aakoshh:
I added a parameter to the Rust function to match Noir; couldn't think of a more elegant way to do it.

Perhaps it's worth taking a note that Barretenberg has a [variable and a fixed length variant](https://github.com/AztecProtocol/aztec-packages/blob/662b61e4c20a2d4217980922d4578f4dfeacae6b/barretenberg/cpp/src/barretenberg/stdlib/hash/poseidon2/sponge/sponge.hpp#L162-L180) of the hash function, while Noir decides based on the actual input length. That is, one can call `hash_variable_length` in Barretenberg with an array that just happens to be as long as the N is, but in theory it could be less, and then Noir would treat it as if it was fixed length without a doubt. So the caller of the C++ could has to know N and do the if/then/else on their side, or Noir would have to have two methods as well.

aakoshh:
I couldn't actually find anything calling `hash_variable_length`. OTOH [poseidon2::hash](https://github.com/AztecProtocol/aztec-packages/blob/1801f5b49fb3b153817a1596c6fd568f1c762fe5/barretenberg/cpp/src/barretenberg/stdlib/hash/poseidon2/poseidon2.cpp#L14-L17) calls `hash_fixed_length` but says it should "just call the variable length" 🤷‍♂️

TomAFrench (approved):
LGTM!

aakoshh:
FWIW I found we already have a test against an external implementation, showcasing the [mapping of configuration](https://github.com/noir-lang/noir/blob/f61ba037c6726e19be4f894d9447fe396df95417/acvm-repo/acvm/tests/solver.rs#L1214-L1231) as well.

PR #6273, opened 2024-10-10
feat(test): Fuzz poseidon hases against an external library

aakoshh:
# Description

## Problem\*

Related to #6141 

## Summary\*

Testing `std::hash::poseidon::bn254::hash_[1-12]` against https://github.com/Lightprotocol/light-poseidon/tree/v0.2.0 

The library doesn't support inputs wider than 12 (doesn't have the parameters for it), so the Noir function `hash_[13-16]` are not covered.

## Documentation\*

Check one:

# PR Checklist\*


Cargo.toml
```diff
@@ -142,7 +142,7 @@ build-data = "0.1.3"
 bincode = "1.3.3"
 hex = "0.4.2"
 const_format = "0.2.30"
-num-bigint = "0.4"
+num-bigint = "0.4.4"
```
aakoshh, line 145: I had to bump this because otherwise cargo couldn't reconcile the versions. It looks like the one needed by `light-poseidon` is compatible, but because it's a `dev-dependency`, by the time cargo got there it was locked at an earlier version.

tooling/nargo_cli/Cargo.toml
```diff
@@ -85,6 +86,7 @@ sha2.workspace = true
 sha3.workspace = true
 iai = "0.1.1"
 test-binary = "3.0.2"
+light-poseidon = { git = "https://github.com/Lightprotocol/light-poseidon.git", tag = "v0.2.0" }
```
TomAFrench, line 89: Looks like it's been published to crates.io (https://crates.io/crates/light-poseidon) so we don't need the git dependency?

aakoshh: Ah cool! 🦅 👁️

tooling/nargo_cli/tests/stdlib-props.rs
```diff
@@ -259,23 +259,16 @@ fn fuzz_sha512_equivalence() {
 fn fuzz_poseidon2_equivalence() {
     use bn254_blackbox_solver::poseidon_hash;
 
-    for max_len in [0, 1, 3, 4, 511, 512] {
+    for max_len in [0, 1, 100] {
```
TomAFrench, line 262: Is there a reason for reducing the number of lengths we test?

TomAFrench (approved):
I'm not clear on one of the changes to poseidon2 but otherwise LGTM.

tooling/nargo_cli/tests/stdlib-props.rs
aakoshh: It took ~35 seconds to test with the larger inputs. Since I just copied them from how you added it to one of the SHAs, I thought the 512 boundary is probably not that relevant to this one. Do you have any preference in this case?

aakoshh: Added back 3 and 4 - there is some logic that kicks in after absorbing 3 inputs, maybe that's a good place to probe.

Issue #5462, opened 2024-07-09
`[T]::pop_front()` panics when inside non-executed `if`


michaeljklein:
### Aim

Attempted to conditionally `pop_front` from a slice, depending on whether it's empty:

```rust
fn main() {
    let empty_slice: [u8] = &[];

    // Also panics:
    // if empty_slice != &[] {
    if empty_slice.len() != 0 {
        // Works:
        // let _ = empty_slice.pop_back();

        // Panics:
        let _ = empty_slice.pop_front();
    }
}
```

### Expected Behavior

I expected the predicate to prevent `pop_front` from executing and failing when the slice is empty.

### Bug

```bash
The application panicked (crashed).
Message:  There are no elements in this slice to be removed
Location: compiler/noirc_evaluator/src/ssa/ir/instruction/call.rs:162
```

### To Reproduce

1.
2.
3.
4.

### Project Impact

Nice-to-have

### Impact Context

NOTE: this is required to finish the slice control flow test [here](https://github.com/noir-lang/noir/blob/ee4cc5afadb3200025b54eb4dc83224f988441bc/test_programs/execution_success/slice_regex/src/main.nr).

### Workaround

None

### Installation Method

Binary (`noirup` default)

### Nargo Version

nargo version = 0.31.0 noirc version = 0.31.0+d44f882be094bf492b1742370fd3896b0c371f59 (git version hash: d44f882be094bf492b1742370fd3896b0c371f59, is dirty: false)

### Would you like to submit a PR for this Issue?

None

jfecher:
Looks like `pop_back` is broken as well but in a different way.

Instead of calling `pop_back` on its slice internally it repeatedly calls push_front to build a new slice. I believe this is to avoid logic on guessing how many times to pop when the element is a struct type which is represented as multiple SSA values.

Anyway, neither `pop_back` nor `pop_front` have checks for if the slice is empty. `pop_front` leads to the error you found while `pop_back` will silently "optimize" to an empty array result. Execution will still fail though because it will leave behind a `v0 = sub u32 0, u32 1` instruction. When this happens the program just exits with "Constraint Failed" and no explanation.

It looks like there is a length check on `SliceRemove` and `SliceInsert` at least.

PR #6274, opened 2024-10-10
fix: prevent compiler panic when popping from empty slices

TomAFrench:
# Description

## Problem\*

Partial fix to https://github.com/noir-lang/noir/issues/5462

## Summary\*

This PR doesn't fully fix https://github.com/noir-lang/noir/issues/5462 but allows full compilation of a program without panicking. The test program from #5462 now successfully compiles (due to us being able to remove the inactive branch during inlining)

We still fail to properly handle branches which are conditional on runtime values such as the below:
```
fn main(active: bool) {
    let empty_slice: [u8] = &[];

    if !active {
        let _ = empty_slice.pop_front();
    }
}
```

This compiles to the below program

```
func 0
current witness index : 2
private parameters indices : [0]
return value indices : []
BLACKBOX::RANGE [(0)] [ ]
INIT (id: 0, len: 0) 
EXPR [ (-1, _1) 0 ]
MEM (id: 0, read at: x1, value: x2) 
```

This will then fail to execute with the below error:
```

   ┌─ /mnt/user-data/tom/noir/test_programs/execution_success/inactive_slice_popping/src/main.nr:11:17
   │
11 │         let _ = empty_slice.pop_front();
   │                 ---------------------
   │
   = Call stack:
     1. /mnt/user-data/tom/noir/test_programs/execution_success/inactive_slice_popping/src/main.nr:11:17

```

Note the memory block has been allocated with a length of 0 so even applying a predicate to the array get will fail.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/acvm-repo/acvm_js/src/lib.rs
+++ b/acvm-repo/acvm_js/src/lib.rs
@@ -17,7 +17,8 @@ mod logging;
 mod public_witness;
 
 pub use black_box_solvers::{
-    and, blake2s256, ecdsa_secp256k1_verify, ecdsa_secp256r1_verify, sha256_compression, xor,
+    and, blake2s256, ecdsa_secp256k1_verify, ecdsa_secp256r1_verify, keccak256, sha256_compression,
+    xor,
 };
 pub use build_info::build_info;
 pub use compression::{
--- a/acvm-repo/acvm_js/src/lib.rs
+++ b/acvm-repo/acvm_js/src/lib.rs
@@ -17,8 +17,7 @@ mod logging;
 mod public_witness;
 
 pub use black_box_solvers::{
-    and, blake2s256, ecdsa_secp256k1_verify, ecdsa_secp256r1_verify, sha256_compression,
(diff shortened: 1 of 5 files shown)


vezenovm (changes requested):
I think if you add `codegen_slice_access_check` to the intrinsic checks during SSA gen we can also avoid the runtime error. This is what is done for insert and remove.

TomAFrench:
@vezenovm I don't think that will help. We'll end up injecting an assertion that 0 >= 1 which will always fail. I think the best way to solve this is to use the tracked slice capacities and then we can inject an `assert(false)` in place of the pop during acirgen.

vezenovm:
> I think the best way to solve this is to use the tracked slice capacities and then we can inject an `assert(false)` in place of the pop during acirgen.

Ah I misread the failing test. Yeah we will have to replace the pop with an assertion.

vezenovm (approved):

PR #6277, opened 2024-10-10
feat: show LSP diagnostic related information

### LINKED ISSUE #6272: Failed assertions in macros are reported in the comptime code, but not the annotation
### Aim

Given: 

```noir
use std::{collections::umap::UHashMap, hash::{BuildHasherDefault, poseidon2::Poseidon2Hasher}};

comptime mut global ARG_STORAGE: UHashMap<FunctionDefinition, str<7>, BuildHasherDefault<Poseidon2Hasher>> = UHashMap::default();

comptime fn this_has_to_run_first(f: FunctionDefinition, arg: str<7>) {
    ARG_STORAGE.insert(f, arg);
}

comptime fn this_has_to_run_after(f: FunctionDefinition) {
    let maybe_arg = ARG_STORAGE.get(f);
    assert(maybe_arg.is_some(), "Function has to run last, please invert the annotation order");
    // Do something with arg
}

#[this_has_to_run_after]
#[this_has_to_run_first("oh noes")]
fn main() {}
```

I would like a hint in the `main` function that the attribute that failed to run was `this_has_to_run_after`

### Expected Behavior

LSP should show a squiggly red line on `#[this_has_to_run_after]`

### Bug

Image

LSP shows the error at the failed assertion

### To Reproduce

Use the example above

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Errors shown in LSP didn't show the "secondaries" we capture, nor the notes.

Resolves #6272

## Summary

Now secondaries and notes are shown as details to the primary error, like in Rust Analyzer:

lsp-related-information-1

lsp-related-information-2

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/lsp/src/notifications/mod.rs
+++ b/tooling/lsp/src/notifications/mod.rs
@@ -8,7 +8,7 @@ use crate::{
 use async_lsp::{ErrorCode, LanguageClient, ResponseError};
 use fm::{FileManager, FileMap};
 use fxhash::FxHashMap as HashMap;
-use lsp_types::{DiagnosticTag, Url};
+use lsp_types::{DiagnosticRelatedInformation, DiagnosticTag, Url};
 use noirc_driver::check_crate;
 use noirc_errors::{DiagnosticKind, FileDiagnostic};
 
@@ -195,12 +195,18 @@ fn publish_diagnostics(
 
     for file_diagnostic in file_diagnostics.into_iter() {
         let file_id = file_diagnostic.file_id;
-        let diagnostic = file_diagnostic_to_diagnostic(file_diagnostic, files);
-
         let path = fm.path(file_id).expect("file must exist to have emitted diagnostic");
-        if let Ok(uri) = Url::from_file_path(path) {
-            diagnostics_per_url.entry(uri).or_default().push(diagnostic);
-        }
+        let Ok(uri) = Url::from_file_path(path) else {
+            continue;
+        };
+
+        let Some(diagnostic) =
+            file_diagnostic_to_diagnostic(file_diagnostic, files, fm, uri.clone())
+        else {
+            continue;
+        };
+
+        diagnostics_per_url.entry(uri).or_default().push(diagnostic);
     }
 
     let new_files_with_errors: HashSet<_> = diagnostics_per_url.keys().cloned().collect();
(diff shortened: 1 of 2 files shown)


tooling/lsp/src/notifications/mod.rs
```diff
@@ -255,13 +265,62 @@ fn file_diagnostic_to_diagnostic(file_diagnostic: FileDiagnostic, files: &FileMa
...
+        let secondary_file = secondary.file.unwrap_or(file_id);
+        let Some(path) = fm.path(secondary_file) else {
+            continue;
+        };
+        let Ok(uri) = Url::from_file_path(path) else {
+            continue;
+        };
+        let Some(range) = byte_span_to_range(files, file_id, secondary.span.into()) else {
+            continue;
+        };
```
michaeljklein, line 280: nit: could these three `let _ = _ else { continue }` be combined into something like:
```
let Some((uri, range)) = {
  let path = fm.path(..)?;
  let uri = ...ok()?;
  let range = ..?;
  Some((uri, range))
} else {
  continue;
}
```
?

tooling/lsp/src/notifications/mod.rs
```diff
@@ -255,13 +265,62 @@ fn file_diagnostic_to_diagnostic(file_diagnostic: FileDiagnostic, files: &FileMa
...
+    for frame in diagnostic.call_stack.into_iter().rev() {
+        let Some(path) = fm.path(frame.file) else {
+            continue;
+        };
+        let Ok(uri) = Url::from_file_path(path) else {
+            continue;
+        };
+        let Some(range) = byte_span_to_range(files, file_id, frame.span.into()) else {
+            continue;
+        };
```
michaeljklein, line 304: nit: same question here r.e. repeated `let .. else { continue }`

tooling/lsp/src/notifications/mod.rs
asterite: I'm not sure... if you use `?` inside a loop, does it continue the loop or return from the function? In this case I need to continue. I could extract the loop body into a separate function and use `?` there 🤔 I'll try that.

jfecher: `?` would be invalid here since it'd try to return from the outer function. You could maybe wrap it in a closure that is immediately invoked though.

asterite: I introduced some helper functions to be able to use `?`. I think the code ended up being clearer, thanks!

Issue #6244, opened 2024-10-08
`std::merkle::compute_merkle_root` returns same values


jewelofchaos9:
## Problem

If last N (size of hash_path) bits of `index` variable are equal, `std::merkle::compute_merkle_root` returns same values

## Bug
```noir
use dep::std;

fn main() {
}

#[test]
fn test_main() {
  let hash_path = [0x1e284aa596cd3d6d3fda0d931804640fff8d3ec437d6741edb730320d711c3b0, 0x169af1edea6b1ee05e68ea2ea6a0ae636d84e0adaa652b7aeac2be73e3a69f53];
  let secret = 0x9532535eab446f71dbfea9dbaac6f16;

  let unique_id_str = "0x4fff9ceb8dd9e0b8d65c59760ea94551";
  let unique_id = U128::from_hex(unique_id_str);
  let note_commitment = std::hash::pedersen_hash([secret, unique_id.to_integer()]);

  let computed_root1 = std::merkle::compute_merkle_root(note_commitment, 8, hash_path);
  let computed_root2 = std::merkle::compute_merkle_root(note_commitment, 16, hash_path);
  let computed_root3 = std::merkle::compute_merkle_root(note_commitment, 32, hash_path);

  assert(computed_root1 == computed_root2);
  assert(computed_root2 == computed_root3);
  assert(computed_root1 == computed_root3);
}
```

## Possible fix
Check if `index` fits in bitarray of size N. 

## To Reproduce
1. Create Noir project for program written in bug section.
2. Run `nargo test`

## Environment
```
nargo version = 0.33.0
noirc version = 0.33.0+325dac54efb6f99201de9fdeb0a507d45189607d
(git version hash: 325dac54efb6f99201de9fdeb0a507d45189607d, is dirty: false)
```

PR #6278, opened 2024-10-10
fix: enforce correctness of decompositions performed at compile time

TomAFrench:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6244

## Summary\*

We were not checking that a radix decomposition simplification performed at compile time was valid, resulting in a truncation to fit the result array. This PR prevents the simplification in this case so we can instead error at runtime.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/ir/instruction/call.rs
```diff
@@ -60,9 +60,14 @@ pub(super) fn simplify_call(
                 } else {
                     unreachable!("ICE: Intrinsic::ToRadix return type must be array")
                 };
-                let result_array = constant_to_radix(endian, field, 2, limb_count, dfg);
-
-                SimplifyResult::SimplifiedTo(result_array)
+                if limb_count < field.num_bits() {
```
guipublic, line 63: I would move this logic inside `constant_to_radix`, where we can check if limb_count is at least decomposed_integer.len() and return an optional valueId

guipublic (changes requested):

compiler/noirc_evaluator/src/ssa/ir/instruction/call.rs
TomAFrench: Yeah, that's a good point. I've made this change.

guipublic (approved):
LGTM

PR #6281, opened 2024-10-11
feat: optimize reading a workspace's files

asterite:
# Description

## Problem

Reading all workspace files is slow.

## Summary

**TL;DR: reading all `.nr` files in the noir-circuits project now takes 10ms instead of 1.5 seconds.**

Using `flamegraph` I noticed reading a workspace files was slow. For example in `noir-contracts` it was taking 1.5 seconds to do that.

The first thing in this PR is using the `walkdir` crate which brings two benefits:
- It reduces the time to 1.3 seconds (not much, but it's something)
- It removes the need for us to reproduce that code and maintain it (we had tests for walking a dir)

Still, 1.3 seconds seemed like too slow. At first I searched if there was alternatives to `std::fs::read_string` but it seems that's the most efficient thing to use. So I created a separate project that essentially did `walkdir` on `noir-contracts`, reading all files and it took around 10ms (!!). So there must be something else going on...
 
It turns out, if we have a workspace with projects A, B and C, with both A and B depending on C, C's files were read twice. So the next thing I did was to avoid reading a file's source if it was already in the file manager. That reduces the time ro 340ms. But the separate project was taking 10ms to do the same! There's still something else...

Well, we still walked C's directories twice (in the above example). So the next thing is to keep track of which packages we already added files for, and not traverse those dirs more than once. With that, now adding all the files for noir-contracts takes around 10ms.

(the difference is so big because in noir-contracts we have many projects depending on many others, resulting in some project files being read like, say, 10 times instead of 1)

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2528,9 +2528,9 @@ dependencies = [
  "rand 0.8.5",
  "rayon",
  "serde",
- "tempfile",
  "thiserror",
  "tracing",
+ "walkdir",
 ]
 
 [[package]]
--- a/compiler/fm/src/lib.rs
+++ b/compiler/fm/src/lib.rs
@@ -100,6 +100,11 @@ impl FileManager {
         self.id_to_path.get(&file_id).map(|path| path.as_path())
     }
 
+    pub fn has_file(&self, file_name: &Path) -> bool {
+        let file_name = self.root.join(file_name);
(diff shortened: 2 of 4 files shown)


michaeljklein (approved):
LGTM

PR #6282, opened 2024-10-11
feat(improve): Remove scan through globals

### LINKED ISSUE #5001: Avoid `O(N^2)` scan through globals during name resolution
### Problem

the following [code](https://github.com/noir-lang/noir/blob/master/compiler/noirc_frontend/src/hir/resolution/resolver.rs#L418)
in the resolver scans through all previous global's whenever a global is added, resulting in `O(num_globals^2)`:

```rust
        // This check is necessary to maintain the same definition ids in the interner. Currently, each function uses a new resolver that has its own ScopeForest and thus global scope.
        // We must first check whether an existing definition ID has been inserted as otherwise there will be multiple definitions for the same global statement.
        // This leads to an error in evaluation where the wrong definition ID is selected when evaluating a statement using the global. The check below prevents this error.
        let mut global_id = None;
        let global = self.interner.get_all_globals();
        for global_info in global {
            if global_info.ident == name
                && global_info.local_id == self.path_resolver.local_module_id()
            {
                global_id = Some(global_info.id);
            }
        }
```

### Happy Case

Store global variables in a set or (hash)map to allow looking them up quickly

### Project Impact

Nice-to-have

### Impact Context

_No response_

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #5001 

## Summary\*

Removes the search for a global by local module ID and name when elaborating a global variable declaration. Instead simply looks up the global by the ID that we have in the definition we are elaborating. In fact the elaboration starts from a global ID of an unresolved global. 

I could be way off the mark on whether this is the right thing to do.

## Additional Context

I found a couple of things odd:
* `NodeInterner::push_global` pushes a `DefinitionKind::Global(id)` with the exact ID that the `GlobalInfo` will have, ie. it is self-referential (just an observation). This is called during the collection of unresolved globals.
* `Elaborator::add_global_variable_decl` only looks for globals by `local_id` and `name`, but _not_ by `definition`, even though at this point we know that the `definition` must contain a `GlobalId`, otherwise we would have stayed in `Elaborator::add_variable_decl`. 
* if a global cannot be found, it calls `NodeInterner::push_definition` with the `definition` in the parameters; since we know it contains a `GlobalId` and that the same ID must have been inserted in `push_global` as well, it will now be part of two definitions at different locations - not sure exactly what this means or looks like in code

I added some assertions in the commits to check that during the tests on CI, the `global_id` and the `definition`, and any results found by the lookup by local+name, are always consistent, which is why in the end I thought that if the only thing we can find is the one we can get by ID, why not get just get it?

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/patterns.rs
+++ b/compiler/noirc_frontend/src/elaborator/patterns.rs
@@ -386,13 +386,8 @@ impl<'context> Elaborator<'context> {
         // This check is necessary to maintain the same definition ids in the interner. Currently, each function uses a new resolver that has its own ScopeForest and thus global scope.
         // We must first check whether an existing definition ID has been inserted as otherwise there will be multiple definitions for the same global statement.
         // This leads to an error in evaluation where the wrong definition ID is selected when evaluating a statement using the global. The check below prevents this error.
-        let mut global_id = None;
-        let global = self.interner.get_all_globals();
-        for global_info in global {
-            if global_info.local_id == self.local_module && global_info.ident == name {
-                global_id = Some(global_info.id);
-            }
-        }
+        let global_id =
+            self.interner.find_global_by_local_and_ident(self.local_module, name.clone());
 
         let (ident, resolver_meta) = if let Some(id) = global_id {
             let global = self.interner.get_global(id);
--- a/compiler/noirc_frontend/src/elaborator/patterns.rs
+++ b/compiler/noirc_frontend/src/elaborator/patterns.rs
@@ -37,7 +37,6 @@ impl<'context> Elaborator<'context> {
             None,
             &mut Vec::new(),
             warn_if_unused,
-            None,
         )
     }
 
@@ -50,7 +49,6 @@ impl<'context> Elaborator<'context> {
         definition_kind: DefinitionKind,
         created_ids: &mut Vec<HirIdent>,
         warn_if_unused: bool,
-        global_id: Option<GlobalId>,
     ) -> HirPattern {
         self.elaborate_pattern_mut(
             pattern,
@@ -59,7 +57,6 @@ impl<'context> Elaborator<'context> {
             None,
             created_ids,
             warn_if_unused,
-            global_id,
         )
     }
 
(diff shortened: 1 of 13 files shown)


compiler/noirc_frontend/src/elaborator/patterns.rs
```diff
@@ -386,13 +388,8 @@ impl<'context> Elaborator<'context> {
...
         // This leads to an error in evaluation where the wrong definition ID is selected when evaluating a statement using the global. The check below prevents this error.
-        let mut global_id = None;
-        let global = self.interner.get_all_globals();
-        for global_info in global {
-            if global_info.local_id == self.local_module && global_info.ident == name {
-                global_id = Some(global_info.id);
-            }
-        }
+        let global_id =
+            self.interner.find_global_by_local_and_ident(self.local_module, name.clone());
```
aakoshh, line 392: So here we could unpack the `definition` (or change the parameter to contain the `GlobalId` itself, which the caller already checked), and look up by ID. Would that make that redundant?

aakoshh: I added an extra assertion in https://github.com/noir-lang/noir/pull/6282/commits/ce3fa081e16ef7888f02e269d1a0951a6b410674 that the result from this `find` matches the input `definition`, so it seems reasonable to me that we should look up by ID and probably don't need the local+ident based search.

If we don't find the global by local+ident, we end up pushing another item into `definitions` with the same content as already exists. If we instead looked up by the `GlobalId` contained in `definition`, we could reuse the same `definition_id`.

aakoshh: Ah, just saw https://github.com/noir-lang/noir/issues/5001#issuecomment-2180721837 saying we should be able to get rid of this check.

jfecher (approved):
👍

PR #6287, opened 2024-10-11
fix: Reject invalid expression with in CLI parser

### LINKED ISSUE #5560: Nargo panics if you specify an expression-width < 3 during compile
### Aim

When compiling a noir program and specifying the expression-width (amount of witnesses in an expression), the compiler expects a width > 2. This is most likely the correct behaviour, but nargo panics with the following message:
```
Message:  assertion failed: width > 2
Location: acvm-repo/acvm/src/compiler/transformers/csat.rs:27

```

### Expected Behavior

The program should terminate and tell me that nargo expects an expression-width > 2, instead of a panic and indicating that there is a bug.

### Bug

The problem is here:
```
 acvm-repo/acvm/src/compiler/transformers/csat.rs:27
```

Here the CSat Solver asserts that the expression-width is > 2. It should return an Result instead telling the user to provide an expression-width > 2

### To Reproduce

1. ```nargo new wrong_expr_width && cd wrong_expr_width```
2. ```nargo compile --expression-width 1```

### Project Impact

None

### Impact Context

_No response_

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Installation Method

Binary (`noirup` default)

### Nargo Version

noirc version = 0.30.0+af57471035e4fa7eaffa71693219df6d029dbcde

### NoirJS Version

_No response_

Maybe

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #5560

## Summary\*

Rejects an invalid `--expression-value` in the CLI, which would later cause a panic in `acvm::compiler::transformers::csat`. 

An example error message looks like:

```text
error: invalid value '1' for '--expression-width <EXPRESSION_WIDTH>': minimum value is 3
```

## Additional Context

The issue suggests that [CSatTransformer::new](https://github.com/noir-lang/noir/blob/ae87d287ab1fae0f999dfd0d1166fbddb927ba97/acvm-repo/acvm/src/compiler/transformers/csat.rs#L24-L27) should return a `Result` explaining the problem it has with `width`, rather than doing an `assert!` (without any message) and crashing the program. I agree, however looking at the chain of calls, none of the half-dozen functions we go through to reach this use `Result`, and the `CSatTransformer` is the only option we have. This suggests to me that this limitation is perhaps supposed to be well-known to the user, ie. it's not the case that one transformer has X limit and another has Y limit. 

For this reason I added a `pub const MIN_EXPRESSION_WIDTH: usize = 3;` to the `acvm::compiler` module and using that as a common value for the assertion as well as the validation in the CLI. Should the assumption of a single global value change, removing that will force us to update the validation logic as well. 

That said if you prefer going the `Result` route I'm not against it, it just seemed like an overkill for this single use case.
 
## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/nargo_cli/benches/criterion.rs
+++ b/tooling/nargo_cli/benches/criterion.rs
@@ -23,6 +23,8 @@ fn compile_program(test_program_dir: &Path) {
     cmd.arg("--program-dir").arg(test_program_dir);
     cmd.arg("compile");
     cmd.arg("--force");
+    cmd.arg("--expression-width");
+    cmd.arg("1");
     cmd.assert().success();
 }
 
--- a/tooling/nargo_cli/benches/criterion.rs
+++ b/tooling/nargo_cli/benches/criterion.rs
@@ -23,8 +23,6 @@ fn compile_program(test_program_dir: &Path) {
     cmd.arg("--program-dir").arg(test_program_dir);
     cmd.arg("compile");
     cmd.arg("--force");
-    cmd.arg("--expression-width");
-    cmd.arg("1");
     cmd.assert().success();
 }
 
--- a/tooling/nargo_cli/src/cli/mod.rs
+++ b/tooling/nargo_cli/src/cli/mod.rs
@@ -112,3 +112,20 @@ pub(crate) fn start_cli() -> eyre::Result<()> {
     println!("{markdown}");
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use clap::Parser;
+    /// Test that parsing an invalid option doesn't
+    #[test]
+    fn test_parse_invalid_expression_width() {
+        let cmd = "nargo --program-dir . compile --expression-width 1";
+        let res = super::NargoCli::try_parse_from(cmd.split_ascii_whitespace());
+
+        let err = res.expect_err("should fail because of invalid width");
+        assert!(err.to_string().contains("expression-width"));
+        assert!(err
+            .to_string()
+            .contains(acvm::compiler::MIN_EXPRESSION_WIDTH.to_string().as_str()));
+    }
+}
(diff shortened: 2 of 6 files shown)


tooling/nargo_cli/benches/criterion.rs
```diff
@@ -23,6 +23,8 @@ fn compile_program(test_program_dir: &Path) {
     cmd.arg("--program-dir").arg(test_program_dir);
     cmd.arg("compile");
     cmd.arg("--force");
+    cmd.arg("--expression-width");
+    cmd.arg("1");
```
asterite, line 27: What does this "1" mean here?

tooling/nargo_cli/src/cli/mod.rs
```diff
@@ -112,3 +112,20 @@ pub(crate) fn start_cli() -> eyre::Result<()> {
     println!("{markdown}");
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use clap::Parser;
+    /// Test that parsing an invalid option doesn't
```
asterite, line 119: Unfinished sentence :-)

We could either finish it or remove this comment, your choice.

tooling/nargo_cli/benches/criterion.rs
TomAFrench: I don't think this should be here right?

TomAFrench (commented):
LGTM pending removal of changes which break benchmarks.

tooling/nargo_cli/benches/criterion.rs
aakoshh: Good cat, I think this is what I used to trigger the error and accidentally checked it in, thanks!

tooling/nargo_cli/src/cli/mod.rs
aakoshh: Thanks, removed. I wanted to say "doesn't cause a panic" but then realised that it never did, froze mid sentence.

TomAFrench (approved):

PR #6291, opened 2024-10-15
fix: Display function name and body when inlining recursion limit hit

### LINKED ISSUE #4828: Compiler panic when you make a recursive call which does not terminate
### Aim

We (https://github.com/FuzzingLabs) found that the compiler panics when you make a recursive call without condition, even if a return call makes you go out of the loop.a 

### Expected Behavior

No panic should occur.

### Bug

When a function makes a recursive call after a return statement, the compiler considers the recursive call, which can lead to a panic crash as it treats it as an infinite loop. As you can see in the code example, the **`rec1`** function has a return statement before the recursive call, preventing any potential looping. However, the compiler still perceives it as an infinite recursive function. 

**`rec2`** is another function that might exhibit the same behavior, but in this case, the compiler doesn’t panic. This seems to be because the compiler never enters the else block. The same phenomenon occurs with **`rec3`**.

The issue occurs with the latest compiler version on the master branch. The error message is as follows:

```
Message:  Attempted to recur more than 1000 times during function inlining.
Location: compiler/noirc_evaluator/src/ssa/opt/inlining.rs:167
```

### To Reproduce

```
/* Function with a recursive call which should end by returning 1, but the compiler treats it as an infinite loop. */
fn rec1() -> u8 {
    if true {
        1
    }
    rec1()
}

/* Function with a recursive call under a condition, so the compiler handles it correctly even though the result should be the same as rec1. */
fn rec2() -> u8 {
    if true {
        1
    } else {
        rec2()
    }
}

/* Another function with the same behavior that doesn't cause the compiler to panic. */
fn rec3() -> u8 {
    if false {
        rec3()
    }
    1
}

fn main() {
    rec1();
    rec2();
    rec3();
}
```

### Project Impact

None

### Impact Context

_No response_

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Installation Method

None

### Nargo Version

_No response_

### NoirJS Version

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #4828 

## Summary\*

Given a program like this:
```
fn main() {
  main();
}
```

Changed this panic message:

```
Attempted to recur more than 1000 times during function inlining.
```

Into this:
```
Attempted to recur more than 1000 times during inlining function 'main': acir(inline) fn main f0 {
  b0():
    call f0()
    return 
}
```

I included the ACIR to help figure out which function is the culprit, in case the name is not enough.

## Additional Context

I added a test to show that there is no compilation error or warning for the example program. It would be nice to issue a warning about unconditional recursion like Rust does. I'll try to do that in a followup PR.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/tests.rs
```diff
@@ -3389,3 +3389,14 @@ fn arithmetic_generics_rounding_fail() {
     let errors = get_program_errors(src);
     assert_eq!(errors.len(), 1);
 }
+
+#[test]
+fn unconditional_recursion() {
```
asterite, line 3394: The tests here (in frontend/tests) don't run the full compiler, just the frontend. So all the SSA code doesn't run here. That's why this doesn't error... but then maybe this test shouldn't be here (maybe it should be in `test_programs` in `compile_failure)`.

asterite (commented):
Looks great! I left a comment about a test, but once that's moved to somewhere else we can merge this.

compiler/noirc_frontend/src/tests.rs
aakoshh: Thanks, I added this because I didn't know about the SSA at all. Then I added tests where they are executed, and figured out where to do the other test above in `inlining.rs`. I thought I would leave this here as a demonstration that this thing is _not_ caught at the moment.

However in https://github.com/noir-lang/noir/pull/6292 I replaced it with failure and successful scenarios after implementing this error detection in the frontend, so we can catch the obvious case early. 

I don't have an opinion on whether it should live here or in the `compile_failure` directory 🤷 

FWIW I collected these places where I look for these kinds of tests, I thought it would be a good addition to the README:

```
Tests:
- tooling/nargo_cli/build.rs                         <- run tests in test_programs
- tooling/nargo_cli/tests/stdlib-builtin.rs          <- run tests in std
- tooling/nargo_cli/tests/stdlib-props.rs            <- fuzz testing code snippets
- tooling/nargo_fmt/tests/execute.rs                 <- expected formatter outcomes
- compiler/noirc_frontend/src/hir/comptime/tests.rs  <- interpreted program snippets
- compiler/noirc_frontend/src/lexer/lexer.rs         <- lexer unit tests
- compiler/noirc_frontend/src/parser/parser/*.rs     <- parser unit tests
- compiler/noirc_frontend/src/tests.rs               <- compilation
- compiler/noirc_frontend/src/tests/*                <- compilation by feature
```

asterite: I think leaving it here might be misleading (when quickly reading it it might seem that infinite recursion does not lead to an error).

asterite:
I'm thinking, that, if possible, it would be nice to show part of the stack trace that led to this infinite recursion. For example if you have a mutual recursion of `foo` and `bar`, seeing `foo` and `bar` would make it clearer that that's the issue. That said, I think the code right now doesn't track which functions it goes into, so this could be part of a separate PR.

asterite (approved):
Looks good!

PR #6292, opened 2024-10-15
feat: Reject programs with unconditional recursion

aakoshh:
# Description

## Problem\*

Related to #4828 
Followup to https://github.com/noir-lang/noir/pull/6291

## Summary\*

Added a new `ResolutionError::UnconditionalRecursion` which should result in a warning or compilation error where the function body recurses into the defining function with no alternative way to return.

## Documentation\*

Check one:

# PR Checklist\*


asterite:
I wonder if this error should be produced on the SSA side instead of panicking when the "too much recursion" is detected. That way it would detect mutual recursion too.

aakoshh:
You mean to change the return value into `Result`? I didn’t investigate the impact, or if it has any other panics that could be covered. There was another PR with a similar case where I changed the panic message and added some CLI side filtering, but didn’t want to cause too much churn by changing the entire call stack. 

I’ll defer to @TomAFrench

aakoshh:
Ah, you mean that instead of a warning, we could rely purely on the SSA.

I did it this way so it can be a warning, like the unused variables. Rust has this warning too, so why not. I didn't implement mutual recursion because Rust doesn't seem to have it either, but in theory I think we could: if we detect a cycle where none of the code paths can return without calling any of the `FuncId` we collected along the way, then it could reject that case too.

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1692,4 +1709,78 @@ impl<'context> Elaborator<'context> {
...
+        if !self.can_return_without_recursing(func_id, body_id) {
+            self.push_err(CompilationError::ResolverError(ResolverError::UnconditionalRecursion {
+                name: func_name,
+                span: func_span,
+            }));
+        }
+    }
+
+    /// Check if an expression will end up calling a specific function.
+    fn can_return_without_recursing(&self, func_id: FuncId, expr_id: ExprId) -> bool {
```
TomAFrench, line 1730: I think this breaks in the scenario where you pass the current function as a function pointer into a call to a separate function. This other function could conditionally recuse but we treat it as unconditional.

There would likely be other cases with lambdas.

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1692,4 +1709,78 @@ impl<'context> Elaborator<'context> {
                 _ => true,
             })
     }
+
+    /// Check that a recursive function *can* return without endlessly calling itself.
+    fn check_for_unbounded_recursion(
```
TomAFrench, line 1714: We have a `lints.rs` module which this could go in.

TomAFrench (commented):
It seems like we're going to end up with false positives so we may want to soften this warning to "potential unconditional recursion" if we're not going to track down all of these?

compiler/noirc_frontend/src/elaborator/mod.rs
aakoshh: I think the reason I didn't put it in there was because all lints are arranged under the `fn run_function_lints(&mut self, func: &FuncMeta, modifiers: &FunctionModifiers) {` function which didn't have all the inputs needed in order to call this check, notably it doesn't have the `body_id: ExprId` parameter and the `hir_func` I passed isn't available yet. I suppose after line 499 it will somehow be associated with the function definition, but it felt like lints were intended to do something else. So I followed the example of [type_check_function_body](https://github.com/noir-lang/noir/blob/4a0a672e3658da36eb73d0d4d542208d34e1c7bf/compiler/noirc_frontend/src/elaborator/mod.rs#L458) of passing the expression.

compiler/noirc_frontend/src/elaborator/mod.rs
aakoshh: Hm, do you have an example? We can add it to the tests. 

I'm not sure I got it right, but if it's a call to another function, it will immediately get a pass, ie. a false negative. For example mutual recursion is not rejected. 

I see I did `HirExpression::Lambda(e) => check(e.body)`, I suppose that might be wrong as it checks more than `HirExpression::Call` does, which doesn't follow into the body 🤔

aakoshh: Added an exemption for lambda bodies. Rust doesn't seem to check them either.

aakoshh:
We may want to soften this warning to "potential unconditional recursion". 

I agree that we can tweak the error message. For the record Rust says:
```
function cannot return without recursing
a `loop` may express intention better if this is on purpose
`#[warn(unconditional_recursion)]` on by default
```

compiler/noirc_frontend/src/elaborator/mod.rs
aakoshh: That said the [call to deprecated_function](https://github.com/noir-lang/noir/blob/00c186bfaea113799a721d08ce348f703d5153fb/compiler/noirc_frontend/src/elaborator/types.rs#L1441-L1449) shows that's not the only place it's called from, and I agree `lints.rs` would be a nice place for this, I'll move it.

aakoshh: Done.

TomAFrench (approved):

PR #6296, opened 2024-10-17
feat: Warn about private types leaking in public functions and struct fields

### LINKED ISSUE #6248: Warn on items more private than others
### Problem

For example in this Rust code:

```rust
struct Foo {}

pub fn bar() -> Foo {
    Foo {}
}
```

Rust says that `Foo` is more private than `bar`.

We should do the same in Noir and also in similar cases (for example when defining a struct with fields that are more visible than the struct).

### Happy Case

Noir warns/errors in the above snippet and similar ones.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #6248

## Summary\*

Issue warnings in more cases when a public item refers to private types:
* public fields of a public struct fields using private types
* a public function taking or returning private types

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/tests/visibility.rs
```diff
@@ -28,29 +28,38 @@ fn errors_once_on_unused_import_that_is_not_accessible() {
         ))
     ));
 }
+
+fn assert_type_visibility_error(src: &str, private_typ: &str, public_item: &str) {
```
asterite, line 32: Should this be named something like `assert_type_is_more_private_than_item_error`?

asterite (approved):
Looks great! I left a comment about a name.

compiler/noirc_frontend/src/tests/visibility.rs
aakoshh: Okay, renamed. I'm notoriously bad at naming 🙏

PR #6300, opened 2024-10-18
feat: new formatter

### LINKED ISSUE #3312: nargo fmt: inconsistent formatting of method chains
### Aim

```rust
object.method1().method2().method3().method4().method5();
```

### Expected Behavior

```rust
object
    .method1()
    .method2()
    .method3()
    .method4()
    .method5();
```

### Bug

The code formatter doesn't seem to handle the formatting of method chains consistently or effectively, possibly leading to readability and maintainability issues. Method chains are essential for writing fluent and clean code, especially in functional programming paradigms, and thus, it's crucial for the formatter to support them well.

### To Reproduce

1.
2.
3.
4.

### Installation Method

None

### Nargo Version

_No response_

### Additional Context

_No response_

No

### Support Needs

_No response_

### LINKED ISSUE #3945: nargo fmt: inconsistent formatting with arrays
### Aim

Run fmt on
```rust
dep::std::hash::pedersen_hash_with_separator([
    contract_address.to_field(),
    storage_slot
], GENERATOR_INDEX__PUBLIC_LEAF_INDEX)
```

### Expected Behavior

```rust
dep::std::hash::pedersen_hash_with_separator([
    contract_address.to_field(),
    storage_slot
], GENERATOR_INDEX__PUBLIC_LEAF_INDEX)
```

Or 

```rust
dep::std::hash::pedersen_hash_with_separator(
    [
        contract_address.to_field(),
        storage_slot
    ], 
    GENERATOR_INDEX__PUBLIC_LEAF_INDEX
)
```

### Bug

The formatted code looks like

```rust
dep::std::hash::pedersen_hash_with_separator(
    [
    contract_address.to_field(),
    storage_slot
],
    GENERATOR_INDEX__PUBLIC_LEAF_INDEX
)
```

### To Reproduce

1. Run nargo fmt
2.
3.
4.

### Installation Method

None

### Nargo Version

_No response_

### Additional Context

Saw the issue as part of https://github.com/AztecProtocol/aztec-packages/pull/3803/files/c6f4b7b9b9efe9a118e57aec0f28219523812563

No

### Support Needs

_No response_

### LINKED ISSUE #4462: Nargo fmt incorrectly formats nested structures
### Aim

Format this code:
```rust
    assert_eq(
        result, [
            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1,
            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2
        ]
    );
```

### Expected Behavior

Do nothing

### Bug

`nargo fmt` removes one level of tabs incorrectly
```rust
    assert_eq(
        result, [
        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1,
        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2
    ]
    );
```

### To Reproduce

1. Copy this code part
2. Run nargo fmt

### Project Impact

Nice-to-have

### Impact Context

Very low pri

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Installation Method

Binary (`noirup` default)

### Nargo Version

noirc version = 0.23.0+5be9f9d7e2f39ca228df10e5a530474af0331704

### NoirJS Version

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #4558: `nargo fmt` should remove unnecessary `{}` in nested imports
### Aim

See https://github.com/noir-lang/noir/pull/4529#discussion_r1525173925 for an example which is not caught by the current formatter:

```rs
use foo::{
    bar::baz,
    qux::{quux},  // unnecessary but not removed
};
```

### Expected Behavior

The formatter should remove `{}` on single imports, even if the import is nested

### Bug

The formatter does not do so

### To Reproduce

1.
2.
3.
4.

### Project Impact

None

### Impact Context

_No response_

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Installation Method

None

### Nargo Version

_No response_

### NoirJS Version

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #4766: Spurious trailing space when formatting arrays
### Aim

Running `nargo fmt` in `tests::format_idempotent_array` adds a spurious trailing space after `10`:

```bash
Differences (-left|+right):
     [
         // hello!
         1,
         // asd
-        10
+        10 
           ^ this is the trailing space
```

See https://github.com/noir-lang/noir/pull/4765 for the test that triggered this behaviour.

### Expected Behavior

The array is successfully formatted without any trailing spaces.

### Bug

Trailing space after `10` 

### To Reproduce

1. See failure [here](https://github.com/noir-lang/noir/actions/runs/8621169588/job/23629569206?pr=4765#step:6:137) or in the PR above
2.
3.
4.

### Project Impact

Nice-to-have

### Impact Context

- The idempotence test is designed to detect divergence, e.g. from trailing spaces such as this one. Repeatedly running `nargo fmt` with this type of bug can produce an unlimited number of trailing spaces.
- Without idempotence, we can't clearly answer the question "Is this file formatted?"

### Workaround

Yes

### Workaround Description

Manually delete trailing spaces

### Additional Context

_No response_

### Installation Method

None

### Nargo Version

_No response_

### NoirJS Version

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #4767: Inconsistent pre-comment space when formatting tuples
### Aim

Running `nargo fmt` in `tests::format_idempotent_tuple` only adds a space between the expression and comment when run twice:

```bash
Differences (-left|+right):
     (//
         1,);
 
     (// 1
-        1,// 2,
+        1, // 2,
           ^ space only added when running twice
```

### Expected Behavior

The tuple is successfully formatted with the space between the expression and comment when run the first time.

### Bug

Space before comment only when running `nargo fmt` twice

### To Reproduce

1. See failure [here](https://github.com/noir-lang/noir/actions/runs/8621169588/job/23629569400?pr=4765#step:6:146) or in the PR above
2.
3.
4.

### Project Impact

Nice-to-have

### Impact Context

- The idempotence test is designed to detect divergence, e.g. from trailing spaces such as this one. Repeatedly running `nargo fmt` with this type of bug can produce an unlimited number of trailing spaces.
- Without idempotence, we can't clearly answer the question "Is this file formatted?"

### Workaround

Yes

### Workaround Description

Manually add space after running `nargo fmt`

### Additional Context

_No response_

### Installation Method

None

### Nargo Version

_No response_

### NoirJS Version

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #4768: Spurious newline when formatting tuples
### Aim

Running `nargo fmt` on the expected output of `tests::format_idempotent_tuple` adds a spurious newline:

```bash
Differences (-left|+right):
     (/*a*/
-        1/*b*/, /*c*/ 2/*d*/, /*c*/ 2/*d*/, /*e*/ 3/*f*/);
+        1/*b*/,
+/*c*/ 2/*d*/, /*c*/ 2/*d*/, /*e*/ 3/*f*/);
 ^ newline only added when running twice
```

### Expected Behavior

The tuple is successfully formatted without a spurious newline

### Bug

Spurious newline (and missing indentation) when formatting a tuple with block comments

### To Reproduce

1. See failure [here](https://github.com/noir-lang/noir/actions/runs/8621169588/job/23629569400?pr=4765#step:6:157) or in the PR above
2.
3.
4.

### Project Impact

Nice-to-have

### Impact Context

- The idempotence test is designed to detect divergence, e.g. from trailing spaces such as this one. Repeatedly running `nargo fmt` with this type of bug can produce an unlimited number of trailing spaces.
- Without idempotence, we can't clearly answer the question "Is this file formatted?"

### Workaround

Yes

### Workaround Description

Manually remove spurious newline after running `nargo fmt`

### Additional Context

_No response_

### Installation Method

None

### Nargo Version

_No response_

### NoirJS Version

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #5281: Poor formatting of comptime blocks
Currently the program below

```rust
fn main() {
    comptime {
        println("hello world");
    }
}
```

is formatted into

```rust
fn main() {
    comptime
    {
        println("hello world");
    }
}
```

Ideally the first curly brace would be on the same line as `comptime`

asterite:
# Description

## Problem

Resolves #5281
Resolves #4768
Resolves #4767
Resolves #4766
Resolves #4558
Resolves #4462
Resolves #3945
Resolves #3312

## Summary

I thought about trying to extend the existing formatter to format more code, but I couldn't understand it very well: it partially implemented things, and it lacked comments and some explanation of how it works. I think some chunks might have been copied from the Rust formatter. I also took a look into it but though it might be too complex.

[I wrote a formatter in the past for Crystal](https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/tools/formatter.cr) with a technique that I saw was used in the Java formatter written for Eclipse. The idea is to traverse the AST together with a Lexer, writing things along the way, bumping into comments and formatting those, etc. It works pretty well! But that's not enough for the Noir formatter because here we also want to automatically wrap long lines (Crystal's formatter doesn't do that). That part (wrapping) is partially based on [this excellent blog post](https://yorickpeterse.com/articles/how-to-write-a-code-formatter/).

Benefits:
- All code kinds are formatted. For example previously traits weren't formatted.
- Comments are never lost.
- Lambdas appearing as the last argument of a call are formatted nicely (this was maybe the most trickier part to get right).
- I believe the new code ends up being simpler than before, even though it's (slightly) longer (previously is was 2138 lines, now it's 6139, though 2608 lines are tests, so it ends up being 3531 lines, but this new formatter does many things the old one didn't). I tried to comment and document it well.
- Adding new formatting rules, new configurations, or enabling formatting of new syntax should be relatively easy to do.
- There are lots and lots of tests for each of the different scenarios. The existing overall formatter tests were kept, but with unit tests it's easier to see how edge cases are formatted.

[Here's Aztec-Packages formatted with the new formatter](https://github.com/AztecProtocol/aztec-packages/pull/9292/files):
- Max line width seems to be respected more (making it more readable)
- `unsafe { ... }` and `comptime { ... }` expressions will be inlined if possible (shortening the code and making it more readable)

## Additional Context

Some things are subjective. For example Rust will put a function `where` clause in a separate line, with each trait bound on its own line. The new formatter does that too. But Rust will sometimes put an `impl` where clause on the same line. It's not clear to me why. I chose to always put `where` clauses on a separate line, but this could easily be changed if we wanted to.

## Documentation

Check one:

# PR Checklist


asterite:
image

I got a bit scared looking at those numbers, thinking that I couldn't have possibly written that much code. Then I realized that's also counting all the diff to the newly formatted Noir files 😅

tooling/nargo_fmt/src/formatter.rs
```diff
@@ -0,0 +1,303 @@
...
+
+    /// Writes whatever is in the given span relative to the file's source that's being formatted.
+    fn write_source_span(&mut self, span: Span) {
+        self.write(&self.source[span.start() as usize..span.end() as usize]);
+    }
+
+    /// Writes the current indentation to the buffer, but only if the buffer
+    /// does not end with a space (otherwise we'd be indenting too much).
+    fn write_indentation(&mut self) {
+        if self.buffer.ends_with(' ') {
```
aakoshh, line 229: I don't fully understand why only space is checked. Shouldn't indentation go only after a new line?

tooling/nargo_fmt/src/formatter.rs
```diff
@@ -0,0 +1,303 @@
...
+        while self.token_span.start() < span.end() {
+            self.bump();
+        }
+    }
+
+    /// Writes a string to the buffer.
+    /// This is the only method that directly appends to the buffer and keeps
+    /// track of the current line width.
+    /// If adding new methods that write to the buffer, always use this method
+    /// instead of directly appending to the buffer.
```
aakoshh, line 256: Would it make sense to encapsulate the buffer and the current line width in a separate structure to enforce this?

tooling/nargo_fmt/src/formatter.rs
```diff
@@ -0,0 +1,303 @@
...
+    source: &'a str,
+    lexer: Lexer<'a>,
+    token: Token,
+    token_span: Span,
+
+    /// The current indentation level.
+    /// We let it be negative because in some cases we just want to decrease indentation
+    /// to compensate an indentation that will come later that we don't need, and we
+    /// don't want to panic when reaching those negative values.
+    indentation: i32,
```
aakoshh, line 47:
```suggestion
    /// We allow it to be negative because in some cases we just want to decrease indentation
    /// to preemptively cancel out an indentation that will come later which we don't want to take effect,
    /// and we don't want to panic when reaching those negative values.
    indentation: i32,
```
When I read "come later that we don't need" I thought it's something like the conditional indentation in the blog post, which either happens or not based on whether something is wrapped, and I thought it might leave some dangling negative unindent. It took me a long time to conclude that you must mean something like `-1; +1`, and not `-1; if whatever() { +1 }`. Not sure if this wording conveys that better, but wanted to highlight that I found the comment not obvious (but super helpful nevertheless!).

tooling/nargo_fmt/src/formatter/chunks.rs
```diff
@@ -0,0 +1,776 @@
...
+//! that we (almost always) never exceed the configurable maximum line width.
+//!
+//! It's heavily inspired by this excellent blog post:
+//!
+//! https://yorickpeterse.com/articles/how-to-write-a-code-formatter/
+//!
+//! However, some changes were introduces to handle comments and other particularities of Noir.
+use super::Formatter;
+
+/// A text chunk. It precomputes the text width and wether it has newlines.
```
aakoshh, line 11:
```suggestion
/// A text chunk. It precomputes the text width and whether it has newlines.
```

tooling/nargo_fmt/src/formatter/chunks.rs
```diff
@@ -0,0 +1,776 @@
...
+            | Chunk::TrailingComma
+            | Chunk::PushIndentation
+            | Chunk::PopIndentation => 0,
+        }
+    }
+
+    /// Computes the width of this chunk considering it's inside an ExpressionList.
+    /// The only thing that changes here compared to `width` is that a LambdaAsLastExpressionInList's
+    /// width is considered to be only the first line, so we can avoid splitting the entire call
+    /// arguments into separate lins.
```
aakoshh, line 82:
```suggestion
    /// arguments into separate lines.
```
On a different note, perhaps an example of the `LambdaAsLastExpressionInList` would be helpful.

tooling/nargo_fmt/src/formatter/chunks.rs
```diff
@@ -0,0 +1,776 @@
...
+    ///   4, 5
+    /// ]
+    ///
+    /// but if one of the elements is long, each one will go in a separate line:
+    ///
+    /// [
+    ///     1,
+    ///     1234567890123,
+    ///     3
+    /// ]
```
aakoshh, line 144:
```suggestion
    /// ```
    /// [
    ///     1,
    ///     1234567890123,
    ///     3
    /// ]
    /// ```
```

Without the surrounding ``` it appears inline as a hyperlink, at least in the Rust Analyser tooltips.
Same goes for the other array.

tooling/nargo_fmt/src/formatter/chunks.rs
```diff
@@ -0,0 +1,776 @@
...
+                        } else {
+                            chunk.width_inside_an_expression_list()
+                        }
+                    } else {
+                        chunk.width_inside_an_expression_list()
+                    }
+                })
+                .sum()
+        } else {
+            self.chunks.iter().map(|chunk| chunk.width_inside_an_expression_list()).sum()
```
aakoshh, line 322: It's not clear to me why the special treatment of `GroupKind::ExpressionList` only applies to the `is_method_call()` case 🤔

tooling/nargo_fmt/src/formatter/chunks.rs
```diff
@@ -0,0 +1,776 @@
...
+            match chunk {
+                Chunk::Text(chunk) => group.text(chunk),
+                Chunk::TrailingComma => {
+                    // If there's a trailing comma after a group, append the text to that group
+                    // so that it glues with the last text present there (if any)
+                    if let Some(Chunk::Group(group)) = group.chunks.last_mut() {
+                        group.add_trailing_comma_to_last_text();
+                    } else {
+                        group.text(TextChunk::new(",".to_string()));
+                    }
```
aakoshh, line 394:
```suggestion
                    group.add_trailing_comma_to_last_text();
```
At least I think it's the same thing 👀

tooling/nargo_fmt/src/formatter/statement.rs
```diff
@@ -0,0 +1,680 @@
...
+            ConstrainKind::AssertEq => Keyword::AssertEq,
+            ConstrainKind::Constrain => {
+                unreachable!("constrain always produces an error, and the formatter doesn't run when there are errors")
+            }
+        };
+
+        group.text(self.chunk(|formatter| {
+            formatter.write_keyword(keyword);
+            formatter.write_left_paren();
+        }));
```
aakoshh, line 157: How do you not forget whether to call things in `self.chunk` or directly like `self.write_left_paren()`?

tooling/nargo_fmt/src/formatter.rs
asterite: Good point! I think this ended up being called twice in some cases (because of nested calls where we can't know if some nested formatting ended up writing a final indentation) so just checking that it doesn't end with a space was enough. But your idea is better (I just tried it and nothing breaks :-)). I'll push a commit soon to fix this.

tooling/nargo_fmt/src/formatter.rs
asterite: Probably yes. I'll try it after addressing the other comments.

tooling/nargo_fmt/src/formatter/statement.rs
aakoshh: Do you see a way to enforce when the formatter can be written to?

tooling/nargo_fmt/src/formatter/chunks.rs
asterite: Good idea, I'll add an example soon.

tooling/nargo_fmt/src/formatter/chunks.rs
asterite: Good point! I think I wrote this before eventually writing the other method.

tooling/nargo_fmt/src/formatter/statement.rs
asterite: > How do you not forget whether to call things in self.chunk or directly like self.write_left_paren()?

It seems that Rust gives an error if you try to use `self` there (it says "cannot borrow `*self` as mutable because previous closure requires unique access second borrow occurs here")

tooling/nargo_fmt/src/formatter/chunks.rs
asterite: I just tried removing the `is_method_call` branch and all tests pass, no change is produced in the standard library, test programs or Aztec-Packages. Go apparently there's no need for this special case. I think in my head I thought there would be a difference 😅

tooling/nargo_fmt/src/formatter.rs
asterite: Done! Thank you, it ended up making the code much clearer.

tooling/nargo_fmt/src/formatter/statement.rs
aakoshh: Sorry, my question wasn't clear. I didn't mean how you remember not to call `self.write_left_paren()` inside the lambda passed to `self.chunk` (which is where the "cannot borrow as mutable" error comes up), but rather how you remember to do this:

```
        group.text(self.chunk(|formatter| {
            formatter.write_keyword(keyword);
            formatter.write_left_paren();
        }));
```
instead of:
```
          self.write_keyword(keyword);
          self.write_left_paren();
```

The compiler wouldn't complain if `group.whatever(...)` is mixed with `self.write_something()`. This is where I find the fact that some `format_*` methods return a `ChunkGroup` while others take a mutable reference to them. 

I was wondering for example if instead of `Formatter::write_*` we could have a `Writer` that is only passed to the lambda by `Formatter::chunk`, so in a `format_*` method we don't have to wonder if we have to call `write_*`, or call `chunk`, because to write we _must_ call `chunk`. But I don't know if this is realistic, just wanted to point out that sometimes I only see that `.chunk` needs to be used because `ChunkGroup` is returned, so calling `self.write` couldn't possibly make sense, but other times it's not returned and it might be harder to see what the right thing to do is.

asterite: Ah, I understand now.

Right now the formatter has two mixed modes: format without accounting for the maximum width (this is done for the first line of a module declaration, trait and impl) or accounting for it (statements and expressions). If the maximum width is not important (or unlikely to be reached) we use `self.write_keyword(...)` as that writes directly to the buffer. Otherwise we use groups and chunks because formatting those will check for the maximum width.

Maybe everything would be more consistent if it was always composed of groups and chunks, but it would make the code slightly more complex. We could try, though! (not sure if this in PR or as a follow-up)

aakoshh: Aha, thanks for that explanation. It would be nice to be able to express this distinction in code, that some formatters require a writer that doesn't deal with chunks, while others need to construct chunks before the real writing begins, but I see that it's a bit intertwined, like `write_token` also calls `skip_comments_and_whitespace`, and I don't have a good idea now how to do it differently. I thought about using a phantom type, but don't think it would work with `&mut self`.

asterite: It's a good point. Sometimes I would write a test to see it fail, then try to implement something and it was still failing... because I was writing directly to the buffer instead of creating a chunk and appending it to an existing group. Luckily the test caught that. I think we could refactor the code to make it more "type safe" in that regards, but with the many tests we have maybe it's good enough: if you use it the wrong way a test will fail.

aakoshh: Here's an idea that is might be relatively easy to implement without radical refactoring: https://gist.github.com/aakoshh/a813dd5291fc48645d0f07b7a894b8b0

It's based on wrapping the mutable reference and implementing formatters on the wrappers.

noir_stdlib/src/option.nr
```diff
@@ -112,19 +116,31 @@ impl<T> Option<T> {
...
     }
 
     /// If self is Some, return self. Otherwise, return `default()`.
     pub fn or_else<Env>(self, default: fn[Env]() -> Self) -> Self {
-        if self._is_some { self } else { default() }
+        if self._is_some {
+            self
+        } else {
+            default()
+        }
```
TomAFrench, line 132: This feels like we're setting `single_line_if_else_max_width` too low by default. I'm imagining that we're measuring the width differently than with the previous formatter.

tooling/nargo_fmt/src/formatter/chunks.rs
TomAFrench, line 1: nit but it would be nice to separate these modules which are language-agnostic components of the compiler from those which define how to format structs, visibilities, etc.

tooling/nargo_fmt/tests/expected/tuple.nr
```diff
@@ -1,32 +1,53 @@
...
+        // 1
+        1,
+        // 2,
+        2,
+    );
 
-    (/*1*/ 1, /*2*/ 2);
+    ( /*1*/ 1, /*2*/  2);
 
     // FIXME:
```
TomAFrench, line 33:
```suggestion
```

Seems to be fixed now

tooling/nargo_fmt/tests/input/impl.nr
```diff
@@ -14,8 +14,6 @@ impl MyType {
     fn method(self: Self) {}
 
     fn method(mut self: Self) {}
-
-    fn method(&mut self: Self) {}
```
TomAFrench, line 18: why was this removed?

TomAFrench (commented):
Looks good, very excited to have the formatter be managable.

noir_stdlib/src/option.nr
asterite: Oh, that's intended. `if` as a statement is always formatted in multiple lines, while `if` as an expression can be inlined. It's the same in Rust if you try it.

I _think_ the reason is that if you have code like this:

```rust
if 1 {
    2 
} else {
    3
}
(4, 5)
```

then the formatter would change it to this:

```rust
if 1 { 2 } else { 3 } (4, 5)
```

and it looks like a method call on that if. I think we also had a similar bug where the formatter would do that, and it was because `if` was always parsed as an expression, never as a statement. Once we changed that it got fixed.

tooling/nargo_fmt/tests/input/impl.nr
asterite: Ah, good catch! I don't know how you find these things in the giant diff, but it's great 😄 It seems that never parsed, I can't remember if the old parser had the same issue. This meant that the file was not formatted, and the test passed because input and expected were the same. I captured a personal note to work on this separately, but I'll work on it now because I think it should be easy to fix.

asterite: I just tried it with the chumsky parser (I checked out one commit before the handwritten parser was merged) and this program:

```noir
struct Foo;

impl Foo {
    fn foo(&mut self: Self) {

    }
}
```

gives this error:

```noir
error: Unexpected Colon, expected one of ), ,
  ┌─ src/main.nr:4:21
  │
4 │     fn foo(&mut self: Self) {
  │                     -
```

so it never worked.

Do you think we should fix it in this PR or in a separate one? I don't know what's the reason that it doesn't work, but I don't know if you do something like `fn foo(&mut self: SomeOtherType)`... should that compile?

aakoshh: I do this mistake in Rust all the time: shouldn't it be `fn foo(self: &mut Self)`?

asterite: Interesting. It seems the old parser (and I copied it for the new parser) parses `&mut self` into a pattern named `self` with type `&mut Self`. So I guess it's a shortcut to write that? And I just tried writing `fn(&mut self: Self)` in Rust and it also gives an error saying it's expecting `,` instead of `:`, so maybe there's nothing to do here.

I'm also starting to remember that I might have added that `&mut self: Self` line in that input file while I was coding the new parser, to make sure it parsed well... but I didn't realize that it was an error, and those errors don't show up in the formatter tests, oops! 😊

TomAFrench (approved):
LGTM, we've got better coverage of the syntax and actual testing so this is a direct upgrade over the previous formatter.

noir_stdlib/src/option.nr
TomAFrench: Ok, sounds good.

tooling/nargo_fmt/src/chunks.rs
```diff
@@ -0,0 +1,973 @@
...
+
+    pub(super) fn bump(&mut self) -> Token {
+        self.0.bump()
+    }
+}
+
+/// Treating a `ChunkFormatter` as a `Formatter` in read-only mode is always fine,
+/// and reduces some boilerplate.
+impl<'a, 'b> Deref for ChunkFormatter<'a, 'b> {
+    type Target = Formatter<'b>;
```
aakoshh, line 546: I'm surprised this isn't `'a`, since `ChunkFormatter<'a, 'b>` has a `'b` reference to a `Formatter<'a>` 🤔

asterite: Oh, good point, it should be `'a`. I don't know why it works, maybe the lifetimes are the same? I'll change it in a follow-up PR after I finish what I'm doing (or feel free to send a PR).

PR #6305, opened 2024-10-21
fix: Allow array map on empty arrays

jfecher:
# Description

## Problem\*

Resolves a private issue sent to me on slack

## Summary\*

Array map is a fairly old method written before we had `std::mem::zeroed`. Now that we have zeroed, we can allow mapping empty arrays since we can use `zeroed` to get the filler `U` value for the starting elements.

## Additional Context

While I was at it I included a small clarification to `reduce`'s docs that it requires a non-empty array

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/noir_stdlib/src/array/mod.nr
+++ b/noir_stdlib/src/array/mod.nr
@@ -43,8 +43,8 @@ impl<T, let N: u32> [T; N] {
     /// assert_eq(b, [2, 4, 6]);
     /// ```
     pub fn map<U, Env>(self, f: fn[Env](T) -> U) -> [U; N] {
-        let first_elem = crate::mem::zeroed();
-        let mut ret = [first_elem; N];
+        let uninitialized = crate::mem::zeroed();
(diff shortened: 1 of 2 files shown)


noir_stdlib/src/array/mod.nr
```diff
@@ -43,10 +43,10 @@ impl<T, let N: u32> [T; N] {
     /// assert_eq(b, [2, 4, 6]);
     /// ```
     pub fn map<U, Env>(self, f: fn[Env](T) -> U) -> [U; N] {
-        let first_elem = f(self[0]);
+        let first_elem = crate::mem::zeroed();
         let mut ret = [first_elem; N];
```
asterite, line 47: Maybe rename `first_elem` to something else, given that it's no longer the first element? Or maybe inline `crate::mem::zeroed()` in the array literal.

asterite (approved):

PR #6306, opened 2024-10-21
feat: support specifying generics on a struct when calling an associated function

### LINKED ISSUE #5584: Enable turbofish for impls
### Problem

Some code like this should compile:

```rust
struct MerkleTree<let N: u32> {
    root: Field,
    hasher: fn([Field; N]) -> Field,
}

impl<let N: u32> MerkleTree<N> {
    fn new(root: Field, hasher: fn([Field; N]) -> Field) -> Self {
        Self { root, hasher }
    }
}

#[test]
fn test_merkle_tree_membership() {
    let root = 0x1c59022dba1d97f63021cc5a23e4fe80f019465e0ccb54de9aa91935495354a3;
    let mt = MerkleTree::<2>::new(root, dep::std::hash::pedersen_hash);
}
```

But it doesn't, because `MerkleTree` doesn't accept turbofish

### Happy Case

The above code would compile and work :)

### Project Impact

None

### Impact Context

_No response_

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #6000: Support specifying generics on a type when calling an associated function
We currently do not allow applying a turbofish to a struct type, attempting to do so will result in a `UnsupportedTurbofishUsage` error

```noir
struct Foo<T> {
    x: T
}

impl <T> Foo<T> {
    fn new(x: T) -> Self {
        Foo { x }
    }
}

fn main() {
    let _ = Foo::<i32>::new(1);
}
```

We should implement support for the turbofish syntax in this situation.

asterite:
# Description

## Problem

Resolves #5584
Resolves #6000

## Summary

Also errors if you have something like this:

```noir
this_is_a_module::<i32>::foo();
                  ^^^^^
            turbofish not allowed on module `this_is_a_module`
```

## Additional Context

Note: this only works for structs. I captured a couple of issues to make this work for traits and aliases, but I'd prefer to do those in separate PRs.

## Documentation

Check one:

# PR Checklist


jfecher (changes requested):
Going to block this temporarily to give me some time to think about the approach some more.

compiler/noirc_frontend/src/elaborator/patterns.rs
```diff
@@ -468,54 +468,111 @@ impl<'context> Elaborator<'context> {
...
+            for error in trait_path_resolution.errors {
                 self.push_err(error);
             }
 
-            HirIdent {
-                location: Location::new(path.span, self.file),
-                id: self.interner.trait_method_id(trait_path_resolution.method.method_id),
-                impl_kind: ImplKind::TraitMethod(trait_path_resolution.method),
-            }
+            // TODO: GenericTypeInPath
```
jfecher, line 558: todo

jfecher (commented):
The generic_type_in_path part seems odd to me - it doesn't really say where that type is and I think it may be possible to have multiple types in the path. E.g. `foo::bar::Baz::<i32>::AssociatedType::<u32>::method::<i8>()`

Maybe it'd be possible to resolve to a module + type bindings? E.g. resolve to the struct module of `Baz` then handle any paths from there separately? Currently this would only be to methods I think since I don't think we support referring to associated types using the syntax I gave above.

asterite:
> The generic_type_in_path part seems odd to me

Makes sense. I thought of different possibilities to do this and that one was the less disruptive to the existing code.

That said... there's something that I always felt wasn't quite right: path resolution/lookup. There are two things here:
- imports (use statements)
- paths in code

Imports can't have generics in their path, while paths do. But: the code to resolve them is the same one.

So we have two choices:
1. Have two different code paths to solve imports and paths
2. Have a single code path, but returning a better type as the output

I'm leaning towards option 2, mainly to avoid code duplication. When resolving an import we can assert/panic when we get a resolution that shouldn't happen.

The better type I'm thinking is, instead of returning `ModuleDefId` we could return some other enum like this:

```rust
enum PathResolutionResult { // I think this name is already taken, but I'll come up with something else
    Module(ModuleId),
    Struct(StructId, Option<Vec<UnresolvedType>>), // The generics, if any (using Option to distinguish the `::<>` case)
    TypeAlias(TypeAliasId, Option<Vec<UnresolvedType>>),
    Trait(TraitId, Option<Vec<UnresolvedType>>),
    ModuleFunction(FuncId),
    StructFunction(StructId, Option<Vec<UnresolvedType>>, FuncId),
    TypeAliasFunction(TypeAliasId, Option<Vec<UnresolvedType>>, FuncId),
    TraitFunction(TraitId, Option<Vec<UnresolvedType>>, FuncId),
}
```

That way it's a bit more clear what a path resolved too... and if eventually we support that associated type line, we'll add a new member to the above enum.

What do you think?

jfecher:
> That way it's a bit more clear what a path resolved too... and if eventually we support that associated type line, we'll add a new member to the above enum.
> What do you think?

I like it - I agree path resolution leaves things to be desired currently. I was trying to avoid suggesting "let's rewrite path resolution entirely" but maybe it is needed :sweat_smile:. I'm fine with the two different code paths for imports and paths in code in general as well though. I think there are enough differences there that using one for both could (unverified) end up being more complicated in the end. E.g. Passing booleans and such to determine which cases you need to support and which you should error for. That's why I was initially thinking of the approach to break a path up into two parts: the prefix path which can only contain modules, functions, structs without turbofish and other things that can be imported, and the suffix path which may contain structs with turbofish, methods on them, and possibly associated types on those structs. Your call on which is cleaner, it can be a bit hard to guess without trying it first.

asterite:
I went down the route of having path and import resolution use the same code but return a better type.

It turns out, resolving a path gave two things:
- a `ModuleDefId`
- a namespace (`PerNs`)

The `ModuleDefId` part was used when looking up paths, the namespace part for imports. So, I changed that `ModuleDefId` to be `PathResolutionItem` in this case, with the enum I mentioned above. Then `GenericTypeInPath` is gone.

I think the code is much cleaner now. I still don't know if we should split the path/import lookup into two... maybe I can work on getting this working for traits and aliases, and then we can see if another refactor is worth it.

Let me know! 🙏  (this is ready for review again)

compiler/noirc_frontend/src/elaborator/scope.rs
```diff
@@ -26,16 +23,6 @@ type Scope = GenericScope<String, ResolverMeta>;
...
 impl<'context> Elaborator<'context> {
-    pub(super) fn lookup<T: TryFromModuleDefId>(&mut self, path: Path) -> Result<T, ResolverError> {
-        let span = path.span();
-        let id = self.resolve_path_or_error(path)?;
-        T::try_from(id).ok_or_else(|| ResolverError::Expected {
-            expected: T::description(),
-            got: id.as_str().to_owned(),
-            span,
-        })
-    }
```
asterite, line 37: Given that the lookup now returns `PathResolutionItem` instead of `ModuleDefId` I had two options:
1. Change `TryFromModuleDefId` to `TryFromPathResolutionItem`
2. Remove this method and inline it in the caller

I found that `lookup` was used 3 times, and with the recent "lookup type" change I think it went down to two... and inlining this and removing `TryFromModuleDefId` results in more or less the same amount of code, or maybe less.

But let me know if this isn't right and I'll bring it back.

jfecher (approved):
👍

asterite:
Sorry I had to force push, yesterday I continued working on this branch and forgot about it, then merged and got unrelated changes.

Issue #6271, opened 2024-10-10
Add a user-specified relative ordering for attributes


jfecher:
### Problem

Users cannot control which order their attributes run leading to issues such as #6143

### Happy Case

We add two new attributes `#[run_after(f)]` and `#[run_before(f)]` which accept a function `f` as an argument and ensure that the attribute function they're applied to is run after or before the attribute function f.

For example:
```noir
#[run_after(derive)]
comptime fn check_if_eq_is_implemented(s: StructDefinition) {
    let eq = quote { Eq }.as_trait_constraint();
    assert(s.as_type().implements(eq));
}
```
We could also have a default step so that attributes can run before other attributes by default. If `derive` used this for example then there'd be no need for other attributes to have to know to run after it:
```noir
#[run_before(_)]
comptime fn derive(s: StructDefinition, traits: [TraitDefinition]) -> Quoted { ... }
```

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

jfecher:
We decided to go with source ordering instead for greater predictability and easier debugging of attributes

PR #6319, opened 2024-10-22
feat: Add a relative ordering for attributes

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6271

## Summary\*

Wanted to get this before the discussion tomorrow for reference.

Adds a user-defined relative ordering for attributes which orders attributes by annotations placed by the library author of the attribute. This is done with the new `#[run_before(attr)]` and `#[run_after(attr)]` attributes which accept another attribute as an argument. Internally, the compiler builds a graph for attributes and topologically sorts it. Then the compiler sorts each attribute usage according to that topological ordering.

TODO:
- [ ] Detect graph for cycles, we can use the same method of our existing cycle detection for the dependency graph
- [ ] Some miscellaneous errors that are left as `todo!()`s in the code
- [ ] Insert CollectedItems into the program after each attribute group.
  - This is a bit of an interesting case since we should still pool results of each `derive` to avoid errors there.
  - This also means attributes of the same name will always run in the same grouping. So https://github.com/noir-lang/noir/issues/6143 would still exist in the case of one attribute applied to several different items, checking for past work. Different attributes will always be on different steps however.

## Additional Context

If needed, users can still override an attribute's order by making their own wrapper for the attribute

```noir
#[run_after(foo)]
comptime fn my_wrapper(s: StructDefinition) -> Quoted {
    library_attribute(s)
}
```

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
Superseded by https://github.com/noir-lang/noir/pull/6326

PR #6322, opened 2024-10-23
feat: merge and sort imports

asterite:
# Description

## Problem

Manually arranging imports is time consuming.

## Summary

Added two new configurations, `reorder_imports` (true by default) and `imports_granularity` (`Preserve` by default), similar to the ones found in rustfmt. With the default configuration imports are reorganized alphabetically but not merged.

While implementing this I bumped into one code in the test programs like this:

```noir
use foo;
use foo::bar;
```

Merging those in Rust ends up like this:

```noir
use foo::{bar, self};
```

We don't have `self` in imports, so I had two choices:
1. Try to leave the above unmodified.
2. Implement `self` in imports.

It turned out that implement `self` in imports is very easy: when we desugar imports, if the last segment is "self" we just remove it. Then `use foo::self` is desugared into `use foo`.

## Additional Context

None.

## Documentation

Check one:

# PR Checklist


noir_stdlib/src/prelude.nr
TomAFrench, line 1: Personally I much prefer the original format here. I'm not sure that we'd want to enforce such an opinionated change in user code.

noir_stdlib/src/uint128.nr
```diff
@@ -313,7 +315,7 @@ impl Shr for U128 {
 }
 
 mod tests {
-    use crate::uint128::{U128, pow64, pow63};
+    use crate::uint128::{pow63, pow64, U128};
```
TomAFrench, line 318: I like the alphabetising of imports within a single `use`.

TomAFrench (changes requested):
I don't think we should be so aggressive on merging different `use` statements together. Other changes look good though.

noir_stdlib/src/prelude.nr
asterite: I see... then maybe this PR isn't good then. I just checked in Rust and it's true, they don't merge imports if you write them by hand and save. Well, it's controlled by [this configuration](https://rust-lang.github.io/rustfmt/?version=v1.4.36&search=#imports_granularity).

So I guess imports are only merged when they are added via auto-import, and if you only use auto-import then all of them will end up being merged.

Hmm... maybe this PR could only be about sorting imports then. Or maybe `merge_imports` can be `false` by default.

TomAFrench: Yeah I think the ideal behaviour for us is:
1. Human written imports are not merged.
2. Entries within braces are sorted alphabetically.
2. injected imports are placed within existing braces if possible but otherwise created on a new line.

I'm happy for us to have the full merging behaviour being opt-in (as long as we're happy to maintain that extra code in the short term)

asterite: We could have two options:
- reorder_imports
- merge_imports

If `reorder_imports` is `true`, they'll be rearranged alphabetically (in a group, but also inside a single `use`).
If `merge_imports` is `true`, import groups will be merged and reordered (so this setting implies `reoder_imports` is `true` too).

Then we can have `merge_imports` false by default, and `reorder_imports` true by default.

What do you think?

(I'll do that after the trait where clause, though)

jfecher (approved):
Code LGTM.

I could see an argument for it either way, but maybe the option should be off by default like Tom said.

noir_stdlib/src/prelude.nr
jfecher: The two option scheme sounds good to me.

asterite: I'll wait for a confirmation before merging, just in case.

TomAFrench: Looks good, I'm not getting the imports autoformatting if I mess with them a bit which is strange but the unit tests are working as expected.

TomAFrench (approved):

noir_stdlib/src/prelude.nr
asterite: Let me know the code snippets that trigger this, it's probably a bug

PR #6324, opened 2024-10-23
feat(perf): Use [u32;16] for message block in sha256

### LINKED ISSUE #6304: feat(sha256_var): Combine building of the message block and conversion to u32s

aakoshh:
# Description

## Problem\*

Resolves #6304 
Follows https://github.com/noir-lang/noir/pull/6318

## Summary\*

Changes the `MSG_BLOCK` type in `sha256.nr` from `[u8; 64]` to `[u32; 16]` to:
* save a step of having to convert before compression
* reduce the number of array writes (which currently incur copying) in favour of more arithmetic operations (byte packing)

### Testing

```
cargo test -p nargo_cli --test stdlib-tests -- run_stdlib_tests sha256
cargo test -p nargo_cli --test stdlib-props fuzz_sha256
```

NB we can run e.g. `-- run_stdlib_tests msg_just_under_block` to execute a specific test.

### Benchmarks

```shell
cargo bench -p nargo_cli --bench criterion sha256_long
```

On my machine it shows that it got moderately faster:

```console
❯ cargo bench -p nargo_cli --bench criterion sha256_long
...
bench_sha256_long_execute
                        time:   [1.3413 ms 1.3477 ms 1.3555 ms]
                        change: [-13.718% -13.172% -12.577%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 20 measurements (10.00%)
  2 (10.00%) high mild

bench_sha256_long_execute_brillig
                        time:   [244.52 µs 259.88 µs 280.67 µs]
                        change: [-26.479% -23.096% -20.123%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 20 measurements (20.00%)
  4 (20.00%) low severe
```

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
Just ran the new `noir-profiler execution-flamegraph` from https://github.com/noir-lang/noir/pull/6327 command on this PR and am getting less than half the same amount of opcodes execution for `sha256`. 
Before this PR with the sha256 from master (28k samples):
image

Current state of this PR (13k samples):
Screenshot 2024-10-23 at 3 42 56 PM

noir_stdlib/src/hash/sha256.nr
```diff
@@ -169,29 +164,74 @@ unconstrained fn build_msg_block<let N: u32>(
...
+
+    // Figure out the number of items in the int array that we have to pack.
+    // e.g. if the input is [0,1,2,3, 4,5] then we need to pack it as 2 items: [0123, 4500]
+    let mut int_input = block_input / INT_SIZE;
+    if block_input % INT_SIZE != 0 {
+        int_input = int_input + 1;
+    };
+
+    for i in 0..int_input {
+        let mut msg_field: u32 = 0;
```
aakoshh, line 176: I switched to `u32` because I used division somewhere and with `Field` I couldn't get back the values. Does it make a difference?

noir_stdlib/src/hash/sha256.nr
```diff
@@ -114,13 +127,13 @@ pub fn sha256_var<let N: u32>(msg: [u8; N], message_size: u64) -> HASH {
     // Pad the rest such that we have a [u32; 2] block at the end representing the length
     // of the message, and a block of 1 0 ... 0 following the message (i.e. [1 << 7, 0, ..., 0]).
     // Here we rely on the fact that everything beyond the available input is set to 0.
-    msg_block[msg_byte_ptr] = 1 << 7;
-    let last_block = msg_block;
+    msg_block = set_msg_byte(msg_block, msg_byte_ptr, 1 << 7);
```
TomAFrench, line 130: This (implementation of `set_msg_byte`) feels like it can be simplified as we know that the the contents of `msg_block` after where we're inserting the byte is all zeros. We then don't need to extra just that single byte but can pull out everything after it as well.

noir_stdlib/src/hash/sha256.nr
```diff
@@ -202,66 +246,209 @@ fn verify_msg_block<let N: u32>(
...
+    let new_field = into_field_byte(msg_byte, msg_byte_ptr);
+    msg_field - old_field + new_field
+}
+
+// Get a byte of a field in a `MSG_BLOCK` according to its overall position in the `BLOCK_SIZE` space.
+fn from_field_byte(mut msg_field: INT_FIELD, msg_byte_ptr: BLOCK_BYTE_PTR) -> u8 {
+    // How many times do we have to shift to the right to get to the position we want?
+    let max_shifts = INT_SIZE - 1;
+    let shifts = max_shifts - msg_byte_ptr % INT_SIZE;
+    for i in 0..max_shifts {
```
TomAFrench, line 346: Can you use an `if is_unconstrained() {}` to just use a bitshift if we're running in an unconstrained context?

noir_stdlib/src/hash/sha256.nr
```diff
@@ -202,66 +246,209 @@ fn verify_msg_block<let N: u32>(
...
+
+// Project a byte into a position in a field based on the overall block pointer.
+// For example putting 1 into pointer 5 would be 100, because overall we would
+// have [____, 0100] with indexes [0123,4567].
+fn into_field_byte(msg_byte: u8, msg_byte_ptr: BLOCK_BYTE_PTR) -> INT_FIELD {
+    let mut msg_field = msg_byte as u32;
+    // How many times do we have to shift to the left to get to the position we want?
+    let max_shifts = INT_SIZE - 1;
+    let shifts = max_shifts - msg_byte_ptr % INT_SIZE;
+    for i in 0..max_shifts {
```
TomAFrench, line 364: ditto here

noir_stdlib/src/hash/sha256.nr
```diff
@@ -202,66 +246,209 @@ fn verify_msg_block<let N: u32>(
...
+
+// Pack the value of byte into a field of an existing message block, shifting the byte according to its position.
+fn set_msg_byte(mut msg_block: MSG_BLOCK, msg_byte_ptr: BLOCK_BYTE_PTR, msg_byte: u8) -> MSG_BLOCK {
+    let i = msg_byte_ptr / INT_SIZE;
+    msg_block[i] = set_field_byte(msg_block[i], msg_byte_ptr, msg_byte);
+    msg_block
+}
+
+// Replace one byte in a field.
+fn set_field_byte(msg_field: INT_FIELD, msg_byte_ptr: BLOCK_BYTE_PTR, msg_byte: u8) -> INT_FIELD {
```
TomAFrench, line 334: I think it would be worth exploring just merging `from_field_byte` and `to_field_byte` into this method. We could then merge the for-loops which would reduce some bytecode size/runtime cost

noir_stdlib/src/hash/sha256.nr
```diff
@@ -202,66 +246,209 @@ fn verify_msg_block<let N: u32>(
...
+    let modulo = msg_byte_ptr % INT_SIZE;
+    if modulo != 0 {
+        // Index of the block in which we find the field we need to partially zero.
+        let i = msg_byte_ptr / INT_SIZE;
+        let mut msg_field = msg_block[i];
+        for _ in 0..(INT_SIZE - modulo) {
+            msg_field = set_field_byte(msg_field, msg_byte_ptr, 0);
+            msg_byte_ptr = msg_byte_ptr + 1;
+        }
+        msg_block[i] = msg_field as u32;
```
TomAFrench, line 400:
```suggestion
        msg_block[i] = msg_field;
```

noir_stdlib/src/hash/sha256.nr
```diff
@@ -202,66 +246,209 @@ fn verify_msg_block<let N: u32>(
...
+    // Check that up to the pointer we match the last block.
+    verify_msg_block_equals_last(msg_block, last_block, msg_byte_ptr);
 
     // We verify the message length was inserted correctly by reversing the byte decomposition.
-    let len = 8 * message_size;
     let mut reconstructed_len: Field = 0;
-    for i in MSG_SIZE_PTR..BLOCK_SIZE {
-        reconstructed_len = 256 * reconstructed_len + msg_block[i] as Field;
+    for i in INT_SIZE_PTR..INT_BLOCK_SIZE {
+        reconstructed_len = 256 * 256 * 256 * 256 * reconstructed_len + msg_block[i] as Field;
```
TomAFrench, line 441:
```suggestion
        let two_pow_32 = 256 * 256 * 256 * 256;
        reconstructed_len = two_pow_32 * reconstructed_len + msg_block[i] as Field;
```

TomAFrench (commented):
Some small suggestions but I think there's extra brillig improvements left on the table in a few places.

noir_stdlib/src/hash/sha256.nr
```diff
@@ -12,15 +12,25 @@ global MSG_SIZE_PTR = 56;
...
+
 // Index of a byte in a 64 byte block; ie. 0..=63
 type BLOCK_BYTE_PTR = u32;
 
 // The foreign function to compress blocks works on 16 pieces of 4-byte integers, instead of 64 bytes.
 type INT_BLOCK = [u32; INT_BLOCK_SIZE];
 
+// We can represent 4-byte integers during construction as `Field` or `u32`.
+// Using `u32` because we shift right by dividing by 256, which doesn't work with `Field`.
+type INT_FIELD = u32;
```
TomAFrench, line 29: Let's remove this type alias, I don't think we're gaining much from it and it just makes it harder to see what type we're using.

noir_stdlib/src/hash/sha256.nr
aakoshh: Yes, I kept it for historical reasons, but I see how a newcomer would be scratching their heads. That's why I added an `INT_FIELD` to explain and indicate in functions that also bear the "field" name what we're returning. Should we call it "item"?

noir_stdlib/src/hash/sha256.nr
```diff
@@ -202,66 +246,209 @@ fn verify_msg_block<let N: u32>(
...
+    // First integer which is supposed to be (partially) zero.
+    let mut int_byte_ptr = msg_byte_ptr / INT_SIZE;
+
+    // Check partial zeroes.
+    let modulo = msg_byte_ptr % INT_SIZE;
+    if modulo != 0 {
+        let msg_field = msg_block[int_byte_ptr];
+        for i in 0..INT_SIZE {
+            if i < INT_SIZE - modulo {
+                assert_eq(from_field_byte(msg_field, msg_byte_ptr), zero as u8);
```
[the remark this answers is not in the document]
aakoshh: Good idea 👍

noir_stdlib/src/hash/sha256.nr
aakoshh: I thought this is one of the most straight forward ways to implement it, but you're right in this instance it might not be the most optimal. Bit of a brain teaser every time I have to touch one of these items, so I thought simpler is good. I'll have a look at a more specialised approach.

noir_stdlib/src/hash/sha256.nr
aakoshh: I suppose we can. Would that work on `Field` as well? Maxim said the compiler turns bit shifts into divisions, and we use bit shift in `1 << 7` in constrained code too. Is there a rule of thumb to follow?

aakoshh: Done in https://github.com/noir-lang/noir/pull/6324/commits/2fa178f19eda14ddc09622bf12923a99c0ea7ece

TomAFrench: We can't do any bitshifts on `Field` types, only on integers.

The general rule of thumb on bitshifts is that in a constrained context we don't have a native way to perform bitshifts within a circuit. We then need to convert it into multiplications/divisions. If we know by how much we're shifting by at compile time this can be easily turned into a single multiplication/division and so is relatively cheap. If we don't know how many bits we're shifting by until runtime then we need to do a lot more work to handle potential cases and so it's more expensive, generally we want to avoid doing bitshifts of this type.

In unconstrained code however we have native support for bitshifts in the VM so there's no need to decompose the bitshifts into multiplications and so we can just do it directly. This helps in this case as we don't need to have the for-loop, etc. and the function then simplifies down to `(msg_field >> shift) as u8`

noir_stdlib/src/hash/sha256.nr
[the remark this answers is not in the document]
aakoshh: Done in https://github.com/noir-lang/noir/pull/6324/commits/b1cb66757b986f269fe8a7a5ac12e9de625aa4da

noir_stdlib/src/hash/sha256.nr
TomAFrench: Ah, sorry I think you misinterpreted me. The `rshift8` functions aren't what I was talking about here as those are cases where we're always shifting by a constant amount so it's basically equivalent in constrained and unconstrained code. I think we can revert the creation of `rshift8` as the multiplication we had previously was fine.

What I was referring to was this whole for-loop where we're emulating a bitshift by a number of bits known at runtime. In this case we need to break down the constrained code to perform a number of divisions as we're doing currently but the unconstrained version can do a standard bitshift.

aakoshh: Gotcha. I added a `shifts` parameter to `lshift8` and `rshift8` in https://github.com/noir-lang/noir/pull/6324/commits/081c8ba6f80a8a6cb3a063dee9abcef81b7a6819 to get rid of the for loop.

noir_stdlib/src/hash/sha256.nr
aakoshh: Added a bunch of new helpers and now setting not just a specific byte in the number but all following bytes to zero as well. This saves us the work of reading what was in that position, because we don't have to preserve the others by subtracting it.

noir_stdlib/src/hash/sha256.nr
aakoshh: This functions is not used any more apart from tests. Does removing it reduce the circuit or bytecode size?

aakoshh:
The latest changes did indeed help Brillig:

```
bench_sha256_long_execute_brillig
                        time:   [225.33 µs 230.03 µs 235.24 µs]
                        change: [-20.162% -17.224% -13.501%] (p = 0.00 < 0.05)
                        Performance has improved.

```

noir_stdlib/src/hash/sha256.nr
```diff
@@ -169,99 +169,352 @@ unconstrained fn build_msg_block<let N: u32>(
     } else {
         BLOCK_SIZE
     };
-    for k in 0..block_input {
-        msg_block[k] = msg[msg_start + k];
+
+    // Figure out the number of items in the int array that we have to pack.
+    // e.g. if the input is [0,1,2,3, 4,5] then we need to pack it as 2 items: [0123, 4500]
```
vezenovm, line 174:
```suggestion
    // e.g. if the input is [0,1,2,3,4,5] then we need to pack it as 2 items: [0123, 4500]
```
nit, the space was bothering me lol

noir_stdlib/src/hash/sha256.nr
```diff
@@ -114,49 +123,40 @@ pub fn sha256_var<let N: u32>(msg: [u8; N], message_size: u64) -> HASH {
     // Pad the rest such that we have a [u32; 2] block at the end representing the length
     // of the message, and a block of 1 0 ... 0 following the message (i.e. [1 << 7, 0, ..., 0]).
     // Here we rely on the fact that everything beyond the available input is set to 0.
-    msg_block[msg_byte_ptr] = 1 << 7;
-    let last_block = msg_block;
+    // msg_block = set_block_byte(msg_block, msg_byte_ptr, 1 << 7);
```
vezenovm, line 126:
```suggestion
```
leftover?

noir_stdlib/src/hash/sha256.nr
```diff
@@ -169,99 +169,352 @@ unconstrained fn build_msg_block<let N: u32>(
...
-            assert_eq(msg_block[msg_byte_ptr], msg[k]);
+    // Reconstructed packed item.
+    let mut msg_item: u32 = 0;
+
+    // Inclusive at the end so that we can compare the last item.
+    let mut i: u32 = 0;
+    for k in msg_start..=msg_end {
+        if k % INT_SIZE == 0 {
+            // If we consumed some input we can compare against the block.
+            if msg_start < message_size & k > msg_start {
```
vezenovm, line 228: Noir wants parens for if conditional statements 🤔 surprised it isn't yelling at you here, I'm wondering if the correct thing is being checked.
```suggestion
            if (msg_start < message_size) & (k > msg_start) {
```

noir_stdlib/src/hash/sha256.nr
```diff
@@ -169,99 +169,352 @@ unconstrained fn build_msg_block<let N: u32>(
...
-        if i >= msg_byte_ptr {
+    // First integer which is supposed to be (partially) zero.
+    let mut int_byte_ptr = msg_byte_ptr / INT_SIZE;
+
+    // Check partial zeros.
+    let modulo = msg_byte_ptr % INT_SIZE;
+    if modulo != 0 {
+        let num_zeros = INT_SIZE - modulo;
+        let mask = if num_zeros == 3 {
+            256 * 256 * 256
```
vezenovm, line 274: Could make these globals to avoid the magic numbers

noir_stdlib/src/hash/sha256.nr
```diff
@@ -169,99 +169,352 @@ unconstrained fn build_msg_block<let N: u32>(
...
+}
+
+// // Pack the value of byte into an item of an existing message block, shifting the byte according to its position.
+// fn set_block_byte(msg_block: MSG_BLOCK, msg_byte_ptr: BLOCK_BYTE_PTR, msg_byte: u8) -> MSG_BLOCK {
+//     update_block_item(
+//         msg_block,
+//         msg_byte_ptr,
+//         |msg_item| set_item_byte(msg_item, msg_byte_ptr, msg_byte),
+//     )
+// }
```
vezenovm, line 346:
```suggestion
```
more leftover looks like, as well as the func below

noir_stdlib/src/hash/sha256.nr
```diff
@@ -169,99 +169,352 @@ unconstrained fn build_msg_block<let N: u32>(
...
+        } else {
+            // Brillig wouldn't shift 0<<4 without overflow.
+            item << (8 * shifts)
+        }
+    } else {
+        // We can do a for loop up to INT_SIZE or an if-else.
+        if shifts == 0 {
+            item
+        } else if shifts == 1 {
+            item * 256
```
vezenovm, line 415: Did you notice a difference with just using `<<`? The compiler should simplify `<<` to `*`

tooling/nargo_cli/tests/stdlib-tests.rs
```diff
@@ -76,7 +76,7 @@ fn run_stdlib_tests() {
                 &bn254_blackbox_solver::Bn254BlackBoxSolver,
                 &mut context,
                 &test_function,
-                false,
+                true,
```
vezenovm, line 79: Is this just for local dev?

noir_stdlib/src/hash/sha256.nr
```diff
@@ -510,4 +768,96 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_byte_into_item() {
+        let fld = make_item(0, 20, 0, 0);
+        assert_eq(byte_into_item(20, 1), fld);
+        assert_eq(byte_into_item(20, 5), fld);
+    }
+
+    // #[test]
```
vezenovm, line 788: Some commented out tests

vezenovm (commented):
Some minor comments

noir_stdlib/src/hash/sha256.nr
aakoshh: Haha, ok, I wanted to make it easier to see what the grouping was.

noir_stdlib/src/hash/sha256.nr
aakoshh: Just a reminder of how it was originally, why those methods existed, as an alternative.

noir_stdlib/src/hash/sha256.nr
aakoshh: I know, left them for reference, they are nice. I'll remove, I guess it's bothering everyone.

noir_stdlib/src/hash/sha256.nr
vezenovm: Ah I see what you were going for. I think it becomes clear in the separate to follow

tooling/nargo_cli/tests/stdlib-tests.rs
aakoshh: I used it to see `println`. Doesn't seem to have any negative effect when the tests pass.

noir_stdlib/src/hash/sha256.nr
aakoshh: I didn't try, just kept this because of how it was previously. Tom recommended to use bit shifting in unconstrained because it can deal with any input, whereas in the circuit it has to be known. I'll try, maybe it can be used to avoid having globals naming all these.

vezenovm: Ah I see it is because we do not necessarily know the `shifts` at compile time.

noir_stdlib/src/hash/sha256.nr
aakoshh: Added TWO_POW... globals.

noir_stdlib/src/hash/sha256.nr
aakoshh: Tried it with bit shifting and the globals `TWO_POW_...`s. The latter is 10% faster.

vezenovm:
> bench_sha256_long_execute_brillig
>                         time:   [225.33 µs 230.03 µs 235.24 µs]
>                         change: [-20.162% -17.224% -13.501%] (p = 0.00 < 0.05)
>                         Performance has improved.

I ran the flamegraph from #6327 again and got the following:
Screenshot 2024-10-24 at 4 29 01 PM

We improved from 139k samples -> 133k

TomAFrench (approved):

PR #6326, opened 2024-10-23
feat: Order attribute execution by their source ordering

jfecher:
# Description

## Problem\*

Resolves 

## Summary\*

Executes attributes by their source ordering. Their source ordering within a crate is defined as:

- Modules that are child modules are executed before their parent modules
- Sibling modules are executed in the order of their `mod foo; mod bar;` declarations in their parent modules
- The module-tree is executed depth-first
- Within a module, attributes are executed from first to last byte index in their source file

## Additional Context

- Currently still needs to insert each CollectedItem in as it is generated rather than grouping them. This does mean that this PR is currently roughly in the same state as #6319.
- I'm assuming sibling modules are declared in order in the code via `add_child_module` for the sibling module ordering. I'll verify that this ordering matches my assumptions later.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/elaborator/comptime.rs
```diff
@@ -40,6 +40,8 @@ struct AttributeContext {
     attribute_module: LocalModuleId,
 }
 
+type CollectedAttributes = Vec<(FuncId, Value, Vec<Expression>, AttributeContext, Span)>;
```
asterite, line 43: Totally optional: should this tuple be its own struct? This is the only comment I have on this PR, and I'm not even sure a struct is needed here (it's kind of clear what each of these is because of the types) but it might slightly improve readability.

asterite (approved):
Looks great! The ordering test is nice too.

compiler/noirc_frontend/src/elaborator/comptime.rs
jfecher: I think it's fine as an alias mostly because it's just something private to this module to reduce typing / clarify signatures a bit. I don't think anything would be gained from making it a struct in that case.

jfecher:
> FAIL [   0.652s] nargo_cli::execute tests::compile_success_empty::test_attribute_sees_result_of_previous_attribute

This is really weird, I can't get this test to fail locally at all.

jfecher:
Wondering if we should wait to merge this since it'd likely largely break aztec macros.

We could also potentially wait until the `run_before`/`run_after` helpers are added in anticipation of aztec-nr's necessary rewrite for the new attribute order.

michaeljklein (approved):
LGTM

TomAFrench (changes requested):
Blocking this until we have a mirror on the asset packages repository to test how broken macros are

jfecher:
Update: added docs

TomAFrench (approved):

TomAFrench:
Seems like the necessary changes in aztec-packages are pretty light-touch so we should be good to merge this.

PR #6327, opened 2024-10-23
feat(profiler)!: New flamegraph command that profiles the opcodes executed

vezenovm:
# Description

## Problem\*

Resolves 

Initial profiling to get better numbers for the opcodes actually being executed by the Brillig VM.

## Summary\*

cc @sirasistant for a lot of the initial code in this PR.

The noir-profiler currently only looks at compiled artifacts to generate a flamegraph showing where constraints counts and/or opcodes counts are most prominent in a program. This PR adds execution profiling sampling to be part of the ACVM/Brillig VM and the noir-profiler. This will enable us to see if there are opcodes which are being unnecessarily executed (such as array copies). 

This initial PR simply adds the `execution-flamegraph` command while a follow-up will add extra metadata to look at Brillig procedures. 

After compiling `execution_success/sha256` with `--force-brillig` we can run the following command:
```
noir-profiler execution-opcodes -a ./target/sha256.json -p ./Prover.toml -o ./target/
```

Result for execution flamegraph:

Screenshot 2024-10-23 at 7 17 48 PM

To compare here is the result of `noir-profiler opcodes-flamegraph`:

Screenshot 2024-10-23 at 7 18 12 PM

## Additional Context

This is a minor breaking change as I changed the names of the commands in the `noir-profiler` to not all be suffixed with `flamegraph`.

## Documentation\*

Check one:

# PR Checklist\*


tooling/profiler/src/flamegraph.rs
```diff
@@ -17,7 +18,7 @@ use super::opcode_formatter::format_opcode;
 
 #[derive(Debug)]
 pub(crate) struct Sample<F: AcirField> {
-    pub(crate) opcode: AcirOrBrilligOpcode<F>,
+    pub(crate) opcode: Option<AcirOrBrilligOpcode<F>>,
```
asterite, line 21: Is the idea that `None` is also captured in the flamegraph?

acvm-repo/brillig_vm/src/lib.rs
```diff
@@ -196,6 +212,13 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>> VM<'a, F, B> {
 
     /// Process a single opcode and modify the program counter.
     pub fn process_opcode(&mut self) -> VMStatus<F> {
+        let call_stack: Vec<usize> = self.get_call_stack();
+        self.profiling_samples.push(BrilligProfilingSample { call_stack });
+
+        self.process_opcode_internal()
+    }
```
asterite, line 219: Is it a concern that this profiling happens regardless of whether profiling is active? I don't know if this makes thing slower. But this is the first time I review a brillig_vm PR, so let me know if what I'm saying doesn't make sense.

acvm-repo/acvm/src/pwg/brillig.rs
```diff
@@ -200,16 +200,16 @@ impl<'b, B: BlackBoxFunctionSolver<F>, F: AcirField> BrilligSolver<'b, F, B> {
     }
 
     pub(crate) fn finalize(
```
asterite, line 202: 💭  There could be two finalize methods:
- `finalize` (the existing one before this PR)
- `finalize_with_profiling` that's the new one in this PR

That way we'd avoid the many `.finalize().0` calls. Similarly, there could be `execute_program` and `execute_program_with_profiling` to avoid some `.map(|(witness_stack, _)| witness_stack)`.

acvm-repo/brillig_vm/src/lib.rs
vezenovm: I don't think it should make things much slower, but it is hard to tell as we do not have comprehensive benchmarks for execution timings like we do bytecode sizes. I will change this either way to mark whether profiling is active and whether to collect samples.

acvm-repo/acvm/src/pwg/brillig.rs
vezenovm: done

tooling/profiler/src/flamegraph.rs
vezenovm: Yes this is handled in the flamegraph. The folding/sorting process checks whether there is some opcode in the sample. If there is not, it will not be added to the list of location names.

acvm-repo/brillig_vm/src/lib.rs
sirasistant: Yes It should be way better with the flag because if not then the process opcode codepath (the hottest) will trigger vec reallocations which are very expensive

asterite (approved):
Looks great!

acvm-repo/acvm/tests/solver.rs
```diff
@@ -319,6 +321,7 @@ fn double_inversion_brillig_oracle() {
         witness_assignments,
         &unconstrained_functions,
         &[],
+        false,
```
aakoshh, line 324: If it's mostly `false`, maybe it would make sense to add a `fn with_profiler(self, enable: bool) -> Self` method to selectively turn it on. It might even return a special wrapper that allows you to call a version of `finalize` that returns an `Option<Samples>`, but the base VM would not collect samples and then no panics are necessary.

acvm-repo/acvm/src/pwg/mod.rs
```diff
@@ -247,6 +262,13 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>> ACVM<'a, F, B> {
     }
 
     /// Finalize the ACVM execution, returning the resulting [`WitnessMap`].
+    pub fn finalize_with_profiling(self) -> (WitnessMap<F>, ProfilingSamples) {
```
aakoshh, line 265: `finalize` could delegate to this for consistency in status handling.

acvm-repo/brillig_vm/src/lib.rs
```diff
@@ -196,6 +220,15 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>> VM<'a, F, B> {
...
     pub fn process_opcode(&mut self) -> VMStatus<F> {
+        if self.profiling_active {
+            let call_stack: Vec<usize> = self.get_call_stack();
+            self.profiling_samples.push(BrilligProfilingSample { call_stack });
+        }
+
+        self.process_opcode_internal()
+    }
+
+    pub fn process_opcode_internal(&mut self) -> VMStatus<F> {
```
aakoshh, line 231:
```suggestion
    fn process_opcode_internal(&mut self) -> VMStatus<F> {
```

tooling/nargo/src/ops/execute.rs
```diff
@@ -184,25 +194,70 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>, E: ForeignCallExecutor<F>>
         // included in a failure case.
         self.call_stack.clear();
 
-        Ok(acvm.finalize())
+        if self.profiling_active {
+            Ok(acvm.finalize_with_profiling())
+        } else {
+            Ok((acvm.finalize(), ProfilingSamples::default()))
```
aakoshh, line 200: I think this double-checking of which version of finalize to call, only then to return an empty list, which then again is double checked to be empty, is a bit too defensive. I reckon an `Option` which is `None` if `profiling_active` is `false` should be enough, we could trust the ACVM to do the right thing, and then all the subsequent processing becomes more streamlined.

tooling/profiler/src/cli/execution_flamegraph_cmd.rs
```diff
@@ -0,0 +1,95 @@
...
+use noirc_abi::input_parser::Format;
+use noirc_artifacts::debug::DebugArtifact;
+
+#[derive(Debug, Clone, Args)]
+pub(crate) struct ExecutionFlamegraphCommand {
+    /// The path to the artifact JSON file
+    #[clap(long, short)]
+    artifact_path: PathBuf,
+
+    #[clap(long, short)]
```
aakoshh, line 22:
```suggestion
    /// The path to the Prover.toml file
    #[clap(long, short)]
```

tooling/profiler/src/cli/mod.rs
```diff
@@ -19,15 +20,17 @@ struct ProfilerCli {
 #[non_exhaustive]
 #[derive(Subcommand, Clone, Debug)]
 enum ProfilerCommand {
-    GatesFlamegraph(gates_flamegraph_cmd::GatesFlamegraphCommand),
-    OpcodesFlamegraph(opcodes_flamegraph_cmd::OpcodesFlamegraphCommand),
+    Gates(gates_flamegraph_cmd::GatesFlamegraphCommand),
+    Opcodes(opcodes_flamegraph_cmd::OpcodesFlamegraphCommand),
+    ExecutionOpcodes(execution_flamegraph_cmd::ExecutionFlamegraphCommand),
```
aakoshh, line 25: You can keep the names backwards compatible using [aliases](https://github.com/clap-rs/clap_derive/blob/acea444f962a0d7c8126273739d4df8bdb152624/examples/subcommand_aliases.rs#L10-L13)

tooling/profiler/src/cli/execution_flamegraph_cmd.rs
```diff
@@ -0,0 +1,95 @@
...
+    )?;
+    println!("Executed");
+
+    let profiling_samples: Vec<Sample<FieldElement>> = profiling_samples
+        .into_iter()
+        .map(|sample| {
+            let call_stack = sample.call_stack;
+            let brillig_function_id = sample.brillig_function_id;
+            let last_entry = call_stack.last();
+            let opcode = brillig_function_id
```
aakoshh, line 67: Can the `sample.opcode` already be `Some` here?

vezenovm: `sample` is a `ProfilingSample` type which does not have an opcode field. `opcode` on this line is already an option as we are mapping `brillig_function_id` which is an option.

tooling/profiler/src/cli/mod.rs
vezenovm: I felt it was fine to rename them fully for now anyway as the flamegraph suffix felt unnecessary (e.g. `opcodes-flamegraph` is more verbose than `opcodes`) and all the commands just return flamegraphs anyway. If the `noir-profiler` ever does anything other than generate flamegraphs we can rename them then I feel. Noted for the future though.

tooling/profiler/src/flamegraph.rs
```diff
@@ -88,37 +89,49 @@ fn generate_folded_sorted_lines<'files, F: AcirField>(
...
+                                    vec![]
+                                }
+                            } else {
+                                vec![]
+                            }
+                        });
+                    let callsite_labels: Vec<_> = source_locations
+                        .into_iter()
+                        .map(|location| location_to_callsite_label(location, files))
+                        .collect();
```
aakoshh, line 122: Perhaps now that there is caching, it would be helpful to put this into a function that turns an opcode location into callsite labels using the debug symbols. Then what's left here could be as simple as 
```rust
let callsite_labels = resolution_cache.entry(opcode_location).or_insert_with(|| 
  find_callsites(&debug_symbols, sample.brillig_function_id.as_ref(), &opcode_location
).clone();`

File: acvm-repo/acvm/tests/solver.rs (hunk 4 above)
Comment (vezenovm) [reply acvm-repo/acvm/tests/solver.rs:324]: I added this method.

File: tooling/nargo/src/ops/execute.rs (hunk 7 above)
Comment (vezenovm) [reply tooling/nargo/src/ops/execute.rs:200]: `None` for the second tuple field of the profiling samples? This feels partially redundant with knowing that we would expect no samples when profiling is not active.

File: acvm-repo/acvm/src/pwg/mod.rs (hunk 5 above)
Comment (vezenovm) [reply acvm-repo/acvm/src/pwg/mod.rs:265]: done

File: tooling/nargo/src/ops/execute.rs (hunk 7 above)
Comment (aakoshh) [reply tooling/nargo/src/ops/execute.rs:200]: Sorry, my comment was confusing because the code involved is a spread out. 

What I meant was that this pattern:
```
struct VM { profiling: bool, samples: Samples }
impl VM {
  fn finalize() -> Witness { 
    assert!(!self.profiling);
    ...
  }
  fn finalize_with_profiling(self) -> (Witness, Samples) {
    assert!(self.profiling);
    ...
  }
}

fn foo(profiling: bool) -> (Witness, Samples) {
  let vm = VM::new(..., profiling = self.profiling);
  ...
  let (result, samples) = if self.profiling { 
    vm.finalize_with_profiling()
  } else { 
    (vm.finalize(), Samples::default())
  };
}

fn bar() -> Witness {
  let (result, samples) = foo(false);
  assert(samples.is_empty());
  result
}

fn baz() -> (Witness, Samples) {
  foo(true)
}
```
is like realising the four eyes principle in code, with the caller and the callee constantly second guessing each other. Like there are samples in VM, but they aren't returned, but then a default is added, which is asserted to be empty, and then discarded...

I just meant that if we trust that the VM itself will return `Some(Samples)` if `profiling_active` is `true` and `None` if it's false, then the rest of them can be without further assertions. 

I would be in support of an API which looks like `let (witness, sample) = VM::new(...).with_profiling().finalize()` or `let witness = VM::new().finalize()` (ie. a wrapper) but that doesn't seem possible because the Solver also just receives a flag from the outside and has to handle both cases. I'm just saying the double checking seems unnecessary. Maybe a `.expect("profiling is on")` by a caller that passed `true` and really wants those samples
 enough?

aakoshh: I understand the motivation that you don't want to accidentally profile when the code isn't even consuming it, just seemed like it's negated here a bit with the re-introduction of a default value.

vezenovm: Ok I see what you mean. And it is a fair point that I should trust the ACVM to act appropriately. I switched to just fetching the profiling samples before calling `finalize`. I then actually check whether the profiling samples returned from the ACVM are empty when I am not profiling rather than that default value.

PR #6332, opened 2024-10-24
feat: Add capacities to brillig vectors and use them in slice ops

sirasistant:
# Description

## Problem\*

Brillig vectors didn't have capacities, so all operations were O(N) since they required copying the vector.
This PR adds capacities to vectors and uses them in the slice operations to avoid copying when possible.

## Additional Context

Unfortunately I can't make an e2e test showcasing a simple program where performance is good thanks to this due to reference counter weirdness:
```noir
unconstrained fn main() {
    let mut slice = &[];

    for i in 0..10000 {
        slice = slice.push_back(i);
    }

    assert(slice.len() == 10000);
}
```
```
brillig(inline) fn main f0 {
  b0():
    v1 = allocate
    store u32 0 at v1
    inc_rc []
    v4 = allocate
    store [] at v4
    jmp b1(u32 0)
  b1(v0: u32):
    v6 = lt v0, u32 2⁴×625
    jmpif v6 then: b3, else: b2
  b3():
    v8 = load v1
    v9 = load v4
    v11, v12 = call slice_push_back(v8, v9, v0)
    inc_rc v12
    store v11 at v1
    store v12 at v4
    v14 = add v0, u32 1
    jmp b1(v14)
  b2():
    v7 = load v1
    constrain v7 == u32 2⁴×625
    return 
}
```

The `inc_rc v12` makes it so the slice stored at v4 always has RC=2 so slice op optimizations do not trigger.
They do trigger in other e2e tests though. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_black_box.rs
+++ b/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_black_box.rs
@@ -8,7 +8,7 @@ use acvm::{
 
 use crate::brillig::brillig_ir::{
     brillig_variable::BrilligVariable, debug_show::DebugToString, registers::RegisterAllocator,
-    BrilligBinaryOp, BrilligContext,
+    BrilligContext,
 };
 
 /// Transforms SSA's black box function calls into the corresponding brillig instructions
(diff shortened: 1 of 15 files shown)


sirasistant:
It's normal for bytecode to be marginally bigger: the procedures need more bytecode to use capacities. However they are deduplicated, so the bytecode size increase is independent of program size.

sirasistant:
After this PR I'll add more PRs to use capacities & RCs in the rest of slice ops

TomAFrench (approved):

PR #6335, opened 2024-10-24
feat: let the formatter remove lambda block braces for single-statement blocks

asterite:
…nt blocks

# Description

## Problem

Not a problem, but rustfmt seems to do the same and it can help reduce some of the code noise.

## Additional Context

Maybe this feature is controversial? I didn't find a way to turn it off in rustfmt.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/nargo_fmt/src/chunks.rs
+++ b/tooling/nargo_fmt/src/chunks.rs
@@ -467,8 +467,7 @@ pub(crate) enum GroupKind {
     /// somewhere that's probably undesired, we'll "turn it" into a block
     /// (write the "{" and "}" delimiters) and write the lambda body in the next line.
     LambdaBody {
-        is_block: bool,
-        block_statements_count: usize,
+        block_statement_count: Option<usize>,
         has_comments: bool,
         lambda_has_return_type: bool,
     },
@@ -767,7 +766,7 @@ impl<'a> Formatter<'a> {
 
             // If a lambda body doesn't fit in the current line and it's not a block,
             // we can turn it into a block and write it in the next line, so its contents fit.
-            if let GroupKind::LambdaBody { is_block: false, .. } = group.kind {
+            if let GroupKind::LambdaBody { block_statement_count: None, .. } = group.kind {
                 // Try to format it again in the next line, but we don't want to recurse
                 // infinitely so we change the group kind.
                 group.kind = GroupKind::Regular;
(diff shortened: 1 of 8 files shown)


tooling/nargo_fmt/src/chunks.rs
```diff
@@ -762,16 +767,17 @@ impl<'a> Formatter<'a> {
...
                 // infinitely so we change the group kind.
                 group.kind = GroupKind::Regular;
                 self.write("{");
                 self.trim_spaces();
                 self.increase_indentation();
-                self.write_line();
+                self.write_line_without_skipping_whitespace_and_comments();
                 self.write_indentation();
                 self.format_chunk_group_impl(group);
+                dbg!(&self.buffer);
```
jfecher, line 780:
```suggestion
```

tooling/nargo_fmt/src/formatter/expression.rs
```diff
@@ -230,12 +232,25 @@ impl<'a, 'b> ChunkFormatter<'a, 'b> {
 
         group.text(params_and_return_type_chunk);
 
-        let body_is_block = matches!(lambda.body.kind, ExpressionKind::Block(..));
+        let (body_is_block, block_statements_count) =
+            if let ExpressionKind::Block(block) = &lambda.body.kind {
+                (true, block.statements.len())
+            } else {
+                (false, 0)
```
jfecher, line 239: Since body_is_block always implies block_statements_count, perhaps we should use an option instead? `block_statement_count: Option<usize>`?

tooling/nargo_fmt/src/chunks.rs
```diff
@@ -803,7 +809,7 @@ impl<'a> Formatter<'a> {
                 // )
                 let comma_trimmed = self.trim_comma();
                 self.decrease_indentation();
-                self.write_line();
+                self.write_line_without_skipping_whitespace_and_comments();
```
jfecher, line 812: When do we want to write a line that skips comments?

asterite: Oh, by default `write_line` will skip any comments and whitespace that exist at the current lexer position. However, here we are writing a bunch of chunks/groups which means that we already lexed the content we want to write. If we call`write_line` here we might be skipping whitespace/comment that we don't want to skip just yet. So this is a bug fix (though I didn't find code that breaks with `write_line`, but I thought it would be best to use the correct method).

tooling/nargo_fmt/src/formatter/expression.rs
asterite: I considered this but wasn't sure which option to use (no pun intended :-D). But given that you are suggesting it too, I'll do the change.

jfecher (approved):

PR #6338, opened 2024-10-24
fix: better formatting of leading/trailing line/block comments in expression lists

asterite:
# Description

## Problem

While fixing the bug [this PR](https://github.com/noir-lang/noir/pull/6337) was fixing I noticed that the formatter destroyed some array formatting, like this one:

https://github.com/noir-lang/noir_json_parser/blob/bc7094394baeaa185c5bf56ae806e302c786bdd3/src/_string_tools/slice_packed_field.nr#L13-L19

## Summary

I decided to try to fix this because it's bad if the formatter doesn't respect this initial formatting. And this change if for any expression list, so it applies to arrays, tuples, call arguments, etc, so we only need to fix this once.

## Additional Context

With this some leading spaces surrounding block comments are gone, but I think it's better because for example array literals don't have spaces after `[` and before `]`, and before this PR the formatter would generate `[ /* comment */ 1]` while now it generates `[/* comment */ 1]` which I think looks better (and is what rustfmt does too). I think I didn't want to spend too much time on these details on the initial formatter pass to avoid getting into an infinite improvement loop.

## Documentation

Check one:

# PR Checklist


jfecher (approved):
Some of the examples still look quite odd but I do mostly like the removal of the leading space in several of them.

Issue #6085, opened 2024-09-18
Investigate usage of "big add" gates in the bb dsl folder


TomAFrench:
Speaking with Kesha today he mentioned that we could accumulate the intermediate variables created when decomposing a large expression more efficiently by using the "big_add" gates in barretenberg.

https://github.com/AztecProtocol/aztec-packages/blob/184cc882b3f1b90d74f149e46100474263a3665d/barretenberg/cpp/src/barretenberg/stdlib_circuit_builders/ultra_circuit_builder.cpp#L265-L297

We should look at having the dsl folder produce these gates and also how to modify how we break up large expressions to ensure that it's compatible.

PR #6341, opened 2024-10-24
feat: experimental acir optimisation *DO NOT MERGE*

guipublic:
# Description

## Problem\*

Related to #6085

## Summary\*
This PR is an attempt to perform the 'as_witness' automatically.
The idea is to create witness for each common expression, instead of having the user to specify the expressions to fit into a witness. Because this strategy generates a lot of 'small' arithmetic opcodes, I benefit from the big-add-gate (cf. #6085) by simplifying those small opcodes into the ones which use them.

## Additional Context
In general this optimisation works well, but some corner cases still need to be investigated. Indeed, the simplification may generate too many opcodes which do not fit the width.
Hence, it is hidden by a non-documented compiler option (experimental-optimisation).
I leave it as a draft for now.

## Documentation\*

Check one:

# PR Checklist\*


guipublic:
I am closing it. Although it improves a bit the number of ACIR opcodes, it has a negative impact on the circuit size.

PR #6355, opened 2024-10-25
fix(ssa): Do not mark an array from a parameter mutable

### LINKED ISSUE #6349: Run tests with a range of values for `inliner_aggressiveness` to test for divergences in behaviour.
@aakoshh noticed in #6340 that compiling the `trait_eq::inequality` test fails with `inliner_aggressiveness = 0` but passes with `inliner_aggressiveness = 9223372036854775807`. We would expect that the resultant brillig bytecode should be equivalent but this is not the case.

We should update the test suite to run at a number of different levels of inliner aggressiveness to check that we don't get different behaviour presented.

cc @sirasistant as non-inlining behaviour is resulting in correctness issues.

vezenovm:
# Description

## Problem\*

Resolves #6349 

## Summary\*

We check that we do not mark arrays mutable which comes from load instructions as that array may be used by multiple values. This is overridden by checking whether we are in the return block, in which case we do mark the array as mutable. This however breaks when we no longer inline everything. I switched the set of arrays from loads to a map of arrays from loads to whether they come from a block parameter.  

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/array_set.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/array_set.rs
@@ -53,7 +53,9 @@ struct Context<'f> {
     is_brillig_runtime: bool,
     array_to_last_use: HashMap<ValueId, InstructionId>,
     instructions_that_can_be_made_mutable: HashSet<InstructionId>,
-    arrays_from_load: HashSet<ValueId>,
+    // Mapping of an array that comes from a load and whether the address 
+    // it was loaded from is a reference parameter.
+    arrays_from_load: HashMap<ValueId, bool>,
     inner_nested_arrays: HashMap<ValueId, InstructionId>,
 }
 
@@ -64,7 +66,7 @@ impl<'f> Context<'f> {
             is_brillig_runtime,
             array_to_last_use: HashMap::default(),
             instructions_that_can_be_made_mutable: HashSet::default(),
-            arrays_from_load: HashSet::default(),
+            arrays_from_load: HashMap::default(),
             inner_nested_arrays: HashMap::default(),
         }
     }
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/array_set.rs
```diff
@@ -133,10 +139,12 @@ impl<'f> Context<'f> {
...
                     }
                 }
-                Instruction::Load { .. } => {
+                Instruction::Load { address } => {
                     let result = self.dfg.instruction_results(*instruction_id)[0];
                     if matches!(self.dfg.type_of_value(result), Array { .. } | Slice { .. }) {
-                        self.arrays_from_load.insert(result);
+                        let is_reference_param =
+                            self.dfg.block_parameters(block_id).contains(address);
+                        self.arrays_from_load.insert(result, is_reference_param);
```
aakoshh, line 147: Can this ever be transitive? 

I'm looking at this initial SSA of a push:
```
acir(inline) fn push f2 {
  b0(v0: &mut [Field; 3], v1: &mut u32, v2: Field):
    v3 = load v0
    inc_rc v3
    v4 = load v0
    v5 = load v1
    v7 = lt v5, u32 3
    constrain v7 == u1 1 '"push out of bounds"'
    v9 = load v0
    inc_rc v9
    v10 = load v0
    v11 = load v1
    v12 = load v0
    v13 = load v1
    v14 = array_set v10, index v13, value v2
    v16 = add v13, u32 1
    store v14 at v0
    store v11 at v1
    v17 = load v0
    inc_rc v17
    v18 = load v0
    v19 = load v1
    v20 = load v0
    v21 = load v1
    v22 = add v21, u32 1
    store v18 at v0
    store v22 at v1
    v23 = load v0
    dec_rc v23
    return 
}
```
I can see that `v0` is loaded multiple times, for example:
```
    v10 = load v0
    v12 = load v0
    v13 = load v1
    v14 = array_set v10, index v13, value v2
    store v14 at v0
    v17 = load v0
    inc_rc v17
    v18 = load v0
```
I'm wondering if there was another `array_set` on, say, `v17` on `v18`, would that be mutable because it's not an input to `b0` like `v0` is?

vezenovm: This checks whether the load of the address comes from a parameter. So `v0` would still be what is checked for `v17` and `v18`.

aakoshh (approved):
The test I did passes with this fix 🎉

Issue #6347, opened 2024-10-24
Type aliases of structs cannot be destructured similar to the underlying struct.


TomAFrench:
### Aim

I'd like to use a type alias to destructure a struct value as if I were using the original struct name.

### Expected Behavior

I should then be able to compile the below program
```noir
struct Foo {
    inner: Field
}

type Bar = Foo;

fn main() {
    let Bar { inner } = Foo { inner: 42 };
    assert_eq(inner, 42);
}
```

### Bug

I get a unification error where instead of resolving to the type we halt after finding that we have a type alias.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #6356, opened 2024-10-25
fix: allow type aliases in let patterns

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6347

## Summary\*

We were calling `lookup` only expecting a StructId, so I added another case for a type alias in general.

## Additional Context

If the type alias resolves to a non-struct type we still issue an error. E.g. if we had `type Bar = ();`

```
  ┌─ src/main.nr:8:9
  │
8 │     let Bar { inner } = Foo { inner: 42 };
  │         ------------- () has no fields to construct it with
  │
```

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/scope.rs
+++ b/compiler/noirc_frontend/src/elaborator/scope.rs
@@ -269,11 +269,25 @@ impl<'context> Elaborator<'context> {
             }
         }
 
-        match self.lookup(path) {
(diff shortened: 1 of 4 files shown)


asterite (approved):
I was just working on that same scope.rs file and noticed that same, though I didn't know if it was intended or not. Looks good!

PR #6375, opened 2024-10-28
feat: do not increment reference counts on arrays through references

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

@jfecher mentioned in 1:1 that we should not be incrementing the reference count here.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
I don't think that the external repo checks will help here as we only run `nargo check` so any errors from SSA wouldn't be caught.

jfecher:
We seem to be getting unrelated errors there anyway

jfecher (approved):

PR #6385, opened 2024-10-29
feat(profiler): Add Brillig procedure info to debug artifact for more informative profiling

vezenovm:
# Description

## Problem\*

Resolves 

Part of general effort to reduce the Brillig execution trace.

## Summary\*

In #6253 we added a command to the noir profiler that allows us to generate a flamegraph of the opcodes executed by the VM. However, the current debug metadata on its own does not provide a full picture. We have reusable procedures in Brillig for common operations such as mem copy, array copy, checking the max stack depth, etc. As procedures are themselves made up of multiple lower level opcodes procedures are executed in a similar way to other calls. This makes it difficult to discern which opcodes come from these common reusable procedures.

This PR adds a mapping of procedure ids to the range of opcodes that they occupy inside of a Brillig artifact. 

The `sha256_regression` test from master:
image

Here is an example of the `sha256_regression` test:
Screenshot 2024-10-29 at 12 22 54 AM

You can see how multiple opcodes with previously unknown locations are now associated with a mem copy procedure (in fact it is 15% of the entire program, which in itself provides us with some valuable insights that can be addressed in optimization follow-ups).

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -70,6 +70,9 @@ pub struct SsaEvaluatorOptions {
 
     /// The higher the value, the more inlined brillig functions will be.
     pub inliner_aggressiveness: i64,
+
+    /// Collect extra debug information for more informative profiling.
+    pub profiling_active: bool,
```
TomAFrench, line 75: Is this necessary? It doesn't seem like it's a ton of extra work to add this debug info (plus it's at compile time rather than at execution).

Unless the debug info generated is truly huge I think that we can just always generate debug data for procedures.

vezenovm: It isn't a ton of extra work, it just may get large if we have many Brillig functions. But either way it wouldn't be _that_ large. A maximum of two usizes to represent the procedure range for each procedure for each function.

vezenovm: I'll remove the flag

vezenovm: Removed

acvm-repo/acir/src/circuit/brillig.rs
```diff
@@ -46,3 +49,66 @@ impl std::fmt::Display for BrilligFunctionId {
         write!(f, "{}", self.0)
     }
 }
+
+/// Procedures are a set of complex operations that are common in the noir language.
+/// Extracting them to reusable procedures allows us to reduce the size of the generated Brillig.
+/// Procedures receive their arguments on scratch space to avoid stack dumping&restoring.
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash, PartialOrd, Ord, Deserialize, Serialize)]
+pub enum ProcedureId {
```
TomAFrench, line 57: Can you move this up into the compiler? Procedures are Noir-only constructs and so shouldn't be in the ACIR crate.

vezenovm: Yeah I was unsure on this change. I did this due to `ProcedureId` being added to the `DebugInfo` which is in `noirc_errors`. I'll make a separate ID mapping for `ProcedureId`.

vezenovm: I'll look at making a separate ID mapping for procedure debug info.

compiler/noirc_evaluator/src/ssa/acir_gen/acir_ir/generated_acir.rs
```diff
@@ -72,13 +73,24 @@ pub(crate) struct GeneratedAcir<F: AcirField> {
     /// As to avoid passing the ACIR gen shared context into each individual ACIR
     /// we can instead keep this map and resolve the Brillig calls at the end of code generation.
     pub(crate) brillig_stdlib_func_locations: BTreeMap<OpcodeLocation, BrilligStdlibFunc>,
+
+    /// Flag that determines whether we should gather extra information for profiling
+    /// such as opcode range of Brillig procedures.
+    pub(crate) profiling_active: bool,
```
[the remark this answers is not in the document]
vezenovm: Yes deleted

acvm-repo/acir/src/circuit/brillig.rs
vezenovm: I have a new `ProcedureDebugId` in `noirc_errors` that we convert into when creating the brillig procedures debug data in ACIR and then convert back into a `ProcedureId` in the flamegraph.

compiler/noirc_evaluator/src/brillig/brillig_ir/procedures/mod.rs
```diff
@@ -45,6 +47,64 @@ pub(crate) enum ProcedureId {
...
+            ProcedureId::ArrayReverse => 1,
+            ProcedureId::VectorCopy => 2,
+            ProcedureId::MemCopy => 3,
+            ProcedureId::PrepareVectorPush(flag) => {
+                if flag {
+                    4
+                } else {
+                    5
+                }
+            }
```
asterite, line 63:
```suggestion
            ProcedureId::PrepareVectorPush(true) => 4,
            ProcedureId::PrepareVectorPush(false) => 5,
```

compiler/noirc_evaluator/src/brillig/mod.rs
```diff
@@ -43,7 +45,11 @@ impl Brillig {
                 self.ssa_function_to_brillig.get(&function_id).map(Cow::Borrowed)
             }
             // Procedures are compiled as needed
-            LabelType::Procedure(procedure_id) => Some(Cow::Owned(compile_procedure(procedure_id))),
+            LabelType::Procedure(procedure_id) => {
+                let mut artifact = compile_procedure(procedure_id);
+                artifact.procedure = Some(procedure_id);
+                Some(Cow::Owned(artifact))
```
asterite, line 51: Would it make sense to assign `procedure` inside `compile_procedure`? I see it's also called in a different place but I don't know if that other place should have it set or not.

asterite (approved):
Looks good! I left a few minor comments.

compiler/noirc_evaluator/src/brillig/mod.rs
vezenovm: The other place is just a test you are right it is more natural in `compile_procedure`. I moved the setting of the procedure setting to the instantiation of the Brillig context and artifact in `new_for_procedure`.

PR #6388, opened 2024-10-29
feat: error if generic type parameter in impl is not mentioned in self type

### LINKED ISSUE #5855: Concrete types do not throw an error if an impl is defined with generics
Here we've got a non-generic struct with a generic impl

https://github.com/AztecProtocol/aztec-packages/blob/e03ea0bd716ccb21ad94414ea393a742dd7f5a65/noir-projects/noir-protocol-circuits/crates/types/src/merkle_tree/variable_merkle_tree.nr#L24

This seems very weird and we should error on this forcing the generic to be moved onto the individual methods.

asterite:
# Description

## Problem

Resolves #5855

## Summary

I decided to do this check separately... that is, another way could have been to keep track of which generics we saw while we solve `self_type`, but I think that would lead to code that's harder to understand. I was also able to use the visitor pattern to quickly find the mentioned generic types (I had to update Visitor so that it correctly traverses unresolved type expressions).

## Additional Context

Some questions:
1. We only produce this error for `impl`... do we also want to do it for trait impls? If yes, in this PR or a separate one?
2. Should we use the same error message as Rust? I find it a bit hard to understand, although it kind of makes sense. Let me know!

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/tests.rs
```diff
@@ -4302,3 +4300,19 @@ fn cannot_determine_type_of_generic_argument_in_enum_constructor() {
...
+        impl<T, U> Foo<T> {}
+                ^ The type parameter `U` is not constrained by the impl trait, self type, or predicates
+                ~ Hint: remove the `U` type parameter
+
+        fn main() {
+            let _ = Foo {};
+        }
+        "#;
+    check_errors(src);
+}
```
michaeljklein, line 4318: Simplified form of the original failing case:

```suggestion
}

#[test]
fn unconstrained_numeric_generic_in_impl() {
    let src = r#"
        pub struct Foo {}
        impl<let N: u32> Foo {}
                 ^ The type parameter `N` is not constrained by the impl trait, self type, or predicates
                 ~ Hint: remove the `N` type parameter
        fn main() {
            let _ = Foo {};
        }
        "#;
    check_errors(src);
}
```

michaeljklein (changes requested):
LGTM, but please add a test that uses a numeric generic as in the original issue, e.g. similar to the one I suggested (untested).

> We only produce this error for `impl...` do we also want to do it for trait impls? If yes, in this PR or a separate one?

Yes, I think we also want this error for trait impls and a separate PR would be fine

> Should we use the same error message as Rust? I find it a bit hard to understand, although it kind of makes sense. Let me know!

I think the current error is clear

compiler/noirc_frontend/src/tests.rs
asterite: Done!

michaeljklein (approved):

PR #6393, opened 2024-10-29
fix: make keccak256 work with input lengths greater than 136 bytes

### LINKED ISSUE #6130: `std::hash::keccak256` cannot hash more than a single block (136 bytes input)
`std::hash::keccak256` is currently written implying that it will support inputs of multiple blocks however attempting to provide inputs in excess of 136 bytes will fail as we attempt to write them into an array with the capacity to hold only a single block.

We should add testing for multi-block hash inputs and fix this issue.

asterite:
# Description

## Problem

Resolves #6130

## Summary

By looking at the algorithm, it seems the initial phase is copying `input` to `block_bytes`. I modified `block_bytes` to be at least of length N, but padded up to a multiple of 136 bytes.

## Additional Context

I read a bit about keccak, checking the algorithm in pseudo-code, C, Python, and all of them seem to start after our comment that says "//1. format_input_lanes". The part that's before that seems to be copying the input bytes to a different array, and it was failing because that array was shorter than the input... so I fixed that, but I don't know if the fix should have been somewhere else. Let me know! 🙏  

Oh, I also removed a few unused imports... they were producing noise when running the Noir tests and it was a bit distracting.

## Documentation

Check one:

# PR Checklist


asterite:
Side note: I found that the algorithm can be done without `Vec`... though doing `nargo info` doesn't change. Is it worth making that change, though?

TomAFrench:
> Side note: I found that the algorithm can be done without Vec... though doing nargo info doesn't change. Is it worth making that change, though?

Replacing slices with arrays can help the brillig codegen so this may still be worthwhile.

asterite:
Should I do that in this PR or a separate one?

TomAFrench (approved):
LGTM. I'm happy merging this as is but we can also roll the other changes into here.

asterite:
I'll do it in a separate PR, given that CI is green :-)

PR #6400, opened 2024-10-30
fix: right shift is not a regular division

### LINKED ISSUE #6201: Operator `>>` does arithmetic or logical shift, depending on rhs
### Aim

Attempted to create a simple program which right-shifts a number.

The following `main.nr`:

```noir
fn main(a: i16) {
    println(a >> 3);
}
```

with `Prover.toml`:

```toml
a = "-769"
```

Results with an arithmetic `>>`, outputing `-96`.

However, the following `main.nr`:

```noir
fn main(a: i16, b: u8) {
    println(a >> b);
}
```

with `Prover.toml`:

```toml
a = "-769"
b = 3
```

Results with a logical `>>`, outputting `8095`.

### Expected Behavior

`>>` to act consistently on the same value

### Bug

`>>` with a constant right hand-side behaves as an arithmetic shift, while with an input value rhs it behaves as a logical shift

### To Reproduce

1. Create two `Noir` projects, for each pair of program and `Prover.toml`
2. Run them with `nargo execute`
3. Observe the different output

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

### Nargo Version

nargo version = 0.34.0 noirc version = 0.34.0+94ed0b8fb13a923a59f6f3ccd8d7c75e191deeca (git version hash: 94ed0b8fb13a923a59f6f3ccd8d7c75e191deeca, is dirty: false)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #6201

## Summary\*
Fix right shift behaviour to match rust's. 

## Additional Context
Regular division performs rounding towards 0 while arithmetic right shift performs a division with rounding towards negative infinite.

## Documentation\*

Check one:

# PR Checklist\*


test_programs/execution_success/bit_shifts_runtime/src/main.nr
```diff
@@ -1,4 +1,4 @@
-fn main(x: u64, y: u8) {
+fn main(x: u64, y: u8, a: i16) {
```
jfecher, line 1: Looks like this test is missing the `a` argument in Prover.toml

jfecher (approved):

test_programs/execution_success/bit_shifts_runtime/src/main.nr
```diff
@@ -17,4 +17,6 @@ fn main(x: u64, y: u8) {
     assert(a << y == -2);
 
     assert(x >> (x as u8) == 0);
+
+    assert_eq(a >> 3, -97);
```
jfecher, line 21: `a` here should be `-1` due to the `let mut a: i8 = y as i8; a = -a;` above.

Also, it looks like we can't have negatives in the Prover.toml unfortunately:
```
input file is badly formed, could not parse, TOML parse error at line 3, column 5
  |
3 | a = -769
  |     ^^^^
data did not match any variant of untagged enum TomlTypes
```

jfecher (changes requested):
Revoking previous approval since fixing the test seems a bit more involved than I guessed.

Could probably still do a quick fix though by multiplying `a` by `-769` so that it has the same value as `a` in bit_shifts_comptime.

test_programs/execution_success/bit_shifts_runtime/src/main.nr
guipublic: We can have negative numbers in Prover.toml, with double quotes: "-769"
The 'a' was copy/pasted from another file where I was testing, I did not see it was already used here... I have renamed it.

jfecher: Looks like bit_shifts_runtime is still failing

PR #6404, opened 2024-10-30
fix: distinguish TypePath with and without turbofish

### LINKED ISSUE #6188: Compiler doesn't infer generics on `Field::from_be_bytes` correctly
https://github.com/noir-lang/noir/pull/6084 added the `Field::from_be_bytes` and `Field::from_le_bytes` functions. These are generic in their inputs however the compiler doesn't seem to be able to determine the value of `N` from the passed array.

This is likely due to the fact that it's an impl on a primitive type, so we should check for inconsistencies between how we handle these relative to struct impls.

asterite:
# Description

## Problem

Resolves #6188

## Summary

We couldn't distinguish between `Field::foo` and `Field::<>::foo`, assuming it was always `Field::<>::foo`, trying to resolve the generics and filling missing ones with `Type::Error`, eventually leading to a compiler error.

## Additional Context

While working on this I noticed that `Type::kind` might be returning the incorrect Kind for some types:

https://github.com/noir-lang/noir/blob/4e44e1f4a71a1a8efc17462c55a85b9cb7fdeae3/compiler/noirc_frontend/src/hir_def/types.rs#L1379

I talked about this with Michael, I might send a follow-up PR if I can find some code that doesn't work if we don't change that.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/parser/parser/expression.rs
+++ b/compiler/noirc_frontend/src/parser/parser/expression.rs
@@ -4,7 +4,7 @@ use noirc_errors::Span;
 use crate::{
     ast::{
         ArrayLiteral, BlockExpression, CallExpression, CastExpression, ConstructorExpression,
-        Expression, ExpressionKind, GenericTypeArgs, Ident, IfExpression, IndexExpression, Literal,
+        Expression, ExpressionKind, Ident, IfExpression, IndexExpression, Literal,
         MemberAccessExpression, MethodCallExpression, Statement, TypePath, UnaryOp, UnresolvedType,
     },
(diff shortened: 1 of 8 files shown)


compiler/noirc_frontend/src/parser/parser/expression.rs
```diff
@@ -553,9 +553,9 @@ impl<'a> Parser<'a> {
             if generics.is_empty() {
                 self.expected_token(Token::Less);
             }
-            generics
+            Some(generics)
```
jfecher, line 556: Can you change the `if self.eat_double_colon() {` above to `self.eat_double_colon().then(|| {` ? :cowboy_hat_face:

jfecher (approved):
LGTM

PR #6405, opened 2024-10-30
fix: numeric generic doesn't have a default type

### LINKED ISSUE #6379: Compiling a function returning an unknown numeric generic panics
### Aim

Attempted to compile the following program with a function that outputs a numeric generic:

```noir
// Using "[(); N]" as a parameter compiles and executes successfully
//
// Using "struct W<N> {}" as a parameter also panics if "foo" is called on "W {}",
// but not if "foo" is called on "let w_1: W<1> = W {};"
fn foo<let N: u32>() -> u32 {
    N
}

fn main() {
    foo();
}
```

### Expected Behavior

The program to compile successfully or fail with a user error

### Bug

Compiling the program panics with the following message:

```bash
Message:  Non-numeric type variable used in expression expecting a value
Location: compiler/noirc_frontend/src/monomorphization/mod.rs:934
```

### To Reproduce

1.
2.
3.
4.

### Workaround

Yes

### Workaround Description

Use turbofish (e.g. `foo::<3>()`) to specify numeric generics when they can't be derived from a parameter's type

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 0.36.0 noirc version = 0.36.0+2f376100d3ee7ab519d6ea30153395bb3e7af7b1 (git version hash: 2f376100d3ee7ab519d6ea30153395bb3e7af7b1, is dirty: false)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #6379

## Summary

Even though `Kind::Numeric` holds a type, say, `u32`, that type is not the default value that should be used if we can't bind that var.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir_def/types.rs
+++ b/compiler/noirc_frontend/src/hir_def/types.rs
@@ -241,11 +241,9 @@ impl Kind {
     /// during monomorphization.
     pub(crate) fn default_type(&self) -> Option<Type> {
         match self {
-            Kind::Any => None,
             Kind::IntegerOrField => Some(Type::default_int_or_field_type()),
             Kind::Integer => Some(Type::default_int_type()),
-            Kind::Normal => None,
-            Kind::Numeric(typ) => Some(*typ.clone()),
+            Kind::Any | Kind::Normal | Kind::Numeric(..) => None,
         }
     }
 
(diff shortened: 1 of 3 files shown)


jfecher (approved):
This makes sense, good find :+1:

PR #6406, opened 2024-10-30
chore(nargo): Remove old profile as part of info cmd

vezenovm:
# Description

## Problem\*

Just something I noticed while working on `nargo info`. The `--profile-info` flag is hidden and I am willing to bet largely unused. We can move to the noir-profiler under `tooling/profiler` as it is much more informative and easier to navigate. 

## Summary\*

I have removed `--profile-info` from `nargo info` and the subsequent methods related to it. 

## Additional Context

I would also like to remove the "Profile" codelens @asterite @TomAFrench. I was trying to use it on `1_mul` and `sha256_regression` and nothing would happen while the other code lens commands worked. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/nargo_cli/src/cli/info_cmd.rs
+++ b/tooling/nargo_cli/src/cli/info_cmd.rs
@@ -1,13 +1,10 @@
-use std::collections::HashMap;
-
 use acvm::acir::circuit::ExpressionWidth;
 use clap::Args;
 use iter_extended::vecmap;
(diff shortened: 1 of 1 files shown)


asterite (approved):

vezenovm:
I was going to add the LSP code lens removal here, but you as you already approved we can just keep the nargo and LSP chores separate.

Issue #4963, opened 2024-05-02
Add foreign call definitions into


TomAFrench:
I've thought that we should add this for a while but this is shown by #4959.

We currently don't store any info for the oracle calls which a circuit will make in the ABI which prevents us from automatically ABI encoding/decoding arguments/return values. Having this information would also allow us to throw better errors when trying to resolve an oracle call which needs a response when there is no resolver for it.

During compilation we should keep track of all the oracle calls which the program may make and add these to the ABI.

TomAFrench:
Once this is done we can also replace the name of the oracle call which is stored in the byte code with just an identifier.

TomAFrench:
Oof, yeah this is a bit unfortunate. We'd have to generate the ABIs for the monomorphized variants of these functions instead so generate for the function `avmOpcodeCalldataCopy<10>` or something.

TomAFrench:
hmmm, this would get be a pretty breaking change however and would screw up aztec's handling of oracle calls a lot.

PR #6412, opened 2024-10-30
feat: include oracles in ABI

asterite:
# Description

## Problem

For #4963

## Summary

Pending defining what to do with generic types: https://github.com/noir-lang/noir/issues/4963#issuecomment-2447792757

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
@asterite I think we're going to have to pull the oracle calls out from the monomorphized AST instead here. This would mean that we can't build the ABI from the HIR alone which is a little sad.

asterite:
Now oracles are computed during monomorphization. The next issues are:
1. Do we need the parameter names of the oracle functions, or are the types alone enough for this?
2. It seems oracles can receive slices and `fmtstr`. These aren't currently allowed in the ABI. Should they?

PR #6413, opened 2024-10-30
feat: nargo command to generate shell completions

### LINKED ISSUE #6284: Setup `bash`/`zsh` command-line completion for `nargo`
### Problem

Tab-completion in [bash](https://www.gnu.org/software/gnuastro/manual/html_node/Bash-TAB-completion-tutorial.html) and zsh for `nargo` would be helpful so that typing partial commands like `nargo co<TAB>` or `nargo n<TAB>`, the commands are completed to `nargo compile`, `nargo new`, respectively.

### Happy Case

Setup tab completion for the commonly used shells `bash` and `zsh`

### Workaround

Yes

### Workaround Description

1. Use "short" versions of commands where available (e.g. `check`, `execute`) or else enable them for other commands
2. Type the whole command

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #6284

## Summary

nargo-shell-completion

## Additional Context

I copied the command name, and learned how to install shell completions, by reading this: https://apple.github.io/swift-argument-parser/documentation/argumentparser/installingcompletionscripts/

I don't know how we can teach users to install completions... maybe some know... for those who don't, maybe they stumble upon the above page. Or should we document this in our documentation? Or in the command itself? (maybe too long)

## Documentation\*

Check one:

# PR Checklist\*


aakoshh:
Nice! It could be hinted at in the output of `noirup`; similar installers mention zsh and fish env vars and path settings. 

https://noir-lang.org/docs/reference/nargo_commands could be another good place to mention it.

asterite:
Oh, it seems that nargo_commands page is automatically generated from the doc comments... so to change it we'd have to also add those comments to the command and to its output, which is maybe not good.

Maybe this can go without docs... people wanting to install shell completions probably know how to install them.

aakoshh:
Yes, I realised that to include any hint in the reference would go against you trying to keep it from the command docs. Perhaps it's worth mentioning under https://github.com/noir-lang/noir/blob/master/docs/docs/getting_started/quick_start.md#nargo instead?

aakoshh:
There is also https://github.com/noir-lang/noir/blob/master/docs/docs/getting_started/noir_installation.md but it doesn’t mention `nargo`, so I thought the other one is a nicer place to make people aware of this goodie

asterite:
I added some docs. I created a new page under "Getting started" for it, then linked it from the "Quick start" and "Standalone Noir Installation" pages. I guess it makes sense that it's a bit more prominent than what I initially thought as it could be very helpful to have those completions.

PR #6415, opened 2024-10-31
feat: ensure that generated ACIR is solvable

### LINKED ISSUE #6389: Add a check for circuit solvability at the end of compilation.
We're currently getting a situation where some aztec contracts are having missing witness assignments: https://github.com/AztecProtocol/aztec-packages/issues/9488

As a final compilation step we should add a check where we ensure that all opcodes are solvable. Starting from the known set of input witnesses we can ensure that all inputs are known for each opcode by the time we reach it, adding the outputs of that opcode to the new set of known witnesses.

This will ensure that there's continuity in the ACIR circuits we generate without needing to know the inputs.

guipublic:
# Description

## Problem\*

Resolves #6389 

## Summary\*
Add a pass at the end of the ACIR transform to check that the result can be solved

## Additional Context
Minimal error handling as this pass should always pass unless there is an issue (i.e a compiler bug) during transformation.

## Documentation\*

Check one:

# PR Checklist\*


tooling/nargo/src/ops/transform.rs
```diff
@@ -1,21 +1,33 @@
...
+        let mut simulator = CircuitSimulator::default();
+        if !simulator.check_circuit(circuit) {
+            let diag = FileDiagnostic {
+                file_id: fm::FileId::dummy(),
+                diagnostic: CustomDiagnostic::from_message("ACVM simulation failed"),
+            };
+            return Err(vec![diag]);
+        }
+    }
+    Ok((compiled_program, Vec::new()))
```
TomAFrench, line 30: On second thoughts, I think we should split this into a separate function which the CLI can run the program through after compilation rather than having it directly integrated into the compilation pipeline.

TomAFrench: I mean the compilation flow not being compile, transform and then having this step included within transform when you do `nargo compile` it does compile, transform, check.

Thete should be a separate function within the nargo crate and the cli should call that after calling transform.

guipublic: I moved the check into a separate function (and file)

TomAFrench (approved):

PR #6424, opened 2024-10-31
chore: revamp attributes

asterite:
# Description

Follow-up to #6414

## Problem

Currently attributes, even "meta" ones (non "tag" ones) were parsed by just consuming all characters until hitting "]". That worked but:
- it didn't allow to use arrays in arguments
- I think in general we wanted to make meta attributes always having the `func(arg1, .., argN)` or just `func` syntax, and kind of enforcing that

## Summary

This PR changes how attributes are parsed and represented. Meta attributes must now always have a function syntax call. The rest of the attributes (the built-in ones) are parsed in a more strict way, generally leading to better error messages.

Some benefits of this PR are:
- We no longer need to parse an attribute to see if it's a function call
- Any kind of expression is allowed as a meta attribute argument
- Comments are now allowed inside attributes
- Attributes are formatted better
- LSP will offer better completions: because now the meta arguments are expressions, if you type `#[derive(E` it will offer `Eq` as a suggestion.

## Additional Context

I had to comment out a `#[deprecated = "..."]` attribute for two reasons:
1. That syntax was wrong (the correct one is `#[deprecated("...")]`
2. If we change it to use the correct syntax, we get deprecation warnings in the standard library... which is probably not something we want. I tried to defined an alternative method that isn't marked as deprecated and use it in the standard library (marked as `pub(crate)`) but some tests rely on that `new` method and I didn't know how to make it work otherwise.

Of course, let me know if the direction this PR is going isn't the desired one!

## Documentation

Check one:

# PR Checklist


jfecher:
> Of course, let me know if the direction this PR is going isn't the desired one!

I haven't seen the line by line changes yet but I'd rather keep the ability to call the functions without an explicit function call syntax. E.g. still allowing `#[aztec_private]` and not requiring `#[aztec_private()]`. I'd think we should still be able to support this so even while parsing these so I'm curious why it was removed?

asterite:
Oh, it should still be allowed. If not I might have made a mistake.

jfecher:
Ah, ok. I just assumed from the description it sounded like a function call form was required. Should be fine then

asterite:
Ah, makes sense. I updated the description.

jfecher (approved):
Looks great, thank you!

PR #6430, opened 2024-11-01
chore: split path and import lookups

asterite:
# Description

## Problem

Path and import lookup behave a bit differently. For example you shouldn't have aliases in the middle of a `use` segment (Rust says they are not modules), and generics can't show up in `use` statements. This made the code more complex than it should.

## Summary

This PR refactors the path lookup code and splits it into two, one to resolve imports and another to resolve paths in places other than imports.

This PR is mainly a refactor that doesn't change how things behave, except that now aliases are forbidden in the middle of a `use` segment (like in Rust).

The logic to resolve a Path is mostly shared in a `resolve_path_kind` function, which determines the plain Path and ModuleId where the path should be resolved in. Then traversing that plain path logic is kind of duplicated, once for imports and once for paths in code, but it kind of makes sense because their logic is a bit different and it will continue to diverge as we'll want to solve types with their generics in order to continue finding methods.

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/hir/resolution/visibility.rs
```diff
@@ -7,34 +7,42 @@ use std::collections::BTreeMap;
...
-// if the given pub(crate) function is being called from another crate. Otherwise returns true.
-pub fn can_reference_module_id(
+/// Returns true if an item with the given visibility in the target module
+/// is visible from the current module. For example:
+///
+/// mod foo {
+///     ^^^ <-- target module
+///   pub(crate) fn bar() {}
+///   ^^^^^^^^^^ <- visibility
+/// }
```
aakoshh, line 17:
```suggestion
/// ```text
/// mod foo {
///     ^^^ <-- target module
///   pub(crate) fn bar() {}
///   ^^^^^^^^^^ <- visibility
/// }
/// ```
```

compiler/noirc_frontend/src/hir/resolution/import.rs
```diff
@@ -159,452 +82,255 @@ impl<'a> From<&'a PathResolutionError> for CustomDiagnostic {
...
+            .get(&crate_name.0.contents)
+            .ok_or_else(|| PathResolutionError::Unresolved(crate_name.to_owned()))?;
+
+        if let Some(references_tracker) = &mut self.references_tracker {
+            let span = crate_name.span();
+            references_tracker.add_reference(ModuleDefId::ModuleId(*dep_module), span, false);
         }
+
+        // Create an import directive for the dependency crate
+        // XXX: This will panic if the path is of the form `use std`. Ideal algorithm will not distinguish between crate and module
```
TomAFrench, line 193: This panic message is stale, no?

TomAFrench (approved):
LGTM

Issue #6315, opened 2024-10-22
Trait aliases


michaeljklein:
### Problem

[Trait inheritance](https://github.com/noir-lang/noir/pull/6252) allows traits to require other traits, but explicitly including all required traits in functions or `impl`'s can become repetitive.

### Happy Case

Once https://github.com/noir-lang/noir/issues/6314 is fixed, support some version of trait aliases, e.g., by desugaring a trait alias declaration:

```noir
trait Baz = Foo + Bar;
```

into (approximately):

```noir
trait Baz: Foo + Bar {}

impl<T> Baz for T where T: Foo + Bar {}
```

Either support something similar to [this experimental Rust feature](https://doc.rust-lang.org/beta/unstable-book/language-features/trait-alias.html) or else ensure that it can be implemented with a simple Noir macro.

### Workaround

Yes

### Workaround Description

`trait Baz: Foo + Bar {}` can currently be defined, but currently needs an explicit `impl Baz for MyStruct` for each type

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

michaeljklein:
After discussing with Tom and Jake, we'd prefer to implement Rust-style trait aliases with the following notes:
1. The error messages when a trait is missing should be clear
2. Error messages when the trait alias is used in other places than intended should be clear
3. ^ If these error messages are tricky, use a first-order enum variant for trait aliases

PR #6431, opened 2024-11-01
feat: trait aliases

michaeljklein:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6315

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/parser/parser/traits.rs
```diff
@@ -12,34 +17,115 @@ use super::Parser;
...
 
         let generics = self.parse_generics();
-        let bounds = if self.eat_colon() { self.parse_trait_bounds() } else { Vec::new() };
-        let where_clause = self.parse_where_clause();
-        let items = self.parse_trait_body();
 
-        NoirTrait {
+        // Trait aliases:
+        // trait Foo<..> = A + B + E where ..;
+        let (bounds, where_clause, items, is_alias) = if self.eat_assign() {
```
asterite, line 40: Optional: instead of desugaring a trait alias in the parser itself, we could introduce a new AST node for it (maybe `TraitAlias`) and desugar it later on, maybe in the Elaborator or collector. That way:
- parsing is a bit simpler
- desugaring can be tested independently
- LSP or tools working with parsed code (like the formatter) don't need to know about how trait aliases are expanded, knowing that they need to skip the synthetic impl, and can focus on better handling this new node

michaeljklein: I originally attempted to implement this by desugaring it in the `Elaborator` (based on the `is_alias` flag), but it was fairly complex to deal with post-parsing types.
- Testing this approach seems a bit simpler since we can check ~equality up to spans between the parsed trait impl and the expected desugaring.

I suppose adding an "is_synthetic: bool` to `NoirTraitImpl` would be easy if that helps handling it in the LSP?

Sure, I'll add a formatter test for this 👍

michaeljklein: @asterite I've added `is_synthetic: bool` to `NoirTraitImpl` and formatter support for trait aliases, with tests for both.

compiler/noirc_frontend/src/parser/parser/parse_many.rs
```diff
@@ -27,6 +40,26 @@ impl<'a> Parser<'a> {
...
+                vec![result]
+            } else {
+                vec![]
+            }
+        };
+        self.parse_many_to_many_return_trailing_separator_if_any(items, separated_by, f)
+    }
+
+    /// Same as parse_many, but returns a bool indicating whether a trailing separator was found.
+    pub(super) fn parse_many_to_many_return_trailing_separator_if_any<T, F>(
```
TomAFrench, line 55:
```suggestion
    fn parse_many_to_many_return_trailing_separator_if_any<T, F>(
```

TomAFrench (approved):
Looks good to me. We've got a number of desugarings happening in the parser atm so I don't think it's blocking to have this here. We can always refactor this to how @asterite mentioned in a later PR.

PR #6073, opened 2024-09-17
feat: avoid branch condition negations in `ValueMerger`

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR modifies how we merge values from `pred * a + (1-pred) * b` to `b + pred * (a-b)` which avoids another multiplication. This is being done manually in [here](https://github.com/noir-lang/noir_string_search/blob/a0a7b9818657da7d447dec43b625c6d2748934ea/src/utils.nr#L35-L49) so it's worth considering as the default merger strategy.

## Documentation\*

Check one:
- [ ] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [ ] I have tested the changes locally.
- [ ] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #6434, opened 2024-11-01
fix: Fix poor handling of aliased references in flattening pass causing some values to be zeroed

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/5771

### Background

Fixes a bug which originates from `merge_stores` in the flattening pass which can in certain cases get confused with aliases and overwrite values incorrectly.

The rough steps to reproduce this issue are as follows, though I've been unsuccessful so far in reproducing it from scratch outside of the original [branch in aztec packages](https://github.com/AztecProtocol/aztec-packages/tree/janb/context-bug-reproduction).
1. Create a mutable reference to something, for example an integer (`context.side_effect_counter` in the original bug)
2. Use a runtime conditional `if input_to_main { ... }`, and ensure that `input_to_main = false`
3. Within the conditional try to confuse the compiler by creating a mutable alias to the reference in a way that the alias only appears after the first mem2reg pass and before the flattening pass. In this case, using an immediately-invoked closure which captures the reference as part of its environment.
4. Use some oracles or other functions which can't be inlined/evaluated currently within that closure.
5. Create another nested if within the closure, also using a runtime condition
6. Inside that if, conditionally mutate the captured reference.
7. Back to the first if, after the closure is invoked, at the end of the if, mutate the original, non-captured reference.

This somewhat specific setup lead to a situation where the flattening pass saw it had a store to what it saw as two different references, but were really both aliases to the same reference. `merge_stores` uses a `then_value`, `else_value`, and `old_value` of each reference to create its `new_value` which it then stores back to the reference. One alias in this case had a correct `then_value`, while the other had a correct `else_value`, with the opposite value of both being equal to the `old_value`. Storing both of these in sequence lead to the final value for the else/false case to be `!input_to_main * (input_to_main * else_value)`. Since one of `input_to_main` and its negation are always zero, the whole value ends up zeroed for the else case. Finally, confusion with aliases also leads the compiler to accept the else case value as the updated value even after the `if` terminates.

I've been unsuccessful trying to reproduce this from scratch so come Monday I'll work on copying over the same test case into an integration test here. I've already removed all macro code but need to figure out how to translate the oracle functions step mainly.

### The Fix

Skipping some alternate fixes I tried, I figured out we could just remove `merge_stores` entirely if we never stored a non-conditional value to a reference to begin with during flattening. E.g. currently in flattening if we see `if c { foo = 3; }` we issue a `store 3 in foo` then after the if we go back and merge stores to issue a `store (if c then 3 else prev_value) in foo` afterward. I've changed it to instead issue `prev_value = load foo; store (if c then 3 else prev_value) in foo` to replace the original store. This means the value of a reference is always correct and we never have to fix it afterward. This greatly simplifies the flattening pass I think.

HOWEVER:
- The first store to an allocation has no previous value: `foo = allocate; store 3 in foo`, so I had to add an `initial_value: ValueId` argument to the Allocate instruction.
  - Not requiring an initial store breaks the mem2reg pass... In particular the logic to remove the last store is breaking. For this reason I'll probably try to revert the Allocate argument change in favor of a hack to not insert a previous value load if allocate is the instruction previous to the current store since we always issue them one after another in practice.
- Issuing an `Instruction::IfElse` for every store leads to more value merging in practice, resulting in increase constraint counts without something like https://github.com/noir-lang/noir/pull/6073.

## Additional Context

Draft because I had to port this over from local changes within aztec packages ([branch jf/fix-zeroed-bug](https://github.com/AztecProtocol/aztec-packages/tree/jf/fix-zeroed-bug)) and there are a few merges that still need to be handled carefully - most notably the mem2reg changes.

Remaining work
- [ ] Reproduce test
- [x] Revert `Instruction::Allocate` changes in favor of hack
- [x] Fix remaining merge diffs from this PR since it was ported from aztec-packages

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
The size of this draft looks large but the actual changes are fairly small and it is mostly removed code. Most of the code size changes are due to merge diffs and the Instruction::Allocate change I'll revert.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -378,11 +378,13 @@ impl<'f> PerFunctionContext<'f> {
...
+                // This causes errors in the sha256 tests
+                //
                 // If there was another store to this instruction without any (unremoved) loads or
                 // function calls in-between, we can remove the previous store.
-                if let Some(last_store) = references.last_stores.get(&address) {
-                    self.instructions_to_remove.insert(*last_store);
-                }
+                // if let Some(last_store) = references.last_stores.get(&address) {
+                //     self.instructions_to_remove.insert(*last_store);
+                // }
```
jfecher, line 387: Had to comment out this portion of mem2reg as it was causing the sha256 tests to fail. There seems to be an issue with tracking `last_stores` for some reason now.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -386,15 +387,18 @@ impl Instruction {
             | IfElse { .. }
             | ArraySet { .. } => true,
 
+            // Store instructions must be removed by DIE in acir code, any load
+            // instructions should already be unused by that point.
+            Store { .. } => matches!(function.runtime(), RuntimeType::Acir(_)),
```
jfecher, line 392: With the change to mem2reg, stores weren't being removed otherwise for some reason (even though the comment claims to just remove certain unneeded stores). So I had to change DIE to remove all stores here.

jfecher:
The brillig size increases are likely from the temporary store changes while mem2reg is fixed (should it be in this PR or separate?)
while the acir changes are likely due to the increased amount of value mergers which could possibly be helped by https://github.com/noir-lang/noir/pull/6073.

jfecher:
Some large bytecode regressions here unfortunately. This is a correctness issue though and the bytecode size fixes in https://github.com/noir-lang/noir/pull/6460 and https://github.com/noir-lang/noir/pull/6490 are looking less certain so I think we should merge this now. We can continue working on these fixes or other ones in the future.

TomAFrench:
Looking at this now. Merging master as there's some strange changes which I would expect to be in master already.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -386,15 +387,21 @@ impl Instruction {
             | IfElse { .. }
             | ArraySet { .. } => true,
 
+            // Store instructions must be removed by DIE in acir code, any load
+            // instructions should already be unused by that point.
+            Store { .. } => {
+                matches!(function.runtime(), RuntimeType::Acir(_))
+                    && function.reachable_blocks().len() == 1
+            }
+
```
TomAFrench, line 396: This is assuming some knowledge about the significance of only having a single block on the part of the reader. Can you flesh this out for the benefits of auditors?

TomAFrench (approved):
Looks good to me

PR #6449, opened 2024-11-04
fix: disallow `#[test]` on associated functions

### LINKED ISSUE #6435: Disallow applying the `#[test]` attribute to impl methods
We recently had an issue in the bignum library not registering that we had multiple tests with the same name due to the tests being defined inside of an impl (for "hacky forcing the compiler to do things it doesn't want to do" reasons).

I don't see a good reason to have tests be anything other than free functions so we should error in cases where they're attached to impls.

asterite:
# Description

## Problem

Resolves #6435

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/hir/def_collector/errors.rs
```diff
@@ -291,6 +293,12 @@ impl<'a> From<&'a DefCollectorErrorKind> for Diagnostic {
                 diag
             }
             DefCollectorErrorKind::UnsupportedNumericGenericType(err) => err.into(),
+            DefCollectorErrorKind::TestOnAssociatedFunction { span } => Diagnostic::simple_error(
+                "The `#[test]` attribute may only be used on a non-associated function".into(),
```
michaeljklein, line 297: What's a "non-associated function"? Closures are a sort of function too, but they don't allow `#[test]` either.

```suggestion
                "The `#[test]` attribute is disallowed on `impl` methods".into(),
```
?

michaeljklein (changes requested):

compiler/noirc_frontend/src/hir/def_collector/errors.rs
asterite: Sounds good. I copied the message from Rust 😊

jfecher (approved):

asterite:
@TomAFrench Let me know if this can be merged. I think with the current macros we can workaround how it's done in the bignum library.

michaeljklein (approved):
LGTM

TomAFrench (approved):

TomAFrench:
Thanks!

PR #6450, opened 2024-11-05
fix: check infix expression is valid in program input

### LINKED ISSUE #6438: Can't pass concrete struct to `main`
Please see this PR which exemplifies the issue https://github.com/zkpassport/noir_rsa/pull/2. See the section "A note to the Noir team".

Edit: here's a dedicated repo focusing on the problem: https://github.com/iAmMichaelConnor/example_sized_types_entrypoint_bug

I'll see if I can come up with a more minimal example.

asterite:
# Description

## Problem

Resolves #6438

## Additional Context

This doesn't fix the panic that's mentioned [here](https://github.com/zkpassport/noir_rsa/pull/2) though, I'll see if I can find out why that happens.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/tests.rs
+++ b/compiler/noirc_frontend/src/tests.rs
@@ -3623,3 +3623,15 @@ fn use_type_alias_to_generic_concrete_type_in_method_call() {
     "#;
     assert_no_errors(src);
 }
+
+#[test]
+fn allows_struct_with_generic_infix_type_as_main_input() {
+    let src = r#"
+        struct Foo<let N: u32> {
+            x: [u64; N * 2],
+        }
+
+        fn main(_x: Foo<18>) {}
+    "#;
+    assert_no_errors(src);
+}
--- a/compiler/noirc_frontend/src/tests.rs
+++ b/compiler/noirc_frontend/src/tests.rs
@@ -3625,13 +3625,39 @@ fn use_type_alias_to_generic_concrete_type_in_method_call() {
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


compiler/noirc_frontend/src/tests.rs
```diff
@@ -3623,3 +3623,15 @@ fn use_type_alias_to_generic_concrete_type_in_method_call() {
...
 }
+
+#[test]
+fn allows_struct_with_generic_infix_type_as_main_input() {
+    let src = r#"
+        struct Foo<let N: u32> {
+            x: [u64; N * 2],
+        }
+
+        fn main(_x: Foo<18>) {}
```
michaeljklein, line 3634: Does this work with cases e.g. `fn main(_x: Foo<2 * 9>)` as well?

michaeljklein (changes requested):
Possible extra test, otherwise LGTM

compiler/noirc_frontend/src/tests.rs
jfecher: Or `(_x: Foo<N * 2>)` where `N` is a constant global?

asterite: It seems both work. Should I add tests for those too?

TomAFrench: Yeah, more tests never hurts.

asterite: Done!

michaeljklein (approved):
LGTM

PR #6454, opened 2024-11-05
fix: don't crash on AsTraitPath with empty path

asterite:
# Description

## Problem

Code like `struct Foo { x: <Y> }` would crash the compiler.

## Summary

`<Y>` in a type position is parsed as an incomplete `AsTraitPath` with an empty Path. The reason for this is that it's useful for LSP: if typing inside `Y` we can complete that trait, even though the `as` and the path are missing. However, some code in the Elaborator checks if a path's first name is `Self` without checking first if the path is not empty, leading to a crash. This PR fixes that.

While typing in an editor, if you tried to rename the type `Foo<N>` by deleting "Foo" first, it would lead to crashing the LSP server.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/path_resolution.rs
+++ b/compiler/noirc_frontend/src/elaborator/path_resolution.rs
@@ -107,7 +107,10 @@ impl<'context> Elaborator<'context> {
     pub(super) fn resolve_path(&mut self, mut path: Path) -> PathResolutionResult {
         let mut module_id = self.module_id();
 
-        if path.kind == PathKind::Plain && path.first_name() == SELF_TYPE_NAME {
+        if path.kind == PathKind::Plain
+            && !path.segments.is_empty()
+            && path.first_name() == SELF_TYPE_NAME
+        {
             if let Some(Type::Struct(struct_type, _)) = &self.self_type {
                 let struct_type = struct_type.borrow();
                 if path.segments.len() == 1 {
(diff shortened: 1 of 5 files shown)


compiler/noirc_frontend/src/elaborator/path_resolution.rs
```diff
@@ -107,7 +107,10 @@ impl<'context> Elaborator<'context> {
     pub(super) fn resolve_path(&mut self, mut path: Path) -> PathResolutionResult {
         let mut module_id = self.module_id();
 
-        if path.kind == PathKind::Plain && path.first_name() == SELF_TYPE_NAME {
+        if path.kind == PathKind::Plain
+            && !path.segments.is_empty()
+            && path.first_name() == SELF_TYPE_NAME
```
jfecher, line 112: Maybe `first_name` should default to `""` if the segments are empty?
That way this'd likely fix any other uses of `first_name` as well.

asterite: I initially made `first_name` return `Option<&str>` but wasn't sure... would you prefer that instead?

jfecher: I'd prefer defaulting to "" or the Option version over panicking internally if there are cases where we don't actually have any path segments

asterite: Done!

Issue #6444, opened 2024-11-04
Test `slice_dynamic_index` fails with inliner = -Inf and brillig


aakoshh:
### Aim

Trying to run `test_programs` with different `--inliner-aggressiveness` in https://github.com/noir-lang/noir/pull/6429

### Expected Behavior

Tests under `execution_success` should work.

### Bug

```console
❯ nargo execute --force --inliner-aggressiveness -9223372036854775808 --force-brillig
error: Failed to solve program: 'Failed to solve brillig function'
   ┌─ /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/slice_dynamic_index/src/main.nr:47:12
   │
47 │     assert(slice[y] == 1);
   │            -------------
   │
   = Call stack:
     1. /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/slice_dynamic_index/src/main.nr:3:5
     2. /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/slice_dynamic_index/src/main.nr:14:5
     3. /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/slice_dynamic_index/src/main.nr:47:12

Failed to solve program: 'Failed to solve brillig function'
```

### To Reproduce

1. `cd test_programs/execution_success/slice_dynamic_index`
3. `nargo execute --force --force-brillig --inliner-aggressiveness -9223372036854775808`

### Workaround

None

### Project Impact

None

### Nargo Version

nargo version = 0.36.0 noirc version = 0.36.0+2f0cb3e80f3d93a1dee77fffacc397811e300257 (git version hash: 2f0cb3e80f3d93a1dee77fffacc397811e300257, is dirty: false)

### Would you like to submit a PR for this Issue?

None

aakoshh:
The SSA for the failing case with inliner = `-Inf` looks as follows

<details>
<summary>After Array Set Optimizations:</summary>

```
brillig(inline) fn main f0 {
  b0(v0: Field):
    v2 = sub v0, Field 1
    v4 = sub v0, Field 4
    call f1(v2, v4)
    return 
}
brillig(inline) fn regression_dynamic_slice_index f1 {
  b0(v0: Field, v1: Field):
    v3 = allocate
    store u32 0 at v3
    inc_rc []
    v6 = allocate
    store [] at v6
    jmp b1(u32 0)
  b1(v2: u32):
    v8 = lt v2, u32 5
    jmpif v8 then: b3, else: b2
[...]

PR #6429, opened 2024-11-01
chore(test): Run test matrix on test_programs

aakoshh:
# Description

## Problem\*

Resolves #6390 

## Summary\*

Runs the `test_programs` with -Inf, 0 and +Inf `--inliner-aggressiveness` before we can change the default from `+Inf` (inline everything) to 0. 

There are some test failures triggered by this, which need to be fixed before merging this:
- #6439
- #6440
- #6441 
- #6442 
- #6443
- #6444

## Additional Context

A specific test can be executed like so:
```shell
cargo test -q -p nargo_cli --test execute <test-name>
```

Consistent test failures are (created tickets for them, see above):
```
    tests::execution_success::test_brillig_cow::false_i64_min_expects
    tests::execution_success::test_brillig_cow::true_i64_min_expects
    tests::execution_success::test_closures_mut_ref::true_0_expects
    tests::execution_success::test_closures_mut_ref::true_i64_min_expects
    tests::execution_success::test_eddsa::true_0_expects
    tests::execution_success::test_eddsa::true_i64_min_expects
    tests::execution_success::test_missing_closure_env::true_0_expects
    tests::execution_success::test_missing_closure_env::true_i64_min_expects
    tests::execution_success::test_references::true_i64_min_expects
    tests::execution_success::test_slice_dynamic_index::true_i64_min_expects
```

Note that instead of the `_brillig` suffix we have `::true_` indicate which tests are run with `--force-brillig`.

All failures are under `execution_success`, so I switched off the inlining matrix for the other test types; doing so impoves the runtime by 15-30%.

Another bug independent from inliner aggressiveness was #6445 This existed in the codebase before and I have seen it locally, but it never seemed to have affected CI; it probably uses less threads. I fixed it in this PR; since it affected test code generation I didn't think a separate PR would be helpful.

## Documentation\*

Check one:
- [x] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [x] I have tested the changes locally.
- [ ] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #6463, opened 2024-11-06
fix(ssa): Change array_set to not mutate slices coming from function inputs

aakoshh:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6444

## Summary\*

https://github.com/noir-lang/noir/pull/6355 changed the logic in the `array_set` SSA pass so that arrays which were loaded from references that are inputs to the block are not set with mutation, as the reference might be shared with something outside the block. 

In this instance what we had was a slice passed to a function like `v1: [Field]`, as opposed to the above which passed a reference `v1: &mut [Field; 3]`, but the effect was similar in that the code ended up mutating the slice backing `v1`, over which a refcount was even increased, instead of making a copy. 

The PR changes the decision so that if we are setting an item on an array that is part of the _function_ inputs (not the block inputs), then it will not use mutation. 

## Additional Context

See https://github.com/noir-lang/noir/issues/6444#issuecomment-2459624919 for the SSA samples.
I have tested this with https://github.com/noir-lang/noir/pull/6429 but opened the PR to `master`.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/array_set.rs
```diff
@@ -109,21 +118,30 @@ impl<'f> Context<'f> {
...
+
+                    let can_mutate = if let Some(is_from_param) = self.arrays_from_load.get(&array)
+                    {
                         // If the array was loaded from a reference parameter, we cannot
                         // safely mark that array mutable as it may be shared by another value.
-                        if !is_from_param && is_return_block {
-                            self.instructions_that_can_be_made_mutable.insert(*instruction_id);
-                        }
-                    } else if !array_in_terminator {
+                        !is_from_param && is_return_block
```
aakoshh, line 139: I wonder if this `is_from_param` also should be looking at function parameters, rather than block parameters?

jfecher:
This only affects brillig code, right? If so, we should probably ignore the mutability on array sets completely there - that was only ever intended for Acir. Brillig should use the ref counts on each array to determine mutability.

compiler/noirc_evaluator/src/ssa/opt/array_set.rs
aakoshh: I started here by checking the block parameters, but then it failed on this SSA:

```
brillig(inline) fn dynamic_slice_index_if f6 {
  b0(v0: u32, v1: [Field], v2: Field):
    v3 = allocate
    store v1 at v3
    inc_rc v1
    v4 = truncate v2 to 32 bits, max_bit_size: 254
    v5 = cast v4 as u32
    v7 = lt v5, u32 10
    jmpif v7 then: b2, else: b1
  b2():
    v13 = cast v2 as u32
    v14 = lt v13, v0
    constrain v14 == u1 1 '"Index out of bounds"'
    v15 = array_get v1, index v13
    constrain v15 == Field 4
    v18 = array_set mut v1, index v13, value Field 2
    store v18 at v3
    jmp b3()
  b3():
    v19 = load v3
    v21 = lt u32 4, v0
    constrain v21 == u1 1 '"Index out of bounds"'
    v22 = array_get v19, index u32 4
    constrain v22 == Field 2
    dec_rc v1
    return 
  b1():
    v8 = cast v2 as u32
    v9 = lt v8, v0
    constrain v9 == u1 1 '"Index out of bounds"'
    v11 = array_get v1, index v8
    constrain v11 == Field 0
    jmp b3()
}
```

The `array_set` on `v18` is in a block without parameters.

aakoshh:
> This only affects brillig code, right? 

Only brillig tests failed with this. It is also my understanding that ACIR doesn't care about this mutability, and that it's a Brillig-only feature. In that light I don't understand why the ACIR code sizes increased in the report 🤔

aakoshh:
So you're saying that if Brillig ignored the `mutable` field on `Instruction::ArraySet` then it wouldn't make a mistake, but ACIR, which ignores the refcounts, should still be given the correct `mutable` field, right? In this instance, for example, adding `mut` is not correct, is it? If that's true, why doesn't the ACIR version of the test fail?

jfecher:
> So you're saying that if Brillig ignored the mutable field on Instruction::ArraySet then it wouldn't make a mistake, but ACIR, which ignores the refcounts, should still be given the correct mutable field, right? In this instance, for example, adding mut is not correct, is it? If that's true, why doesn't the ACIR version of the test fail?

The original intention was for Acir to only use the mutable field (and for that pass to only be for acir code), and for brillig to only use ref counts for mutation.
I'll look into the test more in-depth now to see why the acir version doesn't fail but generally speaking all acir code is inlined into one function and mutable references can't be passed between acir & brillig boundaries, so acir generally doesn't need to worry about function parameters. All loads and stores are also expected to be removed by the end of SSA in acir code.

TomAFrench (approved):
This PR was discussed in scrum and it was agreed to move forwards with this and then follow up with a removal of the concept of array mutability in brillig functions in a followup PR.

jfecher:
@TomAFrench this should be reverted then once the next PR is made. Since it prevents us from mutating arrays passed by parameters to main in acir and is extra logic that is unneeded once brillig is updated to not check the mutable flag

PR #6466, opened 2024-11-06
feat: try to inline brillig calls with all constant arguments

### LINKED ISSUE #6453: Inline results from brillig function calls for which all arguments are known during SSA
We currently execution any brillig function calls for which all arguments are known during ACIR gen at compile-time in order to optimize the generated circuit by replacing the call with its intended results.

Performing this optimization during ACIR-gen however means that this occurs after most of our optimization passes have already run (and so they cannot benefit from the removal of these brillig calls). In SSA we retain the brillig calls which completely block any constant propagation from occurring.

An example of this is shown in the SSA below where the program should be able to be optimized down to an empty program in SSA but because we have some brillig calls we're unable to propagate these results forwards.

```
acir(inline) fn main f0 {
  b0():
    v8, v9 = call f4([Field 1225638183666705271503180972154335964, Field 1149176344733214219505367673588801074, Field 28650], [Field 12, Field 0, Field 0])
    v33, v34, v35 = call f6(u1 0, [Field 0, Field 0, Field 2¹⁶], [u64 0, u64 0, u64 0, u64 0, u64 2¹⁶, u64 0], [u64 0, u64 0, u64 0, u64 0, u64 2¹⁶, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0], [Field 2¹²⁰, Field 1329227995784915872903807060280344575, Field 131071], [Field 0, Field 0, Field 2²²], [[v8]], [[u1 0]], [[[Field 12, Field 0, Field 0]]], [[u1 0]], [[Field 1225638183666705271503180972154335964, Field 1149176344733214219505367673588801074, Field 28650], v9], [u1 1, u1 0])
    v37 = array_get v33, index u32 0
    range_check v37 to 120 bits
    v39 = array_get v33, index u32 1
    range_check v39 to 120 bits
    v41 = array_get v33, index u32 2
    range_check v41 to 120 bits
    range_check v41 to 23 bits
    v42 = array_get v8, index u32 0
    v43 = array_get v8, index u32 1
    v44 = array_get v8, index u32 2
    v45 = array_get v9, index u32 0
    v47 = add Field 103589812118210601400626088126008612, v45
    v48 = array_get v9, index u32 1
    v50 = add Field 180051651051701653398439386691543501, v48
    v51 = array_get v9, index u32 2
 // snip
```

We can address this by 
- Create a copy of the brillig function being called
- Replace all function arguments with the constants the function is being called with.
- Run the optimization passes until the function is reduced to its terminator instruction
- Inline these results in place of the brillig call instruction.

There will be a couple of gotchas related to failing assertions or oracle calls which will need to be handled with some care. As a first cut we can just abort this process if we hit these cases for now.

We started touching on this work with https://github.com/noir-lang/noir/issues/5839

asterite:
# Description

## Problem

Resolves #6453

## Summary

Adds a pass that tries to inline brillig calls where all arguments are constant.

This is done by first creating a copy of the function, setting its runtime to ACIR, replacing its parameters by the call arguments, optimizing the function like we optimize everything else (except inlining no-predicates, because that requires all the functions to be in context: maybe this can be done as a follow-up PR but I'm not sure if it's worth it), then checking if the function was reduced to a single return terminator instruction.

## Additional Context

The changes to the loop unrolling code are mainly there because now loop unrolling can happen inside brillig calls that we are trying to optimize out, so they can contain `break` and generally have a struct that can't happen in ACIR functions. I'm happy to try to find another way to do this, let me know!

## Documentation

Check one:

# PR Checklist


TomAFrench:
We are often going to eliminate all usage of certain brillig functions at this step but they end up being retained (and show up in `show-ssa`) currently. Can you add some tracking to drop these unused functions after this pass?

asterite:
Sure!

Though... I just checked with this program:

```noir
    let (a, _) = unsafe { identity(5) };
    assert_eq(a, 5);
    x + a
}

unconstrained fn identity(x: Field) -> (Field, Field) {
    let y = x;
    let z = y - 1;
    let w = z + 1;
    identity2(w)
}

unconstrained fn identity2(x: Field) -> (Field, Field) {
    (x, x)
}
```

and `--show-ssa` shows me this:

```
acir(inline) fn main f0 {
  b0(v0: Field):
    v2 = add v0, Field 5
    return v2
}
```

It seems after the `inline` pass the unused brillig functions are already being removed. Or... do you have some code where this doesn't happen?

TomAFrench:
I get the below SSA

```
After Inlining Const Brillig Calls:
acir(inline) fn main f0 {
  b0(v0: Field):
    v3 = add v0, Field 5
    return v3
}
brillig(inline) fn identity f1 {
  b0(v0: Field):
    v2 = sub v0, Field 1
    v3 = add v2, Field 1
    return v3, v3
}
```

We do remove this after the final inlining step but there's 9 passes in between these two steps so we get a good amount of unnecessary output in the SSA display

asterite:
Oooh... you mean removing them in this same pass, right? Sure, I'll do it :-)

TomAFrench:
Yep, that's it

TomAFrench:
One extra thing to note is that this branch panics on the testcase added in https://github.com/noir-lang/noir/pull/6464 (as opposed to just compiling to incorrect code).

Obviously there's a different issue in the compiler affecting that code but fyi.

asterite:
Ah, it's also with loop unrolling. I had to change something in this PR to workaround an issue... I guess I'll have to learn how loop unrolling works, and why it doesn't work in this brillig functions turned into ACIR.

asterite:
I managed to reduce the code that triggers the panic in bignum_regression:

```noir
fn main() {
    unsafe { foo() };
}

unconstrained fn foo() {
    for i in 0..1 {
        if i == 0 {
            break;
        }
    }
}
```

It seems it's failing because of that `break`. It's likely that the optimization was made without `break` into account because loop unrolling only happens in ACIR code, and ACIR code doesn't allow breaks.

I can try to fix loop unrolling to account for this case, though it will likely take me a lot of time. Alternatively, we can avoid trying to inline these functions if we know they have `break` (for now).

Which one should I do?

asterite:
For now I pushed a commit that makes it not panic.

I wonder if loop unrolling should be able to work in two modes, ACIR or "ACIR from Brillig". That way we can still panic and use `expect` for the normal case, and just ignore those errors for the "ACIR from Brillig" case. What do you think?

TomAFrench:
> I can try to fix loop unrolling to account for this case, though it will likely take me a lot of time. Alternatively, we can avoid trying to inline these functions if we know they have break (for now).
>
>Which one should I do?

For now let's just skip inlining that function and we can make improvements later.

asterite:
Sounds good, done!

asterite:
(just asked a review to signal that this is ready for review)

compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -149,6 +115,69 @@ pub(crate) fn optimize_into_acir(
...
+        .run_pass(
+            |ssa| ssa.inline_const_brillig_calls(inliner_aggressiveness),
+            "After Inlining Const Brillig Calls:",
+        );
+    let ssa = optimize_ssa_after_inline_const_brillig_calls(
+        builder,
+        inliner_aggressiveness,
+        true, // inline functions with no predicates
+    )?;
+    Ok(ssa)
```
aakoshh, line 134: Is there a reason to break up `optimize_ssa` this way? Why is "after inline cost brillig calls" a significant point?

aakoshh: Ah found it. Maybe some docstrings could explain, like help decide if a new pass should be before or after this?

asterite: Oh, it's mainly because at this point the other optimizations already happened, so to avoid redoing them again (and I think the inlining pass might break things too).

Should I add a comment for this somewhere?

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
```diff
@@ -0,0 +1,220 @@
...
+            for instruction_id in self.dfg[block_id].take_instructions() {
+                if !self.optimize_const_brillig_call(
+                    instruction_id,
+                    brillig_functions,
+                    brillig_functions_we_could_not_inline,
+                    inliner_aggressiveness,
+                    error_selector_to_type,
+                ) {
+                    self.dfg[block_id].instructions_mut().push(instruction_id);
+                }
```
aakoshh, line 86: FWIW I find this code a bit unintuitive at first glance. The docs of `optimize_const_brilling_call` say that:
> If the optimization is successful, returns true. Otherwise returns false. The given instruction is not removed from the function.

The instruction is "not removed from the function", but at the same time `.take_instructions()` removed all instructions from the block, and looking at the `if` it shows that if it optimised it then we are not supposed to add it back... is it _is_ removed, but not _by_ the optimise call?

Then I wonder what it does, and I suppose to optimise it needs to replace the instruction with something, so I wonder if it will push to `self.dfg` itself. And if it does, then why can't it decide whether to push the instruction back as-is, or not?

I'm still looking at the implementation below to understand what's going on, so it might get clearer in a minute, just wanted to mention it.

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        // The function we have is already a copy of the original function, but we need to clone
+        // it again because there might be multiple calls to the same brillig function.
+        let mut function = Function::clone_with_id(*func_id, function);
+
+        // Find the entry block and remove its parameters
+        let entry_block_id = function.entry_block();
+        let entry_block = &mut function.dfg[entry_block_id];
+        let entry_block_parameters = entry_block.take_parameters();
```
aakoshh, line 128: I suppose this is another reason it must be cloned, otherwise we might be mutating the original block. 

I actually don't understand why this works: in `Function::inline_const_brillig_calls` above we started with `self.dfg[block_id].take_instructions()` of all `self.reachable_blocks()`; can't the `entry_block` already have its parameters taken?

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
```diff
@@ -0,0 +1,220 @@
...
+    error_selector_to_type: &BTreeMap<ErrorSelector, Type>,
+) -> Result<Function, RuntimeError> {
+    function.set_runtime(RuntimeType::Acir(InlineType::InlineAlways));
+
+    let ssa = Ssa::new(vec![function], error_selector_to_type.clone());
+    let builder = SsaBuilder { ssa, print_ssa_passes: false, print_codegen_timings: false };
+    let mut ssa = optimize_ssa_after_inline_const_brillig_calls(
+        builder,
+        inliner_aggressiveness,
+        false, // don't inline functions with no predicates
```
aakoshh, line 217: It would be nice to have a comment on why not? Asking for a friend ;)

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
aakoshh: Ah it's `take_parameters` not `take_instructions` 🥹

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
```diff
@@ -0,0 +1,220 @@
...
+        if !entry_block.instructions().is_empty() {
+            brillig_functions_we_could_not_inline.insert(*func_id);
+            return false;
+        }
+
+        let terminator = entry_block.take_terminator();
+        let TerminatorInstruction::Return { return_values, call_stack: _ } = terminator else {
+            brillig_functions_we_could_not_inline.insert(*func_id);
+            return false;
+        };
```
aakoshh, line 158: Just to confirm: the presence of these could not be verified before trying to inline everything, because the act of inlining could have changed them?

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
```diff
@@ -0,0 +1,220 @@
...
+
+        true
+    }
+}
+
+/// Copies a constant from one function to another.
+/// Though it might seem we can just take a value out of `from_function` and call `make_value` on `to_function`,
+/// if the constant is an array the values will still keep pointing to `from_function`. So, this function
+/// recursively copies the array values too.
+fn copy_constant_to_function(
```
aakoshh, line 183: Just curious: why is `optimize_const_brillig_call` defined on `Function`, but this one, which could also be `self.copy_constant_from_function(&mut self, from_function: &Function, constant_id: ValueId)`, defined as a separate function?  I thought maybe you don't want it to pollute the `Function` namespace in other places, but if the other method is there, rather than on some kind of `Context` like in some of the other packages, why not this? It also looks like a method that could find its uses elsewhere.

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
asterite: 😆 

The main reason is that running that pass, it crashed. But then I found that it crashed because to inline things we need to have all other functions in context, and here we just have this single function in context. I'll add a comment for this.

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
asterite: Correct. Or, well, in addition to inlining, jumps and stuff like that is all inlined into a single block.

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
aakoshh: Yeah, I see there is a lot of bookkeeping to be done on `self.dfg` in the successful case, and it's an elegant way to put the instruction back if it didn't happen, and at least the "putting back" is visually close to the taking. It's a shame there is no easy way to return the successful case to say "do this instead to the DFG".

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
asterite: Oh, good point! Actually, I had this defined in `Function` originally, but then I needed to copy constants from the second function into the first one so I thought "I'll need to make this a free function" but I forgot I could call it on the second function 😓 

I'll change it, I think it's clearer as a method.

asterite: Done! I left it in this same file. If later this functionality is generally needed we could move it to the main `Function` impl.

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
```diff
@@ -0,0 +1,221 @@
...
+/// Optimizes a function by running the same passes as `optimize_into_acir`
+/// after the `inline_const_brillig_calls` pass.
+/// The function is changed to be an ACIR function so the function can potentially
+/// be optimized into a single return terminator.
+fn optimize(
+    mut function: Function,
+    inliner_aggressiveness: i64,
+    error_selector_to_type: &BTreeMap<ErrorSelector, Type>,
+) -> Result<Function, RuntimeError> {
+    function.set_runtime(RuntimeType::Acir(InlineType::InlineAlways));
```
aakoshh, line 207: Why do we need to set this particular runtime for this function? What would `Acir(Inline)` or `Brillig(InlineAlways)` do? And doesn't any function called by this one need to have the same thing done to it?

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
```diff
@@ -0,0 +1,221 @@
...
+                brillig_functions.insert(*func_id, cloned_function);
+            };
+        }
+
+        // Keep track of which brillig functions we couldn't completely inline: we'll remove the ones we could.
+        let mut brillig_functions_we_could_not_inline = HashSet::new();
+
+        for func in self.functions.values_mut() {
+            func.inline_const_brillig_calls(
+                &brillig_functions,
```
aakoshh, line 38: Why not loop through `brillig_functions` to get the IDs of the ones we already shortlisted?

aakoshh: Ah because you are trying to inline the brillig function which are called from any function at all.

If brillig function `foo` is called from non-brillig functions `bar` and `baz`, then it is going to be attempted to be optimised twice, with respective constant arguments from both call sites.

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
asterite: I don't know much about this, but I saw in the code that some optimizations don't run for brillig code (for example loop unrolling). I also asked Tom in which mode we should optimize and he said Acir. Maybe @TomAFrench or someone else knows what can or can't be optimized in acir/brillig?

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
```diff
@@ -0,0 +1,221 @@
...
+            // Lookup the argument in the current function and insert it in the function copy
+            let new_argument_id = self.copy_constant_to_function(*argument_id, &mut function);
+            function.dfg.set_value_from_id(*parameter_id, new_argument_id);
+        }
+
+        // Try to fully optimize the function. If we can't, we can't inline it's constant value.
+        let Ok(mut function) = optimize(function, inliner_aggressiveness, error_selector_to_type)
+        else {
+            brillig_functions_we_could_not_inline.insert(*func_id);
+            return false;
```
aakoshh, line 144: Just because you said you didn't like similar repetition, it could possibly be wrapped up into an enum with variants like `NotBrillig`, `CannotInline(FunctionId)`, `Optimized` and then the caller could make sure to insert into the `brillig_functions_we_could_not_inline` instead of having to pass it in, and also to return the instruction to the block unless it's `Optimized`.

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
```diff
@@ -0,0 +1,221 @@
...
+                &brillig_functions,
+                &mut brillig_functions_we_could_not_inline,
+                inliner_aggressiveness,
+                error_selector_to_type,
+            );
+        }
+
+        // Remove the brillig functions that are no longer called
+        for func_id in brillig_functions.keys() {
+            // We never want to remove the main function (it could be brillig if `--force-brillig` was given)
```
aakoshh, line 47: It can even be Brillig on its own like [here](https://github.com/noir-lang/noir/blob/b74b4eaf7acf6becd0f98ab25d0d5ccde3affebe/test_programs/execution_success/brillig_cow/src/main.nr#L45)

compiler/noirc_evaluator/src/ssa.rs
aakoshh: I would mention it on `optimize_ssa_after_inline_const_brillig_calls` that it's used to tentatively try to optimise away functions by the `.inline_const_brillig_calls` itself. 

I keep thinking that it would be nice if the `inline_const_brillig_calls` didn't have to know exactly which steps are following it. The `SsaBuilder` isn't an abstract one that one can pass around and customise, it's tied to an `Ssa` instance and the caller has to drive the passes, but ostensibly `inline_cost_brillig_calls` could get a closure that takes an `Ssa` and runs what we want on it, and then this flow would be visible here.

aakoshh: Something like:
```rust
fn optimize_ssa(builder: SsaBuilder, inliner_aggressiveness: i64) -> Result<Ssa, RuntimeError> {
    fn stage1(builder: SsaBuilder, inliner_aggressiveness: i64) -> Result<SsaBuilder, RuntimeError> {
        Ok(builder
        .run_pass(Ssa::defunctionalize, "After Defunctionalization:")
        .run_pass(Ssa::remove_paired_rc, "After Removing Paired rc_inc & rc_decs:")
        .run_pass(Ssa::separate_runtime, "After Runtime Separation:")
        .run_pass(Ssa::resolve_is_unconstrained, "After Resolving IsUnconstrained:")
        .run_pass(|ssa| ssa.inline_functions(inliner_aggressiveness), "After Inlining:")
        .run_pass(
            |ssa| ssa.inline_const_brillig_calls(|ssa| {
                let builder = SsaBuilder { ssa, print_stuff: false };
                stage2(builder, false).map(|b| b.finish())
            }),
            "After Inlining Const Brillig Calls:",
        ))
    }

    fn stage2(builder: SsaBuilder, inliner_aggressiveness: i64, inline_functions_with_no_predicates) -> Result<SsaBuilder, RuntimeError> {
        Ok(builder
        // Run mem2reg with the CFG separated into blocks
        .run_pass(Ssa::mem2reg, "After Mem2Reg (1st):")
        .run_pass(Ssa::simplify_cfg, "After Simplifying (1st):")
        .run_pass(Ssa::as_slice_optimization, "After `as_slice` optimization")
        .try_run_pass(
            Ssa::evaluate_static_assert_and_assert_constant,
            "After `static_assert` and `assert_constant`:",
        )?
        .try_run_pass(Ssa::unroll_loops_iteratively, "After Unrolling:")?
        .run_pass(Ssa::simplify_cfg, "After Simplifying (2nd):")
        .run_pass(Ssa::flatten_cfg, "After Flattening:")
        .run_pass(Ssa::remove_bit_shifts, "After Removing Bit Shifts:")
        // Run mem2reg once more with the flattened CFG to catch any remaining loads/stores
        .run_pass(Ssa::mem2reg, "After Mem2Reg (2nd):")
        // Run the inlining pass again to handle functions with `InlineType::NoPredicates`.
        // Before flattening is run, we treat functions marked with the `InlineType::NoPredicates` as an entry point.
        // This pass must come immediately following `mem2reg` as the succeeding passes
        // may create an SSA which inlining fails to handle.
        .run_pass(
            |ssa| if inline_functions_with_no_predicates {
                ssa.inline_functions_with_no_predicates(inliner_aggressiveness)
            } else {
                ssa
            },
            "After Inlining:",
        )
        .run_pass(Ssa::remove_if_else, "After Remove IfElse:")
        .run_pass(Ssa::fold_constants, "After Constant Folding:")
        .run_pass(Ssa::remove_enable_side_effects, "After EnableSideEffectsIf removal:")
        .run_pass(Ssa::fold_constants_using_constraints, "After Constraint Folding:")
        .run_pass(Ssa::dead_instruction_elimination, "After Dead Instruction Elimination:")
        .run_pass(Ssa::simplify_cfg, "After Simplifying:")
        .run_pass(Ssa::array_set_optimization, "After Array Set Optimizations:"))
    }

    Ok(stage1(builder, true)?.finish())
}
```

aakoshh: Maybe add a `run_pass_if` variant to avoid printing if a step is turned off (not that it would print in this particular case, but just to not have to break up the chain).

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
TomAFrench: I suggested treating the code as if it were to be ACIR as we don't unroll loops in brillig code and without loop unrolling we're not going to be able to simplify a function down to nothing generally.

This does result in some issues where we can't deal with brillig-only language features such as break however. Perhaps a better solution to this problem is to defer the handling of constant-input brillig functions until after brillig gen. We can then execute the brillig functions directly (so we get the results without needing to be able to fully optimize them in SSA) and then feed those back into the SSA form of the program to do further optimizations.

The alternative to this is to execute the SSA directly but that adds yet another execution environment so I'm reluctant to do so.

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -233,14 +233,17 @@ fn unroll_loop(
...
         .predecessors(loop_.header)
         .filter(|predecessor| *predecessor != loop_.back_edge_start)
         .collect::<Vec<_>>();
 
-    assert_eq!(pre_header.len(), 1);
-    pre_header.remove(0)
+    if pre_header.len() == 1 {
+        Ok(pre_header.remove(0))
+    } else {
+        Err(CallStack::new())
```
aakoshh, line 245: It would be helpful to add some comments to this function. For example it's not obvious to me what the error represents.

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -377,8 +380,7 @@ impl<'f> LoopIteration<'f> {
             }
         }
 
-        self.induction_value
-            .expect("Expected to find the induction variable by end of loop iteration")
+        self.induction_value.ok_or_else(CallStack::new)
```
aakoshh, line 383: Should there be some custom error type to preserve this message?

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
aakoshh: Thanks, it would be great to preserve this explanation as a comment. 

Some followup questions:
* Is this setting transitive, ie. what if this function calls another function that is a brillig function with a loop?
* What happens if we mark a brillig function with `break` as `Acir(InlineAlways)`? I suspect that's why something further down was turned into `Result`, but I don't see it in the `run_passes` returning an error, and I remember seeing errors where the compiler couldn't figure out the loop bounds at compile time in constrained code.

aakoshh: Oh I see that we only pass this function to the SSA in `let ssa = Ssa::new(vec![function],`. 

What I don't fully understand is when [should_inline_call](https://github.com/noir-lang/noir/blob/b74b4eaf7acf6becd0f98ab25d0d5ccde3affebe/compiler/noirc_evaluator/src/ssa/opt/inlining.rs#L638-L639) makes a decision about [calls made by this function](https://github.com/noir-lang/noir/blob/b74b4eaf7acf6becd0f98ab25d0d5ccde3affebe/compiler/noirc_evaluator/src/ssa/opt/inlining.rs#L607-L610) it looks for the definition of the callee in the SSA, and it won't find it. It looks like it would panic there on `let function = &ssa.functions[&called_func_id];` 🤔 

And that function also answers the question of what would happen if the called function is brillig: it's not inlined because the caller is marked as `Acir`.

compiler/noirc_evaluator/src/ssa.rs
aakoshh: > didn't have to know exactly which steps are following it

To illustrate this point, currently it looks something like this:

```rust
fn optimise(builder: SsaBuilder) -> Ssa {
   foo(builder)
}
fn foo(builder: SsaBuiler) -> Ssa {
    builder.run(Ssa::a);
    builder.run(Ssa::b);
    builder.run(Ssa::c);
    bar(builder)
}
fn bar(builder: SsaBuiler) -> Ssa {
    builder.run(Ssa::d);
    builder.run(Ssa::e);
    baz(builder)
}
fn baz(builder: SsaBuiler) -> Ssa {
    builder.run(Ssa::f);
    builder.finish()
}

// Individual modules weave together all methods into Ssa:
impl Ssa {
  fn a(self) -> Self { todo!(); }
  fn b(self) -> Self { todo!(); }
  fn c(self) -> Self { todo!(); let ssa = bar(self); todo!() }
  fn d(self) -> Self { todo!(); let ssa = baz(self); todo!() }
  fn e(self) -> Self { todo!(); }
  fn f(self) -> Self { todo!(); }
}
```

That is, `c` and `d` know to call `bar` and `baz` to do "the rest" of the optimisations, but  if someone was to refactor the top level functions they would have to go an update `c` and `d` to call the correct entry points, make sure there is one, and they would be going up and down in the code to follow the chain of calls. For example consider the following refactorings: 
* moving `c` into `bar` would cause an infinite loop
* moving `d` into `foo` would change what comes after `c` and might invalidate its assumptions about what happens to the Ssa

By passing in a closure to `c` and `d`, the sequencing of the passes would all be visible in the top level builder. The composer would need to understand what `c` needs, but it could be documented and it would be obviously there, rather than a method named `passes_after_c` and e.g. `passes_after_d_but_before_f` which sound a bit more accidental.

aakoshh: I guess another option would be to put all passes into an ordered map and make it possible to resume from anywhere by ID.

asterite: I'll try to refactor this to one of the options you suggested.

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
asterite: > What I don't fully understand is when [should_inline_call](https://github.com/noir-lang/noir/blob/b74b4eaf7acf6becd0f98ab25d0d5ccde3affebe/compiler/noirc_evaluator/src/ssa/opt/inlining.rs#L638-L639) makes a decision about [calls made by this function](https://github.com/noir-lang/noir/blob/b74b4eaf7acf6becd0f98ab25d0d5ccde3affebe/compiler/noirc_evaluator/src/ssa/opt/inlining.rs#L607-L610) it looks for the definition of the callee in the SSA, and it won't find it

That's why inlining is not performed when optimizing this function. Or, put another way, inlining was already done before we start optimizing this function. And the other inline pass (for no predicates) is skipped for this reason too.

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
asterite: Ah, good point. But the code was already returning `Callstack::new` in some other places so I just followed what was done here without really understanding why that is returned. I'll briefly take a look at this to see why a Callstack is returned and try to change it or add a comment.

asterite: It seems all RuntimeErrors have must have a call_stack, and here a dummy one is used, just to comply to that requirement (it's a `call_stack` method on `RuntimeError`). The call stack is not used when turning that error into a message... for now I'll leave it like this because it's unrelated to this PR changes, if that's okay with you.

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
asterite: I think I'll make this pass have two modes (acir and brillig) so that it still panics on the acir case, but not panics on the brillig one.

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
asterite: This is a great idea! I pushed a commit for this. I think it simplified some things, and it's actually the way I wanted the code to look like, but wasn't sure how to do it.

compiler/noirc_evaluator/src/ssa.rs
asterite: I just tried it but it's complex because of the `#[tracing::instrument(level = "trace", skip(self))]` annotation on top of asses: we can't pass closures or `fn` pointers there as they are not `Debug`.

I'm also thinking that the current way is okay: we are running this `inline_const_brillig_calls` optimization so it makes sense to only run passes that comes after it, hence the name I chose.

> I keep thinking that it would be nice if the inline_const_brillig_calls didn't have to know exactly which steps are following it

It doesn't: it just runs the passes that comes after it, but it still doesn't know which are those.

Alternatively... I could just call individual passes when optimizing this function. It's actually what I originally had, but I didn't want us to have to remember to add new optimizations in two places. I guess that's the tradeoff.

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
asterite: Done!

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
aakoshh: Yes, I also had a look and saw that it can only be produced when an unexpected `TerminatorInstruction` is found, which carries a location. Based on the docs of `unroll_loop` this wasn't always returned. The only comment I would add here is that when `pre_header.len() > 1` is I believe when `break` or `continue` is used, in which case we cannot inline and we will discard this entire attempt as an error.

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
aakoshh: Ah, okay, I thought that `false` would have something to do with it. Well I guess if it wasn't correct it would have panicked already 👍

compiler/noirc_evaluator/src/ssa.rs
aakoshh: Thanks for trying. Yeah, I admit it's rather philosophical. I mean now that there is a single deviated "after_inlining_const_brillig" it's easy to say that it doesn't actually know, but in another way it knows that this function is there for it to call; the only way I can say is that if there were more of them, and they weren't so deftly named after the caller itself, it would be more obvious that it has to know which one to pick, rather than being given by the caller. Something like Dependency Injection. Never mind, there is only one :)

aakoshh (approved):
LGTM!

TomAFrench:
@asterite I think we can close this?

Issue #6238, opened 2024-10-08
Fix `eval_global_as_array_length` to support non-array-length types


michaeljklein:
### Aim

https://github.com/noir-lang/noir/pull/6137 adds the warning `EvaluatedGlobalIsntU32` when `eval_global_as_array_length` is used on non-`u32` types. 

This is connected to the following (`test`: issue) pairs:
- `numeric_generic_field_larger_than_u32`: https://github.com/noir-lang/noir/issues/6125
- `numeric_generic_field_arithmetic_larger_than_u32`: https://github.com/noir-lang/noir/issues/6126

### Expected Behavior

`eval_global_as_array_length` to either handle larger types, e.g. `Field` properly, or to only be used when the globals are within the expected range

TODO:
- [ ] See how much overlap there between the evaluate global as array length function and the comptime interpreter code: (some of) the comptime interpreter method(s) may be usable here
- [ ] Test that we error for overflowing in intermediate `u32` kind computations, e.g. the equivalent of `U32_MAX + 1 - X` where `X > 0`

### Bug

- `numeric_generic_field_larger_than_u32`: https://github.com/noir-lang/noir/issues/6125
- `numeric_generic_field_arithmetic_larger_than_u32`: https://github.com/noir-lang/noir/issues/6126

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

michaeljklein:
Current plan:
- Resolve globals to:
  1. `Type::InfixExpr` / `Type::Constant` if possible
    + Where nested globals can be appx. `Type::Global(GlobalId)`
  3. `comptime` globals otherwise (to allow using the existing interpreter)
    + Require such globals to be explicitly `comptime`
- Use existing type-checking and interpreting code paths instead of `eval_global_as_array_length`

Issue #6126, opened 2024-09-23
`Field` computed to be larger than `u32::MAX` with arithmetic generics panics


michaeljklein:
### Aim

Attempted to compile a program that adds two `Field`'s using arithmetic generics, where the result of the computation is larger than `u32::MAX`:

```noir
struct Foo<let F: Field> {}

impl<let F: Field> Foo<F> {
    fn size(self) -> Field {
        F
    }
}

// 2^32 - 1
global A: Field = 4294967295;

// Avoiding overflow succeeds:
// fn foo<let A: Field>() -> Foo<A> {
fn foo<let A: Field>() -> Foo<A + A> {
    Foo {}
}

fn main() {
    // skipping 'size()' succeeds
    let C = foo::<A>();
    let C = foo::<A>().size();
}
```

Note that using a `u64` instead of a `Field` fails with:
```bash
error: The only supported numeric generic types are `u1`, `u8`, `u16`, and `u32`
  ┌─ src/main.nr:5:12
  │
5 │ struct Foo<let F: u64> {}
  │            ----------
  │
```

### Expected Behavior

Expected compilation to succeed or else fail with a user-error

### Bug

```bash
❯ ~/.nargo/bin/nargo execute
The application panicked (crashed).
Message:  Non-numeric type variable used in expression expecting a value
Location: compiler/noirc_frontend/src/monomorphization/mod.rs:925

This is a bug. We may have already fixed this in newer versions of Nargo so try searching for similar issues at https://github.com/noir-lang/noir/issues/.
If there isn't an open issue for this bug, consider opening one at https://github.com/noir-lang/noir/issues/new?labels=bug&template=bug_report.yml
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Nargo Version

nargo version = 0.34.0 noirc version = 0.34.0+5598059576c6cbc72474aff4b18bc5e4bb9f08e1 (git version hash: 5598059576c6cbc72474aff4b18bc5e4bb9f08e1, is dirty: false)

### Would you like to submit a PR for this Issue?

None

Issue #6125, opened 2024-09-23
Using large `Field`'s in numeric generics fails with "array length too large"


michaeljklein:
### Aim

Attempted to use a `Field` larger than `u32::MAX` as a numeric generic:

```noir
global A: Field = 4294967297;

fn foo<let A: Field>() { }

fn main() {
    let C = foo::<A>();
}
```

### Expected Behavior

Expected the program to compile successfully or fail with a clear error

### Bug

Compilation fails with a message that the `Field` is too large to be an array length, but it's not used as an array length:

```bash
❯ ~/.nargo/bin/nargo compile
warning: unused variable C
   ┌─ src/main.nr:12:9
   │
12 │     let C = foo::<A>();
   │         - unused variable 
   │

error: Integer too large to be evaluated to an array-length
  ┌─ src/main.nr:7:19
  │
7 │ global A: Field = 4294967297;
  │                   ---------- Array-lengths may be a maximum size of usize::MAX, including intermediate calculations
  │

Aborting due to 1 previous error
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Nargo Version

nargo version = 0.34.0 noirc version = 0.34.0+5598059576c6cbc72474aff4b18bc5e4bb9f08e1 (git version hash: 5598059576c6cbc72474aff4b18bc5e4bb9f08e1, is dirty: false)

### Would you like to submit a PR for this Issue?

None

michaeljklein:
As a `noirc_frontend` test:

```rust
#[test]
fn use_global_larger_than_u32_as_generic() {
    let src = r#"
        // 2^32 + 1
        global A: Field = 4294967297;

        fn foo<let A: Field>() { }

        fn main() {
            let C = foo::<A>();
        }
    "#;

    let errors = get_program_errors(src);

    // TODO cleanup
    dbg!(&errors);

    assert_eq!(errors.len(), 0);
}
```

PR #6469, opened 2024-11-07
feat: replace `eval_global_as_array_length` with type/interpreter evaluation

michaeljklein:
# Description

## Problem\*

- Resolves https://github.com/noir-lang/noir/issues/6125
- Resolves https://github.com/noir-lang/noir/issues/6126
- Resolves https://github.com/noir-lang/noir/issues/6238

## Summary\*

All globals are now `comptime`

## Additional Context

Cyclic globals now stack overflow since that's the behavior of `comptime` cyclic globals before this PR (https://github.com/noir-lang/noir/issues/6780)

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Discussed in 1:1 that we're going to require globals to have their type declared explicitly which also fixes the issues in this PR.

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1647,25 +1647,19 @@ impl<'context> Elaborator<'context> {
...
 
-        let comptime = let_stmt.comptime;
-
-        let (let_statement, _typ) = if comptime {
-            self.elaborate_in_comptime_context(|this| this.elaborate_let(let_stmt, Some(global_id)))
-        } else {
-            self.elaborate_let(let_stmt, Some(global_id))
-        };
+        let (let_statement, _typ) = self
+            .elaborate_in_comptime_context(|this| this.elaborate_let(let_stmt, Some(global_id)));
```
asterite, line 1657: If I'm understanding this correctly, now all globals are elaborated at comptime, right?

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1647,25 +1647,19 @@ impl<'context> Elaborator<'context> {
             self.push_err(ResolverError::AbiAttributeOutsideContract { span });
         }
 
-        if !let_stmt.comptime && matches!(let_stmt.pattern, Pattern::Mutable(..)) {
+        let explicit_comptime = let_stmt.explicit_comptime;
+        if !explicit_comptime && matches!(let_stmt.pattern, Pattern::Mutable(..)) {
```
asterite, line 1651: Instead of erroring here we could error on the parser side. I think with that there would be no need to track `comptime` and `explicit_comptime` separately. Or at least this is the only use of `LetStatement.explicit_comptime` that I found.

compiler/noirc_frontend/src/hir/comptime/value.rs
```diff
@@ -502,19 +504,35 @@ impl Value {
...
+                | Self::I32(_)
+                | Self::I64(_)
+                | Self::U8(_)
+                | Self::U16(_)
+                | Self::U32(_)
+                | Self::U64(_)
+        )
+    }
+
+    /// Converts any unsigned `Value` into a `FieldElement`.
```
asterite, line 523:
```suggestion
    /// Converts any positive `Value` into a `FieldElement`.
```

compiler/noirc_frontend/src/tests.rs
```diff
@@ -1287,21 +1287,25 @@ fn lambda$f1(mut env$l1: (Field)) -> Field {
...
-    "#;
-
-    let errors = get_program_errors(src);
-    assert_eq!(errors.len(), 1);
-    assert!(matches!(
-        errors[0].0,
-        CompilationError::ResolverError(ResolverError::DependencyCycle { .. })
-    ));
-}
+// TODO(https://github.com/noir-lang/noir/issues/6780): currently failing
```
asterite, line 1290: We could include this test as `#[ignore]` maybe.

compiler/noirc_frontend/src/elaborator/mod.rs
asterite: Or... if we'll consider all globals as `comptime`, do we still need to tag them as such?

compiler/noirc_frontend/src/tests.rs
michaeljklein: Good point: that'll keep it compiling. Will do

compiler/noirc_frontend/src/elaborator/mod.rs
michaeljklein: Yes, that's correct

compiler/noirc_frontend/src/elaborator/mod.rs
michaeljklein: > Instead of erroring here we could error on the parser side.

I think validating here makes sense:
- Since we don't have a spec of what the elaborator expects to be handled/rejected by the parser
- To make it easier to fuzz/test from a parsed AST in the future

Should I add an additional error in the parser?

> Or at least this is the only use of LetStatement.explicit_comptime that I found.

`explicit_comptime` is primarily used for the formatter, i.e. a global `FOO` with `explicit_comptime=true` will be printed as `comptime FOO..`.

> Or... if we'll consider all globals as comptime, do we still need to tag them as such?

I replaced every occurence of `comptime` that's only used for globals with `explicit_comptime`, but the remaining occurrences are in let statements. I believe the possibilities are:

|                | explicit_comptime=true                 | explicit_comptime=false |
| -------------- | -------------------------------------- | ----------------------- |
| comptime=true  | comptime let foo / comptime global FOO | global FOO              |
| comptime=false | N/A                                    | let foo                 |

asterite: Thanks!

There's one LSP test that's failing because it's showing a global as `comptime global` instead of `global`. The problem is that the global definition doesn't keep track of whether it's an explicit comptime or not. So I guess we'd need to track that there too.

compiler/noirc_frontend/src/elaborator/types.rs
```diff
@@ -412,20 +410,57 @@ impl<'context> Elaborator<'context> {
...
                     .interner
                     .get_global_let_statement(id)
-                    .map(|let_statement| Kind::numeric(let_statement.r#type))
+                    .map(|let_statement| {
+                        if let_statement.r#type == Type::FieldElement {
+                            Kind::IntegerOrField
+                        } else {
+                            Kind::numeric(let_statement.r#type)
+                        }
+                    })
```
jfecher, line 419: What is this check for?

compiler/noirc_frontend/src/elaborator/types.rs
```diff
@@ -412,20 +410,57 @@ impl<'context> Elaborator<'context> {
...
+                    } else {
+                        self.push_err(ResolverError::NonIntegralGlobalType { span, global_value });
+                    }
+                    return None;
+                };
+
+                let global_value = match kind.ensure_value_fits(global_value, span) {
+                    Ok(global_value) => global_value,
+                    Err(_err) => {
+                        self.push_err(ResolverError::GlobalLargerThanKind {
```
jfecher, line 454: Keeping with the pattern above, this could be `let Ok(global_value) = kind.ensure_value_fits(...) else {` since the error is unused.

compiler/noirc_frontend/src/hir/comptime/value.rs
```diff
@@ -502,19 +504,35 @@ impl Value {
...
+            Self::Field(_)
+                | Self::I8(_)
+                | Self::I16(_)
+                | Self::I32(_)
+                | Self::I64(_)
+                | Self::U8(_)
+                | Self::U16(_)
+                | Self::U32(_)
+                | Self::U64(_)
+        )
```
jfecher, line 520: Can you tighten up the formatting here? Maybe importing the Value variants would format this onto fewer lines.

jfecher (changes requested):

compiler/noirc_frontend/src/elaborator/mod.rs
michaeljklein: Yes, I ended up added `explicit_comptime` to the HIR let statement definition to fix the hover LSP test. Though r.e. Jake's comment, I'm going to rename it to something like `is_global_let`.

compiler/noirc_frontend/src/hir/comptime/value.rs
michaeljklein: Ok, updated. I could also split it into `is_signed_integer` / `is_unsigned_integer` or similar

compiler/noirc_frontend/src/elaborator/types.rs
```diff
@@ -412,20 +410,57 @@ impl<'context> Elaborator<'context> {
...
-                        expr_kind: kind.to_string(),
-                        expr_span: path.span(),
-                    };
-                    self.push_err(error);
-                }
-                Some(Type::Constant(self.eval_global_as_array_length(id, path).into(), kind))
+                let Some(stmt) = self.interner.get_global_let_statement(id) else {
+                    if let Some(global) = self.unresolved_globals.remove(&id) {
+                        self.elaborate_global(global);
+                        return self.lookup_generic_or_global_type(path);
```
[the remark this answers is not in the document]
michaeljklein: I searched a couple ways for overlaps and pulled out the only one I found into [elaborate_unresolved_global](https://github.com/noir-lang/noir/pull/6469/files#diff-33a1f6d3758430e292c490fe5eaa04e77bcd3834ffe164c5714c22587a2cc133R1698).

I also thought more of this was duplicated somewhere, but perhaps it was part of the removed `eval_global_as_array_length`?

compiler/noirc_frontend/src/elaborator/types.rs
michaeljklein: It's a leftover method to ensure that field element's can be used polymorphically, but it's now unused so I've removed it.

jfecher:
image
image

Maybe we should look into memory usage in the interpreter, particularly now that we know `im::Vector` doesn't actually share anything until its length is > 128

compiler/noirc_frontend/src/hir_def/stmt.rs
```diff
@@ -31,17 +31,31 @@ pub struct HirLetStatement {
...
+    }
+
     pub fn ident(&self) -> HirIdent {
         match &self.pattern {
             HirPattern::Identifier(ident) => ident.clone(),
             _ => panic!("can only fetch hir ident from HirPattern::Identifier"),
         }
     }
+
+    pub fn runs_comptime(&self) -> bool {
```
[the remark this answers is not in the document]
michaeljklein: Good catch

jfecher (approved):
:+1:

asterite:
I think this broke Aztec-Packages contracts:

```
error: Comptime evaluation for builtin function 'as_field' is currently unimplemented
    ┌─ std/uint128.nr:143:17
    │
143 │         let f = crate::as_field(i);
    │                 ------------------
    │

error: Comptime evaluation for builtin function 'as_field' is currently unimplemented
    ┌─ std/uint128.nr:143:17
    │
143 │         let f = crate::as_field(i);
    │                 ------------------
```

I captured an issue for this, I'll try to fix it: https://github.com/noir-lang/noir/issues/6828

asterite:
That said, I don't know what are the valid types `T` for `as_field` and `from_field`... do you know?

jfecher:
@asterite only the various integer types. It's an old builtin from before we had traits. We really should remove it and replace it with `From<Field>` and `Into<Field>`

jfecher:
I think panicking is fine but we should probably verify this. We should at least check the most common uses of this method.

PR #6461, opened 2024-11-06
fix: Discard optimisation that would change execution ordering or that is related to call outputs

guipublic:
# Description

## Problem\*

Resolves #6451 and issue https://github.com/AztecProtocol/aztec-packages/issues/9488 (#6447)

## Summary\*
The eliminate intermediate variable optimisation is disabled when:
- the variable is output of a (brillig) call
- the execution ordering would break

## Additional Context
Eliminating a brillig output is wrong because we will not be able to solve the circuit
The execution ordering can easily be handled in a future PR.

## Documentation\*

Check one:
- [X] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [X] I have tested the changes locally.
- [X] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

Issue #6451, opened 2024-11-05
ICE on merge_expressions.rs


asterite:
### Aim

Reduced from https://github.com/zkpassport/noir_rsa/pull/2

Relevant code:

```noir
fn main(x: Field) {
    let y = unsafe { empty(x) };
    let mut value = 0;

    let term1 = x * x - x * y;
    std::as_witness(term1);
    value += term1;

    let term2 = x * x - y * x;
    value += term2;

    value.assert_max_bit_size::<1>();
}

unconstrained fn empty(_: Field) -> Field {
    0
}
```

### Expected Behavior

It compiles fine.

### Bug

It leads to a crash:

```
The application panicked (crashed).
Message:  assertion failed: i < b
Location: acvm-repo/acvm/src/compiler/optimizers/merge_expressions.rs:77
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #6476, opened 2024-11-07
chore: optimise older opcodes in reverse order

guipublic:
# Description

## Problem\*

Related to #6451 and #6461

## Summary\*
Handle the case where we can optimise an older opcode

## Documentation\*

Check one:

# PR Checklist\*


guipublic:
The increase in circuit size is not expected. This is worrying and we should not merge the PR which is supposed to improve (only a little bit but still) this number.

TomAFrench:
Closes https://github.com/noir-lang/noir/issues/6668?

acvm-repo/acvm/src/compiler/optimizers/merge_expressions.rs
```diff
@@ -46,61 +56,69 @@ impl MergeExpressionsOptimizer {
...
+                                Some(Opcode::AssertZero(expr_define)),
+                            ) = (target_opcode, source_opcode)
+                            {
+                                if let Some(expr) =
+                                    Self::merge_expression(&expr_use, &expr_define, w)
+                                {
+                                    self.modified_gates.insert(target, Opcode::AssertZero(expr));
+                                    self.deleted_gates.insert(source);
+                                    // Update the 'used_witness' map to account for the merge.
+                                    let mut witness_list = CircuitSimulator::expr_wit(&expr_use);
```
aakoshh, line 98: Huh, if I knew these existed I might have tried to use them in https://github.com/noir-lang/noir/pull/6695
Although there I was trying to avoid having to allocate new sets along the way.

acvm-repo/acvm/src/compiler/optimizers/merge_expressions.rs
```diff
@@ -109,17 +127,22 @@ impl MergeExpressionsOptimizer {
...
+            if let Some(op) = self.get_opcode(i, circuit) {
+                new_circuit.push(op);
+                new_acir_opcode_positions.push(*opcode_position);
             }
         }
         (new_circuit, new_acir_opcode_positions)
     }
 
-    fn brillig_input_wit<F>(&self, input: &BrilligInputs<F>) -> BTreeSet<Witness> {
+    fn brillig_input_wit(&self, input: &BrilligInputs<F>) -> BTreeSet<Witness> {
```
aakoshh, line 145: This could also be rewritten in terms of [WitnessFolder::fold_brillig_inputs](https://github.com/noir-lang/noir/blob/70cb16a2d38a1d6dc893371ade88e453ccc99070/acvm-repo/acvm/src/compiler/transformers/mod.rs#L324)

acvm-repo/acvm/src/compiler/optimizers/merge_expressions.rs
```diff
@@ -46,61 +56,69 @@ impl MergeExpressionsOptimizer {
...
+                                    self.deleted_gates.insert(source);
+                                    // Update the 'used_witness' map to account for the merge.
+                                    let mut witness_list = CircuitSimulator::expr_wit(&expr_use);
+                                    witness_list.extend(CircuitSimulator::expr_wit(&expr_define));
+                                    for w2 in witness_list {
+                                        if !circuit_inputs.contains(&w2) {
+                                            let mut v = used_witness[&w2].clone();
+                                            v.insert(b);
+                                            v.remove(&i);
+                                            used_witness.insert(w2, v);
```
aakoshh, line 105: Does this need to be cloned and re-inserted?  Maybe this would work:
```suggestion
                                            used_witness.entry(w2).and_modify(|v| {.clone();
                                                v.insert(b);
                                                v.remove(&i);
                                            };
```

acvm-repo/acvm/src/compiler/optimizers/merge_expressions.rs
```diff
@@ -46,61 +56,69 @@ impl MergeExpressionsOptimizer {
...
+                        };
+                        if i < b {
+                            if merge_opcodes(b, i) {
+                                // We need to stop here and continue with the next opcode
+                                // because the merge invalidates the current opcode.
+                                break;
+                            }
+                        } else if i != b {
+                            // Merge b into i
+                            if merge_opcodes(i, b) {
```
aakoshh, line 121:
```rust
// Merge the lower into the higher opcode.
if i != b && merge_opcodes(i.max(b), i.min(b)) {
    // We need to stop here and continue with the next opcode
    // because the merge invalidates the current opcode.
    break;
}
```
Just an idea, feel free to ignore.

aakoshh (approved):
This looks good as far as I understand it. 

So to be clear, what's happening is that previously we went through opcodes 0, 1, 2, etc and only considered merging `i` into `b` if `i < b`, so earlier opcodes got merged into later opcodes. We did this when we found a witness `w` that was only used by `i` and `b`. 

However, there could have been a witness `w'` that was used by `i`, `j` and `k`, and after merging `i` into `b` we may or may not have recognised that we can merge `j` into `k` now that `w'` was only used in two opcodes: if we visited `j` earlier than `i` then we wouldn't eliminate it. 

But now we can, because when we hit `k` we see that `j` can be merged into it 🎉 

Does this completely eliminate the need for doing multiple passes with the expression merger?

aakoshh:
> Does this completely eliminate the need for doing multiple passes with the expression merger?

I guess it doesn't, because if for example we have a witness that is only used by the last two opcodes, and we eliminate the one before last, it might unlock the merger of earlier opcodes that we won't be visiting again. 

For example:
Opcode | Witness A | Witness B 
:---: | :---: | :---:
0 | x | 
1 | x | 
2 | x | x
3 |   | x

When we reach opcode 2 we recognise that B is only used by 2 and 3, so we merge 2 into 3, but it's too late to merge 0 into 1 now that they are the only ones using A.

acvm-repo/acvm/src/compiler/optimizers/merge_expressions.rs
```diff
@@ -46,61 +56,69 @@ impl MergeExpressionsOptimizer {
...
+                                    self.modified_gates.insert(target, Opcode::AssertZero(expr));
+                                    self.deleted_gates.insert(source);
+                                    // Update the 'used_witness' map to account for the merge.
+                                    let mut witness_list = CircuitSimulator::expr_wit(&expr_use);
+                                    witness_list.extend(CircuitSimulator::expr_wit(&expr_define));
+                                    for w2 in witness_list {
+                                        if !circuit_inputs.contains(&w2) {
+                                            let mut v = used_witness[&w2].clone();
+                                            v.insert(b);
+                                            v.remove(&i);
```
aakoshh, line 104:
```suggestion
                                            v.insert(target);
                                            v.remove(&source);
```
`i` and `b` could be in any order now, right? We could be removing either one.

acvm-repo/acvm/src/compiler/optimizers/merge_expressions.rs
guipublic: Good point!

acvm-repo/acvm/src/compiler/optimizers/merge_expressions.rs
guipublic: Nice spot!

acvm-repo/acvm/src/compiler/optimizers/merge_expressions.rs
guipublic: Since merge_opcodes can now handled the two cases, I put this logic inside the function, I think it is a bit nicer like this.

guipublic:
> > Does this completely eliminate the need for doing multiple passes with the expression merger?
> 
> I guess it doesn't, because (...)

Yes I agree it does not eliminate all in one pass.

aakoshh (approved):
Nice!

Issue #6474, opened 2024-11-07
Mem2reg block parameter aliases handled incorrectly


jfecher:
### Aim

```noir
unconstrained fn main(mut x: i32) {
    mutate_alias(&mut x, &mut x);
    assert_eq(x, 1);
}

unconstrained fn mutate_alias(x: &mut i32, y: &mut i32) {
    *x = 0;
    *y = 1;
    assert_eq(*x, 1);
}
```

### Expected Behavior

The test to pass

### Bug

Test fails because we wrongly optimize the `assert_eq(*x, 1)` to `assert_eq(0, 1)`, wrongly assuming that `x` can't be aliased to `y`.

To fix this, parameters should be pessimistically assumed to be aliases to all other parameters.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #6477, opened 2024-11-07
fix: Treat all parameters as possible aliases of each other

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6474

## Summary\*

We allow mutable aliasing so all parameters can potentially be mutably aliased. We were over-eager in an optimization previously where we assumed parameters weren't aliased, and thus optimized out loads that were needed in some cases.

I've tried to preserve the optimization while fixing the bug best I could. This check is now type-aware so we assume parameters are possibly aliased with any other parameter of the same type, but a `&mut i32` cannot alias a `&mut Field`. The only caveat is that we can't really handle arrays or reference types containing other references within so for these we revert to the pre-optimization behavior of assuming `AliasSet::unknown()` for all references since these nested references may also reference other parameters but we don't have the ValueIds to be able to say "these inner references may _only_ reference other parameters (and not other references inside the function).

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
@@ -330,12 +330,43 @@ impl<'f> PerFunctionContext<'f> {
     fn add_aliases_for_reference_parameters(&self, block: BasicBlockId, references: &mut Block) {
         let dfg = &self.inserter.function.dfg;
         let params = dfg.block_parameters(block);
-        let params = params.iter().filter(|p| dfg.value_is_reference(**p));
+
+        let mut aliases: HashMap<Type, AliasSet> = HashMap::default();
 
         for param in params {
-            let expression =
-                references.expressions.entry(*param).or_insert(Expression::Other(*param));
-            references.aliases.entry(expression.clone()).or_insert_with(|| AliasSet::known(*param));
+            match dfg.type_of_value(*param) {
+                // If the type indirectly contains a reference we have to assume all references
+                // are unknown since we don't have any ValueIds to use.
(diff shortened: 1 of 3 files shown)


compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -324,18 +328,50 @@ impl<'f> PerFunctionContext<'f> {
...
+                    .or_insert_with(|| AliasSet::known(alias));
+            } else if let Some(first) = aliases.first() {
+                let expression = Expression::Other(first);
+                let previous = references.aliases.insert(expression.clone(), aliases.clone());
+                assert!(previous.is_none());
+
+                aliases.for_each(|alias| {
+                    let previous = references.expressions.insert(alias, expression.clone());
+                    assert!(previous.is_none());
+                });
```
TomAFrench, line 373: Do we need the first branch here? Seems like the second can cover both situations (also we can assert that `aliases.first()` is `Some` as we should never end up with an empty or unknown `AliasSet` in this iter.

TomAFrench (approved):
Fine to merge but I think we have a small amount of duplication which can be removed.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
jfecher: I think you're right. Since we expect parameters to not already be in each map the `entry` versus `insert` distinction isn't needed and the only difference afterward is some inexpensive cloning of `expression`. I'll update it.

PR #6481, opened 2024-11-08
feat: use a full `BlackBoxFunctionSolver` implementation when execution brillig during acirgen

TomAFrench:
…

# Description

## Problem\*

Resolves 

## Summary\*

We currently do not solve any field specific blackbox functions when solving brillig calls during acirgen which results in less optimized acir. This PR updates the acir context so that we pass in a full `Bn254BlackBoxSolver`

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
This doesn't affect our benchmarks as we don't use any of these opcodes in our test suite but it gives improvements in the bignum library

michaeljklein (approved):
LGTM

PR #6489, opened 2024-11-08
feat: SSA parser

### LINKED ISSUE #6475: Have a way to parse SSA
### Problem

Right now we have some tests for SSA, but they have this form:

```rust
// fn main f0 {
//   b0(v0: u16):
//     v1 = cast v0 as u32
//     v2 = cast v0 as u32
//     constrain v1 v2
// }
//
// After constructing this IR, we run constant folding which should replace the second cast
// with a reference to the results to the first. This then allows us to optimize away
// the constrain instruction as both inputs are known to be equal.
//
// The first cast instruction is retained and will be removed in the dead instruction elimination pass.
let main_id = Id::test_new(0);

// Compiling main
let mut builder = FunctionBuilder::new("main".into(), main_id);
let v0 = builder.add_parameter(Type::unsigned(16));

let v1 = builder.insert_cast(v0, Type::unsigned(32));
let v2 = builder.insert_cast(v0, Type::unsigned(32));
builder.insert_constrain(v1, v2, None);

let mut ssa = builder.finish();
let main = ssa.main_mut();
let instructions = main.dfg[main.entry_block()].instructions();
assert_eq!(instructions.len(), 3);
```

Essentially, we have a comment for the SSA function we are testing, then code that programmatically creates this.

This works, but:
- it's a slow process, or at least slower than being able to translate the comment to the SSA code
- the code that generates the code could potentially be not generating the code it's supposed to generate, or the two could fall out of sync (or we might need to maintain both so that they are in sync)

So, we could introduce some way to go from SSA code into an `Ssa` object:

```rust
let src = "
    fn main f0 {
    b0(v0: u16):
        v1 = cast v0 as u32
        v2 = cast v0 as u32
        constrain v1 v2
    }
    ";
```

### Happy Case

There's a way to go from SSA source code to an `Ssa` object.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #6475

## Summary

This PR introduces the `Ssa::from_src` method to parse SSA from text into `Ssa`. The code is first parsed into an intermediate (new) AST, then from it an SSA is created by using the existing `FunctionBuilder`.

Then, some existing tests are updated to use this form of creating the SSA instead of manually doing it with `FunctionBuilder`.

Some notes:
- In the tests that I updated I also updated how the final SSA is checked: now we check that the string representation matches the one we expect, instead of manually checking some things. I think this is a bit more robust and shorter. And then, if we trust SSA parsing and SSA printing is correct (and I think we do because when we debug things we rely on the SSA output we get) then I think it should be fine to rely on this.
- I had to change how some SSA are printed, specifically to include some types that otherwise can't be determined when parsing. For example, arrays flatten their values, so if we have `[Field 1, Field 1]` it could actually be an array of `(Field, Field)` but we can't know that. So, now arrays include their type, like this: `[Field 1, Field 1] of Field` (or whatever type they are). Other instructions like `call`, `allocate` and `load` specify what type they return with ` -> types...` at the end. I chose that syntax so that the types come at the end and one can still focus on the important bits on the left side.
- The `IfElse` instruction is still not parsed in this PR. I couldn't find an example of it. But we can include it in a follow-up PR.
- The optional assertion message in `constrain` and `range_check` isn't parsed (yet). This could be done in a follow-up PR, but my thinking here was that it's not important for tests, and it seems no tests actually use these strings (but it could be good to have for completeness)
- I didn't update all tests to avoid making this PR bigger. I updated some tests to verify that the parser is working and that most, if not all, needed features are there. But we can update them in follow-up PRs, and even add more tests to existing passes that aren't tested.

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/ssa/parser/into_ssa.rs
```diff
@@ -0,0 +1,309 @@
...
+        for function in &parsed_ssa.functions {
+            let function_id = FunctionId::new(function_id_counter);
+            function_id_counter += 1;
+
+            functions.insert(function.internal_name.clone(), function_id);
+        }
+
+        let mut translator =
+            Self { builder, functions, variables: HashMap::new(), blocks: HashMap::new() };
+        translator.translate_function_body(main_function)?;
```
aakoshh, line 59: Why is it that main is translated here, but all the other functions are done in `into_ssa` above? I was wondering why `main` is not processed twice when I saw the `.remove(0)`. Would it be equally correct to loop through all the rest here?

asterite: Good question! I tried to make it like you said, but it ends to a panic.

The reason is that it seems `FunctionBuilder` needs to be created with the main function. Once you do that you can call `set_runtime` to set the main function's runtime.

Then other functions need to be created with `new_function` or `new_brillig_function`... so dealing with the main function and the rest of the functions has to be done in different ways.

Maybe to have things a bit more uniform, we could change `FunctionBuilder` to be created without an initial function. Then you can create new functions and trying to insert an instruction before creating an initial function would panic. I guess it's not done like that to avoid an `Option` that could panic. But... we could maybe do this in a separate PR! In any case this special handling is not that much code, so maybe it's fine.

aakoshh: Right, yes, I saw that the builder needs a main, thanks for pointing out the `new_function` vs `new_brillig_function` as well; I didn't think that the special handling of the first function was to be avoided, just found it odd that the main is _removed_ from the mutable `parsed_ssa`, and then the translator is returned to be fed more: at that point it looks like the only reasonable thing to do with the `parsed_ssa` is to consume all of its functions anyway, so I was wondering why it's not done in `new`. But it's all private, so ultimately doesn't matter that much, it was just like "why is it a mutable ref?" when I first saw it.

jfecher: >  I guess it's not done like that to avoid an Option that could panic. But... we could maybe do this in a separate PR! 

It's done for that reason and to avoid the Option entirely so that it is somewhat faster not having to unwrap/expect the current function on every value/instruction insertion.

aakoshh: I would perhaps do it like 
```rust
impl Translator {
  pub fn translate(mut ssa: ParsedSsa) -> Result<Ssa, SsaError> {
    let mut t = Self::new(&mut ssa);
    for f in ssa.function { 
      t.translate_function(f);
    }
    t.finish()
  }
}
```
And then there is no question about who owns the `ssa` during construction.

asterite: Ah, good point. I also thought it wasn't correct to have `new` modify `ParsedSsa`, then the caller doing more things. Making it a method of `Translator` makes more sense 👍

asterite: (but I wasn't sure how to exactly improve it)

asterite: Done! I also added a few comments and renamed a function.

compiler/noirc_evaluator/src/ssa/parser.rs
```diff
@@ -0,0 +1,830 @@
...
+
+        if let Some(instruction) = self.parse_range_check()? {
+            return Ok(Some(instruction));
+        }
+
+        if let Some(instruction) = self.parse_store()? {
+            return Ok(Some(instruction));
+        }
+
+        if let Some(target) = self.eat_identifier()? {
```
aakoshh, line 199: Would perhaps `parse_assignment` be an adequate name for this `if` and its body? It's quite long, all the others are one liners. I was just looking for when `Ok(None)` is returned and it would make it easier to see that it's not from within anywhere in this.

compiler/noirc_evaluator/src/ssa/opt/array_set.rs
```diff
@@ -206,135 +206,47 @@ fn make_mutable(
...
-        let v17 = builder.insert_binary(v5, BinaryOp::Add, one);
-        builder.terminate_with_jmp(b1, vec![v17]);
-
-        builder.switch_to_block(b2);
-        builder.terminate_with_return(vec![]);
+        let src = "
+            brillig(inline) fn main f0 {
+              b0():
+                v1 = allocate -> [Field; 5]
+                store [[Field 0, Field 0, Field 0, Field 0, Field 0] of Field, [Field 0, Field 0, Field 0, Field 0, Field 0] of Field] of [Field; 5] at v1
```
aakoshh, line 220: How do we interpret this? I thought `v1` would be an array of 5 fields, but it looks like `[[Field; 5]; 2]` 🤔

compiler/noirc_evaluator/src/ssa/parser.rs
asterite: Good idea! I didn't do it because to extract each instruction into a method it couldn't work because `target` would have moved, but at least extracting `parse_assignment` for that entire section makes things a bit simpler.

I'm thinking the assignment instruction could be split into `target` and `rest` so we can parse rest without having to move `target`, but I don't know if it's worth it. What do you think?

compiler/noirc_evaluator/src/ssa/opt/array_set.rs
asterite: Ah, right, I was also a bit confused by this.

The type after `of` is the type of the elements, not the type of the array itself.

compiler/noirc_evaluator/src/ssa/parser.rs
aakoshh: I don't fully understand what "moving the target" means, but I think the following would be the nicest pattern in the context of what is above it:

```rust
        if let Some(instruction) = self.parse_store()? {
            return Ok(Some(instruction));
        }

        if let Some(instruction) = self.parse_assignment()? {
            return Ok(Some(instruction));
        }

        Ok(None)
```

If there were multiple options to parse `target` then it would be different, but it looks like `parse_assignment` is the only one. If it was like this:
```
if let Some(target) = self.eat_identifier()? {
   if let Some(ins) = self.parse_foo(target)? { ... }
   if let Some(ins) = self.parse_bar(target)? { ... }
   return Err(self.expected_foo_or_bar());
}
```
then I think it think the placement of `return self.expected_instruction_or_terminator();` would be more justified, but the way it is it might as well be in `parse_assignment`.

compiler/noirc_evaluator/src/ssa/parser.rs
```diff
@@ -0,0 +1,792 @@
...
+
+    fn parse_assignment(&mut self, target: Identifier) -> ParseResult<Option<ParsedInstruction>> {
+        let mut targets = vec![target];
+
+        while self.eat(Token::Comma)? {
+            let target = self.eat_identifier_or_error()?;
+            targets.push(target);
+        }
+
+        self.eat_or_error(Token::Assign)?;
```
aakoshh, line 311: So at this point we're sure that we are dealing with `<target> = `, right? After that an `Ok(None)` value wouldn't be valid, would it?

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -940,47 +848,31 @@ mod test {
...
-        builder.terminate_with_return(vec![v2]);
-
-        let ssa = builder.finish();
+        let src = "
+            acir(inline) fn main f0 {
+              b0(v0: [Field; 2]):
+                inc_rc v0
+                inc_rc v0
+                inc_rc v0
+                v2 = array_get v0, index u32 0 -> Field
```
aakoshh, line 857: Could the parser or the builder plausibly figure out that an `array_get` on `v0: [Field; 2]` means that `v2` is of type `Field` without having to add `-> Field`?

compiler/noirc_evaluator/src/ssa/parser.rs
asterite: Done!

With "moving the target" I meant that right now `parse_assignment` is still a bit long. It could be split into helper functions but `self.parse_foo(target)?` won't work because `target` moves to `parse_foo` and then we can't call `parse_bar`.

compiler/noirc_evaluator/src/ssa/opt/die.rs
asterite: It's a great question, and it's something I wondered at least twice when writing this PR.

Consider this code:

```noir
    let array = [(1, true)];
    let value = array[x];
    value.0
}
```

When accessing an array index, when the array is composite, what Noir will end up doing is accessing each element in the composite type in turn by shifting an index. Here's the SSA the above initially produces:

```
acir(inline) fn main f0 {
  b0(v0: Field):
    inc_rc [Field 1, u1 1] of (Field, u1)
    v4 = cast v0 as u32
    v6 = mul v4, u32 2
    v7 = array_get [Field 1, u1 1] of (Field, u1), index v6 -> Field
    v9 = add v6, u32 1
    v10 = array_get [Field 1, u1 1] of (Field, u1), index v9 -> u1
    return v7
}
```

that's with the type annotations that are added in this PR. Two things to note:
1. An array has all the values flattened, so `1` is on position 0, `true` is on position 1.
2. `array_get` works on these indices, not on the actual indices on the Noir side, so accessing 0 will get you 1, accessing 1 will get you `true`
3. When we do `array_get` we don't know which type will get. Well, actually we can guess because the first one will be one the first type, the second one on the second type, etc... but I didn't want to make the parsing too complex so I decided to just put the expected types in the code. Maybe it also makes it easier to understand the code.

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -317,531 +317,305 @@ mod test {
...
-        let mut ssa = builder.finish();
+        let src = "
+            acir(inline) fn main f0 {
+              b0(v0: Field):
+                v1 = add v0, Field 1
+                v2 = mul v1, Field 3
+                return v2
+            }
+            ";
+        let mut ssa = Ssa::from_str(src).unwrap();
```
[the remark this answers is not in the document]
asterite: Right, I thought about this but was too lazy to do it. I'd need to translate spans to lines/columns... the error is more r less understandable, like it would say that it found some token when it was expecting some other token, so you could guess where the error is. But I agree that showing the error where it's located will be better for all of us in the future... but I think `unwrap` will still show the `Debug` version of the error? Or how does it work?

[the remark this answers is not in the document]
aakoshh: Yes I think `.expect` and `.unwrap` need `Debug`. Just kinda remember the raw compilation errors being somewhat difficult to parse in test failures, but we can add a helper any time 👍

compiler/noirc_evaluator/src/ssa/opt/die.rs
aakoshh: Amazing, thanks for that explanation! I agree in that case it would be difficult to figure out what `index v9` might be, and also agree that it makes it easier to follow 👍

compiler/noirc_evaluator/src/ssa/parser.rs
```diff
@@ -0,0 +1,788 @@
...
+            return Err(ParserError::MultipleReturnValuesOnlyAllowedForCall {
+                second_target: targets[1].clone(),
+            });
+        }
+
+        let target = targets.remove(0);
+
+        if self.eat_keyword(Keyword::Allocate)? {
+            self.eat_or_error(Token::Arrow)?;
+            let typ = self.parse_type()?;
```
aakoshh, line 327: Based on your illustration of how arrays are indexed now I remember that they are all flattened and index arithmetic is used to pick the right element, and when you call `allocate` you just get a pointer to the first element, which I assume could be a slice or an array. 

I notice that the parser accepts any type for an allocation, but the printer would [panic](https://github.com/noir-lang/noir/pull/6489/files#diff-58044f084c8337680eded1741fe40937b6a0e79d4250e5d1b9651de091de99c5R199) if it would be anything other than a reference type. If I'm not mistaken that `Type` is the same as this. Shouldn't this reject non-reference types to catch the problem before the print?

compiler/noirc_evaluator/src/ssa/opt/array_set.rs
aakoshh: Great, TIL that allocate works on references, and I see the original version also does the same thing: `let v3 = builder.insert_allocate(array_type.clone());`, rather than `.insert_allocate(nested_array_type...`

Maybe some package level example would be helpful to illustrate and emphasise this syntax, but now that you spelled it out, it makes sense, thanks!

compiler/noirc_evaluator/src/ssa/parser.rs
aakoshh: Ah, no problem there, I see that `FunctionBuilder::insert_allocate` wraps this type into a reference 👍

aakoshh (approved):
Nice, I should print out the code and put it on my wall! 😃

asterite:
I'll merge this now but try to improve the parsing error message in a follow-up PR.

compiler/noirc_evaluator/src/ssa/parser.rs
aakoshh: So you're suggesting that instead of this:
```
v1 = allocate -> [Field; 5]
```
it would look like this:
```
v1 = allocate -> &mut [Field; 5]
```

Speaking as a newcomer to the codebase, I didn't know all references were mutable; for example looking at this SSA, I thought e.g. `&[Field; 3]` would be an option:
```
brillig(inline) fn push f1 {
  b0(v0: &mut [Field; 3], v1: &mut u32, v2: Field):
```
I finally found out from the print format itself that the `"&mut"` part comes from the [type display format](https://github.com/noir-lang/noir/blob/b65a63d8d898e46cc686baa500f0b8070e45df14/compiler/noirc_evaluator/src/ssa/ir/types.rs#L229), there is no other option. Obviously in the context of a function parameter where it could be passed by value or by mutable reference this differentiation is crucial. 

In the case of `allocate` it's an interesting conundrum. If we all know that `allocate` allocates an `&mut` and the input is a `Type`, then I think we shouldn't have to type the `&mut` just for the sake of it, but the `->` syntax suggests that `allocate` "returns" this thing, which again if you know what `allocate` does, then it makes sense to you, otherwise it might not. For example `allocate::<[Field; 5]>` makes me recognise that it's an input, not an output.

So I agree that `allocate -> &mut <type>` would be more intuitive, because that's how the `Type::Reference` would print itself anyway, and that's what `allocate` inserted, so that's what it should return. Whether people would be annoyed by having to type extra characters I don't know.

asterite: I'll make this change and also improve error messages before merging.

compiler/noirc_evaluator/src/ssa/opt/mod.rs
```diff
@@ -21,3 +22,19 @@ mod resolve_is_unconstrained;
...
+    ssa.normalize_ids();
+
+    let ssa = ssa.to_string();
+    let ssa = trim_leading_whitespace_from_lines(&ssa);
+    let expected = trim_leading_whitespace_from_lines(expected);
+
+    if ssa != expected {
+        println!("Expected:\n~~~\n{}\n~~~\nGot:\n~~~\n{}\n~~~", expected, ssa);
+        similar_asserts::assert_eq!(expected, ssa);
+    }
```
aakoshh, line 39: I'm now trying to migrate [this test](https://github.com/noir-lang/noir/blob/82046be7790bb7128f481ace94b8a6841f76c88d/compiler/noirc_evaluator/src/ssa/opt/unrolling.rs#L516) to use your parser. It turns out that both input and output SSA in the comments are invalid. I figured out how to fix the input based on the errors that `Ssa::from_str(src).unwrap();` printed, but then got here and got confused by the output of `assert_ssa_equals(ssa, expected);`:

```text
Differences (-left|+right):
-fn main f0 {
-b0():
-    constrain Field 0
-    constrain Field 0
-    constrain Field 0
-    constrain Field 0
-    jmp b23()
-b23():
-    constrain Field 0
-    constrain Field 0
-    constrain Field 0
-    constrain Field 0
-    jmp b27()
-b27():
-    constrain Field 0
-    constrain Field 0
-    constrain Field 0
-    constrain Field 0
-    jmp b31()
-b31():
+acir(inline) fn main f0 {
+  b0():
+    constrain u1 0 == Field 1
+    constrain u1 0 == Field 1
+    constrain u1 0 == Field 1
+    constrain u1 0 == Field 1
+    jmp b1()
+  b1():
+    constrain u1 0 == Field 1
+    constrain u1 0 == Field 1
+    constrain u1 0 == Field 1
+    constrain u1 0 == Field 1
+    jmp b2()
+  b2():
+    constrain u1 0 == Field 1
+    constrain u1 0 == Field 1
+    constrain u1 0 == Field 1
+    constrain u1 0 == Field 1
     jmp b3()
-b3():
+  b3():
+    jmp b4()
+  b4():
     return Field 0
 }
```

There are two problems:
1. `similar_asserts::assert_eq!(expected, ssa);` uses the arguments in the opposite order as `assert_ssa_equals(mut ssa: super::Ssa, expected: &str)` receives them, so it's a bit confusing that _left_ in the output corresponds to _right_ in the input. FWIW I would use `assert_eq!(ssa, expected)`. 
2. I was thrown off by seeing expected `constrain Field 0` when I knew that `constrain v5` did not parse, it expected `==` to follow. I didn't understand why it's the one we expect in the end, how did it simplify my input `constrain v5 == Field 1` into this? Then I realised that it's coming from an unparsed and invalid string that was in the test.

On the 2nd point think it would be helpful if, in case there is a mismatch in the strings, to try to parse the expectation and print a warning that it's not even valid. It's okay to print both values because the parsed and processed output might help understand what it should be (assuming it's somewhat correct), but just so I stop treating it as gospel.

asterite:
I pushed a commit that will show errors in a better way. For example, here's one output where there's a typo:

```
called `Result::unwrap()` on an `Err` value:
            acir(inline) fn main f0 {
              b0(v0: Field):

                v1 = ad v0, Field 1
                     ^^
                     Expected an instruction or terminator, found ad

                v2 = mul v1, Field 3
                return v2
            }
```

compiler/noirc_evaluator/src/ssa/opt/mod.rs
asterite: Good ones! I just pushed fixed for both.

aakoshh: Cool, a panic is good too, I guess we can print the SSA or procure an example from somewhere to figure out what it should be 🙏

asterite:
I also renamed `assert_ssa_equals` to `assert_normalized_ssa_equals` because I figured it might be confusing if you print an SSA and get one value, that that value isn't equal to what that assertion is producing...

PR #6460, opened 2024-11-05
chore: Rework mem2reg load & store removal

jfecher:
# Description

## Problem\*

Resolves 

## Summary\*

Draft so that I can compare opcode differences as I work on this PR. Currently I've removed all the extra load/store removals but haven't added them back so initial differences are expected to be bad, similar to https://github.com/noir-lang/noir/pull/6434.

The goal is to rework the additions to mem2reg (notably all the different ways loads & stores are removed) to be more unified and to handle aliases better.

## Additional Context

The end goal of this PR is not necessarily to improve brillig bytecode size (although that'd be a nice bonus) but to rework how we handle load & store removal in mem2reg currently to be less buggy so that https://github.com/noir-lang/noir/pull/6434 can re-enable this feature without bugs and fix the opcode counts in that PR.

## Documentation\*

Check one:
- [ ] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [ ] I have tested the changes locally.
- [ ] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #6490, opened 2024-11-08
chore: Rework DIE pass to operate on a postorder SCC CFG

jfecher:
# Description

## Problem\*

This is currently preventing DIE from being able to remove side-effectful instructions like stores or increment_rcs directly. Instead we rely on either special checks in mem2reg (for stores) or special checks in DIE for inc/decrement rcs.

## Summary\*

This is part of a breakout change for https://github.com/noir-lang/noir/pull/6460 since that PR changes mem2reg to use DIE to remove stores, but it needs this change to handle side-effectful instructions in the presence of loops. When loops are in the program and DIE is run, it previously could remove a store that would be used again next iteration of the loop.

This PR changes DIE to use a graph of strongly-connected components of blocks and to operate on every block within an SCC all at once. This requires a more traditional DIE algorithm which repeats until there are no more changes. This isn't necessary in ours currently since acir is always 1 block and normal scoping semantics prevent creating brillig programs which refer to the same instruction across multiple blocks (unless it is a store).

## Additional Context

Currently a draft while I just verify it does work while merged into #6460, and to see if this has any brillig opcode regressions.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
Closing this to work on an alternate solution for mem2reg (this was only needed for https://github.com/noir-lang/noir/pull/6460)

PR #6491, opened 2024-11-08
Expose poseidon internals

nventuro:
We rely on these functions quite heavily in aztec-packages - it'd be great to have them be public, at least until an alternative API is implemented.


TomAFrench:
Why are you not using the `Hasher` api for this?

PR #6494, opened 2024-11-11
chore: add regression unit tests for failing canonicalization

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

We're getting build failures in https://github.com/AztecProtocol/aztec-packages/pull/9769 due to not being able to solve `[Field; (N - 1) + 1]` to `[Field; N]`. This PR adds a unit test showing a case which we should be able to simplify but aren't.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
@asterite Can you take a look at this? This is blocking a Noir sync atm.

compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
```diff
@@ -111,6 +111,10 @@ impl Type {
 
                 Type::CheckedCast { from: Box::new(from), to: Box::new(to) }
             }
+            Type::Array(length, element) => Type::Array(
+                Box::new(length.canonicalize_helper(found_checked_cast, run_simplifications)),
+                element,
+            ),
```
jfecher, line 117: I can see what this does but can you elaborate why it was added? The intent of canonicalize was to only work on numeric types. If we start trying to find and canonicalize types recursively like this we're missing a bunch of cases, e.g. numerics in struct generics, handling type variables, etc.

jfecher (changes requested):

compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
jfecher: I can see this is blocking a sync but I still want to ensure its the right solution before just merging it

PR #6505, opened 2024-11-12
feat(ssa): Unroll small loops in brillig

### LINKED ISSUE #6470: Improve brillig execution speed by unrolling small loops
```noir

unconstrained fn __validate_gt_remainder(a_u60: [u64; 6]) -> [u64; 6] {
    let mut result_u60: [u64; 6] = [0; 6];

    for i in 0..6 {
        result_u60[i] = a_u60[i] + 1;
    }

    result_u60
}
```

Here we've got a for-loop with a small and simple body, this function compiles to:

```
brillig(inline) fn __validate_gt_remainder f2 {
  b0(v0: [u64; 6]):
    inc_rc v0
    inc_rc [u64 0, u64 0, u64 0, u64 0, u64 0, u64 0]
    v4 = allocate
    store [u64 0, u64 0, u64 0, u64 0, u64 0, u64 0] at v4
    jmp b1(u32 0)                                                 // loop boilerplate
  b1(v1: u32):
    v7 = lt v1, u32 6                                             // loop boilerplate
    jmpif v7 then: b3, else: b2                                   // loop boilerplate
  b3():
    v9 = load v4                                                  // loop boilerplate
    v10 = array_get v0, index v1                                  // logic
    v12 = add v10, u64 1                                          // logic
    v13 = array_set v9, index v1, value v12                       // logic
    v15 = add v1, u32 1                                           // loop boilerplate
    store v13 at v4                                               // loop boilerplate
    v16 = add v1, u32 1                                           // duplicate instruction
    jmp b1(v16)                                                   // loop boilerplate
  b2():
    v8 = load v4                                                  // loop boilerplate
    dec_rc v0
    return v8
}
```

We've then got `3` instructions of actual logic and `8` of boilerplate and when it comes to execution we have `3*num_iterations` instructions of actual logic and `2+6*num_iterations` of boilerplate. This gives us a tradeoff between bytecode size and execution speed.

We should aim to unroll any loops where `loop_instructions * num_iterations <= loop_instructions + 8` => `loop_instructions * (num_iterations - 1) <= 8` as this will reduce bytecode sizes. There's a decent chance that we'd get benefits for unrolling larger loops as well due to constant folding.

aakoshh:
# Description

## Problem\*

Resolves #6470 

## Summary\*

- [x] Look at the runtime during unrolling to preserve ref count instructions, which ACIR did not need and are currently eliminated, but Brilling would need to keep using
- [x] If the runtime is Brillig, enable unrolling the loops if the following check passes:
  - [x] Identify the loop header, the back edge and the pre-header (exists in the module already)
  - [x] Find the lower and upper bounds of the loop, which are expected to be turned into constants by inlining by the time we reach the `Ssa::unroll_loops_iteratively` pass. 
  - [x] Count all instructions in the loop
  - [x] Detect the presence of jumps that look like `break` or `continue` that we couldn't unroll
  - [x] If there are no blocks with terminators coming from `break` or `continue` instructions:
    - [x] Collect all references allocated before the pre-header, for the detection of boilerplate load/stores in the loop body
    - [x] Collect all load/store pairs of pre-header variables in the loop between header and back-edge, to estimate which instructions we can get rid of
    - [x] Compute `loop_instructions = all_instructions_in_loop - 2 * num_load_store_pairs - const_loop_jmp_overhead` 
    - [x] Compare `loop_instructions * num_iterations` vs `all_instructions_in_loop` to find out if it's worth unrolling

I added some extra docstrings to existing functions to illustrate what the code does, and also refactored functions into methods where they already had an obvious parameter corresponding to a `self`; I found it made it easier to orient myself. Sorry about that code churn, here's a [diff](https://github.com/noir-lang/noir/compare/2d14db5...7cd33fe) to make it easier to review, or we can break it up into two PRs.

## Additional Context

Take for example this code:
```rust
fn main(sum: u32) {
    assert(loop(10, 14) == sum);
}

fn loop(from: u32, to: u32) -> u32 {
    let mut sum = 0;
    for i in from..to {
        sum = sum + i;
    }
    sum
}
```

Here's the SSA before and after unrolling. 

```
After `static_assert` and `assert_constant`:
acir(inline) fn main f0 {
  b0(v0: u32):                         // Pre-header
    v2 = allocate -> u32
    store u32 0 at v2                  // This happens even if we do 4 additions in `main`
    jmp b1(u32 10)                     // -: The lower bound is a constant
  b1(v1: u32):                         // Loop header
    v6 = lt v1, u32 14                 // -: the upper bound is a constant
    jmpif v6 then: b3, else: b2        // -: loop one more time?
  b3():
    v9 = load v2 -> u32                // -: Loading v2 from the pre-header
    v10 = add v9, v1                   // +: sum = sum + 1
    store v10 at v2                    // -: Storing modified v2 for next loop
    v12 = add v1, u32 1                // -: Increment loop counter
    jmp b1(v12)                        // -: Back edge
  b2():
    v7 = load v2 -> u32
    v8 = eq v7, v0
    constrain v7 == v0
    return
}

After Unrolling:
acir(inline) fn main f0 {
  b0(v0: u32):
    v1 = allocate -> u32
    store u32 0 at v1
    v3 = load v1 -> u32
    v5 = add v3, u32 10
    store v5 at v1
    v6 = load v1 -> u32
    v8 = add v6, u32 11
    store v8 at v1
    v9 = load v1 -> u32
    v11 = add v9, u32 12
    store v11 at v1
    v12 = load v1 -> u32
    v14 = add v12, u32 13
    store v14 at v1
    jmp b1()
  b1():
    v15 = load v1 -> u32
    v16 = eq v15, v0
    constrain v15 == v0
    return
}
```

Lines marked `// -:` can be removed if we unroll the loops, and then running `mem2reg` gets rid of the redundant `store` and `load` operations:
```
After Mem2Reg (2nd):
acir(inline) fn main f0 {
  b0(v0: u32):
    v1 = allocate -> u32
    v3 = eq u32 46, v0
    constrain u32 46 == v0
    return
}
```

## Documentation\*

Check one:

# PR Checklist\*


aakoshh:
Investigated the increased Brillig opcode count in `fold_2_to_17` like this:
```shell
cargo run -p nargo_cli --  --program-dir . info --force --force-brillig
```
while also printed out which loops were small enough to unroll. 
One of them had these stats:
```
BoilerplateStats { iterations: 3, loads: 16, stores: 4, increments: 1, all_instructions: 32 }
```
The working theory was that if we unroll them we'll get rid of all the stores and the loads, but the original idea was to only do it for pairs. 

Looking at the SSA of this particular loop before unrolling:
```
b28():
    jmp b29(u32 0)
  b29(v6: u32):
    v202 = lt v6, u32 3
    jmpif v202 then: b32, else: b30
  b32():
    v214 = load v15 -> [Field; 3]
    v215 = load v18 -> [Field; 4]
    v216 = load v19 -> u32
    v217 = load v21 -> u1
    v218 = lt v6, v216
    jmpif v218 then: b33, else: b34
  b33():
    v219 = load v15 -> [Field; 3]
    v220 = load v18 -> [Field; 4]
    v221 = load v19 -> u32
    v222 = load v21 -> u1
    v223 = load v15 -> [Field; 3]
    v224 = load v18 -> [Field; 4]
    v225 = load v19 -> u32
    v226 = load v21 -> u1
    v227 = array_get v224, index v6 -> Field
    v228 = load v15 -> [Field; 3]
    v229 = load v18 -> [Field; 4]
    v230 = load v19 -> u32
    v231 = load v21 -> u1
    v232 = array_get v228, index v6 -> Field
    v233 = add v227, v232
    v234 = array_set v220, index v6, value v233
    v235 = add v6, u32 1
    store v219 at v15
    store v234 at v18
    store v221 at v19
    store v222 at v21
    jmp b34()
  b34():
    v236 = add v6, u32 1
    jmp b29(v236)
```
It is strange that we have `load v15 -> [Field; 3]` appear 3 times, and this is true even after `mem2reg` later.

vezenovm:
> It is strange that we have `load v15 -> [Field; 3]` appear 3 times, and this is true even after `mem2reg` later.

We had attempted to solve this previously (https://github.com/noir-lang/noir/pull/6088) but it was showing mixed benefits due to extra ref count instructions remaining due to the removed loads. It has since grown stale, but maybe it is worth reviving.

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -173,73 +201,511 @@ impl Loops {
...
+    ) -> Result<HashSet<ValueId>, CallStack> {
+        // We need to traverse blocks from the pre-header up to the block entry point.
+        let pre_header = self.get_pre_header(function, cfg)?;
+        let function_entry = function.entry_block();
+
+        // The algorithm to collect blocks in the loop is fine for us.
+        let blocks = Self::find_blocks_in_loop(function_entry, pre_header, cfg).blocks;
+
+        // Collect allocations in all blocks above the header.
+        let allocations = blocks.iter().flat_map(|b| {
```
vezenovm, line 521:
```suggestion
        // Collect allocations in all blocks above the header.
        let allocations = self.blocks.iter().flat_map(|b| {
```
Don't we have the blocks for the loop already?

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -173,73 +201,511 @@ impl Loops {
...
+        let Ok(Some((lower, upper))) = self.get_const_bounds(function, cfg) else {
+            return None;
+        };
+        let Some(lower) = lower.try_to_u64() else {
+            return None;
+        };
+        let Some(upper) = upper.try_to_u64() else {
+            return None;
+        };
+        let Ok(refs) = self.find_pre_header_reference_values(function, cfg) else {
```
vezenovm, line 613: This method, `count_loads_and_stores`, and `count_all_instructions` all need to iterate over all blocks in the loop.  We could instead iterate over the loop blocks once and have per block methods.

vezenovm: This probably isn't a big deal as going through only the blocks again shouldn't have a crazy cost, but just noting it.

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -395,14 +822,19 @@ impl<'f> LoopIteration<'f> {
...
                 call_stack,
-            } => self.handle_jmpif(
-                *condition,
-                *then_destination,
-                *else_destination,
-                call_stack.clone(),
-            ),
+            } =>
+            // Check where the loop body wants to jump (e.g if-then blocks).
+            {
```
vezenovm, line 827: nit: why this extra block `{ .. }`? For the comment? I think code states pretty well that we want to handle a if-then block

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -369,14 +791,18 @@ impl<'f> LoopIteration<'f> {
...
         self.induction_value
             .expect("Expected to find the induction variable by end of loop iteration")
     }
 
     /// Unroll a single block in the current iteration of the loop
     fn unroll_loop_block(&mut self) -> Vec<BasicBlockId> {
         let mut next_blocks = self.unroll_loop_block_helper();
+        // If the loop body ends with a JmpIf then eliminate either destinations
+        // which are not one of the block we have seen as a destination of a terminator
+        // of a block we have already inlined.
```
vezenovm, line 805:
```suggestion
        // or a block we have already inlined.
```
?

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -472,23 +910,30 @@ impl<'f> LoopIteration<'f> {
         // instances of the induction variable or any values that were changed as a result
         // of the new induction variable value.
         for instruction in instructions {
-            // Skip reference count instructions since they are only used for brillig, and brillig code is not unrolled
-            if !matches!(
-                self.dfg()[instruction],
-                Instruction::IncrementRc { .. } | Instruction::DecrementRc { .. }
-            ) {
-                self.inserter.push_instruction(instruction, self.insert_block);
+            if self.skip_ref_counts && self.is_refcount(instruction) {
```
vezenovm, line 913:
```suggestion
            if self.inserter.function.runtime().is_acir() && self.is_refcount(instruction) {
```

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -330,6 +748,9 @@ struct LoopIteration<'f> {
     /// This is None until we visit the block which jumps back to the start of the
     /// loop, at which point we record its value and the block it was found in.
     induction_value: Option<(BasicBlockId, ValueId)>,
+
+    /// Whether to discard reference count instructions, which ACIR doesn't need but Brillig does.
+    skip_ref_counts: bool,
```
vezenovm, line 753:
```suggestion
```
We should not need this field with the `inserter`. If we want to reduce the need to call `self.inserter.function.runtime().is_acir()` we can make a method inside of the `LoopIteration` context.

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -340,6 +761,7 @@ impl<'f> LoopIteration<'f> {
         source_block: BasicBlockId,
     ) -> Self {
         Self {
+            skip_ref_counts: function.runtime().is_acir(),
```
vezenovm, line 764:
```suggestion
```
referencing above comment

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -529,166 +971,402 @@ mod tests {
...
-        builder.switch_to_block(b2);
-        let one = builder.field_constant(1u128);
-        let v3 = builder.insert_binary(v1, BinaryOp::Add, one);
-        builder.terminate_with_jmp(b1, vec![v3]);
+    /// Test that we can unroll a small loop.
+    #[test]
+    fn test_brillig_unroll_small_loop() {
+        let ssa = brillig_unroll_test_case();
+
+        // Example taken from an equivalent ACIR program (ie. remove the `unconstrained`) and run
```
vezenovm, line 1132: Confused by this comment. Do you mean to point the reader to a specific program? Also they should remove the `unconstrained` even though main is a brillig runtime?

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -529,166 +971,402 @@ mod tests {
...
+
+    /// Test that with more iterations it's not unrolled.
+    #[test]
+    fn test_brillig_unroll_6470_large() {
+        // More iterations than it can unroll
+        let ssa = brillig_unroll_test_case_6470(6);
+        let stats = loop0_stats(&ssa);
+        assert!(!stats.is_small(), "the loop should be considered large");
+
+        let (ssa, errors) = try_unroll_loops(ssa);
```
vezenovm, line 1216: Do we expect the SSA to be the same? Can we just do a deeper equality check against the SSA output itself?

vezenovm: Same for the other failure cases?

vezenovm (commented):
Mostly nits, but overall looks good. The additional comments are very helpful!

TomAFrench:
I'm a little concerned by the regressions in brillig keccak size. Have you looked at how we can avoid this?

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
aakoshh: No these are different, I'll fix the comment to be clearer. I want the allocations that happen _before_ the loop, so anything between the function entry point and the pre-header of the loop. `self.blocks` is the loop itself, between the actual header and the back edge. By the comment `// The algorithm to collect blocks in the loop is fine for us.` I wanted to say that even though `find_blocks_in_loop` says in its docs that it expects the loop header and the back edge, the algorithm itself works just as fine if I ask it to traverse from the pre-header until the function entry block. 

I suppose I could move that that to a utility method to make it clearer and call that, rather than this hack.

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
aakoshh: True, I thought about it as well. On the other hand I thought that 1) having two dedicated methods makes it easier to understand and test, and 2) `count_all_instructions` doesn't need lookups, just the length of the ID vector. Forcing them into a single loop function felt artificial at the time, but at the same time it's unlikely to be used for anything else, so I suppose I could consolidate them into the stats method.

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
aakoshh: You're right, it's probably not needed, I'll remove it (I added it to the method docstrings anyway). At the time it wasn't clear to me under which condition the blocks will have multiple destinations because I was only thinking about the loops, and wanted to capture that other kinds of blocks in the loop can still occur.

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
vezenovm: Ah I see. I had skimmed over the actual arguments to the function. A clearer comment should be fine!

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
aakoshh: I think "of" was the word, but thanks for flagging because this was a very hand wavy comment about a mechanism I least understood; perhaps you can provide a better explanation of what is the intent here.

I was trying to describe what the `next_blocks.retain(|b| self.loop_.blocks.contains(self.get_original_block(b)))` means. I looked up how things make it into `self.original_block` in `get_or_insert_block`, and it's only called from here:
```
terminator.mutate_blocks(|block| self.get_or_insert_block(block));
```
which is called by `inline_instructions_from_block`, which we do for the loop header and then its body. 

So "one of the block we have seen as a destination of a terminator of a block we have already inlined." is technically what I see happening in the code, but I'm not sure why these are the ones retained and what are the ones we're getting rid of exactly.

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
aakoshh: > with the `inserter` ... we can make a method inside of the `LoopIteration` context.

To be clear, this is a field added to the `LoopIteration`, as you say so that we don't have to call `.runtime().is_acir()`, but if calling that is not a big overhead I'm happy to remove this 👍

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
aakoshh: The program is in the comments of the [brillig_unroll_test_case](https://github.com/noir-lang/noir/blob/176d0fca2a4e8c58a8ef2eb257e1834b6097dbf6/compiler/noirc_evaluator/src/ssa/opt/unrolling.rs#L1293) below. I just wanted to explain where this expected SSA came from: by treating the same code as ACIR and capturing the output, and to trigger the ACIR workflow I removed `unconstrained` from `main`. I'll rephrase it.

The nargo command is probably not needed for anyone used to working with these though, I'll remove it.

aakoshh:
It is strange, I commented out most of the code `keccak256/src/main.nr`, only left the first assertions, which tests a single byte. That means the loops it's unrolling is only a single iteration, yet `nargo info` shows it goes from 591 opcodes to 1291. Unrolling a single iteration will always be seen as worthwhile as it gets rid of at least the constant overhead of jump instructions. Comparing the SSA didn't reveal anything obvious, now looking at the brillig opcodes.

aakoshh:
In the brillig code in the unrolled version there is a huge section like this:
```
Function(Id(0), Some(Id(0))):
  CAST S5, S1 as u8
  CAST S4, S5 as u254
  CAST S1, S4 as u8
  MOV S4, FreeMem
  CONST S5 = 137
  FreeMem = FreeMem + S5
  ICONST S4 = 1
  S5 = S4 + M2
  CONST S6 = 0
  CONST S7 = 136
  S7 = S7 + S5
  MOV S8, S5
  S9 = S8 < S7
  i1_NOT S9 = !S9
  JUMP_IF S9 TO Function(Id(0), Some(Id(0))) - 2
  STORE *S8 = S6
  S8 = S8 + M2
  JUMP_TO Function(Id(0), Some(Id(0))) - 1
  LOAD S5 = *S4
  S5 = S5 + M2
  STORE *S4 = S5
  MOV S5, FreeMem
  FreeMem = FreeMem + M2
  STORE *S5 = S4
  MOV S4, FreeMem
  CONST S7 = 137
  FreeMem = FreeMem + S7
  ICONST S4 = 1
  S7 = S4 + M2
  MOV S8, S7
  STORE *S8 = S1
  S8 = S8 + M2
  STORE *S8 = S6
  S8 = S8 + M2
  STORE *S8 = S6
  S8 = S8 + M2
  STORE *S8 = S6
... repeat until line 870
```
Looking at the final SSA the first block is different. 

Without unrolling:
```
b0(v0: Field, v1: [u8; 32]):
    v11 = truncate v0 to 8 bits, max_bit_size: 254
    v12 = cast v11 as u8
    inc_rc [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] of u8
    v15 = allocate -> &mut [u8; 136]
    store [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] of u8 at v15
    jmp b1(u32 0)
```

With unrolling:
```
b0(v0: Field, v1: [u8; 32]):
    v8 = truncate v0 to 8 bits, max_bit_size: 254
    v9 = cast v8 as u8
    inc_rc [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] of u8
    v12 = allocate -> &mut [u8; 136]
    store [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] of u8 at v12
    store [v9, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] of u8 at v12
    store [v9, u8 1, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] of u8 at v12
    store [v9, u8 1, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 128] of u8 at v12
    inc_rc [u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0] of u64
    v20 = allocate -> &mut [u64; 17]
    store [u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0] of u64 at v20
    jmp b1(u32 0)
```

Not sure what these seemingly redundant stores to `v12` are.

vezenovm:
> Not sure what these seemingly redundant stores to `v12` are.

I'm surprised that mem2reg isn't simplifying these into a single store.

vezenovm:
> I'm surprised that mem2reg isn't simplifying these into a single store.

Oh I see this functionality was removed for correctness (https://github.com/noir-lang/noir/blob/c11a6963a0f79101848e7a6acac36eea74fef5a2/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs#L435)

aakoshh:
The multiple stores appear after `mem2reg`. 

With unrolling it is preceded by this:
```
After Removing Bit Shifts:
brillig(inline) fn main f0 {
  b0(v0: Field, v1: [u8; 32]):
    v8 = truncate v0 to 8 bits, max_bit_size: 254
    v9 = cast v8 as u8
    inc_rc [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] of u8
    v12 = allocate -> &mut [u8; 136]
    store [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] of u8 at v12
    v14 = call is_unconstrained() -> u1
    v15 = load v12 -> [u8; 136]
    v17 = array_set v15, index u32 0, value v9
    store v17 at v12
    v18 = load v12 -> [u8; 136]
    v21 = array_set v18, index u32 1, value u8 1
    store v21 at v12
    v24 = array_set v21, index u32 135, value u8 128
    store v24 at v12
    inc_rc [u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0] of u64
    v27 = allocate -> &mut [u64; 17]
    store [u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0, u64 0] of u64 at v27
    jmp b1(u32 0)
```

Without unrolling it's shorter:
```
After Removing Bit Shifts:
brillig(inline) fn main f0 {
  b0(v0: Field, v1: [u8; 32]):
    v11 = truncate v0 to 8 bits, max_bit_size: 254
    v12 = cast v11 as u8
    inc_rc [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] of u8
    v15 = allocate -> &mut [u8; 136]
    store [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] of u8 at v15
    v17 = call is_unconstrained() -> u1
    jmp b1(u32 0)
```

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
vezenovm: Ok, I think we can rewrite the comment. I think referencing the `JmpIf` here is not entirely clear and if we want to reference what the `JmpIf` does we should do it as a comment on `handle_jmpif `. We just want to make sure that the blocks which we set up next for being unrolled are actually part of the the original loop we are unrolling. e.g. if we have a `jmp_if cond then_block else_block` and the `cond` is not a numeric constant, `unroll_loop_block_helper` will return `then_block` and `else_block`. This retain just makes sure that those blocks are actually part of the loop we are unrolling.

aakoshh:
Here's the complete example without unrolling:
```
brillig(inline) fn main f0 {
  b0(v0: Field, v1: [u8; 32]):
    v11 = truncate v0 to 8 bits, max_bit_size: 254
    v12 = cast v11 as u8
    inc_rc [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] of u8
    v15 = allocate -> &mut [u8; 136]
    // Empty array at v15
    store [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] of u8 at v15
    v17 = call is_unconstrained() -> u1
    jmp b1(u32 0)                                 // loop from 0
  b1(v2: u32):                                    // Single iteration loop header
    v19 = eq v2, u32 0                            // loop to 0, ie. 0..len where len==1
    jmpif v19 then: b30, else: b2
  b30():                                          // loop body
    v119 = load v15 -> [u8; 136]                  // load the empty array
    v121 = array_get [v12] of u8, index v2 -> u8  // get the n-th byte from the input, in our case just the single byte
    v122 = array_set v119, index v2, value v121   // set the input byte in the empty array
    v123 = add v2, u32 1                          
    store v122 at v15                             // store the modified array
    v124 = add v2, u32 1                          // increment induction variable
    jmp b1(v124)                                  // back edge
```

This `b1-to-b30` loop gets unrolled and then instead of working with array references it becomes full array constants.

jfecher:
> Do you think some a strategy like this would make sense to avoid increasing the byte code at all costs?

I don't think it's a goal to avoid increasing byte code size at all costs but I see two directions we can go:
- Use heuristics to determine whether we should unroll
- Use this exact check plus maybe an optimization level multiplier to determine whether we should unroll.
  - Unrolling will always lead to fewer instructions executed since we guarantee we eliminate the increment & jump instructions, but may lead to a larger size. So this multiplier would be to adjust the balance between execution & size we want. E.g. we may say "unrolling is fine, up until it is 2x the size of the original loop, then we just want to keep the original loop"

TomAFrench:
I think we can have a follow-up PR in order to set the exact trade-off. Looking at the benchmarks we're only regressing in keccak256 atm by 25% which is higher than I'd like but manageable.

TomAFrench (approved):

Issue #6479, opened 2024-11-07
Errors when importing traits


nventuro:
### Aim

I noticed some incorrect 'unused import' warnings, and after looking into it some more found some strange incorrect behavior when importing traits. It may extend to other imports.

### Bug

The following program seems correct:

```rust
mod foo {
    pub trait Foo<let N: u32> {
    }
}

use foo::Foo;

pub trait Bar<T, let N: u32> where T: Foo<N> {}

fn main() {
}
```

but it compiles with a seemingly incorrect warning:
```
warning: unused import Foo
  ┌─ src/main.nr:6:10
  │
6 │ use foo::Foo;
  │          --- unused import
```

since `Foo` is clearly used in `Bar`'s declaration.

Removing the `use::foo:Foo` line does not cause a compilation error, which I also found odd.

---

Curiously, adding a second trait causes for the warning to go away:
```rust
mod foo {
    pub trait Bar<let N: u32> {
    }
}

use foo::Bar;

trait Qux {}

pub trait Baz<T, let N: u32> where T: Bar<N> {}

fn main() {
}
```

but this depends on the order of the declarations - the following program does emit the original warning:

```rust
mod foo {
    pub trait Bar<let N: u32> {
    }
}

use foo::Bar;

pub trait Baz<T, let N: u32> where T: Bar<N> {}

trait Qux {}

fn main() {
}
```

---

Finally, placing the new trait inside a new module causes a compilation error instead:

```rust
mod foo {
    pub trait Bar<let N: u32> {
    }
}

use foo::Bar;

pub trait Baz<T, let N: u32> where T: Bar<N> {}

mod qux { 
    trait Qux {}
}

fn main() {
}
```

```
warning: unused import Bar
  ┌─ src/main.nr:6:10
  │
6 │ use foo::Bar;
  │          --- unused import
  │

error: Could not resolve 'Bar' in path
  ┌─ src/main.nr:8:39
  │
8 │ pub trait Baz<T, let N: u32> where T: Bar<N> {}
  │                                       ---
  │
```

PR #6506, opened 2024-11-12
fix: set local_module before elaborating each trait

asterite:
# Description

## Problem

Resolves https://github.com/noir-lang/noir/issues/6479

## Summary

This was actually more than just a visibility bug: it could happen that we just elaborated a trait in a nested module, and `local_module` ended up in that nested module; then finding types for where clauses would find them even if they were normally unreachable. For example this code compiled fine:

```noir
mod foo {
    pub trait Foo {}
}

pub trait Bar<T>
where
    T: Foo, // Here Foo was incorrectly found
{}
```

## Additional Context

I wonder if directly setting `local_module` should be forbidden, and instead we should always use `replace_module`.

## Documentation

Check one:

# PR Checklist


jfecher (approved):
Good catch. I wonder if we're missing this for the other collect cases as well?

Also, should we set the current file or anything else as well? I suppose it is the call to `resolve_trait_constraints` that is why we require the local module here. Makes me wonder if we have any missed assumptions elsewhere.

PR #6507, opened 2024-11-12
fix: parse Slice type in SSa

asterite:
# Description

## Problem

I wanted to add a test for an existing optimization that is untested. While doing so I noticed the SSA parser can't parse slice types and also that some errors were incorrect (there's one `{token}` occurrence that should have been `{found}`).

## Summary

I want to very slowly add tests for existing optimizations that don't have tests. This is just for one of them, but also includes a bunch of fixes.

## Documentation

Check one:

# PR Checklist


aakoshh (approved):
👌

PR #6509, opened 2024-11-13
feat: avoid unnecessary ssa passes while loop unrolling

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

I noticed that when we're unrolling loops we run 2 `mem2reg` passes and a `simplify_cfg` on all functions in the program after every failure. This means that we'll end up doing a lot of redundant work on all the other functions in the program as long as there's a single function which is still having loops unrolled.

This PR changes how we do loop unrolling so that we unroll a single function to completion before we start unrolling the next. This also opens up the opportunity to parallelize this work in future. 

## Documentation\*

Check one:

# PR Checklist\*


jfecher (approved):
Good catch

PR #6511, opened 2024-11-13
chore: Reverse ssa parser diff order

jfecher:
# Description

## Summary\*

The parser diff appears reversed when an SSA test is incorrect so I swapped the order so that it is more natural.

## Additional Context

Before it resembled:
```
- actual line
+ expected line
```
This reverses it to be more like a git diff:
```
- expected line
+ actual line
```

## Documentation\*

Check one:

# PR Checklist\*


aakoshh:
I actually had the opposite remark in the original PR: https://github.com/noir-lang/noir/pull/6489#discussion_r1838211232

My problem was that it starts like this:
```
Differences (-left|+right):
```
which was opposite of how the arguments were passed to the actual function we call from the tests.

TomAFrench (approved):

Issue #6512, opened 2024-11-13
Remove duplicate execution tests


jfecher:
### Aim

We have a few duplicate execution tests, e.g. `slices` and `brillig_slices`, `hash_to_field`, and `brillig_hash_to_field`. Now that we have `--force-brillig`, the brillig variants are unnecessary.

### Expected Behavior

We shouldn't have duplicate tests

### Bug

We have duplicate tests

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #6523, opened 2024-11-15
chore(test): Remove duplicate brillig tests

aakoshh:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6512

## Summary\*

Removed duplicate `test_programs/execution_success/brillig_<test>` from where an equivalent `test_programs/execution_success/<test>` exists, listed by:

```shell
for dir in (ls | grep -v brillig_); if [ -d "brillig_$dir" ]; echo duplicate $dir; end; end;
```

The `brillig_` version of following has been removed from `execution_success`:
* `blake3` -> duplicate; unconstrained main
* `references` -> duplicate; unconstrained main
* `slices` -> duplicate; unconstrained main
* `assert` -> ACIR calls unconstrained; moved conditional
* `hash_to_field` -> ACIR to unconstrained, otherwise identical
* `loop` -> ACIR to unconstrained, otherwise identical
* `array_to_slice` -> not a duplicate, but seems redundant
* `ecdsa_secp256k1` -> ACIR to unconstrained, but very similar
* `ecdsa_secp256r1` -> ACIR to unconstrained, but very similar 
* `sha256`  -> ACIR to unconstrained, subset

## Additional Context

I wasn't sure if the ones where ACIR calls `unconstrained` are considered duplicates, but `hash_to_field` was such an example and it's listed in the ticket, so I removed all of them.

Note that there is `execution_failure/brillig_assert_msg_runtime` which seems like it could be removed in favour of `execution_failure/assert_msg_runtime`, however currently these tests are only executed with the [default settings](https://github.com/noir-lang/noir/blob/852c87ae9ecdd441ee4c2ab3e78e86b2da07d8a4/tooling/nargo_cli/build.rs#L274) which means the `--force-brillig` option isn't applied to them; if we want to test brillig behaviour we need to either run it both ways or have two tests.

## Documentation\*

Check one:

# PR Checklist\*


jfecher (approved):
LGTM

Issue #6471, opened 2024-11-07
Add equivalent to Rust's `std::hint::blackbox`


TomAFrench:
> In the short term (and complementary to debug mode in the future), we could add something similar to Rust's `blackbox` function. This would just be an identity function which is listed as having side-effects, which would block any SSA optimizations from being applied to the outputs. 

 _Originally posted by @TomAFrench in [#2128](https://github.com/noir-lang/noir/issues/2128#issuecomment-1696134105)_

One tool that I've found myself needing a few times when minimising bug reproductions is a way to block our SSA optimization passes being applied to a value. This would also allow us to prevent test programs from being optimized down to nothing and have them resemble runtime code more closely.

PR #6529, opened 2024-11-15
feat: `std::hint::black_box` function.

aakoshh:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6471

## Summary\*

Added a `std::hint` Noir module with a `black_box` builtin function that is treated during SSA optimisation steps as if it could potentially have side effects, which prevents the value it is applied to from being simplified to a known value and the instruction from being removed. 

## Additional Context

Using a `builtin` rather than `foreign` so that this can be called from non-Brillig code. 

I added an `Intrinsic::Hint` because `Intrinsic::BlackBox` was already taken; perhaps `Intrinsic::AsWitness` could be moved into `Hint` as well (not that it matters, but the comments call it a hint too).

## Documentation\*

Check one:

# PR Checklist\*


test_programs/execution_success/hint_black_box/src/main.nr
TomAFrench, line 1: Can you add some tests to check how it handles arrays, etc.?

aakoshh: Good shout, the slice tests uncovered a bug.

aakoshh: Added some tests with mutable references but that doesn't work very well. I left some comments about scenarios that didn't work out, let me know if you think there's anything we can do about them.

TomAFrench:
I'd like to take another look at this today after scrum

aakoshh:
@TomAFrench have you had a chance to look at this?

TomAFrench (approved):

TomAFrench:
Thanks, sorry for the delay on this one.

Savio-Sou:
Looks like this is not currently documented in the Noir docs.

Is the new function potentially useful for the community?

PR #6534, opened 2024-11-15
chore: restructure `noirc_evaluator` crate

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

It's been bugging me how `acir_gen` sits inside of `ssa` whereas `brillig` sits at the same level, this PR pulls the `acir_gen` module to line up with `brillig`. I've also tidied up some of the imports.

## Documentation\*

Check one:

# PR Checklist\*


aakoshh (commented):
Don't you want to go all the way and have `acir/acir_ir` and `acir/acir_gen` similar to how `brillig` is structured?

TomAFrench:
Maybe but I think that would end up being a larger refactor than a simple moving of a folder from one place to another. I don't think we should do it in this PR.

PR #6538, opened 2024-11-17
feat(profiler): Reduce memory in Brillig execution flamegraph

vezenovm:
# Description

## Problem\*

No issue as these are local changes I had to enable executing larger flamegraphs

## Summary\*

Execution of the `blob` circuit when force compiled to Brillig produces 1.1 billion samples. The execution flamegraph process gets killed with an OOM error when running it on the mainframe which has 740G of RAM.  The process would get killed when transforming the profiling samples returned by `execute_program_with_profiling` into the samples that the flamegraph generator expects.

To reduce memory I did a couple things:
1. We format our opcode directly. The `AcirOrBrilligOpcode<F>` enum was over 300 bytes. The opcode strings are mostly less than 20 characters.
2. Use `std::mem::take` when iterating over the profiling samples. This combined with step (1) got us to stop killing the profiling process. We still would use over 240 gigs. 
3. Make a new `Sample` trait with a common interface. Then have different types of samples so that we are not storing unnecessary data. For example, in an execution sample the `count` will always be one. There is no need to store a count field in the sample. When you get to 1 billion samples this extra field starts to make a difference. This looks to have saved us over 20 gigs.

## Additional Context

We can probably get rid of `brillig_function_id` as well by mapping the function ID to its samples. 

## Documentation\*

Check one:

# PR Checklist\*


tooling/profiler/src/flamegraph.rs
```diff
@@ -82,25 +130,26 @@ impl FlamegraphGenerator for InfernoFlamegraphGenerator {
...
 }
 
-fn generate_folded_sorted_lines<'files, F: AcirField>(
-    samples: Vec<Sample<F>>,
+fn generate_folded_sorted_lines<'files, S: Sample>(
+    samples: Vec<S>,
     debug_symbols: &DebugInfo,
     files: &'files impl Files<'files, FileId = fm::FileId>,
 ) -> Vec<String> {
+    println!("About to generate sorted lines");
```
TomAFrench, line 138:
```suggestion
```

tooling/profiler/src/flamegraph.rs
```diff
@@ -13,18 +12,67 @@ use noirc_errors::reporter::line_and_column_from_span;
...
+    }
+
+    fn opcode(self) -> Option<String> {
+        self.opcode
+    }
+}
+
+pub(crate) struct BrilligExecutionSample {
+    pub(crate) opcode: Option<String>,
+    pub(crate) call_stack: Vec<OpcodeLocation>,
```
TomAFrench, line 53: I think interning these callstacks should save us a good deal tbh.

tooling/profiler/src/cli/execution_flamegraph_cmd.rs
```diff
@@ -51,19 +50,21 @@ fn run_with_generator(
...
-        .into_iter()
+    println!("Collecting {} samples", profiling_samples.len());
+
+    let profiling_samples: Vec<BrilligExecutionSample> = profiling_samples
+        .iter_mut()
         .map(|sample| {
-            let call_stack = sample.call_stack;
-            let brillig_function_id = sample.brillig_function_id;
+            let call_stack = std::mem::take(&mut sample.call_stack);
+            let brillig_function_id = std::mem::take(&mut sample.brillig_function_id);
```
TomAFrench, line 67: I'm not clear on why this is preferable over just taking ownership over the samples. Do you know why?

tooling/profiler/src/flamegraph.rs
TomAFrench: Noticed we actually have a decent number of print statements in here so maybe this is on purpose.

tooling/profiler/src/flamegraph.rs
vezenovm: Yeah I had this as a TODO and was experimenting with it, but got blocked testing it against blob with the new frontend stack overflow. For `sha256_regression` the cache I made had about 97k call stacks vs. 126k samples (all with their own call stacks). I didn't include it in this PR as I want to see the improvement against a larger case where it matters more.

tooling/profiler/src/flamegraph.rs
vezenovm: The other ones are on purpose, but this one is leftover from debugging.

tooling/profiler/src/flamegraph.rs
TomAFrench: Yeah, happy to have as a to-do. Can you make an issue for this?

tooling/profiler/src/cli/execution_flamegraph_cmd.rs
vezenovm: I had noticed we were hitting OOM errors during collection of our samples. It was difficult to confirm exactly but I believe `collect` is still allocating memory for `Vec<Sample>` and during collection our memory kept growing after execution. `take` also takes ownership, just it lets us to do it of part of our Vec, so I wanted to guarantee we were taking ownership of the sample rather than reallocating memory. This is not the same issue, but I had stumbled upon some articles that indicated issues with Rust's collect (https://blog.polybdenum.com/2024/01/17/identifying-the-collect-vec-memory-leak-footgun.html) as well. This led me to try indicating I want ownership of this sample data directly.

tooling/profiler/src/flamegraph.rs
```diff
@@ -13,18 +12,67 @@ use noirc_errors::reporter::line_and_column_from_span;
...
+    pub(crate) brillig_function_id: Option<BrilligFunctionId>,
+}
+
+impl Sample for BrilligExecutionSample {
+    fn count(&self) -> usize {
+        1
+    }
+
+    fn brillig_function_id(&self) -> Option<BrilligFunctionId> {
+        // Some(BrilligFunctionId(0))
```
vezenovm, line 63:
```suggestion
```

tooling/profiler/src/flamegraph.rs
vezenovm: https://github.com/noir-lang/noir/issues/6544

TomAFrench (approved):

PR #6540, opened 2024-11-18
fix: iterative unification

### LINKED ISSUE #6457: Thousands of hard-coded constants slows down execution time by 7x.
2 potential bugs here.

Checkout this branch: https://github.com/iAmMichaelConnor/blob-lib/pull/5

Update the Nargo.toml to point to v0.4.0 (or later... I was testing it against `main`, locally).

Turn your attention to the `config.nr` and `negative_roots.nr` files. You'll see thousands of hard-coded constants.

Now observe:

Run `nargo execute` on `main.nr`. This took 28 minutes on my old 2019 MacBook Pro i9.

Now comment-out all those constants and uncomment the "array of zeros" version. Do it in both `config.nr` and `negative_roots.nr`.

E.g. see this pic for what to uncomment:

Image

Again, run `nargo execute` on `main.nr`. It'll break with some kind of "stack too deep" error. Very strange.

Now delete those thousands of lines of commented-out hard-coded constants.

Again, run `nargo execute` on `main.nr`. The "stack too deep" error will disappear. This took 3 mins 50 seconds. 

Observations:
- Those thousands of hard-coded constants slow-down execution by 7x.
- Huge comments break Noir.

Thanks to @MirandaWood for first spotting this!

guipublic:
# Description

## Problem\*

Resolves #6457

## Summary\*
Makes iterative version of:
- type unification
- type conversion (hir to ast)
- type check (in monomorphization)

## Additional Context
In the issue, the problem was due to unification during elaboration and check type during monomorphisation. So type conversion was not failing.
Convert and check type are made iterative only for type variables because the issue is happening with them.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/databus.rs
```diff
@@ -0,0 +1,46 @@
...
+        for func in self.functions.values() {
+            let mut reverse_post_order = PostOrder::with_function(func).into_vec();
+            reverse_post_order.reverse();
+            for block in reverse_post_order {
+                last_use(block, &func.dfg, &mut array_use);
+            }
+        }
+        array_use
+    }
+//la question c comment on remplace  ?
```
vezenovm, line 20: Some leftover french here

compiler/noirc_evaluator/src/ssa/opt/databus.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub(crate) fn array_me(block_id: BasicBlockId, dfg: &mut DataFlowGraph) {
+        let block = &dfg[block_id];
+        for instruction_id in block.instructions() {
+            match &dfg[*instruction_id] {
+                Instruction::ArrayGet { array, index } => {
+                                 // Get operations to call-data parameters are replaced by a get to the call-data-bus array
+            if let Some(call_data) = dfg.data_bus.call_data {
+                let array_id = dfg.resolve(*array);
+                if dfg.data_bus.call_data_map.contains_key(&array_id) {
+                    // on doit remplacer ce get par un autre, mais on doit d'abord fiare un calcul
```
vezenovm, line 32: Here as well and below

compiler/noirc_frontend/src/hir_def/iterative_unification.rs
```diff
@@ -0,0 +1,339 @@
...
+                    for (a, b) in elements_a.iter().zip(elements_b) {
+                        to_unit.push(self.for_unite(a, b));
+                    }
+                    Ok(to_unit)
+                }
+            }
+
+            // No recursive try_unify call for struct fields. Don't want
+            // to mutate shared type variables within struct definitions.
+            // This isn't possible currently but will be once noir gets generic types
```
vezenovm, line 188: This looks like an outdated comment cc @jfecher

compiler/noirc_frontend/src/monomorphization/mod.rs
```diff
@@ -957,6 +958,59 @@ impl<'interner> Monomorphizer<'interner> {
         Ok(ident)
     }
 
+    /// Iterative version of convert_type, for a subset of types
+    /// It defaults to the recursive version for the other types
```
vezenovm, line 962: Can we add a little more info in the comment as to why this subset was chosen for iterative unification vs. recursive unification being used for the rest of the types?

vezenovm (commented):
Looks good to me, seems like we have one last failing test with the inliner though. Wondering if we should just move to iterative unification for all the types? Perhaps this is better for a followup though.

compiler/noirc_evaluator/src/ssa/opt/databus.rs
guipublic: that's an old file that was left-over, I removed it.

compiler/noirc_evaluator/src/ssa/opt/databus.rs
guipublic: file has been removed

vezenovm:
It looks good to me, we just need to figure out that last arithmetic generics failure.

guipublic:
> It looks good to me, we just need to figure out that last arithmetic generics failure.

It was due to the changes for constant terms during unification. I tried to fix this but it was still failing so I end up using the previous recursive code for this case.

TomAFrench:
Is there a good testcase we can add that would break the old type unification code?

compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
```diff
@@ -340,7 +341,8 @@ impl Type {
                     let new_other = Type::InfixExpr(Box::new(other.clone()), inverse, rhs_a);
 
                     let mut tmp_bindings = bindings.clone();
-                    if lhs_a.try_unify(&new_other, &mut tmp_bindings).is_ok() {
+                    if Unifier::try_unify(lhs_a, &new_other, &mut tmp_bindings).is_ok() {
+                        //if lhs_a.try_unify(&new_other, &mut tmp_bindings).is_ok() {
```
TomAFrench, line 345:
```suggestion
```

compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
```diff
@@ -357,7 +359,8 @@ impl Type {
                     let new_self = Type::InfixExpr(Box::new(self.clone()), inverse, rhs_b);
 
                     let mut tmp_bindings = bindings.clone();
-                    if new_self.try_unify(lhs_b, &mut tmp_bindings).is_ok() {
+                    if Unifier::try_unify(&new_self, lhs_b, &mut tmp_bindings).is_ok() {
+                        // if new_self.try_unify(lhs_b, &mut tmp_bindings).is_ok() {
```
TomAFrench, line 363:
```suggestion
```

jfecher:
> https://github.com/noir-lang/noir/issues/6457

How long is this expected to take now? I ran it for 2:12 before halting it because it was taking too long - it didn't finish monomorphization by that point. Comparing to master, master lasts 1:12 for me before stack overflowing

jfecher:
This should be unneeded after https://github.com/noir-lang/noir/pull/6558 - we should be avoiding building up these large type variable chains to begin with now. Sorry for not mentioning this sooner - #6558 was something I just found today while reviewing this PR

compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
guipublic: Done

compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
guipublic: Done

compiler/noirc_frontend/src/hir_def/iterative_unification.rs
guipublic: Done

guipublic:
> This should be unneeded after #6558 - we should be avoiding building up these large type variable chains to begin with now. Sorry for not mentioning this sooner - #6558 was something I just found today while reviewing this PR and trying out a follow_bindings call before unification

I agree it is better to avoir the large binding chain in the first place, but having iterative version is still an improvement over the recursive one.

guipublic:
At least we can see that the additional cloning does not impact the peak memory.

TomAFrench:
I'm inclined to close this if we're making this step harder to follow and adding more cloning.

PR #6547, opened 2024-11-18
feat: simplify constant MSM calls in SSA

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR simplifies MSM calls in SSA if all the inputs are constant. This is necessary for https://github.com/AztecProtocol/aztec-packages/pull/9897

## Documentation\*

Check one:
- [x] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #6548, opened 2024-11-18
feat: try to inline brillig calls with all constant arguments 

### LINKED ISSUE #6453: Inline results from brillig function calls for which all arguments are known during SSA

asterite:
# Description

## Problem

Resolves #6453

## Summary

Alternative to #6466

For now this PR just finds the brillig calls with all constant values, creates a VM for each, executes them, and if we can get some output values, the call results are replaced with those values.

Pending:
- [ ] If we managed to replace some calls, do all SSA optimization passes again as the new values could help further optimize SSA, then check if we can find calls with constant values, and repeat until we can't optimize anymore.

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
```diff
@@ -0,0 +1,463 @@
...
+        self
+    }
+}
+
+/// Result of trying to evaluate an instruction (any instruction) in this pass.
+enum EvaluationResult {
+    /// Nothing was done because the instruction wasn't a call to a brillig function,
+    /// or some arguments to it were not constants.
+    NotABrilligCall,
+    /// The instruction was a call to a brillig function, but we couldn't evaluate it.
```
TomAFrench, line 80:
```suggestion
    /// The instruction was a call to a brillig function, but we couldn't evaluate it.
    /// This can occur in the situation where the brillig function reaches a "trap" or a foreign call opcode.
```

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
```diff
@@ -0,0 +1,463 @@
...
+        };
+
+        let mut calldata = Vec::new();
+        for argument in arguments {
+            value_id_to_calldata(*argument, &self.dfg, &mut calldata);
+        }
+
+        let bytecode = &generated_brillig.byte_code;
+        let foreign_call_results = Vec::new();
+        let black_box_solver = StubbedBlackBoxSolver;
```
TomAFrench, line 250:
```suggestion
        let black_box_solver = bn254_blackbox_solver::Bn254BlackBoxSolver;
```

We can use a full solver here.

TomAFrench (approved):
LGTM, it would also be good to cache the linked bytecode to reuse it in ACIR gen but that can be done in a separate PR.

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_fn.rs
```diff
@@ -68,6 +68,25 @@ impl FunctionContext {
...
+        match typ {
+            Type::Numeric(_) | Type::Reference(_) => {
+                Some(BrilligParameter::SingleAddr(get_bit_size_from_ssa_type(typ)))
+            }
+            Type::Array(item_type, size) => {
+                let mut parameters = Vec::new();
+                for item_typ in item_type.iter() {
+                    let Some(param) = FunctionContext::try_ssa_type_to_parameter(item_typ) else {
+                        return None;
+                    };
```
jfecher, line 81:
```suggestion
                    let param = FunctionContext::try_ssa_type_to_parameter(item_typ)?;
```

compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -141,6 +141,8 @@ pub(crate) fn optimize_into_acir(
         ssa.to_brillig(options.enable_brillig_logging)
     });
 
+    let ssa = ssa.inline_const_brillig_calls(&brillig);
```
jfecher, line 144: Shouldn't we do this earlier? The original issue mentioned doing this during SSA to take advantage of SSA optimizations but if we only do it at the very end of SSA there are no further optimizations.

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
```diff
@@ -0,0 +1,463 @@
...
+            }
+
+            if brillig_functions_we_could_not_inline.contains(func_id) {
+                continue;
+            }
+
+            // We also don't want to remove entry points
+            if self.entry_point_to_generated_index.contains_key(func_id) {
+                continue;
+            }
```
jfecher, line 66: Style nitpick but I'd combine these cases with an `||` and keep the comments

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
```diff
@@ -0,0 +1,463 @@
...
+        for block_id in self.reachable_blocks() {
+            for instruction_id in self.dfg[block_id].take_instructions() {
+                let evaluation_result =
+                    self.evaluate_const_brillig_call(instruction_id, brillig, brillig_functions);
+                match evaluation_result {
+                    EvaluationResult::NotABrilligCall => {
+                        self.dfg[block_id].instructions_mut().push(instruction_id);
+                    }
+                    EvaluationResult::CannotEvaluate(func_id) => {
+                        self.dfg[block_id].instructions_mut().push(instruction_id);
```
jfecher, line 103:
```suggestion
                        self.dfg[block_id].insert_instruction(instruction_id);
                    }
                    EvaluationResult::CannotEvaluate(func_id) => {
                        self.dfg[block_id].insert_instruction(instruction_id);
```

jfecher (changes requested):
I don't think this should be an SSA pass - maybe it should be included in `Instruction::simplify` instead, then we could have a pass just to remove unused functions. The advantage of that being that we can more easily take advantage of SSA optimizations as soon as the call becomes constant. Disadvantage being we'd be checking if it is constant on every pass.

TomAFrench:
I think putting it in instruction simplification may be a bit OTT for now, maybe within the constant folding pass would be a good middle step.

The reason why I don't think it should be in instruction simplification is that we want to only do codegen once. With this we'd need to pass in this big structure of all of the functions whenever we push in any instruction even if there isn't a function call.

compiler/noirc_evaluator/src/ssa.rs
TomAFrench: We need to do this relatively late in the compilation process as the unconstrained bytecode used is ideally the final bytecode.

jfecher: Not sure I understand - wouldn't doing it earlier result in more SSA optimizations? Even if doing it later would optimize the brillig part more so this pass has less to execute, we'd still prefer the SSA optimizations over that I'd think. Or am I missing something?

compiler/noirc_evaluator/src/ssa/opt/inline_const_brillig_calls.rs
```diff
@@ -0,0 +1,463 @@
...
+    NotABrilligCall,
+    /// The instruction was a call to a brillig function, but we couldn't evaluate it.
+    CannotEvaluate(FunctionId),
+    /// The instruction was a call to a brillig function and we were able to evaluate it,
+    /// returning evaluation memory values.
+    Evaluated(Vec<MemoryValue<FieldElement>>),
+}
+
+impl Function {
+    pub(crate) fn inline_const_brillig_calls(
```
jfecher, line 88:
```suggestion
    fn inline_const_brillig_calls(
```

jfecher (approved):
> With this we'd need to pass in this big structure of all of the functions 

That's true, we'd need to pass in a reference to `&Ssa` somehow even though we need mutable access to a function inside it. That or rework globals away from the cranelift model.

Anyways, I'm fine with this as-is as well. Checking for constant calls on every simplification would be slower and we'd still need an extra pass to remove unused functions anyway.

compiler/noirc_evaluator/src/ssa.rs
TomAFrench: Ok I think I mispoke with "need". It essentially boils down to two options. Comparing applying constant brillig function optimizations early in the compilation pipeline vs later, the pros/cons for doing this optimization early are (I may be missing some)

Pros:
- ACIR functions benefit more from the initial pass of SSA optimisations due to constant input brillig functions not blocking constant folding optimizations. (lesser need for additional passes)

Cons:
 - We end up needing to do another codegen of the brillig functions at the end to get optimized bytecode.
 - Brillig bytecode used for optimizations will end up being less efficient to execute so may slow down compilation.
   
   
The pros/cons of applying this optimization late are basically the reverse of the above.

TomAFrench: If we want 100% optimal bytecode then we need to do some iterations of constant folding and cfg_simplification anyway so I'm relatively relaxed about having some extra passes at the end (which we can easily turn off to improve compilation times).

asterite:
As discussed with Tom on our 1 on 1, I moved the logic to constant folding, then we do another pass of DIE. I still need to continue refactoring a bit because I need to pass a mutable reference in a loop and I'm having some trouble doing that.

asterite:
Sorry, I removed from the queue because contracts in Aztec-Packages break. I'll soon push a fix for this, though I'd like to refactor a bit more to avoid always having to pass `&mut brillig_functions_we_could_not_inline` (ideally I'd pass that as a property of `BrilligInfo` but I'm having a small fight with the borrow checker (understandably because I'm still not super familiar with it)).

asterite:
I'm not sure how to refactor the code to avoid passing `&mut brillig_functions_we_could_not_inline` everywhere.

Alternatively, we could have a pass after this one that removes unused brillig functions. It could do that by starting with a set of all brillig functions, and removing from that set whenever we find a call to one of those functions. It would be slightly slower than doing it in one go, but will surely lead to simpler code. What do you think?

I'm also not sure about doing constant folding plus this new thing in one go, it makes the code a bit more complex... though looking at the optimization results it might be worth it.

That said, I'm running `time nargo compile --silence-warnings --inliner-aggressiveness 0` in `noir-contracts` and it's been running since a few minutes ago... it might be that running all of these functions through the VM slows down things a lot.

TomAFrench:
Interesting... Seems like putting this in the constant folding pass allows us to benefit in brillig functions as well. This makes sense as when we're linking brillig functions together we don't do any constant folding between them so this ends up acting as a simple case of that.

asterite:
I pushed another commit and this is now ready for review.

By the way, the `nargo compile` above took 29 minutes on my machine, on `master`. I'll leave it running on this branch to see if it takes much longer... I think previously `nargo compile` took around 1 minute, not sure if something on our SSA optimizations changed, or it's just that now we are passing `--inliner-aggressiveness 0` in bootstrap.

TomAFrench:
Yeah, nargo compilation in aztec-packages is pretty dire atm. A lot of that time is spent inside of the DIE pass, my aim is to get these two codebases back into one and then we can look into it more deeply.

TomAFrench:
> Alternatively, we could have a pass after this one that removes unused brillig functions. It could do that by starting with a set of all brillig functions, and removing from that set whenever we find a call to one of those functions. It would be slightly slower than doing it in one go, but will surely lead to simpler code. What do you think?
>
>I'm also not sure about doing constant folding plus this new thing in one go, it makes the code a bit more complex... though looking at the optimization results it might be worth it.

I'd say lets do this in iterations. Best to create some new issues for followups than do mega-PR.

asterite:
It seems the slowness comes from `MergeExpressionsOptimizer::eliminate_intermediate_variable`: a lot of BTreeMap clones are happening, or at least that's what I get when I sample the process.

asterite:
It seems the merge with master broke this PR. Specifically this commit it seems: https://github.com/noir-lang/noir/commit/7dd71c15cbcbf025ba049b506c94924903b32754#diff-7c64342380b96ff80bc78d976a826185fbf2f08c0a10d77318dde72f697cb3b7

Something in the blackbox solver changed that's making some tests crash:

```
thread 'run_stdlib_tests::true_i64_min_expects' panicked at /home/runner/work/noir/noir/acvm-repo/brillig_vm/src/black_box.rs:160:62:
called `Result::unwrap()` on an `Err` value: MismatchedBitSize { value_bit_size: 254, expected_bit_size: 1 }
```

TomAFrench:
> It seems the merge with master broke this PR.

This is potentially from my PR here https://github.com/noir-lang/noir/pull/6547/files. We're shoving an` EmbeddedCurvePoint` into an array type so that last bool must be causing a problem.

TomAFrench:
> It seems the slowness comes from `MergeExpressionsOptimizer::eliminate_intermediate_variable`: a lot of BTreeMap clones are happening, or at least that's what I get when I sample the process.

ooooof, yeah that function is chewing through clones like candy. I've made a PR [here](https://github.com/AztecProtocol/aztec-packages/pull/10049) to remove some of those.

asterite:
Nice! With that PR it now takes 2 minutes instead of 29! 🎉

asterite:
> This is potentially from my PR here https://github.com/noir-lang/noir/pull/6547/files. We're shoving an EmbeddedCurvePoint into an array type so that last bool must be causing a problem.

It seems the error happens even without that change. What I'm seeing is that the third argument we get is sometimes a U1, sometimes a Field so it explodes when it's a Field. It gets fixed if we handle both cases, but I'll see if I can find why we are getting different types (maybe you know).

asterite:
It's strange because the signature of the blackbox function (if I'm understanding things correctly) is returning `[Field; 3]`, so I think getting a field as the third argument is good and the U1 is bad, but not sure:

https://github.com/noir-lang/noir/blob/68c32b4ffd9b069fe4b119327dbf4018c17ab9d4/noir_stdlib/src/embedded_curve_ops.nr#L114-L118

asterite:
It seems after the last merge a call to `multi_scalar_mul` now has all constants, when before that merge it didn't, so now it's trying to run that and failing. So maybe the bug was always there, or maybe it's not a bug and it's how I'm setting up the VM and calldata.

asterite:
It seems it might be a bug with inlining. I'm trying this code:

```noir
use std::{embedded_curve_ops::{EmbeddedCurveScalar, multi_scalar_mul}, hash::derive_generators};

fn main(x: Field) {
    let mut points = [EmbeddedCurveScalar { lo: 0, hi: 0 }; 1];
    let generators = derive_generators("DEFAULT_DOMAIN_SEPARATOR".as_bytes(), 0);
    let _ = multi_scalar_mul(generators, points);
}
```

Here's what happens to the SSA at some point:

```
After Resolving IsUnconstrained:
brillig(inline) fn main f0 {
  b0(v0: Field):
    v2 = make_array [Field 0, Field 0] : [(Field, Field); 1]
    inc_rc v2
    v3 = allocate -> &mut [(Field, Field); 1]
    store v2 at v3
    v19 = make_array [u8 68, u8 69, u8 70, u8 65, u8 85, u8 76, u8 84, u8 95, u8 68, u8 79, u8 77, u8 65, u8 73, u8 78, u8 95, u8 83, u8 69, u8 80, u8 65, u8 82, u8 65, u8 84, u8 79, u8 82] : [u8; 24]
    v22 = call f1(v19, u32 0) -> [(Field, Field, u1); 1]
    inc_rc v22
    v23 = load v3 -> [(Field, Field); 1]
    v25, v26, v27 = call f2(v22, v23) -> (Field, Field, u1)
    return
}
brillig(inline_always) fn derive_generators f1 {
  b0(v0: [u8; 24], v1: u32):
    call assert_constant(v0)
    v4 = call derive_pedersen_generators(v0, v1) -> [(Field, Field, u1); 1]
    return v4
}
brillig(inline) fn multi_scalar_mul f2 {
  b0(v0: [(Field, Field, u1); 1], v1: [(Field, Field); 1]):
    v3 = call multi_scalar_mul(v0, v1) -> [Field; 3]
    inc_rc v3
    v5 = array_get v3, index u32 0 -> Field
    v7 = array_get v3, index u32 1 -> Field
    v9 = array_get v3, index u32 2 -> Field
    v10 = truncate v9 to 1 bits, max_bit_size: 254
    v11 = cast v10 as u1
    return v5, v7, v11
}

After Inlining (1st):
brillig(inline) fn main f0 {
  b0(v0: Field):
    v2 = make_array [Field 0, Field 0] : [(Field, Field); 1]
    inc_rc v2
    v3 = allocate -> &mut [(Field, Field); 1]
    store v2 at v3
    v19 = make_array [u8 68, u8 69, u8 70, u8 65, u8 85, u8 76, u8 84, u8 95, u8 68, u8 79, u8 77, u8 65, u8 73, u8 78, u8 95, u8 83, u8 69, u8 80, u8 65, u8 82, u8 65, u8 84, u8 79, u8 82] : [u8; 24]
    v23 = make_array [Field 3728882899078719075161482178784387565366481897740339799480980287259621149274, Field -9903063709032878667290627648209915537972247634463802596148419711785767431332, u1 0] : [Field; 3]
    inc_rc v23
    v24 = load v3 -> [(Field, Field); 1]
    v26, v27, v28 = call f1(v23, v24) -> (Field, Field, u1)
    return
}
brillig(inline) fn multi_scalar_mul f1 {
  b0(v0: [(Field, Field, u1); 1], v1: [(Field, Field); 1]):
    v3 = call multi_scalar_mul(v0, v1) -> [Field; 3]
    inc_rc v3
    v5 = array_get v3, index u32 0 -> Field
    v7 = array_get v3, index u32 1 -> Field
    v9 = array_get v3, index u32 2 -> Field
    v10 = truncate v9 to 1 bits, max_bit_size: 254
    v11 = cast v10 as u1
    return v5, v7, v11
}
```

In particular, this instruction after inlining:

```
    v23 = make_array [Field 3728882899078719075161482178784387565366481897740339799480980287259621149274, Field -9903063709032878667290627648209915537972247634463802596148419711785767431332, u1 0] : [Field; 3]
```

it holds `[Field, Field, u1]` but the type says `[Field; 3]`, and that's where it ends up breaking (it should be `[(Field, Field, u1); 1]`.

Now I'll check the inlining pass to see if I can find something there...

asterite:
Found it!

https://github.com/noir-lang/noir/blame/7658c172a2a9725ff90e12b9a525e3b488c039f6/compiler/noirc_evaluator/src/ssa/ir/instruction/call.rs#L809

I'll send a separate PR to fix this.

asterite:
(ah, the link is to a git blame, but the code was already wrong before that commit... I didn't intend to blame anyone :-))

asterite:
This is again ready to review.

TomAFrench:
LGTM

Issue #6515, opened 2024-11-14
Implement full support for blackbox functions within the comptime intepreter


TomAFrench:
We only support a small number of blackbox functions within the comptime interpreter, see https://github.com/noir-lang/noir/blob/35408ab303f1018c1e2c38e6ea55430a2c89dc4c/compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs#L23-L24

We should aim for as close to full coverage of Noir code being able to run through the interpreter as we can, so we should fill this out with all of the other blackbox functions.

PR #6551, opened 2024-11-19
feat(comptime): Implement blackbox functions in comptime interpreter

aakoshh:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6515

## Summary\*

- [x] Test which fails if any of the blackbox functions is not implemented
- [x] Implement missing functions:
    - [x] `aes128_encrypt`
    - [ ] ~`and`~: Handled as a binary op 
    - [ ] ~`xor`~: Handled as a binary op 
    - [x] `range`
    - [x] `blake2s`
    - [x] `blake3`
    - [ ] ~`schnorr_verify`~: To be removed
    - [x] `ecdsa_secp256k1`
    - [x] `ecdsa_secp256r1`
    - [x] `multi_scalar_mul`
    - [ ] ~`recursive_aggregation`~: Can't do at comptime 
    - [x] `embedded_curve_add`
    - [x] `bigint_add`
    - [x] `bigint_sub`
    - [x] `bigint_mul`
    - [x] `bigint_div`
    - [x] `bigint_from_le_bytes`
    - [x] `bigint_to_le_bytes`
    - [x] `sha256_compression`
- [x] Fix `to_le_radix` return type
- [x] Fix bit shift operations to expect `U8` on the RHS, e.g. `(a as i32) >> (b as u8)`
- [x] Add descriptive error for functions that won't be implemented

### Testing 

The integration tests can be executed as follows:
```shell
cd test_programs/noir_test_success/comptime_blackbox
cargo run -p nargo_cli -- --program-dir . test --show-output --debug-comptime-in-file main.nr
```

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Ah no, we have an unimplemented error. What I mean is that we should tell the user that they can't verify a proof inside comptime and it's not a matter of its just us not implementing it yet.

compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs
```diff
@@ -53,21 +132,190 @@ pub(super) fn apply_range_constraint(
...
+    let (bytes, modulus) = check_two_arguments(arguments, location)?;
+
+    let (bytes, _) = get_slice_map(interner, bytes, get_u8)?;
+    let (modulus, _) = get_slice_map(interner, modulus, get_u8)?;
+
+    let id = solver
+        .bigint_from_bytes(&bytes, &modulus)
+        .map_err(|e| InterpreterError::BlackBoxError(e, location))?;
+
+    Ok(Value::U32(id))
```
jfecher, line 153: How does this work if we need to return a BigInt value rather than an ID to one?
Don't we need to return a struct value: `BigInt { pointer: ?, modulus: ? }`?

As-is it looks like users would be able to do `let new_bigint = my_bigint + 2` and offset the id pointer

jfecher (changes requested):

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -232,8 +232,11 @@ impl<'local, 'context> Interpreter<'local, 'context> {
             "unresolved_type_is_field" => unresolved_type_is_field(interner, arguments, location),
             "unresolved_type_is_unit" => unresolved_type_is_unit(interner, arguments, location),
             "zeroed" => zeroed(return_type, location.span),
+            blackbox if BlackBoxFunc::is_valid_black_box_func_name(blackbox) => {
+                self.call_foreign(blackbox, arguments, location)
+            }
```
TomAFrench, line 237: For other reviewers, we discussed this and agreed that this is suboptimal but there's a number of blackbox functions in the stdlib which are marked as builtins and we didn't want to make those changes to the stdlib in this PR. This can be cleaned up in a followup.

compiler/noirc_frontend/src/hir/comptime/errors.rs
```diff
@@ -540,6 +546,10 @@ impl<'a> From<&'a InterpreterError> for CustomDiagnostic {
                 let msg = format!("{item} is currently unimplemented");
                 CustomDiagnostic::simple_error(msg, String::new(), location.span)
             }
+            InterpreterError::WillNotImplement { item, location, explanation } => {
+                let msg = format!("{item} is not going to be implemented");
+                CustomDiagnostic::simple_error(msg, explanation.clone(), location.span)
```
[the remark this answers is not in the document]
aakoshh: The `{item}` part of the message is constructed [here](https://github.com/noir-lang/noir/blob/6f47cf14ef4f4d501980ad30f2db711e919ccd26/compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs#L78-L92), it can be made separate for the `Unimplemented` and `WillNotImplement` case. Note that there is an `explanation` for the second case which tells that you cannot use X in compile time or we shouldn't be calling Y because it's supposed to be handled differently. 

If you have a better name for the variant then please share. I started with `Unavailable`, then went to `Unsupported`, then realised there are some other `UnsupportedSomethingSomething` already and didn't want to add one that sounds like a more general version of those, so I ended up with the one that conveys exactly what we mean, but I agree the message sounds a bit aggressive.

compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs
```diff
@@ -76,25 +324,118 @@ fn poseidon2_permutation(
...
+    use crate::hir::comptime::tests::with_interpreter;
+    use crate::hir::comptime::InterpreterError::{
+        ArgumentCountMismatch, Unimplemented, WillNotImplement,
+    };
+
+    use super::call_foreign;
+
+    /// Check that all `BlackBoxFunc` are covered by `call_foreign`.
+    #[test]
+    fn test_blackbox_implemented() {
```
TomAFrench, line 401: I'm not sure this is necessary tbh. If we want to keep this however then we should remove the `enum_iter` feature flag and just always have it on.

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
```diff
@@ -924,310 +932,186 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
...
+        /// }
+        /// ```
+        macro_rules! match_values {
+            (($lhs_value:ident as $lhs:ident $op:literal $rhs_value:ident as $rhs:ident) {
+                $(
+                    $variant:ident($typ:ty) from [$($promoted:ident),+] to $result:ident => $expr:expr
+                ),*
+                $(,)?
+             }
+            ) => {
```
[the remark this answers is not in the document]
aakoshh: Ah, okay, then I can remove most of these. It was indeed the bit shifting [in sha256](https://github.com/noir-lang/noir/blob/6f47cf14ef4f4d501980ad30f2db711e919ccd26/noir_stdlib/src/hash/sha256.nr#L420-L422) that the interpreter couldn't handle. I didn't check how Noir handles the other cases :-/  

Then I suppose I can narrow it down to only handle the case of shifting. Does Noir always restrict the `rhs` of bit shifting to `u8`?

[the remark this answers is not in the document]
TomAFrench: >Then I suppose I can narrow it down to only handle the case of shifting. Does Noir always restrict the rhs of bit shifting to u8?

yep that's right.

compiler/noirc_frontend/src/hir/comptime/errors.rs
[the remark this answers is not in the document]
jfecher: Maybe `InvalidInComptimeContext` for a name?

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin/builtin_helpers.rs
```diff
@@ -93,6 +98,47 @@ pub(crate) fn get_array(
...
+        _ => {
+            let expected = StructType::new(
+                StructId::dummy_id(),
+                Ident::new(name.to_string(), location.span),
+                location,
+                Vec::new(),
+                Vec::new(),
+            );
+            let expected = Type::Struct(Shared::new(expected), Vec::new());
+            type_mismatch(value, expected, location)
```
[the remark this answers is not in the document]
aakoshh: The return type is actually different in this case, it takes `EllipticCurvePoint` for example and returns `[Field; 3]`. 
I suppose we can look up the function attributes, although this would be the only function what needs it so far. 

I did it this way because for example `get_array` and all the other `get_<thing>` functions in `builtin_helpers` that take the `interner` seem to generate some kind of generic type like this:
```
let type_var = Box::new(interner.next_type_variable());
let expected = Type::Array(type_var.clone(), type_var);
type_mismatch(value, expected, location)
```
I thought you just want the general idea and the location.

compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs
aakoshh: I think it's nice to have one, if we care about providing nice error messages, might as well force ourselves to do so. 

I checked and found I think one instance where the enum iter was always on, not just in tests, all the others were conditional. It's a bit of a dance to have it and use it in tests, and I'm sure it doesn't help compile times to have it two ways, so I'm in favour of enabling it always.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin/builtin_helpers.rs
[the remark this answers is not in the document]
jfecher: Approximating the types in the other cases is usually fine, it's just a bit worse in the case of structs since you'd be hard coding the struct name, generic counts, and can panic if the dummy id leaks. Since we don't pass in the argument types at all though it'll have to do.

compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs
jfecher: A fun (unfun?) trick is that you can probably trick the type system using a macro:
```rs
comptime {
    let a = bigint::Secpk1Fq::from_le_bytes(&[9]);
    let b = bigint::BigInt::from_le_bytes(a.to_le_bytes(), bigint::bn254_fq);

    let b: u32 = std::meta::unquote!(quote { $b });
    println(b + 2);
}
```
That should probably work but it's not a big deal in that case since it's somewhat expected to be a type system hole anyway.

I think a case that would fail would be if a user or library author tried to access a struct field on the U32 id value.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin/builtin_helpers.rs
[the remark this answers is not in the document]
aakoshh: > It should be retrievable from either the function's parameters

Just so I understand: we could pass in the parameters from the `FuncMeta` that we can still look up in `call_special`, however if the argument is actually _not_ a `Value::Struct`, then wouldn't it be reasonable to assume that the type in the function parameter will also not be a `Type::Struct`? Otherwise the compiler would have rejected it already. By hardcoding the struct name (and the fields), we're a bit more suggestive of the assumptions on the interpreter side, otherwise it would say it expected exactly what it got, but it's not a struct.

compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs
aakoshh: Thanks for pointing out the holes! 😃 
Yes, like I mentioned in the PR a valid use case would IMO be to add a `clone()` method to `BigInt` that seems reasonable to stay on Noir side and copy the two fields. Best to go with a Struct.

compiler/noirc_frontend/src/hir/comptime/errors.rs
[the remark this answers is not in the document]
aakoshh: Renamed to `InvalidInComptimeContext` and using different `item`.

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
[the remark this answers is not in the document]
aakoshh: Changed the macros to only allow different types in case of bit shifting, and only `U8` on the RHS, like Noir. Kept the helpers so the formulas are only stated once (so we have 5 blocks of match arms, rather than 16).

compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs
aakoshh: Changed to construct `BigInt` based on the `return_type`.

compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs
aakoshh: Removed the feature flag and enabled `EnumIter` for good.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
aakoshh: Created https://github.com/noir-lang/noir/issues/6594

compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs
TomAFrench: @michaeljklein fyi on the above as you mentioned looking for hacky ways to convert between types for the enum experiment.

acvm-repo/blackbox_solver/src/bigint.rs
```diff
@@ -97,3 +97,51 @@ impl BigIntSolver {
         Ok(())
     }
 }
+
+/// Wrapper over the generic bigint solver to automatically assign bigint IDs.
+#[derive(Default, Debug, Clone, PartialEq, Eq)]
+pub struct BigintSolverWithId {
```
vezenovm, line 103: Just for consistency with the camel case can we rename this symbol to `BigIntSolverWithId`? This looks like it was moved from `acvm-repo/brillig_vm/src/black_box.rs` which also had `Bigint`. Let's just rename it here.

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin/builtin_helpers.rs
```diff
@@ -114,6 +160,47 @@ pub(crate) fn get_slice(
...
+pub(crate) fn get_fixed_array_map<T, const N: usize>(
+    interner: &NodeInterner,
+    (value, location): (Value, Location),
+    f: impl Fn((Value, Location)) -> IResult<T>,
+) -> IResult<([T; N], Type)> {
+    let (values, typ) = get_array_map(interner, (value, location), f)?;
+
+    values.try_into().map(|v| (v, typ.clone())).map_err(|_| {
+        // Assuming that `values.len()` corresponds to `typ`.
+        let Type::Array(_, ref elem) = typ else { unreachable!("it's an array") };
```
vezenovm, line 198: Can we improve this panic message to something like "Expected an array type"

compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs
```diff
@@ -1,40 +1,127 @@
...
+                "recursive_aggregation" => Some("A proof cannot be verified at comptime.".into()),
+                _ => None,
+            };
+            if let Some(explanation) = explanation {
+                let item = format!("Attempting to evaluate foreign function '{name}'");
+                Err(InterpreterError::InvalidInComptimeContext { item, location, explanation })
+            } else {
+                let item = format!("Comptime evaluation for foreign function '{name}'");
+                Err(InterpreterError::Unimplemented { item, location })
+            }
```
vezenovm, line 101:
```suggestion
            let explanation = match name {
                "schnorr_verify" => "Schnorr verification will be removed.".into(),
                "and" | "xor" => "It should be turned into a binary operation.".into(),
                "recursive_aggregation" => "A proof cannot be verified at comptime.".into(),
                _ => {
                    let item = format!("Comptime evaluation for foreign function '{name}'");
                    return Err(InterpreterError::Unimplemented { item, location })
                }
            };

            let item = format!("Attempting to evaluate foreign function '{name}'");
            Err(InterpreterError::InvalidInComptimeContext { item, location, explanation })
```
style nit to avoid the additional pattern matching on the option

compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs
```diff
@@ -76,25 +334,135 @@ fn poseidon2_permutation(
...
     arguments: Vec<(Value, Location)>,
     location: Location,
 ) -> IResult<Value> {
     let input = check_one_argument(arguments, location)?;
-    let input_location = input.1;
-
-    let (input, typ) = get_array(interner, input)?;
 
-    let input = try_vecmap(input, |integer| get_u64((integer, input_location)))?;
+    let (state, typ) = get_fixed_array_map(interner, input, get_u64)?;
```
vezenovm, line 345: nice!

vezenovm (approved):
Some nits, but LGTM!

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin/builtin_helpers.rs
aakoshh: Sure, I made it longer and state the that `get_array_map` is expected to have checked it was an array, which is why it's unreachable.

acvm-repo/blackbox_solver/src/bigint.rs
aakoshh: Renamed both.

TomAFrench:
@jfecher I know you're on holiday but can you approve? Github is busted so I can't dismiss your review.

aakoshh:
@jfecher do you think this can be merged?

PR #6558, opened 2024-11-19
fix: Do a shallow follow_bindings before unification

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6457

## Summary\*

Finishes monomorphization in the program above in ~1.3s on my machine compared to master which stack overflows after 1min 12s.

Fixes the issue by performing a shallow follow bindings call before each unification. This follows immediate type variable links but stops once a non-type variable is reached. This prevents the issue in the program above because previously in unification we'd unify as soon as possible. If `self` is unbound but `other` is a bound type variable, we could be continuously adding another link to the chain. E.g. `tv3 -> tv2 -> tv1 -> Field` unified with `tv4` may become `tv4 -> tv3 -> tv2 -> tv1 -> Field`.

With this change we follow bindings beforehand so we'd unify `tv4` to `Field` and would never build up the large chain to begin with.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir_def/types.rs
+++ b/compiler/noirc_frontend/src/hir_def/types.rs
@@ -1547,12 +1547,15 @@ impl Type {
     ) -> Result<(), UnificationError> {
         use Type::*;
 
-        let lhs = match self {
+        let lhs = self.follow_bindings_shallow();
+        let rhs = other.follow_bindings_shallow();
+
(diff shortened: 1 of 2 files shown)


TomAFrench (approved):

TomAFrench:
Nice!

PR #6563, opened 2024-11-20
feat(ssa): Loop invariant code motion

### LINKED ISSUE #6517: Add SSA optimization pass to hoist instructions out of blocks to deduplicate them or hoist them out of loops
### Problem

Certain instructions used in loops may be unnecessarily executed on each iteration of the loop. This can happen if the instruction only depends on other values used outside the loop. For example, the program:

```noir
unconstrained fn main(x: u32) {
    for _ in 0 .. 10 {
        println(x + 1);
    }
}
```
Currently produces the SSA:
```
brillig(inline) fn main f0 {
  b0(v0: u32):
    jmp b1(u32 0)
  b1(v1: u32):
    v4 = lt v1, u32 10
    jmpif v4 then: b3, else: b2
  b3():
    v6 = add v0, u32 1
    call v7(u1 1, v6, [u8 123, u8 34, u8 107, u8 105, u8 110, u8 100, u8 34, u8 58, u8 34, u8 117, u8 110, u8 115, u8 105, u8 103, u8 110, u8 101, u8 100, u8 105, u8 110, u8 116, u8 101, u8 103, u8 101, u8 114, u8 34, u8 44, u8 34, u8 119, u8 105, u8 100, u8 116, u8 104, u8 34, u8 58, u8 51, u8 50, u8 125] of u8, u1 0)
    v30 = add v1, u32 1
    jmp b1(v30)
  b2():
    return
}
```

### Happy Case

We can hoist instructions like `v6` out of their blocks to their earliest ancestor to get the SSA:
```
brillig(inline) fn main f0 {
  b0(v0: u32):
    v6 = add v0, u32 1
    jmp b1(u32 0)
  b1(v1: u32):
    v4 = lt v1, u32 10
    jmpif v4 then: b3, else: b2
  b3():
    call v7(u1 1, v6, [u8 123, u8 34, u8 107, u8 105, u8 110, u8 100, u8 34, u8 58, u8 34, u8 117, u8 110, u8 115, u8 105, u8 103, u8 110, u8 101, u8 100, u8 105, u8 110, u8 116, u8 101, u8 103, u8 101, u8 114, u8 34, u8 44, u8 34, u8 119, u8 105, u8 100, u8 116, u8 104, u8 34, u8 58, u8 51, u8 50, u8 125] of u8, u1 0)
    v30 = add v1, u32 1
    jmp b1(v30)
  b2():
    return
}
```
The downside of this optimization is that if we apply this for blocks in general we could move instructions out of rarely-executed else blocks and into the happy path.

Another benefit of this optimization is that it allows more instructions to be deduplicated by constant folding since moving instructions to their dependency's common ancestor is much more permissive than checking if the first usage of an instruction dominates each other usage. For example, if we have the program `if foo { ... }; if bar { ... }; if baz { ... }`, each if body could be duplicated but none dominate each other since it is possible the else branch could have been taken each time. If we hoist to before the first if however, we'd deduplicate the blocks at the cost of executing extra instructions in the else case.

Since this could be a tradeoff in guessing how many times certain blocks may be executed, we could implement an alternate version of this change to start out with: only hoist if the instruction is used multiple times. That way, we could still execute unnecessary instructions but we'd at least guarantee to optimize out (deduplicate) 1 instance of the same instruction.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #6517 

The issue references both hoisting instructions into their earliest ancestors or simply hoisting instructions out of loops. This PR looks to add a pass that simply tries to hoist loop invariants into their pre-header.  

## Summary\*

Description of the pass in the codebase:
```
The loop invariant code motion pass moves code from inside a loop to before the loop
if that code will always have the same result on every iteration of the loop.

To identify a loop invariant, check whether all of an instruction's values are:
- Outside of the loop
- Constant
- Already marked as loop invariants
- 
We also check that we are not hoisting instructions with side effects.
```

## Additional Context

I re-used some of the context structures and methods from the unrolling and constant folding passes. I felt we could do some consolidation of the loop optimization passes. 

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -84,7 +82,8 @@ impl Function {
     }
 }
 
-struct Loop {
+#[derive(Debug, Clone)]
```
vezenovm, line 85:
```suggestion
```

compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -104,6 +104,9 @@ pub(crate) fn optimize_into_acir(
         "After `static_assert` and `assert_constant`:",
     )?
     .try_run_pass(Ssa::unroll_loops_iteratively, "After Unrolling:")?
+    // We unroll small Brillig loops based off a comparison between the number of unrolled instructions
+    // and the loop overhead. Loop invariant code motion can affect this calculation so we run it after unrolling.
+    .run_pass(Ssa::loop_invariant_code_motion, "After Loop Invariant Code Motion:")
```
michaeljklein, line 109: Wouldn't `loop_invariant_code_motion` be expected to have a positive effect on unrolling small Brillig loops? I.e. we'd be able to have more small loops unrolled because their loop bodies could get smaller from `loop_invariant_code_motion`?

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -0,0 +1,329 @@
...
+                    let (instruction, _) = inserter.map_instruction(instruction_id);
+                    instruction.for_each_value(|value| {
+                        // We are implicitly checking whether the values are constant as well.
+                        // The set of values defined in the loop only contains instruction results and block parameters
+                        // which cannot be constants.
+                        is_not_loop_invariant |=
+                            defined_in_loop.contains(&value) && !loop_invariants.contains(&value);
+                    });
+
+                    if !is_not_loop_invariant
```
michaeljklein, line 91:
```suggestion
                    if is_loop_invariant
```
Double-negated boolean (`!` and `_not_`)

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -0,0 +1,329 @@
...
+                let (instruction, _) = inserter.map_instruction(instruction_id);
+                let new_results = push_instruction(
+                    instruction_id,
+                    instruction,
+                    &old_results,
+                    pre_header,
+                    &mut inserter.function.dfg,
+                );
+
+                replace_result_ids(&mut inserter.function.dfg, &old_results, &new_results);
```
michaeljklein, line 123: Do we expect `old_results` and `new_results` to always have the same number of elements? If so, please add an assertion here or in `replace_result_ids`, which currently `zip`'s the two `Vec`'s, which will silently skip results if one of the `Vec`'s is shorter than the other

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -0,0 +1,329 @@
...
+                    let (instruction, _) = inserter.map_instruction(*instruction_id);
+                    let new_results = push_instruction(
+                        *instruction_id,
+                        instruction,
+                        &old_results,
+                        block,
+                        &mut inserter.function.dfg,
+                    );
+
+                    replace_result_ids(&mut inserter.function.dfg, &old_results, &new_results);
```
michaeljklein, line 143: Same here r.e. `replace_result_ids`

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -0,0 +1,329 @@
...
+            return
+        }
+        ";
+
+        let ssa = ssa.loop_invariant_code_motion();
+        assert_normalized_ssa_equals(ssa, expected);
+    }
+
+    #[test]
+    fn hoist_invariant_with_invariant_as_argument() {
```
michaeljklein, line 270: Please add a test that a function call with a side effect inside a loop body fails to hoist

michaeljklein (changes requested):
A couple nits and possible additional test, but otherwise LGTM

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -0,0 +1,329 @@
...
+}
+
+impl Function {
+    fn loop_invariant_code_motion(&mut self) {
+        Loops::find_all(self).hoist_loop_invariants(self);
+    }
+}
+
+impl Loops {
+    fn hoist_loop_invariants(self, function: &mut Function) {
```
jfecher, line 49: Can you break this up into helper functions to show the individual steps a bit better?

jfecher (changes requested):

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: Switched to not do the final double negation. I had to change the logic of the instruction value mapping above as well.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: done

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -0,0 +1,329 @@
...
+                            inserter.function.dfg.instruction_results(instruction_id).to_vec();
+                        let results = results.into_iter().map(|value| inserter.resolve(value));
+
+                        loop_invariants.extend(results);
+                        instructions_to_hoist.push(instruction_id);
+                    } else {
+                        instructions_to_keep.push(instruction_id);
+                    }
+                }
+                block_to_instructions.insert(*block, instructions_to_keep);
```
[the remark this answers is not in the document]
vezenovm: I first operated on the current empty block directly but was having some issues. I think you are right though that it isn't needed, I'll look at how I can rewrite it. Most likely am just missing some mapping/resolution.

compiler/noirc_evaluator/src/ssa.rs
vezenovm: Yeah good catch. I had moved it as with the original pass I wrote `slice_loop` was marking one less loop as small. But as you pointed out this is actually indicative of a deeper error as it would positively effect unrolling small loops. The error looks to be in how I was mapping values, thus why normalizing IDs also fixed the regression. I have moved it back to be before unrolling without a regression.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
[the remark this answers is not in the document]
vezenovm: I have removed this extra map.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: I have added a new `LoopInvariantContext` struct and have broken up this method into a few smaller methods.

vezenovm:
The blob binary goes from ~760 million executed opcodes on master to ~474 million opcodes executed when compiled with this branch.

jfecher (approved):
Looks much more readable now :+1:

Issue #6565, opened 2024-11-20
Performance regression after "Skip array_set pass for Brillig functions"


sirasistant:
### Aim

-

### Expected Behavior

-

### Bug

After this PR https://github.com/noir-lang/noir/pull/6513 in conjunction with this one https://github.com/noir-lang/noir/pull/6552 some functions of aztec contracts have blown up in opcodes executed. Particularly, Uniswap::swap_public is consuming 3 times more gas than before.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

sirasistant:
Added an `execution_success` test case with one of the functions that blew up in executed opcodes.

PR #6568, opened 2024-11-20
feat: Avoid incrementing reference counts in some cases

jfecher:
# Description

## Problem\*

Fixes https://github.com/noir-lang/noir/issues/6565

## Summary\*

Avoids incrementing array reference counts in some cases:
- If the parameter is an immutable array (we used to rely on the remove paired rc pass for this case)
- No longer increment rcs when dereferencing lvalues

## Additional Context

Since immutable arrays no longer issue rc instructions, the "remove paired rcs" pass is less needed. Now it only really applies when a user passes a mutable array but never mutates it (although it does still apply if that array is only passed to other mutating functions which is useful).

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/function_builder/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/function_builder/mod.rs
@@ -442,25 +442,20 @@ impl FunctionBuilder {
     /// within the given value. If the given value is not an array and does not contain
     /// any arrays, this does nothing.
     pub(crate) fn increment_array_reference_count(&mut self, value: ValueId) {
-        self.update_array_reference_count(value, true, true);
+        self.update_array_reference_count(value, true);
     }
 
     /// Insert instructions to decrement the reference count of any array(s) stored
     /// within the given value. If the given value is not an array and does not contain
     /// any arrays, this does nothing.
     pub(crate) fn decrement_array_reference_count(&mut self, value: ValueId) {
-        self.update_array_reference_count(value, false, true);
+        self.update_array_reference_count(value, false);
     }
 
     /// Increment or decrement the given value's reference count if it is an array.
     /// If it is not an array, this does nothing. Note that inc_rc and dec_rc instructions
     /// are ignored outside of unconstrained code.
-    pub(crate) fn update_array_reference_count(
-        &mut self,
-        value: ValueId,
-        increment: bool,
-        found_ref: bool,
-    ) {
+    fn update_array_reference_count(&mut self, value: ValueId, increment: bool) {
         match self.type_of_value(value) {
             Type::Numeric(_) => (),
             Type::Function => (),
@@ -468,18 +463,16 @@ impl FunctionBuilder {
                 if element.contains_an_array() {
                     let reference = value;
                     let value = self.insert_load(reference, element.as_ref().clone());
-                    self.update_array_reference_count(value, increment, true);
+                    self.update_array_reference_count(value, increment);
                 }
             }
             Type::Array(..) | Type::Slice(..) => {
                 // If there are nested arrays or slices, we wait until ArrayGet
                 // is issued to increment the count of that array.
-                if found_ref {
-                    if increment {
-                        self.insert_inc_rc(value);
-                    } else {
-                        self.insert_dec_rc(value);
-                    }
+                if increment {
+                    self.insert_inc_rc(value);
+                } else {
+                    self.insert_dec_rc(value);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


jfecher:
> keccak256 	+1,547 ❌ 	+70.16%

Hmm, not sure what's going on with this test. I haven't added any new instructions, just removed some existing inc_rcs

jfecher:
> sha256_brillig_performance_regression 	-411,727 ✅ 	-93.64%

Looks like this fixes https://github.com/noir-lang/noir/issues/6565

michaeljklein:
> Now it only really applies when a user passes an immutable array but never mutates it.

You mean when a user passes a _mutable_ array, but never mutates it?

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -735,7 +736,7 @@ impl<'a> FunctionContext<'a> {
             // Reference counting in brillig relies on us incrementing reference
             // counts when arrays/slices are constructed or indexed.
             // Thus, if we dereference an lvalue which happens to be array/slice we should increment its reference counter.
-            self.builder.increment_array_reference_count(reference);
+            // self.builder.increment_array_reference_count(reference);
```
michaeljklein, line 739: leftover comment

compiler/noirc_evaluator/src/ssa/function_builder/mod.rs
```diff
@@ -442,37 +442,44 @@ impl FunctionBuilder {
     /// within the given value. If the given value is not an array and does not contain
     /// any arrays, this does nothing.
     pub(crate) fn increment_array_reference_count(&mut self, value: ValueId) {
-        self.update_array_reference_count(value, true);
+        self.update_array_reference_count(value, true, true);
```
michaeljklein, line 445: 1. Is this always called when `found_ref == true` or does `found_ref` mean something different than "a reference has been found"?
2. I'd prefer these `true`'s to be labeled, at least when used outside of the helper functions

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -933,7 +936,7 @@ impl<'a> FunctionContext<'a> {
         dropped_parameters.retain(|parameter| !terminator_args.contains(parameter));
 
         for parameter in dropped_parameters {
-            self.builder.decrement_array_reference_count(parameter);
+            self.builder.update_array_reference_count(parameter, false, false);
```
michaeljklein, line 939: Please comment in these cases (or use labeled `bool`'s) to indicate that the `update_array_reference_count`'s are incrementing, decrementing the reference count, resp.

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
jfecher, line 739:
```suggestion
```

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
jfecher: I think I'm just going to restructure this to avoid the extra parameters and the exposing of `update_array_reference_count`

compiler/noirc_evaluator/src/ssa/function_builder/mod.rs
jfecher: I struggled naming this parameter. For parameters we only want to issue rc instructions after we find a reference type and recur, hence `found_ref`, but for other cases like assigning to a `let` binding we always want to inc_rc so `found_ref` is just overridden to true. I'm going to change this to remove the need for the parameter.

jfecher:
@asterite No, this is just a small ssa-gen change to not issue ref count instructions for immutable array parameters like `foo: [Field; 3]` (but still for `mut foo: [Field; 3]` and `foo: &mut [Field; 3]`), and also avoid issuing them within lvalues.
I don't think it's major enough to require a test for it (aside ensuring it works for our existing test_programs), but if we wanted a test we can't use the existing SSA test infrastructure since it's a ssa-gen change which needs to go from monomorphized ast -> SSA instead of SSA -> SSA.

compiler/noirc_evaluator/src/ssa/function_builder/mod.rs
jfecher: Removed

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
jfecher: Done

TomAFrench (approved):

vezenovm:
Just noting this PR also reduces opcodes executed from array copies in the blob binary. Went from ~171 million samples to ~71 million samples from array copies after this PR (the total samples when compiled with the current master is ~760 million).

michaeljklein (approved):

PR #6570, opened 2024-11-20
feat: Don't increment parameter ref counts at all

jfecher:
# Description

## Summary\*

We may not need to increment reference counts in parameters at all since https://github.com/noir-lang/noir/pull/6568 eliminates them for arrays, and previously we had eliminated them for references as well, though this was reverted. This PR is speculation that the reversion was only needed because we forgot to also remove the dec_rc in that case, leading to reference counts drifting down too far.

## Additional Context

This is passing locally but I think we're missing a inc_rc when dereferencing that is now required with this change.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
> fold_2_to_17 	1,168,465 (+21,730) 	+1.89%

I wonder what more is being executed here? When the only change is to issue fewer inc/decs, reference counts should in theory remain at 1 more often and just lead to more direct mutation.

jfecher:
Approach taken here is incorrect. It'd lead to mutation if the same parameter was passed as a reference twice.

Issue #6583, opened 2024-11-21
DIE is removing necessary `inc_rc` instructions


jfecher:
### Aim

Writing the function:
```noir
unconstrained fn borrow_mut(array: &mut [Field; 3]) {
    assert_refcount(*array, 2); // normal function call
    array[0] = 5;
    println(array[0]);
}
```
(with `--inliner-aggressiveness -1000`)

### Expected Behavior

SSA before DIE is as expected:
```
brillig(inline) fn borrow_mut f2 {
  b0(v0: &mut [Field; 3]):
    v1 = load v0 -> [Field; 3]
    inc_rc v1                                    // increase rc of array/v0
    v2 = load v0 -> [Field; 3]
    call f5(v2, u32 2)
    v5 = load v0 -> [Field; 3]
    inc_rc v5                                    // extra rc (removed in https://github.com/noir-lang/noir/pull/6568)
    v6 = load v0 -> [Field; 3]
    v9 = array_set v6, index u32 0, value Field 5
    store v9 at v0
    call f6(Field 5)
    dec_rc v9
    return
}
```

### Bug

Both `v1` and `v5` are otherwise unused in the above SSA so the program is incorrectly reduced down to:
```
brillig(inline) fn borrow_mut f2 {
  b0(v0: &mut [Field; 3]):
    v1 = load v0 -> [Field; 3]
    call f5(v1, u32 2)
    v4 = load v0 -> [Field; 3]
    v7 = array_set v4, index u32 0, value Field 5
    store v7 at v0
    call f6(Field 5)
    dec_rc v7
    return
}
```
Where all the `inc_rc`s are removed and the `dec_rc` even incorrectly remains

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #6584, opened 2024-11-21
feat: Add `array_refcount` and `slice_refcount` builtins for debugging

jfecher:
# Description

## Problem\*

Resolves 

## Summary\*

Added debug functions to retrieve the hidden reference count field on arrays and slices in unconstrained code. This always returns `0` in constrained and comptime code.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
This is failing under different inliner aggressiveness settings due to https://github.com/noir-lang/noir/issues/6583.

Currently it isn't possible to write the `reference_counts` test since the count differs depending on the inliner level.

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -402,210 +402,251 @@ impl<'block> BrilligBlock<'block> {
...
+                    // This match could be combined with the above but without it rust analyzer
+                    // can't automatically insert any missing cases
+                    match intrinsic {
+                        Intrinsic::ArrayLen => {
+                            let result_variable = self.variables.define_single_addr_variable(
+                                self.function_context,
+                                self.brillig_context,
+                                dfg.instruction_results(instruction_id)[0],
+                                dfg,
+                            );
```
jfecher, line 415: These cases are identical, I just moved them into a nested match so that I could use the code action to fill in the missing cases.

compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -2762,6 +2762,13 @@ impl<'a> Context<'a> {
             Intrinsic::FieldLessThan => {
                 unreachable!("FieldLessThan can only be called in unconstrained")
             }
+            Intrinsic::ArrayRefCount | Intrinsic::SliceRefCount => {
+                let zero = self.acir_context.add_constant(FieldElement::zero());
```
michaeljklein, line 2766: Why not fail here instead of returning an always-zero value?

test_programs/execution_success/reference_counts/src/main.nr
```diff
@@ -0,0 +1,40 @@
...
+    copy_mut(array, std::mem::array_refcount(array));
+}
+
+fn borrow(array: [Field; 3], rc_before_call: u32) {
+    assert_refcount(array, rc_before_call);
+    println(array[0]);
+}
+
+fn borrow_mut(array: &mut [Field; 3], rc_before_call: u32) {
+    assert_refcount(*array, rc_before_call + 0); // Issue! This should be rc_before_call + 1
```
michaeljklein, line 16: Follow up issue?

compiler/noirc_evaluator/src/acir/mod.rs
michaeljklein: (not blocking)

jfecher: I'm not sure - after your message I was thinking an error would be a better option so I added one in acir. Then we had:
- acir: error
- brillig: return refcount
- comptime: return 0

which felt odd with the comptime behavior. I thought about erroring in comptime too but I think we should still allow comptime to run code with this function in it. Particularly since `comptime { std::runtime::is_unconstrained()) }` is true. In the end I ended up keeping both acir & comptime at zero but would be open to reintroducing the error.

michaeljklein: The main reason I thought of for an error would be when testing constrained vs. unconstrained vs. comptime: an error makes it clear that we called one of these intrinsics.

PR #6592, opened 2024-11-22
chore!: Require types of globals to be specified

michaeljklein:
# Description

## Problem\*

Resolves [global type inference issues](https://github.com/noir-lang/noir/pull/6469/#issuecomment-2491997152) in https://github.com/noir-lang/noir/pull/6469/

## Summary\*

The approach in https://github.com/noir-lang/noir/pull/6469/ is to evaluate all globals as comptime globals, but that interacts poorly with polymorphic types, e.g. the type of `x` in `let x = 3`.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/tests.rs
```diff
@@ -3223,6 +3237,50 @@ fn infer_globals_to_u32_from_type_use() {
...
+        pub global ARRAY: [Field; _] = [0; 3];
+        pub global STR: str<_> = "hi";
+        pub global FMT_STR: fmtstr<_, _> = f"hi {ARRAY}";
+
+        fn main() { }
+    "#;
+
+    let errors = get_program_errors(src);
+    assert_eq!(errors.len(), 0);
+}
```
TomAFrench, line 3268: A little torn on this. I can see some benefits for the code writer in the case of strings but I think that the consistency of requiring these to be fully specified is good.

michaeljklein: Ok, sounds good

asterite:
Would a compiler flag to add missing types make sense? I'm thinking most if not all libraries break with this change.

Issue #6550, opened 2024-11-19
Profile Nargo's high memory usage during compilation and determine optimization opportunities


TomAFrench:
We're currently getting very high memory usage when compiling the protocol circuits in aztec-packages so we should put in some work in order to determine where this memory is going.

There's a couple of potential tools which can help with this:
- [Heaptrack](https://github.com/KDE/heaptrack): Seems like it wouldn't need any changes to our code and just works externally.
- [Tracy](https://github.com/wolfpld/tracy): Cody has recommended this as what the crypto team uses. This is targetted more at C++ but there's a [rust crate](https://docs.rs/tracing-tracy/latest/tracing_tracy/) to hook into it (It does mention the need to use https://docs.rs/tracy-client/0.17.4/tracy_client/index.html to get memory data out though)
- Others?

I don't know which option is best here so some exploration on how the various tools work and which is best would be very useful

Ideal outcomes:
- Internal docs for how to install the chosen tool/tools.
- Internal docs or scripts for usage of the chosen tool/tools.
- Shortlist of locations where we can improve memory usage.
- Memory reports which we can track in CI (nice to have)

guipublic:
I have investigated Heaptrack and Tracy:

## Heaptrack
[Heaptrack](https://github.com/KDE/heaptrack) is straightforward to use.

You can install it with: 
`sudo apt-get install heaptrack`
And check the installation with: `heaptrack -v`
Nargo can be profiled very easily: `heaptrack nargo compile` 
This generates a dump, which must be processed like this:
`heaptrack --analyze heaptrack.nargo.409749.gz`
The analyse reports all the allocation done by nargo. In particular, the peaks are interesting.
For instance, compiling the blake3 test, which simply performs one call to a blackbox function, mention several peak memory around the elaborator. It makes sense because in this case, most of the work is done on the frontend.

The summary contains the following information:
*total runtime: 1.47s.*
*calls to allocation functions: 1156118 (789159/s)*
*temporary memory allocations: 272381 (185925/s)*
*peak heap memory consumption: 78.39M*
*peak RSS (including heaptrack overhead): 105.05M*
*total memory leaked: 1.38M*

Heaptrack could be added to the CI in order to keep track of the *peak heap memory consumption*.

## Tracy
[Tracy](https://github.com/wolfpld/tracy) is more involved and cannot really be integrated with CI.
In fact Tracy is used to profile Barretenberg, via scripts and code instrumentation, and the scripts are not added to the CI.

[...]

PR #6600, opened 2024-11-22
chore: remove temporary allocations from `num_bits`

### LINKED ISSUE #4759: Update `FieldElement.num_bits()` to return 1 for a zero field element.
`FieldElement.num_bits()` currently returns 0 for a zero field element:
https://github.com/noir-lang/acvm/blob/39b8a41293e567971f700f61103852cb987a8d16/acir_field/src/generic_ark.rs#L204-L211

See attached comment: https://github.com/noir-lang/noir/pull/893#pullrequestreview-1308154733

TomAFrench:
# Description

## Problem\*

Closes #4759 

## Summary\*

I found this due to #6550.

We make heavy use of the `num_bits` method during compilation but this is currently doing a lot of unnecessary allocations in order to perform a full bit decomposition of the field element. We can rework this to work on the byte decomposition which removes 96% of all the temporary allocations from compilation.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/acvm-repo/acir_field/src/field_element.rs
+++ b/acvm-repo/acir_field/src/field_element.rs
@@ -371,6 +371,12 @@ mod tests {
     use super::{AcirField, FieldElement};
     use proptest::prelude::*;
 
+    #[test]
+    fn requires_one_bit_to_hold_zero() {
+        let field = FieldElement::<ark_bn254::Fr>::zero();
+        assert_eq!(field.num_bits(), 1);
+    }
+    
     proptest! {
         #[test]
         fn num_bits_agrees_with_ilog2(num in 1u128..) {
--- a/acvm-repo/acir_field/src/field_element.rs
+++ b/acvm-repo/acir_field/src/field_element.rs
@@ -213,7 +213,11 @@ impl<F: PrimeField> AcirField for FieldElement<F> {
         let mut iter = bytes.iter().skip_while(|x| (**x) == 0);
 
         // The first non-zero byte in the decomposition may have some leading zero-bits.
-        let head_byte = iter.next().copied().unwrap_or(0);
+        let Some(head_byte) = iter.next() else {
+            // If we don't have a non-zero byte then the field element is zero,
+            // which we consider to require a single bit to represent.
+            return 1;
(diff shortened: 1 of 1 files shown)


asterite:
Incredible! On Friday I was looking at small and quick optimization opportunities and found a couple of them, one of them was this one (not with this algorithm, but I noticed `num_bits` was allocating too much).

Another thing I found is that `FieldElement::hash` does `state.write(&self.to_be_bytes());` but I was wondering if we could directly call `self.0.hash(state)`, which seems to compile but I don't know why that's not done (though I don't know if it's more efficient).

Another one I found was this double hashmap lookup:

https://github.com/noir-lang/noir/blob/7311d8ca566c3b3e0744389fc5e4163741927767/acvm-repo/acvm/src/compiler/transformers/csat.rs#L274-L275

but I didn't send any PR because I wasn't sure if any of those improved the performance, really.

acvm-repo/acir_field/src/field_element.rs
```diff
@@ -224,12 +207,26 @@ impl<F: PrimeField> AcirField for FieldElement<F> {
...
+
+        // Iterate through the byte decomposition and pop off all leading zeroes
+        let mut iter = bytes.iter().skip_while(|x| (**x) == 0);
+
+        // The first non-zero byte in the decomposition may have some leading zero-bits.
+        let Some(head_byte) = iter.next() else {
+            // If we don't have a non-zero byte then the field element is zero,
+            // which we consider to require a single bit to represent.
+            return 1;
+        };
```
asterite, line 220: Is there a test for this case?

asterite (approved):
Looks good! I left a comment for something that might require an additional test.

acvm-repo/acir_field/src/field_element.rs
TomAFrench: Just added one.

TomAFrench:
> Another one I found was this double hashmap lookup:

We end up hitting our peak memory usage during SSA rather than in the post-compilation transformations so I think we can ignore this for now.

Issue #6604, opened 2024-11-22
Integrate Heapstack into CI to track the peak memory usage of the compiler on a set of programs.


TomAFrench:
Heaptrack has proven to be a useful tool so we should try to automate its usage. One easy way to do this is to add tracking of the peak memory usage in PRs, any significant changes should be reported with a warning comment.

We shouldn't need too many programs for this: a small, medium and large one should suffice as it'll allow us to see the fixed size memory costs vs worst case.

Ideally this should emit a workflow artifact of the profiling run so we can easily just download it and open in heaptrack.

PR #6606, opened 2024-11-22
feat: Reduce memory consumption by storing array length as `u32` during SSA

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

We're currently representing the length of a `Type::Array` using a usize. This means that despite the fact that Noir indexes its arrays with `u32`s we need to use 64bits to store this length within rust (assuming the user is on a 64bit system).

This PR switches how we store `Type`s within SSA so that we always store an array length as a u32 which saves significant amounts of memory (100 MB reduction on compiling `eddsa`).

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -307,13 +307,20 @@ impl DataFlowGraph {
...
+            .map(|(position, typ)| {
+                self.values.insert(Value::Instruction {
+                    typ,
+                    position,
+                    instruction: instruction_id,
+                })
+            })
+            .collect();
+
+        self.results.insert(instruction_id, results);
```
TomAFrench, line 323: Also includes this change which inlines an unnecessary function and avoid repeatedly pushing onto a `Vec` in favour of iterators.

jfecher:
> This PR switches how we store Types within SSA so that we always store an array length as a u32 which saves significant amounts of memory (100 MB reduction on compiling eddsa).

That must be a lot of array types being created for a 32 byte change to have such an impact. Maybe we should consider caching array types (or nested types in general) if they have such an impact.

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -307,13 +307,20 @@ impl DataFlowGraph {
...
+            .into_iter()
+            .enumerate()
+            .map(|(position, typ)| {
+                self.values.insert(Value::Instruction {
+                    typ,
+                    position,
+                    instruction: instruction_id,
+                })
+            })
+            .collect();
```
jfecher, line 321:
```suggestion
        let result_types = self.instruction_result_types(instruction_id, ctrl_typevars);
        let results = vecmap(result_types.into_iter().enumerate(), |(position, typ)| {
            let instruction = instruction_id; // could also rename the parameter
            self.values.insert(Value::Instruction { typ, position, instruction })
        });
```

jfecher (approved):

TomAFrench:
On second thoughts, I think we should hold off on merging this PR (and similar memory-motivated PRs) until we have #6604 .

Issue #6533, opened 2024-11-15
Deduplicate intrinsics with side effects when they're under the same predicate


TomAFrench:
We currently do not deduplicate any intrinics which have side effects. This results in us not deduplicating instrinsics which we should be able to (example given in #6527 where we do not deduplicate `Intrinsic::ToRadix` calls of the same variable in the same block).

We essentially need to make deduplication of these instructions aware of `deduplicate_with_predicate` similarly to how we do for other instructions.

jfecher:
We can add the `&& deduplicate_with_predicate` clause but we also need to actually match on them more closely in that case since we still wouldn't want to deduplicate e.g. `println(1)`

TomAFrench:
I don't think we have any intrinsics which result in foreign calls afaik but agreed in general that we should be careful when deduplicating these.

PR #6615, opened 2024-11-25
feat(ssa): Deduplicate intrinsics with predicates

aakoshh:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6533

## Summary\*

Allows `Instruction::Intrinsic` types to be deduplicated in `constant_folding` if the `Context` is in `use_constraint_info` mode, and the `Intrinsic` is one that doesn't have a side effect other than the possibility of violating an implicit constraint on its inputs, which should be preserved in the first occurrence.

## Additional Context

I found it a bit confusing why we're using `use_constraint_info` for these when`requires_acir_gen_predicate` returns `false` for all these instructions, which will cause `cached_instruction_results` to contain `None` for the predicate regardless of `use_constraint_info`. For some, but not all intrinsics, the predicate that tells whether the side effect is enabled is attached to their inputs during `Ssa::flatten_cfg`, and they don't depend any more on the `enable_side_effect` instruction. But it seems like this flag is the one that communicates whether we _want_ such instructions to be deduplicated, so I used it even though I'm not sure what adverse effect we would see if we just returned `true` for these, other than the implicit coupling by knowing that `flatten_cfg` has done things in a certain way.

Another method that seems to have some of the truth is `remove_enable_side_effects::Context::responds_to_side_effects_var`

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -152,6 +155,35 @@ impl Intrinsic {
...
+            // `constant_folding` we only use this information if the caller shows interest in it.
+            Intrinsic::ToBits(_)
+            | Intrinsic::ToRadix(_)
+            | Intrinsic::BlackBox(
+                BlackBoxFunc::MultiScalarMul
+                | BlackBoxFunc::EmbeddedCurveAdd
+                | BlackBoxFunc::RecursiveAggregation,
+            ) => deduplicate_with_predicate,
+
+            // As long as the constraints are applied on the same inputs.
```
vezenovm, line 177: Is this not implied by `can_be_deduplicated`? Why the separate match here?

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -152,6 +155,35 @@ impl Intrinsic {
...
+            Intrinsic::AssertConstant
+            | Intrinsic::StaticAssert
+            | Intrinsic::ApplyRangeConstraint
+            | Intrinsic::AsWitness => deduplicate_with_predicate,
```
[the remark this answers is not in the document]
vezenovm: The comment above `// As long as the constraints are applied on the same inputs.` this case also feels implied by "can_be_deduplicated"

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
aakoshh: See above; they don't have their input attached to the predicate already.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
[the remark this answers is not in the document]
aakoshh: Yes, I thought it might be enough to produce one witness for the same inputs, and hoed to clarify in the PR review with y’all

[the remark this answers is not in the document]
TomAFrench: I think for the purposes of `std::as_witness`, it doesn't require that distinct witnesses are created so I'd deduplicate these.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -112,6 +112,9 @@ impl Intrinsic {
     /// Returns whether the `Intrinsic` has side effects.
     ///
     /// If there are no side effects then the `Intrinsic` can be removed if the result is unused.
+    ///
+    /// An obvious example of a side effect is `println`, or pushing to a slice, but functions
+    /// which can fail due to implicit constraints are also considered to have a side effect.
```
TomAFrench, line 117: The `println` example doesn't apply here as that's a foreign call rather than an intrinsic.

aakoshh: I used it because [it's still in the docs](https://github.com/noir-lang/noir/blob/7d612425728a9d5852ec26a96af629aec7cd0a21/compiler/noirc_evaluator/src/ssa/ir/instruction.rs#L48-L49) of `Intrinsic` and it also appeared in [this comment](https://github.com/noir-lang/noir/issues/6533#issuecomment-2479731097), so I thought it might be some kind of honorary intrinsic, or a vestigial one. 

I'm happy to remove it from both places to avoid the confusion.

vezenovm (approved):

PR #6616, opened 2024-11-25
fix: LSP code action wasn't triggering on beginning or end of identifier

asterite:
# Description

## Problem

I noticed that LSP code actions weren't triggering when the cursor was right at the beginning (or end) of an identifier. This is a very common use case so this PR fixes that.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_errors/src/position.rs
+++ b/compiler/noirc_errors/src/position.rs
@@ -95,7 +95,7 @@ impl Span {
     }
 
     pub fn intersects(&self, other: &Span) -> bool {
-        self.end() > other.start() && self.start() < other.end()
+        self.end() >= other.start() && self.start() <= other.end()
     }
 
     pub fn is_smaller(&self, other: &Span) -> bool {
--- a/compiler/noirc_errors/src/position.rs
+++ b/compiler/noirc_errors/src/position.rs
@@ -97,6 +97,8 @@ impl Span {
     /// Returns `true` if any point of `self` intersects a point of `other`.
     /// If either span is empty, the point where the span is empty is considered
     /// to be part of the span (so, for example, `1..1` intersects `1..3`).
+    /// If both spans are empty, they are considered to intersect if they are the same
+    /// span (so `1..1` intersects `1..1`).
     pub fn intersects(&self, other: &Span) -> bool {
         self.end() >= other.start() && self.start() <= other.end()
     }
--- a/compiler/noirc_errors/src/position.rs
+++ b/compiler/noirc_errors/src/position.rs
@@ -95,10 +95,7 @@ impl Span {
     }
 
     /// Returns `true` if any point of `self` intersects a point of `other`.
-    /// If either span is empty, the point where the span is empty is considered
-    /// to be part of the span (so, for example, `1..1` intersects `1..3`).
-    /// If both spans are empty, they are considered to intersect if they are the same
-    /// span (so `1..1` intersects `1..1`).
+    /// Adjacent spans are considered to intersect (so, for example, `0..1` intersects `1..3`).
     pub fn intersects(&self, other: &Span) -> bool {
         self.end() >= other.start() && self.start() <= other.end()
     }
--- a/compiler/noirc_errors/src/position.rs
+++ b/compiler/noirc_errors/src/position.rs
@@ -94,6 +94,9 @@ impl Span {
         self.start() <= other.start() && self.end() >= other.end()
     }
 
+    /// Returns `true` if any point of `self` intersects a point of `other`.
+    /// If either span is empty, the point where the span is empty is considered
+    /// to be part of the span (so, for example, `1..1` intersects `1..3`).
     pub fn intersects(&self, other: &Span) -> bool {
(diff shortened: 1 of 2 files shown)


tooling/lsp/src/requests/code_action/import_or_qualify.rs
```diff
@@ -183,6 +183,31 @@ mod foo {
...
+    async fn test_import_code_action_for_struct_at_beginning_of_name() {
+        let title = "Import foo::bar::SomeTypeInBar";
+
+        let src = r#"mod foo {
+    pub mod bar {
+        pub struct SomeTypeInBar {}
+    }
+}
+
+fn foo(x: >|<SomeTypeInBar) {}"#;
```
aakoshh, line 201: I see in [other tests](https://github.com/noir-lang/noir/blob/c6ed325ebe0343d14150386f57cf141c9a7c9441/tooling/lsp/src/requests/code_action/import_or_qualify.rs#L143-L161) in the module that the `src` and `expected` can be indented along with the code, which would be much more readable.

compiler/noirc_errors/src/position.rs
```diff
@@ -95,7 +95,7 @@ impl Span {
     }
 
     pub fn intersects(&self, other: &Span) -> bool {
-        self.end() > other.start() && self.start() < other.end()
+        self.end() >= other.start() && self.start() <= other.end()
```
aakoshh, line 98: I see this is called through [includes_span](https://github.com/noir-lang/noir/blob/c6ed325ebe0343d14150386f57cf141c9a7c9441/tooling/lsp/src/requests/code_action/import_or_qualify.rs#L26). What worries me a bit about changing the definition of `intersects` is that the current definition seems correct: a `Span` is exclusive at the end, so when we say `self` intersects `other` if `self.end() >= other.start()` then the actual intersect is not part of `self`, which is weird.

Couldn't we instead implement `includes_span` in a way to say it's either intersect or is adjacent to the span?

tooling/lsp/src/requests/code_action/import_or_qualify.rs
asterite: Makes sense. Is it okay if I make that large change in a follow-up PR?

compiler/noirc_errors/src/position.rs
aakoshh: Hm, I see there is actually a `Span::empty(position)` constructor for a 0 length span, so technically we _could_ return a `Span` for the intersect that returns `true` when passed to `contains`. Previously I only saw `Span::single_character`. 

Perhaps some docstrings would be helpful for `intersect` to state that it considers adjacent texts intersecting with an empty string.

asterite: I'll add a few tests to Span, I also wasn't sure about this change.

asterite: I added a few tests and a comment. I think the logic is good (I think I got it wrong the first time 😊)

aakoshh: To clarify on the docs: it considers non-empty spans intersecting as well if one ends where the other starts, doesn't it? That is, `0..1` intersects `1..3` as well because `1 >= 1 && 0 <= 3`

aakoshh: What I wanted to point out with `Span::empty` earlier was that if it was defined like this:
```
impl Span { 
  fn intersect(&self, &other) -> Option<Span> { ... }
  fn intersects(&self, &other) -> bool { self.intersect(other).is_some() }
}
```
Then it would be possible to implement your semantics by returning `Span::empty(self.end)` for example in the edge case.

aakoshh: A counter example logic would be [this one](https://docs.rs/range-ext/latest/src/range_ext/intersect.rs.html#102) 🤷

asterite: This method is only used in a couple of places, and only in LSP code. I'm not sure splitting it into two makes sense now... or, maybe it would make things a bit more complex than necessary, not sure!

aakoshh: I only wanted to point out that the docstring says “if one of the spans is empty then it considers adjacent ones intersecting”, but this is suggests if they are both non-empty then a different rule applies, which isn’t the case. So I’m just saying the docs could be changed, not advocating for a split.

asterite: I adjusted a bit the comment. Let me know if it's okay now... though I don't think the original comment suggested something different (if one span is empty then the point where is empty is considered for intersection, so with two empty spans that logic still applies)

aakoshh: Haha, thanks, but I meant that two _non-empty_ ones are also intersecting if they touch. I am trying to say that there is no reason to mention empty input at all!

although at this point I am not sure any more I am not missing something myself 🙈

asterite: Done!

PR #6617, opened 2024-11-25
fix(LSP): use generic self type to narrow down methods to complete

### LINKED ISSUE #6608: LSP suggest unavailable methods
### Aim

Given a struct with some `impl`s, the LSP suggests autocompletion with fns that are not available in the current context.

### Expected Behavior

Consider the following program

```rust
struct Foo<T> {
    t: T,
}

impl Foo<Field> {
    fn bar(_self: Self) -> Field {
        5
    }
}

impl Foo<u32> {
    fn bar(_self: Self) -> Field {
        5
    }

    fn baz(_self: Self) -> Field {
        6
    }
}

fn main() -> pub Field {
    let foo: Foo<Field> = Foo { t: 5 };
    foo.bar()
}
```

In the last line, only `bar()` can be called on `foo`.

### Bug

The LSP suggests incorrect autocompletes:

Image

Note that two versions of `bar` are suggested, one for each `impl`. However, the second one cannot be called since I have `Foo<Field>` and not `Foo<u32>`.

`baz` is also suggested as an alternative, but this is invalid: that method is not available on the type since it belongs to the other specialization.

Finally, `t` is suggested even though it is not public (and currently results in a warning).

asterite:
# Description

## Problem

Resolves #6608

## Summary

I didn't know about having to use `find_matching_method` when looking up methods, so back then I ended up doing some similar logic that wasn't quite right.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/node_interner.rs
+++ b/compiler/noirc_frontend/src/node_interner.rs
@@ -2351,7 +2351,7 @@ impl Methods {
     }
 
     /// Select the 1 matching method with an object type matching `typ`
-    fn find_matching_method(
+    pub fn find_matching_method(
         &self,
         typ: &Type,
         has_self_param: bool,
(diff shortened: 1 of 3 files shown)


aakoshh (approved):
LGTM!

PR #6619, opened 2024-11-25
fix(ssa): Track all local allocations during flattening

vezenovm:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/pull/6434#discussion_r1828469417

## Summary\*

When there is a duplicate store in the same block without a load or call in between we used to remove that store in mem2reg. After https://github.com/noir-lang/noir/pull/6434 reactivating the following snippet:
```
                if let Some(last_store) = references.last_stores.get(&address) {
                    self.instructions_to_remove.insert(*last_store);
                }
``` 
gives this error:
Screenshot 2024-11-25 at 6 11 15 PM

In https://github.com/noir-lang/noir/pull/6434, when handing side effects during flattening we switched to doing `prev_value = load foo; store (if c then 3 else prev_value) in foo`. However, the first store to an allocation has no previous value (`foo = allocate; store 3 in foo`). So if we replace the first store with a load, the load is not going to be removed by mem2reg as it now does not have a known value. Now that the load is not being removed the allocation will also remain and not be removed by DIE as expected. We then finally error in ACIR gen for having a remaining allocation. 

#### Fix

I brought back the set of local allocations we had prior to #6434. In #6434 we also tried to account for this first store to an allocation case, however, we are currently only looking at stores which immediately follow an allocation instruction. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
@@ -203,10 +203,12 @@ struct Context<'f> {
     arguments_stack: Vec<Vec<ValueId>>,
 
     /// Stores all allocations local to the current branch.
-    /// Since these branches are local to the current branch (ie. only defined within one branch of
+    ///
+    /// Since these branches are local to the current branch (i.e. only defined within one branch of
     /// an if expression), they should not be merged with their previous value or stored value in
-    /// the other branch since there is no such value. The ValueId here is that which is returned
-    /// by the allocate instruction.
+    /// the other branch since there is no such value.
+    ///
+    /// The `ValueId` here is that which is returned by the allocate instruction.
     local_allocations: HashSet<ValueId>,
 }
 
--- a/compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
@@ -327,7 +327,6 @@ impl<'f> Context<'f> {
         // If this is not a separate variable, clippy gets confused and says the to_vec is
         // unnecessary, when removing it actually causes an aliasing/mutability error.
         let instructions = self.inserter.function.dfg[block].instructions().to_vec();
-        let mut previous_allocate_result = None;
 
         for instruction in instructions.iter() {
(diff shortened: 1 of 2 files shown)


compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -902,8 +903,8 @@ mod tests {
 
         let b1_instructions = main.dfg[b1].instructions();
 
-        // We expect the last eq to be optimized out, only the store from above remains
-        assert_eq!(b1_instructions.len(), 1);
+        // We expect the last eq to be optimized out
+        assert_eq!(b1_instructions.len(), 0);
```
TomAFrench, line 907: Seeing as we're touching this test can you convert it to use the parser?

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -201,6 +201,13 @@ struct Context<'f> {
     /// When processing a block, we pop this stack to get its arguments
     /// and at the end we push the arguments for his successor
     arguments_stack: Vec<Vec<ValueId>>,
+
+    /// Stores all allocations local to the current branch.
+    /// Since these branches are local to the current branch (ie. only defined within one branch of
+    /// an if expression), they should not be merged with their previous value or stored value in
+    /// the other branch since there is no such value. The ValueId here is that which is returned
+    /// by the allocate instruction.
```
TomAFrench, line 209:
```suggestion
    /// Stores all allocations local to the current branch.
    ///
    /// Since these branches are local to the current branch (i.e. only defined within one branch of
    /// an if expression), they should not be merged with their previous value or stored value in
    /// the other branch since there is no such value.
    ///
    /// The `ValueId` here is that which is returned by the allocate instruction.
```

TomAFrench (commented):
This looks very reasonable but it's late so will approve tomorrow after I've had another look.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
vezenovm: switched

TomAFrench (approved):

PR #6499, opened 2024-11-11
feat: Deduplicate instructions across blocks

jfecher:
# Description

## Summary\*

Deduplicates instructions across multiple blocks, mostly for brillig improvements.

Uses a dominator tree to ensure the instruction we're replacing the current with dominates the current block.

## Additional Context

This should resolve the duplicate `make_array` instructions in https://github.com/noir-lang/noir/pull/6071

## Documentation\*

Check one:
- [x] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #6627, opened 2024-11-26
fix(ssa): don't deduplicate constraints in blocks that are not dominated

### LINKED ISSUE #6588: Cross-block deduplication in constant folding has correctness bugs.
One of the components which has been causing issues on Noir syncs has been the changes made in #6499 so it's had to be pulled out of the syncs. 
I opened [this PR](https://github.com/AztecProtocol/aztec-packages/pull/9972) to test adding it in isolation and there's a number of test failures.

Sadly all of these tests rely on foreign calls so it's going to be a bit messy to get a nice reproduction case but we should find it and bring it back into this repo to fix the underlying issue.

asterite:
# Description

## Problem

Resolves #6588

## Summary

The PR fixes two issues in `constant_folding` introduced by https://github.com/noir-lang/noir/pull/6499 : 
1. The `constraint_simplification_mapping` didn't take into account the block in which the constraint was encountered to limit where the simplification can be used. 
2. Hoisting instructions into the common dominator should only happen when the instruction is free of side effects, because there can be other paths in which it's invalid. 

### Simplification

Here's an example where things went wrong:
```rust
fn main(x: Field, y: bool) {
    if y {
        assert_eq(x, 1);
    }
    print(x);
}
```
Calling the above with `main(0, false)` prints `1` on the console, because even though the body of the `if` is not hit, the `constrain x, 1` ended up taking effect outside the `if` and "simplified" `print(x)` to `print(1)`. 

The PR fixes this by remembering that the constraint was encountered in the block corresponding to the _then_ branch, and only uses the simplification in blocks which are dominated by it.

### Hoisting

For this to kick in we needed two blocks have the same expression:
```rust
fn main(x: Field, y: bool, z: bool) {
    if y {
        assert_eq(x, 1); // Caches the result of the instruction
    } 
    if z {
        assert_eq(x, 1); // Finds the result in the cache, but it's from a sibling block, so it hoists it up in the common dominator
    }
}
```

Calling the above with `main(0, false, false)` causes the circuit to fail, because it is rewritten to something like this:

```rust
fn main(x: Field, y: bool, z: bool) {
    assert_eq(x, 1); // Instruction hoisted to the block before both `if`s
    if y {
        assert_eq(x, 1); // Original instruction still in place until the next deduplication
    } 
    if z {
        // Instruction result now reusable from cache
    }
    // Now we enforce the constraint even when both y and z are false
}
```

The PR fixes this by only hoisting instructions which don't have side effects (such as failing the circuit, or modifying an array).

## Additional Context

Curiously the results cache does take into account the side effect variable a.k.a. predicate:
```rust
type InstructionResultCache = HashMap<Instruction, HashMap<Option<ValueId>, ResultCache>>;
                                                             ^
                                                             optional predicate
```
So it should be able to isolate the results based on whether we are before the `if`, in the _then_ or in the _else_ block. However it's use [depends](https://github.com/noir-lang/noir/blob/53f16c7fe75da04c54517b3d3199094b15195ce4/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs#L415) on whether `Instruction::requires_acir_gen_predicate` indicates that it should be used, and for `Constrain` this is [not the case](https://github.com/noir-lang/noir/blob/53f16c7fe75da04c54517b3d3199094b15195ce4/compiler/noirc_evaluator/src/ssa/ir/instruction.rs#L509), so it always ends up being stored with `None` for key.

```rust
            let use_predicate =
                self.use_constraint_info && instruction.requires_acir_gen_predicate(dfg);
            let predicate = use_predicate.then_some(side_effects_enabled_var);

            self.cached_instruction_results
                .entry(instruction)
                .or_default()
                .entry(predicate)
                .or_default()
                .cache(block, instruction_results);
```

NB previously every block had its own cache, so maybe the intent behind this was completely different, but it would be good to have clarification on how `flatten_cfg`, `enable_side_effect`, and this, are supposed to work together.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Nice! Good find.

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -226,9 +270,12 @@ impl<'brillig> Context<'brillig> {
         let mut side_effects_enabled_var =
             function.dfg.make_constant(FieldElement::one(), Type::bool());
 
+        let mut dom = DominatorTree::with_function(function);
```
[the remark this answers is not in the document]
aakoshh: I moved it from `self` because of borrow checking. `self.get_constraint_map` requires `&mut self` and then the lookup of the value in `SimplificationCache::get` requires `&mut DominatorTree` which would be another borrow of `&mut self`. I tried by making the `DominatorTree` use a `RefCell` to no avail.

Why is it slowing down, it's just created once I think.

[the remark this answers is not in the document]
aakoshh: Ah it's per block. Then it should be moved further up, and passed to `fold_constants_in_block` 🤦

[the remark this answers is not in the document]
aakoshh: Moved up.

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -611,6 +664,20 @@ impl<'brillig> Context<'brillig> {
...
+    dfg: &mut DataFlowGraph,
+    deduplicate_with_predicate: bool,
+) -> bool {
+    // These two can never be hoisted as they have a side-effect
+    // (though it's fine to de-duplicate them, just not fine to hoist them)
+    if matches!(instruction, Instruction::Constrain(..) | Instruction::RangeCheck { .. }) {
+        return false;
+    }
+
+    instruction.can_be_deduplicated(dfg, deduplicate_with_predicate)
```
aakoshh, line 678: This is implicit in the fact that we found a cached result: if the instruction couldn't be deduplicated it [would not be cached](https://github.com/noir-lang/noir/blob/7b28c597498fdbb16f336c3eea2c68ea9cfbb148/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs#L458).

Do you think this could be simplified to just: don't hoist `Instruction::Constrain`, which is the only instruction that feeds into the simplification mappings?

aakoshh: @TomAFrench I think the pattern we see is something like this:
```rust
fn main(x: Field, y: bool, z: bool) {
    if y {
        assert_eq(x, 1);
    } else if z {
        assert_eq(x, 1);
    } 
    // Here we don't state anything about x, but because it appears in both branches above
    // the constraint was hoisted above the `if` and then fails any time x isn't 1.
}
```

Do you see any other instruction that can cause issues? I'd think all other [assertion intrinsics](https://github.com/noir-lang/noir/blob/7b28c597498fdbb16f336c3eea2c68ea9cfbb148/compiler/noirc_evaluator/src/ssa/ir/instruction.rs#L117-L130) are suspect to this, aren't they? So with https://github.com/noir-lang/noir/pull/6615 in mind, we should be calling `has_side_effect`, not `can_be_deduplicated(deduplicate_with_predicate)`. 

I'm a bit confused by the `use_predicate` thing because for example `Constrain` doesn't interact with the predicate and will use `None` as a key. It sounds like should be able to signal that we cannot use an expression under a different side-effect-variable, but it's not stored there under any, even though it can clearly depend on the `if`.

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -188,12 +190,56 @@ pub(crate) struct BrilligInfo<'a> {
...
+        let existing = self.simplifications.entry(block).or_insert(simple);
+        // Keep the simpler expression in this block.
+        if *existing != simple {
+            match simplify(dfg, *existing, simple) {
+                Some((complex, simple)) if *existing == complex => {
+                    *existing = simple;
+                }
+                _ => {}
+            }
+        }
```
TomAFrench, line 216:
```suggestion
       self.simplifications
            .entry(block)
            .and_modify(|existing| {
                // `SimplificationCache` may already hold a simplification in this block
                // so we check whether `simple` is a better simplification than the current one.
                if let Some((_, simpler)) = simplify(dfg, *existing, simple) {
                    *existing = simpler;
                };
            })
            .or_insert(simple);
```

I think this better shows the intent of this function


`cargo fmt` is gonna not like this so will need to be run.

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
TomAFrench: I agree, we shouldn't hoist any instructions with side-effects out of their block as otherwise we can break ordering of side effects so we should switch to `has_side_effect`.

TomAFrench: Let me remind myself of `deduplicate_with_predicate`.

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -188,12 +190,56 @@ pub(crate) struct BrilligInfo<'a> {
...
+    }
+
+    /// Try to find a simplification in a visible block.
+    fn get(&self, block: BasicBlockId, dom: &mut DominatorTree) -> Option<ValueId> {
```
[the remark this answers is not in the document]
aakoshh: You're right, not any more 👍

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
aakoshh: Nice!

TomAFrench (approved):

TomAFrench:
@asterite @aakoshh Can you pull these fixes into https://github.com/AztecProtocol/aztec-packages/pull/9972?

PR #6628, opened 2024-11-26
fix: don't deduplicate constraints in blocks that are not dominated

asterite:
# Description

## Problem

Resolves https://github.com/noir-lang/noir/issues/6588

## Summary

Alternative to the approach in #6627. Here we keep track of all the simplifications a value could have (one per block) while #6627 only kept one simplification... which works, but this PR should result in more simplifications 🤞 

## Additional Context

There are a lot of nested structures in this file and it would be nice to somehow refactor it... but maybe it can be done in a separate PR, now that we have a regression test for the bug.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
@@ -169,12 +169,17 @@ struct Context<'a> {
 
     /// Contains sets of values which are constrained to be equivalent to each other.
     ///
-    /// The mapping's structure is `side_effects_enabled_var => (constrained_value => simplified_value)`.
+    /// The mapping's structure is `side_effects_enabled_var => (constrained_value => [(block, simplified_value)])`.
     ///
     /// We partition the maps of constrained values according to the side-effects flag at the point
     /// at which the values are constrained. This prevents constraints which are only sometimes enforced
     /// being used to modify the rest of the program.
-    constraint_simplification_mappings: HashMap<ValueId, HashMap<ValueId, ValueId>>,
+    ///
+    /// We also keep track of how a value was simplified to other values per block. That is,
+    /// a same ValueId could have been simplified to one value in one block and to another value
+    /// in another block.
+    constraint_simplification_mappings:
+        HashMap<ValueId, HashMap<ValueId, Vec<(BasicBlockId, ValueId)>>>,
 
     // Cache of instructions without any side-effects along with their outputs.
     cached_instruction_results: InstructionResultCache,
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -611,6 +623,20 @@ impl<'brillig> Context<'brillig> {
...
 }
 
+fn instruction_can_be_hoisted(
+    instruction: &Instruction,
+    dfg: &mut DataFlowGraph,
+    deduplicate_with_predicate: bool,
+) -> bool {
+    // These two can never be hoisted as they have a side-effect
+    // (though it's fine to de-duplicate them, just not fine to hoist them)
+    if matches!(instruction, Instruction::Constrain(..) | Instruction::RangeCheck { .. }) {
```
aakoshh, line 633: Wouldn't it be enough to do this for `Constrain`, because that's the one that interacts with simplifications that can replace other instructions' results? Other instructions can have side effects too, but `Constrain` has special role in `constant_folding`. I don't see `RangeCheck` have the same.

asterite: I think any instruction can be simplified. At least `instruction` here comes from `resolve_instruction` where any `Instruction` type can be returned.

aakoshh: What I meant was that we only have simplifications that [come from Constrain](https://github.com/noir-lang/noir/blob/3306f490bf100899cf46f366be2afe3732054f35/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs#L405), and I thought the hoisting of `Constrain` in the dominator is what's causing trouble further on. Just not sure what `RangeCheck` would do.

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -611,6 +623,20 @@ impl<'brillig> Context<'brillig> {
...
+        return false;
+    }
+
+    instruction.can_be_deduplicated(dfg, deduplicate_with_predicate)
```
aakoshh, line 637: Another thing that messes with my head is what if a side effecting instruction is followed by a non-side effecting one that depends on its results: would we hoist the latter but not the former? Would it still make sense on its own?

asterite: Good point, I'm not sure how that works.

aakoshh: I guess the we consider the instructions together with their inputs, which are value IDs. If they were visible in both blocks, then the IDs must come from their dominators, otherwise they would be different, as the SSA makes sure all values created have a unique ID.

TomAFrench:
I'd like to avoid having 2 PRs with 99% of the same code. Especially as #6627 is getting refactorings which disguise the meaningful differences in approach being taken.

Let's park this PR and get the other one merged to fix the correctness bug. We can then revisit how we can maintain more optimizations.

asterite:
Sounds good. The other PR also has more refactors and the code is a bit more readable 👍

PR #6635, opened 2024-11-27
fix: used signed division for signed modulo

### LINKED ISSUE #6609: Incorrect `%` (modulo) operator behavior for negative divisors
### Aim

Attempted to create a simple program which does the modulo operation between a positive and a negative number.

The following main.nr:
```rust
fn main(x: i16, y: i16) {
    println(x % -13);
    println(x % y);
}
```
with Prover.toml:
```toml
x = "169"
y = "-13"
```
Outputs
```
169
169
```

This output is unexpected. It should be
```
0
0
```

Another example is the following main.nr:
```rust
fn main(x: i16, y: i16) {
    println(x % -107);
    println(x % y);
}
```
with Prover.toml:
```toml
x = "769"
y = "-107"
```
Outputs
```
769
769
```
In Rust the same program would have the output 20. In some other programming languages the output would be -20

### Expected Behavior

Consisted behavior of the modulo operator. Currently the modulo operation between negative constants fails to compile but it has undefined behavior for negative integer parameters.

### Bug

The modulo operator produces an incorrect value when working with negative values passed as parameters.

### To Reproduce

1. Create a Noir project with the provided code
2. `nargo check`
3. Insert the provided values into the `Prover.toml`
4. `nargo execute`

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 0.39.0 noirc version = 0.39.0+45eb7568d56b2d254453b85f236d554232aa5df9 (git version hash: 45eb7568d56b2d254453b85f236d554232aa5df9, is dirty: false)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #6609

## Summary\*
Noir was using unsigned division for modulo, instead of signed division when the inputs are signed integers.

## Documentation\*

Check one:
- [] No documentation needed.

# PR Checklist\*


test_programs/execution_success/regression/src/main.nr
```diff
@@ -130,4 +130,8 @@ fn main(x: [u8; 5], z: Field) {
     assert(result_0 == 1);
     let result_4 = bitshift_variable(4);
     assert(result_4 == 16);
+
+    // Issue 6609
+    assert(u % -13 == 0);
```
TomAFrench, line 135: I think it's worth documenting how we define the modulo operator in the case where the rhs is negative.

test_programs/execution_success/regression/src/main.nr
```diff
@@ -130,4 +130,8 @@ fn main(x: [u8; 5], z: Field) {
...
+    assert(u % v == 0);
```
TomAFrench, line 136: Can you add another example for a non-zero result?

compiler/noirc_evaluator/src/acir/acir_variable.rs
```diff
@@ -1084,11 +1084,22 @@ impl<F: AcirField, B: BlackBoxFunctionSolver<F>> AcirContext<F, B> {
...
+        typ: AcirType,
         bit_size: u32,
         predicate: AcirVar,
     ) -> Result<AcirVar, RuntimeError> {
-        let (_, remainder) = self.euclidean_division_var(lhs, rhs, bit_size, predicate)?;
-        Ok(remainder)
+        let numeric_type = match typ {
+            AcirType::NumericType(numeric_type) => numeric_type,
+            AcirType::Array(_, _) => {
+                todo!("cannot modulo arrays. This should have been caught by the frontend")
```
TomAFrench, line 1094:
```suggestion
                unreachable!("cannot modulo arrays. This should have been caught by the frontend")
```

`todo` implies that we don't yet disallow this in the frontend, which I'd be very surprised by.

TomAFrench (approved):
Looks good. Some small docs/testing suggestions.

compiler/noirc_evaluator/src/acir/acir_variable.rs
guipublic: Done

test_programs/execution_success/regression/src/main.nr
guipublic: I have updated the signed integer section of the documentation

TomAFrench (approved):
LGTM

PR #6636, opened 2024-11-27
feat: allow filtering which SSA passes are printed

asterite:
# Description

## Problem

When debugging SSA issues we often want to see the result of one pass: the pass where the issue is. Right now there's no easy way to print just that.

## Summary

This PR adds a `--show-ssa-pass-name="..."` flag. Only SSA passes that have a name that contain the given string will be printed. You can still use `--show-ssa` to show them all.

I also renamed the passes so that you can always choose just one (for example, we had two "Dead Instruction Elimination" and there was no easy way to just print the first one).

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_driver/src/lib.rs
+++ b/compiler/noirc_driver/src/lib.rs
@@ -13,7 +13,7 @@ use noirc_abi::{AbiParameter, AbiType, AbiValue};
 use noirc_errors::{CustomDiagnostic, FileDiagnostic};
 use noirc_evaluator::create_program;
 use noirc_evaluator::errors::RuntimeError;
-use noirc_evaluator::ssa::SsaProgramArtifact;
+use noirc_evaluator::ssa::{SsaLogging, SsaProgramArtifact};
 use noirc_frontend::debug::build_debug_crate_file;
 use noirc_frontend::hir::def_map::{Contract, CrateDefMap};
 use noirc_frontend::hir::Context;
@@ -69,6 +69,11 @@ pub struct CompileOptions {
     /// Emit debug information for the intermediate SSA IR to stdout
     #[arg(long, hide = true)]
     pub show_ssa: bool,
+
+    /// If `show_ssa` is true, setting this will only show SSA passes whose name
+    /// contains the provided string.
+    #[arg(long, hide = true)]
+    pub show_ssa_pass_name: Option<String>,
 
     /// Emit the unoptimized SSA IR to file.
     /// The IR will be dumped into the workspace target directory,
@@ -577,7 +582,16 @@ pub fn compile_no_check(
     }
     let return_visibility = program.return_visibility;
     let ssa_evaluator_options = noirc_evaluator::ssa::SsaEvaluatorOptions {
-        enable_ssa_logging: options.show_ssa,
+        ssa_logging: match &options.show_ssa_pass_name {
+            Some(string) => SsaLogging::Equals(string.clone()),
+            None => {
+                if options.show_ssa {
+                    SsaLogging::All
+                } else {
+                    SsaLogging::None
+                }
+            }
+        },
         enable_brillig_logging: options.show_brillig,
         force_brillig_output: options.force_brillig,
         print_codegen_timings: options.benchmark_codegen,
--- a/compiler/noirc_evaluator/src/ssa.rs
+++ b/compiler/noirc_evaluator/src/ssa.rs
@@ -44,9 +44,16 @@ mod opt;
 pub(crate) mod parser;
 pub mod ssa_gen;
 
+#[derive(Debug, Clone)]
+pub enum SsaLogging {
+    None,
+    All,
+    Equals(String),
+}
+
 pub struct SsaEvaluatorOptions {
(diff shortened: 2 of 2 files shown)


compiler/noirc_driver/src/lib.rs
```diff
@@ -70,6 +70,11 @@ pub struct CompileOptions {
     #[arg(long, hide = true)]
     pub show_ssa: bool,
 
+    /// If `show_ssa` is true, setting this will only show SSA passes whose name
+    /// contains the provided string.
```
aakoshh, line 74: It will also print names if `show_ssa` is false, _but_ `options.show_ssa_pass_name` is given, won't it?

compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -85,46 +92,46 @@ pub(crate) fn optimize_into_acir(
...
         options.force_brillig_output,
         options.print_codegen_timings,
         &options.emit_ssa,
     )?
-    .run_pass(Ssa::defunctionalize, "After Defunctionalization:")
-    .run_pass(Ssa::remove_paired_rc, "After Removing Paired rc_inc & rc_decs:")
-    .run_pass(Ssa::separate_runtime, "After Runtime Separation:")
-    .run_pass(Ssa::resolve_is_unconstrained, "After Resolving IsUnconstrained:")
-    .run_pass(|ssa| ssa.inline_functions(options.inliner_aggressiveness), "After Inlining (1st):")
+    .run_pass(Ssa::defunctionalize, "Defunctionalization")
```
aakoshh, line 100: Instead of removing the "After" prefix and the colon, why not do what the option description says and check if the label _contains_ the pass name rather than _equals_ to it?

aakoshh: Oh I see you kept them further down, nice 👍

compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -44,9 +44,16 @@ mod opt;
 pub(crate) mod parser;
 pub mod ssa_gen;
 
+#[derive(Debug, Clone)]
+pub enum SsaLogging {
+    None,
+    All,
+    Equals(String),
```
aakoshh, line 51:
```suggestion
    Contains(String),
```
I found it misleading, I thought in the end it will really look for equality.

compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -460,9 +464,22 @@ impl SsaBuilder {
...
+                let string = string.to_lowercase();
+                let string = string.strip_prefix("after ").unwrap_or(&string);
+                let string = string.strip_suffix(':').unwrap_or(string);
+
+                if msg.to_lowercase().contains(string) {
+                    self.ssa.normalize_ids();
+                    println!("After {msg}:\n{}", self.ssa);
+                }
+            }
         }
```
aakoshh, line 483:
```suggestion
        let print_ssa_pass = match &self.ssa_logging {
            SsaLogging::None => false,
            SsaLogging::All => true,
            SsaLogging::Equals(string) => {
                let string = string.to_lowercase();
                let string = string.strip_prefix("after ").unwrap_or(&string);
                let string = string.strip_suffix(':').unwrap_or(string);
                msg.to_lowercase().contains(string)
            }
        };
        if print_ssa_pass {
            self.ssa.normalize_ids();
            println!("After {msg}:\n{}", self.ssa);
        }
```
So we only have that normalise-and-format logic once.

aakoshh (commented):
Cool! 

One extra feature that I think would possibly be useful is to be able to see the SSA before and after a step, which we could perhaps support by allowing the option to be like `--show-ssa-pass-name "remove ifelse|constant folding"`, ie to split on `|` and check if any such strings are contained.

compiler/noirc_driver/src/lib.rs
aakoshh: Maybe like this?

> Only show SSA passes whose name contains the provided string. If `show_ssa` is true, this setting takes precedence.

asterite: > If show_ssa is true

Did you mean "if show_ssa is false"?

compiler/noirc_evaluator/src/ssa.rs
asterite: Oh, I changed between the behaviors a couple of times and I forgot to change this one.

compiler/noirc_driver/src/lib.rs
asterite: I used your comment but changed a bit. Let me know if it's okay.

aakoshh (approved):
LGTM! Filtering by multiple fragments could be done separately as well, if people need it.

aakoshh:
I never thought about using diff on before vs after, only between branches. I only look at _before_ understand what the inputs of the pass were. Sure, we can run it twice 👍 👍

asterite:
That said, allowing "|" to choose multiple passes at once seems like something useful and easy to do, we can do it.

Issue #6637, opened 2024-11-27
`noirc_abi` cannot parse negative integer strings


TomAFrench:
I noticed that we're handling signed integers differently for JSON files compared to TOML.

https://github.com/noir-lang/noir/blob/6167b7384722821b3ad2d604802b4880f55cc3b4/tooling/noirc_abi/src/input_parser/toml.rs#L129-L145

https://github.com/noir-lang/noir/blob/6167b7384722821b3ad2d604802b4880f55cc3b4/tooling/noirc_abi/src/input_parser/json.rs#L146-L158

That is we don't have handling for negative integer strings in JSON. This means that the WASM ABI encoder can't handle these inputs which is bad.

We currently have proptests for ABI encoding but we should probably extend these to cover the input parsing and fix any issues which arise.

TomAFrench:
I've scoped out the tests necessary in #6638 but I'd appreciate others taking this up.

TomAFrench:
You've written the same expression as the code you're saying is incorrect but I would say that 255 is correct here as it's the two's complement of -1 which is how we represent signed integers.

asterite:
Oh, I see what you mean. I'll rollback that commit, or fix it and the fix the test.

asterite:
I think I got it right this time 🤞

asterite:
Closed by https://github.com/noir-lang/noir/pull/6638

PR #6638, opened 2024-11-27
fix: correct signed integer handling in `noirc_abi`

TomAFrench:
# Description

## Summary\*

This is a starting point for #6637. The tests are currently failing on both inputs as a negative input will be written back to file as if it were a positive integer.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/noirc_abi/proptest-regressions/input_parser/json.txt
+++ b/tooling/noirc_abi/proptest-regressions/input_parser/json.txt
@@ -1,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
+cc b3f9ae88d54944ca274764f4d99a2023d4b0ac09beb89bc599cbba1e45dd3620 # shrinks to (typ, value) = (Integer { sign: Signed, width: 1 }, -1)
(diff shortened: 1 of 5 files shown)


TomAFrench:
Nit picky but the issue isn't in the test. The problem is that when we try to write a negative integer value to file we end up writing the positive integer which is its two's complement representation.

We should update the serialization logic to correctly apply the two's complement conversion so that we get back the original without the need for any conversion.

asterite:
Would serializing back to integers be fine too?

TomAFrench:
For the purposes of this PR let's keep to hex strings. Currently JS always returns hex strings for fields and integers and adding extra degrees of freedom there seems bad.

TomAFrench:
Let's add a test in the JS ABI encoder package to double check that we handle them properly as well.

asterite:
Done!

> Let's add a test in the JS ABI encoder package to double check that we handle them properly as well.

I don't know where that is but I also ran out of time today. Feel free to add it tomorrow or let me know where and I'll do it.

TomAFrench:
Nice!

> I don't know where that is but I also ran out of time today. Feel free to add it tomorrow or let me know where and I'll do it.

We have a test here which we can add this to: https://github.com/noir-lang/noir/blob/master/tooling/noirc_abi_wasm/test/node/abi_encode.test.ts

guipublic:
I mark it as ready for review, test is fixed.

asterite (approved):

PR #6642, opened 2024-11-28
chore: pin foundry version in CI

TomAFrench:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/.github/workflows/test-js-packages.yml
+++ b/.github/workflows/test-js-packages.yml
@@ -478,6 +478,9 @@ jobs:
 
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1.2.0
+        with:
+          version: nightly-8660e5b941fe7f4d67e246cfd3dafea330fb53b1
+          
 
       - name: Install `bb`
         run: |


.github/workflows/test-js-packages.yml
```diff
@@ -478,6 +478,9 @@ jobs:
 
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1.2.0
+        with:
+          version: nightly-8660e5b941fe7f4d67e246cfd3dafea330fb53b1
```
aakoshh, line 482: Shouldn't we use the same commit as [aztec-packages](https://github.com/AztecProtocol/aztec-packages/blob/8835b31d76b2f7c45416eaf67a748d8df9dbc753/bootstrap.sh#L79)? Although it did require a different Rust version than the repo, so if this work then great.

aakoshh (approved):
Does this have something to do with [this failure](https://github.com/noir-lang/noir/actions/runs/12066976367/job/33649104119)?

.github/workflows/test-js-packages.yml
TomAFrench: Nah, there's no need to align with them on this.

TomAFrench:
Yep, this PR fixes that error.

PR #6650, opened 2024-11-29
chore: simplify MSM with constant folding

### LINKED ISSUE #6549: Simplify MSM calls for which only a subset of inputs are constant.
We currently simplify any MSMs for which all the inputs are known however we can also do partial simplifications of MSM opcodes.

As ECadd/ECmul are commutative and associative, we can group together any constant terms and simplify them down to a simple curve point. This then just needs to be added to the result of the non-constant terms.

guipublic:
# Description

## Problem\*

Resolves #6549

## Summary\*
fold constant values in MSM into one point

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/ir/instruction/call/blackbox.rs
```diff
@@ -70,52 +71,127 @@ pub(super) fn simplify_msm(
...
+                    }
+                    (Some(lo), Some(hi), Some(x), Some(y), Some(infinity)) => {
+                        is_constant = true;
+                        constant_scalars_lo.push(lo);
+                        constant_scalars_hi.push(hi);
+                        constant_points.push(x);
+                        constant_points.push(y);
+                        constant_points.push(infinity);
+                    }
+                    _ => {
```
michaeljklein, line 119: Since this is run in a loop, won't it unset `is_constant` if _any_ of the values in the MSM are non-constant? I.e. currently appx. `is_constant = scalars.all(|scalar| scalar.is_constant())`?

I think setting `let mut is_constant = true;` at the top and then skipping this `is_constant = false` will have behavior closer to `scalars.any(|scalar| scalar.is_constant())`, which is what I'd expect from your comment:

>  We decompose points and scalars into constant and non-constant parts in order to simplify MSMs where a subset of the terms are constant.

michaeljklein (changes requested):

TomAFrench:
The test for this simplification doesn't actually test anything can you add a test in SSA that shows the expected SSA being simplified?

compiler/noirc_evaluator/src/ssa/ir/instruction/call/blackbox.rs
guipublic: `is_constant` is properly set at each iteration because it is set in all branches of the match, so I think it is correct.

michaeljklein (approved):
LGTM

compiler/noirc_evaluator/src/ssa/ir/instruction/call/blackbox.rs
```diff
@@ -70,52 +71,127 @@ pub(super) fn simplify_msm(
...
+            for i in 0..len {
+                match (
+                    dfg.get_numeric_constant(scalars[2 * i]),
+                    dfg.get_numeric_constant(scalars[2 * i + 1]),
+                    dfg.get_numeric_constant(points[3 * i]),
+                    dfg.get_numeric_constant(points[3 * i + 1]),
+                    dfg.get_numeric_constant(points[3 * i + 2]),
+                ) {
+                    (Some(lo), Some(hi), _, _, _)
+                        if lo == FieldElement::zero() && hi == FieldElement::zero() =>
```
TomAFrench, line 94:
```suggestion
                        if lo.is_zero() && hi.is_zero() =>
```

compiler/noirc_evaluator/src/ssa/ir/instruction/call/blackbox.rs
```diff
@@ -70,52 +71,127 @@ pub(super) fn simplify_msm(
...
+                    {
+                        is_constant = true;
+                        constant_scalars_lo.push(lo);
+                        constant_scalars_hi.push(hi);
+                        constant_points.push(FieldElement::zero());
+                        constant_points.push(FieldElement::zero());
+                        constant_points.push(FieldElement::one());
+                    }
+                    (_, _, _, _, Some(infinity)) if infinity == FieldElement::one() => {
```
TomAFrench, line 103:
```suggestion
                    (_, _, _, _, Some(infinity)) if infinity.is_one() => {
```

TomAFrench (approved):
LGTM, some nits

compiler/noirc_evaluator/src/ssa/ir/instruction/call/blackbox.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/ir/instruction/call/blackbox.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/ir/instruction/call/blackbox.rs
```diff
@@ -70,52 +71,127 @@ pub(super) fn simplify_msm(
...
+                return SimplifyResult::SimplifiedTo(result_array.first());
+            }
+            // If there is only one non-null constant term, we cannot simplify
+            if constant_scalars_lo.len() == 1 && result_is_infinity != FieldElement::one() {
+                return SimplifyResult::None;
+            }
+            // Add the constant part back to the non-constant part, if it is not null
+            if result_is_infinity != FieldElement::one() {
+                let one = dfg.make_constant(FieldElement::one(), Type::field());
+                let zero = dfg.make_constant(FieldElement::zero(), Type::field());
```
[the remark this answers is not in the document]
guipublic: Done

PR #6652, opened 2024-11-29
Add regression test for inline brillig call bug

asterite:
# Description

## Problem

Resolves https://github.com/AztecProtocol/aztec-packages/pull/10291#issuecomment-2506735220

## Summary

For now just a regression test. I don't know why it's failing.

## Documentation\*

Check one:

# PR Checklist\*


asterite:
@TomAFrench 

I noticed the SSA code ends up having this line:

```
    v22 = mul u1 0, Field -8965431648830121816183193850757721314014926105976498373483500724795934473540
```

Is that valid? It's multipling u1 with Field.

TomAFrench:
Looking at this now. Yeah that looks dodgy, maybe we've got a bad type being returned from a simplified blackbox

TomAFrench:
Yeah, that's it.

TomAFrench:
We're pushing a bool into an array of 3 `Field`s here: https://github.com/noir-lang/noir/blob/aa143a75d3460785ed88ea7ab3337c880c1153fd/compiler/noirc_evaluator/src/ssa/ir/instruction/call/blackbox.rs#L49-L53

asterite:
Nice find!

TomAFrench (approved):

asterite:
Is there an assertion we could add somewhere to check that a simplified value has the same type as the expected type on the Noir side? I think this is the second time we had a wrong type in an instrinsic.

TomAFrench:
Up in `simplify_call` we have access to `ctrl_typevars` which we can assert to be compatible with the `SimplifyResult`. Currently checking that this is `Some` for this case.

asterite:
Great! But now I realized that even if we check that type, it wouldn't have worked in this case because we were inserting a wrong type inside an array. So I guess for now this PR like this is good.

TomAFrench:
Ah, that's very true. Well, I guess is can't hurt :shrug:

TomAFrench:
>  Great! But now I realized that even if we check that type, it wouldn't have worked in this case because we were inserting a wrong type inside an array. So I guess for now this PR like this is good.

Actually now I think of it, we can recursively pull out all of the value ids which are contained in the array and check those as well.

TomAFrench:
@asterite Can you transplant this fix across into aztec-packages and we can open an issue for the verification of the types of array elements.

asterite:
Sounds good, I'll add it to https://github.com/AztecProtocol/aztec-packages/pull/10291

PR #6657, opened 2024-11-29
fix: allow multiple `_` parameters, and disallow `_` as an expression you can read from

asterite:
# Description

## Problem

I noticed you couldn't define a function like `fn foo(_: i32, _: Field)`, it errored saying that `_` is defined multiple times.
Then I noticed you can have code like `let _ = 1; let x = _;`.

## Summary

Both things above are fixed with this PR (the first one doesn't error anymore, the second one errors with a message that's the same Rust produces).

## Documentation

Check one:

# PR Checklist


asterite:
Hm, it seems the debugger is inserting some "_" for "_" variables but I don't know where...

TomAFrench (approved):

Issue #5425, opened 2024-07-05
Underconstrained value detector


Rumata888:
### Problem

```rust
unconstrained fn maximum_price(options: [u32; 3]) -> u32 {
    let mut maximum_option= 0;
    for option in options {
        if (option > maximum_option) {
            maximum_option=option;
        }
    }
    maximum_option
}

fn main(sandwiches: pub [u32; 3], drinks: pub [u32; 3], snacks: pub [u32; 3], best_value: u32) {
    let meal_deal_cost:u32= 390;
    let most_expensive_sandwich= maximum_price(sandwiches);
    let mut sandwich_exists= false;
    for sandwich_price in sandwiches {
        assert(sandwich_price <= most_expensive_sandwich);
        sandwich_exists|=sandwich_price==most_expensive_sandwich;
    }
    assert(sandwich_exists);

    let most_expensive_drink= maximum_price(drinks);
    let mut drink_exists= false;
    for drink_price in drinks {
        assert(drink_price <= most_expensive_drink);
        drink_exists|=drink_price==most_expensive_drink;
    }
    assert(drink_exists);

    let most_expensive_snack= maximum_price(snacks);
    assert(
        best_value
        == (most_expensive_sandwich + most_expensive_drink + most_expensive_snack - meal_deal_cost)
    );
}
```

In the following code the `most_expensive_snack` is underconstrained. There are no checks making sure that the value is actually derived from previous witnesses.

### Happy Case

We should be able to detect this in Ssa. When there is a call to a brillig function in the Ssa we can construct a graph of ValueIds derived from the arguments of the functions and of ValueIds derived from the results. Two of those graphs need to have elements constrained together with an assert in most usecases. I think a warning to the user showing that there is no assert between any derived values would be useful.

### Project Impact

Nice-to-have

### Impact Context

Circuit security

### Workaround

None

### Would you like to submit a PR for this Issue?

Yes

PR #6658, opened 2024-11-29
feat(ssa): Implement missing brillig constraints SSA check

rkarabut:
# Description

This adds a new check to the security SSA passes module warning the developer about insufficiently constrained brillig calls.

## Problem

Brillig function calls not properly constrained by a later assert should be reported as bugs, as they present potential soundness vulnerabilities. This is referenced in and resolves [#5425](https://github.com/noir-lang/noir/issues/5425). 

## Summary

For now, we consider descendant values of every call result value and at least one of the argument values (if there are any) being involved in a set of constraint instructions proper coverage. If a result value itself (and not its descendant) is constrained (against a constant, for example), that too is currently considered proper. To check for it, we collect the descendant value ids for each brillig call met in a function and check them against constraints met later. If any of the brillig calls aren't sufficiently covered, we issue security bug warnings for them.

To illustrate this, the program mentioned in #5425 is also included in test-programs.

## Additional Context

This check has been tested on `noir-protocol-circuits` to measure how it affects compiler performance (employing --benchmark-codegen) and see how many bug occurrences there would be. At the time, the check catches 607 brillig calls there, and the performance hit seems to be negligible. In testing, some false positives have been revealed and addressed, with tests added accordingly.

For some reason, seemingly unrelated to the new check (should happen way before it, and anyway it wouldn't touch a function with no brillig calls) after merging in the latest master the `unconditional_recursion` test in the inliner fails on CI environment (getting caught with stack overflow _before_ reaching `RECURSION_LIMIT`) while passing both on mainframe and my dev machine. CI test reruns didn't help. Mitigated this by lowering `RECURSION_LIMIT` in the inliner for now, please decide if that should stay.

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/errors.rs
```diff
@@ -93,7 +93,10 @@ impl From<SsaReport> for FileDiagnostic {
                 let message = bug.to_string();
                 let (secondary_message, call_stack) = match bug {
                     InternalBug::IndependentSubgraph { call_stack } => {
-                        ("There is no path from the output of this brillig call to either return values or inputs of the circuit, which creates an independent subgraph. This is quite likely a soundness vulnerability".to_string(),call_stack)
+                        ("There is no path from the output of this brillig call to either return values or inputs of the circuit, which creates an independent subgraph. This is quite likely a soundness vulnerability".to_string(), call_stack)
+                    }
+                    InternalBug::UncheckedBrilligCall { call_stack } => {
+                        ("There is no assert checking this brillig call's inputs against its return values in some way. This should be done to prevent potential soundness vulnerabilities".to_string(), call_stack)
```
michaeljklein, line 99: Would this be more accurate since constraints other than `assert`'s are checked? It would agree with the wording in other messages.
```suggestion
                        ("There is no constraint between this brillig call's inputs and its return values. This should be done to prevent potential soundness vulnerabilities".to_string(), call_stack)
```

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
```diff
@@ -22,7 +22,7 @@ use fxhash::FxHashMap as HashMap;
 
 /// An arbitrary limit to the maximum number of recursive call
 /// frames at any point in time.
-const RECURSION_LIMIT: u32 = 1000;
+const RECURSION_LIMIT: u32 = 500;
```
michaeljklein, line 25: This appears to be an early indicator that this new pass is either recursing too much or else interfering with other passes. Because we've had other performance regressions recently, it will need need to be fixed if this pass is on by default.

Perhaps this could be addressed by reducing the amount of recursion used in this new pass? I wouldn't expect it to interfere with other passes. (But that could be tested, i.e. that running this pass leaves the result unchanged.)

test_programs/compile_success_with_bug/underconstrained_value_detector_5425/src/main.nr
```diff
@@ -0,0 +1,37 @@
+unconstrained fn maximum_price(options: [u32; 3]) -> u32 {
```
michaeljklein, line 1: Please add similar tests for when an unconstrained value in Brillig is:
- A single element of an array with at least two values
- In a nested brillig call, e.g. `foo(bar(xs))` where `xs: [u8; 3]`, `foo` constrains `xs[0]` and `bar` constrains `xs[1]` (i.e. `xs[2]` is unconstrained)
- In a nested type, e.g. `SomeStruct { field_a: [(value, ..), ..], .. }`

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,335 @@ fn check_for_underconstrained_values_within_function(
...
+                // of one loaded from the same memory slot
+                Instruction::Store { address, value } => {
+                    self.memory_slots.insert(*address, function.dfg.resolve(*value));
+                }
+                Instruction::Load { address } => {
+                    // Recall the value stored at address as parent for the results
+                    if let Some(value_id) = self.memory_slots.get(address) {
+                        self.update_children(&[function.dfg.resolve(*value_id)], &results);
+                    } else {
+                        debug!("load instruction {} has attempted to access previously unused memory location, skipping",
```
michaeljklein, line 262: I'd expect that encountering this would represent a compiler error so it should be a `panic`

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,335 @@ fn check_for_underconstrained_values_within_function(
...
+                                    BrilligTaintedIds::new(&arguments, &results),
+                                );
+                            }
+                            RuntimeType::Acir(..) => {
+                                // Record all the function arguments as parents of the results
+                                self.update_children(&arguments, &results);
+                            }
+                        },
+                        Value::ForeignFunction(..) => {
+                            debug!("should not be able to reach foreign function from non-Brillig functions, {func_id} in function {}", function.name());
```
michaeljklein, line 331: I'd expect that encountering this would represent a compiler error so it should be a `panic`

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,335 @@ fn check_for_underconstrained_values_within_function(
...
+                        }
+                        Value::Instruction { .. }
+                        | Value::NumericConstant { .. }
+                        | Value::Param { .. } => {
+                            debug!(
```
michaeljklein, line 336: I'd expect that encountering this would represent a compiler error so it should be a `panic`

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -32,6 +35,32 @@ impl Ssa {
             })
             .collect()
     }
+
+    /// Detect Brillig calls left unconstrained with manual asserts
+    /// and return a vector of bug reports if any have been found
+    pub(crate) fn check_for_missing_brillig_constraints(&mut self) -> Vec<SsaReport> {
+        // Skip the check if there are no Brillig functions involved
+        if !self.functions.values().any(|func| func.runtime().is_brillig()) {
```
michaeljklein, line 43: Can this be skipped since we're filtering for Brillig functions later on? If the pass is fairly performant, I'd expect this extra run through all functions to be slower than attempting the pass and finding zero Brillig functions.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -32,6 +35,32 @@ impl Ssa {
...
+            return vec![];
+        };
+
+        let functions_id = self.functions.values().map(|f| f.id().to_usize()).collect::<Vec<_>>();
```
michaeljklein, line 47: Why are you converting function ID's to `usize` and then back to `FunctionId` in the loop?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,335 @@ fn check_for_underconstrained_values_within_function(
...
+        // consider the call properly constrained
+        self.results.len() == self.results_constrained.len()
+    }
+
+    /// Remember partial constraints (involving some of the results and an argument)
+    /// along the way to take them into final consideration
+    fn store_partial_constraints(&mut self, constrained_values: &HashSet<ValueId>) {
+        let mut results_involved: Vec<usize> = vec![];
+
+        // For a valid partial constrain, a value descending from
```
michaeljklein, line 166:
```suggestion
        // For a valid partial constraint, a value descending from
```

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,335 @@ fn check_for_underconstrained_values_within_function(
...
+                // Check the constrain instruction arguments against those
+                // involved in Brillig calls, remove covered calls
+                Instruction::Constrain(value_id1, value_id2, _) => {
+                    self.clear_constrained(&[
+                        function.dfg.resolve(*value_id1),
+                        function.dfg.resolve(*value_id2),
+                    ]);
+                }
+                // Consider range check to also be constraining
+                Instruction::RangeCheck { value, .. } => {
```
michaeljklein, line 275: What if a range check were "empty," i.e. it checked that a value is within `0..Field::MAX`?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,335 @@ fn check_for_underconstrained_values_within_function(
...
+                    if let Some(value_id) = self.memory_slots.get(address) {
+                        self.update_children(&[function.dfg.resolve(*value_id)], &results);
+                    } else {
+                        debug!("load instruction {} has attempted to access previously unused memory location, skipping",
+                            instruction);
+                    }
+                }
+                // Check the constrain instruction arguments against those
+                // involved in Brillig calls, remove covered calls
+                Instruction::Constrain(value_id1, value_id2, _) => {
```
michaeljklein, line 268: What if `value_id1 == value_id2`, e.g. from `assert(x == x)`? Then shouldn't `x` still be unconstrained?

I think there might be similar cases with cycles of simple redundant constraints, e.g.

```noir
// x is unconstrained
unconstrained fn foo(x: Field) {
    let y = x;
    let z = y;
    // equivalent to "assert(x == x)"
    assert(x == z);
}
```

But I'm not sure what other cases could be in scope for this pass, because we can have cases like the following:

```noir
unconstrained fn foo(x: Field)
    assert(x == (x + CONSTANT - CONSTANT))
}
```

On that note: especially if it's on by default, this pass needs docs and you can add any caveats there.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,335 @@ fn check_for_underconstrained_values_within_function(
...
+                | Instruction::IfElse { .. }
+                | Instruction::Not(..)
+                | Instruction::Truncate { .. } => {
+                    // Record all the used arguments as parents of the results
+                    self.update_children(&arguments, &results);
+                }
+                // These instructions won't affect the dependency graph
+                Instruction::Allocate { .. }
+                | Instruction::DecrementRc { .. }
+                | Instruction::EnableSideEffectsIf { .. }
```
michaeljklein, line 366: Could using a value as the condition in `EnableSideEffectsIf` be considered a constraint?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,335 @@ fn check_for_underconstrained_values_within_function(
...
+        }
+    }
+
+    /// Check if any of the recorded Brillig calls have been properly constrained
+    /// by given values after recording partial constraints, if so stop tracking them
+    fn clear_constrained(&mut self, constrained_values: &[ValueId]) {
+        trace!("attempting to clear Brillig calls constrained by values: {:?}", constrained_values);
+
+        // For now, consider array element constraints to be array constraints
+        // TODO: this probably has to be further looked into, to ensure _every_ element
```
michaeljklein, line 407: Leftover TODO: please make a follow-up issue

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -433,4 +799,250 @@ mod test {
...
+            dec_rc v0
+            dec_rc v1
+            return
+        }
+        */
+        let type_u32 = Type::Numeric(NumericType::Unsigned { bit_size: 32 });
+        let type_u1 = Type::Numeric(NumericType::Unsigned { bit_size: 1 });
+
+        let main_id = Id::test_new(0);
+        let mut builder = FunctionBuilder::new("main".into(), main_id);
```
michaeljklein, line 863: Please use the new SSA parser instead of the `FunctionBuilder` for these tests

michaeljklein (changes requested):

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
TomAFrench: These are automatically ignored when being added: https://github.com/noir-lang/noir/blob/dc96de7c7a3d3c325b955edb277534653204cc5a/compiler/noirc_evaluator/src/ssa/ir/instruction.rs#L855

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Done

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: @Rumata888 what do you think?

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
rkarabut: I'm not sure how it could be interfering though as the stack overflow seemingly happens well before the check, in the inlining pass (where it should actually panic instead). Can you advise how to run this on CI with an option to skip the check passed to the compiler? The panic works correctly on any other environment I tried.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: The pass can be slow sometimes since it's going over every instruction sometimes gathering quite a footprint for brillig calls, especially in case they are unrolled large and brillig calls are in fact left unchecked. This filter doesn't get into function bodies at all, it just checks if any of involved functions is a brillig. 

By the way, this also should be ensuring the check doesn't do anything beyond this in the aforementioned `unconditional_recursion` test even if it the execution gets there somehow, as there are no brillig functions in the test.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: This is a leftover from copying some boilerplate of the previous check. Will fix it in both places.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Should be okay then

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Ok

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Ok

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Ok

compiler/noirc_driver/src/lib.rs
```diff
@@ -131,6 +131,12 @@ pub struct CompileOptions {
     #[arg(long)]
     pub skip_underconstrained_check: bool,
 
+    /// Flag to turn off the compiler check for missing Brillig call constrains.
+    /// Warning: This can improve compilation speed but can also lead to correctness errors.
+    /// This check should always be run on production code.
+    #[arg(long)]
+    pub skip_missing_brillig_constraints_check: bool,
```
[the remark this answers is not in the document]
rkarabut: Agreed

test_programs/compile_success_with_bug/underconstrained_value_detector_5425/src/main.nr
rkarabut: Can do 3 I believe, for a nested structure. For 1, checking if every element of an array is constrained is for now unsupported as mentioned further below (will make an issue for that).

2 I'll have to make sense of looking at the SSA. Still, won't work with array elements for now in any case.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Fixed

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Interesting point. Will look into it.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
Rumata888: No idea. Could you give an example of a specific case where it is used?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Where can I get a usage example?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Okay, as with RangeCheck this actually gets handled earlier in https://github.com/noir-lang/noir/blob/dc96de7c7a3d3c325b955edb277534653204cc5a/compiler/noirc_evaluator/src/ssa/ir/instruction.rs#L748

I'm not sure at a glance if `decompose_constrain` would handle the other sufficiently trivial cases, but anyway if they have to be handled it's seemingly should be the simplifier's job and not this pass. @TomAFrench do you agree with this?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Although yes, in case there are no brillig calls happening at all it would be a simple no-op run through every instruction. If you think it's alright to do it (that certainly doesn't take much time anyway), we can get rid of this initial check.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
TomAFrench: These are the current examples: https://github.com/search?q=repo%3Anoir-lang%2Fnoir%20ssa%3A%3Afrom_str&type=code

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
TomAFrench: I don't think this makes much difference either way tbh. The only situation where I see this causing a performance regression would be if someone wrote a program with a) no brillig calls (so the pass is should be skipped)  b) The program makes heavy use of folding so that there's lots of ACIR functions still in SSA at this point.

I think this is a very unlikely case as if the're a program which uses an advanced feature like folding they're likely to find _some_ opportunity for brillig calls to optimise their program. It's also a very cheap check to run on each function so I don't see a good reason not to.

TomAFrench: That said, I think a complete lack of brillig usage in a program which experiences significant slowdown due to this pass would be odd.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Thank you! On it

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
jfecher: `EnableSideEffectsIf` is a marker instruction to mark where `if` blocks were once they were flattened and removed. In pseudo code
```noir
if foo {
    print("foo");
} else {
    print("!foo");
}
```
would translate to the pseudo-ssa post-flattening:
```
enable_side_effects_if foo
call print("foo")
v0 = not foo
enable_side_effects_if v0
call print("!foo")
enable_side_effects_if true
... rest of program ...
```
It is mostly used to track whether we should actually perform or ignore calls since in our target there are no blocks or control-flow. Other instructions like `constrain` may have their own mechanisms for this. In the case of constrain we can multiply the constrain value by the current condition, e.g. `constrain a == b` becomes `constrain a*cond == b*cond` which will zero both out if `cond` is false because we're in a branch that wasn't taken.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
jfecher: I'd expect trivial constrains like `assert_eq(a, a)` to be removed before this pass, yes

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Doesn't seem like a valid constraint in terms of this check then. @Rumata888 do you agree?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Leaving this in then. @michaeljklein agreed?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Looks like it's okay then. @michaeljklein agreed?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Created https://github.com/noir-lang/noir/issues/6698, TODO updated

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: All the tests have been reworked, including the ones for the existing pass.

test_programs/compile_success_with_bug/underconstrained_value_detector_5425/src/main.nr
michaeljklein: Ok, then please add to the follow up issue to add support for array elements

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
rkarabut: So, to confirm the pass does not directly affect the issue I ran an experimental commit restoring the values _and_ disabling the check by default on CI. The `unconditional_recursion` test still fails: https://github.com/noir-lang/noir/actions/runs/12178227484/job/33967982930?pr=6658#step:6:737

@michaeljklein what do we do in this case? If the limit is arbitrary anyway serving to prevent this exact stack overflow, I don't see any harm in reducing it for the test that checks the panic works out. I could investigate further but the change in behavior looks pretty esoteric and doesn't seem to follow in any obvious way from the pass being introduced.

test_programs/compile_success_with_bug/underconstrained_value_detector_5425/src/main.nr
rkarabut: Added a test for briilig calls with nested type results. Mentioned this conversation in https://github.com/noir-lang/noir/issues/6698.

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
aakoshh: I came across this in a PR of mine where I thought it was caused by something I did. It seems to be that different test runners have different stack sizes, depending on the platform, which makes this test flaky. I ended up [using](https://github.com/noir-lang/noir/pull/6487/files#diff-dafe6612cf3a7b2e7c2890bd92574ea38432c2449c6f4a43fa0339a0dfd6a772R90) the `RUST_MIN_STACK` env var on CI to make it uniform.

This PR never got merged, but it's another way to solve the issue, in case lowering the recursion limit is not desired.

rkarabut: Seems like we've got to choose between these two.

test_programs/compile_success_with_bug/underconstrained_value_detector_5425/src/main.nr
```diff
@@ -0,0 +1,37 @@
...
+        if (option > maximum_option) {
+            maximum_option = option;
+        }
+    }
+    maximum_option
+}
+
+fn main(sandwiches: pub [u32; 3], drinks: pub [u32; 3], snacks: pub [u32; 3], best_value: u32) {
+    let meal_deal_cost: u32 = 390;
+    let most_expensive_sandwich = maximum_price(sandwiches);
```
aakoshh, line 13: Hm, I thought we would have to surround this call with `unsafe` as explained in the [docs](https://github.com/noir-lang/noir/blob/d86891bea9e01d12988d4ee344883d018be32ea7/docs/docs/noir/concepts/unconstrained.md?plain=1#L97-L98) 🤔

test_programs/compile_success_with_bug/underconstrained_value_detector_5425/Nargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "underconstrained_value_detector_5425"
```
aakoshh, line 2: I see this `compile_success_with_bug` directory is new. I don't think CI will do anything with it as it's not handled in [build.rs](https://github.com/noir-lang/noir/blob/d86891bea9e01d12988d4ee344883d018be32ea7/tooling/nargo_cli/build.rs#L32-L39). Perhaps `generate_compile_success_no_bug_tests` could be an example of what to do with it?

test_programs/compile_success_with_bug/underconstrained_value_detector_5425/src/main.nr
```diff
@@ -0,0 +1,37 @@
...
+    }
+    assert(sandwich_exists);
+
+    let most_expensive_drink = maximum_price(drinks);
+    let mut drink_exists = false;
+    for drink_price in drinks {
+        assert(drink_price <= most_expensive_drink);
+        drink_exists |= drink_price == most_expensive_drink;
+    }
+    assert(drink_exists);
```
aakoshh, line 27: What is the difference between the sandwich and the drinks case? 👀 
I see that snacks don't have it, but do we need all three?

It would be great to have some comments in the code describing what the bug is, to make it easier for anyone not familiar with the rules to see what's going on 🙏

test_programs/compile_success_with_bug/underconstrained_value_detector_5425/Nargo.toml
rkarabut: Oh right. Thank you, working on it

test_programs/compile_success_with_bug/underconstrained_value_detector_5425/src/main.nr
rkarabut: Commented

test_programs/compile_success_with_bug/underconstrained_value_detector_5425/Nargo.toml
rkarabut: Done

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
Rumata888: Unfortunately, I think we should treat it as one and mark all the results of instructions affected by a condition that uses brillig output as also tainted by brillig output. I'd argue that it doesn't work the same for the arguments of the brillig function, since it doesn't constrain them. However, brillig function calls could be used to enable a particular constraint.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,340 @@ fn check_for_underconstrained_values_within_function(
...
+        // descendant has been constrained, e.g. against a constant)
+        if !results_involved.is_empty()
+            && (self.arguments.is_empty()
+                || self.root_results.intersection(constrained_values).next().is_some()
+                || self.arguments.intersection(constrained_values).next().is_some())
+        {
+            // Remember the partial constraint
+            self.results_constrained.extend(&results_involved);
+            // We can clear the unneeded sets now
+            results_involved.iter().for_each(|i| self.results[*i].clear());
```
aakoshh, line 189: This makes me think that instead of this:
```rust
    results: Vec<HashSet<ValueId>>,
    // Indices in `results` already constrained
    results_constrained: HashSet<usize>,
```
You could have something like
```rust
enum ResultStatus { 
  Unconstrained { descendants: HashSet<ValueId> },
  Constrained,
}
results: Vec<ResultStatus>
```
Just a thought, maybe there are other reasons to keep the two separate.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,340 @@ fn check_for_underconstrained_values_within_function(
...
+    /// following constraints
+    fn process_instructions(
+        &mut self,
+        block: BasicBlockId,
+        function: &Function,
+        all_functions: &BTreeMap<FunctionId, Function>,
+    ) {
+        trace!("processing instructions of block {} of function {}", block, function);
+        let mut arguments = Vec::new();
+        let mut results = Vec::new();
```
aakoshh, line 218: Any reason not to declare a fresh pair inside the loop?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Yeah thanks this would be indeed cleaner. Done

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,340 @@ fn check_for_underconstrained_values_within_function(
...
+                // of one loaded from the same memory slot
+                Instruction::Store { address, value } => {
+                    self.memory_slots.insert(*address, function.dfg.resolve(*value));
+                }
+                Instruction::Load { address } => {
+                    // Recall the value stored at address as parent for the results
+                    if let Some(value_id) = self.memory_slots.get(address) {
+                        self.update_children(&[function.dfg.resolve(*value_id)], &results);
+                    } else {
+                        panic!("load instruction {} has attempted to access previously unused memory location, skipping",
```
aakoshh, line 275:
```suggestion
                        panic!("load instruction {} has attempted to access previously unused memory location",
```

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Extra allocations, although they probably get optimized out anyway. I don't think it matters a lot actually though. You thinking bringing them into the loop for clarity?

aakoshh: Yeah, just because then we don't have to think about how these are used after the loop, which they aren't. I don't typically see loop variables bubbling up to outside the loop very often.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,340 @@ fn check_for_underconstrained_values_within_function(
...
+        let mut results = Vec::new();
+
+        for instruction in function.dfg[block].instructions() {
+            // Optimization: there is no reason to do anything until a Brillig
+            // call is encountered
+            if self.tainted.is_empty() {
+                if let Instruction::Call { func: func_id, .. } = &function.dfg[*instruction] {
+                    if let Value::Function(callee) = &function.dfg[*func_id] {
+                        if !all_functions[&callee].runtime().is_brillig() {
+                            continue;
```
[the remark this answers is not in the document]
rkarabut: Yeah you're right, good catch. Fixing

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,340 @@ fn check_for_underconstrained_values_within_function(
...
+            match &function.dfg[*instruction] {
+                // For memory operations, we have to link up the stored value as a parent
+                // of one loaded from the same memory slot
+                Instruction::Store { address, value } => {
+                    self.memory_slots.insert(*address, function.dfg.resolve(*value));
+                }
+                Instruction::Load { address } => {
+                    // Recall the value stored at address as parent for the results
+                    if let Some(value_id) = self.memory_slots.get(address) {
+                        self.update_children(&[function.dfg.resolve(*value_id)], &results);
```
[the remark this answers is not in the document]
rkarabut: Resolve removed

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Okay

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,353 @@ fn check_for_underconstrained_values_within_function(
...
+        trace!("making following reports for function {}: {:?}", function.name(), warnings);
+        warnings
+    }
+
+    /// Update sets of value ids that can be traced back to the Brillig calls being tracked
+    fn update_children(&mut self, parents: &[ValueId], children: &[ValueId]) {
+        for (_, tainted_ids) in self.tainted.iter_mut() {
+            for parent in parents {
+                tainted_ids.update_children(parent, children);
+            }
```
aakoshh, line 412: This method is called as `self.update_children(arguments, results)`, then iterates over each argument calling the [update_children](https://github.com/noir-lang/noir/blob/55f991dc3b4111b768dc81a2e55be2ec6f481efe/compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs#L139-L152) of the tainted structure. 

NB the `children` will be processed multiple times, once for each argument that is already part of of `BrilligTaintedIds::arguments`. It could be more efficient to do the intersection trick on all parents, and if _any_ parent is found then add _all_ children once.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,353 @@ fn check_for_underconstrained_values_within_function(
...
+                            }
+                        },
+                        Value::ForeignFunction(..) => {
+                            panic!("should not be able to reach foreign function from non-Brillig functions, {func_id} in function {}", function.name());
+                        }
+                        Value::Instruction { .. }
+                        | Value::NumericConstant { .. }
+                        | Value::Param { .. } => {
+                            panic!(
+                                "should not be able to call {func_id} in function {}",
```
aakoshh, line 356:
```suggestion
                                "calling non-function value with ID {func_id} in function {}",
```

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,355 @@ fn check_for_underconstrained_values_within_function(
...
+                    if descendants.intersection(constrained_values).next().is_some() {
+                        results_involved.push(i);
+                    }
+                }
+            }
+        }
+
+        // Also, one of the argument descendants should be constrained
+        // (skipped if there were no arguments, or if an actual result and not a
+        // descendant has been constrained, e.g. against a constant)
```
aakoshh, line 195: Can you add a comment of why this is the rule?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -354,83 +732,268 @@ impl Context {
...
+                maximum_option = options[1];
+            }
+            maximum_option
+        }
+
+        fn main(sandwiches: pub [u32; 2], drinks: pub [u32; 2], best_value: u32) {
+            let most_expensive_sandwich = maximum_price(sandwiches);
+            let mut sandwich_exists = false;
+            sandwich_exists |= (sandwiches[0] == most_expensive_sandwich);
+            sandwich_exists |= (sandwiches[1] == most_expensive_sandwich);
```
aakoshh, line 802: Is there a test that shows what happens if only 1 element of the input array is constrained?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Alright. This should go into a separate issue then.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: No, checking individual array elements isn't supported currently. It's been moved into a separate issue: https://github.com/noir-lang/noir/issues/6698

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,355 @@ fn check_for_underconstrained_values_within_function(
...
+                let mut brillig_encountered = false;
+                if let Instruction::Call { func: func_id, .. } = &function.dfg[*instruction] {
+                    if let Value::Function(callee) = &function.dfg[*func_id] {
+                        if all_functions[&callee].runtime().is_brillig() {
+                            brillig_encountered = true;
+                        }
+                    }
+                }
+                if !brillig_encountered {
+                    continue;
```
aakoshh, line 244: So the rules to mark a Brillig call as constrained are that when we find a constraint, it should:
* be a descendant of one of the results of the call, and
  * be a descendant of one of the arguments of the call, or
  * constrain one of the immediate results of the call

The processing of instructions doesn't have a memory of constraints _before_ the Brillig call; only constraints that happen after the call are considered. 

I wonder what happens in an example such as this:

```rust
unconstrained fn times10(a: Field) -> Field { 
  a * 10
}
fn main(a: Field) {
  assert(a == 5);
  let b = times10(a);
  let c = 2 * b;
  assert(b == 100);
}
```

It feels like it shouldn't matter whether `a` gets constrained before or after the call to the Brillig function, but it looks like it might not realise because it would ignore that instruction.

aakoshh: Ah, never mind, the second assertion should contain descendants of `a` _as well as_ `b`.

aakoshh, line 244: What would happen in this example?
```rust
fn main(a: Field, b: Field) {
  let c = a + b;
  let d = unsafe { foo(a, b) };
  assert_eq(c, d);
}

unconstrained fn add(a: Field, b: Field) -> Field { 
  a + b
}
```

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Cleared the relationship between the result and argument descendant up a bit

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Hmm. Good point. Perhaps this optimization is not a good idea, at least being as sweeping. Removing it for now

aakoshh: I'm not sure that solves the issue. 

To be clear what we see here is that `c` is a descendant of `a` and `b`, so a constraint on `c` puts a constraint on `a` and `b`, which are the arguments of `add` and `sub`, and the `assert_eq` puts `c` in relation with the outputs of both brillig calls, and they aren't direct descendants. 

The problem is that even if you remove the optimisation, the only mechanism to remember that `c` had anything to do with `a` and `b` is the `tainted` collection, ie. until we see a call to `add` and `sub` we forget how `c` came to be, even if we process the instruction.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -63,6 +92,341 @@ fn check_for_underconstrained_values_within_function(
...
+                }
+            }
+        }
+
+        // Along with it, one of the argument descendants should be constrained
+        // (skipped if there were no arguments, or if an actual result and not a
+        // descendant has been constrained, e.g. against a constant)
+        if !results_involved.is_empty()
+            && (self.arguments.is_empty()
+                || self.root_results.intersection(constrained_values).next().is_some()
```
aakoshh, line 200: What happens in this example?

```rust
fn main(a: Field, b: Field) -> pub Field {
  let c = unsafe { add(a, b) };
  let d = unsafe { add(a, b) };
  assert_eq(c, d);
  (c + d)
}

unconstrained fn add(a: Field, b: Field) -> Field { 
  a + b
}
```

So here we have a constraint on the direct outputs of `add` but it doesn't have anything to do with its inputs. It looks like any implementation can be plugged in.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
aakoshh: Maybe we should do two passes:
1. one to collect all descendants of any Brillig call, then 
2. to collect all ancestors of constraints that intersect with that.

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
rkarabut: Switched to using the `RUST_MIN_STACK` solution.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: That's true. Probably could go backwards from the calls found unconstrained by the end only. I think this should go into a possible false positive issue like EnableSideEffectsIf though. Also, could lead to a significant performance hit, especially for huge unrolled functions.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Indeed. Fixing

aakoshh:
Maybe an illustration like this could be helpful at the top of the method:
```
  v1     v2      v3
   \   /  \    /
    \ /    \  /
     v4     v5 = call(v2, v3)
     |\     | 
     | \    |    
     |  \   |
     |   \  |
     |    \ |
     |      v6 = call(v5, v4) 
     |     /
     |    /
     |   /
     |  /
constrain(v4, v6)
```
With an explanation that we're looking for the ancestors or the constraint to intersect both the i) descendants (results) and the ii) ancestors (arguments) of the Brillig calls.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Moved into a separate issue: https://github.com/noir-lang/noir/issues/6735

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: This requires creating a HashSet of `parents` on the go, but yes it's probably worth it. Changed

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Done

Issue #6643, opened 2024-11-28
Add flag to `nargo test` to ignore failures due to missing foreign calls


TomAFrench:
For repositories such as aztec-packages there's a number of tests which rely on foreign call handlers, however we want to be able to run tests on the remainder of functions to see if any failures occur in pure Noir functions.

We should then add a flag which tells nargo to ignore any of these functions and mark them as skipped. For now we can just run the test and catch any unhandled foreign calls so there's no need to inspect the functions beforehand.

TomAFrench:
It's kinda annoying to pull out that a given test has failed due to an unhandled foreign call. I think the cleanest way to go about this is to have a specified foreign call handler for tests which tracks whether it's encountered an unknown foreign call and then we can modify the test status based off of that.

I'm doing some work to make the foreign call handlers more modular so we don't keep making the default one more and more complex.

PR #6659, opened 2024-11-29
chore: refactor foreign call executors

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This is a refactor to make it easier to implement #6643. We're probably going to want to have a new foreign call handler which we use for testing so I'd like to break these up a bit.

See https://github.com/noir-lang/noir/issues/6643#issuecomment-2508435660

As a bonus it becomes way easier to add a bunch of tests for how we deal with mocked foreign calls and external resolvers, etc.

## Documentation\*

Check one:
-  [x] No documentation needed.

# PR Checklist\*


asterite (approved):

tooling/nargo/src/foreign_calls/mocker.rs
```diff
@@ -0,0 +1,176 @@
...
+        foreign_call: &ForeignCallWaitInfo<F>,
+    ) -> Result<ForeignCallResult<F>, ForeignCallError> {
+        let foreign_call_name = foreign_call.function.as_str();
+        match ForeignCall::lookup(foreign_call_name) {
+            Some(ForeignCall::CreateMock) => {
+                let mock_oracle_name = Self::parse_string(&foreign_call.inputs[0]);
+                assert!(ForeignCall::lookup(&mock_oracle_name).is_none());
+                let id = self.last_mock_id;
+                self.mocked_responses.push(MockedCall::new(id, mock_oracle_name));
+                self.last_mock_id += 1;
```
aakoshh, line 95: It's not obvious to me why the mocks aren't stored in a `BTreeMap` indexed by ID. I looks like we won't have repeated IDs; I thought there might be multiple single-use responses lined up for the same ID, but not sure that's a use case.

TomAFrench: Our testing of mocked oracle calls is woefully incomplete atm so this may not be ideal. Let's handle this in a followup.

tooling/nargo/src/foreign_calls/rpc.rs
```diff
@@ -0,0 +1,227 @@
...
+use acvm::{acir::brillig::ForeignCallResult, pwg::ForeignCallWaitInfo, AcirField};
+use jsonrpc::{arg as build_json_rpc_arg, minreq_http::Builder, Client};
+use noirc_printable_type::ForeignCallError;
+use serde::{Deserialize, Serialize};
+
+use super::ForeignCallExecutor;
+
+#[derive(Debug)]
+pub(super) struct RPCForeignCallExecutor {
+    /// A randomly generated id for this `DefaultForeignCallExecutor`.
```
aakoshh, line 12: Should be `RPCForeignCallExecutor`

tooling/nargo/src/foreign_calls/rpc.rs
```diff
@@ -0,0 +1,227 @@
...
+    ///
+    /// This is used so that a single `external_resolver` can distinguish between requests from multiple
+    /// instantiations of `DefaultForeignCallExecutor`.
```
aakoshh, line 15: And this

Issue #6647, opened 2024-11-28
Update `run-external-checks` to run `nargo test` instead of `nargo check`


TomAFrench:
Blocked on:

https://github.com/noir-lang/noir/issues/6643
https://github.com/AztecProtocol/aztec-packages/issues/10278

TomAFrench:
Closed by #6660

PR #6660, opened 2024-11-29
feat: allow ignoring test failures from foreign calls

### LINKED ISSUE #6643: Add flag to `nargo test` to ignore failures due to missing foreign calls

TomAFrench:
# Description

## Problem\*

Resolves #6643 

## Summary\*

This PR adds the snappily named `NARGO_IGNORE_TEST_FAILURES_FROM_FOREIGN_CALLS` environment variable which if set tells nargo to ignore any failures which are thought to be a result from a missing foreign call handler.

We do this by creating a new `TestForeignCallResolver` which tracks any unknown foreign calls and marks any test which encounters an unknown foreign call and then fails as skipped.

## Additional Context

I'm purposefully not adding a proper user flag here as this feature is motivated by https://github.com/noir-lang/noir/issues/6647. If we need users to be able to do this then we can do a followup PR

## Documentation\*

Check one:

# PR Checklist\*


.github/workflows/test-js-packages.yml
```diff
@@ -554,9 +554,12 @@ jobs:
           # Github actions seems to not expand "**" in globs by default.
           shopt -s globstar
           sed -i '/^compiler_version/d' ./**/Nargo.toml
-      - name: Run nargo check
+      
+      - name: Run nargo test
         working-directory: ./test-repo/${{ matrix.project.path }}
         run: nargo check
```
TomAFrench, line 560:
```suggestion
        run: nargo test
```

.github/workflows/test-js-packages.yml
```diff
@@ -519,12 +519,26 @@ jobs:
...
+        - { repo: noir-lang/mimc, path: ./  }
+        - { repo: noir-lang/noir_sort, path: ./  }
+        - { repo: noir-lang/noir-edwards, path: ./  }
+        - { repo: noir-lang/noir-bignum, path: ./  }
+        - { repo: noir-lang/noir_bigcurve, path: ./  }
+        - { repo: noir-lang/noir_base64, path: ./  }
+        - { repo: noir-lang/noir_string_search, path: ./  }
+        - { repo: noir-lang/sparse_array, path: ./  }
+        - { repo: noir-lang/noir_rsa, path: ./lib  }
         - { repo: AztecProtocol/aztec-nr, path: ./ }
```
TomAFrench, line 533:
```suggestion
        - { repo: AztecProtocol/aztec-packages, path: ./noir-projects/aztec-nr }
```

.github/workflows/test-js-packages.yml
```diff
@@ -519,12 +519,26 @@ jobs:
...
+        - { repo: noir-lang/noir-edwards, path: ./  }
+        - { repo: noir-lang/noir-bignum, path: ./  }
+        - { repo: noir-lang/noir_bigcurve, path: ./  }
+        - { repo: noir-lang/noir_base64, path: ./  }
+        - { repo: noir-lang/noir_string_search, path: ./  }
+        - { repo: noir-lang/sparse_array, path: ./  }
+        - { repo: noir-lang/noir_rsa, path: ./lib  }
+        - { repo: AztecProtocol/aztec-packages, path: ./noir-projects/aztec-nr }
         - { repo: AztecProtocol/aztec-packages, path: ./noir-projects/noir-contracts }
         # Disabled as aztec-packages requires a setup-step in order to generate a `Nargo.toml`
```
TomAFrench, line 535:
```suggestion
```

tooling/nargo/src/ops/test.rs
```diff
@@ -217,3 +249,93 @@ fn check_expected_failure_message(
...
+}
+
+impl<F: AcirField + Serialize + for<'a> Deserialize<'a>> ForeignCallExecutor<F>
+    for TestForeignCallExecutor<F>
+{
+    fn execute(
+        &mut self,
+        foreign_call: &ForeignCallWaitInfo<F>,
+    ) -> Result<ForeignCallResult<F>, ForeignCallError> {
+        // If the circuit has reached a new foreign call opcode then it must have
```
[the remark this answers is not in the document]
TomAFrench: Thanks, fixed this.

.github/workflows/test-js-packages.yml
```diff
@@ -519,12 +519,25 @@ jobs:
       fail-fast: false
       matrix:
         project:
-        # Disabled as these are currently failing with many visibility errors
-        - { repo: AztecProtocol/aztec-nr, path: ./ }
```
aakoshh, line 523: What exactly happened with this?

aakoshh: Ah it moved further down into a `path`.

TomAFrench: Yeah, I noticed that aztec sometimes pushes out broken commits to this repository so we need to pull from aztec-packages.

tooling/nargo/src/ops/test.rs
```diff
@@ -217,3 +249,93 @@ fn check_expected_failure_message(
...
+
+                if let Some(external_resolver) = &mut self.external {
+                    // If the user has registered an external resolver then we forward any remaining oracle calls there.
+                    match external_resolver.execute(foreign_call) {
+                        Err(ForeignCallError::NoHandler(_)) => (),
+                        response_or_error => return response_or_error,
+                    };
+                }
+
+                self.encountered_unknown_foreign_call = true;
```
aakoshh, line 328: Maybe it would be helpful to allow a message to be attached to the test failure status, so that it can hint at why something has been skipped. `Skipped` suggests it's related to an `#[ignore]`-like attribute; in ScalaTest I used to use the [cancel](https://www.scalatest.org/scaladoc/3.0.5/org/scalatest/Assertions$.html#cancel(message:String)(implicitpos:org.scalactic.source.Position):Nothing) method to stop a running test because it could not be completed on CI for example.

PR #6666, opened 2024-12-02
feat: skip tests having oracle calls

### LINKED ISSUE #6643: Add flag to `nargo test` to ignore failures due to missing foreign calls

guipublic:
# Description

## Problem\*

Resolves #6643

## Summary\*
Add a 'skip-oracle' flag for the nargo test command.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_driver/src/lib.rs
```diff
@@ -139,6 +139,10 @@ pub struct CompileOptions {
     /// A lower value keeps the original program if it was smaller, even if it has more jumps.
     #[arg(long, hide = true, allow_hyphen_values = true)]
     pub max_bytecode_increase_percent: Option<i32>,
+
+    /// Flag to skip tests using oracles.
+    #[arg(long, hide = true)]
+    pub skip_oracle: bool,
```
aakoshh, line 145: Shouldn't this be added to the [TestCommand](https://github.com/noir-lang/noir/blob/d7a3207b6c3c1c487ce29282730b0ba44481410c/tooling/nargo_cli/src/cli/test_cmd.rs#L26) instead? It sounds more like a testing option than a general compile option which is shared with e.g. `execute` and `info`.

tooling/nargo/src/ops/foreign_calls.rs
```diff
@@ -60,6 +60,10 @@ impl ForeignCall {
             _ => None,
         }
     }
+
+    pub(crate) fn invalid_name(name: &str) -> bool {
+        ForeignCall::lookup(name).is_none()
```
aakoshh, line 65: Perhaps turning this into an `is_valid_name` would be more idiomatic, for example we have [BlackBoxFun::is_valid_black_box_func_name](https://github.com/noir-lang/noir/blob/d7a3207b6c3c1c487ce29282730b0ba44481410c/acvm-repo/acir/src/circuit/black_box_functions.rs#L249-L250) which is very similar.

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -263,33 +263,53 @@ fn display_test_report(
...
                 .expect("Failed to write to stderr");
         }
-
+        if count_skipped != 0 {
+            writer
+                .set_color(ColorSpec::new().set_fg(Some(Color::Yellow)))
+                .expect("Failed to set color");
+            write!(writer, ", {count_skipped} test{plural_skipped} skipped")
+                .expect("Failed to write to stderr");
+        }
```
aakoshh, line 312: I think at this point this could be simplified to something like this, so both passed and skipped are only handled once:
```
if count_passed > 0 || count_failed == 0 && count_skipped == 0 {
    // Show passed if there are some passed or if nothing else would be shown.
}
if count_skipped > 0 {
    ...
}
if count_failed > 0 {
    ...
} 
```

tooling/nargo/src/ops/foreign_calls.rs
guipublic: I changed the logic, this function is not idiomatic anymore. I tried to use a meaningful name.

tooling/nargo/src/ops/test.rs
```diff
@@ -11,17 +11,24 @@ use noirc_frontend::hir::{def_map::TestFunction, Context};
...
+    Skipped,
     CompileError(FileDiagnostic),
 }
 
 impl TestStatus {
     pub fn failed(&self) -> bool {
-        !matches!(self, TestStatus::Pass)
+        matches!(self, TestStatus::Fail { .. } | TestStatus::CompileError(_))
+    }
+    pub fn pass(&self) -> bool {
```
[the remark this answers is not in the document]
guipublic: Done

compiler/noirc_driver/src/lib.rs
guipublic: Done

tooling/nargo/src/foreign_calls/mod.rs
```diff
@@ -62,6 +66,20 @@ impl ForeignCall {
...
         }
     }
+
+    /// Returns the foreign call status based on its name:
+    /// create_mock => 'mocked'
+    /// valid ForeignCall name => 'handled'
+    /// otherwise => 'unhandled'
+    pub(crate) fn check_oracle_status(name: &str) -> OracleResult {
+        if name == ForeignCall::CreateMock.name() {
+            OracleResult::Mocked
```
aakoshh, line 76: Why aren't all the other [mock functions](https://github.com/noir-lang/noir/blob/14bb5ab228fcf0f995bac4b3a07773de6ae5b644/tooling/nargo/src/foreign_calls/mod.rs#L61-L65) considered to be `Mocked` status?

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -49,6 +49,10 @@ pub(crate) struct TestCommand {
     /// JSON RPC url to solve oracle calls
     #[clap(long)]
     oracle_resolver: Option<String>,
+
+    /// Flag to skip tests using oracles.
```
aakoshh, line 53:
```suggestion
    /// Flag to skip tests using unhandled oracles without mocks.
```
The way it's phrased sounds like any test with an oracle will be skipped.

tooling/nargo/src/ops/test.rs
```diff
@@ -60,6 +68,26 @@ pub fn run_test<B: BlackBoxFunctionSolver<FieldElement>>(
 
     match compile_no_check(context, config, test_function.get_id(), None, false) {
         Ok(compiled_program) => {
+            if skip_oracle {
+                let mut has_oracle = false;
+                for brillig_function in &compiled_program.program.unconstrained_functions {
+                    match brillig_function.get_oracle_status(ForeignCall::check_oracle_status) {
+                        OracleResult::Mocked => {
+                            has_oracle = false;
```
aakoshh, line 76:
```suggestion
                            // Assume that mocks will handle any call.
                            has_oracle = false;
```

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -275,32 +283,39 @@ fn display_test_report(
...
     let count_failed = test_report.iter().filter(|(_, status)| status.failed()).count();
-    let plural = if count_all == 1 { "" } else { "s" };
-    if count_failed == 0 {
+    let count_passed = test_report.iter().filter(|(_, status)| status.passed()).count();
+    let count_skipped = test_report.iter().filter(|(_, status)| status.skipped()).count();
+    let plural_failed = if count_failed == 1 { "" } else { "s" };
+    let plural_passed = if count_passed == 1 { "" } else { "s" };
+    let plural_skipped = if count_skipped == 1 { "" } else { "s" };
+    let mut previous = false;
+    if count_passed > 0 {
```
aakoshh, line 293:
```suggestion
    if count_passed > 0 || count_skipped == 0 && count_failed == 0 {
```
Previously if there were 0 tests executed it still gave feedback that "0 tests passed" as a confirmation that something happened. Maybe not the best message, but just wanted to highlight that it was there.

aakoshh: Arguably Rust's version is unambiguous:
```
running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 3 filtered out; finished in 0.00s
```

aakoshh: Actually, never mind, I see that the number of filtered tests is displayed before and after, so not displaying "0 tests passed" will make even more sense 👍 

```console
❯ cargo run -q -p nargo_cli -- --program-dir . test blah

[keccak256] Running 0 test functions
[keccak256] 0 tests passed
Found 0 tests containing 'blah'.
```

aakoshh (commented):
This looks good to me, just a couple of suggestions for additional docs/comments. I'll defer to @TomAFrench whether this should be the default on CI or the env var he did that detects a failure after attempting to run the test.

tooling/nargo/src/foreign_calls/mod.rs
guipublic: I wanted to identify mocked oracles by looking at the create_mock() function, but it is not easy. Instead I  used an heuristic that consider that all oracles are mocked if there is one create_mock.

Issue #6680, opened 2024-12-02
`Instruction::MakeArray` deduplication is incompatible with mutating arrays in unconstrained code


jfecher:
### Aim

```noir
fn main(x: Field, y: pub Field) {
    let qs = [0];

    let mut dynamic = [0];
    dynamic[x] = 1000;

    println(qs.as_slice()); // &[0]
    println(as_slice_push(qs)); // (loop to create a slice from qs): &[0]
}
```

### Expected Behavior

The above code to run with the slices given in comments to be printed

### Bug

In ssa, `dynamic` gets deduplicated to the same `make_array` instruction as `qs`. This occurs in master and is potentially problematic.

This latent issue was made easier to trigger failing cases for in https://github.com/noir-lang/noir/pull/6607 since reference counts were changed to effectively start at 1 instead of 2 so these arrays are actually being mutated now. This issue is still present on master though, just more difficult to trigger.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

vezenovm:
This issue has been resolved. See additional context in #6685

PR #6577, opened 2024-11-21
feat(opt): Reduce array cloning by initializing refcounts to 0, not 1

jfecher:
# Description

## Summary\*

I noticed after storing an array in a variable we were always issuing an `inc_rc` which lead to arrays having an rc of `2` to begin with instead of `1`, meaning the first mutation would always be cloned.

## Documentation\*

Check one:
- [x] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #6685, opened 2024-12-02
fix: Optimize array ref counts to copy arrays much less often

jfecher:
# Description

## Problem\*

This is an alternate fix for https://github.com/noir-lang/noir/pull/6577. Where that PR re-aligns ref counts by starting them at zero so that the initial `inc_rc` increments them up to 1, this PR removes the inc_rc in a few cases known to be correct.

Making this PR since https://github.com/noir-lang/noir/pull/6577 is running into issues with DIE removing the initial inc_rc leading to an invalid starting RC of zero.

## Summary\*

Starts reference counts at 1 instead of 2 for array literals. Instead of changing brillig to start ref counts at zero, avoid issuing the initial `inc_rc` instruction if the rhs of a let binding is known to be an array literal (and should thus be moved, not shared)

## Additional Context

~~Since this PR also lets us mutate arrays directly more often, it is also susceptible to https://github.com/noir-lang/noir/issues/6680 like #6577 is. That's where the CI failures are currently coming from.~~

To fix the above, we no longer deduplicate arrats willy-nilly in brillig. We:
1. Only deduplicate MakeArray instructions which have not been stored to or used in an ArraySet
2. When we deduplicate we insert an `inc_rc` to that array to increase the reference count. Although there should be no mutations before that point, there may still be some after so it needs to see that now the ref count is higher.

**Additionally**, I've added a new optimization: we don't issue `inc_rc` instructions for the first mutable array parameter of each type (but we do for each subsequent one). This means if you have `fn foo(a: &mut [Field; 3])`, `a` is considered mutably borrowed and we don't need a inc_rc since you can only access it through `a` anyway. If you add another `&mut [Field; 3]` parameter (with any number of nested references to it), then we issue an inc_rc for the next instance so that the rc is 2+. In practice this leads to pretty good improvements with any function that mutates arrays. You can see this in our opcodes executed even though most arrays in our tests tend to be small, we still copy them considerably less now.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
This PR should be ready to merge now. It has decent opcodes executed improvements with the optimization I mentioned (reference_counts is only larger because I expanded the test).

This PR may or may not require https://github.com/noir-lang/noir/pull/6700 - but either way the expected rcs in the reference_counts test will need to be updated once that PR is merged.

compiler/noirc_evaluator/src/brillig/brillig_ir/procedures/array_copy.rs
```diff
@@ -69,6 +69,9 @@ pub(super) fn compile_array_copy_procedure<F: AcirField + DebugToString>(
                 BRILLIG_MEMORY_ADDRESSING_BIT_SIZE,
                 1_usize.into(),
             );
+            // Decrease the rc of the original ref count now that this copy is
+            // no longer pointing to it
```
jfecher, line 73:
```suggestion
            // Decrease the original ref count now that this copy is no longer pointing to it
```

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -688,6 +699,26 @@ impl<'brillig> Context<'brillig> {
...
     }
+
+    fn remove_possibly_mutated_cached_make_arrays(
+        &mut self,
+        instruction: &Instruction,
+        function: &Function,
+    ) {
+        use Instruction::{ArraySet, Store};
+
+        // Should we consider calls to slice_push_back and similar to be mutating operations as well?
```
michaeljklein, line 710: Does this need a follow up issue?

michaeljklein (approved):
LGTM

Small comment r.e. possible follow-up issue

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
jfecher: Yes, I'll make one

test_programs/execution_success/array_dedup_regression/src/main.nr
```diff
@@ -0,0 +1,21 @@
...
+        a2[x + i] = 128;
+        println(a2);
+
+        if i != 0 {
+            assert(a2[x + i - 1] != 128);
+        }
+    }
+
+    // Can't use `== [1, 2, 3, 4, 5]` here, that make_array may get
+    // deduplicated to equal a1 in the bugged version
```
vezenovm, line 15: This still occurs in this PR?

vezenovm (approved):

test_programs/execution_success/array_dedup_regression/src/main.nr
jfecher: No but in some versions of the buggy version it'd get deduplicated to the mutated version. To avoid confusion in that version I just checked each index manually.

Issue #6670, opened 2024-12-02
Add some CI to test that `nargo::ops::transform_program` is idempotent.


TomAFrench:
This has caused an issue where invalid witnesses are generated in #6667 so we should have regression testing for this.

PR #6689, opened 2024-12-03
feat(tooling): Skip program transformation when loaded from cache

### LINKED ISSUE #6669: Skip running the transformation step on cached programs
This is unnecessary at best (if transformation is idempotent) and causes correctness issues at worst (if not idempotent) so we should aim to avoid it.

aakoshh:
# Description

## Problem\*

Resolves #6669 

## Summary\*

`compile_cmd::compile_programs` now detects the case when `nargo::ops::compile::compile_program` has the compiled artefact loaded from cache, rather than recompiled from scratch. When it's loaded from cache, it is assumed to have already gone through the rest of what `compile_cmd::compile_programs` does, which is to apply `nargo::ops::transform_program` on it, which includes ACVM optimisations and setting the target expression width. If the current target expression width is the same as the one the package was compiled with, we can skip doing these steps again. 

That eliminates the risk of running an non-idempotent optimisation that changes our circuit from what has been persisted before. If somebody were to delete the artefacts and compile from scratch with the same nargo version, then again the transformations were only applied once, which should deterministically yield the same result. 

### What about compiled contracts?

Curiously `compile_cmd::compile_contracts` does not look for persisted artefacts, nor does it apply `nargo::ops::check_program` on the compiled circuits. This is consistent with how [noir_wasm does it](https://github.com/noir-lang/noir/blob/304403f24e2a15b57bb054c4402a8d7f8d275668/compiler/wasm/src/compile.rs#L182-L219), but I'm not sure why these steps aren't done, since under the hood there are the same types involved in both cases. 

### Added `Hash` to a lot of Brillig types

To support the comparison between cached and compiled artefact I added `#[derive(Hash)]` to a lot of types that comprise a `CompiledProgram`. Most of these implemented `Eq`, so that would have been another option. I didn't want to add a `bool` flag to the return value of `noirc_driver::compile_main` to not have a breaking change in the API, although in retrospect I could have added a new method instead. Given that the types seem to implement `Eq` anyway, I don't see the harm in having `Hash` as well, and it could be useful in implementing #6670 

NB this hash is different than the one we use to decide if we can reuse the artefact, which is based roughly on the AST, and doesn't include the transformations. We don't use the hash to detect whether the transformations result in the same hash, only to decide whether the cached version or a fresh one was returned.

### What about Wasm?

The Wasm version doesn't use caching, so it will only apply the transformations once, never repeatedly.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2789,6 +2789,7 @@ dependencies = [
  "dirs",
  "file-lock",
  "fm",
+ "fxhash",
  "iai",
  "iter-extended",
  "lazy_static",
--- a/acvm-repo/acir/src/circuit/brillig.rs
+++ b/acvm-repo/acir/src/circuit/brillig.rs
@@ -5,7 +5,7 @@ use serde::{Deserialize, Serialize};
 
 /// Inputs for the Brillig VM. These are the initial inputs
 /// that the Brillig VM will use to start.
-#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Debug)]
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Debug, Hash)]
 pub enum BrilligInputs<F> {
     Single(Expression<F>),
     Array(Vec<Expression<F>>),
@@ -14,7 +14,7 @@ pub enum BrilligInputs<F> {
 
 /// Outputs for the Brillig VM. Once the VM has completed
 /// execution, this will be the object that is returned.
-#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Debug)]
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Debug, Hash)]
 pub enum BrilligOutputs {
     Simple(Witness),
     Array(Vec<Witness>),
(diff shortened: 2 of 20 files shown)


TomAFrench:
> Curiously compile_cmd::compile_contracts does not look for persisted artefacts, nor does it apply nargo::ops::check_program on the compiled circuits.

This is mostly down to keeping things simple compared to hashing all the various contract functions, etc.

aakoshh:
> In the new hash you've added however we're including the debug info which includes the full source which was used to compile the program so this hash will change due to comment changes.

Maybe I'm missing something, but this will only happen if the `CompiledProgram` returned from `compile_no_check` is _not_ `cached_program`. If we just return `cached_program` as-is, then we're going to be returning the cached debug info as well, and the hash will not change. If you change a comment in the program, and it passes the "inner" hash check because no meaningful change has been done, then we still return `cached_program`, not the one that contains the debug info of the new comment.

> We can then have situations where the "inner" hash check will pass but the "outer" one will not 

If the "inner" has check passed we return the `cached_program` without modification, so the "outer" check _must_ pass 🤔

aakoshh:
What think I'm doing is basically this:

```rust
fn compile_package(package) {
    let cached_program = load_cached_program(package);
    let program = compile_program(package, cached_program);
    if program == cached_program {
      return;
    }
    optimize(program);
    save(package, program);
}
```

TomAFrench:
>Maybe I'm missing something, but this will only happen if the CompiledProgram returned from compile_no_check is not cached_program.

Yeah, this is a brainfart on my end. This shouldn't be an issue.

TomAFrench (approved):

Issue #6668, opened 2024-12-02
Fix `MergeExpressionsOptimizer` so that it fully optimized programs


TomAFrench:
Currently the `MergeExpressionsOptimizer` needs multiple iterations to fully optimize a program. We should update the tranformation step to run this until it converges or make necessary some logic changes to fully optimize on the first iteration

TomAFrench:
FYI @guipublic on this along with parent issue.

Issue #6667, opened 2024-12-02
Nargo further optimizes cached programs after loading them.


TomAFrench:
In this snippet we're either compiling the program from scratch or returning a cached artifact. This will then be unconditionally fed into `nargo::ops::transform`.

https://github.com/noir-lang/noir/blob/4ff308128755c95b4d461bbcb7e3a49f16145585/tooling/nargo_cli/src/cli/compile_cmd.rs#L182-L199

This is fine when `nargo::ops::transform` is a noop on a cached artifact but it seems that `MergeOptimizer` is not fully optimizing programs on the first instance so when we run the cached program through it again it can further optimize the program resulting in a different bytecode being executed compared to what's saved on disk.

PR #6695, opened 2024-12-04
fix: Make `nargo::ops::transform_program` idempotent

aakoshh:
# Description

## Problem\*

Followup for https://github.com/noir-lang/noir/pull/6694 

## Summary\*

Makes `nargo::ops::transform_program` idempotent by adding two loops, both up to 3 passes, exiting early if the opcode hash doesn't change:
* an outer loop in `acvm::compiler::compile` that includes `optimize_internal` and `transform_internal` 
* an inner loop in `acvm::compiler::transformers::transform_internal`

## Additional Context

Here's the journey I went through investigating the source of changes across optimisation runs.

### Ever increasing `current_witness_index`

On the `slice_loop` example the `Circuit::current_witness_index` field increased by 2 after each pass. This seems to be because:
1. It is increased by the number of intermediate variables created [here](https://github.com/noir-lang/noir/blob/dc96de7c7a3d3c325b955edb277534653204cc5a/acvm-repo/acvm/src/compiler/transformers/mod.rs#L83) even though the mutable reference is [also increased](https://github.com/noir-lang/noir/blob/dc96de7c7a3d3c325b955edb277534653204cc5a/acvm-repo/acvm/src/compiler/transformers/csat.rs#L279-L281), hence the double increment
2. The intermediate variables are (in this case entirely) [eliminated](https://github.com/noir-lang/noir/blob/dc96de7c7a3d3c325b955edb277534653204cc5a/acvm-repo/acvm/src/compiler/transformers/mod.rs#L162-L163) later, but the current index isn't adjusted accordingly. 

The PR adds a brute force step to visit each opcode and collect the remaining witnesses to set the next one correctly. 

Alternatively we could implement `PartialEq` for `Circuit` in a way that it ignores `current_witness_index`. 

After this fix we have the following tests still flagging the transformation as non-idempotent:
* `7_function`
* `conditional_1`
* `fold_fibonacci`
* `hashmap`
* `regression_5252`
* `regression_6451`
* `sha256`
* `sha256_regression`
* `sha256_var_size_regression`
* `sha256_var_witness_const_regression`
* `slices`
* `to_be_bytes`

### Multiple passes

For the above I found that adding the following to the test makes all of them pass:

```rust
                for _ in 0..2 {
                    program_0 = nargo::ops::transform_program(program_0, width);
                }
```

So two extra full passes to stabilise the optimisation currently. An example where 1 extra pass is not enough was `conditional_1`. Ideally we would find which part of the transformation needs to be repeated to avoid having to do a full pass; our initial hypothesis was to look at #6668 

I had to add a loop around `transform_internal` allowing 3 passes before the tests indicated more stability than before. The expectation was that we only need to loop around `MergeExpressionOptimizer`, but that doesn't seem to be true. Even after this, the following two programs fail:

* `7_function`
* `slices`
The difference in both cases is the removal of a range check:
```console
assertion failed: `(left == right)`: optimization not idempotent for test program 'slices''
  left: `"func 0\ncurrent witness index : 617\nprivate parameters indices : [0]\npublic parameters indices : [1]\nreturn value indices : []\nEXPR [ (1, _1) -10 ]\nEXPR [ (-1, _0) (-1, _2) 10 ]\nBRILLIG CALL func 0: in..."` (truncated)
 right: `"func 0\ncurrent witness index : 617\nprivate parameters indices : [0]\npublic parameters indices : [1]\nreturn value indices : []\nEXPR [ (1, _1) -10 ]\nEXPR [ (-1, _0) (-1, _2) 10 ]\nBRILLIG CALL func 0: in..."` (truncated)

Differences (-left|+right):
 BRILLIG CALL func 0: inputs: [Single(Expression { mul_terms: [], linear_combinations: [(1, Witness(13))], q_c: 0 })], outputs: [Simple(Witness(14))]
 EXPR [ (1, _13, _14) (1, _15) -1 ]
 EXPR [ (1, _13, _15) 0 ]
 EXPR [ (1, _7) (-1, _16) 5 ]
-BLACKBOX::RANGE [(16)] [ ]
 EXPR [ (-1, _17) 0 ]
 EXPR [ (1, _0, _4) (-1, _489) 0 ]
 EXPR [ (-10, _4, _8) (-1, _490) 0 ]
 EXPR [ (10, _8) (-1, _18) (1, _489) (1, _490) 0 ]
```

For these to stabilise had to move the loop to be around the entire transformation that includes the initial backend agnostic `optimize_internal` as well:

```rust
/// Applies [`ProofSystemCompiler`][crate::ProofSystemCompiler] specific optimizations to a [`Circuit`].
pub fn compile<F: AcirField>(
    acir: Circuit<F>,
    expression_width: ExpressionWidth,
) -> (Circuit<F>, AcirTransformationMap) {
    let mut pass = 0;
    let mut prev_acir = acir;

    let (mut acir, acir_opcode_positions) = loop {
        let (acir, acir_opcode_positions) = optimize_internal(prev_acir);
        let (acir, acir_opcode_positions) =
            transform_internal(acir, expression_width, acir_opcode_positions);

        if pass == 2 {
            break (acir, acir_opcode_positions);
        }
        pass += 1;
        prev_acir = acir;
    };

    let transformation_map = AcirTransformationMap::new(&acir_opcode_positions);
    acir.assert_messages = transform_assert_messages(acir.assert_messages, &transformation_map);

    (acir, transformation_map)
}
```

The final solution has an inner and an outer loop, treating both as black boxes.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/acvm-repo/acvm/src/compiler/transformers/mod.rs
+++ b/acvm-repo/acvm/src/compiler/transformers/mod.rs
@@ -1,5 +1,3 @@
-use std::collections::BTreeSet;
-
 use acir::{
     circuit::{
         self,
@@ -163,13 +161,12 @@ pub(super) fn transform_internal<F: AcirField>(
         ..acir
     };
     let mut merge_optimizer = MergeExpressionsOptimizer::new();
+
     let (opcodes, new_acir_opcode_positions) =
         merge_optimizer.eliminate_intermediate_variable(&acir, new_acir_opcode_positions);
 
     // n.b. if we do not update current_witness_index after the eliminate_intermediate_variable pass, the real index could be less.
     let mut acir = Circuit {
-        current_witness_index,
-        expression_width,
         opcodes,
         // The optimizer does not add new public inputs
         ..acir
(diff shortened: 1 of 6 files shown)


TomAFrench (commented):
> So we need two extra passes to stabilise the optimisation currently. An example were 1 extra pass is not enough was conditional_1.

Performing the full transformation step again is a little overkill. The only non-idempotent optimization is #6668 afaik so if anything we should run multiple instances of this but avoid the rest.

acvm-repo/acvm/src/compiler/transformers/mod.rs
```diff
@@ -160,13 +165,267 @@ pub(super) fn transform_internal<F: AcirField>(
...
         ..acir
     };
+
+    // After the elimination of intermediate variables the `current_witness_index` is potentially higher than it needs to be,
+    // which would cause gaps if we ran the optimization a second time, making it look like new variables were added.
+    // Here we figure out what is the final state of witnesses by visiting each opcode.
+    let witnesses = WitnessCollector::collect_from_circuit(&acir);
+    if let Some(max_witness) = witnesses.last() {
+        acir.current_witness_index = max_witness.0;
+    }
```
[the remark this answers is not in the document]
aakoshh: You mean that the collector should be a visitor that only keep the last? Sure, no problem.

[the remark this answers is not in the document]
TomAFrench: Yeah, I think we can keep traversal here but we can avoid some memory usage.

aakoshh:
> The only non-idempotent optimization is https://github.com/noir-lang/noir/issues/6668 afaik so if anything we should run multiple instances of this but avoid the rest.

I pushed a commit which allows multiple passes with the `MergeExpressionsOptimizer`, but despite this another full cycle still changes the results 🤔

aakoshh:
For example on the following example it looks like the MEO only runs once, which suggests there is another source of unstable output:
```shell
cargo test -p nargo_cli --bins -- test_transform_program_is_idempotent to_be_bytes
```

asterite (approved):
Looks great!

acvm-repo/acvm/src/compiler/transformers/mod.rs
```diff
@@ -12,22 +17,28 @@ pub use csat::MIN_EXPRESSION_WIDTH;
...
 use super::{
     optimizers::MergeExpressionsOptimizer, transform_assert_messages, AcirTransformationMap,
+    MAX_OPTIMIZER_PASSES,
 };
 
 /// Applies [`ProofSystemCompiler`][crate::ProofSystemCompiler] specific optimizations to a [`Circuit`].
-pub fn transform<F: AcirField>(
-    acir: Circuit<F>,
+///
+/// TODO: Can this be removed?
```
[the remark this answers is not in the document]
aakoshh: Done, then undone :) I didn't notice first the `pub use`, maybe somebody uses `acvm-repo` as a library 🤷

tooling/nargo_cli/src/cli/compile_cmd.rs
```diff
@@ -413,11 +413,22 @@ mod tests {
 
                 let width = get_target_width(package.expression_width, None);
 
+                // Figure out if there is an upper bound to repeating the entire optimization step after which the results are stable:
```
vezenovm, line 416: Could you open up an issue for this

acvm-repo/acvm/src/compiler/transformers/mod.rs
```diff
@@ -79,8 +128,6 @@ pub(super) fn transform_internal<F: AcirField>(
                     &mut next_witness_index,
                 );
 
-                // Update next_witness counter
-                next_witness_index += (intermediate_variables.len() - len) as u32;
```
vezenovm, line 83: Could you elaborate why this was removed again?

aakoshh: Yes, I added it in the PR description:

> It is increased by the number of intermediate variables created [here](https://github.com/noir-lang/noir/blob/dc96de7c7a3d3c325b955edb277534653204cc5a/acvm-repo/acvm/src/compiler/transformers/mod.rs#L83) even though the mutable reference is [also increased](https://github.com/noir-lang/noir/blob/dc96de7c7a3d3c325b955edb277534653204cc5a/acvm-repo/acvm/src/compiler/transformers/csat.rs#L279-L281), hence the double increment

I might be missing the significance of this call here, but it looks like it's always incremented in tandem with any the actual variable being added. I guess this would have been needed if instead of passing a mutable reference, the method that adds the intermediate variables would increment their copy.

tooling/nargo_cli/src/cli/compile_cmd.rs
aakoshh: Maybe this comment is misleading: this was my way to figure out what value I could use for `MAX_OPTIMIZATION_PASSES`, so in that sense this is done, I just left the code here commented out while I was going back and forth between running this vs placing the loop in a smaller scope. I guess the value has been established to be 3 passes. I can remove this comment to reduce confusion, but I'm not sure what further issue to create for it.

There are sub-issues under https://github.com/noir-lang/noir/issues/6667 

Are you suggesting that we should open an issue to investigate _why_ we need 3 passes?

vezenovm: > I guess the value has been established to be 3 passes. I can remove this comment to reduce confusion, but I'm not sure what further issue to create for it.

No I was just confused by the comment and thought you wanted to investigate further and just picked 3 as an arbitrary number for now. Let's remove the comment then.

vezenovm (approved):

PR #6703, opened 2024-12-04
fix!: several format string fixes and improvements

### LINKED ISSUE #4642: Format and raw strings ignore escape characters
### Aim

Attempted to use escape characters in format or raw strings

### Expected Behavior

1. reject invalid escape character `\999..`
2. convert escaped characters to expected values, e.g. `\n` becoming a newline
3. the same behavior between the different string types, besides exceptions for formatting and `#`'s

### Bug

Format strings `f".."` and raw strings `r".."` currently ignore `\n`, and other escaped characters.

For example:
```rust
    print(f"hello world!\n \0 \99999999");
    print(r#"hello world!\n \0 \99999999"#);
```

`nargo execute` gives:

```bash
hello world!\n \0 \99999999[name_of_program] Circuit witness successfully solved
```

### To Reproduce

1. put one of the `print`'s above into a test file
2. `nargo execute`

### Project Impact

None

### Impact Context

_No response_

### Workaround

Yes

### Workaround Description

use regular strings for escaped characters and format strings for formatting, like so:

```rust
  print(f"hi there, {name}!");
  print("\n");
```

### Additional Context

_No response_

### Installation Method

Binary (`noirup` default)

### Nargo Version

_No response_

### NoirJS Version

0.24.0+722dc969e8b09e5a6fd56b094b6939e9330233f6

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #4642

## Summary

This PR includes several fixes and improvements to format strings:
1. Format strings (`f"..."`), as tokens, are now represented as fragments which can either be strings or interpolated pieces. The interpolated pieces also have a Span to know where they are. The advantage of this is that we don't need to use a regex to find the interpolations and we can provide better error messages when, for example, the interpolated variable doesn't exist. That said, the print oracle gets the raw string and needs to do the interpolation, and that still uses regex (but at least compile-time code won't need to do regex matching each time it finds a fmtstr).
2. We now disallow non-identifiers in interpolations. This actually uncovered a few "bugs" in the standard library where we were using interpolations in a wrong way.
3. Escape sequences like `\n` and `\t` are now allowed inside format strings.
4. We also now allow escaping `{` and `}` and we do that just like in Rust: `{{` to mean `{`, `}}` to mean `}`.
5. Like in Rust, if you write `}` but there was no single `{` before it (we were not in an interpolation) you get an error that says you must use `}}` to write `}`.

Because `{{` now means to write `{`, I had to slightly change the print oracle to understand this.

## Additional Context

I'm marking this PR as a breaking change because it could break someone's code if they had incorrect interpolations, or if they had `\n` inside a format string and wanted to actually print "\\" and "n" separately.

I left LSP features as a follow-up PR, mainly because I realized autocompletion doesn't automatically kick in when you are typing inside a string literal, so implementing completion there doesn't seem very useful. What we could do is at least implement "hover" in these pieces (go-to-definition already works).

A question: do print oracles exist somewhere else? For example in a web context (what I'm saying might not make sense, sorry). I'm just wondering if the print oracle logic of handling interpolations must be updated somewhere else too.

## Documentation

This PR might need to document format string escape sequences... but I couldn't find this explained for strings so maybe all of that could be done in one go in a separate PR (if really needed now).

Check one:

# PR Checklist


jfecher:
> However, if you write f"{foo.bar}" then I think that should error but it doesn't (you get "{foo.bar}" printed) so given that we don't handle such errors I didn't want to handle any error at all.

We should definitely error here eventually, it's quite an annoyance while writing format strings

jfecher (approved):

asterite:
Awesome! I'll tackle that next, it should be relatively easy to do.

I also noticed that if you write `f"Hello {x}"` and `x` gives an error (say, it doesn't exist) then the error is on the entire format string. I'll try to improve that too (in yet another PR).

asterite:
Oooh... a test is failing because it has this:

```noir
        let s1 = f"x is {x}, fake interpolation: \{y}, y is {y}";
```

I guess `\{` was used to escape that `{` (though that `\` is still printed). In Rust to do that you write `{{`. Which way do we want to use?

jfecher:
Let's stick with Rust in this case

aakoshh (approved):
LGTM 👍

compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs
```diff
@@ -232,10 +233,26 @@ impl<'a> FunctionContext<'a> {
...
             }
             ast::Literal::Str(string) => Ok(self.codegen_string(string)),
-            ast::Literal::FmtStr(string, number_of_fields, fields) => {
+            ast::Literal::FmtStr(fragments, number_of_fields, fields) => {
+                let mut string = String::new();
+                for fragment in fragments {
+                    match fragment {
+                        FmtStrFragment::String(value) => {
+                            // Escape curly braces in non-interpolations
+                            let value = value.replace('{', "{{").replace('}', "}}");
```
asterite, line 242: I'm realizing that for codegen purposes we could just escape `{` as `\{`, though maybe for consistency with everything else it's fine if we use the same approach as the Noir syntax.

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
```diff
@@ -665,13 +664,10 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
                         } else {
                             result.push_str(&value.display(self.elaborator.interner).to_string());
                         }
+                    } else {
+                        // TODO: should we panic here?
```
asterite, line 668: TODO: should we panic if we are interpreting a fmtstr, get into an interpolation but we can't find the value to interpolate?

asterite:
Asking for a re-review because this PR now does a lot more than what I originally planned.

compiler/noirc_printable_type/src/lib.rs
```diff
@@ -282,11 +290,17 @@ impl<F: AcirField> std::fmt::Display for PrintableValueDisplay<F> {
                 let mut display_iter = values.iter();
                 let re = Regex::new(r"\{([a-zA-Z0-9_]+)\}").map_err(|_| std::fmt::Error)?;
```
asterite, line 291: This could also be done by using a regex that matches `{...}` but not `{{...}}` but I couldn't figure out how to do it so instead I check if there's `{` before each match.

asterite: Hmm... I'm realizing that, like this, this program doesn't work as expected:

```rust
fn main() {
    let hello_world = 1234;
    let x = f"{{{hello_world}}}";
    println(x);
}
```

It prints "{{hello_world}}" instead of interpolating it, because there's "{" before "{hello_world}". I'm not sure how to solve this... maybe if there's another "{" before it we do need to interpolate it (but not if there's yet one more "{" before it). I'll try to fix it.

asterite: It worked!

compiler/noirc_printable_type/src/lib.rs
```diff
@@ -282,11 +307,17 @@ impl<F: AcirField> std::fmt::Display for PrintableValueDisplay<F> {
...
```
asterite, line 308: Maybe this logic could be done with regex, just going char by char and checking if we get `{` but not `{` after it... but for now I left it as is, maybe it's fine.

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
aakoshh: I see that you one expression per fragment [here](https://github.com/noir-lang/noir/pull/6703/files#diff-3c108dc343290b597e508f0b0cc501f5a848f3a644aa08a406a9bfb1a2305f73R273), but if one of the fragments was undeclared this is [skipped](https://github.com/noir-lang/noir/pull/6703/files#diff-3c108dc343290b597e508f0b0cc501f5a848f3a644aa08a406a9bfb1a2305f73R264), so the length of them can differ. I'm not sure such a string will be interpreted, but if it can, we shouldn't panic.

aakoshh: OTOH maybe the two independent vectors are not the right way to represent a `HirLiteral::FmtString`; it could be a single `Vec<(FmtStrFragment, Option<ExprId>)>` instead, because currently it would conceptually cause fragments to be interpolated into the wrong spot if one is missing.

asterite: That makes sense! I'll do this change.

asterite: Hm, or maybe not... if we can't type-check an interpolation we could simply push the fragment as-is without interpolation. I don't know if that part of the code will be reached, but just in case... all of this probably doesn't matter much as we are already in an error state 🤷

aakoshh: yeah so currently as I understand it if we have, say `f"foo={foo}, bar={non_existing}, baz={baz}"`, then it would become something like 
```rust
HirLiteral::FmtStr {
fragments: vec![
  S("foo="), 
  I("{foo}"), 
  S(", bar="), 
  I("{non_existing}"), 
  S(", baz="), 
  I("{baz}")
], 
captures:  vec![
  foo_id, 
  baz_id
]}
```
and then you go over the `fragments` and if it's an `S` you push as-is, if it's an `I` then it pops the next value from `captures`. This will result in: `"foo={eval(foo_id)}, bar={eval(baz_id)}, baz="`. 

What I'm suggesting is to represent it like this:
```rust
HirLiteral::FmtStr(
vec![
  S("foo="), 
  I("{foo}", Some(foo_id)), 
  S(", bar="), 
  I("{non_existing}", None), 
  S(", baz="), 
  I("{baz}", Somę(baz_id))
])
```

And then as you go through them, you see that the 2nd `I` has no corresponding expression to evaluate, so you can display it as `"foo={eval(foo_id)}, bar={{non_existing}}, baz={eval(baz_id)}"` or `bar=, baz={eval(...` or whatever you fancy.

compiler/noirc_printable_type/src/lib.rs
```diff
@@ -253,8 +253,15 @@ fn to_string<F: AcirField>(value: &PrintableValue<F>, typ: &PrintableType) -> Op
...
-fn replace_all<E>(
+/// Assumes `re` is a regex that matches interpolation sequences like `{...}`.
+/// Replaces all matches by invoking `replacement`, but only if a match isn't preceded by an odd
+/// number of '{'.
+/// For example, if we have `{{x}}` then there's one `{` before the interpolation, and we can
+/// see it's actually not an interpolation.
+/// However, if we have `{{{x}}}` there are two `{` before the interpolation, so the third
+/// `{` actually starts an interpolation.
+/// With four (`{{{{x}}}}`) it's again even and it's not an interpolation, etc.
+fn replace_all_interpolations<E>(
```
aakoshh, line 264: So previously this was a general `replace_all` function that could replace all captures in any regex, but now it's specific to format strings because it knows that it has to count the number of curly braces. 

I would suggest to go either of two ways:
* Remove the `re` parameter and create the regex looking for interpolations inside the function. I don't see a reason why it should _assume_ that it's given a regex that does the right thing. It's called from one place anyway, so there is no need for that regex to be created there, when it's the only one a `replace_all_interpolations` can probably work with anyway. 
* Leave it as `replace_all` and move the curly brace counting logic into the `replacement` function implementation. the `Captures` has access to the `haystack`, so I think it can do the same there as well, except you have to return an actual string to "leave" an odd number of stuff in place.

Or do both, to leave the generic `replace_all` around and also have an interpolation specific version.

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
asterite: I like this! I'll do it.

compiler/noirc_printable_type/src/lib.rs
asterite: I made it work without regex. The code is likely a bit simpler and also faster (maybe?), and I added a few tests too.

aakoshh: You da man!

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
asterite: Hm, I think for this I'll need to introduce a new type that's that `S` or `I` but where `I` has the string fragment plus `Option<ExprId>`. That makes the code a bit longer and complex, but I'm still thinking this is for an error case that won't go to the monomorphizer, etc., so handling all of this with extra logic might not be worth it.

aakoshh: Another way of looking at it is that the interpreter already does handle it, but in a weird way where captures appear in the wrong position 🤷 So we have extra logic that might be illogical :D

jfecher: I think this would be good to test for to see if we can actually trigger this behavior in real comptime code. The variable not declared error comes from elaborating the format string, so it should be sufficient to test if we can do this by calling a comptime function with a format string in it.

As far as the original TODO goes, we should definitely error. If there was an error evaluating the string we should return an error instead of a partial string or something. We may still need the partial string approach @aakoshh is talking about for the elaborator though depending on if the error mentioned above stops the interpreter from interpreting the bad string or not though.

asterite: Huh, interesting. This code:

```noir
fn main() {
    comptime {
        let x = 1;
        let s = f"hello {x} {y}";
        println(s);
    }
}
```

gives an error, but before erroring it prints this:

```
hello 1
```

If we change it to `f"hello {y} {x}" it prints the same thing, because it ignores which one goes where (as you noticed).

But... I still don't know if it's worth doing this more correctly. It's also not clear what should be printed here (do we print "{y}" or do we show "ERROR" or somehting like that). Also, this code was working like this before this PR, so...

Maybe others from @noir-lang/core could give their opinion here.

asterite: As usual, I wrote the comment before seeing Jake replied...

aakoshh: Okay, so when you `execute` the code above it gets compiled first, and during compilation the interpreter evaluates the code and prints the incorrect string. Then the compiler prints the errors from the elaborator, and finally the `execute` command errors because there were errors. 

If we panicked in the interpreter the entire program would crash before the user could see what was wrong:

```console
❯ cargo run -q -p nargo_cli -- --program-dir . execute
Message:  oh noes!
Location: compiler/noirc_frontend/src/hir/comptime/interpreter.rs:678

```

I'd say printing the correct string would still be helpful, otherwise it might confuse come macro developer. Although the error is pretty clear so it probably doesn't matter that much:

```console
❯ cargo run -q -p nargo_cli -- --program-dir . execute
hello y=1 x=
error: cannot find `y` in this scope
  ┌─ src/main.nr:4:17
  │
4 │         let s = f"hello y={y} x={x}";
  │                 -------------------- not found in this scope
  │

```

I tried returning an `Err` from `evaluate_format_string` but it did not improve the above, mostly because I used some random `InterpreterError`. I suppose one that would stay "cannot evaluate format string" could be added. It depends on whether we think there is a point in trying to carry on with the _rest_ of the program after producing a partially invalid string. For println it's probably ok, but if they use it for something it might just cause further problems.

jfecher: > If we panicked in the interpreter the entire program would crash before the user could see what was wrong:
> ...
> I tried returning an Err from evaluate_format_string but it did not improve the above, mostly because I used some random InterpreterError.

We should definitely return an Err when evaluating a bad format string in the interpreter. It's what we do for other errors, even ones that could already be reported by the elaborator, such as type errors. Worst case we may get 2 similar errors printed for the same issue but it is worth it to stop code after the format string from being executed.

This is very forced but imagine a code with side-effects at compile-time:

```
comptime fn foo(name, balance, other_number) {
    let s = f"{{ name: {name}, balance: {balance}, other: {other_number} }}";
    process_transaction(s);
}
```
If this gets formatted using `other_number` in place of `balance` that could be bad™️ for users. Not that they should use format strings in their transaction logic either way.

asterite: Makes sense. Then I'll change the code to how Akosh suggested it.

asterite: Actually, I'll just return `Err` in the case where we can't find a value to interpolate (easier).

aakoshh: I agree, this is what I meant by
> but if they use it for something it might just cause further problems.

asterite: Done!

asterite: We can't do `fragments.len() != captures.len()` because `fragments` also has both the raw string values and the pieces we need to interpolate to.

aakoshh: Right, you would have to count the `Interpolate` types. I guess the only difference would be that this way it might error during the evaluation of one of the arguments, before you run out of captures.

aakoshh (approved):
Looks good!

Savio-Sou:
> but I couldn't find this explained for strings

Format strings are documented in the [fmtstr page](https://noir-lang.org/docs/noir/standard_library/fmtstr) instead of the Strings page.

Lemme create a corresponding Issue.

PR #6709, opened 2024-12-05
feat: show printable byte arrays as byte strings in SSA

asterite:
# Description

## Problem

Doesn't resolve any issue, but might make debugging SSA a bit easier.

## Summary

In Noir string literals are just arrays of bytes. When going from code to SSA we always get byte arrays and it's hard to relate those back to string literals.

In this PR, when printing a `make_array` instruction whose array (or slice) consists of all printable ascii bytes, we now show that as a byte string literal (`b"..."`). I know Noir doesn't have `b"..."` like in Rust, but I used the same notation to make it clear that it's really a byte array.

For example, for this Noir program:

```noir
fn main() {
    let x = 1;
    println(f"X is {x}");
}
```

The final SSA before this PR was:

```
acir(inline) fn main f0 {
  b0():
    v7 = make_array [u8 88, u8 32, u8 105, u8 115, u8 32, u8 123, u8 120, u8 125] : [u8; 8]
    call f1(u1 1, v7, u32 1, Field 1)
    return
}
brillig(inline) fn print_unconstrained f1 {
  b0(v0: u1, v1: [u8; 8], v2: Field, v3: Field):
    v15 = make_array [u8 123, u8 34, u8 107, u8 105, u8 110, u8 100, u8 34, u8 58, u8 34, u8 102, u8 105, u8 101, u8 108, u8 100, u8 34, u8 125] : [u8; 16]
    call v16(v0, v1, v2, v3, v15, u1 1)
    return
}
```

And with this PR:

```
acir(inline) fn main f0 {
  b0():
    v7 = make_array b"X is {x}"
    call f1(u1 1, v7, u32 1, Field 1)
    return
}
brillig(inline) fn print_unconstrained f1 {
  b0(v0: u1, v1: [u8; 8], v2: Field, v3: Field):
    v15 = make_array b"{\"kind\":\"field\"}"
    call v16(v0, v1, v2, v3, v15, u1 1)
    return
}
```

I think the last one is a bit clearer.

This will also help when debugging SSA issues: we can include a couple of `println(...)` around a code segment that doesn't work well, and we can easily find out where those `println` statements ended up being.

## Additional Context

It could happen that one wrote a byte array that happens to have all printable ascii characters. For example:

```noir
fn main() {
    let x: [u8; _] = [65, 66, 67, 123];
    println(x);
}
```

In that case the SSA will be:

```
acir(inline) fn main f0 {
  b0():
    v3 = make_array b"ABC{"
    inc_rc v3
    call f1(u1 1, v3)
    return
}
```

which might be confusing, though in the end if we understand that `b"..."` is just another way to print a byte array then it shouldn't cause any issues while debugging SSA (we could have an env var to disable this special printing, though let's see if we really need it).

Some context around why I did this: I was debugging format strings and I couldn't easily see how a format string ended up being in SSA. With this PR it's clear the format string with the formatting remains as-is.

By the way, I think Erlang also behaves this way: strings are just byte arrays and Erlang will print them as strings if all the bytes are printable.

## Documentation

Check one:

# PR Checklist

- [] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/parser/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/parser/mod.rs
@@ -467,7 +467,7 @@ impl<'a> Parser<'a> {
                 return Ok(ParsedInstruction::MakeArray { target, elements, typ });
             } else if let Some(string) = self.eat_byte_str()? {
                 let u8 = Type::Numeric(NumericType::Unsigned { bit_size: 8 });
-                let typ = Type::Array(Arc::new(vec![u8.clone()]), string.bytes().count() as u32);
+                let typ = Type::Array(Arc::new(vec![u8.clone()]), string.len() as u32);
                 let elements = string
                     .bytes()
                     .map(|byte| ParsedValue::NumericConstant {
--- a/compiler/noirc_evaluator/src/ssa/parser/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/parser/mod.rs
@@ -4,7 +4,10 @@ use std::{
 };
 
 use super::{
-    ir::{instruction::BinaryOp, types::Type},
+    ir::{
+        instruction::BinaryOp,
+        types::{NumericType, Type},
+    },
     Ssa,
 };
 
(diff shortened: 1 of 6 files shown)


compiler/noirc_evaluator/src/ssa/parser/mod.rs
```diff
@@ -796,6 +826,18 @@ impl<'a> Parser<'a> {
         }
     }
 
+    fn eat_byte_str(&mut self) -> ParseResult<Option<String>> {
+        if matches!(self.token.token(), Token::ByteStr(..)) {
+            let token = self.bump()?;
```
michaeljklein, line 831:
```suggestion
            // skip the leading 'b'
            let token = self.bump()?;
```

michaeljklein (approved):
LGTM w/ small comment

compiler/noirc_evaluator/src/ssa/parser/mod.rs
asterite: Oh, this is actually consuming the current token which is the byte string. The other methods (for example `eat_str`) are the same.

PR #6715, opened 2024-12-05
fix: print ssa blocks without recursion

### LINKED ISSUE #6520: nargo crashes while printing the SSA
### Aim

Tried to print the SSA of `fold_2_to_17`.

### Expected Behavior

I expected all SSA passes to be printed all the way down to "After array set optimisation".

### Bug

Nargo crashes while printing the SSA "After Unrolling"

```console
~/Work/aztec/noir/test_programs/execution_success/fold_2_to_17 
❯ cargo run -p nargo_cli --  --program-dir . compile --show-ssa
...
thread '<unknown>' has overflowed its stack
fish: Job 1, 'cargo run -p nargo_cli --  --pr…' terminated by signal SIGABRT (Abort)
```

It won't print a stack trace with the `RUST_BACKTRACE=1` option either.

### To Reproduce

1. `cd test_programs/execution_success/fold_2_to_17`
2. `cargo run -p nargo_cli --  --program-dir . compile --show-ssa`

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 0.38.0 noirc version = 0.38.0+efd763cb77b3709948ef54ae534cbc3f1a8de0e9 (git version hash: a42c643b9016debc99120ec26ebe0fff05596c59, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #6520

## Additional Context

Curiously, `--show-ssa` for `fold_2_to_17` now gives another error, but at least it's not the old stack overflow anymore.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/function.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/function.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeSet, VecDeque};
+use std::collections::BTreeSet;
 
 use iter_extended::vecmap;
 use noirc_frontend::monomorphization::ast::InlineType;
@@ -169,12 +169,11 @@ impl Function {
     /// want to iterate only reachable blocks.
     pub(crate) fn reachable_blocks(&self) -> BTreeSet<BasicBlockId> {
         let mut blocks = BTreeSet::new();
-        let mut deque = VecDeque::new();
-        deque.push_back(self.entry_block);
+        let mut stack = vec![self.entry_block];
 
-        while let Some(block) = deque.pop_front() {
+        while let Some(block) = stack.pop() {
             if blocks.insert(block) {
-                deque.extend(self.dfg[block].successors());
+                stack.extend(self.dfg[block].successors());
             }
         }
         blocks
--- a/compiler/noirc_evaluator/src/ssa/ir/function.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/function.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeSet;
+use std::collections::{BTreeSet, VecDeque};
 
 use iter_extended::vecmap;
 use noirc_frontend::monomorphization::ast::InlineType;
@@ -169,11 +169,12 @@ impl Function {
     /// want to iterate only reachable blocks.
     pub(crate) fn reachable_blocks(&self) -> BTreeSet<BasicBlockId> {
         let mut blocks = BTreeSet::new();
-        let mut stack = vec![self.entry_block];
+        let mut deque = VecDeque::new();
+        deque.push_back(self.entry_block);
 
-        while let Some(block) = stack.pop() {
+        while let Some(block) = deque.pop_front() {
             if blocks.insert(block) {
-                stack.extend(self.dfg[block].successors());
+                deque.extend(self.dfg[block].successors());
             }
         }
         blocks
--- a/compiler/noirc_evaluator/src/ssa/ir/printer.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/printer.rs
@@ -21,23 +21,20 @@ use super::{
 /// Helper function for Function's Display impl to pretty-print the function with the given formatter.
 pub(crate) fn display_function(function: &Function, f: &mut Formatter) -> Result {
     writeln!(f, "{} fn {} {} {{", function.runtime(), function.name(), function.id())?;
-    display_block_with_successors(function, function.entry_block(), &mut HashSet::new(), f)?;
+    display_function_blocks(function, f)?;
     write!(f, "}}")
 }
 
-/// Displays a block followed by all of its successors recursively.
-/// This uses a HashSet to keep track of the visited blocks. Otherwise
-/// there would be infinite recursion for any loops in the IR.
-pub(crate) fn display_block_with_successors(
-    function: &Function,
-    block_id: BasicBlockId,
-    visited: &mut HashSet<BasicBlockId>,
-    f: &mut Formatter,
-) -> Result {
+/// Displays all of a function's blocks by printing the entry block
+/// and its successors, recursively.
+pub(crate) fn display_function_blocks(function: &Function, f: &mut Formatter) -> Result {
     // The block chain to print might be really long so we use a deque instead of recursion
     // to avoid potentially hitting stack overflow (see https://github.com/noir-lang/noir/issues/6520)
     let mut blocks_to_print = VecDeque::new();
-    blocks_to_print.push_back(block_id);
+    blocks_to_print.push_back(function.entry_block());
+
+    // Keep track of the visited blocks. Otherwise there would be infinite recursion for any loops in the IR.
+    let mut visited = HashSet::new();
 
     while let Some(block_id) = blocks_to_print.pop_front() {
         if !visited.insert(block_id) {
--- a/compiler/noirc_evaluator/src/ssa/ir/printer.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/printer.rs
@@ -1,6 +1,6 @@
 //! This file is for pretty-printing the SSA IR in a human-readable form for debugging.
 use std::{
-    collections::HashSet,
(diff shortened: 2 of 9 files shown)


aakoshh (approved):
Awesome! 🕺

compiler/noirc_evaluator/src/ssa/ir/printer.rs
```diff
@@ -21,27 +21,31 @@ use super::{
...
+    let mut blocks_to_print = VecDeque::new();
+    blocks_to_print.push_back(function.entry_block());
 
-    for successor in function.dfg[block_id].successors() {
-        if !visited.contains(&successor) {
-            display_block_with_successors(function, successor, visited, f)?;
+    // Keep track of the visited blocks. Otherwise there would be infinite recursion for any loops in the IR.
+    let mut visited = HashSet::new();
+
+    while let Some(block_id) = blocks_to_print.pop_front() {
```
aakoshh, line 38: Maybe it should be popping from the back to keep the same order as the original?

aakoshh: And insert them in opposite order too 🤷

compiler/noirc_evaluator/src/ssa/ir/printer.rs
```diff
@@ -21,27 +21,31 @@ use super::{
...
+        if !visited.insert(block_id) {
+            continue;
+        };
+        display_block(function, block_id, f)?;
+
+        for successor in function.dfg[block_id].successors() {
+            blocks_to_print.push_back(successor);
```
jfecher, line 45: Instead of repeating this logic, you could also consider using `reachable_blocks` and iterating over that in order. Peak memory usage will be a tad higher since you'd be holding onto all of the blocks at once but it should be fine since that method is already used in other passes.

asterite: Nice, it works! And it's much simpler this way.

aakoshh: Based on this failure:
```
Differences (-left|+right):
     jmpif v5 then: b2, else: b6
   b2():
     v7 = eq v0, u32 2
     jmpif v7 then: b7, else: b3
-  b7():
-    v18 = add v0, u32 1
-    jmp b1(v18)
   b3():
     v9 = eq v0, u32 5
     jmpif v9 then: b5, else: b4
+  b4():
+    v10 = load v1 -> Field
+    v12 = add v10, Field 1
+    store v12 at v1
+    v14 = add v0, u32 1
+    jmp b1(v14)
   b5():
     jmp b6()
   b6():
     v15 = load v1 -> Field
     v17 = eq v15, Field 4
     constrain v15 == Field 4
     return
-  b4():
-    v10 = load v1 -> Field
-    v12 = add v10, Field 1
-    store v12 at v1
-    v14 = add v0, u32 1
-    jmp b1(v14)
+  b7():
+    v18 = add v0, u32 1
+    jmp b1(v18)
 }
```

And the [code](https://github.com/noir-lang/noir/blob/2f46f5a6832067dd1cfbba89c866e6c119cb5438/compiler/noirc_evaluator/src/ssa/ir/function.rs#L174-L176) of `reachable_blocks` the problem is again that it inserts the successors in the opposite order of how the previous depth-first recursion visited them. 

I think the original was nicer in that for example above it visited `b7` and `b3`, in that order, after `b2` which ends with `jmpif v7 then: b7, else: b3`, and in a typical loop the `then` part is the one with the back-edge, and the `else` part will end up going down to the rest of the program.

asterite: I'll try changing how `reachable_blocks` works so that we get the correct order.

jfecher: > And the [code](https://github.com/noir-lang/noir/blob/2f46f5a6832067dd1cfbba89c866e6c119cb5438/compiler/noirc_evaluator/src/ssa/ir/function.rs#L174-L176) of reachable_blocks the problem is again that it inserts pops the successors in the opposite order of how the previous depth-first recursion visited them.

Reachable_blocks is already correct, it returns a BTreeSet so all blocks will be ordered numerically which I think is fine. Passes should generally try to match this ordering and I think it's more clear if we always print in numerical block order, then just try to fix any passes that place blocks in a different order if possible.

jfecher (approved):

compiler/noirc_evaluator/src/ssa/ir/printer.rs
jfecher: In the now failing `ssa::opt::simplify_cfg::test::swap_negated_jmpif_branches_in_brillig` test for example, the else branch is numbered before the then branch for some reason. Seems the `simplify_cfg` pass reorders then/else blocks - we can fix this in a separate PR.

asterite: I changed it so that blocks are pushed on the back, but they are popped from the front. This way blocks are always printed in the order they were created, so `b0`, `b1`, `b2`, `b3`, etc.

I don't know if this is desired by everyone, but I personally like it. When debugging SSA I always wonder "where is this block..." but now that they'll show up in order I know exactly where it'll be (I know I can search, but sometimes it's nearby and it's easier to scroll).

compiler/noirc_evaluator/src/ssa/ir/function.rs
```diff
@@ -169,11 +169,12 @@ impl Function {
     /// want to iterate only reachable blocks.
     pub(crate) fn reachable_blocks(&self) -> BTreeSet<BasicBlockId> {
         let mut blocks = BTreeSet::new();
-        let mut stack = vec![self.entry_block];
+        let mut deque = VecDeque::new();
+        deque.push_back(self.entry_block);
 
-        while let Some(block) = stack.pop() {
+        while let Some(block) = deque.pop_front() {
```
aakoshh, line 175: So this would do a BFS instead of the DFS that the display originally did, however it's all defeated by the fact that `BTreeSet` will order by numeric ID.

compiler/noirc_evaluator/src/ssa/ir/printer.rs
asterite: Oh, I wrote the above comment before seeing your replies @jfecher (I was on an old tab). For some reason now blocks are ordered numerically, where previously they weren't... I don't know if `BTreeSet` also keeps insertion order and that's what it uses when iterating it. But let me know if this is not right, we could probably revert to the duplicated code and not touch `reachable_blocks`.

asterite: Oooh... nevermind, now I understand. I'll revert the code.

compiler/noirc_evaluator/src/ssa/ir/function.rs
asterite: Oh, now I understand. I missed how `BTreeSet` works

aakoshh: I think the insertion-order sensitive version is https://docs.rs/indexmap/latest/indexmap/
I came across it yesterday in the optimiser

compiler/noirc_evaluator/src/ssa/ir/function.rs
```diff
@@ -169,11 +169,12 @@ impl Function {
     /// want to iterate only reachable blocks.
     pub(crate) fn reachable_blocks(&self) -> BTreeSet<BasicBlockId> {
         let mut blocks = BTreeSet::new();
-        let mut stack = vec![self.entry_block];
+        let mut deque = VecDeque::new();
+        deque.push_back(self.entry_block);
```
jfecher: Can you change this back? This code does nothing since we return a BTreeSet which is ordered by block number anyway

jfecher: Was looking at older version of code - nevermind

Issue #6275, opened 2024-10-10
Compilation flag to add


michaeljklein:
### Problem

Some ACIR opcodes have built-in assumptions about their inputs/outputs, e.g.
- Whether a point is on the curve and/or at infinity
- An input/output being within a range (if integral) or of a specific size (if array-like)
- Public key(s) are valid(?)

These assumptions are expected to be checked at _some_ point throughout compilation, but
1. They're exclusively runtime checks
2. They're checked at different locations and/or passes
3. They could be checked earlier in compilation and then invalidated later, e.g. by an SSA pass

### Happy Case

Add a compilation flag to generate runtime code to check such assumptions, perhaps `--pedantic` or `--pedantic-gen`
- This flag would only be used for testing, since it would be expected to have a negative impact on performance (and should be redundant)
- This would be especially useful for fuzzing, where otherwise assumptions can be broken silently

### Workaround

Yes

### Workaround Description

A partial workaround is to manually trace the dependencies of a piece of user code to their ACVM opcodes and add such runtime checks, however:
- This only works for individual and specific pieces of user code
- It would be challenging to keep up to date, especially without feature flags

### Project Impact

Nice-to-have

### Would you like to submit a PR for this Issue?

None

Rumata888:
This would be really nice to protect us from shooting ourselves in the foot. If we implement fuzzing for noir programs, it would also have a benefit as an extended bug oracle. Would be great

PR #6716, opened 2024-12-05
feat: `--pedantic-solving` flag

michaeljklein:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6275

## Summary\*

Adds a compile-time flag `--pedantic-solving` to run extra sanity-check assertions at the opcode level in ACVM.

## Additional Context

1. This is expected to be [useful for fuzzing](https://github.com/noir-lang/noir/issues/6275#issuecomment-2419234239)
2. This will need a follow-up for similar types of assertions

## Documentation\*

Check one:

# PR Checklist\*


tooling/nargo/src/ops/test.rs
```diff
@@ -75,6 +75,7 @@ pub fn run_test<B: BlackBoxFunctionSolver<FieldElement>>(
                     WitnessMap::new(),
                     blackbox_solver,
                     &mut foreign_call_executor,
+                    config.pedantic_solving,
```
TomAFrench, line 78: Seems like we should be pedantic during tests by default right?

acvm-repo/acvm/src/pwg/blackbox/logic.rs
```diff
@@ -49,11 +51,13 @@ fn solve_logic_opcode<F: AcirField>(
...
     result: Witness,
+    pedantic_solving: bool,
     logic_op: impl Fn(F, F) -> F,
 ) -> Result<(), OpcodeResolutionError<F>> {
-    // TODO(https://github.com/noir-lang/noir/issues/5985): re-enable these once we figure out how to combine these with existing
+    // TODO(https://github.com/noir-lang/noir/issues/5985): re-enable these by
+    // default once we figure out how to combine these with existing
     // noirc_frontend/noirc_evaluator overflow error messages
-    let skip_bitsize_checks = true;
+    let skip_bitsize_checks = !pedantic_solving;
```
aakoshh, line 60: I see we have stand-alone functions here, so the `pedantic_solving` can only be a function parameter. Would it make sense to create a `LogicSolver` similar to the `BigintSolver` and make it a config parameter, so as not to complicate the API methods?

aakoshh: Or maybe introduce some kind os `SolverContext` that could contain the `initial_witness` and the `pedantic_solving`; most functions I look at take the `&mut WitnessMap` as parameters, and by wrapping these up we could have a bunch of ambient staff we always pass around.

acvm-repo/acvm/src/pwg/mod.rs
```diff
@@ -588,15 +608,19 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>> ACVM<'a, F, B> {
...
+            return Err(OpcodeResolutionError::AcirMainCallAttempted { opcode_location });
         }
 
-        if is_predicate_false(&self.witness_map, predicate)? {
+        if is_predicate_false(
+            &self.witness_map,
+            predicate,
+            self.pedantic_solving,
+            &opcode_location,
+        )? {
```
aakoshh, line 623: Maybe `is_predicate_false` could be added to that `SolverContext` I suggested, resulting in a `ctx.is_predicate_false(predicate, &opcode_location)` which is a bit less busy.

acvm-repo/acvm/src/pwg/mod.rs
```diff
@@ -737,9 +759,30 @@ fn any_witness_from_expression<F>(expr: &Expression<F>) -> Option<Witness> {
...
+            let pred_value = get_value(pred, witness);
+            if pedantic_solving {
+                pred_value.and_then(|predicate_value| {
+                    if predicate_value.is_zero() {
+                        Ok(true)
+                    } else if predicate_value.is_one() {
+                        Ok(false)
+                    } else {
+                        let opcode_location = *opcode_location;
+                        Err(OpcodeResolutionError::PredicateLargerThanOne {
```
aakoshh, line 776: Is this an `Err` just because `is_predicate_false` already returned a `Result`, while other checks added `panic!` because we don't want to turn them into `Result`? Should this panic too, or is a predicate > 1 a user error?

acvm-repo/blackbox_solver/src/bigint.rs
```diff
@@ -84,8 +96,11 @@ impl BigIntSolver {
             }
             BlackBoxFunc::BigIntMul => lhs * rhs,
             BlackBoxFunc::BigIntDiv => {
+                if pedantic_solving && rhs == BigUint::ZERO {
+                    panic!("--pedantic-solving: BigIntDiv by zero");
```
aakoshh, line 100: Should this be a `BlackBoxResolutionError` instead? Dividing by zero sounds like a user error.

acvm-repo/blackbox_solver/src/bigint.rs
```diff
@@ -141,7 +194,27 @@ impl BigIntSolverWithId {
...
+fn all_allowed_bigint_moduli_are_prime() {
+    use num_prime::nt_funcs::is_prime;
+    use num_prime::Primality;
+
+    for modulus in BigIntSolver::allowed_bigint_moduli() {
+        let modulus = BigUint::from_bytes_le(&modulus);
+        let prime_test_config = None;
+        match is_prime(&modulus, prime_test_config) {
+            Primality::Yes => (),
+            Primality::No => panic!("not all allowed_bigint_moduli are prime"),
```
aakoshh, line 212:
```suggestion
            Primality::No => panic!("not all allowed_bigint_moduli are prime: {modulus}"),
```

acvm-repo/blackbox_solver/src/bigint.rs
```diff
@@ -141,7 +194,27 @@ impl BigIntSolverWithId {
...
+            Primality::Probable(probability) => {
+                if probability < 0.90 {
+                    panic!("not all allowed_bigint_moduli are prime within the allowed probability: {} < 0.90", probability);
```
aakoshh, line 215:
```suggestion
                    panic!("not all allowed_bigint_moduli are prime within the allowed probability: {modulus}, {probability} < 0.90");
```

tooling/acvm_cli/src/cli/execute_cmd.rs
```diff
@@ -34,12 +34,19 @@ pub(crate) struct ExecuteCommand {
     /// Set to print output witness to stdout
     #[clap(long, short, action)]
     print: bool,
+
+    /// Use pedantic ACVM solving, i.e. double-check some black-box function
+    /// assumptions when solving.
+    /// This is disabled by default.
```
aakoshh, line 40: Just wondering why not enable it by default? Is it expected to be on for production runs?

acvm-repo/acvm/src/pwg/mod.rs
michaeljklein: This is an error because I'm not sure whether it can occur and want to include the location so users can debug. It seems plausible that this could occur in low-level code or in the future if we add more low-level optimization helpers like `as_witness`.

tooling/acvm_cli/src/cli/execute_cmd.rs
michaeljklein: It's expected to decrease performance, especially if we add additional checks for e.g. valid public keys.

aakoshh: Okay, so it's something to do once. Performance is why I am not bullish on global-like approaches, they seem to involve some kind of sync primitive.

acvm-repo/acvm/src/pwg/blackbox/logic.rs
michaeljklein: I looked into using `WitnessMap`, but 1) I don't want to modify a native type, and 2) `WitnessMap` is used more often than `pedantic_solving`. I already moved `pedantic_solving` to the blackbox/bigint solver structs, so will look into the remaining cases next.

aakoshh: To be clear I did not want to suggest modifying `WitnessMap` but to create a `struct SolverContext { initial_witness: WistnessMap, pedantic_solver: bool }` which could be used by the numerous functions that take both of these. If something only takes one we could always pass e.g. `self.ctx.initial_witness`.

acvm-repo/acvm/src/pwg/mod.rs
```diff
@@ -737,9 +752,30 @@ fn any_witness_from_expression<F>(expr: &Expression<F>) -> Option<Witness> {
...
+                            opcode_location,
+                            predicate_value,
+                        })
+                    }
+                })
+            } else {
+                pred_value.map(|pred_value| pred_value.is_zero())
+            }
+        }
```
TomAFrench, line 778:
```suggestion
            let pred_value = get_value(pred, witness)?;
            let predicate_is_false = pred_value.is_zero();
            if pedantic_solving {
               // We expect that the predicate should resolve to either 0 or 1.
                if !predicate_is_false && !predicate.is_one() {
                     let opcode_location = *opcode_location;
                        return Err(OpcodeResolutionError::PredicateLargerThanOne {
                            opcode_location,
                            predicate_value,
                        })
                }
            }
            Ok(predicate_is_false)
```

I think we can separate the calculation of the result from the pedantic check.

acvm-repo/acvm/src/pwg/memory_op.rs
```diff
@@ -150,7 +153,10 @@ mod tests {
         block_solver.init(&init, &initial_witness).unwrap();
 
         for op in trace {
-            block_solver.solve_memory_op(&op, &mut initial_witness, &None).unwrap();
+            let pedantic_solving = true;
```
TomAFrench, line 156: pedantic solving is always on in this file.

TomAFrench (commented):
Looks sensible but we seem to be being pedantic all the time for memory ops.

acvm-repo/bn254_blackbox_solver/src/lib.rs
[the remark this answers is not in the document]
TomAFrench: Actually, we want to exclude any points at infinity. The set of assumptions we make are documented in the stdlib.

acvm-repo/blackbox_solver/src/bigint.rs
```diff
@@ -51,6 +65,18 @@ impl BigIntSolver {
...
+        {
+            if !self.is_valid_modulus(modulus) {
+                panic!("--pedantic-solving: bigint_from_bytes: disallowed modulus {:?}", modulus);
+            } else {
+                panic!(
+                    "--pedantic-solving: bigint_from_bytes: inputs.len() > modulus.len() {:?}",
+                    (inputs.len(), modulus.len())
+                );
+            }
+        }
```
aakoshh, line 79:
```suggestion
        if self.pedantic_solving {
            if !self.is_valid_modulus(modulus) {
                panic!("--pedantic-solving: bigint_from_bytes: disallowed modulus {:?}", modulus);
            }
            if inputs.len() > modulus.len() {
                panic!(
                    "--pedantic-solving: bigint_from_bytes: inputs.len() > modulus.len() {:?}",
                    (inputs.len(), modulus.len())
                );
            }
        }
```
When I saw just the inner `if` I thought we would panic either way. I think this is cleaner because:
* We only check `is_valid_modulus` once
* The check on `inputs` vs `modulus` is close to the panic, more obvious

acvm-repo/bn254_blackbox_solver/src/lib.rs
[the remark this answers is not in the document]
michaeljklein: Ok, we're currently checking that all points are on curves for `multi_scalar_mul` and `embedded_curve_add` [here](https://github.com/noir-lang/noir/blob/36bca82c9a25fdb5f1cd657dace11e46e59491c8/acvm-repo/bn254_blackbox_solver/src/embedded_curve_ops.rs#L100)

I've added checks for points at infinity: it looks like we support them for `multi_scalar_mul`?

acvm-repo/acvm/src/pwg/blackbox/logic.rs
michaeljklein: I'd prefer to keep it as-is since there's only ~5 cases left where `pedantic_solving` is passed along with a `WitnessMap` and I'd have to modify a bunch of places where the `WitnessMap` is created/updated.

acvm-repo/acvm/src/pwg/mod.rs
michaeljklein: General response [above](https://github.com/noir-lang/noir/pull/6716#discussion_r1880892264)

michaeljklein:
NOTE: the following test appears to be failing due to a bug that was fixed in https://github.com/noir-lang/noir/pull/6771:

```bash
tests::compile_success_empty::test_embedded_curve_add_simplification::forcebrillig_false_inliner_0_expects
```

TomAFrench:
Can you update our CI to run with pedantic checks on?

acvm-repo/bn254_blackbox_solver/src/embedded_curve_ops.rs
```diff
@@ -75,11 +80,28 @@ pub fn multi_scalar_mul(
 pub fn embedded_curve_add(
     input1: [FieldElement; 3],
     input2: [FieldElement; 3],
+    pedantic_solving: bool,
 ) -> Result<(FieldElement, FieldElement, FieldElement), BlackBoxResolutionError> {
+    if pedantic_solving && input1[2] > FieldElement::one() && input2[2] > FieldElement::one() {
```
TomAFrench, line 85:
```suggestion
    if pedantic_solving && (input1[2] > FieldElement::one() || input2[2] > FieldElement::one()) {
```

examples/prove_and_verify/prove_and_verify.sh
```diff
@@ -3,12 +3,12 @@ set -eu
 
 BACKEND=${BACKEND:-bb}
 
-nargo execute witness
+cargo run execute --pedantic-solving witness
```
TomAFrench, line 6: Why did this change from nargo to cargo?

michaeljklein: Leftover from testing: reverted

TomAFrench (approved):
LGTM once merge conflicts fixed.

PR #6728, opened 2024-12-06
feat: several `nargo test` improvements

### LINKED ISSUE #4489: Restore streaming test output to `nargo test`
We removed the behaviour where tests would be printed to the console as they were started and completed as this would have broken when tests were made to run in parallel in #4484.

This has the downside of when there's a package with many tests such that running the tests take a long time, nargo appears to hang until all of the tests are run and then they're all printed at once. We should improve how we print out the test results so that we can display the test results as they come in.

### LINKED ISSUE #6724: Restrict the number of tests being run in parallel for `nargo test`
We're currently unable to run the tests for https://github.com/noir-lang/noir-bignum due to `nargo test` attempting to run all of the tests in parallel, resulting in it just running out of memory and the process being killed.

As a short term fix, we should restrict the number of threads which `nargo test` will use so that we can run these tests in github CI. We can set the `RAYON_NUM_THREADS` environment variable internally to set the max number of threads.

### LINKED ISSUE #6727: Track the time taken to run test functions in `nargo test` and print beside the test output
Pretty much clear from the title. We currently have no real good way to determine which tests in a package are particularly slow to compile so it would be helpful to have a readout of the time taken for each one.

This would be useful for tracking potential regressions in compilation speed in the noir compiler or when it would be sensible to split up a test into smaller components.

asterite:
# Description

## Problem

Resolves #4489
Resolves #6724
Resolves #6727

## Summary

This PR does a few things:
2. We now use a maximum number of threads to run tests. The default is the system's default (taken from a Rust function) but it can be overwritten with the new `--num-threads` CLI option (let me know if this should be removed, though, probably always parallelizing is never an issue, unlike in Rust)
3. We now compile all packages and show all errors before starting to run tests (previously the first error would abort the run)
4. Before showing how many tests passed and failed in a package, a list of the tests that failed is shown. This is similar to Rust too. I personally find this useful because otherwise to know which failed you have to scroll up and fine the red "fail" ones, which is a bit time-consuming. But let me know and we could revert this.
5. Output is shown by package. That is, first we show all output for a package, then all output for another package, etc. However, we do run all tests in parallel.
6. Output fron Noir's `println` isn't printed immediately. Instead, it's captured in a String and associated to the test. Then, if `--show-output` is given, that output is printed after the test name status. That makes it very easy to see what output each test produced, and output from different tests is never mixed up.
7. For slow tests (those that take more than 30 seconds) we now show the time.

An example of 6:

image

And consider this Noir program:

```noir
#[test]
pub fn one() {
    println("one (1)");
    println("one (2)");
    println("one (3)");
}

#[test]
pub fn two() {
    println("two (1)");
    println("two (2)");
    println("two (3)");
}

#[test]
pub fn three() {
    println("three (1)");
    println("three (2)");
    println("three (3)");
}
```

The new output for `nargo test` is:

image

The old output was this:

image

## Documentation

Check one:

# PR Checklist


tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -77,29 +88,65 @@ pub(crate) fn run(args: TestCommand, config: NargoConfig) -> Result<(), CliError
...
+        )?;
+        tests.extend(package_tests);
+    }
+
+    let num_tests = tests.len();
+
+    let num_threads = std::thread::available_parallelism().ok().map(Into::into).unwrap_or(1);
+
+    let (sender, receiver) = mpsc::channel();
+    let iter = Mutex::new(tests.into_iter());
```
aakoshh, line 112: Maybe consider using [crossbeam channels](https://docs.rs/crossbeam/latest/crossbeam/channel/index.html#sharing-channels) which can be shared between multiple consumers, so there would be no need for an explicit mutex around an iterator. You could even start running tests while they are being collected.

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -77,29 +88,65 @@ pub(crate) fn run(args: TestCommand, config: NargoConfig) -> Result<(), CliError
...
+                    let _ = sender.send((test.name, test.package_name, test_status));
+                })
+                .unwrap();
+        }
+    });
+
+    let mut test_reports_per_package = HashMap::<String, Vec<(String, TestStatus)>>::new();
+
+    // Print results of tests that already dinished
+    for (test_name, package_name, test_status) in receiver.iter().take(num_tests) {
```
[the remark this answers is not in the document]
asterite: I copied all of this code from libtest-mimic 😊 (which is what the issue suggested)

https://github.com/LukasKalbertodt/libtest-mimic/blob/63e4ec13f15c4adc5b36e3cc0c5f49e19b50f480/src/lib.rs#L492-L551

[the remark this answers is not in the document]
asterite: To answer the question, I'm not sure the senders will go out of the scope because they would be dropped after this loop.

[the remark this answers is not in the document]
aakoshh: Maybe it would be more obviously useful in a setup like this?
```
let (test_tx, test_rx) = crossbeam::channel::unbounded();
let (result_tx, result_rx) = mpsc::channel::unbounded();

for i in 0 .. num_workers { 
  let mut test_rx = test_rx.clone();
  let mut result_tx = result_tx.clone();
  thread::spawn(|| {
    while Ok(test) = test_rx.recv() {
      let result = todo();
      let _ = result_tx.send(result);
    }
  });
}

// Sequential between packages
for package in collect_packages() {
  let package_tests = collect_tests(package);
  let num_tests = package_tests.len();  
  // Parallel inside packages
  for test in package_tests {
    test_tx.send(test);
  }
  // Collect results
  println!("running {num_tests} tests in {package}");
  for _ in 0 .. num_tests {
    if let Ok(result) = result_rx.recv() {
      println!(result);
    }
  }
  println!("tests completed for {package}");
}
```
That is, for each package we know how many results are coming in, so we can display an interim report, before resuming with the next cohort. And it displays results in a streaming fashion as well, as they come. 

The only danger is if a worker crashes while handling a report, we might never get the expected number of results; but other workers are still online, blocked on waiting for new tasks, and the whole thing is deadlocked.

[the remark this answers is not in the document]
aakoshh: > I'm not sure the sender will go out of the scope because it would be dropped after this loop.

I was actually surprised that you did not have to make a clone of the sender, one for each thread. If you do then the receiver should work as long as there are items in the channel, or if it's empty but some sender instances are still around.

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -143,27 +190,38 @@ fn run_tests<S: BlackBoxFunctionSolver<FieldElement> + Default>(
...
+                    package_name_clone.clone(),
+                    compile_options,
+                );
+                display_test_status(
+                    &test_name,
+                    &package_name_clone,
+                    &test_status,
+                    file_manager,
+                    compile_options,
+                );
```
aakoshh, line 240: Maybe I'm misunderstanding what this will do, but I reckon if you put the display into the parallel bit it might get garbled, which is why it would be good to collect the results into a channel, which a single worker displays one by one.

aakoshh: Oh I see there is a lock on the writer. But is it a different instance within each tester?

tooling/nargo_cli/src/cli/test_cmd.rs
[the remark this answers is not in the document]
asterite: > That is, for each package we know how many results are coming in, so we can display an interim report, before resuming with the next cohort. And it displays results in a streaming fashion as well, as they come.

What's the advantage of doing this instead of what there is right now? I wonder if this alternative approach could show the second package output before the first one, if there's a `println` inside Noir, which currently is done outside of this sending/receiving (I'm thinking about improving this but I'm not sure it's worth it).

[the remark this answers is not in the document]
asterite: Or, put another way, maybe it could be done in a separate PR (I'm thinking about making `--show-output` send the output of Noir's `println` to a buffer, then display that buffer at the end... but it requires a larger refactor)

tooling/nargo_cli/src/cli/test_cmd.rs
asterite: Yes, I don't know if the lock locks the underlying stream or not. I checked the docs and it's not clear.

I'm also thinking we could remove the lock as now all the printing happens sequentially... except that Noir's `println` might get in the middle... but then I guess the writer's lock doesn't make sense, it can't possibly lock the underlying stream.

That said, I tried printing stuff from Noir and the output never got garbled, not sure why. Maybe Rust's println is atomic?

tooling/nargo_cli/src/cli/test_cmd.rs
asterite: Maybe this could be done in a separate PR?

tooling/nargo_cli/src/cli/test_cmd.rs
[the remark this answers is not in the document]
aakoshh: I was also just an idea of why `.take(num_tests)` would be useful in that we can keep our workers alive while intermittently print a summary of tests, because we know how many they have to complete before we give them more tasks 🤷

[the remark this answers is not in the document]
asterite: I stopped pushing the remaining code changed I had in mind. If you want, you can try this alternative approach by pushing new commits here, or making a PR on top of this one, mainly because I think you have a clear idea of how to do it and me trying to do it might not result in exactly what you are thinking. Then we can compare the time to run tests to see if it ends up being faster (or, regardless of this, we could keep this alternative approach if it's better in some way).

tooling/nargo_cli/src/cli/test_cmd.rs
aakoshh: I added a test like this and can confirm that the `.lock()` does have the desired effect:
```rust
#[cfg(test)]
mod tests {
    use std::io::Write;
    use std::{thread, time::Duration};
    use termcolor::{ColorChoice, StandardStream};

    #[test]
    fn test_stderr_lock() {
        for i in 0..4 {
            thread::spawn(move || {
                let mut writer = StandardStream::stderr(ColorChoice::Always);
                let mut writer = writer.lock();

                let mut show = |msg| {
                    thread::sleep(Duration::from_millis(10));
                    // println!("{i} {msg}");
                    writeln!(writer, "{i} {msg}").unwrap();
                };

                show("a");
                show("b");
                show("c");
            });
        }
        thread::sleep(Duration::from_millis(100));
    }
}
```

As above it prints:
```
0 a
0 b
0 c
1 a
1 b
1 c
2 a
2 b
2 c
3 a
3 b
3 c
```

If we remove the ` let mut writer = writer.lock();` line it gets garbled:
```
1 3 a
0 a
2 a
a
01 b
3 b
2 b
 b
1 c
3 c
2 c
0 c
```

And if we then use `println!` it gets just mixed (so, no `.lock()`):
```
2 a
0 a
1 a
3 a
2 b
0 b
3 b
1 b
2 c
0 c
3 c
1 c
```

So I'd expect that if we execute tests in parallel then yes, `println!` will not be garbled per-se, but the output of different Noir tests can get intertwined.

The reports at the end will be good, so if the output was collected instead of printed, then the display can happens as you moved it into the tester, rather than the way I suggested with the async collector.

asterite: > but the output of different Noir tests can get intertwined

What I noticed is that the output of different Rust tests can also get intertwined. That was my motivation for not making it work better in Noir 😅

aakoshh: Fair enough :D 
Is that always the case, or only when `--nocapture` is used?

TomAFrench: I think we should do this in a followup PR as there's the same potential for garbling currently, right?

tooling/nargo_cli/src/cli/test_cmd.rs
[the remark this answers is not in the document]
aakoshh: Now that we [confirmed](https://github.com/noir-lang/noir/pull/6728#discussion_r1876110088) that the report is correct I don't have any problem with displaying the results on the go. Maybe it's slower, maybe not, we can try after this PR if we want, please do finish your what you had in mind 🙏

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -77,31 +95,25 @@ pub(crate) fn run(args: TestCommand, config: NargoConfig) -> Result<(), CliError
...
-                    Some(package.name.to_string()),
-                    &args.compile_options,
-                )
-            })
-            .collect::<Result<_, _>>()
-    })?;
-    let test_report: Vec<(String, TestStatus)> = test_reports.into_iter().flatten().collect();
+    let num_threads = args
+        .test_threads
+        .unwrap_or_else(|| std::thread::available_parallelism().ok().map(Into::into).unwrap_or(1));
```
TomAFrench, line 100:
```suggestion
        .unwrap_or_else(|| rayon::max_num_threads());
```

tooling/nargo_cli/src/cli/test_cmd.rs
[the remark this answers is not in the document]
aakoshh: So the only real problem with https://github.com/noir-lang/noir/issues/4489 was that [this lock](https://github.com/noir-lang/noir/blob/c58d69141b54a918cd1675400c00bfd48720f896/tooling/nargo_cli/src/cli/test_cmd.rs#L152-L156) had to happen _inside_ the loop, not before it.

TomAFrench:
Looks good but I'm seeing a number of new test failures in the `run-external-checks` tests (the `private_rollup_lib` tests).

TomAFrench:
Hmm, we'd need to buffer the test results in that case though wouldn't we otherwise we'd get tests from different packages potentially interleaved.

asterite:
I managed to do the above (first compile all packages in parallel, then run tests in parallel for each package) but it's still slower than before. For example if there are many, many packages with a very small number of tests, for example 3, and we have 10 threads, we'll just run 3 in parallel in sequential batches.

Maybe running everything in parallel like before is fine? We could still show a summary at the end for each package separately, which is maybe what matters most. What do you think?

aakoshh:
Couple of questions:
* are you using the same number of threads?
* what if you disable the printing of results after each test?

I think it wasn’t about going faster but to limit the amount if resource usage; the ticket mentioned running out of memory.

TomAFrench:
Yeah, I don't mind if things are a little slower as long as we can run the protocol tests with a simple `nargo test`. We can always improve the speed in future PRs.

aakoshh:
If you consider my earlier pseudo code that set up workers up front, you can imagine setting up a pipeline where you feed tests in order into the work queue, so it will be package A first then package B, and in another component you tell up front how many results are to be expected from each package. That component gets the results in a channel, and if they are out of order it buffers them, if it’s for the current item it shows them, when it reaches the expected number of reports it shows the summary and starts outputting the next one from the buffer

asterite:
@aakoshh Right, I looked at your code to see if I could use it, because it seems what we'd need to do in the end. The problem is that to get the tests to run we need to compile a package. In the pseudocode packages are compiled sequentially and that's what's slow.

However, I'll at least change things so that all packages are compiled in parallel. There's still a slowdown because we don't run tests from different packages in parallel. If the issue was too many threads and out of memory, using worker threads solved this but then I suggested running packages sequentially and that's what caused the additional slowdown. What I'm saying is that we could revert that functionality, get rid of the "out of memory", and still have things work really fast.

asterite:
Well, running tests for `noir-contracts` (with oracles failing) now takes 6 seconds vs. 2, so maybe it's not a big deal. I'll just push a commit that parallelizes the compilation.

aakoshh:
I'll talk to you on Slack, but I agree if you can compile in parallel, great! I thought the out-of-memory might be the proving part, I don't know. What I suggested was exactly to run tests in parallel, limited by the number of tests, even across packages, while displaying the results in a streaming fashion, using buffering to reorder them when a transition happens. 

Something like this:
```
let test_chan = channel::unbounded([a0, a1, a2, a3, a4, b0, b1, b2, c0, c1, d0]);
let test_stats = VecDequeue::from([(a, 5), (b, 3), (c, 2), (d, 1)]);
let result_chan = channel::unbounded();
for _ in 0 .. N { 
  spawn(|| {
    while Some(test) = test_chan.recv(); {
      result_chan.send(test.run());
    }
  });
}
spawn(|| {
  let mut buffer = HashMap::default();
  while Some((curr_package, num_tests)) = work_stats.pop_first() {
    let tests_received = 0;
    if let Some(buffered) = buffer.remove(curr_package) {
      for result in buffered { 
        show(result);
        tests_received += 1;
      }
    }
    while Some(result) = result_chan.recv() {      
      if result.package == curr_package { 
        show(result);
        tests_received += 1;
        break;
      } else {
        buffer[result.package].push(result);
      }
    }
  }
});
```

asterite:
I was seeing a difference between the test output of this PR vs. master and I just understood what was going on. In `master` as soon as there's a compilation error the program stops. This could happen even after already running a few tests. In this PR because we compile everything up-front all of the errors were shown and then the program exited.

With that said, I'm not sure CI has new failures compared to master. For rollup_lib I'm seeing this:

image

For private_kernel_lib this:

image

And then a few errors in contract_instance_deployer_contract:

image

But those are the errors I'm also seeing locally.

I do wonder if we should still run tests for packages that were able to compile... (but maybe for a follow-up PR)

asterite:
> I do wonder if we should still run tests for packages that were able to compile... (but maybe for a follow-up PR)

Checking with Akosh, it seems Rust doesn't run any tests if at least one workspace member has compile errors in it, so it's probably okay if we do the same.

This PR is now ready again for review.

asterite:
I just reverted the changes to `master` to see what are the failures for external checks: they are the same as in this PR. The `rollup_lib` ones seem to timeout (on my machine it takes 2:30 minutes to run them) so maybe CI for that has a timeout, not sure.

asterite:
I decided to keep working on this branch to avoid having multiple PRs on top of this one and later having to solve merge conflicts. I updated the main PR description with the new changes.

asterite:
Now that I think about it, I wonder if stdout for each test should be shown before each test instead of after. Or ir could be after, like now, but maybe precede it with "standard outfor for test:"

tooling/nargo/src/ops/test.rs
```diff
@@ -64,7 +66,8 @@ pub fn run_test<B: BlackBoxFunctionSolver<FieldElement>>(
                 // Run the backend to ensure the PWG evaluates functions like std::hash::pedersen,
                 // otherwise constraints involving these expressions will not error.
                 let mut foreign_call_executor = TestForeignCallExecutor::new(
-                    show_output,
+                    output,
+                    // show_output,
```
aakoshh, line 70:
```suggestion
```

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -42,8 +48,28 @@ pub(crate) struct TestCommand {
     /// JSON RPC url to solve oracle calls
     #[clap(long)]
     oracle_resolver: Option<String>,
+
+    /// Number of threads used for running tests in parallel
+    #[clap(long)]
+    test_threads: Option<usize>,
```
aakoshh, line 54: If you want you can provide the default value right here, see an [example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=731f0008d73402fd93f44c54b5c7e2f0)

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -68,258 +94,410 @@ pub(crate) fn run(args: TestCommand, config: NargoConfig) -> Result<(), CliError
...
+    workspace: Workspace,
+    args: &'a TestCommand,
+    pattern: FunctionNameMatch<'a>,
+    num_threads: usize,
+}
+
+impl<'a> TestRunner<'a> {
+    fn run(&self) -> Result<(), CliError> {
+        // First compile all packages and collect their tests
+        let package_test_results = self.collect_packages_tests();
```
aakoshh, line 122: When I read `package_test_results` I think _results_ as in the results of test execution, a collection of `TestResult` type above. I see it's actually a `HashMap<String, Result<Vec<Test<'a>>, CliError>>`. 

I see why you return `Result`, so we inspect compilation failures on a package-by-package basis, I reckon just `package_tests` would be a more intuitive name, especially because in iteration you use `(package_name, package_tests)`, not e.g. `(package_name, package_tests_result)`.

TomAFrench:
> Or ir could be after, like now, but maybe precede it with "standard outfor for test:"

I would say this.

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -68,258 +94,410 @@ pub(crate) fn run(args: TestCommand, config: NargoConfig) -> Result<(), CliError
...
+
+        // Now gather all tests and how many are per packages
+        let mut tests = Vec::new();
+        let mut test_count_per_package = HashMap::new();
+
+        for (package_name, package_tests) in package_test_results {
+            let package_tests = package_tests?;
```
aakoshh, line 129: So we get the report from the first package that failed to compile, but not the rest? Or did it already print something about why they failed earlier?

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -68,258 +94,410 @@ pub(crate) fn run(args: TestCommand, config: NargoConfig) -> Result<(), CliError
...
+                    // (the default is 2MB)
+                    .stack_size(STACK_SIZE)
+                    .spawn_scoped(scope, move || loop {
+                        // Get next test to process from the iterator.
+                        let Some(test) = iter.lock().unwrap().next() else {
+                            break;
+                        };
+
+                        let time_before_test = std::time::Instant::now();
+                        let (status, output) = (test.runner)();
```
aakoshh, line 193: Should we use [catch_unwind](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) to make sure no assertion in mocks or pedantic solvers bubble up?

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -68,258 +94,410 @@ pub(crate) fn run(args: TestCommand, config: NargoConfig) -> Result<(), CliError
...
+        let package_test_results = self.collect_packages_tests();
+
+        // Now gather all tests and how many are per packages
+        let mut tests = Vec::new();
+        let mut test_count_per_package = HashMap::new();
+
+        for (package_name, package_tests) in package_test_results {
+            let package_tests = package_tests?;
+            test_count_per_package.insert(package_name, package_tests.len());
+            tests.extend(package_tests);
```
aakoshh, line 131: I would use a `BTreeMap` instead of `HashMap` in all these cases, to fix the iteration order. One is to provide a more deterministic output among packages, and the other to make absolutely sure that the order we see while iterating `package_test_results`, which is going to be the order in which tests go into `tests`, is exactly going to be reflected in `test_count_per_package` as well, which we will iterate over when waiting for results to appear in that order. Or use a `Vec` for `test_count_per_package`.

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -68,258 +94,410 @@ pub(crate) fn run(args: TestCommand, config: NargoConfig) -> Result<(), CliError
...
+
+/// Display a report for all tests in a package
+fn display_test_report(package_name: &String, test_report: &[TestResult]) -> std::io::Result<()> {
+    let writer = StandardStream::stderr(ColorChoice::Always);
+    let mut writer = writer.lock();
+
+    let failed_tests: Vec<_> = test_report
+        .iter()
+        .filter_map(
+            |test_result| if test_result.status.failed() { Some(&test_result.name) } else { None },
```
aakoshh, line 464:
```suggestion
            |test_result| test_result.status.failed().then(&test_result.name),
```

aakoshh (commented):
Very nice!

tooling/nargo_cli/src/cli/test_cmd.rs
asterite: Right, the way it works is a bit confusing: the error is printed right away and `Err` is returned. So by the time we get here all the errors were already printed 🤷

asterite: That is, it's confusing but it was already working like that (it's how compilation works)

aakoshh: Then maybe return `Result<BTreeMap<String, Vec<Test<'a>>>, CliError>` from `collect_packages_tests` to indicate that there's nothing more to do with this?

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -68,258 +94,410 @@ pub(crate) fn run(args: TestCommand, config: NargoConfig) -> Result<(), CliError
...
+                        format!("Found 0 tests containing '{pattern}'.",),
+                    ))
+                }
+                // If we are running all tests in a crate, having none is not an error
+                FunctionNameMatch::Anything => {}
+            };
+        }
+
+        if test_reports.iter().any(|test_result| test_result.status.failed()) {
+            Err(CliError::Generic(String::new()))
```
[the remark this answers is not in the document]
aakoshh: Ah right, they assume everything that needed to be said has been printed already 👍

tooling/nargo_cli/src/cli/test_cmd.rs
aakoshh: What I'd like to understand is that if somebody writes a test that hits for example a panic [like this one](https://github.com/noir-lang/noir/blob/4b0df947a3e74b633412743ff286954305f39938/acvm-repo/blackbox_solver/src/bigint.rs#L71-L72) by using some invalid value in a test, is that going to crash `nargo` itself, and if it does, is it going to tell before it happens which test was to blame?

tooling/nargo_cli/src/cli/test_cmd.rs
asterite: Great idea!

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -68,258 +94,410 @@ pub(crate) fn run(args: TestCommand, config: NargoConfig) -> Result<(), CliError
...
+        root_path: Option<PathBuf>,
+        package_name: String,
+    ) -> (TestStatus, String) {
+        // This is really hacky but we can't share `Context` or `S` across threads.
+        // We then need to construct a separate copy for each test.
+
+        let (mut context, crate_id) =
+            prepare_package(self.file_manager, self.parsed_files, package);
+        check_crate(&mut context, crate_id, &self.args.compile_options)
+            .expect("Any errors should have occurred when collecting test functions");
```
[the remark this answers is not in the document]
asterite: I think it's because it was compiled before?

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -68,258 +94,410 @@ pub(crate) fn run(args: TestCommand, config: NargoConfig) -> Result<(), CliError
...
+                        remaining_test_count -= 1;
+                        if remaining_test_count == 0 {
+                            break;
+                        }
+                    }
+                }
+
+                display_test_report(package_name, &test_report)
+                    .expect("Could not display test report");
+                test_reports.extend(test_report);
```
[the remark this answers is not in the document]
asterite: Good point. I just kept it like that because it was like that, but I think we only need to know if there are any tests results, and if there are failures. I'll change it to track those things.

tooling/nargo_cli/src/cli/test_cmd.rs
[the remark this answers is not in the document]
aakoshh: I was just surprised, I thought some kind of compiled bytecode will go along with the tests, with a function ID to select the entry point.

tooling/nargo_cli/src/cli/test_cmd.rs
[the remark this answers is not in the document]
aakoshh: Cool, thanks. It's only because that way we don't have to think about whether the reporting has already happened: it must have, because we can't do it on the return value. Also frees up the output strings.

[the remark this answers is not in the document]
asterite: Done!

tooling/nargo_cli/src/cli/test_cmd.rs
asterite: Done! Though I'm sure there's a more idiomatic Rust way to do this...

aakoshh: With stuff like `Vec<Result<_, _>>` you can use `let foo: Result<Vec<_>, _> = results.collect();` but I'm not sure it works with a `BTreeMap`.

TomAFrench:
> At this point I'm wondering if we actually need to prefix all the output with the package name. We could maybe print it at the beginning and the end, but not for every test. Thoughts?

Let's keep it for now. This is consistent with other packages and we have quite a few changes in this PR already.

tooling/nargo_cli/src/cli/test_cmd.rs
asterite: I just tried it. It seems that if there's a panic inside `run_test` then that panic is printed(it could happen in multiple threads) and then the program aborts.

If we use `catch_unwind` the program continues, but we can at least know the tests that failed because of an unexpected errors. So I'll make this change 👍

aakoshh (approved):
👏

Savio-Sou:
Looks like the new `nargo test` flags introduced in this PR and https://github.com/noir-lang/noir/pull/6776 are not documented in https://noir-lang.org/docs/reference/nargo_commands#nargo-test.

Are they all not meant to for use by the community?

PR #6736, opened 2024-12-09
fix: allow empty loop headers

### LINKED ISSUE #6734: Compiler panic during SSA unrolling
The following noir program crashes:

```noir
pub fn subarray_extended<let SRC_LEN: u32, let DST_LEN: u32>(
    src: [Field; SRC_LEN],
    offset: u32,
) -> [Field; DST_LEN] {
    let available_elements_to_copy = SRC_LEN - offset;
    let elements_to_copy = if DST_LEN > available_elements_to_copy {
        available_elements_to_copy
    } else {
        DST_LEN
    };

    let mut dst: [Field; DST_LEN] = std::mem::zeroed();
    for i in 0..elements_to_copy {
        dst[i] = src[i + offset];
    }

    dst
}

#[test]
unconstrained fn subarray_extended_into_empty() {
    assert_eq(subarray_extended([], 0), []);
}
```

`nargo test` yields:

```
[example] Running 1 test function
Message:  assertion `left == right` failed: The header should just compare the induction variable and jump
  left: 0
 right: 1
Location: compiler/noirc_evaluator/src/ssa/opt/unrolling.rs:305
```

This seems to be related to the for loop iterating over `elements_to_copy`. An empty loop body also causes the crash, and `elements_to_copy`  can be simplified a bit and still cause a crash, for example:

```noir
let elements_to_copy = if DST_LEN > available_elements_to_copy {
    1
} else {
    DST_LEN
};
```

guipublic:
# Description

## Problem\*

Resolves #6734

## Summary\*
Allow empty loop headers

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/test_programs/execution_success/regression_6734/Nargo.toml
+++ b/test_programs/execution_success/regression_6734/Nargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "regression_6736"
+name = "regression_6734"
 type = "bin"
 authors = [""]
 


TomAFrench:
Can you add a regression test? Looks like #6734 relies on constants being folded so that we simplify the jump condition at just the wrong moment so using the SSA parser may be difficult.

test_programs/execution_success/regression/src/main.nr
```diff
@@ -138,4 +138,31 @@ fn main(x: [u8; 5], z: Field, u: i16, v: i16) {
     assert(-u % -11 == -4);
     assert(u % -11 == u % (v + 2));
     assert(-u % -11 == -u % (v + 2));
+
+    // Issue 6736
```
aakoshh, line 142:
```suggestion
    // Issue 6734
```

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -302,6 +302,11 @@ impl Loop {
     pub(super) fn get_const_upper_bound(&self, function: &Function) -> Option<FieldElement> {
         let block = &function.dfg[self.header];
         let instructions = block.instructions();
+        if instructions.is_empty() {
+            // If the loop condition is constant time, the loop header will be
+            // simplified to a simple jump.
```
aakoshh, line 307: So this loop will have zero iterations, right, there is nothing whatsoever to unroll?

aakoshh: Just curious:would it work if we returned `Some(0)` instead of `None`?

jfecher (approved):
LGTM

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
jfecher: That'd set the upper bound at zero which could work differently if we ever allow looping starting from a negative index: `for _ in -5 .. 0 { ... }`. Returning None is more correct here

test_programs/execution_success/regression/src/main.nr
```diff
@@ -138,4 +138,32 @@ fn main(x: [u8; 5], z: Field, u: i16, v: i16) {
     assert(-u % -11 == -4);
     assert(u % -11 == u % (v + 2));
     assert(-u % -11 == -u % (v + 2));
+
+    // Issue 6734
+    unsafe {
+        sub_array_extended_into_empty();
+    }
```
TomAFrench, line 145: Can you create a separate directory for this new test? I'd like to avoid `regression` becoming a "kitchen sink" test.

TomAFrench (approved):

PR #6737, opened 2024-12-09
chore: separate test compilation from test running

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR separates out compilation from the `run_test` function. I've chosen to do this as compilation errors are all global as far as I know (i.e. if we can't compile one test program in a package then that will prevent any others from being compiled)

We can then have the `run_test` program take in a compiled program in order to improve separation of concerns.

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
Nice refactor!

PR #6743, opened 2024-12-09
chore: Cleanup unrolling pass

jfecher:
# Description

## Summary\*

Minor code cleanup of some unrolling methods

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
@@ -279,10 +279,10 @@ impl Loop {
         &self,
         function: &Function,
         cfg: &ControlFlowGraph,
-    ) -> Result<Option<FieldElement>, CallStack> {
-        let pre_header = self.get_pre_header(function, cfg)?;
-        let jump_value = get_induction_variable(function, pre_header)?;
-        Ok(function.dfg.get_numeric_constant(jump_value))
+    ) -> Option<FieldElement> {
+        let pre_header = self.get_pre_header(function, cfg).ok()?;
+        let jump_value = get_induction_variable(function, pre_header).ok()?;
+        function.dfg.get_numeric_constant(jump_value)
     }
 
     /// Find the upper bound of the loop in the loop header and return it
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -603,13 +596,11 @@ impl Loop {
...
-            .map(|block| {
-                let block = &function.dfg[*block];
-                block.instructions().len() + block.terminator().map(|_| 1).unwrap_or_default()
-            })
-            .sum()
+        let iter = self.blocks.iter().map(|block| {
+            let block = &function.dfg[*block];
+            block.instructions().len() + block.terminator().is_some() as usize
+        });
+        iter.sum()
```
aakoshh, line 603: Is the breaking up of `.iter()` here and with `instructions` above something we should be doing consistently, or just something up to personal preference on a case-by-case basis?

aakoshh (approved):
Yeah that's better 👍

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
jfecher: It's just to nudge the formatter into not separating `map` and it's brackets onto separate lines just because there's a `sum` afterward. Without it, it gets formatted to:
```rust
        self.blocks
            .iter()
            .map(|block| {
                let block = &function.dfg[*block];
                block.instructions().len() + block.terminator().is_some() as usize
            })
            .sum()
```

TomAFrench (approved):

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
aakoshh: Indeed, I was just wondering if this is something that I should strive to achieve, or it's not a big deal if it appears on separate lines either.

jfecher: I just think this formatting choice by rustfmt is a bit ugly so I avoid it in my code at least. It's just a personal preference though and I've no actual objective reasons backing me up besides the code being a tad shorter.

aakoshh: Okay, just wanted to make sure I'm not systematically causing debt for others to clean up after me 🙂

PR #6747, opened 2024-12-09
chore: Try replace callstack with a linked list

jfecher:
# Description

## Problem\*

Resolves 

## Summary\*

Test to see memory usage change from replacing the callstack with a (shared) linked list.

## Additional Context

Surprisingly, I had trouble finding an existing crate which provides a shared linked list. I ended up writing my own. (`std::collections::LinkedList` boxes each item so it isn't shared). Callstacks come up several times in parallel iterators unfortunately so I had to use `Arc` over `Rc`.

Still need to reverse the display order of callstacks now that they're stored in reverse order due to being lists.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
> regression_4709 	300.69M 	-10%

Maybe not big enough to warrant merging?

Also ram_blowup_regression is 2.55G on other PRs but is showing as 0% changed here

TomAFrench:
Let's keep this PR open until https://github.com/noir-lang/noir/pull/6750 is landed to see what the effects are on the protocol circuits.

jfecher:
Comparing this to https://github.com/noir-lang/noir/pull/6753, the peak memory usage reduction is much closer to optimal, -4% and -10% compared to a maximum of -5% and -15% respectively for those tests.

TomAFrench:
The report isn't showing this correctly but it's cutting the memory usage of private-kernel-reset by ~400MB.

TomAFrench (approved):
Approving but not going to merge as it was left in draft.

TomAFrench:
> Still need to reverse the display order of callstacks now that they're stored in reverse order due to being lists.

Ah should have read the OP more closely.

jfecher:
Reversed iteration direction of the call stacks so this should be ready for review / merge now.

TomAFrench (approved):
LGTM

compiler/noirc_evaluator/src/ssa/ir/list.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    pub fn len(&self) -> usize {
+        self.len
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.len == 0
+    }
+
+    fn pop_front(&mut self) -> Option<T>
```
aakoshh, line 67: Both `push_back` and `pop_front` operate on the head; I understand what you meant by `push_back` actually being a "push_front", but wouldn't it be better to name both of these head operations to be the same? I thought `pop_front` will go to the opposite end from `push_back` in an O(n) operation. Why not just `push` and `pop`?

aakoshh:
My all time favourite introduction to Rust has some hilarious discussion about linked lists that might explain why there aren't many implementations around: https://rust-unofficial.github.io/too-many-lists/index.html :)

compiler/noirc_evaluator/src/ssa/ir/list.rs
jfecher: Yeah, that's my mistake. They were originally both `*_back` to make it more easy for me to migrate existing code, then I changed `pop_front` to be the front to be more accurate but thought renaming the public `push_back` to front as well could be confusing to CallStack users who want to add something to the conceptual "back" of the structure, unaware that the structure is essentially reversed internally.

I'd be for renaming it back to pop_back.

PR #6748, opened 2024-12-09
chore: free memory for silenced warnings early

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

The protocol circuits currently print out 250k lines of warnings to the console which reminded me that we track all of the typechecking warnings through the full compilation flow even if we've decided to silence them (as they do in aztec-packages).

This PR updates `check_crate` to drop all these warnings before we start the SSA passes so that we can free up some memory.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_driver/src/lib.rs
+++ b/compiler/noirc_driver/src/lib.rs
@@ -316,7 +316,7 @@ pub fn check_crate(
         })
         .filter(|diagnostic| {
             // We filter out any warnings if they're going to be ignored later on to free up memory.
-            !options.silence_warnings || diagnostic.diagnostic.kind == DiagnosticKind::Warning
+            !options.silence_warnings || diagnostic.diagnostic.kind != DiagnosticKind::Warning
         })
         .collect();
 
--- a/compiler/noirc_driver/src/lib.rs
+++ b/compiler/noirc_driver/src/lib.rs
@@ -10,7 +10,7 @@ use clap::Args;
 use fm::{FileId, FileManager};
 use iter_extended::vecmap;
 use noirc_abi::{AbiParameter, AbiType, AbiValue};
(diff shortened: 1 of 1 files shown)


compiler/noirc_driver/src/lib.rs
```diff
@@ -301,23 +301,29 @@ pub fn check_crate(
...
-    }));
+    let warnings_and_errors: Vec<FileDiagnostic> = diagnostics
+        .into_iter()
+        .map(|(error, file_id)| {
+            let diagnostic = CustomDiagnostic::from(&error);
+            diagnostic.in_file(file_id)
+        })
+        .filter(|diagnostic| {
+            // We filter out any warnings if they're going to be ignored later on to free up memory.
+            !options.silence_warnings || diagnostic.diagnostic.kind == DiagnosticKind::Warning
```
aakoshh, line 319:
```suggestion
            !options.silence_warnings || diagnostic.diagnostic.kind != DiagnosticKind::Warning
```

compiler/noirc_driver/src/lib.rs
```diff
@@ -301,23 +301,29 @@ pub fn check_crate(
...
+        })
+        .collect();
 
-    if has_errors(&errors, options.deny_warnings) {
-        Err(errors)
+    if has_errors(&warnings_and_errors, options.deny_warnings) {
```
aakoshh, line 323: Would it make sense to only filter the warnings _after_ `deny_warnings` had a chance to look at them, or does `silence_warnings` imply `!deny_warnings` as well?

aakoshh: Never mind, I see only one of them can be true at a time 👍

aakoshh (approved):
👍

PR #6750, opened 2024-12-09
chore(ci): Handle external libraries in compilation timing report

### LINKED ISSUE #6722: Add tracking for compilation times of important circuits in CI
We currently have no good way to quantitatively track whether a PR will cause a regression in compilation time for important user circuits. A recent example of this coming back to bite us is in the bloblib circuits where compilation times jumped from 10 to 90mins.

Similar to `run-external-checks`, we should add a workflow to CI which will compile a number of protocol circuits (the rollup circuits in particular) and measure the time taken to do so. This can then:

- Print a comment to the PR showing the time and highlight any regressions (would require a modification of https://github.com/noir-lang/noir-bench-report to support this new metric)
- Define a hard limit on compilation times which will automatically block the PR if exceeded.

As a first cut we can just use `time nargo compile --force` to extract the compilation time so this is just a devops issue and shouldn't need modifying the compiler.

vezenovm:
# Description

## Problem\*

Resolves #6722 

## Summary\*

Just pushing a draft for now to see that things are working

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
We'd also like to track memory usage for the protocol circuits which we can pretty easily add in the same way in this PR.

TomAFrench:
Had a bit of a deeper look at this PR. Looks like you're trying to pass around a single compilation report from job to job and append to it?

If so have you considered having all the individual reports being generated in parallel and then having a single step to merge them all? This could just be a simple `jq` call.

vezenovm:
> If so have you considered having all the individual reports being generated in parallel and then having a single step to merge them all? This could just be a simple `jq` call.

Yeah I was hoping to switch to something like this. I wanted to get the basics of the github action working first and then look at how it can be done better.

.github/workflows/reports.yml
```diff
@@ -262,8 +261,145 @@ jobs:
...
+    needs: [generate_compilation_report]
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - project: { repo: AztecProtocol/aztec-packages, path: ./noir-projects/aztec-nr }
+          - project: { repo: AztecProtocol/aztec-packages, path: ./noir-projects/noir-contracts }
+          - project: { repo: AztecProtocol/aztec-packages, path: ./noir-projects/noir-protocol-circuits/crates/parity-lib }
```
TomAFrench, line 293:
```suggestion
          - project: { repo: AztecProtocol/aztec-packages, path: ./noir-projects/noir-contracts }
```

These other two are libraries so we won't get a meaningful compilation time out of them.

.github/workflows/reports.yml
```diff
@@ -262,8 +261,145 @@ jobs:
...
+          chmod +x $nargo_binary
+          echo "$(dirname $nargo_binary)" >> $GITHUB_PATH
+          export PATH="$PATH:$(dirname $nargo_binary)"
+          nargo -V
+
+      - name: Download compilation report script
+        uses: actions/download-artifact@v4
+        with:
+          name: compilation_report_script
+          path: ./test-repo/${{ matrix.project.path }}
```
TomAFrench, line 326: You can do a sparse checkout of the noir repository to just pull in this script which should make this simpler.

https://github.com/actions/checkout?tab=readme-ov-file#fetch-only-the-root-files-and-github-and-src-folder

vezenovm:
> Nice, I'll close all the performance issues :p

Yeah not sure why this started displaying such times...

.github/workflows/reports.yml
vezenovm: I have been making some edits locally, so I updated this manually

.github/workflows/reports.yml
```diff
@@ -316,17 +310,10 @@ jobs:
...
-          name: script_for_compilation_report
-          path: ./test-repo/${{ matrix.project.path }}
-
-      # - name: Download compilation report
-      #   uses: actions/download-artifact@v4
-      #   with:
-      #     name: in_progress_compilation_report
-      #     path: ./test-repo/${{ matrix.project.path }}
+          sparse-checkout: |
+            ./test_programs/compilation_report.sh
```
TomAFrench, line 317: We're downloading this script to a different directory now, I'm not sure if we can do the checkout into `./test_repo` or if we need to move the file after the fact however.

We may need to update this to not use cone mode: https://github.com/actions/checkout?tab=readme-ov-file#fetch-only-a-single-file

TomAFrench: I imagine that it's placed at `test_repo/test_programs/compilation_report.sh` so that if you were to add more files into the sparse checkout then it won't move any of the old files.

vezenovm: Yeah I just noticed that I was not specifying cone mode as false.

vezenovm: > I'm not sure if we can do the checkout into `./test_repo` or if we need to move the file after the fact however.

I guess we shall see

.github/workflows/reports.yml
```diff
@@ -262,8 +261,134 @@ jobs:
...
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+            ./test_programs/compilation_report.sh
+          sparse-checkout-cone-mode: false
+
+      - name: Move compilation report
+        run: |
+          echo $(ls ./)
+          echo $(lhs ./test_programs)
```
TomAFrench, line 322:
```suggestion
          echo $(ls ./test_programs)
```

.github/workflows/reports.yml
```diff
@@ -262,8 +261,133 @@ jobs:
...
+
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+            ./test_programs/compilation_report.sh
+          sparse-checkout-cone-mode: false
+
+      - name: Move compilation report
+        run: |
+          echo $(ls ./)
```
TomAFrench, line 321:
```suggestion
          ls ./
```

vezenovm: yeah I removed the echo

.github/workflows/reports.yml
```diff
@@ -262,8 +261,134 @@ jobs:
...
+        with:
+          sparse-checkout: |
+            test_programs/compilation_report.sh
+          sparse-checkout-cone-mode: false
+
+      - name: Move compilation report
+        run: |
+          ls ./
+          ls ./test_programs
+          mv ./test_programs/compilation_report.sh ./test-repo/${{ matrix.project.path }}/compilation_report.sh
```
TomAFrench, line 323:
```suggestion
          mv ./test_programs/compilation_report.sh ./test-repo/${{ matrix.project.path }}/
```

.github/workflows/reports.yml
```diff
@@ -262,8 +261,134 @@ jobs:
...
+
+      - name: Move compilation report
+        run: |
+          ls ./
+          ls ./test_programs
+          mv ./test_programs/compilation_report.sh ./test-repo/${{ matrix.project.path }}/
+
+      - name: Generate compilation report
+        working-directory: ./test-repo/${{ matrix.project.path }}
+        run: |
```
TomAFrench, line 327:
```suggestion

      - name: Generate compilation report
        working-directory: ./test-repo/${{ matrix.project.path }}
        run: |
            mv /home/runner/test_programs/compilation_report.sh ./compilation_report.sh
```

Maybe try pulling `compilation_report.sh` into the correct directory rather than pushing?

.github/workflows/reports.yml
```diff
@@ -262,8 +261,139 @@ jobs:
...
+          nargo_binary="${{ github.workspace }}/nargo/nargo"
+          chmod +x $nargo_binary
+          echo "$(dirname $nargo_binary)" >> $GITHUB_PATH
+          export PATH="$PATH:$(dirname $nargo_binary)"
+          nargo -V
+
+      - uses: actions/checkout@v4
+        with:
+          # path: test-repo/${{ matrix.project.path }}
```
vezenovm, line 315:
```suggestion
          path: test-repo
```

.github/workflows/reports.yml
```diff
@@ -262,8 +261,141 @@ jobs:
...
+          repository: ${{ matrix.project.repo }}
+          path: test-repo
+          ref: ${{ matrix.project.ref }}
+
+      - name: Generate compilation report
+        working-directory: ./test-repo/${{ matrix.project.path }}
+        run: |
+          ls .
+          ls /home/runner/work/
+          ls ~
```
vezenovm, line 332:
```suggestion
          echo " "
          ls /home/runner/work/noir
```

.github/workflows/reports.yml
```diff
@@ -262,8 +261,143 @@ jobs:
...
+      - name: Upload compilation report
+        uses: actions/upload-artifact@v4
+        with:
+          name: in_progress_compilation_report
+          path: compilation_report.json
+          retention-days: 3
+          overwrite: true
+
+  external_repo_compilation_report:
+    needs: [generate_compilation_report]
```
TomAFrench, line 276: We should be able to relax this restriction (and then add `generate_compilation_report` as a new dependency of the report accumulation job)

.github/workflows/reports.yml
```diff
@@ -262,8 +261,143 @@ jobs:
...
+      matrix:
+        include:
+          - project: { repo: AztecProtocol/aztec-packages, path: ./noir-projects/noir-contracts }
+          - project: { repo: AztecProtocol/aztec-packages, path: ./noir-projects/noir-protocol-circuits/crates/parity-root }
+          - project: { repo: AztecProtocol/aztec-packages, path: ./noir-projects/noir-protocol-circuits/crates/private-kernel-inner }
+          - project: { repo: AztecProtocol/aztec-packages, path: ./noir-projects/noir-protocol-circuits/crates/private-kernel-reset }
+          - project: { repo: AztecProtocol/aztec-packages, path: ./noir-projects/noir-protocol-circuits/crates/rollup-base-private }
+          - project: { repo: AztecProtocol/aztec-packages, path: ./noir-projects/noir-protocol-circuits/crates/rollup-base-public }
+
+    name: External repo compilation report - ${{ matrix.project.repo }}
```
TomAFrench, line 290: We should include the path as we have a lot of projects from the same repository.

.github/workflows/reports.yml
vezenovm: done

.github/workflows/reports.yml
vezenovm: done

.github/workflows/reports.yml
```diff
@@ -262,8 +261,133 @@ jobs:
...
+    timeout-minutes: 15
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-contracts }
+          - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/parity-root }
+          - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/private-kernel-inner }
+          - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/private-kernel-reset }
+          # TODO: Bring these back once they no longer time out
```
vezenovm, line 287: These were taking over 30 minutes to compile in CI. However, running nargo compile locally against aztec-packages master these rollup circuits compiled in just over 4 minutes.

.github/workflows/reports.yml
```diff
@@ -262,8 +261,133 @@ jobs:
...
+          # - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/rollup-base-private }
+          # - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/rollup-base-public }
```
TomAFrench, line 289:
```suggestion
          - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/rollup-base-private }
          - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/rollup-base-public }
```

TomAFrench: Testing these out again.

.github/workflows/reports.yml
TomAFrench: Yeah, I'm getting the same compilation time on the mainframe as you. I can only imagine that we're running into memory issues on the github runner?

TomAFrench: In any case, these are the circuits we're most interested in so it's pretty important to include them in these benchmarks.

.github/workflows/reports.yml
```diff
@@ -262,8 +261,133 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        include:
+          - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-contracts }
+          - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/parity-root }
+          - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/private-kernel-inner }
+          - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/private-kernel-reset }
+          # TODO: Bring these back once they no longer time out
+          - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/rollup-base-private }
+          - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/rollup-base-public }
```
TomAFrench, line 289:
```suggestion
          #- project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/rollup-base-private }
          #- project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/rollup-base-public }
```

TomAFrench (approved):

PR #6752, opened 2024-12-10
feat(cli): Run command on the package closest to the current directory

### LINKED ISSUE #2606: Run commands only on a single package if current working directory is within that package
### Problem

Within a workspace, we currently run all commands under the `DefaultOrAll` strategy no matter where you are within the workspace. This is unexpected coming from rust where if I run the command from within a package then it will apply to that package only unless the `--workspace` flag is set.

### Happy Case

When running commands within a package, the command should be equivalent to us running `nargo <command> --package <package>`, otherwise we should perform the current workspace behaviour.

### Alternatives Considered

_No response_

### Additional Context

_No response_

No

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #2606

## Summary\*

Changed the behaviour of the CLI to look for:
* the innermost manifest if neither `--workspace` nor `--package` is given, ie. find the manifest of the current directory
* the outermost manifest if either `--workspace` or `--package` is given (the latter so we can test package `b` when we're in the directory of package `a`)

Added a new `PackageOptions` struct to contain the above two options, to cut down some of the repetition in the commands. 

Added a `nargo_toml::find_root` function that takes a `workspace` flag to indicate whether we want the outer or inner package.

I did not change the LSP which calls `find_package_manifest` in test requests, it will keep finding the workspace manifest:
* [on_test_request_inner](https://github.com/noir-lang/noir/blob/52bcb0f24b27090e812611c23ed326a541d37153/tooling/lsp/src/requests/tests.rs#L30-L45) tests all packages, so in theory it could use the innermost package, and then if the editor was opened in a package, rather than on the whole workspace, it could limit the scope of tests; I'm not sure if we want to change this, so I left it as-is.
* [on_test_run_request_inner](https://github.com/noir-lang/noir/blob/52bcb0f24b27090e812611c23ed326a541d37153/tooling/lsp/src/requests/test_run.rs#L31-L41) gets the name of the crate to test in the request parameters - this will only work reliably if we allow the package to be found in the workspace.

### Testing

Using the `overlapping_dep_and_mod` for testing because it has two members in its workspace: `foo` and `bin`; the latter actually has `overlapping_dep_and_mod` as package name.

```console
❯ cd test_programs/compile_success_empty/overlapping_dep_and_mod/foo/src

❯ cargo run -q -p nargo_cli -- --program-dir . test
[foo] Running 0 test functions
[foo] 0 tests passed

❯ cargo run -q -p nargo_cli -- --program-dir . test --package foo
[foo] Running 0 test functions
[foo] 0 tests passed

❯ cargo run -q -p nargo_cli -- --program-dir . test --package bin
Selected package `bin` was not found

❯ cargo run -q -p nargo_cli -- --program-dir . test --package overlapping_dep_and_mod
[overlapping_dep_and_mod] Running 0 test functions
[overlapping_dep_and_mod] 0 tests passed

❯ cargo run -q -p nargo_cli -- --program-dir . test --workspace
[overlapping_dep_and_mod] Running 0 test functions
[overlapping_dep_and_mod] 0 tests passed
[foo] Running 0 test functions
[foo] 0 tests passed

❯ cd ../..

❯ cargo run -q -p nargo_cli -- --program-dir . test 
[overlapping_dep_and_mod] Running 0 test functions
[overlapping_dep_and_mod] 0 tests passed

❯ cargo run -q -p nargo_cli -- --program-dir . test --workspace
[overlapping_dep_and_mod] Running 0 test functions
[overlapping_dep_and_mod] 0 tests passed
[foo] Running 0 test functions
[foo] 0 tests passed

```

## Documentation\*

Check one:

# PR Checklist\*


michaeljklein (changes requested):
Do the docs need to be updated?

I see your comment:
> Using the overlapping_dep_and_mod for testing because it has two members in its workspace:
> ..

Please add one or more unit tests along these lines. I only see `test_find_root_does_not_panic`

aakoshh:
I couldn’t see anything in the docs that mentioned the current behaviour, and both the `—package` and `—workspace` options are hidden, so I just added explanation to those for the folks who know where to look. Because the ticket says this change outs to bring it in line with how `cargo` works I thought this behaviour would not need explanation. 

Regarding unit test, okay, will add something. Didn’t do it because it relies on finding TOML files in the file system, so it requires a bit of cumbersome setup.

aakoshh:
@michaeljklein added docs and tests, PTAL.

TomAFrench:
There's been a couple of instances of people wanting to put a nargo project within another nargo project without creating a workspace. This has been blocked by the fact that the outer nargo toml will fail to be registered as a workspace config but with this PR that will no longer happen.

Can you add some documentation saying that we don't support nested packages except within a workspace so that if we re-break this in future then it's not an issue.

tooling/nargo_toml/src/lib.rs
```diff
@@ -556,6 +602,124 @@ fn parse_package_expression_width_toml() {
...
+    )]
+    fn test_find_root_does_not_panic(workspace: bool, path: &str) {
+        let path = PathBuf::from_str(path).unwrap();
+        let error = find_root(&path, workspace).expect_err("non-existing paths");
+        assert!(matches!(error, ManifestError::MissingFile(_)));
+    }
+
+    /// Test to demonstrate how `find_root` works.
+    #[test]
+    fn test_find_root_example() {
```
TomAFrench, line 623: Very snazzy!

tooling/nargo_toml/src/lib.rs
```diff
@@ -47,6 +47,35 @@ pub fn find_file_manifest(current_path: &Path) -> Option<PathBuf> {
...
+/// When `workspace` is `true` it returns the topmost directory, when `false` the innermost one.
+///
+/// Returns a [ManifestError] if no parent directories of `current_path` contain a manifest file.
+pub fn find_root(current_path: &Path, workspace: bool) -> Result<PathBuf, ManifestError> {
+    if workspace {
+        find_package_root(current_path)
+    } else {
+        find_file_root(current_path)
+    }
+}
```
TomAFrench, line 59: This is great and an improvement over the current behaviour but there's still some holes where someone can have a workspace which sits within another workspace.

If someone tries to run a command within the inner workspace they'll end up with very confusing errors messages rather than it complaining about nested workspaces. We can handle this in a followup PR as it's low priority.

tooling/nargo_cli/src/cli/mod.rs
```diff
@@ -83,22 +115,8 @@ pub(crate) fn start_cli() -> eyre::Result<()> {
...
-        | NargoCommand::Info(..) => {
-            config.program_dir = find_package_root(&config.program_dir)?;
-        }
-        NargoCommand::New(..)
-        | NargoCommand::Init(..)
-        | NargoCommand::Lsp(..)
-        | NargoCommand::Dap(..)
-        | NargoCommand::GenerateCompletionScript(..) => (),
+    if let Some(workspace) = is_workspace_rooted(&command) {
+        config.program_dir = nargo_toml::find_root(&config.program_dir, workspace)?;
```
TomAFrench, line 119: This is a nice refactor but this causes an unwanted change in behaviour, I think. We want to move to the root of any packages we're in for these commands so that we can run `nargo execute` anywhere within a noir package rather than just at the root. The issue is it seems we only do so if we've got the `--workspace` flag set now.

TomAFrench: Ohhh, I'm being slow. The `Option` handles this.

tooling/nargo_cli/src/cli/mod.rs
```diff
@@ -127,6 +145,27 @@ pub(crate) fn start_cli() -> eyre::Result<()> {
...
+fn is_workspace_rooted(cmd: &NargoCommand) -> Option<bool> {
+    match &cmd {
+        NargoCommand::Check(cmd) => Some(cmd.package_options.is_workspace_rooted()),
+        NargoCommand::Compile(cmd) => Some(cmd.package_options.is_workspace_rooted()),
+        NargoCommand::Execute(cmd) => Some(cmd.package_options.is_workspace_rooted()),
+        NargoCommand::Export(cmd) => Some(cmd.package_options.is_workspace_rooted()),
+        NargoCommand::Test(cmd) => Some(cmd.package_options.is_workspace_rooted()),
+        NargoCommand::Info(cmd) => Some(cmd.package_options.is_workspace_rooted()),
+        NargoCommand::Debug(cmd) => Some(cmd.package.is_some()),
+        NargoCommand::Fmt(..) => Some(true),
```
TomAFrench, line 160: For consistency with cargo, let's not always run this on the full workspace.

tooling/nargo_cli/src/cli/mod.rs
TomAFrench: A enum wouldn't be the worst thing in the world tbh, using an `Option` for this puts of a little bit of a hurdle for people reading this.

```noir 
fn get_command_location(&Command) -> CommandLocation

enum CommandLocation {
   Workspace,
   Package,
   Any
}
````

aakoshh: I started adding something like that, but then it becomes so so noisy to invoke `find_root`. Renamed to `manifest_scope` now with an explanation. But if you think an enum is better I'll give it another go.

aakoshh: It's an enum now 👍

TomAFrench:
@michaeljklein Can you take another look as you're currently blocking this pr.

michaeljklein (approved):
LGTM

Issue #6751, opened 2024-12-09
`Instruction::map_values` is allocating a lot of memory


TomAFrench:
![Image](https://github.com/user-attachments/assets/191f39fd-3fee-4fed-a07f-954d067405b7)

This image shows the memory usage for compiling the `rollup-base-private`. 50% of this memory usage comes from `Instruction::map_values` so we should focus on this method in order to reduce memory usage.

PR #6756, opened 2024-12-10
chore: Add `Instruction::map_values_mut`

jfecher:
# Description

## Problem\*

Working towards #6751

## Summary\*

I noticed in a few places we use `Instruction::map_values` we clone the instruction beforehand to get around borrowing restrictions. This means we should mutate in place though instead of using `map_values` to produce another `Instruction` value, so I've added `map_values_mut` for `Instruction` and `DataBus` and searched through to find each case we could use these as well as the existing `TerminatorInstruction::mutate_values` which I've renamed.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -599,8 +599,9 @@ impl<'f> PerFunctionContext<'f> {
     }
 
     fn update_data_bus(&mut self) {
-        let databus = self.inserter.function.dfg.data_bus.clone();
-        self.inserter.function.dfg.data_bus = databus.map_values(|t| self.inserter.resolve(t));
+        let mut databus = self.inserter.function.dfg.data_bus.clone();
```
TomAFrench, line 602: We can avoid this clone, no?

compiler/noirc_evaluator/src/ssa/ir/function_inserter.rs
```diff
@@ -73,25 +73,24 @@ impl<'f> FunctionInserter<'f> {
...
     }
 
     /// Maps the data bus in place, replacing any ValueId in the data bus with the
     /// resolved version of that value id from this FunctionInserter's internal value mapping.
     pub(crate) fn map_data_bus_in_place(&mut self) {
-        let data_bus = self.function.dfg.data_bus.clone();
-        let data_bus = data_bus.map_values(|value| self.resolve(value));
+        let mut data_bus = self.function.dfg.data_bus.clone();
+        data_bus.map_values_mut(|value| self.resolve(value));
         self.function.dfg.data_bus = data_bus;
```
TomAFrench, line 94: ditto here.

compiler/noirc_evaluator/src/ssa/ir/function_inserter.rs
```diff
@@ -73,25 +73,24 @@ impl<'f> FunctionInserter<'f> {
 
     /// Get an instruction and make sure all the values in it are freshly resolved.
     pub(crate) fn map_instruction(&mut self, id: InstructionId) -> (Instruction, CallStack) {
-        (
-            self.function.dfg[id].clone().map_values(|id| self.resolve(id)),
-            self.function.dfg.get_call_stack(id),
-        )
+        let mut instruction = self.function.dfg[id].clone();
+        instruction.map_values_mut(|id| self.resolve(id));
+        (instruction, self.function.dfg.get_call_stack(id))
```
TomAFrench, line 78: and here.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
jfecher: No, `self.inserter.resolve(t)` borrows inserter mutably so we can't also access `self.inserter.function.dfg.data_bus`

compiler/noirc_evaluator/src/ssa/ir/function_inserter.rs
jfecher: Also no, since self.resolve requires a mutable reference to self

compiler/noirc_evaluator/src/ssa/ir/function_inserter.rs
jfecher: Nope.

If we changed `self.resolve` to be immutable (maybe with an internal refcell or similar) we could avoid it though.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
TomAFrench: Ah that's unfortunate.

TomAFrench:
I'm not seeing a big benefit to the peak memory usage here although the heaptrack profiling was done on the `rollup-base-private` circuit which our CI doesn't currently track. I'll profile this again tomorrow to see if we should go ahead and merge.

jfecher:
@TomAFrench I'd think this change would generally benefit runtime speed more than memory use since the extra clones we did before would just get freed afterward anyway. I'm not sure why the time for private-kernel-inner seems to have noticeably increased though.

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -1207,20 +1207,17 @@ mod test {
...
+                let mut constants = vec![];
 
-                let mut values: Vec<_> = values
-                    .into_iter()
-                    .flat_map(|value| get_all_constants_reachable_from_instruction(dfg, value))
-                    .collect();
+                dfg[instruction].for_each_value(|value| {
+                    for constant in get_all_constants_reachable_from_instruction(dfg, value) {
+                        constants.push(constant);
+                    }
```
asterite, line 1215:
```suggestion
                    constants.extend(get_all_constants_reachable_from_instruction(dfg, value));
```

asterite (approved):

PR #6758, opened 2024-12-10
feat: show test duration for tests that took more than 30 seconds

### LINKED ISSUE #6727: Track the time taken to run test functions in `nargo test` and print beside the test output

asterite:
# Description

## Problem

Resolves #6727

## Summary

Snippet from a run with some slow tests:

image

## Documentation

Check one:

# PR Checklist


asterite:
Done in #6728

PR #6757, opened 2024-12-10
fix: Do not merge expressions that contain output witnesses

aakoshh:
# Description

## Problem\*

Resolves #6516 

## Summary\*

Stop the `MergeExpressionOptimizer` from optimising away an opcode that would assign to an output witness.

TODO: Add a test.

## Additional Context

Using this example program for debug:
```rust
pub fn main(in0: Field) -> pub (Field, Field) {
    let out0 = (in0 * in0);
    //let out1 = (in0 + in0);
    let out1 = (in0 * in0);
    (out0, out1)
}
```

Running it with `--force-brillig` assigns all output witnesses:
```console
❯ cargo run -q -p nargo_cli -- --program-dir . execute --force-brillig 
[0_test] Circuit witness successfully solved
[0_test] Circuit output: Vec([Field(100), Field(100)])
[0_test] Witness saved to /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/0_test/target/0_test.gz
```

However with ACIR it only assigns one witness, which is only evident with extra prints added to the program:
```console
❯ cargo run -q -p nargo_cli -- --program-dir . execute
main witness = WitnessMap({Witness(0): 10, Witness(2): 100})
[0_test] Circuit witness successfully solved
[0_test] Witness saved to /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/0_test/target/0_test.gz
```

Adding more prints to show the opcodes:
```console
❯ cargo run -q -p nargo_cli -- --program-dir . execute --print-acir
Compiled ACIR for main (unoptimized):
func 0
current witness index : 2
private parameters indices : [0]
public parameters indices : []
return value indices : [1, 2]
EXPR [ (-1, _0, _0) (1, _1) 0 ]
EXPR [ (-1, _1) (1, _2) 0 ]


SOLVING OPCODE EXPR [ (-1, _0, _0) (1, _2) 0 ]
SOLVING EXPR Expression { mul_terms: [(-1, Witness(0), Witness(0))], linear_combinations: [(1, Witness(2))], q_c: 0 }
SOLVING EVAL Expression { mul_terms: [], linear_combinations: [(1, Witness(2))], q_c: -100 }
INSERTING Witness(2): 100
main witness = WitnessMap({Witness(0): 10, Witness(2): 100})
[0_test] Circuit witness successfully solved
[0_test] Witness saved to /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/0_test/target/0_test.gz
```

[...]

Issue #6516, opened 2024-11-14
Circuit output is not printed after successful execution


LukiMueller:
### Aim

I want to check the result of my program executions but for some special cases the results are not shown although the witness is generated.
Here are some examples when this is the case.

**Examples NOT showing the** `Circuit output:`

~~~rust
pub fn main(in0 : Field) -> pub (Field, Field) {
    let out0 = (in0 * in0);
    let out1 = (in0 * in0);
    (out0, out1)
}

// OR

pub fn main(in0 : Field) -> pub (Field, Field) {
    let out0 = (in0 * in0);
    let out1 = out0;
    (out0, out1)
}
~~~

**Example showing the** `Circuit output:`

~~~rust
pub fn main(in0 : Field) -> pub (Field, Field) {
    let out0 = (in0 * in0);
    let out1 =  (in0 + in0);
    (out0, out1)
}

// OR

pub fn main(in0 : Field) -> pub (Field, Field) {
    let out0 = (in0 * in0);
    let out1 = 0;
    (out0, out1)
}
~~~

In all cases I get `Circuit witness successfully solved`. Older versions seem to give me the expected `Circuit output:` for all of the provided test cases.
Was there a change that removed this for some special cases, do I have to provide a flag now, or is this a bug?

### Expected Behavior

I expected to receive the circuit output `Circuit output: ...`  after the message `Circuit witness successfully solved`

### Bug

For some programs I do no longer receive a `Circuit output: ...`  although `nargo execute` prints  `Circuit witness successfully solved`.

### To Reproduce

1. Create a new `nargo` project
2. Create a file `main.nr` with the content from above and a `Prover.toml` (with e.g. in0="1") 
3. `nargo execute`
3.
4.

### Workaround

None

### Project Impact

None

### Nargo Version

nargo version = 0.38.0 noirc version = 0.38.0+35408ab303f1018c1e2c38e6ea55430a2c89dc4c (git version hash: 35408ab303f1018c1e2c38e6ea55430a2c89dc4c, is dirty: false)

### Would you like to submit a PR for this Issue?

None

TomAFrench:
Hmm, this is odd. I've confirmed that this exists in master.

I think that this is likely caused by the `MergeExpressionOptimizer` doing something it shouldn't as the issue pops up when a single witness is returned twice.

PR #6765, opened 2024-12-10
feat(cli): Verify `return` against ABI and `Prover.toml`

aakoshh:
# Description

## Problem\*

Followup to https://github.com/noir-lang/noir/pull/6757
Related to https://github.com/noir-lang/noir/issues/6516

## Summary\*

Modified the `execute` command to:
* check that if the circuit should return a value according to the ABI, then there is indeed `Some` value we can parse from the `WitnessMap`
* check that if there is a `return` entry in `Prover.toml` then it matches the actual value returned by the circuit

These checks are done after the witness is saved, so we can inspect what went into the file if necessary; the main goal is to be able to use this integration testing, since the ABI parser already looks for the `return` key. `build.rs` will now check return values in `execution_success`, if present. 

Added the test case from #6516 as extra coverage in addition to the unit test in #6757

## Additional Context

Example of running the test with wrong output in `Prover.toml`:

```console
❯ cargo test -q -p nargo_cli --test execute return_twice::forcebrillig_false_inliner_0

running 1 test
F
failures:

---- tests::execution_success::test_return_twice::forcebrillig_false_inliner_0_expects stdout ----
thread 'tests::execution_success::test_return_twice::forcebrillig_false_inliner_0_expects' panicked at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/ops/function.rs:250:5:
Unexpected failure.
code=1
stderr=```"Unexpected return value: expected Vec([Field(100), Field(20)]); got Some(Vec([Field(100), Field(100)]))\n"```
command=`"/Users/aakoshh/Work/aztec/noir/target/debug/nargo" "--program-dir" "/Users/aakoshh/Work/aztec/noir/test_programs/execution_success/return_twice" "execute" "--force" "--inliner-aggressiveness" "0" "--check-return"`
code=1
stdout=
[return_twice] Circuit witness successfully solved
[return_twice] Circuit output: Vec([Field(100), Field(100)])
[return_twice] Witness saved to /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/return_twice/target/return_twice.gz


stderr="Unexpected return value: expected Vec([Field(100), Field(20)]); got Some(Vec([Field(100), Field(100)]))\n"

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    tests::execution_success::test_return_twice::forcebrillig_false_inliner_0_expects

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1664 filtered out; finished in 0.59s

error: test failed, to rerun pass `-p nargo_cli --test execute`
```

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Is this reading necessary to be honest so it's kind of surprised that we didn't error out when ABI decoding already.

If we have a witness map from which we want to decode return values and we find empty witnesses at indices that should hold return values then we should just error and that should avoid the need for this flag I think.

aakoshh:
@TomAFrench the reason it did not complain about the gaps in the witness map is explained [here](https://github.com/noir-lang/noir/blob/f9abf724abd674ea4ccb342a770d237c70864ee1/tooling/noirc_abi/src/lib.rs#L343-L360): 

> Unlike for the circuit inputs, we tolerate not being able to find the witness values for the return value.
> This is because the user may be decoding a partial witness map for which is hasn't been calculated yet.
> If a return value is expected, this should be checked for by the user.

You are right that based on the ABI we can tell if the circuit has any return values. I'm not sure if `execute` can always demand that `abi.decode` returns `Some`; it sounds like fair assumption in the context of the `execute` command 🤷 

But I'd argue that the flag still has value to easily assert the actual value of what is being returned, not just that _something_ is, at least in the context of integration testing, where we use `Prover.toml` for inputs and it readily handles the output as well. At least it can catch inconsistencies like in the `diamond_deps` example.

TomAFrench:
Ok, sorry I didn't read fully as I was on the tube. I'll have a look at this properly tomorrow.

aakoshh:
I added another check to see if the return value is empty when it should be non-empty according to the ABI; all the tests are passing so I suppose that means this is a reasonable expectation. Again doing it after saving the (partial) witness map to file, but could be one earlier if the value cannot be printed.

jfecher:
> Maybe we could compare return in Prover.toml whenever it is present, without any flags?

This sounds good to me. I thought we already compared the return value against the verifier toml by default.

asterite (approved):
Looks great!

PR #6768, opened 2024-12-10
chore(ci): Extend compiler memory report to external repos

vezenovm:
# Description

## Problem\*

Followup to https://github.com/noir-lang/noir/pull/6750

## Summary\*

This report essentially copies the process in #6750 for recording additional benchmarks from external repos. I made two new jobs `external_repo_memory_report` and `upload_memory_report`. I debating making the uploading and the uploading of the compilation report to be one job but I decided it would be better to keep them separate.

## Additional Context

The memory report job for `noir-contracts` takes ~29 minutes with ~9G for peak memory. I commented it out for now due to the time it takes and was not sure we would want to be running this on every job.

I was thinking it may be best to simply generate benchmark reports for external repos only when specified by the label `run-external-checks`, similar to our testing of external repos. 

## Documentation\*

Check one:

# PR Checklist\*


.github/workflows/reports.yml
```diff
@@ -403,3 +395,132 @@ jobs:
...
+          cat memory_report.json
+
+      - name: Parse memory report
+        id: memory_report
+        uses: noir-lang/noir-bench-report@0d7464a8c39170523932d7846b6e6b458a294aea
+        with:
+          report: memory_report.json
+          header: |
+            # Memory Report
+          memory_report: false
```
vezenovm, line 519:
```suggestion
          memory_report: true
```

TomAFrench (approved):

PR #6770, opened 2024-12-11
feat: Fuzzer for noir programs

Rumata888:
# Description

A coverage-guided fuzzer for automatically testing Noir programs. The fuzzer generates and mutates test inputs to find bugs while maximizing code coverage.

## Key Features
- Coverage-guided fuzzing using both ACIR and Brillig execution modes
- Parallel execution support for improved performance
- Maintains and evolves a corpus of high-coverage test cases
- Mutation-based fuzzing with crossover capabilities
- Detects discrepancies between ACIR and Brillig execution modes
- Supports corpus minimization
- Can be used with an oracle for differential fuzzing
- Configurable failure conditions via `only_fail_with` attribute

## Documentation\*

Check one:

# PR Checklist\*


tooling/greybox_fuzzer/src/Readme.md
```diff
@@ -0,0 +1,116 @@
+# Noir Greybox Fuzzer
```
asterite, line 1: Could this file be moved one level up? I think it's usual for Readme.md to be at the root of a project.

tooling/nargo_cli/src/cli/fuzz_cmd.rs
```diff
@@ -0,0 +1,489 @@
...
+use noirc_driver::{check_crate, CompileOptions};
+use noirc_frontend::hir::{FunctionNameMatch, ParsedFiles};
+use rayon::prelude::{ParallelBridge, ParallelIterator};
+use termcolor::{Color, ColorChoice, ColorSpec, StandardStream, WriteColor};
+
+use crate::{cli::check_cmd::check_crate_and_report_errors, errors::CliError};
+
+use super::{fs::inputs::write_inputs_to_file, LockType, PackageOptions, WorkspaceCommand};
+
+/// Run the tests for this program
```
asterite, line 26: This description should be updated.

tooling/nargo/src/ops/fuzz.rs
```diff
@@ -0,0 +1,349 @@
...
+        // Good for us, run fuzzer
+        (Ok(acir_program), Ok(brillig_program)) => {
+            #[cfg(target_arch = "wasm32")]
+            {
+                // We currently don't support fuzz testing on wasm32 as the u128 strategies do not exist on this platform.
+                FuzzingRunStatus::ExecutionFailure {
+                    message: "Fuzz tests are not supported on wasm32".to_string(),
+                    error_diagnostic: None,
+                }
+            }
```
TomAFrench, line 129: We do have u128 strategies on wasm now afaik

TomAFrench:
btw, @Rumata888 and I spoke in DMs about how this fuzzer relates to the existing one. As it's a direct improvement on the current fuzzer the idea is that we can replace the old one so `nargo test` will make use of the greybox fuzzer.

For the sake of simplicity we don't need to clear out the old fuzzer as part of this PR. However we may want to have the greybox fuzzer target any `#[test]` functions with arguments in the same way as the old fuzzer.

acvm-repo/brillig_vm/src/lib.rs
```diff
@@ -66,6 +69,10 @@ pub enum VMStatus<F> {
 // A sample for each opcode that was executed.
 pub type BrilligProfilingSamples = Vec<BrilligProfilingSample>;
 
+pub type Branch = (usize, usize);
```
asterite, line 72: Could these two `usize` be documented? Alternative, maybe it could be a struct so the meaning is immediately clear.

acvm-repo/brillig_vm/src/lib.rs
```diff
@@ -66,6 +69,10 @@ pub enum VMStatus<F> {
...
+// A map for translating encountered branching logic to features for fuzzing
+pub type BranchToFeatureMap = HashMap<Branch, usize>;
```
asterite, line 74: Is this `usize` an index into something?

TomAFrench:
> How will #[fuzz(only_fail_with = "This is the message that will be checked for")] be supported inside the test attribute?

`#[test(should_fail_with = "This is the message that will be checked for")]`

We already have this used here:

https://github.com/noir-lang/noir/blob/1fa9b33aab796dbc2a61f3062bf80e120831b462/test_programs/noir_test_success/assert_message/src/main.nr#L4

asterite:
I thought about reusing that attribute, but I think it's a bit different: in a test it expects it to always fail with that message. When fuzzing it's the only message we expect to fail with if fuzzing triggers a failure.

I guess it could be something like this:

```noir
#[test(only_fail_fuzz_if = "...")]`
```

acvm-repo/brillig_vm/src/lib.rs
```diff
@@ -228,7 +258,209 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>> VM<'a, F, B> {
 
         self.process_opcode_internal()
     }
+    fn trace_branching(&mut self, destination: usize) {
+        if self.fuzzing_active {
```
asterite, line 262: We could do:

```rust
if !self.fuzzing_active {
  return;
}
```

in all of these to return the indentation level.

acvm-repo/brillig_vm/src/lib.rs
```diff
@@ -228,7 +258,209 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>> VM<'a, F, B> {
...
+                | BinaryIntOp::Or
+                | BinaryIntOp::Xor
+                | BinaryIntOp::Shl
+                | BinaryIntOp::Shr => {}
+                BinaryIntOp::Equals | BinaryIntOp::LessThan | BinaryIntOp::LessThanEquals => {
+                    let lhs_value: u128;
+                    let rhs_value: u128;
+                    match bit_size {
+                        IntegerBitSize::U1 => {
+                            lhs_value = match lhs.expect_u1() {
```
asterite, line 349: Would it make sense to introduce a `MemoryValue::to_u128()` method? That said, this code seems to also be checking that the types of the arguments matches `bit_size` so using a method like `to_u128` would lose that, but I don't know if we need to do these assertions (maybe the binary op already checks this when it's built)

acvm-repo/brillig_vm/src/lib.rs
```diff
@@ -228,7 +258,209 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>> VM<'a, F, B> {
...
+                        [&(self.program_counter, if c { usize::MAX } else { usize::MAX - 1 })];
+                    self.fuzzer_trace[condition_index] += 1;
+                    self.fuzzer_trace[approach_index] += 1;
+                }
+            }
+        }
+    }
+    pub fn get_fuzzing_trace(&self) -> Vec<u32> {
+        if !self.fuzzing_active {
+            Vec::new()
```
asterite, line 459: I think in this case `self.fuzzer_trace` will be an empty Vec so we could simplify this method to always return `self.fuzzer_trace.clone()`

test_programs/fuzzing_failure/branching/src/main.nr
```diff
@@ -0,0 +1,5 @@
+// This program will fail when the input is 0xdeadbeef
```
asterite, line 1: Nice! 😄

test_programs/fuzzing_failure/comparison/src/main.nr
```diff
@@ -0,0 +1,7 @@
+// This program will fail when there is a specific difference between the two fields
+#[fuzz]
+fn comparison(a: Field, b: Field) {
+    let c = b + 0xffffffffffff0000000000;
+    let d = c - 0xeeeeeeeeeeeeeeeee;
+    assert((a !=d) | (b == 0) | (a == 0));
```
asterite, line 6: It would destroy the compactness of some of these programs, but we should probably have the formatter run through these files (it must be added in `test_programs/format.sh`)

tooling/nargo_cli/build.rs
```diff
@@ -211,6 +212,40 @@ fn test_{test_name}(force_brillig: ForceBrillig, inliner_aggressiveness: Inliner
     .expect("Could not write templated test file.");
 }
 
+/// Generate all test cases for a given test name (expected to be unique for the test directory),
+/// based on the matrix configuration. These will be executed serially, but concurrently with
+/// other test directories. Running multiple tests on the same directory would risk overriding
+/// each others compilation artifacts, which is why this method injects a mutex shared by
+/// all cases in the test matrix, as long as the test name and directory has a 1-to-1 relationship.
```
asterite, line 219: In any case, maybe the entire description should be updated to be about fuzzing.

tooling/nargo_fmt/src/formatter/attribute.rs
```diff
@@ -136,6 +140,29 @@ impl<'a> Formatter<'a> {
         self.write_right_bracket(); // ]
     }
 
+    fn format_fuzz_attribute(&mut self, fuzz_scope: FuzzingScope) {
+        self.write_current_token_and_bump(); // #[
+        self.skip_comments_and_whitespace();
+        self.write_current_token_and_bump(); // test
```
asterite, line 146:
```suggestion
        self.write_current_token_and_bump(); // fuzz
```

though if they `fuzz` attribute is going to be removed then it doesn't matter.

Rumata888:
> I thought about reusing that attribute, but I think it's a bit different: in a test it expects it to always fail with that message. When fuzzing it's the only message we expect to fail with if fuzzing triggers a failure.
> 
> I guess it could be something like this:
> 
> ```nr
> #[test(only_fail_fuzz_if = "...")]`
> ```

Yes, the fuzz tests don't have a `should_fail`  attribute and are in general expected not to target failure, so it would take some work to converge both methods in this PR

acvm-repo/acvm/src/pwg/mod.rs
```diff
@@ -244,6 +256,24 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>> ACVM<'a, F, B> {
         self.profiling_active = profiling_active;
     }
 
+    // Enable brillig fuzzing
+    pub fn with_brillig_fuzzing(
+        &mut self,
+        brillig_fuzzing_active: bool,
+        brillig_branch_to_feature_map: Option<&'a BranchToFeatureMap>,
+    ) {
```
asterite, line 264: This could be:

```suggestion
    pub fn with_brillig_fuzzing(
        &mut self,
        brillig_branch_to_feature_map: &'a BranchToFeatureMap,
    ) {
```

That if, when this is called it's because we want fuzzing active, and the map must be a `Some` value.

Then, maybe `brillig_fuzzing_active` isn't needed anymore, we could know if it's active or not based on whether the map is None or Some.

tooling/nargo/src/ops/execute.rs
```diff
@@ -228,15 +258,45 @@ pub fn execute_program_with_profiling<
...
+pub(crate) fn execute_program_with_brillig_fuzzing<
+    F: AcirField,
+    B: BlackBoxFunctionSolver<F>,
+    E: ForeignCallExecutor<F>,
+>(
+    program: &Program<F>,
+    initial_witness: WitnessMap<F>,
+    blackbox_solver: &B,
+    foreign_call_executor: &mut E,
+    brillig_branch_to_feature_map: Option<&BranchToFeatureMap>,
```
asterite, line 281: Related to another comment about this, if this method enables fuzzing then the map should also be present (and in fact we only call this method with `Some`, never with `None`):

```suggestion
    brillig_branch_to_feature_map: &BranchToFeatureMap,
```

tooling/nargo/src/ops/execute.rs
```diff
@@ -58,9 +70,24 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>, E: ForeignCallExecutor<F>>
...
+            brillig_fuzzing_active: false,
+            brillig_branch_to_feature_map: None,
+            last_fuzzing_trace: None,
         }
     }
 
+    fn with_brillig_fuzzing(
+        &mut self,
+        brillig_fuzzing_active: bool,
+        brillig_branch_to_feature_map: Option<&'a BranchToFeatureMap>,
```
asterite, line 82: Same here, this could just receive `&'a BranchToFeatureMap`

tooling/greybox_fuzzer/src/corpus.rs
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    /// Returns a vector containing all parsed inputs from the corpus
+    pub fn get_full_corpus(&self) -> Vec<&InputMap> {
+        let file_ids = self.file_manager.as_file_map().all_file_ids();
+        let mut full_corpus = Vec::new();
+        for file_id in file_ids {
+            full_corpus.push(&self.parsed_map[file_id]);
+        }
+        full_corpus
```
asterite, line 143:
```suggestion
        file_ids.map(|file_id| &self.parsed_map[file_id]).collect()
```

tooling/greybox_fuzzer/src/corpus.rs
```diff
@@ -0,0 +1,431 @@
...
+        let mut builder = DirBuilder::new();
+        match builder.recursive(true).create(&self.corpus_path) {
+            Ok(_) => (),
+            Err(_) => {
+                return Err(format!(
+                    "Couldn't create directory {:?}",
+                    self.corpus_path.as_os_str()
+                ));
+            }
+        }
```
asterite, line 164: This could be moved into a helper method (it also exists in `load_corpus_from_disk`)

tooling/greybox_fuzzer/src/corpus.rs
```diff
@@ -0,0 +1,431 @@
...
+    pub fn new_testcase(&mut self, testcase_id: TestCaseId) {
+        self.executions_per_testcase.insert(testcase_id, 0);
+        self.sequence_number.insert(testcase_id, 0);
+    }
+
+    /// Removes a testcase from being managed by the orchestrator.
+    /// Updates execution counts and clears current sequence if it was using this testcase.
+    pub fn remove(&mut self, testcase_id: TestCaseId) {
+        let executions = self.executions_per_testcase[&testcase_id];
+        self.executions_per_testcase.remove(&testcase_id);
```
asterite, line 247:
```suggestion
        let executions = self.executions_per_testcase.remove(&testcase_id).unwrap();
```

or `expect` instead of `unwrap`

tooling/greybox_fuzzer/src/corpus.rs
```diff
@@ -0,0 +1,431 @@
...
+
+            // Create new sequence with exponentially increasing length based on sequence number
+            self.current_sequence = Sequence {
+                testcase_id: chosen_id,
+                executions_left: min(1u64 << self.sequence_number[&chosen_id], average / 2),
+            };
+
+            // Update execution counts
+            self.total_executions += 1;
+            self.executions_per_testcase.entry(chosen_id).and_modify(|executions| *executions += 1);
```
asterite, line 293: It seems the first statement could be moved outside of this `if`.

Then, at this point using `chosen_id` or `self.current_sequence.testcase_id` is the same, so this similar line from the other branch could also be moved outside of this `if`.

tooling/greybox_fuzzer/src/corpus.rs
```diff
@@ -0,0 +1,431 @@
...
+            .collect()
+    }
+
+    /// Returns the path where corpus files are stored
+    pub fn get_corpus_storage_path(&self) -> &Path {
+        self.corpus_file_manager.get_corpus_path()
+    }
+
+    /// Adds a testcase to the cache (stored but not active)
+    pub fn insert_into_cache(&mut self, testcase_id: TestCaseId, new_testcase_value: InputMap) {
```
asterite, line 372: It seems this doesn't need to be `pub`

tooling/greybox_fuzzer/src/corpus.rs
```diff
@@ -0,0 +1,431 @@
...
+        self.discovered_testcases.remove(&testcase_id);
+    }
+
+    /// Returns the input values for a testcase by its ID, checking both active and cached testcases
+    pub fn get_testcase_by_id(&self, id: TestCaseId) -> &InputMap {
+        if self.discovered_testcases.contains_key(&id) {
+            &self.discovered_testcases[&id]
+        } else {
+            &self.cached_testcases[&id]
+        }
```
asterite, line 409:
```suggestion
        self.discovered_testcases.get(&id).unwrap_or_else(|| &self.cached_testcases[&id])
```

tooling/nargo/src/ops/execute.rs
```diff
@@ -228,15 +258,45 @@ pub fn execute_program_with_profiling<
...
-    execute_program_inner(
+    match execute_program_inner(
         program,
         initial_witness,
         blackbox_solver,
         foreign_call_executor,
         profiling_active,
-    )
+    ) {
+        Ok((witness_stack, profiling_samples, ..)) => Ok((witness_stack, profiling_samples)),
```
asterite, line 268: It seems `execute_program_inner` returns a tuple of two elements, so this `match` and mapping shouldn't be necessary.

tooling/nargo_cli/src/cli/fuzz_cmd.rs
```diff
@@ -0,0 +1,489 @@
...
+                )));
+            }
+            FunctionNameMatch::Contains(pattern) => {
+                let single_pattern = pattern[0].clone();
+                return Err(CliError::Generic(format!(
+                    "Found 0 fuzzing_harnesses containing '{single_pattern}'.",
+                )));
+            }
+            // If we are running all tests in a crate, having none is not an error
+            FunctionNameMatch::Anything => {}
```
asterite, line 181: Should it be an error for fuzzing? Maybe someone used `#[test]` instead of `#[fuzz]`, ran the fuzzer, it produced no errors so they would think there's no issue.

tooling/nargo_cli/src/cli/fuzz_cmd.rs
```diff
@@ -0,0 +1,489 @@
...
+            if args.exact {
+                FunctionNameMatch::Exact(names)
+            } else {
+                FunctionNameMatch::Contains(names)
+            }
+        }
+        None => FunctionNameMatch::Anything,
+    };
+
+    if args.list_all {
```
asterite, line 100: The contents of this `if` could be moved to a helper function.

tooling/nargo_cli/src/cli/fuzz_cmd.rs
```diff
@@ -0,0 +1,489 @@
...
+        );
+        fuzzing_reports.push((fuzzing_harness_name, status));
+        // Display the latest report
+        display_fuzzing_report_and_store(
+            root_path.clone(),
+            fuzz_folder_config.fuzzing_failure_dir.clone(),
+            file_manager,
+            package,
+            compile_options,
+            &fuzzing_reports[(&fuzzing_reports.len() - 1)..fuzzing_reports.len()],
```
asterite, line 254: It seems this will always be one element? If so, could `display_fuzzing_report_and_store` be changed to take just one of these?

tooling/nargo_cli/src/cli/fuzz_cmd.rs
```diff
@@ -0,0 +1,489 @@
...
+    let mut writer = writer.lock();
+    let fuzzing_failure_path =
+        fuzzing_failure_folder.map(PathBuf::from).unwrap_or(root_path.clone().unwrap_or_default());
+    if !fuzzing_failure_path.exists() {
+        std::fs::create_dir_all(&fuzzing_failure_path)
+            .expect("Failed to create fuzzing failure directory");
+    }
+
+    for (fuzzing_harness_name, test_status) in fuzzing_report {
+        write!(writer, "[").expect("Failed to write to stderr");
```
asterite, line 335: Given how unlikely these are to fail, we could use `write!(...)?;` here and everywhere to reduce the noise and make the code clearer (the test_cmd code used to be like this but it was eventually changed).

tooling/nargo_cli/src/cli/fuzz_cmd.rs
```diff
@@ -0,0 +1,489 @@
...
+                        abi,
+                        input_map,
+                    )
+                    .expect("Couldn't write toml file");
+                    writeln!(writer, "saved input to:").expect("Failed to write to stderr");
+                    writer
+                        .set_color(ColorSpec::new().set_fg(Some(Color::Yellow)))
+                        .expect("Failed to set color");
+                    let mut full_path_of_example = fuzzing_failure_path.clone().join(file_name);
+                    full_path_of_example.set_extension(PathBuf::from("toml"));
```
asterite, line 432: After this, we could try to make `full_path_of_example` relative to the current directory so the output is a bit shorter. This is done for example here: https://github.com/noir-lang/noir/blob/7a9c8c13f626575f23d25f9cb8689e44dc1c6116/compiler/fm/src/file_map.rs#L91-L96

asterite:
I didn't take a look at how the fuzzer does for finding inputs that fail so quickly, but I just tried it with this program:

```noir
#[fuzz]
fn fuzz_equality(a: i64) {
    // This is always true
    assert(a <= 9223372036854775807);
}
```

and `nargo fuzz` never finishes (it keeps printing `LOOP` lines). Is this expected?

asterite:
Ah, nevermind, maybe it's expected because the default is "no timeout".

tooling/greybox_fuzzer/src/types.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub case: InputMap,
+
+    /// Resulting witness (only available if the acir program has been run)
+    pub witness: Option<WitnessStack<FieldElement>>,
+
+    /// Coverage from brillig execution (only available if  brillig program has been run)
+    pub brillig_coverage: Option<Vec<u32>>,
+
+    /// How much time executing the acir program took (0 if it hasn't been run)
+    pub acir_time: u128,
```
asterite, line 53: Maybe include the time unit in the name? Alternatively this and the next field could be a `Duration`.

tooling/nargo/src/ops/fuzz.rs
```diff
@@ -0,0 +1,349 @@
...
+        message: String,
+    },
+    CompileError(FileDiagnostic),
+}
+
+impl FuzzingRunStatus {
+    pub fn failed(&self) -> bool {
+        !matches!(
+            self,
+            FuzzingRunStatus::ExecutionPass | FuzzingRunStatus::MinimizationFailure { .. }
```
asterite, line 67: Is `MinimizationPass` a failure?

If not, it might be good to `match` over self so we can make sure that if a new variant is added, it's handled here.

tooling/nargo/src/ops/fuzz.rs
```diff
@@ -0,0 +1,349 @@
...
+        .function_meta(&fuzzing_harness.get_id())
+        .function_signature()
+        .0
+        .is_empty();
+
+    if fuzzing_harness_has_no_arguments {
+        return FuzzingRunStatus::ExecutionFailure {
+            message: ("Fuzzing harness has no arguments".to_owned()),
+            counterexample: (None),
+            error_diagnostic: (None),
```
asterite, line 98:
```suggestion
            message: "Fuzzing harness has no arguments".to_owned(),
            counterexample: None,
            error_diagnostic: None,
```

tooling/nargo/src/ops/fuzz.rs
```diff
@@ -0,0 +1,349 @@
...
+        compile_no_check(context, &acir_config, fuzzing_harness.get_id(), None, false);
+    let acir_program_copy = if let Ok(acir_program_internal) = &acir_program {
+        Some(acir_program_internal.clone())
+    } else {
+        None
+    };
+    let brillig_program =
+        compile_no_check(context, &brillig_config, fuzzing_harness.get_id(), None, false);
+    let brillig_program_copy = if let Ok(brillig_progam_internal) = &brillig_program {
+        Some(brillig_progam_internal.clone())
```
asterite, line 115:
```suggestion
    let brillig_program_copy = if let Ok(brillig_program_internal) = &brillig_program {
        Some(brillig_program_internal.clone())
```

tooling/nargo/src/ops/fuzz.rs
```diff
@@ -0,0 +1,349 @@
...
+                        #[cfg(not(feature = "rpc"))]
+                        let build_foreign_call_executor = |output, base| {
+                            DefaultForeignCallBuilder { output, enable_mocks: true }
+                                .build_with_base(base)
+                        };
+                        // Use a base layer that doesn't handle anything, which we handle in the `execute` below.
+                        let inner_executor = build_foreign_call_executor(
+                            if show_output { PrintOutput::Stdout } else { PrintOutput::None },
+                            layers::Unhandled,
+                        );
```
asterite, line 203: If this is the same as before it could be extracted to a helper function. I actually tried it because sometimes in Rust I get confused with lifetimes and abandon some refactors, but this worked:

```rust
fn build_foreign_call_executor<
    'a,
    F: AcirField + Serialize + for<'de> Deserialize<'de> + 'a + 'static,
>(
    show_output: bool,
    foreign_call_resolver_url: Option<&str>,
    root_path: Option<PathBuf>,
    package_name: Option<String>,
) -> impl ForeignCallExecutor<F> {
    #[cfg(feature = "rpc")]
    let build_foreign_call_executor = |output, base| {
        DefaultForeignCallBuilder {
            output,
            enable_mocks: true,
            resolver_url: foreign_call_resolver_url.map(|s| s.to_string()),
            root_path: root_path.clone(),
            package_name: package_name.clone(),
        }
        .build_with_base(base)
    };
    #[cfg(not(feature = "rpc"))]
    let build_foreign_call_executor = |output, base| {
        DefaultForeignCallBuilder { output, enable_mocks: true }.build_with_base(base)
    };
    // Use a base layer that doesn't handle anything, which we handle in the `execute` below.
    let inner_executor = build_foreign_call_executor(
        if show_output { PrintOutput::Stdout } else { PrintOutput::None },
        layers::Unhandled,
    );

    TestForeignCallExecutor::new(inner_executor)
}
```

tooling/nargo/src/ops/fuzz.rs
```diff
@@ -0,0 +1,349 @@
...
+                    brillig_executor,
+                    &package_name.clone().unwrap(),
+                    context.def_interner.function_name(&fuzzing_harness.get_id()),
+                    FuzzedExecutorExecutionConfiguration {
+                        num_threads: fuzz_execution_config.num_threads,
+                        timeout: fuzz_execution_config.timeout,
+                    },
+                    FuzzedExecutorFailureConfiguration {
+                        fail_on_specific_asserts: fuzzing_harness.only_fail_enabled(),
+                        failure_reason: fuzzing_harness.failure_reason(),
```
asterite, line 243: It seems that if `failure_reason()` is `Some` then `only_fail_enabled()` will be `true`, so carrying these two things is a bit redundant. This could be simplified to just carry `failure_reason` (dropping the bool).

tooling/nargo/src/ops/fuzz.rs
```diff
@@ -0,0 +1,349 @@
...
+                        #[cfg(not(feature = "rpc"))]
+                        let build_foreign_call_executor = |output, base| {
+                            DefaultForeignCallBuilder { output, enable_mocks: true }
+                                .build_with_base(base)
+                        };
+                        // Use a base layer that doesn't handle anything, which we handle in the `execute` below.
+                        let inner_executor = build_foreign_call_executor(
+                            if show_output { PrintOutput::Stdout } else { PrintOutput::None },
+                            layers::Unhandled,
+                        );
```
asterite, line 281: This too could be deduplicated using that helper function.

tooling/greybox_fuzzer/src/mutation/mod.rs
```diff
@@ -0,0 +1,769 @@
...
+    second_testcase_weight: usize,
+    total_weight: usize,
+}
+
+impl UnbalancedSpliceConfiguration {
+    #[allow(unused)]
+    pub fn new(first_testcase_weight: usize, second_testcase_weight: usize) -> Self {
+        let total_weight = first_testcase_weight + second_testcase_weight;
+        Self { first_testcase_weight, second_testcase_weight, total_weight }
+    }
```
asterite, line 80: Can this be removed if it's unused? Same goes for `second_testcase_weight`.

tooling/greybox_fuzzer/src/mutation/mod.rs
```diff
@@ -0,0 +1,769 @@
...
+const STRING_WEIGHT_MULTIPLIER: u32 = 4;
+impl InputMutator {
+    /// Create a new input mutator given the ABI and a hashset of field elements used in the circuit (for the dictionary)
+    pub fn new(abi: &Abi, original_dictionary: &HashSet<FieldElement>) -> Self {
+        let mut weight_tree = Self::count_all_input_weights(abi);
+        weight_tree.calculate_offsets(0);
+        Self {
+            abi: abi.clone(),
+            weight_tree,
+            full_dictionary: FullDictionary::new(original_dictionary),
```
asterite, line 109: `FullDictionary::new(original_dictionary)` could be extracted to `let full_dictionary = ...` and it will reduce the number of lines here.

tooling/greybox_fuzzer/src/mutation/mod.rs
```diff
@@ -0,0 +1,769 @@
...
+        }
```
asterite, line 110:
```suggestion
        let full_dictionary = FullDictionary::new(original_dictionary);
        Self { abi: abi.clone(), weight_tree, full_dictionary }
```

tooling/greybox_fuzzer/src/mutation/mod.rs
```diff
@@ -0,0 +1,769 @@
...
+            AbiType::Array { length, typ } => {
+                let length = *length as usize;
+                let subnode_weight = Self::count_single_input_weight(typ);
+                let node_weight = subnode_weight.get_weight() * length as u32;
+
+                NodeWeight {
+                    start: 0,
+                    end: node_weight,
+                    subnodes: Some(vec![subnode_weight; length]),
+                }
```
asterite, line 137:
```suggestion
                let subnodes = Some(vec![subnode_weight; length]);
                NodeWeight { start: 0, end: node_weight, subnodes }
```

tooling/greybox_fuzzer/src/mutation/dictionary.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub fn new(original_dictionary: &[FieldElement]) -> Self {
+        Self { width_dictionaries: Self::filter_dictionary_by_width(original_dictionary) }
+    }
+    pub fn get_dictionary_by_width(&self, width: u32) -> &Vec<FieldElement> {
+        match width {
+            8 => &self.width_dictionaries[0],
+            16 => &self.width_dictionaries[1],
+            32 => &self.width_dictionaries[2],
+            64 => &self.width_dictionaries[3],
+            _ => panic!("Only widths 8, 16, 32, 64 are supported"),
```
asterite, line 25: We could allow 128 now that `u128` was recently added.

tooling/greybox_fuzzer/src/mutation/mod.rs
```diff
@@ -0,0 +1,769 @@
...
+                    self.full_dictionary.get_int_dictionary(),
+                ),
+                None,
+            ),
+            AbiType::Array { length, typ } => {
+                let length = *length as usize;
+                let input_vector = match previous_input {
+                    InputValue::Vec(previous_input_vector) => previous_input_vector,
+                    _ => panic!("Mismatch of AbiType and InputValue should not happen"),
+                };
```
asterite, line 249: nit:

```suggestion
                let InputValue::Vec(input_vector) = previous_input else {
                    panic!("Mismatch of AbiType and InputValue should not happen")
                };
```

asterite: This can also be done in the `Struct` and `Tuple` cases.

tooling/greybox_fuzzer/src/mutation/string.rs
```diff
@@ -0,0 +1,290 @@
...
+        let insertion_count = self.prng.gen_range(1..=(buffer_length - insertion_position));
+
+        // Pick value
+        let value = self.prng.gen_range(MIN_ASCII..=MAX_ASCII);
+        for item in result.iter_mut().skip(insertion_position).take(insertion_count) {
+            *item = value;
+        }
+        result
+    }
+    /// Create a spliced version of 2 buffers, where each element in the result is at the same index as in the original ones
```
asterite, line 180: Could you include an example of this? That is, an example first_buffer and second_buffer, with one possible outcome.

I also think it would be good to include small examples like this in all mutations so it's easier to understand what they do without having to read the code.

tooling/greybox_fuzzer/src/mutation/mod.rs
```diff
@@ -0,0 +1,769 @@
...
+    }
+
+    /// Use one of splicing mechanisms to produce a new testcase generated by merging two previous ones
+    pub fn splice_two_maps(
+        &self,
+        first_input_map: &InputMap,
+        second_input_map: &InputMap,
+        prng: &mut XorShiftRng,
+    ) -> InputMap {
+        // TODO: add the most basic splice where the whole map is converted into a witness map and then we splice those
```
asterite, line 629: Maybe capture a follow-up issue for this (or drop the TODO if it's not needed anymore)

tooling/greybox_fuzzer/src/mutation/string.rs
```diff
@@ -0,0 +1,290 @@
...
+
+struct StringMutator<'a> {
+    dictionary: &'a Vec<FieldElement>,
+    prng: &'a mut XorShiftRng,
+    value_mutation_configuration: &'static ByteValueMutationConfiguration,
+}
+
+impl<'a> StringMutator<'a> {
+    pub fn new(dictionary: &'a IntDictionary, prng: &'a mut XorShiftRng) -> Self {
+        let u8_dictionary = dictionary.get_dictionary_by_width(8);
```
asterite, line 43: Would it make sense for the dictionary here to only consist of bytes that are ascii?

tooling/greybox_fuzzer/src/mutation/configurations.rs
```diff
@@ -0,0 +1,874 @@
...
+            return IntTopLevelMutation::SmallValueUpdate;
+        }
+        IntTopLevelMutation::DictionaryValueUpdate
+    }
+}
+
+/// Default configurations for all mutations that are currently used
+
+pub(crate) const BASIC_SPLICE_MUTATION_CONFIGURATION: SpliceMutationConfiguration =
+    SpliceMutationConfiguration {
```
asterite, line 728: Would it make sense to use `::new` here? It requires `new` to be `const` but it works, and then we don't have to manually compute `total_weight` (though I don't know if the reason for now doing that is making it more clear what the numbers mean)

tooling/greybox_fuzzer/src/mutation/array.rs
```diff
@@ -0,0 +1,201 @@
...
+    pub fn new(prng: &'a mut XorShiftRng) -> Self {
+        Self { prng }
+    }
+
+    /// Create a spliced version of 2 buffers, where each element in the result is at the same index as in the original ones
+    fn structured_splice(
+        &mut self,
+        first_buffer: &[InputValue],
+        second_buffer: &[InputValue],
+    ) -> Vec<InputValue> {
```
asterite, line 27: I wonder if this one and the one in string could be done by calling a generic helper function with `T` being `InputValue` or `u8` (the logic seems to be the same)

asterite: Same goes for `swap` and `duplicate_chunk`

tooling/greybox_fuzzer/src/mutation/array.rs
```diff
@@ -0,0 +1,201 @@
...
+            StructuralMutation::ChunkDuplication => self.duplicate_chunk(input_buffer),
+            StructuralMutation::Swap => self.swap(input_buffer),
+            StructuralMutation::RandomValueDuplication => {
+                panic!("Vector mutations should have a value duplication weight of zero")
+            }
+        }
+    }
+
+    /// Swap 2 random chunks in the buffer
+    fn swap(&mut self, buffer: &Vec<InputValue>) -> Vec<InputValue> {
```
asterite, line 126: This method could be tested by having a separate function that receives all the values that are generated randomly (mainly to have a small proof that it works correctly).

tooling/greybox_fuzzer/src/mutation/int.rs
```diff
@@ -0,0 +1,558 @@
...
+    let mut initial_i128 = input.to_i128();
+    if initial_i128 >= 2i128.pow(width - 1) {
+        initial_i128 -= 2i128.pow(width);
+    }
+    initial_i128
+}
+
+/// Convert a value from integer representation to field representation
+/// Width is needed to detect negative values in 2-complement form
+fn i128_to_field(value: i128, width: u32) -> FieldElement {
```
asterite, line 288: Could use `mut value` so there's no need for `new_value`.

tooling/greybox_fuzzer/src/mutation/int.rs
```diff
@@ -0,0 +1,558 @@
...
+    let mut new_value = value;
+    if value < 0 {
+        new_value += 2i128.pow(width);
+    }
+    FieldElement::from(new_value)
+}
+
+/// Generate a random unsigned integer of given width and convert to field
+fn generate_random_for_width(prng: &mut XorShiftRng, width: u32) -> FieldElement {
+    FieldElement::from(prng.gen_range(0..(2i128 << width)))
```
asterite, line 298: Is this 2 intentional here or should it be 1?

tooling/greybox_fuzzer/src/mutation/configurations.rs
asterite: Actually, nevermind. After seeing a few of these I understand now that the unused value is the last case that will be considered, and it's better to have explicit names here (in other cases it's more than just two possibilities).

tooling/greybox_fuzzer/src/coverage.rs
```diff
@@ -0,0 +1,751 @@
...
+
+        // Go through the list of perceived boolean witnesses
+        for witness_index in self.witness.iter().copied() {
+            let value = first_func_witnesses
+                .witness
+                .get(&witness_index)
+                .expect("There should be a witness in the witness map");
+
+            // Check that the values are zero or 1
+            if *value != FieldElement::zero() && *value != FieldElement::one() {
```
asterite, line 77: For consistency we could use `is_zero`/`is_one` here

tooling/greybox_fuzzer/src/coverage.rs
```diff
@@ -0,0 +1,751 @@
...
+    /// Generate a bool witness list by parsing the witnesses in the program
+    fn from(witness_stack: &WitnessStack<FieldElement>) -> Self {
+        let mut witness_set = HashSet::new();
+        // Should be only one function
+        assert!(witness_stack.length() == 1);
+        let first_func_witnesses = witness_stack.peek().unwrap();
+
+        // Look for witnesses that are either 0 or 1
+        for (witness_index, value) in first_func_witnesses.witness.clone().into_iter() {
+            if value == FieldElement::one() || value == FieldElement::zero() {
```
asterite, line 50: For consistency we could use `is_zero`/`is_one` here

tooling/greybox_fuzzer/src/coverage.rs
```diff
@@ -0,0 +1,751 @@
...
+            add_to_leavers(Some(self.bool_state_to_testcase_id[&state]));
+        }
+    }
+
+    /// Returns true if there is new coverage in the presented testcase
+    pub fn detect_new_coverage(&self, new_coverage: &SingleTestCaseCoverage) -> bool {
+        // Go through all single branch coverage ranges and check that either:
+        // 1. A new branch is taken
+        // 2. A branch is taken more times than ever encountered before in a single execution
+        // 3. A branch is taken pow2 times that hasn't been previously observed
```
asterite, line 564: Are 2 and 3 dependent on 1 here?

tooling/greybox_fuzzer/src/coverage.rs
```diff
@@ -0,0 +1,751 @@
...
+                continue;
+            }
+            let mut least_different_bits = u32::MAX;
+            let mut last_value = 0;
+            for i in 0..(cmp_approach.bits + 1) {
+                if !new_coverage.brillig_coverage[i + cmp_approach.raw_index].is_zero() {
+                    least_different_bits = (cmp_approach.bits - i) as u32;
+                    last_value = new_coverage.brillig_coverage[i + cmp_approach.raw_index];
+                }
+            }
```
asterite, line 596: Could this use `find_closest_comparison`?

tooling/greybox_fuzzer/src/coverage.rs
```diff
@@ -0,0 +1,751 @@
...
+                feature_to_index_map.insert((opcode_index, opcode_index + 1), total_features + 1);
+                coverage_items.push(BrilligCoverageItemRange::Branch(BranchCoverageRange {
+                    index: total_features,
+                }));
+                total_features += 2;
+            }
+            // Conditional mov
+            &BrilligOpcode::ConditionalMov { .. } => {
+                feature_to_index_map.insert((opcode_index, usize::MAX - 1), total_features);
+                feature_to_index_map.insert((opcode_index, usize::MAX), total_features + 1);
```
asterite, line 666: I looked at the tracing code first and couldn't understand why `usize::MAX` and `usize::MAX - 1` were used there, but now seeing this code I understand. I wonder if these could be extracted into documented constants so that it's easier to see where they come from and what they mean.

tooling/greybox_fuzzer/src/coverage.rs
```diff
@@ -0,0 +1,751 @@
...
+                acvm::acir::brillig::BinaryFieldOp::Add
+                | acvm::acir::brillig::BinaryFieldOp::Sub
+                | acvm::acir::brillig::BinaryFieldOp::Mul
+                | acvm::acir::brillig::BinaryFieldOp::Div
+                | acvm::acir::brillig::BinaryFieldOp::IntegerDiv => {}
+                // Equality and LessThan(Equals) are comparisons that interest us
+                acvm::acir::brillig::BinaryFieldOp::Equals
+                | acvm::acir::brillig::BinaryFieldOp::LessThan
+                | acvm::acir::brillig::BinaryFieldOp::LessThanEquals => {
+                    let features_per_comparison= 1 /*true */+1/*false */+255 /*possible bits() results*/;
```
asterite, line 683: Should this be 254?

tooling/greybox_fuzzer/src/coverage.rs
```diff
@@ -0,0 +1,751 @@
...
+                | acvm::acir::brillig::BinaryIntOp::LessThan
+                | acvm::acir::brillig::BinaryIntOp::LessThanEquals => {
+                    let features_per_comparison = 1 /*true */+1/*false */+1/*when ilog is zero*/+ match bit_size{
+                    acvm::acir::brillig::IntegerBitSize::U1 => 1,
+                    acvm::acir::brillig::IntegerBitSize::U8 => 8,
+                    acvm::acir::brillig::IntegerBitSize::U16 => 16,
+                    acvm::acir::brillig::IntegerBitSize::U32 => 32,
+                    acvm::acir::brillig::IntegerBitSize::U64 => 64,
+                    acvm::acir::brillig::IntegerBitSize::U128 => 128,
+                };
```
asterite, line 719: It seems there's `impl From<IntegerBitSize> for u32`

tooling/greybox_fuzzer/src/coverage.rs
```diff
@@ -0,0 +1,751 @@
...
+                    // Insert features for each potential difference log
+                    for i in 0..features_per_comparison {
+                        feature_to_index_map
+                            .insert((opcode_index, usize::MAX - i), total_features + i);
+                    }
+                    coverage_items.push(BrilligCoverageItemRange::Comparison(CmpCoverageRange {
+                        index: total_features,
+                        bits: features_per_comparison - 3,
```
asterite, line 727: Maybe keep the features_for_bits as a separate variable so we don't need to subtract here

tooling/greybox_fuzzer/src/mutation/configurations.rs
```diff
@@ -0,0 +1,874 @@
...
+}
+
+pub(crate) struct ByteValueMutationConfiguration {
+    random_byte_mutation_weight: usize,
+    #[allow(unused)]
+    dictionary_byte_mutation_weight: usize,
+    total_weight: usize,
+}
+
+/// Configuration for selecting a byte value mutation
```
asterite, line 20: This comment should go on top of the struct

tooling/greybox_fuzzer/src/mutation/configurations.rs
```diff
@@ -0,0 +1,874 @@
+use rand::Rng;
```
asterite, line 1: The logic for selecting a mutation based on weights inside a configuration is repeated here a lot. I was thinking how this could be avoided and came up with this: https://gist.github.com/asterite/cbea91f23ec600be288bcc39b0a3d14e

The only downside to it that I see is that you might forget to include an option in the config... however, I think that downside also exists in the current code: if you add a new option you must remember to add it to the configuration, and you must also remember to add to the total weight.

tooling/greybox_fuzzer/src/mutation/dictionary.rs
asterite: It's also fine to leave this as a follow-up PR 👍

tooling/greybox_fuzzer/src/corpus.rs
```diff
@@ -0,0 +1,431 @@
...
+            // Skip files that are already loaded
+            if self.file_manager.has_file(&path) {
+                continue;
+            }
+            let source = std::fs::read_to_string(path.as_path())
+                .unwrap_or_else(|_| panic!("could not read file {:?} into string", path));
+
+            let parsed_source = parse_json(&source, &self.abi).map_err(|parsing_error| {
+                format!("Error while parsing file {:?}: {:?}", path.as_os_str(), parsing_error)
+            })?;
```
asterite, line 127: If you run `nargo fuzz` against one function and then change that function's signature and run `nargo fuzz` again it's almost sure you'll get an error here:

```
[one] Executed fuzzing task on test...issue with corpus:
Error while parsing file "corpus/one/test/060a619ba2124046f9b324b52ba4526b7896ed40ed7b2bc52a962eac1a178637.json": MissingArgument("y")
[one] 1 test failed
```

I guess that's fine, but I wonder if we can guide the user towards fixing that error. For example I deleted that file and it failed again because there was another file. After deleting both files it succeeded.

Not something to improve here, but maybe capture a follow-up issue for this.

tooling/greybox_fuzzer/src/lib.rs
```diff
@@ -0,0 +1,1356 @@
...
+            witness,
+            brillig_coverage.clone().unwrap(),
+            bool_witness_list,
+        );
+        // Quickly detect if there is any new coverage so that later single-threaded check can quickly discard this testcase
+        accumulated_coverage.detect_new_coverage(&new_coverage)
+    }
+
+    /// Start the fuzzing campaign
+    pub fn fuzz(&mut self) -> FuzzTestResult {
```
asterite, line 428: It would be nice if it were possible to break this method down into using some helper functions, mainly because it's very long and also to turn individual processing chunks into separate functions.

asterite (commented):
Incredible work!

The code looks great. I must admit that I still don't fully understand 100% of it but mainly because it's a lot to absorb (just yesterday I learned about greybox fuzzing). That said, I think I understand the gist of it, also thanks to our pairing today. Also, it seems to work very well, for the example programs and for some things I tried.

I did left many suggestions, mainly about organizing code, removing duplication and making code clearer.

test_programs/fuzzing_failure/comparison/src/main.nr
Rumata888: Done

tooling/greybox_fuzzer/src/types.rs
Rumata888: Done

tooling/nargo_fmt/src/formatter/attribute.rs
Rumata888: Thanks

tooling/nargo_cli/build.rs
```diff
@@ -211,6 +212,40 @@ fn test_{test_name}(force_brillig: ForceBrillig, inliner_aggressiveness: Inliner
     .expect("Could not write templated test file.");
 }
 
+/// Generate all test cases for a given test name (expected to be unique for the test directory),
+/// based on the matrix configuration. These will be executed serially, but concurrently with
+/// other test directories. Running multiple tests on the same directory would risk overriding
+/// each others compilation artifacts, which is why this method injects a mutex shared by
```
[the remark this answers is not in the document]
Rumata888: Yeah, replaced

tooling/nargo/src/ops/fuzz.rs
Rumata888: No, this was an error

acvm-repo/brillig_vm/src/lib.rs
Rumata888: Added

tooling/greybox_fuzzer/src/coverage.rs
Rumata888: Thanks!

acvm-repo/acvm/src/pwg/mod.rs
Rumata888: Done

acvm-repo/brillig_vm/src/lib.rs
Rumata888: Done

acvm-repo/brillig_vm/src/lib.rs
Rumata888: Right

acvm-repo/brillig_vm/src/lib.rs
```diff
@@ -266,15 +498,19 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>> VM<'a, F, B> {
                 // We use 0 to mean false and any other value to mean true
                 let condition_value = self.memory.read(*condition);
                 if condition_value.expect_u1().expect("condition value is not a boolean") {
+                    self.trace_branching(*destination);
```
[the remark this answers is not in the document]
Rumata888: Yep

tooling/greybox_fuzzer/src/coverage.rs
Rumata888: Done

tooling/greybox_fuzzer/src/coverage.rs
Rumata888: Done

tooling/greybox_fuzzer/src/lib.rs
Rumata888: This is already broken up, but I'll try

tooling/greybox_fuzzer/src/lib.rs
```diff
@@ -0,0 +1,1356 @@
...
+fn display_starting_info(
+    minimize_corpus: bool,
+    seed: u64,
+    starting_corpus_size: usize,
+    num_threads: usize,
+    package_name: &str,
+    fuzzing_harness_name: &str,
+    corpus_path: &Path,
+    minimized_corpus_path: &Path,
+) {
```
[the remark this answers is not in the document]
Rumata888: Done

tooling/greybox_fuzzer/src/mutation/int.rs
Rumata888: My mistake

tooling/greybox_fuzzer/src/coverage.rs
Rumata888: No

tooling/greybox_fuzzer/src/coverage.rs
Rumata888: Updated

tooling/greybox_fuzzer/src/mutation/mod.rs
Rumata888: Moved it into configurations

tooling/greybox_fuzzer/src/mutation/configurations.rs
Rumata888: Changed all of that

tooling/greybox_fuzzer/src/mutation/mod.rs
Rumata888: Thanks

asterite:
I left a bunch of suggestions that might be good to address in this PR. Otherwise later it will be hard to find what needs to be changed (I think they are small things, like updating descriptions and tiny refactors).

tooling/greybox_fuzzer/src/corpus.rs
Rumata888: Maybe in a following PR

tooling/greybox_fuzzer/src/corpus.rs
Rumata888: Yes, I will fix in a follow-up PR

tooling/greybox_fuzzer/src/mutation/mod.rs
Rumata888: Thanks

tooling/greybox_fuzzer/src/mutation/string.rs
Rumata888: Added

tooling/nargo/src/ops/execute.rs
Rumata888: Oh, stuff changed I guess.

tooling/nargo/src/ops/fuzz.rs
Rumata888: Done

tooling/nargo/src/ops/fuzz.rs
Rumata888: I'm going to add should_fail_with compatibility to be able to run tests on greybox fuzzing next. So bools will be needed

tooling/nargo/src/ops/fuzz.rs
Rumata888: Done

tooling/greybox_fuzzer/src/corpus.rs
Rumata888: ok

tooling/nargo_cli/src/cli/fuzz_cmd.rs
Rumata888: Done

tooling/nargo_cli/src/cli/fuzz_cmd.rs
Rumata888: Done

tooling/nargo_cli/src/cli/fuzz_cmd.rs
Rumata888: Maybe in a follow-up. It already uses cli-error, so I'd have to put these into a helper to convert

tooling/greybox_fuzzer/src/coverage.rs
Rumata888: I changed it

tooling/greybox_fuzzer/src/mutation/array.rs
Rumata888: Added

Issue #6733, opened 2024-12-07
Add machine readable test reports


TomAFrench:
We currently run tests from external repositories only if the `run-external-checks` flag is set manually. This means that we can have regressions sneak in through PRs where this flag has not been set however we also cannot run it by default as any test failures due to breaking changes will cause our CI to fail until the external repository has been updated.

One solution to this would be for us to add a machine-readable report option to `nargo test`.  This would allow us to filter out any expected failures to be ignored when when considering whether to fail CI or not.

See https://nexte.st/docs/machine-readable/libtest-json/ for inspiration.

TomAFrench:
Note that we'd want to have some mechanism to flag when ignored tests should be reactivated again.

asterite:
I'm thinking how to do this. For each repo, here's what can happen:
1. The project compiles fine and all tests pass
2. The project compiles fine but some tests fail
3. The project doesn't compile at all

So, for each repo, we want to have a file that tells it's status, one of:
1. We know the project compiles fine and all tests pass (this would be signaled by a lack of this file, or with an empty file)
2. We know the project compiles but some tests fail. Here I guess we'll list the full names of the tests that fail, and the script should tell us if some of these started to pass again.
3. We know the project doesn't compile (but let us know if it starts compiling again)

So maybe it can be a yaml file.

```yaml
# All good
compiles: true
known_failures: []
```

```yaml
# Some tests fail
compiles: true
known_failures:
  - test_one
  - test_two
````

```yaml
# Does not compile
compiles: false

# this list could be non-empty and it's just ignored, but it's useful to keep around
# when the project starts compiling again
known_failures: [] 
```

But then this logic seems a bit too complex to have in a `.sh` file. What's the usual way we do these things? Could it be a Rust script? Or maybe it can be directly in `nargo test` as an extra option?

TomAFrench:
I don't think that it's necessary to add this into nargo (and would prefer we didn't). I think some use of jq would avoid this.

TomAFrench:
https://github.com/noir-lang/noir/blob/bc03152366f242a6976f6e006e12520989e5e112/.github/workflows/test-js-packages.yml#L537

Something like this to get a JSON object which contains tests along with their success statuses which we then compare against the yaml/JSON file.

PR #6776, opened 2024-12-11
feat: `nargo test -q` (or `nargo test --format terse`)

asterite:
# Description

Made on top of #6728

## Problem

Towards #6733

## Summary

Eventually we'll have `nargo test --format json`, which is what #6733. As a first step I wanted to introduce the `format` flag and, to try that abstraction in code I introduced a terse formatter similar to Rust.

On failure we don't immediately print the error. Instead, it's printed at the end (like in Rust). I can imagine this could be used in CI to reduce the output verboseness, because I guess all that matters in CI is seeing which tests failed. NOTE: I already added this `-q` flag to our external checks, you can see how it looks, I think it's a bit better as it's less noisy.

nargo-test-q

Like in Rust, there are two ways to trigger this terse behavior: `-q` (or `--quiet`) or `--format terse`. We could just keep `--format terse` but I think `-q` is convenient (and `format` is more general as it will soon support `json`).

Here's the output when all tests pass:

image

## Additional Context

1. The proptest output that shows up in the middle of the run will be gone with #6777
2. I just noticed that we have duplicate and now outdated code to run stdlib tests [here](https://github.com/noir-lang/noir/blob/master/tooling/nargo_cli/tests/stdlib-tests.rs). We've been able to run `nargo test` inside `noir_stdlib` for a while now so we can probably move those to a CI test matrix (in this PR I tried to reuse the code from `nargo_cli` but I couldn't because `nargo_cli` is a binary and we can't use that binary as a dependency in test integrations).

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/.github/workflows/test-js-packages.yml
+++ b/.github/workflows/test-js-packages.yml
@@ -559,7 +559,7 @@ jobs:
           - project: { repo: AztecProtocol/aztec-packages, path: ./noir-projects/noir-protocol-circuits/crates/rollup-lib }
           - project: { repo: AztecProtocol/aztec-packages, path: ./noir-projects/noir-protocol-circuits/crates/types }
 
-    name: Check external repo - ${{ matrix.project.repo }}
+    name: Check external repo - ${{ matrix.project.repo }}/${{ matrix.project.path }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
--- a/.github/workflows/test-js-packages.yml
+++ b/.github/workflows/test-js-packages.yml
@@ -591,7 +591,7 @@ jobs:
       
       - name: Run nargo test
         working-directory: ./test-repo/${{ matrix.project.path }}
-        run: nargo test --silence-warnings
+        run: nargo test -q --silence-warnings
         env:
           NARGO_IGNORE_TEST_FAILURES_FROM_FOREIGN_CALLS: true
 
--- a/.github/workflows/test-js-packages.yml
+++ b/.github/workflows/test-js-packages.yml
@@ -534,7 +534,7 @@ jobs:
       - name: Build list of libraries
         id: get_critical_libraries
         run: |         
-          LIBRARIES=$(grep -Po "^https://github.com/\K.+"  ./CRITICAL_NOIR_LIBRARIES | jq -R -s -c 'split("\n") | map(select(. != "")) | map({ repo: ., path: "./"})')
+          LIBRARIES=$(grep -Po "^https://github.com/\K.+"  ./CRITICAL_NOIR_LIBRARIES | jq -R -s -c 'split("\n") | map(select(. != "")) | map({ repo: ., path: ""})')
           echo "libraries=$LIBRARIES"
           echo "libraries=$LIBRARIES" >> $GITHUB_OUTPUT
         env:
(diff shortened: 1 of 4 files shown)


tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -53,6 +55,40 @@ pub(crate) struct TestCommand {
...
+
+    /// Display one character per test instead of one line
+    #[clap(short = 'q', long = "quiet", action)]
+    quiet: bool,
+}
+
+#[derive(Debug, Copy, Clone, clap::ValueEnum)]
+enum Format {
+    /// Print verbose output
+    Pretty,
```
aakoshh, line 71: Would `Verbose` itself be a more fitting name, as the opposite of "terse"? Not sure it's trying to be beautiful. 
OTOH might have `JsonPretty` at some point I suppose, whereas `JsonVerbose` is not a thing 🤷

tooling/nargo_cli/src/cli/test_cmd/formatters.rs
```diff
@@ -0,0 +1,319 @@
...
+                writer.reset()?;
+            }
+            TestStatus::Skipped => {
+                writer.set_color(ColorSpec::new().set_fg(Some(Color::Yellow)))?;
+                write!(writer, "s")?;
+                writer.reset()?;
+            }
+        }
+
+        if current_test_count % 88 == 0 && current_test_count < total_test_count {
```
aakoshh, line 211: I think deserves a comment, perhaps a `const` for the magic 88?

aakoshh (approved):
Awesome, nice to see it broken up!

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -53,6 +55,40 @@ pub(crate) struct TestCommand {
...
     #[clap(long, default_value_t = rayon::current_num_threads())]
     test_threads: usize,
+
+    /// Configure formatting of output
+    #[clap(long)]
+    format: Option<Format>,
+
+    /// Display one character per test instead of one line
+    #[clap(short = 'q', long = "quiet", action)]
+    quiet: bool,
```
TomAFrench, line 65: These two flags can contradict each other and should be made mutually exclusive.

tooling/nargo_cli/src/cli/test_cmd.rs
asterite: I copied this from Rust:

```
        --format pretty|terse|json|junit
                        Configure formatting of output:
                        pretty = Print verbose output;
                        terse = Display one character per test;
                        json = Output a json document;
                        junit = Output a JUnit document
```

It's interesting because even the description for "pretty" says "verbose" 🤣 

I'll change it to verbose. It'll be different from Rust, which is something we try to avoid, but it's unlikely someone will pass `--format verbose` because that's the default.

aakoshh: No no leave it, let’s not surprise anyone who is familiar with it!

tooling/nargo_cli/src/cli/test_cmd.rs
asterite: Good point. I just tried it in Rust and it seems `format` takes precedence over `-q`. Should we do the same? We could also drop this flag and just keep `format`.

aakoshh: Ah yes, like [these](https://github.com/noir-lang/noir/blob/303a9c1c88256bc6714490e42a5641ac903eeb89/tooling/nargo_cli/src/cli/mod.rs#L64)

What does the `action` do?

aakoshh: FWIW I like `-q`

tooling/nargo_cli/src/cli/test_cmd.rs
asterite: Ah, okay. Right, I copied Rust for those familiar with Rust. I'll rever it then :-)

tooling/nargo_cli/src/cli/test_cmd.rs
asterite: Oh, it's a left-over. I first wanted it to be `Option<bool>` but then you have to pass `--quiet true` or `--quiet false` and I saw `action` could solve that, but I ended up not using it.

aakoshh:
It would be nice to change [this](https://github.com/noir-lang/noir/blob/0925a332dbaa561aad195c143079588158498dad/.github/workflows/test-js-packages.yml#L562) to contain the `path` like [this one](https://github.com/noir-lang/noir/blob/0925a332dbaa561aad195c143079588158498dad/.github/workflows/reports.yml#L284) does. 

Curiously even though it fails [on another PR](https://github.com/noir-lang/noir/actions/runs/12296073463/job/34320079472?pr=6789) that one has a green status. Maybe we don't want to be blocked by other repos checking in broken code?

asterite:
> It would be nice to change [this](https://github.com/noir-lang/noir/blob/0925a332dbaa561aad195c143079588158498dad/.github/workflows/test-js-packages.yml#L562) to contain the path like [this one](https://github.com/noir-lang/noir/blob/0925a332dbaa561aad195c143079588158498dad/.github/workflows/reports.yml#L284) does.

Done!

.github/workflows/test-js-packages.yml
```diff
@@ -559,7 +559,7 @@ jobs:
           - project: { repo: AztecProtocol/aztec-packages, path: ./noir-projects/noir-protocol-circuits/crates/rollup-lib }
           - project: { repo: AztecProtocol/aztec-packages, path: ./noir-projects/noir-protocol-circuits/crates/types }
 
-    name: Check external repo - ${{ matrix.project.repo }}
+    name: Check external repo - ${{ matrix.project.repo }}/${{ matrix.project.path }}
```
vezenovm, line 562: nit: If we add this, let's remove `./` from the path above as the output of this line will then be `AztecProtocol/aztec-packages/./noir-projects/noir-protocol-circuits/crates/types` which is a bit uglier than if it were not there

aakoshh: If you remove it, it also appears [here](https://github.com/noir-lang/noir/blob/78f5e87c3fee3a51a79b1c7f6e41572e8e35029f/.github/workflows/test-js-packages.yml#L537)

asterite: Will it work if I change "./" to "" there?

Maybe I can leave these changes to a separate PR? (and undo the one I did)

aakoshh: It should work, it's only used as a [suffix](https://github.com/noir-lang/noir/blob/78f5e87c3fee3a51a79b1c7f6e41572e8e35029f/.github/workflows/test-js-packages.yml#L593) to `/test-repo/`

But sure, it's unrelated, I thought the PR may have tightened some reporting, but it fails everywhere.

asterite: Done!

PR #6777, opened 2024-12-11
fix: disable failure persistance in nargo test fuzzing

asterite:
# Description

## Problem

Running `nargo test` with fuzz tests outputs

> proptest: FileFailurePersistence::SourceParallel set, but no source file known

which is just noise.

## Summary

To reproduce, have a project with this file:

```noir
# src/main.nr
#[test]
fn one(_input: u8) {}
```

Run `nargo test`. You get:

```
[one] Running 1 test function
proptest: FileFailurePersistence::SourceParallel set, but no source file known
[one] Testing one... ok
[one] 1 test passed
```

With this PR:

```
[one] Running 1 test function
[one] Testing one... ok
[one] 1 test passed
```

## Additional Context

It seems the default failure persistence is a directory called "proptest-regressions". But even when that directory exists we get the error. I assume here that `nargo test` never produces proptest regressions and that this PR doesn't break things.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/nargo/src/ops/test.rs
+++ b/tooling/nargo/src/ops/test.rs
@@ -13,6 +13,7 @@ use noirc_driver::{compile_no_check, CompileError, CompileOptions};
 use noirc_errors::{debug_info::DebugInfo, FileDiagnostic};
 use noirc_frontend::hir::{def_map::TestFunction, Context};
 use noirc_printable_type::ForeignCallError;
+use proptest::test_runner::Config;
 use rand::Rng;
 use serde::{Deserialize, Serialize};
 
@@ -114,7 +115,9 @@ pub fn run_test<B: BlackBoxFunctionSolver<FieldElement>>(
                     use acvm::acir::circuit::Program;
                     use noir_fuzzer::FuzzedExecutor;
                     use proptest::test_runner::TestRunner;
-                    let runner = TestRunner::default();
+
+                    let runner =
+                        TestRunner::new(Config { failure_persistence: None, ..Config::default() });
 
                     let executor =
                         |program: &Program<FieldElement>,


tooling/nargo/src/ops/test.rs
```diff
@@ -113,8 +113,11 @@ pub fn run_test<B: BlackBoxFunctionSolver<FieldElement>>(
                 {
                     use acvm::acir::circuit::Program;
                     use noir_fuzzer::FuzzedExecutor;
+                    use proptest::test_runner::Config;
                     use proptest::test_runner::TestRunner;
-                    let runner = TestRunner::default();
+
+                    let runner =
+                        TestRunner::new(Config { failure_persistence: None, ..Config::default() });
```
michaeljklein, line 120: Did you try using either of the following `failure_persistence` options?
- `FileFailurePersistence::WithSource`
- `FileFailurePersistence::Direct`
?

It appears that the `SourceParallel` option looks for `.rs` source files, which might explain why it wasn't finding the `proptest-regressions` folder.

michaeljklein (changes requested):

tooling/nargo/src/ops/test.rs
asterite: I didn't try it because I assume that whoever wrote this didn't intend to persist these.

TomAFrench: Yeah, I think for the purposes of this pull request just silencing these errors is correct. We may be deprecating this fuzzer due to keshas work anyway.

TomAFrench (approved):

michaeljklein (approved):
LGTM

PR #6781, opened 2024-12-12
Fix optimizer to keep track of changing opcode locations

aakoshh:
# Description

## Problem\*

Resolves the issue while trying to merge https://github.com/AztecProtocol/aztec-packages/pull/10483
Mirroring https://github.com/AztecProtocol/aztec-packages/pull/10483/commits/13539806da191988e44a9335c8aa82098dfa3566

## Summary\*

Unlike `transform_internal`, `optimize_internal` did not take an `acir_opcode_positions` which would have allowed it to be chained. This caused the second iteration of the loop to reset `acir_opcode_positions` to the default values, which indicated no change when the `AcirTransformationMap` was derived from it. The consequence was that the `assert_messages` and `debug` information did not align with the new opcode locations, which changed in the first iteration. 

## Additional Context

The way to debug this on `aztec-packages` went like this. 
First build everything with `./bootstrap.sh`, then in one terminal start an oracle:

```shell
export LOG_LEVEL=debug
cd aztec-packages/yarn-project/txe
yarn start
```

In another terminal run a test (one of the failing ones):
```shell
export NARGO=$(pwd)/aztec-packages/noir/noir-repo/target/release/nargo 
cd aztec-packages/noir-projects/noir-contracts
$NARGO test --silence-warnings --oracle-resolver http://localhost:8080 --package nft_contract transfer_to_public_failure_not_an_owner 
```

Then change something in the Noir codebase and recompile:
```shell
cd aztec-packages/noir 
./bootstrap.sh
```

With `MAX_OPTIMIZER_PASSES=1` the test passed, otherwise failed. The `LOG_LEVEL=debug` showed the index of the failing opcode. Curiously it was the same in both tests (the oracle is expected to fail, the test verifies it fails with the right message).

One of the things I tried was to print the ACIR opcodes with, with a change in the code so it prints the optimised version, after transformations:
```
$NARGO compile --package nft_contract --print-acir 
```
Comparing the ACIR indicated that 1 or 3 passes doesn't make any difference. Since it was also the same opcode, I started to suspect that it is actually the assertion message that gets lost.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/acvm-repo/acvm/src/compiler/mod.rs
+++ b/acvm-repo/acvm/src/compiler/mod.rs
@@ -79,42 +79,47 @@ fn transform_assert_messages<F: Clone>(
 ///
 /// Runs multiple passes until the output stabilizes.
 pub fn compile<F: AcirField>(
-    acir: Circuit<F>,
+    mut acir: Circuit<F>,
     expression_width: ExpressionWidth,
 ) -> (Circuit<F>, AcirTransformationMap) {
-    if MAX_OPTIMIZER_PASSES == 0 {
-        let acir_opcode_positions = (0..acir.opcodes.len()).collect::<Vec<_>>();
-        let transformation_map = AcirTransformationMap::new(&acir_opcode_positions);
-        return (acir, transformation_map);
+    let mut acir_opcode_positions = (0..acir.opcodes.len()).collect::<Vec<_>>();
+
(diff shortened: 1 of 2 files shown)


asterite (approved):
Looks good!

PR #6783, opened 2024-12-12
feat(ssa): Bring back tracking of RC instructions during DIE

vezenovm:
# Description

## Problem\*

Brings back the optimization with a fix that was [removed for correctness ](https://github.com/noir-lang/noir/pull/6700)

## Summary\*

This came to mind while working on https://github.com/noir-lang/noir/pull/6782 as actually being a very beneficial optimization. After #6782 we will have multiple locations (loop invariant code motion and constant folding) where upon moving a make_array instructions, we will conservatively issue inc_rc instructions in case that array is mutable. However, if the array is never mutated or mutably borrowed that will lead to us unnecessary incrementing the reference counter. 

Compared to the original RcTracker: 
- We now check ref counts based upon the type of an array per block
- We check whether on an array type has been mutated or passed into a call. Calls can accept mutable arguments as inputs and contain calls themselves. If the mutation instruction is simplified after inlining we can potentially remove a necessary inc rc

This lets us get rid of most increment rcs on arrays that are never mutated.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
Maybe it'd be easier if we promote this to its own pass rather than basing it off of DIE logic

vezenovm:
> Maybe it'd be easier if we promote this to its own pass rather than basing it off of DIE logic

Perhaps, I was thinking the same thing as it makes DIE more complex. On the flip side, I wanted to avoid another pass for this logic which can be contained to DIE and in a way does remove useless instructions (even if they are not "dead").

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -517,6 +526,112 @@ fn apply_side_effects(
...
+    // IncrementRc and DecrementRc instructions are normally side effectual instructions, but we remove
+    // them if their value is not used anywhere in the function. However, even when their value is used, their existence
+    // is pointless logic if there is no array set between the increment and the decrement of the reference counter.
+    // We track per block whether an IncrementRc instruction has a paired DecrementRc instruction
+    // with the same value but no array set in between.
+    // If we see an inc/dec RC pair within a block we can safely remove both instructions.
+    rcs_with_possible_pairs: HashMap<Type, Vec<RcInstruction>>,
+    rc_pairs_to_remove: HashSet<InstructionId>,
+    // We also separately track all IncrementRc instructions and all arrays which have been mutably borrowed.
+    // If an array has not been mutably borrowed we can then safely remove all IncrementRc instructions on that array.
```
vezenovm, line 541:
```suggestion
    // We also separately track all IncrementRc instructions and all array types which have been mutably borrowed.
    // If an array is the same type as one of those non-mutated array types, we can safely remove all IncrementRc instructions on that array.
```

jfecher (changes requested):
I'd like to see if this breaks an external repos if we change rcs to start at 1 again instead of 2. No errors in cargo test locally at least, but I'm checking in https://github.com/noir-lang/noir/pull/6797 in the meantime.

vezenovm:
> **regression_4709**	-51,320,073 ✅	**-97.72%**

This PR is where we get the largest benefit for patterns like this test where we are simply reading from a global array in a loop.

jfecher:
My main hesitancy with this PR is that it was removed for correctness before and is re-added now but what changed? As far as I can tell it's identical. Won't we run into sync issues again? Or run into them once https://github.com/noir-lang/noir/pull/6797 is added as well?

vezenovm:
> My main hesitancy with this PR is that it was removed for correctness before and is re-added now but what changed? As far as I can tell it's identical. Won't we run into sync issues again? Or run into them once #6797 is added as well?

This PR also adds a check for whether an array type was passed to a call. I have also tested it here (https://github.com/AztecProtocol/aztec-packages/pull/10399) which passes all noir-projects tests in atec-packages (the yarn-project-test failures look to be from the p2p network tests). 

It does look this PR would run into issues with #6797. But don't we have these issues already if we were to add back #6797?

jfecher (approved):
Hmm alright, I'm okay with merging this then. Are you confident passing arrays into calls was the original issue or is this something we should continue investigating?

> It does look this PR would run into issues with https://github.com/noir-lang/noir/pull/6797. But don't we have these issues already if we were to add back https://github.com/noir-lang/noir/pull/6797?

I'm not concerned with #6797, the only failure there was from the failure that was originally specific to that rc change rather than the RcTracker changes here.

vezenovm:
> Hmm alright, I'm okay with merging this then. Are you confident passing arrays into calls was the original issue or is this something we should continue investigating?

Working with the original RC tracker (no `Call` checks), I had not finished reproducing the error from the noir protocol circuits failure yet. I then decided to try things against the `reference_counts` test with different inlining settings. We were hitting the case where `copy_mut` was failing as `array` was not being incremented as expected. This made sense to me as I was already thinking we need to treat `Call` the same way we treat `Store` for checking non mutated arrays. The failure from `reference_counts` is the one laid out in the `do_not_remove_inc_rc_if_used_as_call_arg` unit test (one of the only other new things in this PR). 

I can continue trying to reproduce the exact Noir snippet from aztec-packages, but I do think not checking arrays passed to calls was the culprit of the broken RC tracker.

PR #6791, opened 2024-12-12
chore: manage call stacks using a tree

### LINKED ISSUE #6603: Reduce memory usage of tracking instruction callstacks in `dfg.locations`
We currently track the callstacks for each instruction in the `locations` field of the `DataflowGraph`:

https://github.com/noir-lang/noir/blob/3c361c9f78a5d9de1b1bcb5a839d3bc481f89898/compiler/noirc_evaluator/src/ssa/ir/dfg.rs#L94

Heaptrack is reporting that inserting entries into this hashmap is using 17% of the compiler's peak memory. 

Image

We should definitely not require 157MB of memory in order to track this as the number of unique callstacks is going to be relatively low. It should be relatively easy to rework this so that we don't track N different instances of the same callstack.

guipublic:
# Description

## Problem\*

Resolves #6603 

## Summary\*
Call stacks are stored inside a big tree, which allows to share identical prefixes between call stacks

## Additional Context
The only drawback is that we need to re-create the trees among function contexts during inlining

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Looks like this is an improvement over https://github.com/noir-lang/noir/pull/6747 and gets us closer to the optimal https://github.com/noir-lang/noir/pull/6753#issuecomment-2531629009

TomAFrench:
We are eating a performance penalty however.

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -21,6 +22,75 @@ use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 use serde_with::DisplayFromStr;
 
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub(crate) struct CallStackId(u32);
+
+impl CallStackId {
+    pub(crate) fn root() -> Self {
+        Self::new(0)
```
jfecher, line 30: Can you move call stack related things to their own file?

Also, since we no longer need the linked list type - can you remove that file too?

jfecher (commented):
Some good improvements here - I wonder why regression_4709 increased so much in compilation time though. From the description it sounds like inlining could take longer but regression_4709 seems to be dominated by a large loop instead.

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -21,6 +22,75 @@ use serde::{Deserialize, Serialize};
...
+
+    pub(crate) fn index(&self) -> usize {
+        self.0 as usize
+    }
+
+    pub(crate) fn is_root(&self) -> bool {
+        self.0 == 0
+    }
+
+    // Retrieves a CallStack from a CallStackId
```
aakoshh, line 45: I'm not sure "Retrieve" is right word, maybe "reconstruct"? I usually associate it with some kind of database retrieval.

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -21,6 +22,75 @@ use serde::{Deserialize, Serialize};
...
+    fn get_call_stack(&self, locations: &[LocationNode]) -> CallStack {
+        let mut call_stack = List::new();
+        let mut current_location = *self;
+        while let Some(parent) = locations[current_location.index()].parent {
+            call_stack.push_back(locations[current_location.index()].value);
```
aakoshh, line 50: Just to confirm: if a location doesn't have a parent, its own value is not going to be part of the call stack either? Like I guess `root()` doesn't have a parent, and you don't expect `root.value` to be interesting?

guipublic:
> I wonder why regression_4709 increased so much in compilation time though. From the description it sounds like inlining could take longer but regression_4709 seems to be dominated by a large loop instead.

My guess is that it is due to getting a big list of flattened call stacks; not too deep but very wide, which means some elements have a lot of children and searching among them would be significant. I will try to use some sorted container for the children, like a btree.

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
guipublic: yes, root means an empty call stack.

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -21,6 +22,75 @@ use serde::{Deserialize, Serialize};
...
+        {
+            return *result;
+        }
+        locations.push(LocationNode { parent: Some(*self), children: vec![], value: location });
+        let new_location = CallStackId::new(locations.len() - 1);
+        locations[self.index()].children.push(new_location);
+        new_location
+    }
+
+    // Returns a new CallStackId which extends the current one with the provided call_stack.
```
aakoshh, line 71: Can these be elevated to be docstrings?

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -21,6 +22,75 @@ use serde::{Deserialize, Serialize};
...
+        }
+        call_stack
+    }
+
+    // Adds a location to the call stack
+    fn add_location(&self, location: Location, locations: &mut Vec<LocationNode>) -> CallStackId {
+        if let Some(result) = locations[self.index()]
+            .children
+            .iter()
+            .find(|child| locations[child.index()].value == location)
```
aakoshh, line 61: There are quite a few invariants assumed about the `locations` vector, e.g. that `self` is in it, and that the `children` of a `LocationNode` are all in it as well; it would be worth creating a _newtype_ to wrap the `Vec` and make sure these invariants hold, and you're not dealing with just any odd `vec!` someone made.

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -486,21 +558,91 @@ impl DataFlowGraph {
         destination.set_terminator(terminator);
     }
 
-    pub(crate) fn get_call_stack(&self, instruction: InstructionId) -> CallStack {
+    pub(crate) fn get_instruction_call_stack(&self, instruction: InstructionId) -> CallStack {
+        let call_stack = self.locations.get(&instruction).cloned().unwrap_or_default();
```
aakoshh, line 562:
```suggestion
        let call_stack = self. get_instruction_call_stack_id(instruction);
```

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -21,6 +22,75 @@ use serde::{Deserialize, Serialize};
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct LocationNode {
+    parent: Option<CallStackId>,
+    children: Vec<CallStackId>,
+    value: Location,
+}
+
+impl LocationNode {
+    fn new() -> Self {
+        Self { parent: None, children: vec![], value: Location::dummy() }
```
aakoshh, line 90: Where is this used?

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -486,21 +558,91 @@ impl DataFlowGraph {
...
+        call_stack.add_location(location, &mut self.location_array);
+    }
+
+    pub(crate) fn add_location_to_root(&mut self, location: Location) -> CallStackId {
+        if self.location_array.is_empty() {
+            self.location_array.push(LocationNode {
+                parent: None,
+                children: vec![],
+                value: location,
+            });
```
aakoshh, line 585: I think the fact that here we have to think about the array being empty is a sign that the array itself should be a _newtype_ that holds these methods, like `self.location_array.add_location_to_root` would establish the root ID, and then `self.location_array.add_location(parent_id, location)` would be an example of adding more stuff, rather than the IDs manipulating the collection.

aakoshh: I mean the sign that something is not right is that a person cannot do `CallStackId::root().add_location(location, &mut self.location_array)` because it might panic, so `CallStackId` alone is not responsible for the invariants, although it could special case `add_location` for when `self.is_root()` is true and the target is empty.

compiler/noirc_evaluator/src/ssa/ir/function.rs
```diff
@@ -85,6 +86,8 @@ impl Function {
     /// Note that any parameters or attributes of the function must be manually added later.
     pub(crate) fn new(name: String, id: FunctionId) -> Self {
         let mut dfg = DataFlowGraph::default();
+        // Adds root node for the location tree
+        dfg.add_location_to_root(Location::dummy());
```
aakoshh, line 90: Is this something that could be done in `DataFlowGraph::default()` itself?

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
guipublic: It is not, I removed it.

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -21,6 +22,75 @@ use serde::{Deserialize, Serialize};
...
+        let mut current_location = *self;
+        while let Some(parent) = locations[current_location.index()].parent {
+            call_stack.push_back(locations[current_location.index()].value);
+            current_location = parent;
+        }
+        call_stack
+    }
+
+    // Adds a location to the call stack
+    fn add_location(&self, location: Location, locations: &mut Vec<LocationNode>) -> CallStackId {
```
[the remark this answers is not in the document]
guipublic: Done

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
guipublic: Done, I used 'construct'

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -21,6 +22,69 @@ use serde::{Deserialize, Serialize};
...
+        }
+        call_stack
+    }
+
+    /// Adds a location to the call stack
+    fn add_child(&self, location: Location, locations: &mut Vec<LocationNode>) -> CallStackId {
+        if let Some(result) = locations[self.index()]
+            .children
+            .iter()
+            .find(|child| locations[child.index()].value == location)
```
jfecher, line 61: If we can remove this .iter .find that may fix the performance concerns

guipublic: I have modified it with a heuristic that works well for regression_4709.

I have a more complicated version that I'd rather do as a follow-up PR because I am not sure it would be useful since the heuristic does not degrade the memory improvement at all, so using an exact version for the search would only get less performance.

jfecher: Sounds good to me

jfecher (approved):
LGTM - we may still want to change call stacks back to Vecs from lists now that we no longer need the sharing there

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
guipublic: I tried to refactor this.

compiler/noirc_evaluator/src/ssa/ir/call_stack.rs
```diff
@@ -0,0 +1,114 @@
...
+        call_stack
+    }
+
+    /// Adds a location to the call stack
+    pub(crate) fn add_child(&mut self, call_stack: CallStackId, location: Location) -> CallStackId {
+        if let Some(result) = self.locations[call_stack.index()]
+            .children
+            .iter()
+            .rev()
+            .take(1000)
```
aakoshh, line 69: I assume we're searching in reverse so that we find recent additions faster; is there any significance as to why give up after 1000 children, after which a duplicate entry is acceptable? Is that value something that was actually observed? 

I wonder if you thought some reverse index from e.g. the hash of the location to the child `CallStackId` could be used to make this faster and easier to reason about? Like `children: HashMap<u64, CallStackId>` where the key is `fxhash::hash64(location)`.

compiler/noirc_evaluator/src/ssa/ir/function.rs
guipublic: Done

PR #6796, opened 2024-12-12
feat: add `nargo test --format json`

asterite:
# Description

## Problem

Part of #6733

## Summary

Adds `--format json` to `nargo test`.

Here's the output for a single suite and test that passes:

```json
{"event":"started","name":"one","test_count":1,"type":"suite"}
{"event":"ok","exec_time":0.049250625,"name":"one","type":"test"}
{"event":"ok","failed":0,"passed":1,"skipped":0,"type":"suite"}
```

Here's one when there's a failure:

```json
{"event":"started","name":"one","test_count":1,"type":"suite"}
{"event":"failed","exec_time":0.435959917,"name":"keys::getters::test::test_get_public_keys_unknown","stdout":"error: Test failed with the wrong message. \nExpected: Invalid public keys hint for address \nGot: 0 output values were provided as a foreign call result for 13 destination slots\nAssertion failed: '0 output values were provided as a foreign call result for 13 destination slots'\nat aztec/src/test/helpers/cheatcodes.nr\nCall stack:\n1. aztec/src/keys/getters/test.nr:17:19\n2. aztec/src/keys/getters/test.nr:10:19\n3. aztec/src/test/helpers/cheatcodes.nr:51:5\n","type":"test"}
{"event":"failed","failed":1,"passed":0,"skipped":0,"type":"suite"}
```

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/nargo_cli/src/cli/test_cmd/formatters.rs
+++ b/tooling/nargo_cli/src/cli/test_cmd/formatters.rs
@@ -443,7 +443,7 @@ impl Formatter for JsonFormatter {
                 }
             }
             TestStatus::Skipped => {
-                json.insert("event".to_string(), json!("skipped"));
+                json.insert("event".to_string(), json!("ignored"));
             }
             TestStatus::CompileError(diagnostic) => {
                 json.insert("event".to_string(), json!("failed"));
@@ -491,16 +491,16 @@ impl Formatter for JsonFormatter {
     ) -> std::io::Result<()> {
         let mut passed = 0;
         let mut failed = 0;
-        let mut skipped = 0;
+        let mut ignored = 0;
         for test_result in test_results {
             match &test_result.status {
                 TestStatus::Pass => passed += 1,
                 TestStatus::Fail { .. } | TestStatus::CompileError(..) => failed += 1,
-                TestStatus::Skipped => skipped += 1,
+                TestStatus::Skipped => ignored += 1,
             }
         }
         let event = if failed == 0 { "ok" } else { "failed" };
-        let json = json!({"type": "suite", "event": event, "passed": passed, "failed": failed, "skipped": skipped});
+        let json = json!({"type": "suite", "event": event, "passed": passed, "failed": failed, "ignored": ignored});
         println!("{json}");
         Ok(())
     }
--- a/tooling/nargo_cli/src/cli/test_cmd/formatters.rs
+++ b/tooling/nargo_cli/src/cli/test_cmd/formatters.rs
@@ -13,16 +13,17 @@ use super::TestResult;
 /// The order of events is:
 /// 1. Compilation of all packages happen (in parallel). There's no formatter method for this.
 /// 2. If compilation is successful, one `package_start_async` for each package.
-/// 3. For each test, one `test_start_async` event.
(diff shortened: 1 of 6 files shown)


tooling/nargo_cli/src/cli/test_cmd/formatters.rs
```diff
@@ -317,8 +319,132 @@ impl Formatter for TerseFormatter {
...
+        show_output: bool,
+        _deny_warnings: bool,
+        silence_warnings: bool,
+    ) -> std::io::Result<()> {
+        let mut json = Map::new();
+        json.insert("type".to_string(), json!("test"));
+        json.insert("name".to_string(), json!(&test_result.name));
+        json.insert("exec_time".to_string(), json!(test_result.time_to_run.as_secs_f64()));
+
+        let mut stdout = String::new();
```
michaeljklein, line 346: Not blocking, but I'd expect it to be useful to distinguish `stdout` from `stderr`: where `stdout` is the `test_result.output` and `stderr` includes outputs from `TestStatus::Fail` / `TestStatus::CompileError`. Otherwise, how can I get e.g. _just_ the test output from the JSON when a test fails?

michaeljklein (approved):
LGTM

tooling/nargo_cli/src/cli/test_cmd/formatters.rs
asterite: Good point. Stderr goes directly without json formatting and will only happen if compilation fails. This is the same as in Rust. So there's no need to distinguish it here (I think!)

TomAFrench:
Just comparing this to the libtest output (would be nice to be compatible)

TomAFrench:
Differences I see are:
- The `name` key for suites doesn't exist in libtest.
- We include an `exec_time` key for test completion.
- We don't print a start event for individual tests.

Can you add this start event? The other two fields should be fine to keep (`exec_time` in particular is useful).

asterite:
One difference I remember with libtest is that they show the elapsed time (`exec_time`) of the entire suite but not for individual tests. Here I chose to show it for individual tests and not for the suite, because:
1. that's what we had before this PR
2. I think it's more useful, to know which tests are slow
3. because tests run in parallel, I'm not sure how useful it is for the entire suite

Another difference is that because we use "skipped" instead of "ignored" I use that word... but we could change it.

TomAFrench:
Yeah, let's change "skipped" to "ignored".

TomAFrench:
:shrug: It's cheap to do and improves compatibility so might as well.

> I wasn't sure how useful it is, or if there are tools that already use the JSON output from libtest (the other formatters don't "use" the start event)

If `nargo test` hangs we'll be able to tell which test caused it (or at least narrow down to a couple).

asterite:
Done! If there are tools that use the libtest output and rely on the specific order they use we could later adjust the code on our side.

asterite:
@TomAFrench Let me know if the output looks fine now.

TomAFrench (approved):

asterite:
I think I'll undo the last commit in this PR. I wanted to start working on the filtering so the first step was running external checks, but I'll do that in a separate PR.

PR #5657, opened 2024-08-01
Document gate savings around if, loop, and no_predicates

jzaki:

PR #6806, opened 2024-12-13
chore: add if/loop tip (separate from no-predicate #5657)

jzaki:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*


docs/docs/explainers/explainer-writing-noir.md
```diff
@@ -140,6 +140,18 @@ Use arrays and indices that are known at compile time where possible.
 Using `assert_constant(i);` before an index, `i`, is used in an array will give a compile error if `i` is NOT known at compile time.
 :::
 
+### Reduce what is inside loops and conditional logic
+
+Putting less logic  inside an `if` (`else`, etc) paths, or inside a loop, translates to less gates required to represent the program.
+
+A loop duplicates the gates for each iteration of the loop, or put another way, "unwraps" or "flattens" the loop. Any calculations/calls that are unchanged in the loop should be calculated once before, and the result used in the loop.
+
+An `if` will create gates representing each path even though execution will use only one. Furthermore, there are additional operations required for each path. Sometimes this can have a multiplying effect on the operations in the `if` and `else` etc.
```
vezenovm, line 149: It would be nice to have a small example. Like what does the following translate into? 
```
if cond {
  y = x;
}
```

docs/docs/explainers/explainer-writing-noir.md
```diff
@@ -140,6 +140,18 @@ Use arrays and indices that are known at compile time where possible.
 Using `assert_constant(i);` before an index, `i`, is used in an array will give a compile error if `i` is NOT known at compile time.
 :::
 
+### Reduce what is inside loops and conditional logic
+
+Putting less logic  inside an `if` (`else`, etc) paths, or inside a loop, translates to less gates required to represent the program.
+
+A loop duplicates the gates for each iteration of the loop, or put another way, "unwraps" or "flattens" the loop. Any calculations/calls that are unchanged in the loop should be calculated once before, and the result used in the loop.
```
vezenovm, line 147:
```suggestion
A loop duplicates the gates for each iteration of the loop, or put another way, "unrolls" the loop. Any calculations/calls that are unchanged in the loop should be calculated once before, and the result used in the loop.
```
I think "unroll" is clearer as it is the term used more generally. We say "flatten" when we refer to if statements.

vezenovm: FYI, the compiler also now accounts for calculations that are unchanged in the loop and hoists them to be calculated before the loop. It is possible the compiler may have missed things, and if there is conditional logic inside the loop that could be moved out of the loop this statement still holds true.

vezenovm (approved):
Just some nits, but approved

docs/docs/explainers/explainer-writing-noir.md
jzaki: https://github.com/noir-lang/noir/issues/6921

PR #6807, opened 2024-12-13
chore: Change Id to use a u32

jfecher:
# Description

## Summary\*

Changes `Id` to be represented by a `u32` instead of a `usize` to save memory.

## Additional Context

Unsure on performance implications since I also used `.try_into().unwrap()` to panic if the length is greater than a u32 over `as u32` which would be faster but silently wrap.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
image
Not bad for a 18 line diff.

PR #6810, opened 2024-12-13
feat: configurable external check failures

### LINKED ISSUE #6733: Add machine readable test reports

asterite:
# Description

## Problem

Resolves #6733

## Summary

WIP, for now I'll let this run the external checks to see which one we need to ignore.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/.github/workflows/test-js-packages.yml
+++ b/.github/workflows/test-js-packages.yml
@@ -591,7 +591,7 @@ jobs:
       
       - name: Run nargo test
         working-directory: ./test-repo/${{ matrix.project.path }}
-        run: nargo test -q --silence-warnings
+        run: nargo test --silence-warnings --format json
         env:
           NARGO_IGNORE_TEST_FAILURES_FROM_FOREIGN_CALLS: true
 
--- a/scripts/check-critical-libraries.sh
+++ b/scripts/check-critical-libraries.sh
@@ -31,7 +31,7 @@ for REPO in ${REPOS_TO_CHECK[@]}; do
     TAG=$(getLatestReleaseTagForRepo $REPO)
     git clone $REPO -c advice.detachedHead=false --depth 1 --branch $TAG $TMP_DIR
     
-    nargo test -q --program-dir $TMP_DIR
+    nargo test --program-dir $TMP_DIR --format json
 
     rm -rf $TMP_DIR
 done


asterite:
Strange, `rollup-lib` always gets stuck, even though it works fine on my machine.

vezenovm:
> Strange, `rollup-lib` always gets stuck, even though it works fine on my machine.

I noticed the same thing when making the compilation report (https://github.com/noir-lang/noir/pull/6750#discussion_r1878881848). Have not had the chance to investigate yet, I'll look a bit deeper at why this is occurring.

michaeljklein (approved):
LGTM

> Let me know if we should document all of this in a README file. 

It's not blocking, but I'd like to see a bit of docs in a `README`, e.g. a minimal one linking to this PR

.github/workflows/test-js-packages.yml
```diff
@@ -557,11 +555,17 @@ jobs:
           - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/parity-lib }
           - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/private-kernel-lib }
           - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/reset-kernel-lib }
-          - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/rollup-lib }
           - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/types }
+          # This one currently doesn't compile (it runs out of memory or it takes too long)
+          # - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/rollup-lib, nargo_args: "--test-threads 1" }
```
TomAFrench, line 560: Does this fail in a reasonable amount of time? If so can we reenable this so we can track how it fails (and when it becomes to pass again)?

asterite: Around 4 minutes. I'll try re-enabling it and removing/renaming the corresponding file. I think the big curve one takes more time than this.

asterite: I think the problem is that GitHub cancels the job and it fails. It doesn't even reach the comparison step. So I'll undo that commit (but feel free to try something else).

asterite:
I increased the job's timeout so now the rollup-lib tests might pass. On my machine running those tests took about an hour, so maybe in CI they'll take a bit longer.

There are 3 tests that take about 15 minutes each. I wonder if we could exclude just those tests in CI to speed up things a bit (we'd need a new `nargo test` option for that) or if we are fine with waiting that much.

asterite:
> we'd need a new nargo test option for that

`cargo test` has `--skip` to skip a test from running (it can be used multiple times). I think it does a substring but then there's another `--exact` option that changes it to do an exact match.

asterite:
It took 1 hour and 20 minutes.

asterite:
Oh, the flag already exists: `--skip-brillig-constraints-check`.

If we use it in the rollup-lib tests, at least on my machine it takes 4 minutes instead of 1+ hour.

@TomAFrench Is it safe to pass this flag in these tests? Should we do it for all tests or just for rollup-lib?

TomAFrench:
I think we can disable it for all external checks tbh. Fixing any issues associated with it falls to the external dev.

scripts/check-critical-libraries.sh
```diff
@@ -31,7 +31,7 @@ for REPO in ${REPOS_TO_CHECK[@]}; do
     TAG=$(getLatestReleaseTagForRepo $REPO)
     git clone $REPO -c advice.detachedHead=false --depth 1 --branch $TAG $TMP_DIR
     
-    nargo test -q --program-dir $TMP_DIR
+    nargo test --program-dir $TMP_DIR --format json
```
TomAFrench, line 34:
```suggestion
    nargo test -q --program-dir $TMP_DIR
```

Reverting this change as we don't need the full output in this script.

TomAFrench (approved):

PR #6814, opened 2024-12-13
chore: Do not print entire functions when running debug trace

vezenovm:
# Description

## Problem\*

I was running `NOIR_LOG=trace nargo` on some programs and my entire output was entirely filled with the SSA of my program. I was unsure where we were printing this SSA, but then I found that a couple trace macros are called in the under constrained values check where we state which function we are processing. We should print the ID, not the entire function.

## Summary\*

I switched to using func IDs instead of the entire function. This makes the trace actually readable. I could turn off the underconstrained check, but I think we should switch this to use an ID either way. Someone attempting to use `trace` should not have to know that they need this flag in order to get some reasonable debug output.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/acvm-repo/acvm/src/compiler/optimizers/mod.rs
+++ b/acvm-repo/acvm/src/compiler/optimizers/mod.rs
@@ -37,7 +37,7 @@ pub fn optimize<F: AcirField>(acir: Circuit<F>) -> (Circuit<F>, AcirTransformati
 /// Applies [`ProofSystemCompiler`][crate::ProofSystemCompiler] independent optimizations to a [`Circuit`].
 ///
 /// Accepts an injected `acir_opcode_positions` to allow optimizations to be applied in a loop.
-#[tracing::instrument(level = "trace", name = "optimize_acir" skip(acir))]
+#[tracing::instrument(level = "trace", name = "optimize_acir" skip(acir, acir_opcode_positions))]
 pub(super) fn optimize_internal<F: AcirField>(
     acir: Circuit<F>,
     acir_opcode_positions: Vec<usize>,
--- a/compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
+++ b/compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
@@ -387,7 +387,11 @@ impl DependencyContext {
             })
             .collect();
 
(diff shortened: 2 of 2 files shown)


compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -108,7 +108,7 @@ struct DependencyContext {
 }
 
 /// Structure keeping track of value ids descending from Brillig calls'
-/// arguments and results, also storing information on results
+/// arguments and results, also storing information osn results
```
vezenovm, line 111:
```suggestion
/// arguments and results, also storing information on results
```

TomAFrench (approved):

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -407,8 +402,6 @@ impl DependencyContext {
     /// Check if any of the recorded Brillig calls have been properly constrained
     /// by given values after recording partial constraints, if so stop tracking them
     fn clear_constrained(&mut self, constrained_values: &[ValueId], function: &Function) {
-        trace!("attempting to clear Brillig calls constrained by values: {:?}", constrained_values);
```
vezenovm, line 410: I deleted this as well as it is also taking up a lot of the trace. We can bring it back as needed for debugging.

acvm-repo/acvm/src/compiler/optimizers/mod.rs
```diff
@@ -37,7 +37,7 @@ pub fn optimize<F: AcirField>(acir: Circuit<F>) -> (Circuit<F>, AcirTransformati
 /// Applies [`ProofSystemCompiler`][crate::ProofSystemCompiler] independent optimizations to a [`Circuit`].
 ///
 /// Accepts an injected `acir_opcode_positions` to allow optimizations to be applied in a loop.
-#[tracing::instrument(level = "trace", name = "optimize_acir" skip(acir))]
+#[tracing::instrument(level = "trace", name = "optimize_acir" skip(acir, acir_opcode_positions))]
```
vezenovm, line 40: Another place we need to skip for tracing that was messing with tracing large ACIR programs

PR #6673, opened 2024-12-02
feat: revert changes to `ValueMerger` and `Instruction::IfElse`

TomAFrench:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:
- [ ] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [ ] I have tested the changes locally.
- [ ] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #6815, opened 2024-12-14
feat: use more efficient field merger code

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR replaces the value merger code reverted in #6673 but solely for fields.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/flatten_cfg/value_merger.rs
```diff
@@ -116,20 +116,41 @@ impl<'a> ValueMerger<'a> {
...
+            let conditional_diff_value = dfg
+                .insert_instruction_and_results(conditional_diff, block, None, call_stack.clone())
+                .first();
+
+            let merged_field =
+                Instruction::binary(BinaryOp::Add, else_value, conditional_diff_value);
+            dfg.insert_instruction_and_results(merged_field, block, None, call_stack.clone())
+                .first()
+        } else {
+            let cast = Instruction::Cast(else_condition, else_type);
```
aakoshh, line 139: Related to this I found that when we were casting to `Field` we saved a lot of range constraints implied on the binary operations performed on `Unsigned` types. First I thought that must be unsafe as it would mask a potential overflow, but with the `c*a + !c*b` formula below it seems safe (or actually it has an `else_condition` below, but assuming that they are mutually exclusive). 

I assume you're not doing the `c * (a-b) + b` trick because of a potential overflow, but isn't it true that mathematically this should be okay, if the cost of the extra `Cast`s to `Field` would be acceptable?

TomAFrench: Yeah, we're not concerned about overflows here as the merged value is constrained to `then_value` or `else_value` and any underflow in the difference term will disappear (This is why we needed to cast to Fields as the integer types imply checked arithmetic)

I did this just for fields in this PR as this doesn't need the inputs to be casted to fields and back (as it's already the correct type).

TomAFrench:
I think we should wait for #6817 before pushing this further due to negligible changes within the test suite.

PR #6827, opened 2024-12-16
chore(ci): Execution time report

### LINKED ISSUE #6773: Add execution time benchmarks for protocol circuits
After https://github.com/AztecProtocol/aztec-packages/pull/10608 we now have sample inputs to execute some of the protocol circuits with. Let's add this to our benchmarking suite in order to track execution time.

Currently this will show compilation + execution time which isn't ideal but we can look at the difference between the execution and compilation times to get the pure execution cost.

vezenovm:
# Description

## Problem\*

Resolves #6773 

## Summary\*

Adds a report similar to the already existing memory and compilation reports. 

The main addition in this pass is that we do not force recompile the program we profiling. The `execution_report.sh` script expects a compilation artifact to already be built. Thus, the execution report relies on the compilation reports which upload its resulting compilation artifacts. 

## Documentation\*

Check one:

# PR Checklist\*


.github/workflows/reports.yml
```diff
@@ -490,3 +522,194 @@ jobs:
...
+              mkdir ./test-repo/${{ matrix.project.path }}/target
+              mv ./artifacts/$ARTIFACT_NAME/$ARTIFACT_NAME.json ./test-repo/${{ matrix.project.path }}/target/$PACKAGE_NAME.json
+              break
+            fi
+          done
+
+      - name: Generate execution report
+        working-directory: ./test-repo/${{ matrix.project.path }}
+        run: |
+          mv /home/runner/work/noir/noir/scripts/test_programs/execution_report.sh ./execution_report.sh
```
TomAFrench, line 652: Is it worth just inlining this into the compilation report job so that it emits multiple report artifacts? That's what I'm doing in https://github.com/noir-lang/noir/pull/6832 plus it will make life easier once we tackle #6844

vezenovm: Yeah it may be worth just inlining this, especially as the execution reports depends on the compilation report. We do have certain packages we compile but do not execute, such as `noir-contracts` as it is a library. I could add an extra flag to the matrix as to whether we should execute?

TomAFrench: Hmm, yeah the contracts are a little bit of an issue. I think adding a flag is the best option for now.

vezenovm: I have inlined the execution report into the compilation report.

.github/workflows/reports.yml
```diff
@@ -316,28 +336,57 @@ jobs:
...
           PACKAGE_NAME=$(basename $PACKAGE_NAME)
           mv ./test-repo/${{ matrix.project.path }}/compilation_report.json ./compilation_report_$PACKAGE_NAME.json
           echo "compilation_report_name=$PACKAGE_NAME" >> $GITHUB_OUTPUT
 
+      - name: Move execution report
+        id: execution_report
+        shell: bash
+        run: |
+          IS_LIBRARY=${{ matrix.project.is_library }}
+          if [[ "$IS_LIBRARY" != "true" ]]; then
```
TomAFrench, line 363: You can add a `if: ${{ !matrix.project.is_library }}` to the job directly (along with any later steps). I think this is cleaner.

TomAFrench (approved):
Looks good, one nit on how we disable the execution report for libraries

Issue #6738, opened 2024-12-09
Support fuzz tests in wasm


TomAFrench:
We disalllowed running fuzz tests in wasm due to lack of `i128` support however we've since restricted the type system to only allow `u64`s as the maximum integer type. We then don't need to use `i128`s at all and can remove this conditional compilation by changing the `UintStrategy` to use `u64`s

PR #6835, opened 2024-12-17
feat(test): Enable the test fuzzer for Wasm

aakoshh:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6738

## Summary\*

* Upgrades `proptest` to 1.6 to be able to use `u128` and `i128` strategies with Wasm
* Creates the `UIntStrategy` with max 64 bit size, still returning `u128`
* Creates the `IntStrategy` with max 64 bit size, still returning `i128`
* Enables the `FuzzedExecutor` for `#[test]`s with inputs on Wasm
* Disables the [fork and timeout](https://altsysrq.github.io/proptest-book/proptest/forking.html) features, otherwise it doesn't seem to compile for Wasm (the compiler needs to be told if the target OS is `unix` or `windows`).

## Additional Context

The ticket says:
> we've since restricted the type system to only allow u64s as the maximum integer type. 

I couldn't find where exactly this happened, it would be nice to include a link. 

I started with using `u64` instead of `u128`, and `i64` instead of `i128`; the latter covers half the range as if we used `u64` and mapped to negative or positive based on whether the random `start` value we picked was negative or positive. I did it that way because `proptest` [enabled](https://github.com/proptest-rs/proptest/commit/fdc53faf031f374019ab15857043cf9ab9ad7a20) `i128` and `u128` for Wasm just last week, and I thought it would be easier to to go back to using `i128` when it's released. Then I noticed that `proptest` 1.6 was actually just released yesterday, and it [includes](https://docs.rs/proptest/1.6.0/src/proptest/num.rs.html#368) the above change, it just didn't have a tag in the repo 🎉 

On a different note [this shifting](https://github.com/noir-lang/noir/blob/f037c36f6bfcb8efb1950e444b50bc3eff28ffc4/tooling/fuzzer/src/strategies/mod.rs#L28-L33) makes it look like we don't even want negative values for signed numbers, so maybe it would be easier to just use the `UIntStrategy` without `fixtures` and covert the returned value to `i128`.

### Compilation

CI uses `wasm-pack`, but we can also compile it locally with: 
```shell
cargo build -p noir_wasm --target wasm32-unknown-unknown
```

## Documentation\*

Check one:

# PR Checklist\*


tooling/fuzzer/src/strategies/int.rs
```diff
@@ -40,35 +45,45 @@ impl IntStrategy {
...
-        Ok(proptest::num::i128::BinarySearch::new(start))
+        Ok(BinarySearch::new(start))
     }
 
+    /// Maximum allowed positive number.
     fn type_max(&self) -> i128 {
-        if self.bits < 128 {
-            (1i128 << (self.bits - 1)) - 1
+        let bits = self.type_max_bits();
+        if bits < 128 {
```
asterite, line 60: Wouldn't this always be true now?

aakoshh: Yes; I thought, what if someone changes it back, it would fail to produce the maximum shift (at least in the `uint` version). 

But you're right, it looks daft. I moved the restriction to 64 bits to where the strategy is created, and then here it can still generate the maximum range, it just will never be asked to do so.

TomAFrench (approved):
Nice!

Issue #6841, opened 2024-12-17
Separate runtimes before SSA


vezenovm:
### Problem

We have two types of functions, ACIR and Brillig (constrained vs unconstrained). Brillig functions are simply marked with the `unconstrained` function modifier. 

In Noir code, we allow ACIR functions to be called in the unconstrained environment. The goal was to reduce code duplication and the need to have the user write two different Noir functions for both constrained and unconstrained if they wanted to compute the same thing in both runtimes. 

Right now, we separate runtimes during an SSA pass called `runtime_separation`. This converts any remaining functions with the Brillig runtime. If an ACIR function was called from the Brillig runtime, this pass will clone the function and change its runtime to Brillig. The pass is restricted to only work after defunctionalization.

We should move to have runtime separation occur during monomorphization in the frontend before SSA generation. This would enable us then to code gen different SSA based upon the runtime of a function such as #6231. Logic from other issues such as https://github.com/noir-lang/noir/issues/6831 could also just be included as part of SSA generation as we will already know the runtime of the function we are generating. 

### Happy Case

We should be able to safely access function runtimes during SSA generation. 

### Workaround

Yes

### Workaround Description

The workaround is to keep doing what we are doing and work with separated runtimes only after the `runtime_separation` pass. I find this generally will lead to more complexity.

### Project Impact

Blocker

### Blocker Context

This work is a blocker for PR #6231. We could probably hack around not having the runtime during SSA gen, but it would be ugly and introduce a lot of complexity.

### Would you like to submit a PR for this Issue?

None

PR #4832, opened 2024-04-17
fix: defer overflow checks for unsigned integers to acir-gen

guipublic:
# Description

## Problem\*

Resolves #4456

## Summary\*

Overflow checks for unsigned integers are done in acir-gen, when converting arithmetic operations into field arithmetic. That way the semantic of the ssa unsigned operations is preserved.

## Additional Context

As indicated in the issue, I did not touch signed integers, but it could be cleaner to convert them into unsigned operations (and have the signed overflow checks done with additional ssa instructions, like we do currently)

## Documentation\*

Check one:
- [X] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [X] I have tested the changes locally.
- [X] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #6848, opened 2024-12-17
fix: don't deduplicate binary math of unsigned types

### LINKED ISSUE #6834: Constraints mistakenly failing until a pointless print or assert statement is added
A nice new example from Theo, which might be easier to reproduce than other examples: https://aztecprotocol.slack.com/archives/C0768TZ8KT4/p1734426057449759

Related examples:
https://github.com/AztecProtocol/aztec-packages/issues/10558
https://github.com/noir-lang/noir/issues/5771

asterite:
# Description

## Problem

Resolves #6834

## Documentation

Check one:

# PR Checklist


test_programs/execution_success/regression_6834/src/main.nr
```diff
@@ -0,0 +1,13 @@
+fn main(year: u32, min_age: u8) {
+    if min_age == 0 {
+        if year > 2028 {
+            assert((((year - 1) - 2028) / 4 + 1) as i32 >= 0);
+        }
+    } else {
+        if year > 2028 {
+            // Uncomment the next line and the assertion on line 10 will pass
+            // assert(year > 2028);
```
asterite, line 9: It's also interesting that if we make this line be `println("hello");` and uncomment it, the program also passes.

asterite:
Here's the diff if it's a `println` instead of an assert:

image

asterite:
The failure we get in both cases is this:

```
error: Failed constraint
   ┌─ /Users/asterite/Sandbox/one/src/main.nr:10:21
   │
10 │             assert((((year - 1) - 2028) / 4 + 1) as i32 >= 0);
   │                     -----------------------
   │
   = Call stack:
     1. /Users/asterite/Sandbox/one/src/main.nr:10:21
```

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -745,15 +745,17 @@ impl ResultCache {
         has_side_effects: bool,
     ) -> Option<CacheResult> {
         self.result.as_ref().and_then(|(origin_block, results)| {
+            if has_side_effects {
+                return None
+            }
```
asterite, line 750: Should this be right at the beginning of the function?

asterite: Or it could even be done in `get_cached`

vezenovm: Yes it can be moved

asterite:
I pushed a commit just to try something and see how opcodes change.

vezenovm:
> I pushed a commit just to try something and see how opcodes change.

I think marking overflowing ops as requiring acir gen predicates is the most accurate.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -556,10 +556,22 @@ impl Instruction {
     /// If true the instruction will depend on `enable_side_effects` context during acir-gen.
     pub(crate) fn requires_acir_gen_predicate(&self, dfg: &DataFlowGraph) -> bool {
         match self {
-            Instruction::Binary(binary)
-                if matches!(binary.operator, BinaryOp::Div | BinaryOp::Mod) =>
-            {
-                true
+            Instruction::Binary(binary) => {
+                // Some binary math can overflow or underflow
+                match binary.operator {
```
vezenovm, line 561: Let's case for just unsigned ops as we do in acir gen https://github.com/noir-lang/noir/blob/0d7642cb2071fbee148978a89a0922bfffe5be6a/compiler/noirc_evaluator/src/acir/mod.rs#L1984

vezenovm:
> **ram_blowup_regression**	+98,719 ❌	**+12.68%**
> **brillig_cow_regression**	+64,620 ❌	**+12.45%**

We may be able to optimize this regression a bit in a follow-up as loop invariant code motion currently checks whether a binary operation requires a predicate inside of the check for whether an instruction can be deduplicated https://github.com/noir-lang/noir/blob/0d7642cb2071fbee148978a89a0922bfffe5be6a/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs#L208.

Looking at the overflow check in brillig gen:
https://github.com/noir-lang/noir/blob/0d7642cb2071fbee148978a89a0922bfffe5be6a/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs#L1471

We can see that there are a few instructions generated during brillig gen (the Mul check is the most expensive) that we should be able to move to SSA. All instructions other than the final constrain we should then be able to hoist out of the loop.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
asterite: Strange, why is this check only done for unsigned types?

asterite:
> I think marking overflowing ops as requiring acir gen predicates is the most accurate

Given that the name is `requires_acir_gen_predicate`, and that there brillig opcodes executed has increased in this PR... does this new logic is the same in acir and brillig?

vezenovm:
> Given that the name is `requires_acir_gen_predicate`, and that there brillig opcodes executed has increased in this PR... does this new logic is the same in acir and brillig?

It is the same in the case of loop invariant code motion. For ACIR this just helps compilation time as we unroll all loops, but this can change the final Brillig bytecode and execution trace. So in both cases we do not want to be hoisting these instructions. This means we want this check:
[noir/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs](https://github.com/noir-lang/noir/blob/0d7642cb2071fbee148978a89a0922bfffe5be6a/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs#L208)
to return `false`.

`can_be_deduplicated` currently calls `requires_acir_gen_predicate` internally. A lot of the instructions which require a predicate have the same status in Brillig and ACIR (such as array get and checked binary ops), thus `can_be_deduplicated` was re-using `requires_acir_gen_predicate`. `can_be_deduplicated` in fact has more strict conditions than `requires_acir_gen_predicate`. 

As noted above, the method to determine whether we can deduplicate instructions is muddled as it is used for both ACIR and Brillig runtimes. The intertwining of this logic has happened before and is why these methods were originally written, but we should re-assess. We may simply want better documentation surrounding these methods. We could also make a separate `can_be_deduplicated` method for Brillig vs. ACIR runtimes as to have a clear distinction. For example, in the Brillig runtime case, we would instead deduplicate Binary instructions based off of the `deduplicate_with_predicate` flag. This is more accurate for Brillig as the reason it cannot be hoisted is not due to its reliance on the acir gen predicate, but rather due to that the brillig gen overflow check code gens constrains:
https://github.com/noir-lang/noir/blob/0d7642cb2071fbee148978a89a0922bfffe5be6a/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs#L1506

This can happen in follow-up work though, and updating just `requires_acir_gen_predicate` should be fine for this PR.

vezenovm (approved):

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
vezenovm: Looks like we perform some overflow checks in SSA:
https://github.com/noir-lang/noir/blob/a97402aa3b3a1a7d40f61242e02d3772175846f2/compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs#L344

We deferred those specific ACIR gen checks in this PR https://github.com/noir-lang/noir/pull/4832 and you can view the issue in that PR's description for more context. The TLDR for why we deferred the checks is that how we perform an overflow check can differ in ACIR vs. Brillig. Rather than deferring all the way to ACIR we should still generate these checks during SSA so the instructions can benefit from the SSA optimizations. This would be another place #6841 would benefit as we could simply code gen differing SSAs immediately.

Issue #6742, opened 2024-12-09
jsonrpc dependency is unmaintained and is failing `cargo deny`


TomAFrench:
We're currently using https://github.com/paritytech/jsonrpc which is causing `cargo deny` to fail due to an issue in one of its dependencies: https://github.com/noir-lang/noir/actions/runs/12236926106/job/34131483409#step:4:6638

jsonrpc is no longer maintained as the maintainers have moved on to https://github.com/paritytech/jsonrpsee so we should probably update to this library.

TomAFrench:
See https://github.com/noir-lang/noir-bignum/pull/62/files for usage.

PR #6849, opened 2024-12-17
feat!: Switch to using `jsonrpsee` for foreign calls; refactor `run_test`; foreign call layering

aakoshh:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6742

## Summary\*

Instead of `jsonrpc` we use `jsonrpsee` to make calls to the Oracle function resolver.

Also includes https://github.com/noir-lang/noir/pull/6857 and https://github.com/noir-lang/noir/pull/6858 which effectively make `nargo::ops::test` receive a generic `ForeignCallExecutor` (a breaking change in the API) so that we can use it with Wasm, which otherwise could not compile the `RpcForeignCallExecutor`. 

## Additional Context

Our codebase is mostly synchronous, but `jsonrpsee` is async. I created a `tokio::runtime::Runtime` in the executor where I can block on the calls, however this doesn't work well with the `#[tokio::test]`, runtimes don't nest. Because of this the tests require a bit more supporting machinery than before.

## Documentation\*

Check one:

# PR Checklist\*


aakoshh:
@TomAFrench it looks like it doesn't compile to Wasm because the `http-client` feature brings in `tokio:net` which brings `mio` as a dependency. It doesn't look like `noir_wasm` uses the `foreign_calls` module, so I guess I can try putting that behind a feature flag, so we don't have these network related dependencies when we're compiling to Wasm.

TomAFrench:
Yeah we don't need an RPC server in wasm as we expect that to use callbacks directly. We can put it behind a feature flag then.

aakoshh:
@TomAFrench the problem is that I not with the RPC server, which was only used as a `dev-dependency`, it's the RPC _client_ itself. I wanted to it behind a feature flag for foreign calls, but then realised that `ops::test` directly depends on being able to make HTTP calls: `TestForeignCallExecutor` uses `RPCForeignCallExecutor`. By contrast, `ProgramExecutor` in `ops::execute` only uses the `ForeignCallExecutor` trait, which would be easier to isolate from the network related stuff in  `foreign_calls` module. 

I created a `test` and `execute` feature flag, so that `noir_wasm` can depend only on compilation machinery, which is the only one it has `#[wasm_bindgen]` for right now, but by disabling `test` in Wasm, https://github.com/noir-lang/noir/pull/6835 is moot at the moment.

We can refactor the whole module so that a `ForeignCallExecutor` is passed to the `ops::test::run_test` as a dependency, similar to how it's done in `ProgramExecutor::new`, rather than as a URL, which I think is useless within Wasm, at least I don't see any injected methods that would handle network calls. It would be a breaking change in the test API, though.

> in wasm as we expect that to use callbacks directly

Perhaps this is related, can you elaborate on what you wanted to do?

tooling/nargo/Cargo.toml
```diff
@@ -21,20 +21,27 @@ noirc_errors.workspace = true
...
-jsonrpc-derive = "18.0"
-jsonrpc-core = "18.0"
+jsonrpsee = { workspace = true, features = ["server"] }
+
+[features]
+default = ["execute", "test"]
+
+# Execution currently uses HTTP based Oracle resolvers; does not compile to Wasm.
+execute = ["jsonrpsee/http-client", "jsonrpsee/macros", "tokio/rt", "serde"]
+test = ["execute", "noir_fuzzer", "proptest", "rand"]
```
aakoshh, line 47:
```suggestion
execute = ["jsonrpsee/http-client", "jsonrpsee/macros", "tokio/rt", "serde", "rand"]
test = ["execute", "noir_fuzzer", "proptest"]
```

compiler/wasm/Cargo.toml
```diff
@@ -33,6 +33,11 @@ gloo-utils.workspace = true
 tracing-subscriber.workspace = true
 tracing-web.workspace = true
 
+# Disable the execution feature so we don't get the HTTP dependency which doesn't compile to Wasm.
```
TomAFrench, line 36:
```suggestion
# Disable the rpc feature so we don't get the HTTP dependency which doesn't compile to Wasm.
```

tooling/nargo/Cargo.toml
```diff
@@ -21,20 +21,26 @@ noirc_errors.workspace = true
...
 
 [dev-dependencies]
-jsonrpc-http-server = "18.0"
-jsonrpc-core-client = "18.0"
-jsonrpc-derive = "18.0"
-jsonrpc-core = "18.0"
+jsonrpsee = { workspace = true, features = ["server"] }
+
+[features]
+default = ["rpc"]
```
TomAFrench, line 43: Can we make this a non-default flag and have `nargo_cli` turn it on? This should allow us to have `compiler/wasm` use the workspace dependency no? (maybe you tried this already.

aakoshh: I haven’t tried that, no. I just wanted to keep it backwards compatible for any potential user who already have this as a dependency, and have the override for noir_wasm, hidden in our build. We can try, I don’t have a strong opinion either way.

aakoshh: Yes, I can confirm that it would work that way. The following need to add the `rpc` feature:
* `/tooling/acvm_cli/Cargo.toml`
* `/tooling/debugger/Cargo.toml`
* `/tooling/profiler/Cargo.toml`
Not sure why `nargo_cli` didn't complain 🤔 

Up to you, like I say I didn't want to disrupt anyone who is used to having access to the `DefaultForeignCallExecutor`, which is not available without `rpc`.

aakoshh: Ah so the workspace unification is still in play: as soon as one crate required the `rpc` feature it became the default, which is why `nargo_cli` didn't need to specify it. It's weird that the Wasm crate even compiled 😵 

I was trying to figure out a way where we can keep the `DefaultForeignCallExecutor` but change its API based on whether the RPC config is accepted or not. For example the `acvm-cli` and `profiler` crates don't pass any URL, and it's not really enabled in the `debugger` crate either.

aakoshh: Changed it so `rpc` is an opt-in feature.

compiler/wasm/Cargo.toml
```diff
@@ -33,6 +33,11 @@ gloo-utils.workspace = true
 tracing-subscriber.workspace = true
 tracing-web.workspace = true
 
+# Disable the rpc feature so we don't get the HTTP dependency which doesn't compile to Wasm.
+# Cannot use `workspace = true` because it would unify all packages and compile with the default
+# features on, ignoring the setting here, but with `path` it works.
+nargo.workspace = true
```
aakoshh, line 39:
```suggestion
# Cannot use the `rpc` feature because the HTTP dependency wouldn't compile to Wasm.
# We could use `path` if `rpc` was a default feature, but we made it opt-in so we don't get any problems when publishing the workspace.
nargo.workspace = true
```

tooling/nargo/src/ops/test.rs
```diff
@@ -278,85 +266,41 @@ fn check_expected_failure_message(
...
-
-                // If all executors have no handler for the given foreign call then we cannot
-                // return a correct response to the ACVM. The best we can do is to return an empty response,
-                // this allows us to ignore any foreign calls which exist solely to pass information from inside
-                // the circuit to the environment (e.g. custom logging) as the execution will still be able to progress.
-                //
-                // We optimistically return an empty response for all oracle calls as the ACVM will error
-                // should a response have been required.
-                Ok(ForeignCallResult::default())
+                layers::Empty.execute(foreign_call)
```
TomAFrench, line 301:
```suggestion
 // If all executors have no handler for the given foreign call then we cannot
                // return a correct response to the ACVM. The best we can do is to return an empty response,
                // this allows us to ignore any foreign calls which exist solely to pass information from inside
                // the circuit to the environment (e.g. custom logging) as the execution will still be able to progress.
                //
                // We optimistically return an empty response for all oracle calls as the ACVM will error
                // should a response have been required.
                layers::Empty.execute(foreign_call)
```

tooling/nargo/src/foreign_calls/mocker.rs
```diff
@@ -174,3 +176,32 @@ impl<F: AcirField + Serialize + for<'a> Deserialize<'a>> ForeignCallExecutor<F>
...
+                | ForeignCall::GetMockLastParams
+                | ForeignCall::SetMockReturns
+                | ForeignCall::SetMockTimes
+                | ForeignCall::ClearMock => {
+                    panic!("unexpected mock call: {}", foreign_call.function)
+                }
+                _ => {}
+            }
+        }
+        Err(ForeignCallError::NoHandler(foreign_call.function.clone()))
```
TomAFrench, line 205:
```suggestion
        if let Some(
            ForeignCall::CreateMock
            | ForeignCall::SetMockParams
            | ForeignCall::GetMockLastParams
            | ForeignCall::SetMockReturns
            | ForeignCall::SetMockTimes
            | ForeignCall::ClearMock,
        ) = ForeignCall::lookup(foreign_call_name)
        {
            panic!("unexpected mock call: {}", foreign_call.function)
        }
        Err(ForeignCallError::NoHandler(foreign_call.function.clone()))
```

tooling/nargo/src/ops/test.rs
TomAFrench: Maybe something shorter as this paragraph is now in documentation for `Empty` but we want to explain why the `TestForeignCallExecutor` is doing what it does.

tooling/nargo/src/ops/test.rs
```diff
@@ -44,16 +35,19 @@ impl TestStatus {
...
     context: &mut Context,
     test_function: &TestFunction,
-    output: PrintOutput<'_>,
-    foreign_call_resolver_url: Option<&str>,
-    root_path: Option<PathBuf>,
-    package_name: Option<String>,
+    output: PrintOutput<'a>,
     config: &CompileOptions,
-) -> TestStatus {
+    foreign_call_executor: F,
```
TomAFrench, line 44:
```suggestion
    build_foreign_call_executor: F,
```

To make it clearer that we're not passing in an actual executor.

tooling/nargo/src/ops/test.rs
```diff
@@ -278,85 +266,41 @@ fn check_expected_failure_message(
...
+impl<E> TestForeignCallExecutor<E> {
+    #[allow(clippy::new_ret_no_self)]
+    fn new<'a, F>(output: PrintOutput<'a>, foreign_call_executor: &F) -> Self
+    where
+        F: Fn(PrintOutput<'a>, layers::Unhandled) -> E + 'a,
+    {
+        // Use a base layer that doesn't handle anything, which we handle in the `execute` below.
+        let executor = foreign_call_executor(output, layers::Unhandled);
+
+        Self { executor, encountered_unknown_foreign_call: false }
```
TomAFrench, line 283:
```suggestion
    fn new(executor: &E) -> Self {
        Self { executor, encountered_unknown_foreign_call: false }
```

I don't think there's any benefit of passing in anything other than the internal foreign call executor, right? We can call `foreign_call_executor` within `run test` and just pass it in here.

I'm also skeptical about passing ``layers::Unhandled` as we're doing here, we have no guarantees that `foreign_call_executor` will use the specified `base` layer so we're just relying on convention. What do you think of just documenting this requirement and relying on the user to use this base layer?

TomAFrench (commented):
Looks good, some nits on how we construct the `TestForeignCallExecutor`.

TomAFrench:
I'm going to fix the conflicts in this PR caused by #6865

TomAFrench:
As you're out today I'll just make the above changes so we can get master green again.

TomAFrench (approved):
LGTM

PR #6850, opened 2024-12-17
chore: Remove ValueIds from SSA

jfecher:
# Description

## Summary\*

Removes `ValueId`s from SSA to reduce memory usage and hopefully improve compiler performance as well.

- `ValueId`s are replaced by the `Value` enum itself, which is now thinner and implements `Copy`.
- Several functions for iterating over values (`block_parameters`, `instruction_results`) now no longer require references to the DFG, so we don't need as much cloning in several passes.
- Related to the previous point, values can be created much more often without referencing the DFG at all now, e.g. `Value::constant` just needs a field and a type since we don't need the separate map to keep each valueid unique to each constant value any more. Similarly, even `Value::Instruction` only needs an instruction id and a position. No need to store instruction id values in an external map in the dfg. This means we can get rid of quite a few fields on the DFG to save memory.
- `DFG::set_value_from_id` can no longer mutate the value a ValueId points to, and instead relies solely on the internal map between values.

## Additional Context

The integration tests are currently all passing, but there is a stack overflow somewhere within the test runner when running the unit tests, in between running the actual tests.

I have not started migrating all the SSA parser tests in the evaluator, these are expected to be failing currently since the display of valueids has changed.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
Initial results are very ungood, I think this is because FieldElement is 256bits causing Value to be 320 bits.
I'll re-add constant ids which should bring Value down to 96 bits. If we could embed the enum tag in the value somehow we could get it to 64.

jfecher:
Value's size has been reduced, although I'm unsure why the changes aren't more dramatic. I'd think removing the `instruction_results` map alone would have been a decent improvement. We're also cloning Vecs less throughout. I think I'll have to profile this more before fixing the ssa parser but as is this may not be worth merging with the regression in ram_blowup_regression.

TomAFrench:
Potentially an increased size of the replaced values map due to holding the actual values now?

jfecher:
Yes, but even then the increase is only from 32 to 64 bits which I'd expect to be offset by the other removals in this PR.

jfecher:
From manually timing programs, it seems this PR is about ~7% faster compiling programs compared to master

jfecher:
After some more perf testing, I'm going to close this PR. Testing it against master on rollup-base-private, this PR is compiling in ~90s compared to master's ~60s. Unsure if this is from the same regression as conditional_1 but this PR has been enough of a rabbit hole already that I'd rather close it.

PR #6855, opened 2024-12-18
fix: double alias in path

### LINKED ISSUE #6854: Function call with path that points to alias that points to alias panics
### Aim

This code:

```noir
struct Foo {}

impl Foo {
    fn new() -> Self {
        Self {}
    }
}

type FooAlias1 = Foo;
type FooAlias2 = FooAlias1;

fn main() { 
    let _ = FooAlias2::new();
}
```

### Expected Behavior

Should compile.

### Bug

Currently panics:

```
Message:  Type alias in path not pointing to struct not yet supported
Location: compiler/noirc_frontend/src/elaborator/path_resolution.rs:231
```

The reason is that we currently support aliases that point to structs when looking up methods like that, but we forgot that an alias could also point to an alias that eventually points to a struct. This should be relatively easy to solve.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #6854

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/path_resolution.rs
+++ b/compiler/noirc_frontend/src/elaborator/path_resolution.rs
@@ -9,7 +9,7 @@ use crate::hir::resolution::visibility::item_in_module_is_visible;
 
 use crate::locations::ReferencesTracker;
 use crate::node_interner::{FuncId, GlobalId, StructId, TraitId, TypeAliasId};
-use crate::Type;
+use crate::{Shared, Type, TypeAlias};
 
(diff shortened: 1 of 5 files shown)


jfecher (approved):
LGTM

PR #6857, opened 2024-12-18
feat!: Composable foreign call handlers

aakoshh:
# Description

## Problem\*

Followup for https://github.com/noir-lang/noir/pull/6849

## Summary\*

Adds a `foreign_calls::layer::Layer` construct which is used to compose `ForeignCallExecutor` instances. `DefaultForeignCallExecutor`, `DebugForeignCallExecutor` and `TestForeignCallExecutor` are reimplemented with this layering. 

The `DefaultForeignCallExecutor` allows us to inject a base layer, which the `TestForeignCallExecutor` uses to change the behaviour when nothing handles the call; this way we can reuse the default composition without having to repeat the logic in the test, similar to how the `DefaultDebugForeignCallExecutor` did it earlier.

Follow up for this would be:
* Make the _mock_ layer in the `DefaultForeignCallExecutor` optional, so that during normal execution no contract can install a mock and then handle foreign calls that should hit URLs. A `DisabledMockForeignCallExecutor` has been added in this PR, but it's unused. (I'm not sure this is a legit concern, but I didn't see any reason it wouldn't work).
* Make `run_test` accept a generic function to create a `ForeignCallExecutor` on top of the `Unhandled` base handler, instead of explicitly creating a `DefaultForeignCallExecutor`, so that we don't depend on `RPCForeignCallExecutor` in the tests, and then we can compile them to Wasm.  https://github.com/noir-lang/noir/pull/6858

## Additional Context

It's a breaking change because it changes the return type of `DefaultForeignCallHandler::new`, although the way it's called is the same for now.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/debugger/src/foreign_calls.rs
+++ b/tooling/debugger/src/foreign_calls.rs
@@ -4,8 +4,8 @@ use acvm::{
     AcirField, FieldElement,
 };
 use nargo::{
-    foreign_calls::{DefaultForeignCallExecutor, ForeignCallExecutor},
-    PrintOutput,
+    foreign_calls::{layers::Layer, ForeignCallExecutor},
+    PrintForeignCallExecutor, PrintOutput,
 };
 use noirc_artifacts::debug::{DebugArtifact, DebugVars, StackFrame};
 use noirc_errors::debug_info::{DebugFnId, DebugVarId};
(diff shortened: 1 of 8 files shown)


asterite (approved):
Looks great!

If I'm understanding things correctly, `DefaultForeignCallExecutor` used to have three internal executors and it would try them one after the other. Now in this PR they are layered on top of each other, so the effect is the same... except that we end up without duplicated code.

aakoshh:
That's right, the `DefaultForeignCallExecutor` knew exactly which requests should go where, and then the handlers it dispatched to _also_ inspected the call to decide whether they should be the ones handling it. Sometimes the `DefaultForeignCallExecutor` looked for this `NoHandler` error as a legitimate signal to carry on to the next handler. Namely the `MockForeignCallExecutor` may or may not be able to handle arbitrary calls, depending on the calls history came before. 

Then the `DefaultForeignCallExecutor` returned an empty result. To inject further logic before this happens, the `TestForeignCallExecutor` had to repeat the whole setup, which now it doesn't have to any more, although it could have been achieved in different ways I suppose.

With layering there is a chance to combine layers differently, and also to remove duplication by passing around builder functions. There is more reliance on the `NoHandlers` error mechanism, at the cost of cloning more strings, but I think we could even remove the wrapped `String` as the content might never be accessed.

In the next extension we could add something like `layers::Either` that could be for example either the `MockExecutor` or the `DisabledMockExecutor`, or add a flag to the `MockExecutor` would be fine as well.

TomAFrench (approved):
Very nice, I really like the removal of duplication between test and default executors.

PR #6860, opened 2024-12-18
feat: add a warning when using unsafe blocks without safety comments

### LINKED ISSUE #6478: Warn on `unsafe` blocks which don't have a safety comment
@nventuro posted in slack about how the fairies have been just wrapping `unsafe` blocks around unconstrained function calls without explaining why it's safe:

> tldr: whenever we use unsafe there should be a comment explaining why what's being done is ok
> either:
> a) we wrap the unsafe function in a safe function because we apply proper constraints (e.g. see the historical state reads)
> b) we explain why unconstrained usage is ok in this specific context (e.g. because we trust the sender already with our privacy, and so we trust them to choose good randomness)
> c) we don't use an unsafe block and leave the warning as a reminder that at some point in the future we need to revisit this and apply either a) or b)
> snippets like the above can be forgotten about, and the reader is clueless as to why unconstrained values are suddenly injected, and if/why that's ok

We should add a warning to encourage users to explain why their usage of `unsafe` blocks are safe rather then just wrapping the call in `unsafe` just to silence the other warning:

See https://github.com/rust-lang/rust-clippy/issues/9330

guipublic:
# Description

## Problem\*

Resolves #6478 

## Summary\*

Warn when an unsafe block does not start with a '//@safety' comment.

## Documentation\*

Check one:

# PR Checklist\*


noir_stdlib/src/meta/expr.nr
```diff
@@ -669,7 +669,12 @@ comptime fn new_unary_op(op: UnaryOp, rhs: Expr) -> Expr {
 
 comptime fn new_unsafe(exprs: [Expr]) -> Expr {
     let exprs = join_expressions(exprs, quote { ; });
-    quote { unsafe { $exprs }}.as_expr().unwrap()
+    quote { unsafe { 
+        //@safety: generated by macro
+        $exprs 
+        }}
```
michaeljklein, line 675:
```suggestion
    }}
```
Are these leading spaces from the formatter?

noir_stdlib/src/field/bn254.nr
```diff
@@ -74,7 +79,12 @@ pub fn decompose(x: Field) -> (Field, Field) {
 
 pub fn assert_gt(a: Field, b: Field) {
     if is_unconstrained() {
-        assert(unsafe { field_less_than(b, a) });
+        assert(
+            unsafe {
+                //@safety : unsafe in unconstrained*
```
michaeljklein, line 84: I'm not sure what this and the following "unsafe in unconstrained" means: does the `assert` make it safe? what's unsafe?

Also: spurious `*`
```suggestion
                //@safety : unsafe in unconstrained
```

michaeljklein (changes requested):
LGTM besides some docs/nits

asterite:
Have you considered alternatives to a new comment type?

One alternative is using doc comments, either on top of `unsafe`:

```noir
/// This is a safety comment
unsafe { ... }
```

or inside it:

```noir
unsafe {
    /// This is a safety comment
}
```

docs/docs/noir/concepts/unconstrained.md
```diff
@@ -96,6 +97,7 @@ This ends up taking off another ~250 gates from our circuit! We've ended up with
 
 Note that in order to invoke unconstrained functions we need to wrap them in an `unsafe` block,
 to make it clear that the call is unconstrained.
+Furthermore, the `unsafe` block is required to start with a `//@safety: ...` comment explaining why it is fine to call the unconstrained function.
```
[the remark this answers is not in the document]
guipublic: It's a warning for now, and I think it should not become a hard requirement, so I will change the doc.

noir_stdlib/src/field/bn254.nr
```diff
@@ -55,6 +59,7 @@ pub fn decompose(x: Field) -> (Field, Field) {
         compute_decomposition(x)
     } else {
         unsafe {
+            /*@safety  : decomposition is properly checked below*/
```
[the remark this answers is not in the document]
guipublic: Done. Long story short; I had to fight with the formatter :)

noir_stdlib/src/field/bn254.nr
guipublic: It's because this code is unconstrained due to the test above. I rephrased it to 'already unconstrained', it should be clearer?

guipublic:
> Have you considered alternatives to a new comment type?
> 
> One alternative is using doc comments, either on top of `unsafe`:
> 
> ```nr
> /// This is a safety comment
> unsafe { ... }
> ```
> 
> or inside it:
> 
> ```nr
> unsafe {
>     /// This is a safety comment
> }
> ```
> 
> Mainly because those doc comments already reach the parser and go into the AST (in some cases).

Yes I did:
- The comment on top of unsafe is harder to implement because the parser does not expect to have any comment left inside blocks, and I have don't have much preference regarding top vs inside.
- I am already using doc comments (but with a specific 'style'), so I believe it would be easy to switch to normal block comments. However I was thinking their semantic is a bit different.

So basically, I don't have strong opinion on this, I'm happy to change if you think it would be better.

noir_stdlib/src/meta/expr.nr
guipublic: I would have guess yes, but no, the formatter does not override them. It's a bit weird because it is not aligned with the next line however.

noir_stdlib/src/array/mod.nr
```diff
@@ -185,9 +185,10 @@ where
     /// ```
     pub fn sort_via<Env>(self, ordering: fn[Env](T, T) -> bool) -> Self {
         unsafe {
-            // Safety: `sorted` array is checked to be:
-            //   a. a permutation of `input`'s elements
-            //   b. satisfying the predicate `ordering`
+            /*@safety : `sorted` array is checked to be:
```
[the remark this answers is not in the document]
guipublic: Done

michaeljklein (approved):
LGTM

PR #6867, opened 2024-12-19
feat: warn on unnecessary unsafe blocks

### LINKED ISSUE #6866: Warn on unnecessary unsafe blocks
### Problem

Code like this:

```noir
fn main() {
    unsafe {
        foo();
    }
}

fn foo() {}
```

does not warn that the `unsafe` block is not needed. If the compiler does not tells us it's unneeded we could think the call is unsafe when it actually isn't.

Similarly, if there's `unsafe` nested inside another `unsafe`, the compiler should warn about it.

### Happy Case

The compiler detects `unsafe` blocks that don't have any unconstrained calls, or `unsafe` blocks that happen inside unconstrained calls.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #6866

## Additional Context

1. I checked `Aztec-Packages` and there are some occurrences of this warning (and also in the bignum library)
2. I learned that using `unsafe` inside a comptime context isn't needed, so the compiler will automatically warn about this (this happened in one place in the stdlib)

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/expressions.rs
+++ b/compiler/noirc_frontend/src/elaborator/expressions.rs
@@ -147,6 +147,12 @@ impl<'context> Elaborator<'context> {
     ) -> (HirExpression, Type) {
         // Before entering the block we cache the old value of `in_unsafe_block` so it can be restored.
         let old_in_unsafe_block = self.unsafe_block_status;
+        let is_nested_unsafe_block =
+            !matches!(old_in_unsafe_block, UnsafeBlockStatus::NotInUnsafeBlock);
+        if is_nested_unsafe_block {
+            self.push_err(TypeCheckError::NestedUnsafeBlock { span });
+        }
+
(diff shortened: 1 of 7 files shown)


compiler/noirc_frontend/src/elaborator/expressions.rs
```diff
@@ -410,6 +429,8 @@ impl<'context> Elaborator<'context> {
         method_call: MethodCallExpression,
         span: Span,
     ) -> (HirExpression, Type) {
+        dbg!(&method_call);
+
```
jfecher, line 433:
```suggestion
```

jfecher (approved):

asterite:
Oh, I just remembered the warning currently covers the entire `unsafe` block, while Rust only shows it on top of the `unsafe` keyword. I'll change it to do the same thing.

PR #6869, opened 2024-12-19
feat!: Disable mocks in `execute`

aakoshh:
# Description

## Problem\*

Follow up for https://github.com/noir-lang/noir/pull/6857

## Summary\*

Disables the `MockForeignCallExecutor` in `nargo execute` by returning a new `ForeignCallError::Disabled`. Also disables it in the ACVM execution.

## Additional Context

It looks like `std::test::OracleMock` is part of the Noir standard library regardless of whether we called `nargo execute` or `nargo test`, and as such we can create mocks in any program, e.g. the ones under `test_programs/execution_success`. AFAIU the intended use is for them to be used in Noir `#[test]` functions, to facilitate oracle testing in `aztec-packages`, and it wouldn't make sense for them to be used in production code. But if they were, I don't see anything that would prevent them from hijacking real oracle calls, which sounds to me like a security hole.

I thought about adding a CLI argument to `ExecuteCmd` to force-enable them, but I can't see under what conditions one wants to test these where they can't write a `#[test]`. Let me know if that is a use case.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_printable_type/src/lib.rs
```diff
@@ -69,6 +69,9 @@ pub enum PrintableValueDisplay<F> {
 
 #[derive(Debug, Error)]
 pub enum ForeignCallError {
+    #[error("Attempted to call disabled foreign call `{0}`")]
+    Disabled(String),
```
TomAFrench, line 73: Any reason why we can't just use `NoHandler` for this?

tooling/nargo/src/foreign_calls/layers.rs
```diff
@@ -122,6 +122,28 @@ impl<T> Layering for T {
     }
 }
 
+/// A case where we can have either this or that type of handler.
+pub enum Either<L, R> {
```
TomAFrench, line 126: We could then just have an optional layer here rather than needing an `Either` case.

aakoshh: Yeah, so the same argument went against the optional layer: because I didn't want to silently let calls to mocks in production programs go unhandled, I wanted the programmer to realise what there was something wrong. 

Instead of adding an `Either` I thought about adding an `enabled` flag to the existing mock handler, but somehow this seemed more attractive from an open-closed principle perspective.

tooling/nargo/src/foreign_calls/default.rs
```diff
@@ -15,24 +15,51 @@ use super::rpc::RPCForeignCallExecutor;
...
+
 impl<'a> DefaultForeignCallBuilder<'a> {
     /// Override the output.
     pub fn with_output(mut self, output: PrintOutput<'a>) -> Self {
         self.output = output;
         self
     }
 
+    /// Enable or disable mocks.
+    pub fn with_mocks(mut self, enabled: bool) -> Self {
```
michaeljklein, line 58: unused?

michaeljklein (approved):
LGTM with small comment r.e. unused `with_mocks`

tooling/nargo/src/foreign_calls/default.rs
aakoshh: It is, it looks like everywhere I wanted control over it I was already using the struct syntax, but because `with_output` exists, it looked like `with_mocks` is a nice to have as well for anyone who prefers the fluid syntax. It's a public type, so someone might use it I suppose. Or we could remove both, since the rest don't have them, although they could, because they go together I think, a method could nudge people to provide both.

michaeljklein: All good: just want to avoid accumulating unused code since `pub` interfaces can be tricky to remove.

PR #6871, opened 2024-12-19
chore: Bump arkworks to version `0.5.0`

winderica:
# Description

## Problem\*

(N/A)

## Summary\*

This PR bumps arkworks to version `0.5.0`, where `noir_grumpkin` is additionally replaced by `ark-grumpkin`.

The MSRV is also updated to `1.75.0` due to the requirement of the new arkworks.

## Additional Context

Note that two of the dependencies still rely on arkworks `0.4.0`:
- `light-poseidon`: There is already a [PR](https://github.com/Lightprotocol/light-poseidon/pull/44) that updates arkworks, which is thus used in https://github.com/noir-lang/noir/blob/5cf5e5c37e4388e6df1b35862b830397608dbe06/tooling/nargo_cli/Cargo.toml#L94
- `zkhash`: Unfortunately I failed to find the repository for `zkhash` and couldn't directly contribute to it. Instead of maintaining a fork, I tentatively copied its implementation into https://github.com/noir-lang/noir/blob/5cf5e5c37e4388e6df1b35862b830397608dbe06/acvm-repo/acvm/tests/solver.rs#L1214-L1468

Please let me know if you are fine with the solutions. I am happy to make changes based on your feedback :)

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Awesome, thank you for this PR! Ideally we'd have a crates.io release of light-poseidon before merging this as we're looking to remove our git dependencies.

We'll probably take a proper look at getting this merged in the new year.

winderica:
Thanks! In fact, we are also replacing git dependencies with crates.io ones in [sonobe](https://github.com/privacy-scaling-explorations/sonobe), and `acvm` is among the [last few libraries](https://github.com/privacy-scaling-explorations/sonobe/issues/186) that are still git-based forks, which is why I opened this PR 😄

I will watch `light-poseidon` and update this PR once a new version is released!

acvm-repo/acvm/tests/solver.rs
```diff
@@ -1210,6 +1209,264 @@ where
     fields.into_iter().map(|field| into_repr_vec(field)).collect()
 }
 
+// From https://docs.rs/crate/zkhash/0.2.0/source/ .
+// The original code depends on `ark_ff` v0.4.0, but here we use v0.5.0.
+mod zkhash {
```
TomAFrench, line 1214: One quick thought: I'd be more comfortable keeping the external dependency rather than vendoring this code as we're using is precisely to have a reference implementation. We could add an arkworks-0.4 dep and have `run_both_poseidon2_permutations` convert from 0.5 to 0.4 fields internally to get around this.

tooling/nargo_cli/Cargo.toml
```diff
@@ -91,7 +91,7 @@ iai = "0.1.1"
 test-binary = "3.0.2"
 test-case.workspace = true
 lazy_static.workspace = true
-light-poseidon = "0.2.0"
+light-poseidon = { git = "https://github.com/Lightprotocol/light-poseidon", rev = "c3bb062" }
```
TomAFrench, line 94: Looks like this dependency is completely unused so we can just get rid of it! (we ignore unused dev-deps due to false positives)

TomAFrench:
> acvm is among the https://github.com/privacy-scaling-explorations/sonobe/issues/186 that are still git-based forks, which is why I opened this PR

We've not pushed acvm releases to crates.io for a while so but we can take a look at prioritising fixing this publishing flow

tooling/nargo_cli/Cargo.toml
TomAFrench: Ah, nevermind. I searched for `light-poseidon` but forgot that `-` would get changed to `_` in the actual source.

TomAFrench: This is just in the test suite so we can tolerate the performance degradation from needing to convert to ark0.4 fields however.

TomAFrench (approved):

acvm-repo/acvm/tests/solver.rs
winderica: Sorry for being inactive recently due to some personal stuff. Yes, the conversion from ark 0.5 to 0.4 makes a lot of sense because both `zkhash` and `light-poseidon` are used only for tests, and it is definitely better as we no longer need to depend on a pending PR or manually copy the code. Thanks so much for pushing the changes already!

PR #6875, opened 2024-12-19
feat: flatten nested if-else statements with equivalent conditions

### LINKED ISSUE #6786: Example of an unexpectedly-huge contract artifact.
https://github.com/AztecProtocol/aztec-packages/issues/10546

In particular, `create_note` and `create_note_no_init_check`

https://github.com/AztecProtocol/aztec-packages/blob/6f209fb69fcce868c6e0fe9b79b5ac3f3a1e5c48/noir-projects/noir-contracts/contracts/stateful_test_contract/src/main.nr#L45

TomAFrench:
# Description

## Problem\*

Resolves #6786 

## Summary\*

The new way of handling predicates on store instructions was resulting in us having a stack of `load->if-else ->store->load ->if-else->store` where the else value of each if-else was the previous if-else. This means we get a huge stack of nested if-else instructions which share the same `then_condition` so they can be squashed down to just have the top "then" and bottom "else".

I've solved this by having the `simplify` instruction inspect the `then_value` and `else_value` of any if-else instructions to see if we're building a nested structure and then flatten it automatically.

This PR drops the size of `stateful-test-contract` (after being reduced to just having the `create_note` function) from 31.2MB to 12.2MB.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/instruction.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/instruction.rs
@@ -958,9 +958,11 @@ impl Instruction {
                 }
             }
             Instruction::IfElse { then_condition, then_value, else_condition, else_value } => {
+                let then_condition = dfg.resolve(*then_condition);
+                let else_condition = dfg.resolve(*else_condition);
                 let typ = dfg.type_of_value(*then_value);
 
-                if let Some(constant) = dfg.get_numeric_constant(*then_condition) {
+                if let Some(constant) = dfg.get_numeric_constant(then_condition) {
                     if constant.is_one() {
                         return SimplifiedTo(*then_value);
                     } else if constant.is_zero() {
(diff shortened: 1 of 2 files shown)


TomAFrench:
Need to modify some test names.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -974,10 +976,51 @@ impl Instruction {
...
+                            let instruction = Instruction::IfElse {
+                                then_condition,
+                                then_value,
+                                else_condition: inner_else_condition,
+                                else_value: inner_else_value,
+                            };
+                            return SimplifiedToInstruction(instruction);
+                        }
+                        // TODO: We could check to see if `then_condition == inner_else_condition`
+                        // but we run into issues with duplicate NOT instructions.
```
TomAFrench, line 1019:
```suggestion
                        // but we run into issues with duplicate NOT instructions having distinct ValueIds.
                    }
                };

                if let Value::Instruction { instruction, .. } = &dfg[else_value] {
                    if let Instruction::IfElse {
                        then_condition: inner_then_condition,
                        else_condition: inner_else_condition,
                        else_value: inner_else_value,
                        ..
                    } = dfg[*instruction]
                    {
                        if then_condition == inner_then_condition {
                            let instruction = Instruction::IfElse {
                                then_condition,
                                then_value,
                                else_condition: inner_else_condition,
                                else_value: inner_else_value,
                            };
                            return SimplifiedToInstruction(instruction);
                        }
                        // TODO: We could check to see if `then_condition == inner_else_condition`
                        // but we run into issues with duplicate NOT instructions having distinct ValueIds.
```

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -974,10 +976,51 @@ impl Instruction {
...
+                        if then_condition == inner_then_condition {
+                            let instruction = Instruction::IfElse {
+                                then_condition,
+                                then_value,
+                                else_condition: inner_else_condition,
+                                else_value: inner_else_value,
+                            };
+                            return SimplifiedToInstruction(instruction);
+                        }
+                        // TODO: We could check to see if `then_condition == inner_else_condition`
```
vezenovm, line 1018: How about `else_condition  == inner_else_condition`? And should we make an issue for these two TODOs

vezenovm (approved):

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
TomAFrench: This would never trigger as we always insert a new NOT operation when creating an if_else instruction. This means that we're never gonna be able to match on the else condition value ID as they are all completely fresh and only used in that if-else.

TomAFrench: In order to do this we would I think we would need a map to track known NOTted value ids so that we can reuse these values in future rather than inserting new a instruction

TomAFrench:
Chucking in the merge queue while I create issues.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
vezenovm: Ah yeah I see

jfecher (approved):

PR #6874, opened 2024-12-19
chore(ci): Take averages for compilation and execution report of small programs

vezenovm:
# Description

## Problem\*

No issue as just a small improvement on existing compiler reports

## Summary\*

For very small programs (e.g. less than 2 seconds) the compilation and execution reports can be deceiving as they can vary greatly per run. This PR adds logic to take an average of certain small programs to get more accurate reports.

## Documentation\*

Check one:
- [X] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [X] I have tested the changes locally.
- [X] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #6880, opened 2024-12-19
chore(ci): Display times in compilation and execution reports only with seconds

vezenovm:
# Description

## Problem\*

In https://github.com/noir-lang/noir/pull/6874 we moved to computing an average of multiple runs of `nargo compile` and `nargo execute`. The noir-bench-report still expects our time input to have the format of the output from the `time` command (e.g. `[0-9]+m[0-9]+.[0-9]+s`). In that PR we just prepended `0m` as to not to have to convert back to minutes. This suffix clutters the report output. 

## Summary\*

In this PR I decided to just keep with total seconds rather than converting back to minutes and seconds. It is simpler to work with one unit in my opinion. If we have reports that really need to be represented in minutes, we can address that at the time (if we have reports of that size it most likely means we have larger issues than how we display our reports). 

Updates to the commit in this PR https://github.com/noir-lang/noir-bench-report/pull/6

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/.github/workflows/reports.yml
+++ b/.github/workflows/reports.yml
@@ -425,7 +425,7 @@ jobs:
 
       - name: Parse compilation report
         id: compilation_report
-        uses: noir-lang/noir-bench-report@0d7464a8c39170523932d7846b6e6b458a294aea
+        uses: noir-lang/noir-bench-report@81d04cbb962b6dbc84db866d5dd5f1fdaba81ad0
         with:
           report: compilation_report.json
           header: |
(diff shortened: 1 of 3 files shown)


vezenovm:
Although the diff percentage shows `Infinity%`, the next report will display correct numbers based off the report generated on the parent branch https://github.com/noir-lang/noir/pull/6881.

jfecher (approved):
> Infinity%

I'm not sure, may have to block this until perf improves. Joking of course :+1:

PR #6882, opened 2024-12-19
feat: require trait function calls (`Foo::bar()`) to have the trait in scope (imported)

### LINKED ISSUE #6864: Require trait definition to be in scope to be able to call trait methods on type.
In the below program we have a trait `Foo` which is private to the module `private_mod`, however we can call the `foo()` method on `Bar` in `main` despite the root module not being supposed to know about `Foo` at all.

```nr
fn main() {
    assert_eq(Bar::foo(), 42);
}

pub struct Bar {}

mod private_mod {
    trait Foo {
        fn foo() -> Field;
    }

    impl Foo for super::Bar {
        fn foo() -> Field {
            42
        }
    }
}
```

This means that a trait used for internal implementation details (such as in #6845) leaks into user code. The way that Rust addresses this issue is by requiring the trait definition to be in scope in order to call trait methods.

asterite:
# Description

## Problem

Part of #6864

## Summary

This PR is only about calls (`Foo::bar()`).

There are many more cases to handle (methods, but also calls on primitive types like `Field::foo()` and method calls on primitive types) but it might be better to do all of this across many PRs so they are easier to review.

## Additional Context

1. To avoid introducing a huge breaking change, if you call a trait function, you don't have it imported, and there's only one possible trait function to call, that's reported as a warning.
2. We can document these new requirements in the final PR.

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/tests/traits.rs
TomAFrench, line 1: For completeness sake, maybe we should add a test for the case where there's two potential implementations defined but only one is imported?

```nr
    use private_mod::Foo;
    fn main() {
        let _ = Bar::foo();
    }
    pub struct Bar {
    }
    mod private_mod {
        pub trait Foo {
            fn foo() -> i32;
        }
        impl Foo for super::Bar {
            fn foo() -> i32 {
                42
            }
        }
        pub trait Foo2 {
            fn foo() -> i32;
        }
        impl Foo2 for super::Bar {
            fn foo() -> i32 {
                42
            }
        }
    }
```

TomAFrench (approved):
LGTM with extra test case.

Issue #6878, opened 2024-12-19
Add handling for negated conditions when folding `IfElse` instructions.


TomAFrench:
In #6875 we added an optimisation such that we'd flatten the below nested if statement such that it became the simple if-statement below.

```nr
// Let then_condition_1 == then_condition_2

if then_condition_1 {
   then_value_1
} else if else_condition_1 {
   if then_condition_2 {
     then_value_1
   } else if else_condition_2 {
     else_value
   }
}

// Propagate conditions such that

if then_condition {
   then_value_1
} else if else_condition_1 {
     else_value
}
```

In theory we could perform a similar simplification for the case below:

```nr
Let then_condition_1 == else_condition_2

if then_condition_1 {
   then_value_1
} else if else_condition_1 {
   if then_condition_2 {
     then_value_1
   } else if else_condition_2 {
     else_value
   }
}

// Propagate conditions such that

if then_condition {
   then_value_1
} else if else_condition_1 {
    then_value_1
}
```

This is hindered by the fact that the `else_condition`s for if-else instructions in Noir's SSA are constructed on the fly so that they're very unlikely to share a valueid with a different value in the circuit. When we make this more reliable we can enable this optimization.

PR #6888, opened 2024-12-19
chore: Add missing `IfElse` simplification cases

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6878

## Summary\*

Wanted a separate PR for this change so I could see the effect it had alone

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Yeah, I think this is a much less likely optimization tbh as we need to get reversed conditions as opposed to being able to simplify all of the if-elses in a particular branch which will naturally share the same condition.

jfecher:
Closing this for now, it's unclear that it's needed so I'm erring on the side of less code to maintain

PR #6890, opened 2024-12-20
fix: remove unnecessary cast in bit-shift

### LINKED ISSUE #6788: Unoptimized BinaryOp::Shl
In `insert_wrapping_shift_left`, `rhs` value [casted](https://github.com/noir-lang/noir/blob/1b0dd4149d9249f0ea4fb5e2228c688e0135618f/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs#L130) to unsinged with size equal to size of lhs. Therefore, the number of iterations in the pow function increases.

`insert_shift_right‎` works fine

## u32 << u8
```
acir(inline) fn main f0 {
  b0(v0: u32, v1: u8):
    v5 = lt v1, u8 32
    v6 = cast v5 as u32
    v7 = cast v1 as u32
    v9 = call to_le_bits(v7) -> [u1; 32]
    v12 = array_get v9, index Field 31 -> u1
...
```

## u80 << u8
```
acir(inline) fn main f0 {
  b0(v0: u80, v1: u8):
    v5 = lt v1, u8 80
    v6 = cast v5 as u80
    v7 = cast v1 as u80
    v9 = call to_le_bits(v7) -> [u1; 80]
    v12 = array_get v9, index Field 79 -> u1
...
```

## u80 >> u8
```
acir(inline) fn main f0 {
  b0(v0: u80, v1: u8):
    v5 = call to_le_bits(v1) -> [u1; 8]
    v8 = array_get v5, index Field 7 -> u1

...
```

guipublic:
# Description

## Problem\*

Resolves #6788 

## Summary\*
Since bit shift are now forced to be done with u8 rhs operand, we do not need to cast it to unsigned type anymore.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
@@ -128,9 +128,7 @@ impl Context<'_> {
             let bit_size_var = self.numeric_constant(FieldElement::from(bit_size as u128), u8_type);
             let overflow = self.insert_binary(rhs, BinaryOp::Lt, bit_size_var);
             let predicate = self.insert_cast(overflow, typ);
-            // we can safely cast to unsigned because overflow_checks prevent bit-shift with a negative value
-            let rhs_unsigned = self.insert_cast(rhs, NumericType::unsigned(bit_size));
-            let pow = self.pow(base, rhs_unsigned);
+            let pow = self.pow(base, rhs);
             let pow = self.insert_cast(pow, typ);
             (FieldElement::max_num_bits(), self.insert_binary(predicate, BinaryOp::Mul, pow))
         };


TomAFrench:
I'm surprised that we don't automatically noop casts when we're inserting them into the dfg. Can you add this to the `simplify` function?

jfecher:
There is:
>     } else if *dst_typ == dfg.type_of_value(value) {
>        SimplifiedTo(value)

already in `simplify_cast`. We should investigate why this isn't sufficient (are we returning before that else if?)

guipublic:
Cast are properly simplified, the issue was that the number of bits to shift was casted to the type of the number to be shifted. This is because originally:
1. the shift operation required operands to be of the same type
2. In case of signed operands, we were converting the number of bits to unsigned value for internal computation (the power operation)

When we shifted to u8 operand, the unsigned cast was left, which does not add any constraint per-se but change the type of the operand which indirectly affected the power operation (in the to_bits() function).

jfecher (approved):

Issue #6831, opened 2024-12-16
Strip `IncRC` instructions from SSA for ACIR functions before reaching ACIRgen


TomAFrench:
In ACIR functions we currently allow `IncrementRC` instructions to sit in the SSA all the way up to ACIRgen where they're eventually discarded. This means that we need to spend time processing them and memory storing them, their (empty) results and callstack entries.

We can instead just discard them at any point after the runtime separation pass so we don't need to consider them at all.

aakoshh:
If I am reading it correctly, `Rc` instructions can be created in `loop_invariants.rs` and `constant_folding.rs`, so if I was to change `convert_ssa_instruction` to raise treat `Rc` as `unreachable!` the removal of these instructions from ACIR runtimes would need to happen potentially twice: once after runtime separation, which would reduce the number of instructions, and once again before ACIR generation, because the constant folding happens towards the end of the passes. Or those two passes would need to be smarter and not insert these instructions unless the runtime is Brillig, which would probably be a better idea, as I we wouldn't have to second guess in each pass whether anything could have re-introduced these instructions. 

To be clear, you're worried about the fact that they are there taking up extra slots in the instruction vectors; the compiler will force us to consider them anyway as long as they are part of `Instructions` and we have common SSA passes for both runtimes, right?

I wonder if there would be a different way of handling this issue via #6841, some kind of separation of the instruction set itself, with some phantom types indicating what kind of runtime we're passing over, so that no pass can create such instructions, and then we can be sure that we really don't need to consider them because they cannot exist 🤔

jfecher:
Yes, if unconstrained functions are separated before SSA we could have a simple SSA check to see if the function is constrained, and if so avoid inserting any rc instructions to begin with.

PR #6893, opened 2024-12-20
feat(ssa): Immediately simplify away RefCount instructions in ACIR functions

aakoshh:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6831

## Summary\*

Assuming the runtime separation has already happened, prevents `IncrementRc` and `DecrementRc` from being added to ACIR functions by moving the `runtime` field from the `Function` to the `DataFlowGrap`, and modifying `insert_instruction_and_results` to immediately return `InstructionRemoved` if the `Instruction` would not be handled by the current runtime. This is to prevent e.g. `constant_folding` and `loop_invariants` to re-introduce these instructions, and so that we don't constantly have to keep it in mind that we should add them only in one runtime, but not the other.

## Documentation\*

Check one:

# PR Checklist\*


asterite (commented):
Looks good so far!

TomAFrench:
I think we should merge https://github.com/noir-lang/noir/pull/6894 before this as we can then remove the runtime separation changes in this PR.

TomAFrench (approved):
LGTM

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -100,6 +105,34 @@ pub(crate) struct DataFlowGraph {
...
+            values: Default::default(),
+            constants: Default::default(),
+            functions: Default::default(),
+            intrinsics: Default::default(),
+            foreign_functions: Default::default(),
+            blocks: Default::default(),
+            replaced_value_ids: Default::default(),
+            locations: Default::default(),
+            call_stack_data: Default::default(),
+            data_bus: Default::default(),
```
vezenovm, line 122:
```suggestion
            ..Default::default()
```
Should be able to just do this

aakoshh: Ah yes, I just accepted the editor filling it in without thinking 🤦‍♂️ Thanks.

aakoshh: Hm, no, `Runtime` doesn't have a `Default`, so even if I can't add back `#[derive(Default)]` to `DataFlowGraph` to enable this syntax.

vezenovm: We could simply add a default for `RuntimeType` (we do this for `InlineType`). But I won't block on this as this is a minor syntax choice.

aakoshh: Okay, added it! I suppose it was the default anyway by the virtue of `Function::new` assigning a default value to its own field.

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -194,14 +218,18 @@ impl DataFlowGraph {
...
     pub(crate) fn insert_instruction_and_results(
         &mut self,
         instruction: Instruction,
         block: BasicBlockId,
         ctrl_typevars: Option<Vec<Type>>,
         call_stack: CallStackId,
     ) -> InsertInstructionResult {
+        if !self.is_handled_by_runtime(&instruction) {
+            return InsertInstructionResult::InstructionRemoved;
+        }
```
jfecher, line 232: This check shouldn't be needed anymore since https://github.com/noir-lang/noir/pull/6894 already prevents inc/dec rc from being inserted into acir functions. They'd only be able to be reintroduced via at inlining optimization pass which inlines brillig code into an acir function, but that shouldn't be possible anyway.

Perhaps we'll have more brillig-only instructions in the future though? (e.g. separating array semantics)?

jfecher: Also (unrelated), with runtime in the dfg now... do we really need both dfg and Function? Function now is just a wrapper around a dfg and an entry block. IIRC we copied this from cranelift originally but I'm unsure of the motivation there. Perhaps Function had more fields.

TomAFrench: Ah that's true. I didn't notice the circuit diff disappearing from this PR.

aakoshh: I saw a runtime check added [here](https://github.com/noir-lang/noir/pull/6894/files#diff-ac8c4aa901333c53b981ac317926a12485d73931bb79c1de9d51a707ba017c9dR472), but AFAICS [constant_folding](https://github.com/noir-lang/noir/blob/ab8807d27b0442bd7e9ff73050f44bacea3add16/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs#L319) and [loop_invariant](https://github.com/noir-lang/noir/blob/ab8807d27b0442bd7e9ff73050f44bacea3add16/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs#L121) can still add them directly. 

FWIW if I return `true` from `is_handled_by_runtime` then we get the following integration test failures because of the panic "Expected all Rc instructions to be removed before acir_gen":
```
    tests::execution_success::test_6::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_array_dynamic_blackbox_input::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_array_to_slice::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_bigint::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_brillig_identity_function::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_conditional_regression_short_circuit::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_debug_logs::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_hashmap::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_hint_black_box::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_merkle_insert::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_nested_array_dynamic::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_pedersen_check::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_pedersen_hash::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_regression_4449::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_sha256::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_sha256_regression::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_sha256_var_padding_regression::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_sha256_var_size_regression::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_sha256_var_witness_const_regression::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_sha2_byte::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_simple_shield::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_slice_dynamic_index::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_slice_regex::forcebrillig_false_inliner_i64_min_expects
    tests::execution_success::test_slices::forcebrillig_false_inliner_i64_min_expects
```

vezenovm: > I saw a runtime check added [here](https://github.com/noir-lang/noir/pull/6894/files#diff-ac8c4aa901333c53b981ac317926a12485d73931bb79c1de9d51a707ba017c9dR472), but AFAICS [constant_folding](https://github.com/noir-lang/noir/blob/ab8807d27b0442bd7e9ff73050f44bacea3add16/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs#L319) and [loop_invariant](https://github.com/noir-lang/noir/blob/ab8807d27b0442bd7e9ff73050f44bacea3add16/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs#L121) can still add them directly.

We probably just need to have similar runtime checks in those passes which add the rc instructions directly.

aakoshh: Yes, if we add the checks in the passes directly then we can roll back this entire PR. Adding them in the passes might be the only option if the runtime-specific pattern is more complicated than an individual instruction. 

I did it this way because 
1) the ticket suggested there is value in not having instructions which will be just ignored later on and 
2) I thought that the passes that added them could have always been written in a way that they don't add them unless the runtime is meant to process them, so if they weren't, maybe the intention was to keep them agnostic, or at least not more complicated. 

It seemed like silently skipping such instructions was the lightest touch, but I see the value in handling them in the passes as well, saving the extra checks for every instruction that is added.

PR #6894, opened 2024-12-20
chore: Separate unconstrained functions during monomorphization

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6841
Resolves https://github.com/noir-lang/noir/issues/6831

## Summary\*

Adds a boolean is_unconstrained as an extra key during monomorphization. This should separate constrained from unconstrained functions. This key is infectious so every function called from an unconstrained function will also be unconstrained.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/monomorphization/mod.rs
```diff
@@ -111,24 +125,29 @@ type HirType = crate::Type;
...
+    monomorphizer.in_unconstrained_function = force_unconstrained;
     let function_sig = monomorphizer.compile_main(main)?;
 
     while !monomorphizer.queue.is_empty() {
-        let (next_fn_id, new_id, bindings, trait_method, location) =
+        let (next_fn_id, new_id, bindings, trait_method, is_unconstrained, location) =
             monomorphizer.queue.pop_front().unwrap();
         monomorphizer.locals.clear();
 
+        monomorphizer.in_unconstrained_function = is_unconstrained;
```
vezenovm, line 149:
```suggestion
        monomorphizer.in_unconstrained_function = is_unconstrained || force_unconstrained;
```

vezenovm: Then I think we can leave this assert https://github.com/noir-lang/noir/blob/186fbb75c11fe575afd0d462f837331ccc307889/compiler/noirc_evaluator/src/ssa.rs#L279.

compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -276,19 +271,12 @@ pub fn create_program(
         (generated_acirs, generated_brillig, brillig_function_names, error_types),
         ssa_level_warnings,
     ) = optimize_into_acir(program, options)?;
-    if options.force_brillig_output {
```
vezenovm, line 279: I do think this assert is correct and should remain.

compiler/noirc_frontend/src/monomorphization/mod.rs
vezenovm: Reading the pass some more it looks like this should not be necessary

compiler/noirc_evaluator/src/ssa.rs
vezenovm: We could remove the option but case on whether we have a single main function that is marked unconstrained

vezenovm: As otherwise we are going to error out when asserting the number of acir artifacts against the number of function signatures.

vezenovm: We can also just update how we collect function sigs 
https://github.com/noir-lang/noir/blob/186fbb75c11fe575afd0d462f837331ccc307889/compiler/noirc_frontend/src/monomorphization/mod.rs#L146

`is_entry_point` is going to return true for functions marked with `#[fold]`. However, if we are force compiling to Brillig we do not need to collect this signature. The func sigs are only used for constructing acir artifacts. This feels like the best solution here.

TomAFrench:
This causes some regressions on circuit sizes which is not being shown due to a bug in the `noir-gates-diff` action. The fix is in https://github.com/noir-lang/noir/pull/6944.

TomAFrench:
> This causes some regressions on circuit sizes

Ah nice, looks like I was reading things backwards.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable.rs
```diff
@@ -0,0 +1,75 @@
...
+        }
+    };
+
+    for block_id in func.reachable_blocks() {
+        let block = &func.dfg[block_id];
+
+        for instruction_id in block.instructions() {
+            let instruction = &func.dfg[*instruction_id];
+
+            if matches!(instruction, Instruction::Store { .. } | Instruction::Call { .. }) {
```
aakoshh, line 66: Just for my understanding: what exactly does it mean to `Store` but never `Call` a function?

vezenovm: If we store a function in another value, we may eventually load the function from that value and call it.

vezenovm (approved):
Looks good once the debugger is patched. I see that we do have some monomorphization logic for the debugger under `monomorphization/debugger.rs`

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable.rs
aakoshh: That makes sense, I was just wondering if it's considered reachable even if it's never called, ie. if we only ever see a `Store` but never a `Call` for it.

vezenovm: That could possibly occur and yes technically that function be an unreachable. But we want to leave any `Store`/`Load` resolution to mem2reg. If a `Store` remains we should assume that the value we are storing is potentially used elsewhere in a separate block or even passed as a reference to another function.

aakoshh: Right, so removing a function that it referenced by something, even if not called, could be a problem. Thanks!

jfecher:
Looks like the debug tests are failing because the debugger generates code like the following:
```
brillig(inline) fn __debug_var_assign_inner f63 {
  b0(v0: u32, v1: Field):
    call v2(v0, v1)
    return
}
```
Which is invalid since there is no `v2`. This occurs on master as well

jfecher:
Made an issue for the debugger issue here: https://github.com/noir-lang/noir/issues/6955.
For now, I added a short circuit if no such function exists in the pass. In the future we probably shouldn't have the inliner removing functions still referenced by debugger functions.

PR #6895, opened 2024-12-20
feat: require trait method calls (`foo.bar()`) to have the trait in scope (imported)

asterite:
# Description

## Problem

Part of https://github.com/noir-lang/noir/issues/6864

## Summary

Similar to 6882 but for method calls (`foo.bar()`).

## Documentation

Check one:
- [ ] No documentation needed.
- [ ] Documentation included in this PR.
- [x] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #6901, opened 2024-12-20
feat!: require trait primitive functions/calls to have their trait in scope

asterite:
# Description

Based on top of #6895

## Problem

Part of https://github.com/noir-lang/noir/issues/6864

## Summary

Similar to the previous PRs but this time for primitive types (`Field::foo()` and `field.foo()`). In reality `Field::foo()` already worked before this PR (as a result of some previous PR) but I added tests to make sure that's the case.

This PR also introduces a refactor: previously methods were kept in two places, one for structs and one for everything else (keyed under `TypeMethodKey`). Now they are all in a single place, and there's a new `TypeMethodKey` variant for structs. This greatly simplifies/unifies the code.

## Documentation

Check one:

# PR Checklist


michaeljklein (approved):
LGTM

PR #6910, opened 2024-12-21
feat(ssa): Hoist add and mul binary ops using known induction variables

vezenovm:
# Description

## Problem\*

Followup to https://github.com/noir-lang/noir/pull/6848 whose largest regression was in brillig opcodes executed https://github.com/noir-lang/noir/pull/6848#issuecomment-2549554990. 

## Summary\*

In #6848 we marked certain binary ops as not being able to be hoisted due to the overflow checks inside of ACIR gen and Brillig gen. However, if a binary operation has one variable that is from an outer loop's induction variable we can determine whether or not there would be an overflow. This re-uses logic introduced in https://github.com/noir-lang/noir/pull/6639.

This PR currently just tries to hoist add and mul instructions as they are simpler. For example, `Sub` overflows depends on if the induction variable is on the lhs or the rhs (e.g. if it is the rhs we should check the induction variable's upper bound but if it is the lhs we should check the lower bound). These can be done in follow-ups if we are comfortable with this approach.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
@@ -15,7 +15,7 @@ use crate::ssa::{
         basic_block::BasicBlockId,
         function::Function,
         function_inserter::FunctionInserter,
-        instruction::{Instruction, InstructionId},
+        instruction::{binary::eval_constant_binary_op, Instruction, InstructionId},
         types::Type,
         value::ValueId,
     },
@@ -207,6 +207,7 @@ impl<'f> LoopInvariantContext<'f> {
 
         let can_be_deduplicated = instruction.can_be_deduplicated(self.inserter.function, false)
             || matches!(instruction, Instruction::MakeArray { .. })
+            || matches!(instruction, Instruction::Binary(_))
             || self.can_be_deduplicated_from_upper_bound(&instruction);
 
         is_loop_invariant && can_be_deduplicated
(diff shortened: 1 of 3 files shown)


compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -231,6 +232,31 @@ impl<'f> LoopInvariantContext<'f> {
...
+                let rhs_const =
+                    self.inserter.function.dfg.get_numeric_constant_with_type(binary.rhs);
+                let (lhs, rhs) = match (
+                    lhs_const,
+                    rhs_const,
+                    self.outer_induction_variables.get(&binary.lhs),
+                    self.outer_induction_variables.get(&binary.rhs),
+                ) {
+                    (Some((lhs, _)), None, None, Some(upper_bound)) => (lhs, *upper_bound),
+                    (None, Some((rhs, _)), Some(upper_bound), None) => (*upper_bound, rhs),
```
jfecher, line 254: Do we care that these middle cases are None or should any of them be match-all cases?

vezenovm: I'm not totally sure I understand the question. Could you elaborate? If we have `None` it just means we do not have a constant or an outer induction variable. We only care about the cases where if we have a constant, that the other side of the binary op is an induction variable from the outer loop.

jfecher: I was somewhat trying to understand if we needed no constant/outer induction variable or if using a wildcard match like `(Some((lhs, _)), _, _, Some(upper_bound))` would also be applicable here

vezenovm: Ah I see. I could use the wildcard. If we have something like `(Some((lhs, _)), _, _, Some(upper_bound))` it should be impossible for rhs_const to return anything other than `None` as that binary op should already be simplified. Same with fetching the induction variable for `binary.lhs`. If we have a constant lhs, that value should not also be an induction variable. I was thinking the explicit `None` makes this logic a bit clearer to understand, but I am also fine switching to the wildcard syntax.

jfecher: Ah, no that's alright. If it's clearer without the wildcard / is expected to always be None that's fine. I was just trying to understand this logic if it needed / was expected to be None or if there could ever be Some cases and what that'd mean in this context.

jfecher (approved):
:+1:

PR #6913, opened 2024-12-23
fix: require generic trait impls to be in scope to call them

asterite:
# Description

Based on top of #6901

## Problem

Part of https://github.com/noir-lang/noir/issues/6864

## Summary

Similar to the previous PRs, but for traits which are implemented over all types with an optional `where` clause, for example:

```noir
fn main() {
    let x: i32 = 1;
    let _ = x.foo(); // This will warn because there's no `use private_mod::Foo;`
}

mod private_mod {
    pub trait Foo<T> {
        fn foo(self) -> i32;
    }

    impl<T> Foo<T> for T {
        fn foo(self) -> i32 {
            42
        }
    }
}
```

This is the last one! 🎉 

## Documentation

Check one:

# PR Checklist


jfecher (approved):
LGTM

PR #6915, opened 2024-12-23
feat(LSP): suggest trait methods from where clauses

asterite:
# Description

## Problem

No issue, just something I noticed while working with traits.

## Summary

Previously only `into()` was suggested here:

lsp-suggest-trait-methods

## Additional Context

This currently doesn't work inside trait methods. While figuring out why I noticed we don't track the trait method location, so I added that. It still didn't work but then I remembered that we currently don't type-check trait methods at all... but I still left the code I changed before that so that when we implement type-checking trait methods it will start working.

## Documentation

Check one:

# PR Checklist


jfecher (approved):
:+1:

PR #6918, opened 2025-01-01
feat[embedded_curve_ops]: `Hash` and `Eq` for 

skaunov:
`EmbeddedCurvePoint` and `EmbeddedCurveScalar`

# Description

## Problem\*

Resolves none.

Allows comparing of the embedded curve values and using them in a `HashMap` among other `Hash` perks.

## Summary\*

`derive` macro applications to the structs.

## Additional Context

none

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/noir_stdlib/src/embedded_curve_ops.nr
+++ b/noir_stdlib/src/embedded_curve_ops.nr
@@ -1,9 +1,11 @@
 use crate::cmp::Eq;
 use crate::ops::arith::{Add, Neg, Sub};
+use crate::hash::Hash;
 
 /// A point on the embedded elliptic curve
 /// By definition, the base field of the embedded curve is the scalar field of the proof system curve, i.e the Noir Field.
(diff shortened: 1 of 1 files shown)


jfecher (approved):
LGTM

TomAFrench:
@skaunov looks like this is missing some imports.

jfecher (changes requested):
Sorry, I approved this a bit early thinking I could just suggest an import, but there's more than an import that'd need to change. After testing this locally, both EmbeddedCurvePoint and EmbeddedCurveScalar already have `Eq` impls - and the impl for EmbeddedCurvePoint differs from the default implementation.

We also can't derive Hash for these because of an internal issue with the Hash deriver - it creates code referencing `std::hash::Hasher` but you cannot refer to `std` within the `std` crate, you must use `crate::hash::Hasher`, so an error is produced. We'd need a manual Hash impl.

PR #6926, opened 2025-01-02
feat: don't report warnings for dependencies

### LINKED ISSUE #6909: Automatically filter out any warnings which are thrown in dependencies
It's not possible for the consumer of libraries to fix these themselves so there's no point in warning here. This just results in us having to go round and push out unnecessary new releases in various libraries to prevent these from showing in user code.

asterite:
# Description

## Problem

Resolves #6909

## Summary

The way this is done here is:
1. When collecting all files to parse, we also collect which files belong to root packages
2. When reporting errors and warnings we drop warnings from files that are not in root packages

## Additional Context

The name of the branch has "2" in it because at first I tried a different approach: avoid collecting warnings as soon as we encounter them in dependencies. However, this is much more complex because we can't immediately know if an error is a warning or not (we have to turn it into a Diagnostic first, or introduce a large refactor to track this separately) and also in LSP it makes sense to track errors from dependencies if they are tracked by the editor, so it doesn't make sense to exclude them right away.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_driver/src/lib.rs
+++ b/compiler/noirc_driver/src/lib.rs
@@ -311,6 +311,7 @@ pub fn check_crate(
     crate_id: CrateId,
     options: &CompileOptions,
 ) -> CompilationResult<()> {
+    let crate_files = context.crate_files(&crate_id);
     let diagnostics =
         CrateDefMap::collect_defs(crate_id, context, options.debug_comptime_in_file.as_deref());
     let warnings_and_errors: Vec<FileDiagnostic> = diagnostics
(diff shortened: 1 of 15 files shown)


TomAFrench:
Hmm, forcing the user to pass around another piece of compiler state is a little sad. I'd like to avoid this if possible but would need to think more on this.

asterite:
The set of root files could be tracked by `FileManager` so we'd need less things to pass around... though at that point maybe `FileManager` would be doing more just than a single thing.

TomAFrench:
I'm thinking that there's enough information for what we want inside of the `Context` struct, following this down we can get the `ModuleData` for a particular crate.

https://github.com/noir-lang/noir/blob/ebc4d2cf2b91658a10393733407f33d50a0faaf1/compiler/noirc_frontend/src/hir/mod.rs#L36
https://github.com/noir-lang/noir/blob/ebc4d2cf2b91658a10393733407f33d50a0faaf1/compiler/noirc_frontend/src/hir/def_map/mod.rs#L67
https://github.com/noir-lang/noir/blob/ebc4d2cf2b91658a10393733407f33d50a0faaf1/compiler/noirc_frontend/src/hir/def_map/module_data.rs#L15

By reading the `ModuleData` for a particular crate we can get a tree of all the modules within a particular crate. Working down from the root we can then use logic similar to `find_module` in order to get all of the `FileId`s connected to this crate.

https://github.com/noir-lang/noir/blob/ebc4d2cf2b91658a10393733407f33d50a0faaf1/compiler/noirc_frontend/src/hir/def_collector/dc_mod.rs#L1121-L1125

We should then be able to generate this list after the fact rather than needing to pass it around externally to the compiler.

Note that this method will mean that we ignore any warnings outside of the current crate as opposed to the current workspace. This seems to be closer to what we want imo as it will avoid a workspace library from having its warnings emitted many times (this also matches up with cargo behaviour).

TomAFrench:
We could also track this a little more explicitly inside of the compiler state but the above is a "zero footprint" solution.

asterite:
Almost there! In `compile_cmd` and `debug_cmd` there's no `Context` so I don't know where to get the files from... Those files end up calling `compile_program` that returns `CompilationResult<CompiledProgram>`, maybe that type (`CompiledProgram`) could hold that, or maybe a tuple could be returned... but I think it would make the code messier again.

TomAFrench:
We can filter the returned warnings inside of the compiler before we pass it back up to `nargo_cli` though, no? Is there a reason why we need to do this externally?

asterite:
> Is there a reason why we need to do this externally?

The main reason is that for LSP we'd still want to show warnings in crates that are open in the UI but are not the main crate. Though I guess not doing would be fine too...

asterite:
It worked! And for some reason it works well in LSP too, maybe the code there is slightly different, or maybe all crates are checked, not sure.

TomAFrench:
I'm getting zero warnings being thrown in aztec-packages when using this PR despite there definitely being warnings that should be thrown.

asterite:
Good catch, I might have changed something in the end that broke it. Now it's fixed.

TomAFrench (approved):
LGTM

PR #6941, opened 2025-01-03
feat: lock on Nargo.toml on several nargo commands

asterite:
# Description

## Problem

Something mentioned over Slack.

## Summary

nargo_cli's `build.rs` was using a combination of mutexes and file locks. Instead of only doing it there, this PR moves the file lock to almost every `nargo` command. That means that if you do `nargo execute` twice on the same directory, the second invocation will wait until the first one ends.

I changed from `file_lock` to `fs2` because it allows doing `try_lock_exclusive` so we can show the user a message saying we are waiting for a lock...

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/nargo_cli/build.rs
+++ b/tooling/nargo_cli/build.rs
@@ -177,32 +177,12 @@ fn generate_test_cases(
     }
     let test_cases = test_cases.join("\n");
 
-    // We need to isolate test cases in the same group, otherwise they overwrite each other's artifacts.
-    // On CI we use `cargo nextest`, which runs tests in different processes; for this we use a file lock.
-    // Locally we might be using `cargo test`, which run tests in the same process; in this case the file lock
-    // wouldn't work, becuase the process itself has the lock, and it looks like it can have N instances without
-    // any problems; for this reason we also use a `Mutex`.
-    let mutex_name = format! {"TEST_MUTEX_{}", test_name.to_uppercase()};
     write!(
         test_file,
         r#"
-lazy_static::lazy_static! {{
-    /// Prevent concurrent tests in the matrix from overwriting the compilation artifacts in {test_dir}
-    static ref {mutex_name}: std::sync::Mutex<()> = std::sync::Mutex::new(());
-}}
-
 {test_cases}
 fn test_{test_name}(force_brillig: ForceBrillig, inliner_aggressiveness: Inliner) {{
     let test_program_dir = PathBuf::from("{test_dir}");
-
-    // Ignore poisoning errors if some of the matrix cases failed.
-    let mutex_guard = {mutex_name}.lock().unwrap_or_else(|e| e.into_inner());
-
-    let file_guard = file_lock::FileLock::lock(
-        test_program_dir.join("Nargo.toml"),
-        true,
-        file_lock::FileOptions::new().read(true).write(true).append(true)
-    ).expect("failed to lock Nargo.toml");
 
     let mut nargo = Command::cargo_bin("nargo").unwrap();
     nargo.arg("--program-dir").arg(test_program_dir);
@@ -221,9 +201,6 @@ fn test_{test_name}(force_brillig: ForceBrillig, inliner_aggressiveness: Inliner
     }}
 
     {test_content}
-
-    drop(file_guard);
-    drop(mutex_guard);
 }}
 "#
     )
--- a/tooling/nargo_cli/src/cli/check_cmd.rs
+++ b/tooling/nargo_cli/src/cli/check_cmd.rs
@@ -1,4 +1,5 @@
 use crate::errors::CliError;
+use crate::lock::Lock;
 
 use clap::Args;
 use fm::FileManager;
@@ -34,6 +35,8 @@ pub(crate) struct CheckCommand {
 
 pub(crate) fn run(args: CheckCommand, config: NargoConfig) -> Result<(), CliError> {
     let toml_path = get_package_manifest(&config.program_dir)?;
+    let lock = Lock::lock(toml_path.clone());
+
     let selection = args.package_options.package_selection();
     let workspace = resolve_workspace_from_toml(
         &toml_path,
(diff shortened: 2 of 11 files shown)


tooling/nargo_cli/src/lock.rs
```diff
@@ -0,0 +1,26 @@
...
+        let file = File::open(toml_path).expect("Expected Nargo.toml to exist");
+        if file.try_lock_exclusive().is_err() {
+            eprintln!("Waiting for lock on Nargo.toml...");
+        }
+
+        file.lock_exclusive().expect("Failed to lock Nargo.toml");
+        Self { file }
+    }
+
+    pub(crate) fn unlock(&self) {
```
aakoshh, line 23: This looks like it could be called from `Drop` of `Lock` to make sure it happens.

tooling/nargo_cli/src/cli/check_cmd.rs
```diff
@@ -34,6 +35,8 @@ pub(crate) struct CheckCommand {
 
 pub(crate) fn run(args: CheckCommand, config: NargoConfig) -> Result<(), CliError> {
     let toml_path = get_package_manifest(&config.program_dir)?;
+    let lock = Lock::lock(toml_path.clone());
```
aakoshh, line 38: Have you considered doing the lock somewhere [here](https://github.com/noir-lang/noir/blob/cb0628b86a0e37db9d62d84e340c1a9ae254c340/tooling/nargo_cli/src/cli/mod.rs#L128-L129), so the lock/unlock pattern doesn't have to be repeated in every command? It would apply the ones that as you say don't produce artefacts, but would also make things simpler elsewhere; maybe they could use [lock_shared](https://docs.rs/fs2/latest/fs2/trait.FileExt.html#tymethod.lock_shared) so they don't block each other.

tooling/nargo_cli/build.rs
```diff
@@ -177,33 +177,13 @@ fn generate_test_cases(
     }
     let test_cases = test_cases.join("\n");
 
-    // We need to isolate test cases in the same group, otherwise they overwrite each other's artifacts.
-    // On CI we use `cargo nextest`, which runs tests in different processes; for this we use a file lock.
-    // Locally we might be using `cargo test`, which run tests in the same process; in this case the file lock
-    // wouldn't work, becuase the process itself has the lock, and it looks like it can have N instances without
-    // any problems; for this reason we also use a `Mutex`.
-    let mutex_name = format! {"TEST_MUTEX_{}", test_name.to_uppercase()};
```
aakoshh, line 185: Great, you solution is nicer indeed, no need for this or that kind of lock 👍 
I didn't want to modify the CLI, but I agree it's better this way.

tooling/nargo_cli/src/lock.rs
asterite: I thought about this, I'll try it. When I used it it was less clear, I think we'd still need to explicitly call `drop`, otherwise it's unlocked as soon as possible.

tooling/nargo_cli/src/cli/check_cmd.rs
asterite: Oh, good point! I'll move it there.

> maybe they could use [lock_shared](https://docs.rs/fs2/latest/fs2/trait.FileExt.html#tymethod.lock_shared) so they don't block each other

I'm not sure about this. The file already exists and nobody writes to it... and if they don't block each other then it kind of defeats the purpose of this PR? But I must be misunderstanding something.

tooling/nargo_cli/src/lock.rs
aakoshh: I know what you mean, I am also never quite sure when Rust would drop an unused guard, however as a precaution for if it’s not unlocked I think it’s a good idea to implement Drop. If not to unlock then to panic and say it needs to be unlocked explicitly, it hasn’t been done already.

tooling/nargo_cli/src/cli/check_cmd.rs
aakoshh: Sorry, I meant that whatever writes artifacts would use lock_exclusively, while the other commands that just read artifacts use lock_shared. That way they can’t be mutated while they are used by read-only commands

tooling/nargo_cli/src/lock.rs
asterite: Oh, you mean, also implement Drop. Sure, I'll do it! I wasn't exactly sure what's the semantic of Drop, if it could call anything.

asterite: Or, I guess, if one calls `unlock` and the `drop` comes in, will it unlock it twice? I guess it's not issue, I just wasn't sure about this. And also, since now it happens in a single place in the code, maybe we don't even need the `Lock` struct, I'll check...

aakoshh: Yes I suppose it would unlock twice unless you prevent it, eg with a flag, or using `ManuallyDrop` like [here](https://github.com/aakoshh/async-session-types-rs/blob/6fc5dff2d73a9c20932584fad24a553a3c3213e9/src/lib.rs#L113) 

I think Drop can do anything, but it’s difficult to do async stuff there.

tooling/nargo_cli/src/cli/check_cmd.rs
asterite: Oooh... I see! Makes sense, I'll make this change.

asterite: I just realized some commands don't need a lock at all, like `new` and `init`. Then LSP runs in the background... I don't know if it would be bad for that to lock things because then you couldn't run some commands on the command line.

Given that this PR was about solving the issue of running multiple `execute` commands in parallel in CI without issues, maybe only exclusively locking in a few commands and just that is good enough for now.

aakoshh (approved):
Like I said I'd consider using `lock_shared` for any that read `Nargo.toml`, but otherwise LGTM!

PR #6847, opened 2024-12-17
chore(ci): Add basic non determinism check

vezenovm:
# Description

## Problem\*

Resolves #2036

## Summary\*

Still a draft as I want to see all the failures that we get. 

## Documentation\*

Check one:
- [X] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [X] I have tested the changes locally.
- [X] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #6945, opened 2025-01-04
fix: Non-determinism from under constrained checks

vezenovm:
# Description

## Problem\*

In https://github.com/noir-lang/noir/pull/6847 the `--check-non-determinism` flag was only run when force compiling to Brillig. Not sure why I put the flag here in #6847, it looks to be a mistake. Running the non-determinism check for normal ACIR compilation revealed that the under constrained value checks were generating their warnings in a non-deterministic manner.  

## Summary\*

The under constrained value checks iterate over a HashSet in many locations. I switched the sets which are iterated over to all be a BTreeSet rather than a HashSet.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/nargo_cli/build.rs
+++ b/tooling/nargo_cli/build.rs
@@ -208,6 +208,7 @@ fn test_{test_name}(force_brillig: ForceBrillig, inliner_aggressiveness: Inliner
     nargo.arg("--program-dir").arg(test_program_dir);
     nargo.arg("{test_command}").arg("--force");
     nargo.arg("--inliner-aggressiveness").arg(inliner_aggressiveness.0.to_string());
+    nargo.arg("--check-non-determinism");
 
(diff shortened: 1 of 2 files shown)


TomAFrench (approved):

tooling/nargo_cli/build.rs
```diff
@@ -217,7 +218,6 @@ fn test_{test_name}(force_brillig: ForceBrillig, inliner_aggressiveness: Inliner
         nargo.arg("50");
 
         // Check whether the test case is non-deterministic
```
aakoshh, line 220: Comment did not move 👆

vezenovm: 🤦 updated here https://github.com/noir-lang/noir/pull/6959

PR #6947, opened 2025-01-06
feat: acir_formal_proofs

defkit:
# Description
The ACIR formal verification. Combines a test generator in the Noir repository with a formal verifier in Barretenberg to mathematically prove the correctness of ACIR instructions using SMT solving. Verifies range of operations including 127-bit arithmetic (addition, subtraction, multiplication), 126-bit division, bitwise operations (though currently limited to 32-bit for AND/OR/XOR), shift operations, field operations (ADD, MUL, DIV), and comparison operations

## Additional Context
[aztec-packages PR](https://github.com/AztecProtocol/aztec-packages/pull/10973)

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/ssa_test/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "ssa_test"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
```
Rumata888, line 6: Please look at the root Cargo.toml and the ones from other packages and do them in the same style

compiler/noirc_evaluator/src/ssa/ir/map.rs
```diff
@@ -28,7 +28,7 @@ impl<T> Id<T> {
     ///
     /// This is private so that we can guarantee ids created from this function
     /// point to valid T values in their external maps.
-    fn new(index: u32) -> Self {
+    pub(crate) fn new(index: u32) -> Self {
```
Rumata888, line 31: Why?

compiler/noirc_evaluator/src/acir_instruction_builder.rs
```diff
@@ -0,0 +1,287 @@
...
+    let v0 = builder.add_parameter(variable_type);
+    builder.insert_range_check(v0, 64, Some("Range Check failed".to_string()));
+    builder.terminate_with_return(vec![v0]);
+
+    return builder.finish()
+}
+
+/// Creates an SSA function for truncate operations
+fn truncate_function(variable_type: Type) -> Ssa {
+    // truncate v0: field 10, 20?..
```
Rumata888, line 265: Why the question mark?

Rumata888 (changes requested):
Please change how you import dependencies (Cargo.toml) and remove .gitignore

compiler/noirc_evaluator/src/ssa/ir/map.rs
jewelofchaos9: It was changed in #6807 from `pub(crate)` to private. I can not use `test_new` instead, because of `#[cfg(test)]` attribute. I need `Id` for function building.

jewelofchaos9: > Why?

It was changed in #6807 from `pub(crate)` to private. I can not use `test_new` instead, because of `#[cfg(test)]` attribute. I need `Id` for function building.

compiler/noirc_evaluator/ssa_test/src/main.rs
```diff
@@ -0,0 +1,111 @@
...
+    let mut file = File::create(path)?;
+    file.write_all(data)?;
+    Ok(())
+}
+
+/// Saves instruction artifacts to files in the artifacts directory
+/// Prints the formatted SSA for each artifact and saves the decompressed ACIR
+fn save_artifacts(all_artifacts: Vec<InstructionArtifacts>, dir: &str) {
+    for artifacts in all_artifacts.iter() {
+        println!("{}", artifacts.formatted_ssa);
```
aakoshh, line 38: Perhaps you can print a header of the artefact between the SSA to make it easier to recognise which one it is?

compiler/noirc_evaluator/ssa_test/src/main.rs
```diff
@@ -0,0 +1,111 @@
...
+#[command(author, version, about, long_about = None)]
+struct Args {
+    #[arg(short, long, default_value = "../../../../../barretenberg/cpp/src/barretenberg/acir_formal_proofs/artifacts/")]
+    dir: String,
+}
+
+/// Decompresses gzipped data into a byte vector
+fn ungzip(compressed_data: Vec<u8>) -> Vec<u8> {
+    let mut decompressed_data: Vec<u8> = Vec::new();
+    let mut decoder = GzDecoder::new(&compressed_data[..]);
```
aakoshh, line 21: I looks like `unzip` could take a reference, and then the `acir.clone()` in `save_artifacts` could be just `&acir`.

compiler/noirc_evaluator/ssa_test/src/main.rs
```diff
@@ -0,0 +1,111 @@
...
+/// Saves instruction artifacts to files in the artifacts directory
+/// Prints the formatted SSA for each artifact and saves the decompressed ACIR
+fn save_artifacts(all_artifacts: Vec<InstructionArtifacts>, dir: &str) {
+    for artifacts in all_artifacts.iter() {
+        println!("{}", artifacts.formatted_ssa);
+        let filename = format!("{}{}{}", dir, artifacts.instruction_name, ".acir");
+        let acir = &artifacts.serialized_acir;
+        match save_to_file(&ungzip(acir.clone()), &filename) {
+            Ok(_) => (),
+            Err(error) => println!("Error saving data: {}", error),
```
aakoshh, line 43: Does it make sense to carry on with the other artefacts? 
Perhaps you could exit the process with a non-zero error code to show that it failed, just in case this needs to be automated.
Also, errors can be printed with `eprintln!` to go to `stderr`.

compiler/noirc_evaluator/src/acir_instruction_builder.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    /// SSA representation formatted as "acir(inline) {...}"
+    pub formatted_ssa: String,
+
+    /// JSON serialized SSA
+    pub serialized_ssa: String,
+
+    /// Gzipped bytes of ACIR program
+    pub serialized_acir: Vec<u8>,
+}
```
aakoshh, line 38: This looks quite specific to the use case you have. Could it live in the new crate you created?

compiler/noirc_evaluator/ssa_test/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "ssa_test"
```
aakoshh, line 2: This sounds like it will test the SSA, which we do elsewhere. Maybe `ssa_verification` would be more suggestive of a broader scope?

compiler/noirc_evaluator/src/acir_instruction_builder.rs
jewelofchaos9: > This looks quite specific to the use case you have. Could it live in the new crate you created?

No, `InstructionArtifacts` needs to stay in noirc_evaluator since its implementation uses internal methods and structs of this crate.  I don't know how to move this struct to the new crate without breaking the current architecture, if I understand you correctly

aakoshh: Right, I see even `FunctionBuilder` is only `pub(crate)`. Maybe there is a case for making that public, as a more general purpose SSA related utility, than making `InstructionArtifacts` part of the public interface, given its limited scope. I could be wrong, just seeing different serialised SSA formats in it seems overly specific to how you are generating the inputs for the verification, not something anyone else will use.

jewelofchaos9: I can try to move the entire `ssa_verification` crate  into the `noirc_evaluator` crate -- this way we wouldn't need to expose anything.  However, I'm not sure if this is the right approach since `ssa_verification` is very specific.

aakoshh: I agree, I don’t think that’s the right approach either. Maybe @TomAFrench can weigh in on how important it is to keep things private to the crate?

compiler/noirc_evaluator/ssa_verification/src/main.rs
```diff
@@ -0,0 +1,132 @@
...
+- Bitvector operations (up to 127 bits): add, sub, mul, mod, xor, and, div, eq, lt, not
+- Shift operations (32 and 64 bits): shl, shr  
+- Binary operations (32-bit): xor, and, or
+- Field operations: add, mul, div
+- Signed integer operations: div (126-bit)
+
+Each test case generates formatted SSA representation and serialized ACIR output.
+
+FLAGS:
+    -d, --dir <PATH>    Output directory for test artifacts [default: ../../../../../barretenberg/cpp/src/barretenberg/acir_formal_proofs/artifacts/]"
```
aakoshh, line 27: I think `clap` creates these on its own.

compiler/noirc_evaluator/src/acir_instruction_builder.rs
rkarabut: @aakoshh after looking this over, I wonder if it would be better to expose the SSA parser we use in tests and not FunctionBuilder :thinking: To me it looks like it makes more sense as a public (debugging etc) interface

aakoshh: I agree with exposing the parser, it's much easier to get started with if we have an SSA to start off from. I suppose the benefit of the `FunctionBuilder` is compile time warnings about any breaking change, but the parser is pretty good at pointing out what's wrong with a snippet.

Issue #6842, opened 2024-12-17
Consider adding a no-op instruction to SSA


jfecher:
### Problem

Certain passes like `resolve_is_unconstrained` go through the ssa to handle only a few instructions (in this case only `vN = call is_unconstrained()`) but don't have a good way to remove these instructions without re-inserting every instruction into the block as they go (and choosing not to re-insert these handled instructions) or iterating through the block again afterward and retaining only the non-handled instructions.

### Happy Case

We can consider adding an `Instruction::Nop { result_types }` instruction which conceptually "returns" value(s) of any type but in practice will always be removed from the program or skipped over in acir-gen.

This Nop instruction can be used to replace instructions like `call is_unconstrained()` with so that they will later automatically be removed by calling `Instruction::Nop { ... }.simplify()` which will always return `SimplifyResult::Remove`.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

aakoshh:
> Certain passes ... don't have a good way to remove these instructions without re-inserting every instruction ... or ... retaining only the non-handled instructions.

Just looked at `resolve_is_unconstrainted` at it's doing neither re-insertion of all instructions nor retaining them per-se, rather, it [disconnects](https://github.com/noir-lang/noir/blob/fd816cf1e747ef036be84771a6f3fe0a651d80a6/compiler/noirc_evaluator/src/ssa/opt/resolve_is_unconstrained.rs#L42-L52) the results of the call from their original return value and assumes the DIE pass will remove the now-unused call. Does that imply that in this case there _is_ a way, but it's not a _good_ way?

To make sure I understand correctly, what you're suggesting is not only to replace the original result with a constant, but to replace the call instruction itself with a different `Nop` instruction in-place, to avoid having to shift the instruction vector in the block, and then, similar to how `resolve_is_unconstrained` relies on DIE removing unused instructions, that there will be some other pass that _will_ actually re-insert all instructions, while calling `simplify`. 

IIUC the DIE pass would not eliminate `Nop` because it still has return values, unless those are similarly disconnected from the originals. If I just return `None` from `simplify`, wouldn't that leave the original result IDs unassigned and somehow invalid?

TomAFrench:
The general idea is this no-op instruction is a way that we can tell compiler that we don't care about the results. The component can then whenever it says reinserting instructions ignore this instruction and so we automatically filter them out without needing a whole pass devoted to that. 

[...]

PR #6948, opened 2025-01-06
feat: don't simplify SSA instructions when creating them from a string

### LINKED ISSUE #6822: Separate instruction simplification from insertion into DFG
We currently don't have a particularly good way to create exact SSA programs as we automatically perform any available simplifications as we push instructions into the function. This has been flagged as an issue in https://github.com/noir-lang/noir/pull/6434 and has been something that I've definitely run into a couple of times.

It would then be nice to be able to add a flag to the `FunctionBuilder` so that we can get it to skip any instruction simplification when constructing a function which would allow us to construct test programs more easily.

asterite:
# Description

## Problem

Resolves #6822

## Summary

In a couple of tests simplification is still done because those tests are testing the simplification.

In one test, though, a new instruction popped up. I noticed that `normalize_ids()` would previously simplify instructions so we would miss if a pass didn't correctly simplify things... so I think the flattening pass might be leaving some extra `enable_side_effects` instructions, though I don't know if later those are simplified anyway.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/instruction/call.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/instruction/call.rs
@@ -730,7 +730,7 @@ mod tests {
                 return v2
             }
             "#;
-        let ssa = Ssa::from_str(src).unwrap();
+        let ssa = Ssa::from_str_simplifying(src).unwrap();
 
         let expected = r#"
             brillig(inline) fn main f0 {
--- a/compiler/noirc_evaluator/src/ssa/function_builder/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/function_builder/mod.rs
@@ -40,6 +40,10 @@ pub(crate) struct FunctionBuilder {
     finished_functions: Vec<Function>,
     call_stack: CallStackId,
     error_types: BTreeMap<ErrorSelector, HirType>,
+
+    /// Whether instructions are simplified as soon as they are inserted into this builder.
+    /// This is true by default unless changed to false after constructing a builder.
+    pub(crate) simplify: bool,
 }
 
 impl FunctionBuilder {
@@ -56,6 +60,7 @@ impl FunctionBuilder {
             finished_functions: Vec::new(),
             call_stack: CallStackId::root(),
             error_types: BTreeMap::default(),
+            simplify: true,
         }
     }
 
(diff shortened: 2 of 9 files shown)


TomAFrench:
> think the flattening pass might be leaving some extra enable_side_effects instructions, though I don't know if later those are simplified anyway.

Yep, we have another SSA pass which will clear these up: https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs

compiler/noirc_evaluator/src/ssa/ir/instruction/call.rs
```diff
@@ -730,7 +730,7 @@ mod tests {
                 return v2
             }
             "#;
-        let ssa = Ssa::from_str(src).unwrap();
+        let ssa = Ssa::from_str_simplifying(src).unwrap();
```
TomAFrench, line 733: Why do we simplify in these situations?

aakoshh: I think it's just this particular test that is actually about the correctness of the simplification.

aakoshh: I see there is another test below where this is less obvious (`simplify_derive_generators_has_correct_type` vs `constraint_decomposition`). 

Just wondering if it would make sense to add a method to the `Ssa` which re-inserts all instructions with simplification? As I understand some mechanism actually assume that this will happen sooner or later, like what we discussed in https://github.com/noir-lang/noir/issues/6842 Having an explicit way to trigger this would perhaps not be out of place, and then a test that needs it can parse the SSA without simplification and then call this method. Then the "simple" act of parsing a piece of SSA doesn't need to be burdened with the knowledge of further effects we want to see happen.

TomAFrench:
Merging master to trigger an update to the reports as I've made changes in aztec-packages which should greatly improve these.

compiler/noirc_evaluator/src/ssa/ir/instruction/call.rs
asterite: I don't know if this would be what the "old" normalize pass would do. So maybe this could be done by having the normalize pass receive a boolean flag. We could try it in a follow-up PR.

TomAFrench: Yeah, it's essentially bringing back that old behaviour.

I'd argue that if we're only using this in a couple of tests then having the "special behaviour" only in test code is a good thing (and so we shouldn't make that follow-up). We could quite possibly refactor these tests to not rely on this anyway.

TomAFrench (approved):

PR #6951, opened 2025-01-06
chore: simplify boolean in a mul of a mul

### LINKED ISSUE #6879: Fix regression caused by #6875
#6875 caused a small regression in some of our circuits where we started to multiply a predicate into a value multiple times unnecessarily.

We should nail down how this regression is coming in (notably the optimization we added prevents the automatic value merging of if-else instructions for numeric values if the new optimization fires) and address it.

guipublic:
# Description

## Problem\*

Resolves #6879 

## Summary\*
When multiplying a boolean with the result of another multiplication involving the same boolean, we can discard one multiplication.

## Additional Context
The regression is a bit subtle, the optimisation introduced by PR #6875 was simplifying an ifesle instruction if one of its operand is also an ifelse with the same condition.
When the operands are an array and the array with one element modified, the merge of values only happens on the modified element, which induces only one multiplication.
However, when the array is a result of of an ifelse, the optimisation from #6875 replace it with another array (the operand of the inner ifselse). So we are no longer merging two almost identical arrays, and we get one multiplication per array element instead.
So basically, the optimisation is preventing another optimisation.
I could not find a way to get the two optimisation to be effective at the same time. The value merge is done when we eliminate IfElse, so the #6875 must be done before. But then the link between the arrays is not obvious.
Fortunately, since this happens when we have inner If statements with the same condition, that means we are multiplying several times with the condition, and this can be optimised from the instruction simplification pass.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
This PR drops the number of opcodes in `rollup-base-private` from 617798 (ab8807d27b0442bd7e9ff73050f44bacea3add16) to 215613. Not bad!

TomAFrench (approved):
LGTM!

PR #6969, opened 2025-01-07
fix: don't fail parsing macro if there are parser warnings

### LINKED ISSUE #6968: Missing safety comments in code being unquoted into top-level item results in failure to parse
This was first noted by @kashbrti.

The snippet below is expected to insert a test function into the program which consists of an empty `unsafe` block.

```nr
#[make_test]
pub mod foo {}

comptime fn make_test(_m: Module) -> Quoted {
    quote {
#[test]
fn bar() {
    unsafe { }
}
}
}
```

Running nargo test however produces the below output (and no test is inserted as shown by the final lines)

```
$ nargo test
warning: Failed to parse macro's token stream into top-level item
  ┌─ src/main.nr:1:1
  │
1 │ #[make_test]
  │ ------------
  │ │
  │ Unsafe block must start with a safety comment: //@safety
  │ Missing Safety Comment
  │
  = The resulting token stream was: (stream starts on next line)
      #[test]fn bar() {
        unsafe {
            
        }
    }
  = To avoid this error in the future, try adding input validation to your macro. Erroring out early with an `assert` can be a good way to provide a user-friendly error message

[tmep] Running 0 test functions
[tmep] 0 tests passed

```

We should find out why this warning results in the macro failing to insert this function and fix it. We should also throw a hard error if we reach a situation where we're not properly unquoting values like this as opposed to failing silently.

asterite:
# Description

## Problem

Resolves #6968

## Summary

The issue was that parser warnings made macro parsing failed. Instead, in this PR we report those warnings and move on (only if all parser errors were warnings).

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/comptime.rs
+++ b/compiler/noirc_frontend/src/elaborator/comptime.rs
@@ -247,7 +247,7 @@ impl<'context> Elaborator<'context> {
 
         if value != Value::Unit {
             let items = value
-                .into_top_level_items(location, self.interner)
+                .into_top_level_items(location, self)
                 .map_err(|error| error.into_compilation_error_pair())?;
 
             self.add_items(items, generated_items, location);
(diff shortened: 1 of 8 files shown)


TomAFrench (approved):

kashbrti:
solves the issue! Thanks!

Issue #6929, opened 2025-01-02
Bytecode size blowup not completely resolved


benesjan:
A while ago I stumbled upon a bytecode size issues when re-constraining encryption in a token contract. This is the original post on slack:

>After re-constraining encryption token contract artifact size blows up from 4.8MB to 125MB. When I look inside the artifact the culprit lies in bytecode and debug_symbols sizes of the changed functions. (e.g. one of those functions is `transfer_to_public` where bytecode blows up to more than 20 MB and debug_symbols to more than 9 MB).

After rebasing [that PR ](https://github.com/AztecProtocol/aztec-packages/pull/10535) on master there was a significant improvement and the bytecode size dropped to 50MB (`noir-projects/noir-contracts/target/token_contract-Token.json` file).

This however still causes some of the tests to fail (when fetching the bytecode from aztec node).

it seems to me that the bytecode is still too big and the bytecode size blowup issue have not been completely resolved.

### Nargo Version

nargo version = 1.0.0-beta.0 noirc version = 1.0.0-beta.0+644ac8064bf63de3 (git version hash: 644ac8064bf63de3, is dirty: false)

aakoshh:
For reference this method was also the one responsible for the increase investigated in https://github.com/noir-lang/noir/issues/6786

PR #6972, opened 2025-01-07
fix: Reproduce and fix bytecode blowup

aakoshh:
# Description

## Problem\*

Fixes https://github.com/noir-lang/noir/issues/6929 

## Summary\*

Reproduces the bytecode blowup when switching from unconstrained to constrained encrypted note emission in https://github.com/AztecProtocol/aztec-packages/pull/10535

Fixes the issue by doing another mem2reg pass before flattening the CFG.

## Additional Context

See the flame graph in the comments below for what we're trying to reproduce. It looks like [this line](https://github.com/AztecProtocol/aztec-packages/blob/b42756bc10175fea9eb60544759e9dbe41ae5e76/noir-projects/aztec-nr/aztec/src/encrypted_logs/payload.nr#L149) results in lots of opcodes/gates coming from `AssertZero`, `MemOp` and `Range` instructions.

NB the size blowup goes away if we use `let size = M - 32 - HEADER_SIZE - OVERHEAD_PADDING;` instead of `let size = M - offset;`, but that's in `aztec-packages`, it doesn't work here. We created https://github.com/noir-lang/vscode-noir/issues/92 to capture that. 

Examining the original SSA (see the comments in this PR 👇) revealed that the source of RANGE checks was a bug with unsigned multiplications, which was fixed by https://github.com/noir-lang/noir/pull/6983 , but it doesn't make the array-copy phenomenon go away. 

The item-by-item array copying is related to how `Remove IfElse` pass handles conditional variables and arrays. The bytecode size in this example depends on where we place the `flag` variable (not part of the original program). 

### Testing

To see gate counts:
```
cargo run -p nargo_cli -- --program-dir . info --skip-underconstrained-check --skip-brillig-constraints-check
```

To see the final SSA:
```
cargo run -p nargo_cli -- --program-dir . info --skip-underconstrained-check --skip-brillig-constraints-check --show-ssa --show-ssa-pass-name "Dead Instruction Elimination (2nd)"
```

### The fix

Looking at the SSA with smaller numbers (an array of size 15 instead of 527 items) just before the `Remove IfElse` looked like this:

<details>
  <summary>After Inlining (2nd)</summary>

```
After Inlining (2nd):
acir(inline) fn main f0 {
  b0(v0: [u8; 3], v1: [u8; 2], v2: [u8; 9], v3: u1):
    v5 = make_array [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v6 = allocate -> &mut [u8; 15]
    v7 = allocate -> &mut u32
    enable_side_effects v3
    v9 = array_get v0, index u32 0 -> u8
    v10 = make_array [v9, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v12 = array_get v0, index u32 1 -> u8
    v13 = make_array [v9, v12, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v15 = array_get v0, index u32 2 -> u8
    v16 = make_array [v9, v12, v15, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v17 = array_get v1, index u32 0 -> u8
    v18 = make_array [v9, v12, v15, v17, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v19 = array_get v1, index u32 1 -> u8
    v20 = make_array [v9, v12, v15, v17, v19, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v21 = not v3
    v22 = if v3 then v20 else (if v21) v5
    v23 = cast v3 as u32
    v24 = cast v21 as u32
    v26 = mul v23, u32 6
    v27 = array_get v2, index u32 0 -> u8
    v28 = array_set v22, index v26, value v27
    v29 = add v26, u32 1
    v30 = if v3 then v28 else (if v21) v5
    v31 = add v26, u32 1
    v32 = array_get v2, index u32 1 -> u8
    v33 = array_set v30, index v31, value v32
    v34 = add v31, u32 1
    v35 = if v3 then v33 else (if v21) v5
    v36 = add v26, u32 2
    v37 = array_get v2, index u32 2 -> u8
    v38 = array_set v35, index v36, value v37
    v39 = add v36, u32 1
    v40 = if v3 then v38 else (if v21) v5
    v42 = add v26, u32 3
    v43 = array_get v2, index u32 3 -> u8
    v44 = array_set v40, index v42, value v43
    v45 = add v42, u32 1
    v46 = if v3 then v44 else (if v21) v5
    v48 = add v26, u32 4
    v49 = array_get v2, index u32 4 -> u8
    v50 = array_set v46, index v48, value v49
    v51 = add v48, u32 1
    v52 = if v3 then v50 else (if v21) v5
    v54 = add v26, u32 5
    v55 = array_get v2, index u32 5 -> u8
    v56 = array_set v52, index v54, value v55
    v57 = add v54, u32 1
    v58 = if v3 then v56 else (if v21) v5
    v59 = add v26, u32 6
    v60 = array_get v2, index u32 6 -> u8
    v61 = array_set v58, index v59, value v60
    v62 = add v59, u32 1
    v63 = if v3 then v61 else (if v21) v5
    v65 = add v26, u32 7
    v66 = array_get v2, index u32 7 -> u8
    v67 = array_set v63, index v65, value v66
    v68 = add v65, u32 1
    v69 = if v3 then v67 else (if v21) v5
    v71 = add v26, u32 8
    v72 = array_get v2, index u32 8 -> u8
    v73 = array_set v69, index v71, value v72
    v74 = add v71, u32 1
    v75 = if v3 then v73 else (if v21) v5
    enable_side_effects u1 1
    return v75
}
```
</details>

What we see there is that for each field, we make a copy of the array, then either use it or discard it and keep the initial value `v5` depending on the conditional variable `v3`. After this, in `Remove IfElse`, the `ValueMerger` would take all the instructions like `v75 = if v3 then v73 else (if v21) v5` and merge the arrays item-by-item, multiplying each item with the conditional, before making a new array out of them. This results in a quadratic blowup of instructions.

Instead what we'd ideally do is modify a copy of the array with each item, and then at the end have a _single_ if-then-else to merge it with `v5`, depending on `v3`.

Now take a look at the step before `Flattening`:

<details>
  <summary>After Simplifying (2nd)</summary>

```
After Simplifying (2nd):
acir(inline) fn main f0 {
  b0(v0: [u8; 3], v1: [u8; 2], v2: [u8; 9], v3: u1):
    v5 = make_array [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v6 = allocate -> &mut [u8; 15]
    store v5 at v6
    v7 = allocate -> &mut u32
    store u32 0 at v7
    jmpif v3 then: b1, else: b2
  b1():
    v9 = array_get v0, index u32 0 -> u8
    v10 = make_array [v9, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v12 = array_get v0, index u32 1 -> u8
    v13 = make_array [v9, v12, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v15 = array_get v0, index u32 2 -> u8
    v16 = make_array [v9, v12, v15, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v17 = array_get v1, index u32 0 -> u8
    v18 = make_array [v9, v12, v15, v17, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v19 = array_get v1, index u32 1 -> u8
    v20 = make_array [v9, v12, v15, v17, v19, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    store v20 at v6
    store u32 6 at v7
    v22 = load v6 -> [u8; 15]
    v23 = load v7 -> u32
    v24 = array_get v2, index u32 0 -> u8
    v25 = array_set v22, index v23, value v24
    v26 = add v23, u32 1
    store v25 at v6
    v27 = load v6 -> [u8; 15]
    v28 = load v7 -> u32
    v29 = add v28, u32 1
    v30 = array_get v2, index u32 1 -> u8
    v31 = array_set v27, index v29, value v30
    v32 = add v29, u32 1
    store v31 at v6
    v33 = load v6 -> [u8; 15]
    v34 = load v7 -> u32
    v35 = add v34, u32 2
    v36 = array_get v2, index u32 2 -> u8
    v37 = array_set v33, index v35, value v36
    v38 = add v35, u32 1
    store v37 at v6
    v39 = load v6 -> [u8; 15]
    v40 = load v7 -> u32
    v42 = add v40, u32 3
    v43 = array_get v2, index u32 3 -> u8
    v44 = array_set v39, index v42, value v43
    v45 = add v42, u32 1
    store v44 at v6
    v46 = load v6 -> [u8; 15]
    v47 = load v7 -> u32
    v49 = add v47, u32 4
    v50 = array_get v2, index u32 4 -> u8
    v51 = array_set v46, index v49, value v50
    v52 = add v49, u32 1
    store v51 at v6
    v53 = load v6 -> [u8; 15]
    v54 = load v7 -> u32
    v56 = add v54, u32 5
    v57 = array_get v2, index u32 5 -> u8
    v58 = array_set v53, index v56, value v57
    v59 = add v56, u32 1
    store v58 at v6
    v60 = load v6 -> [u8; 15]
    v61 = load v7 -> u32
    v62 = add v61, u32 6
    v63 = array_get v2, index u32 6 -> u8
    v64 = array_set v60, index v62, value v63
    v65 = add v62, u32 1
    store v64 at v6
    v66 = load v6 -> [u8; 15]
    v67 = load v7 -> u32
    v69 = add v67, u32 7
    v70 = array_get v2, index u32 7 -> u8
    v71 = array_set v66, index v69, value v70
    v72 = add v69, u32 1
    store v71 at v6
    v73 = load v6 -> [u8; 15]
    v74 = load v7 -> u32
    v76 = add v74, u32 8
    v77 = array_get v2, index u32 8 -> u8
    v78 = array_set v73, index v76, value v77
    v79 = add v76, u32 1
    store v78 at v6
    jmp b2()
  b2():
    v80 = load v6 -> [u8; 15]
    return v80
}
```
</details>

@TomAFrench 's insight is that to achieve what we outlined above (the single merge at the end), we just have to get rid of the redundant store and load instructions:

> If we know at compile-time that none of these array_sets are going to be out of bounds then we should do as you suggest.
> the issue is that we're not propagating the fact that offset is a known constant (and so all the write indices are known) until after we've flattened them and multiplied in the predicate.
> in b1 we're doing a lot of storing and then immediately loading from that location again which we should be able to remove.

Adding another mem2reg before flattening indeed had the desired effect 🎉 

<details>
  <summary>After Flattening</summary>

```
After Flattening:
acir(inline) fn main f0 {
  b0(v0: [u8; 3], v1: [u8; 2], v2: [u8; 9], v3: u1):
    v5 = make_array [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v6 = allocate -> &mut [u8; 15]
    store v5 at v6
    v7 = allocate -> &mut u32
    enable_side_effects v3
    v9 = array_get v0, index u32 0 -> u8
    v10 = make_array [v9, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v12 = array_get v0, index u32 1 -> u8
    v13 = make_array [v9, v12, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v15 = array_get v0, index u32 2 -> u8
    v16 = make_array [v9, v12, v15, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v17 = array_get v1, index u32 0 -> u8
    v18 = make_array [v9, v12, v15, v17, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v19 = array_get v1, index u32 1 -> u8
    v20 = make_array [v9, v12, v15, v17, v19, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v21 = array_get v2, index u32 0 -> u8
    v22 = make_array [v9, v12, v15, v17, v19, u8 0, v21, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v23 = array_get v2, index u32 1 -> u8
    v24 = make_array [v9, v12, v15, v17, v19, u8 0, v21, v23, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v25 = array_get v2, index u32 2 -> u8
    v26 = make_array [v9, v12, v15, v17, v19, u8 0, v21, v23, v25, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v28 = array_get v2, index u32 3 -> u8
    v29 = make_array [v9, v12, v15, v17, v19, u8 0, v21, v23, v25, v28, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v31 = array_get v2, index u32 4 -> u8
    v32 = make_array [v9, v12, v15, v17, v19, u8 0, v21, v23, v25, v28, v31, u8 0, u8 0, u8 0, u8 0] : [u8; 15]
    v34 = array_get v2, index u32 5 -> u8
    v35 = make_array [v9, v12, v15, v17, v19, u8 0, v21, v23, v25, v28, v31, v34, u8 0, u8 0, u8 0] : [u8; 15]
    v37 = array_get v2, index u32 6 -> u8
    v38 = make_array [v9, v12, v15, v17, v19, u8 0, v21, v23, v25, v28, v31, v34, v37, u8 0, u8 0] : [u8; 15]
    v40 = array_get v2, index u32 7 -> u8
    v41 = make_array [v9, v12, v15, v17, v19, u8 0, v21, v23, v25, v28, v31, v34, v37, v40, u8 0] : [u8; 15]
    v43 = array_get v2, index u32 8 -> u8
    v44 = make_array [v9, v12, v15, v17, v19, u8 0, v21, v23, v25, v28, v31, v34, v37, v40, v43] : [u8; 15]
    v45 = load v6 -> [u8; 15]
    v46 = not v3
    v47 = if v3 then v44 else (if v46) v45
    store v47 at v6
    enable_side_effects u1 1
    v49 = load v6 -> [u8; 15]
    return v49
}

```
</details>

## Documentation\*

Check one:

# PR Checklist\*


aakoshh:
Here's a flame graph of the problematic contract in `aztec-packages`:

main gates bug

To reproduce:
```bash
PROGRAM_DIR=$PWD/noir-projects/noir-contracts                                                                                                                                                                  
cd noir/noir-repo  
git checkout origin/12-09-chore_removing_noir_bug_workaround
export GIT_DIRTY=true
export GIT_COMMIT=$(git rev-parse HEAD) 

cargo run --release -p nargo_cli -- --program-dir $PROGRAM_DIR compile --package token_contract     
ls -lh $PROGRAM_DIR/target/token_contract-Token.json 
jq '.functions | sort_by(.name) | .[] | {name: .name, size: .bytecode | length }' $PROGRAM_DIR/target/token_contract-Token.json

$PROGRAM_DIR/scripts/flamegraph.sh Token transfer_to_public
```
The SVG is at `noir-projects/noir-contracts/dest/main::gates.svg`

aakoshh:
Here's the baseline flame graph for reference:

main gates master

test_programs/compile_success_no_bug/encrypted_log_regression/src/main.nr
```diff
@@ -0,0 +1,116 @@
...
+
+    // NOTE: This made the bytecode size blowup disappear in aztec packages,
+    // but in this reproduction the size seems to be statically known regardless.
+    //let size = M - 32 - HEADER_SIZE - OVERHEAD_PADDING;
+
+    assert_eq(size, incoming_body_ciphertext.len(), "ciphertext length mismatch");
+    for i in 0..size {
+        // NOTE: Adding `shift` makes the index dynamic enough to force ACIR to use memory operations.
+        // Without `shift` it assigns directly to witnesses, which is good, but not what we wanted to reproduce.
+        encrypted_bytes[offset + i + shift] = incoming_body_ciphertext[i];
```
vezenovm, line 97: I have a feeling we are seeing a blow-up here because every `array_set` from this unrolled loop is performing a copy. If we see `v3 = array_set v1 index v2 value incoming_body_ciphertext[i]`, we will copy `v1` so that we do not mutate it directly.

However, in this case, we know that the dynamic value `shift` is not changed between `array_set` instructions and in fact `i` is different for each one, thus we should be able to safely mutate the array and avoid copying. Could you post the final SSA before ACIR gen? You can probably get rid of all the code preceding this loop to shorten the SSA.

vezenovm: > With the shift in place the SSA ends like this

Are there `array_set` instructions in the instructions you snipped between v178 and v1944

vezenovm: Hmm it does look like we are correctly marking the array sets mutable here.

aakoshh:
I am looking at the original SSA now, printed by this command:
```shell
cargo run -p nargo_cli -- --program-dir $PROGRAM_DIR compile --package token_contract --show-ssa-pass-name "Array Set Optimizations" > compile-output.txt
```
It's 2 million lines, but looking for the first occurrence of "ciphertext length mismatch" reveals a pattern like this:
```
    v28334 = add v28332, v28333
    v28335 = array_set mut v28328, index u32 0, value v28334
    inc_rc v28335
    enable_side_effects v22441
    v28336 = array_get v28335, index u32 1 -> u8
    ... get all indices of the array ...
    v28366 = array_get v28335, index u32 31 -> u8
    v28367 = cast v22441 as u8
    v28368 = mul v28367, v28334
    ... multiply all variables created above with the side effect variable ...
    v28447 = mul v28367, v27047
    v28448 = make_array [v28368, v28369, ..., v28447, u8 0, u8 0, ..., u8 0] : [u8; 527]
    v28449 = mul v22495, u32 95
    v28451 = mul u32 432, v22495
    v28452 = mul v27866, v22495
    constrain v28451 == v28452, "ciphertext length mismatch"
    v28453 = lt u32 0, v27866
    v28454 = mul v28453, v22441
    constrain v28454 == v22441, "Index out of bounds"
    v28455 = array_get v27867, index u32 0 -> u8
    v28456 = array_set mut v28448, index v28449, value v28455
    v28457 = add v28449, u32 1
    v28458 = array_get v28456, index Field 0 -> u8
    v28459 = mul v28367, v28458
    v28460 = array_get v28456, index Field 1 -> u8
    v28461 = mul v28367, v28460
    ... get all fields of the array and multiply the value with the side effect variable
    v29955 = array_get v28456, index Field 526 -> u8
    v29956 = mul v28367, v29955
    v29957 = make_array [v28459, v28461, ..., v29953, v29956] : [u8; 527]
    v29958 = lt u32 1, v27866
    v29959 = mul v29958, v22441
    constrain v29959 == v22441, "Index out of bounds"
```
I suspect that explain all the RANGE checks, coming from [check_unsigned_overflow](https://github.com/AztecProtocol/aztec-packages/blob/70cf870f98ab486d8803b1cdd6a795419a57dffd/noir/noir-repo/compiler/noirc_evaluator/src/acir/mod.rs#L1984) after inserting the [multiplication](https://github.com/AztecProtocol/aztec-packages/blob/70cf870f98ab486d8803b1cdd6a795419a57dffd/noir/noir-repo/compiler/noirc_evaluator/src/acir/mod.rs#L1951).

TomAFrench:
`check_unsigned_overflow` should not insert range checks for these multiplications due to the fact the condition is an upcasted boolean so it can tell that the multiplication won't overflow.

aakoshh:
@TomAFrench that's what I thought, on the other hand isn't it a `u8` multiplied by a `u32`? How does it know it can't be `u8::MAX` which could overflow a `u32::MAX`, at the point where it just sees the types? Can you link to the code?

I'm looking at the ACIR and see this pattern repeating:
```
INIT (id: 399, len: 527) 
EXPR [ (95, _19831, _19831) (-1, _26963) 0 ]
MEM (id: 399, read at: x26963, value: x26964) 
EXPR [ (-1, _19831, _26964) (1, _19831, _25930) (1, _26964) (-1, _26965) 0 ]
MEM (id: 399, write x26965 at: x26963) 
EXPR [ (95, _19831, _19831) (1, _19831) (-1, _26966) 0 ]
BLACKBOX::RANGE [(26966)] [ ]
MEM (id: 399, read at: x813, value: x26967) 
EXPR [ (1, _19831, _26967) (-1, _26968) 0 ]
EXPR [ (1, _19831, _26968) (-1, _26969) 0 ]
...
MEM (id: 399, read at: x25926, value: x28254) 
EXPR [ (1, _19831, _28254) (-1, _28255) 0 ]
EXPR [ (1, _19831, _28255) (-1, _28256) 0 ]
BLACKBOX::RANGE [(28256)] [ ]
MEM (id: 399, read at: x25928, value: x28257) 
EXPR [ (1, _19831, _28257) (-1, _28258) 0 ]
EXPR [ (1, _19831, _28258) (-1, _28259) 0 ]
BLACKBOX::RANGE [(28259)] [ ]
EXPR [ (-1, _28260) 431 ]
MEM (id: 399, read at: x28260, value: x28261) 
EXPR [ (1, _19831, _28261) (-1, _28262) 0 ]
EXPR [ (1, _19831, _28262) (-1, _28263) 0 ]
BLACKBOX::RANGE [(28263)] [ ]
EXPR [ (-1, _28264) 432 ]
...
EXPR [ (-1, _28640) 526 ]
MEM (id: 399, read at: x28640, value: x28641) 
EXPR [ (1, _19831, _28641) (-1, _28642) 0 ]
EXPR [ (1, _19831, _28642) (-1, _28643) 0 ]
BLACKBOX::RANGE [(28643)] [ ]
```
And then another array gets initialised with `id: 400`, and so on all the way to `id: 830`. Maybe Maxim is right and there copies involved, although I can't see where, it's all reads except for the one write in the beginning.

aakoshh:
> how does it know it can't be

Ah, the `get_value_max_num_bits` 💡

TomAFrench:
We can handle this more cleanly after #6979

aakoshh:
I added a simple print and see tons of this kind of range checks:
```
inserting range check for mul (8): 1 vs 8
inserting range check for mul (32): 1 vs 32
```

Changed the code like this:
```rust
            BinaryOp::Mul => {
                if bit_size == 1
                    || max_lhs_bits + max_rhs_bits <= bit_size
                    || max_lhs_bits == 1
                    || max_rhs_bits == 1
                {
                    // Either performing boolean multiplication (which cannot overflow),
                    // or `lhs` and `rhs` have both been casted up from smaller types and so cannot overflow.
                    return Ok(());
                }
                "attempt to multiply with overflow".to_string()
            }
```
That brings the code size down from 50MB to 30MB, which is still higher than expected. Need to investigate the repeated arrays.

But that change seem correct, doesn't it?

aakoshh:
I can confirm that the SSA switches from using `array_set` to `make_array` during the `Remove IfElse` step. The SSA of `transfer_to_public` goes from 64K to 1M lines during that pass. 

Prior to it we have this; NB it's not using `array_set mut`:

```
    v45841 = make_array [u8 0, u8 0, ..., u8 0, u8 0] : [u8; 527]
    ...
    v46558 = make_array [v46447, v46449, ..., v43416, v43420, u8 0, u8 0, ..., u8 0, u8 0] : [u8; 527]
    v46559 = if v36715 then v46558 else (if v36714) v45841
    ...
    constrain v46566 == v46567, "ciphertext length mismatch"
    v46568 = lt u32 0, v45839
    v46569 = mul v46568, v36715
    constrain v46569 == v36715, "Index out of bounds"
    // read item 0
    v46570 = array_get v45840, index u32 0 -> u8
    v46571 = array_set v46559, index v46562, value v46570
    // increment i
    v46572 = add v46562, u32 1
    v46573 = if v36715 then v46571 else (if v36714) v45841
    // offset + 1
    v46574 = add v46562, u32 1
    v46575 = lt u32 1, v45839
    v46576 = mul v46575, v36715
    constrain v46576 == v36715, "Index out of bounds"
    // read item 1
    v46577 = array_get v45840, index u32 1 -> u8
    v46578 = array_set v46573, index v46574, value v46577
    // increment i
    v46579 = add v46574, u32 1
    v46580 = if v36715 then v46578 else (if v36714) v45841
    // offset + 2
    v46581 = add v46562, u32 2
    v46582 = lt u32 2, v45839
    v46583 = mul v46582, v36715
    constrain v46583 == v36715, "Index out of bounds"
    // read item 2
    v46584 = array_get v45840, index u32 2 -> u8
    v46585 = array_set v46580, index v46581, value v46584

```

test_programs/compile_success_no_bug/encrypted_log_regression/src/main.nr
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    encrypted_bytes
+}
+
+pub fn compute_incoming_body_ciphertext<let P: u32>(
+    plaintext: [u8; P],
+    aes_secret: [u8; 32],
+) -> [u8] {
```
TomAFrench, line 106: This return type is coercing the result to a slice which seems to mess with ACIRgen enough so we get a bunch of unnecessary memory operations. We could either make this return an array or we can have `array_get` instructions with constant indices which act on a `as_slice` instruction "look through it" and act on the underlying array instead.

test_programs/compile_success_no_bug/encrypted_log_regression/src/main.nr
TomAFrench: Yeah, I think we're handling this optimally here (minus the range constraints). Having a dynamic shift amount will result in this being a RAM write so this should be replaced with an unconstrained helper.

aakoshh: I agree I think `shift` is a "false flag". I added it to force memory operations, which ended up resembling what the original flame graphs suggested in terms of opcode composition, but looking at the [original SSA](https://github.com/noir-lang/noir/pull/6972#issuecomment-2576190022) revealed that the range checks are coming from conditional variables, rather than indexing, so I'll try to rewrite the test to include if-then-else instead. 

I'll also check why making `size` statically known changes the original SSA to be so much smaller, because that seems unrelated to any conditionals. Maybe a known `size` gets rid of the memory operations, and then the `Remove IfElse` doesn't cause copies in a loop 🤔

aakoshh:
With a statically known `size`, ie. using `let size = M - 32 - HEADER_SIZE - OVERHEAD_PADDING;`, the original SSA looks like this:

```fsharp 
    ...
    enable_side_effects v62654
    v62666 = mul v52946, v62652
    enable_side_effects v52946
    inc_rc v62665
    // initialize encrypted_bytes 
    v62667 = make_array [v62664, u8 0, u8 0, ..., u8 0] : [u8; 527]
    ...
    // re-initialize encrypted_bytes to contain the header
    v62777 = make_array [v62664, ..., v59629, u8 0, ..., u8 0] : [u8; 527]
    v62779 = eq u32 432, v62053
    v62780 = cast v52946 as u32
    v62781 = mul u32 432, v62780
    v62782 = mul v62053, v62780
    constrain v62781 == v62782, "ciphertext length mismatch"
    v62783 = lt u32 0, v62053
    v62784 = mul v62783, v52946
    constrain v62784 == v52946, "Index out of bounds"
    // read incoming_header_ciphertext[0]
    v62785 = array_get v62054, index u32 0 -> u8
    // re-initialize encrypted_bytes to contain the header, padding, then incoming_header_ciphertext[0]
    v62786 = make_array [v62664, ..., v59629, u8 0, ..., u8 0, v62785, u8 0, ..., u8 0] : [u8; 527]
    v62787 = lt u32 1, v62053
    v62788 = mul v62787, v52946
    constrain v62788 == v52946, "Index out of bounds"
    // read incoming_header_ciphertext[1]
    v62789 = array_get v62054, index u32 1 -> u8
    // re-initialize encrypted_bytes to contain the header, padding, then incoming_header_ciphertext[0..=1] 
    v62790 = make_array [v62664, ..., v59629, u8 0, ..., u8 0, v62785, v62789, u8 0, ..., u8 0] : [u8; 527]
    v62791 = lt u32 2, v62053
    v62792 = mul v62791, v52946
    constrain v62792 == v52946, "Index out of bounds"
    // read incoming_header_ciphertext[2]
    v62793 = array_get v62054, index u32 2 -> u8
    ... and so on, no looping copy ...
```

aakoshh:
I managed reproduce the issue by placing an artificial conditional into the program. Depending on where it is placed we either get [this](https://github.com/noir-lang/noir/pull/6972#issuecomment-2576333003) or [this](https://github.com/noir-lang/noir/pull/6972#issuecomment-2576360388) kind of SSA.

test_programs/compile_success_no_bug/encrypted_log_regression/src/main.nr
aakoshh: I tried returning an array here, but it didn't make a difference in the bytecode size. Removed this function completely now.

TomAFrench:
Interesting we're getting regressions here :thinking:

aakoshh:
Just compared the SSA of `sha256_bytes` with and without the extra mem2reg. The difference is that before is that there are many changes like these:

```
v6559 = truncate v6558 to 64 bits, max_bit_size: 114
```
into
```
v6559 = truncate v6558 to 64 bits, max_bit_size: 304
```
The `max_bit_size` increased in all cases 🤷‍♂️

TomAFrench:
> max_bit_size: 304

That's _very_ wrong. Can we figure out what's happening here and fix before we merge?

aakoshh:
Further weirdness:
* If I run `cargo run -p nargo_cli -- --program-dir . info --skip-underconstrained-check --skip-brillig-constraints-check` then I get the exact same bytecode size as on `master`, ~15K
* If I run `cargo run -p nargo_cli -- --program-dir . info --skip-underconstrained-check --skip-brillig-constraints-check --show-ssa` then we get ~20K

So I suppose it's something to do with the ID normalisation.

TomAFrench:
> v6559 = truncate v6558 to 64 bits, max_bit_size: 304


This looks like it's coming in through the removal of the bit shift instructions.

aakoshh:
Okay so what happens (I don't know why yet), is that with the extra mem2reg we have something like this:
```
v1 = Field 1
v2 = cast v1 as u64
v3 = shlift left v2 by 45 bits
```
When `remove_bit_shifts` sees the `Shl` operation it looks up the maximum bit size of the variable using `get_value_max_num_bits`, which sees the instruction is a `Cast` and looks up the original value, which happens to be a `Field`, which has 254 bytes, which results in this example a max value of 254+45=299 bits. 

Whereas without the extra `mem2reg` it sees the original type as `u64` and the max bits is just 109. 

I reckon `get_value_max_num_bits` could recognise a downcast as having got rid of the potential extra bytes. It makes sense the other way, that if we cast `u1` into `u32` then we want to recover that it's 1 bit, but if we cast `u64` to `u32` we don't want 64 again.

TomAFrench:
> I reckon get_value_max_num_bits could recognise a downcast as having got rid of the potential extra bytes. It makes sense the other way, that if we cast u1 into u32 then we want to recover that it's 1 bit, but if we cast u64 to u32 we don't want 64 again.

Yeah, I would say the correct behaviour is to take the smaller of the two types.

TomAFrench:
>  which results in this example a max value of 254+45=299 bits.

I would also say that we should cap the maximum number of bits at the size of the field.

aakoshh:
> I would also say that we should cap the maximum number of bits at the size of the field.

Okay, I added the limit to `insert_truncate` in `remove_bit_shifts`. Didn't change the formula that came up with 299, so that it still does the casts and multiplications to actually shift, although on second thought I think it would be doing the right thing anyway [here](https://github.com/noir-lang/noir/blob/8d7f1f909c50887bca6ff1243d70622c2a3b3ec8/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs#L137) 🤔

aakoshh:
I checked what was the difference between the SSA after `Remove IfElse` in the case of `nested_array_dynamic` which saw a 2.62% regression in size. Curiously the only difference seems to be that it changes some array operations to work on the input array, rather than another one derived from it:

Screenshot 2025-01-09 at 10 28 30

There is a similar case further down the SSA, but that's less clear, it favours one array over the other but not all the way back to the inputs.

TomAFrench:
```

struct Bar {
    inner: [Field; 3],
}

struct Foo {
    a: Field,
    b: [Field; 3],
    bar: Bar,
}

struct FooParent {
    array: [Field; 3],
    foos: [Foo; 4],
}

fn main(mut x: [Foo; 4], y: pub Field) {
    let foo_parent_one = FooParent { array: [0, 1, 2], foos: x };
    let foo_parent_two = FooParent { array: [3, 4, 5], foos: x };
    let mut foo_parents = [foo_parent_one, foo_parent_two];

    assert(foo_parents[1].foos[1].b == [5, 6, 21]);
    if y == 2 {
        foo_parents[y - 2].foos[y - 2].b = [10, 9, 8];
    } else {
        foo_parents[y - 2].foos[y - 2].b = [20, 19, 18];
    }
    assert(foo_parents[1].foos[1].b == [20, 19, 18]);

    assert(foo_parents[1].foos[1].b[2] == 18);
    if y == 3 {
        foo_parents[y - 2].foos[y - 2].b[y - 1] = 5000;
    } else {
        foo_parents[y - 2].foos[y - 2].b[y - 1] = 1000;
    }
    assert(foo_parents[1].foos[1].b[2] == 5000);
    // Set a dynamic array value
    // foo_parents[y - 2].foos[y - 3].b = foo_parents[y - 2].foos[y - 2].b;
    // assert(foo_parents[1].foos[0].b == [20, 19, 5000]);
}
```

This seems to be the minimal example I can get showing a regression

TomAFrench:
@aakoshh Let's create an issue reminding us to revisit `nested_array_dynamic` as we're doing some really bad ACIR generation here which is waaaaay more significant than this regression (and will solve this issue at the same time).

This program above should be solvable with a single memory block which we read and write to directly (with the internal arrays just resulting in offsets to the indices we're reading/writing to) however we're currently:

- Initializing a memory block fo `foo_parents` with everything
- Reading this at dynamic indices to pull out the `FoosParent`  at `y-2` and placing that in a new memory block
- Reading the `foos` field at dynamic indices to pull out the `Foo`  at `y-2` and placing that in a new memory block
- etc.

We're then getting a total of 19 memory blocks being generated for something which should need a couple at most.

@vezenovm's proposed changes to how we handle arrays should help here.

test_programs/compile_success_no_bug/encrypted_log_regression/src/main.nr
TomAFrench, line 1: Can you move this to `execution_success` and generate some inputs for this?

TomAFrench:
image
That jump in memory usage is also interesting :thinking: This doesn't need to be solved in this PR but can you create a followup issue?

test_programs/compile_success_no_bug/encrypted_log_regression/src/main.nr
aakoshh: Huh, I must be blind, I thought I confirmed that `rebuild.sh` will include this directory in the gate size comparison, but now I see [it does not](https://github.com/noir-lang/noir/blob/8d7f1f909c50887bca6ff1243d70622c2a3b3ec8/test_programs/rebuild.sh#L52) 🙈 
Thanks!

TomAFrench (approved):
LGTM

aakoshh:
Created followup tickets:
* https://github.com/noir-lang/noir/issues/7000
* https://github.com/noir-lang/noir/issues/7001

PR #6973, opened 2025-01-07
feat!: update `aes128_encrypt` to return an array

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR changes the signature of the `aes128_encrypt` blackbox function so that it returns an array rather than a slice.

Docs do not mention the return type of this function so no changes are necessary.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/.github/workflows/reports.yml
+++ b/.github/workflows/reports.yml
@@ -299,7 +299,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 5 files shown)


vezenovm (approved):

aakoshh:
This is great! For the record I just compiled the Token contract as it is in https://github.com/noir-lang/noir/issues/6929 with this change, but it doesn't seem to make a difference in bytecode size, it's all coming from the setter side.

PR #6974, opened 2025-01-07
fix: wrong module to lookup trait when using crate or super

asterite:
# Description

## Problem

Some warnings found in noir-bignum regarding traits.

## Summary

When doing path lookup there's a concept of "importing module", which is the module the triggered the path lookup, and "starting module", which is where we start looking names. Most of the time these are the same, for example if we do `use foo::Bar` we start at the current module and also look up `foo` in the current module. But when doing `use crate::...` or `use super::..` the starting module changes.

The code was using "starting module" instead of "importing module" to find traits in scope. This PR fixes that to use "importing module".

## Documentation

Check one:

# PR Checklist


TomAFrench (approved):
Nice!

PR #6983, opened 2025-01-08
fix: Do not emit range check for multiplication by bool

aakoshh:
# Description

## Problem\*

Found a bug during investigating https://github.com/noir-lang/noir/pull/6972

## Summary\*

Side effect variables are handled by multiplying values with them to turn them on/off. Because side effect variables are bools, multiplying an unsigned value with them will not result in an overflow: the value either stays what it is, or turns into zero. Yet, range checks were emitted.

The PR fixes the condition that causes `check_unsigned_overflow` to return early to consider the bit size of the LHS or the RHS on its own as sufficient evidence that we are multiplying with a bool value.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/.github/workflows/reports.yml
+++ b/.github/workflows/reports.yml
@@ -299,7 +299,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-contracts, is_library: true }
+          # - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-contracts, is_library: true }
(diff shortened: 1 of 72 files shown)


TomAFrench (commented):
We have a similar overflow check function in brillig gen (named the same thing afaik) can you make an equivalent change there? Otherwise LGTM.

aakoshh:
@TomAFrench thanks for pointing out the Brillig equivalent; I factored out both to call the same logic to avoid having to add another test. The ACIR test covers both cases, although it could be covered with a simpler unit test now.

TomAFrench:
@aakoshh Looks like tests are failing after this change.

aakoshh:
Yeah, I'm trying to run the report generation locally...

aakoshh:
Fixed 🙈

TomAFrench (approved):

Issue #6801, opened 2024-12-12
SSA: Global variables


vezenovm:
### Problem

Now that the compiler supports inlining Brillig functions based off of an aggressiveness setting, we can potentially have non-inlined functions in our final SSA. Previously, when we used to inline all functions a global would simply be a local constant that was constructed in the top block of a function. With multiple functions being able to re-use the same global, we may now be duplicating initialization of the those constants across functions. 

e.g. let's take the modified `regression_4709` test:
```
global EXPONENTIATE: [[Field; 2]; 2] = [
    [1, 1],
    [0, 0],
];
fn main(x: Field, y: pub Field) {
    let mut acc: Field = 0;
    for i in 0..2 {
        for j in 0..2 {
            acc += EXPONENTIATE[i][j];
        }
    }
    assert(!acc.lt(x));
    assert(x != y);

    dummy_again(x, y);
}
fn dummy_again(x: Field, y: pub Field) {
    let mut acc: Field = 0;
    for i in 0..2 {
        for j in 0..2 {
            acc += EXPONENTIATE[i][j];
        }
    }
    assert(!acc.lt(x));
    assert(x != y);
} 
```
If compiled with a really low inliner aggressiveness, while compiling to Brillig we will get the following SSA for `main`:
```
brillig(inline) fn main f0 {
  b0(v0: Field, v1: Field):
    v3 = allocate -> &mut Field
    store Field 0 at v3
    v6 = make_array [Field 1, Field 1] : [Field; 2]
    v7 = make_array [Field 0, Field 0] : [Field; 2]
    v8 = make_array [v6, v7] : [[Field; 2]; 2]
    jmp b1(u32 0)
  b1(v2: u32):
    v11 = lt v2, u32 2
    jmpif v11 then: b3, else: b2
  b2():
    v12 = load v3 -> Field
    v14 = call f2(v12, v0) -> u1
    constrain v14 == u1 0
    v16 = eq v0, v1
    constrain v16 == u1 0
    call f3(v0, v1)
    return
  b3():
    v21 = array_get v8, index v2 -> [Field; 2]
    v22 = load v3 -> Field
    ...
    inc_rc v21
    v23 = array_get v21, index u32 0 -> Field
    v24 = add v22, v23
    ...
    inc_rc v21
    v26 = array_get v21, index u32 1 -> Field
    v27 = add v24, v26
    store v27 at v3
    v28 = add v2, u32 1
    jmp b1(v28)
}
```
[...]

PR #6985, opened 2025-01-08
feat: SSA globals in monomorphization and SSA gen 

vezenovm:
# Description

## Problem\*

Working towards #6801 

## Summary\*

This PR does the initial work to move towards handling globals in SSA. Actually utilizing globals in Brillig gen will come in a follow-up PR. I wanted to split out this work for easier reviewing as all the globals work together will be a good amount of code. 

I wanted to add globals to SSA without any major changes to our existing data structures and or having to thread lifetimes throughout our structures. In follow-up work we most likely will want to further abstract some our concepts such as the `DataFlowGraph` or pass a global context into all SSA functions. For now, I wanted to focus on minimal changes without causing regressions. 

The main changes:
1. During monomorphization we now keep track of globals and have a new `Definition::Global`. If we have a new global we monomorphize the global's expression but then wrap it in an ident with a `Definition::Global`. For closures we keep the old strategy of inlining the global during monomorphization. 
2. The `SharedContext` now contains a `GlobalsContext` 
3. The `GlobalsContext` re-uses the `DataFlowGraph` to build globals. The newly built globals are stored in a `globals: BTreeMap<monomorphization::ast::GlobalId, Values>` map to be accessed by the `FunctionContext`. 
4. Before performing any code gen for a function we full up the value IDs that are reserved for globals. 
5. A global is represented in a function's DFG simply with a `Value::Global`. When we see `Value::Global` it indicates we have a reserved value id and that we need to access some separate global context for the actual value. Currently, the only pass that makes use of this is `inlining`. We are currently still inlining all globals so that the globals work can be split up for easier reviewing.

Example SSA for the new `global_var_regression_simple` test (commenting out `assert(!acc.lt(x))` for a simpler SSA):

<details>
  <summary>SSA output with globals</summary>

```
After Initial SSA:
g0 = Field 1
g1 = make_array [Field 1, Field 1] : [Field; 2]
g2 = Field 0
g3 = make_array [Field 0, Field 0] : [Field; 2]
g4 = make_array [v1, v3] : [[Field; 2]; 2]

brillig(inline) fn main f0 {
  b0(v5: Field, v6: Field):
    v9 = allocate -> &mut Field
    store Field 0 at v9
    jmp b1(u32 0)
  b1(v7: u32):
    v13 = lt v7, u32 2
    jmpif v13 then: b3, else: b2
  b2():
    v14 = eq v5, v6
    v15 = not v14
    constrain v14 == u1 0
    call f1(v5, v6)
    return
  b3():
    jmp b4(u32 0)
  b4(v8: u32):
    v18 = lt v8, u32 2
    jmpif v18 then: b6, else: b5
  b5():
    v20 = add v7, u32 1
    jmp b1(v20)
  b6():
    v21 = load v9 -> Field
    v22 = array_get g4, index v7 -> [Field; 2]
    inc_rc v22
    v23 = array_get v22, index v8 -> Field
    v24 = add v21, v23
    store v24 at v9
    v25 = add v8, u32 1
    jmp b4(v25)
}
brillig(inline) fn dummy_again f1 {
  b0(v5: Field, v6: Field):
    v9 = allocate -> &mut Field
    store Field 0 at v9
    jmp b1(u32 0)
  b1(v7: u32):
    v13 = lt v7, u32 2
    jmpif v13 then: b3, else: b2
  b2():
    v14 = eq v5, v6
    v15 = not v14
    constrain v14 == u1 0
    return
  b3():
    jmp b4(u32 0)
  b4(v8: u32):
    v17 = lt v8, u32 2
    jmpif v17 then: b6, else: b5
  b5():
    v19 = add v7, u32 1
    jmp b1(v19)
  b6():
    v20 = load v9 -> Field
    v21 = array_get g4, index v7 -> [Field; 2]
    inc_rc v21
    v22 = array_get v21, index v8 -> Field
    v23 = add v20, v22
    store v23 at v9
    v24 = add v8, u32 1
    jmp b4(v24)
}
```
</details>

## Additional Context

Actual implementation of globals and handling them as we do existing constants will come in followups. It may be helpful to wait for the PR with the Brillig gen implementation to see how globals will actually be used.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
I'm sure you're thinking about this but looks like we don't have global arrays yet which I expect would be the biggest contributor here.

TomAFrench:
Ah gotcha, I'm jumping the gun in that case.

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -1068,3 +1098,207 @@ pub(super) enum LValue {
...
+                })
+            }
+            ast::Literal::Slice(array) => {
+                let elements = self.codegen_array_elements(&array.contents)?;
+
+                let typ = FunctionContext::convert_type(&array.typ).flatten();
+                Ok(match array.typ {
+                    ast::Type::Slice(_) => {
+                        // let slice_length =
+                        // self.builder.length_constant(array.contents.len() as u128);
```
vezenovm, line 1199:
```suggestion
```

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -1068,3 +1098,208 @@ pub(super) enum LValue {
...
+                } else {
+                    value
+                };
+
+                Ok(self.dfg.make_constant(value, numeric_type).into())
+            }
+            ast::Literal::Bool(value) => {
+                Ok(self.dfg.make_constant((*value).into(), NumericType::bool()).into())
+            }
+            ast::Literal::Unit => todo!(),
```
vezenovm, line 1239:
```suggestion
            ast::Literal::Unit => Ok(Self::unit_value()),
```

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -1068,3 +1098,208 @@ pub(super) enum LValue {
...
+                };
+
+                Ok(self.dfg.make_constant(value, numeric_type).into())
+            }
+            ast::Literal::Bool(value) => {
+                Ok(self.dfg.make_constant((*value).into(), NumericType::bool()).into())
+            }
+            ast::Literal::Unit => Ok(Self::unit_value()),
+            ast::Literal::Str(string) => Ok(self.codegen_string(string)),
+            ast::Literal::FmtStr(_, _, _) => todo!(),
```
vezenovm, line 1241:
```suggestion
            ast::Literal::FmtStr(_, _, _) => {
                unreachable!("Format strings are lowered as normal strings as they are already interpolated");
            }
```

test_programs/execution_success/global_var_regression_simple/src/main.nr
```diff
@@ -0,0 +1,25 @@
...
+}
+
+fn dummy_again(x: Field, y: Field) {
+    let mut acc: Field = 0;
+    for i in 0..2 {
+        for j in 0..2 {
+            acc += EXPONENTIATE[i][j];
+        }
+    }
+    // assert(!acc.lt(x));
```
vezenovm, line 23:
```suggestion
    assert(!acc.lt(x));
```

compiler/noirc_evaluator/src/ssa/ssa_gen/program.rs
```diff
@@ -101,13 +107,39 @@ impl Ssa {
...
         for function in self.functions.values() {
             writeln!(f, "{function}")?;
         }
         Ok(())
     }
 }
 
+impl Display for GlobalsContext {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "Globals: ")?;
```
asterite, line 123: When printing the SSA, "Globals:" shows up as the first thing. Is this intended? For example:

```
Globals:
@v0 = Field 1

acir(inline) fn main f0 {
  b0():
    call f1(u1 1, Field 1)
    return
}
```

Just asking because eventually we'd have to parse globals, and maybe not having "Globals:" here could be easier (also less noise). Also this could be changed later on so feel free to ignore this comment.

asterite (approved):
LGTM!

compiler/noirc_evaluator/src/ssa/ssa_gen/program.rs
vezenovm: I did add it intentionally but was not sure on it because as you mentioned it adds some noise. I guess you think it is clear on its own we are in the global space without the "Globals:"?

asterite: Oh, maybe it's fine, we can easily parse that. I was looking at how it's in LLVM, it's like this:

```
@variable = global i32 21
```

In our case that could be:

```
@v0 = global Field 1
```

but I'm not sure I actually like it 😄 

Another alternative is to print "// Globals:" instead of "Globals": the SSA parser already supports comments so that would be ignored.

test_programs/execution_success/global_var_regression_simple/src/main.nr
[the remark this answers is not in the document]
vezenovm: With the current state of globals in this PR it will not. In my follow-up work that implements this in Brillig we do see benefits and that is why I added it. I could just include it in that PR as a regression though if that is preferred. Or perhaps extend `regression_4709` (that test is basically a much simplified version of `regression_4709`)

compiler/noirc_evaluator/src/ssa/ssa_gen/program.rs
vezenovm: I gated printing globals 
```rust
if self.globals.dfg.values_iter().len() > 0 {
     write!(f, "{}", self.globals)?;
}
```
as to avoid triggering the SSA parser for now. I could switch it to `// Globals: ` and remove that check? I also feel `@v*` states clearly that we have a global as that is the convention we are following.

asterite: Either way is fine. Thank you!

compiler/noirc_evaluator/src/ssa/ir/printer.rs
```diff
@@ -53,6 +53,9 @@ fn value(dfg: &DataFlowGraph, id: ValueId) -> String {
         Value::Intrinsic(intrinsic) => intrinsic.to_string(),
         Value::ForeignFunction(function) => function.clone(),
         Value::Param { .. } | Value::Instruction { .. } => id.to_string(),
+        Value::Global(_) => {
+            format!("@{id}")
+        }
```
jfecher, line 58: Can you change the format of globals to avoid the LLVM `@`? The rest of our IR is based on cranelift, and I think we can use our existing convention of changing the letter before the id instead. E.g. `g0`, `g1`, etc.

compiler/noirc_evaluator/src/ssa/ssa_gen/program.rs
```diff
@@ -101,13 +107,39 @@ impl Ssa {
...
+        }
+
         for function in self.functions.values() {
             writeln!(f, "{function}")?;
         }
         Ok(())
     }
 }
 
+impl Display for GlobalsContext {
```
jfecher, line 121: Can you move this Display impl (and probably Display for Ssa as well) to the ssa printer module?

compiler/noirc_evaluator/src/ssa/ssa_gen/program.rs
jfecher: I agree that the leading `Globals: ` or `// Globals:` is probably unnecessary and can be removed

compiler/noirc_frontend/src/hir/comptime/value.rs
```diff
@@ -521,6 +521,10 @@ impl Value {
         )
     }
 
+    pub(crate) fn is_closure(&self) -> bool {
+        matches!(self, Value::Closure(_, _, _, _, _))
```
jfecher, line 525:
```suggestion
        matches!(self, Value::Closure(..))
```

compiler/noirc_frontend/src/monomorphization/mod.rs
```diff
@@ -928,16 +948,51 @@ impl<'interner> Monomorphizer<'interner> {
...
             DefinitionKind::Global(global_id) => {
                 let global = self.interner.get_global(*global_id);
-
-                let expr = if let GlobalValue::Resolved(value) = global.value.clone() {
-                    value
-                        .into_hir_expression(self.interner, global.location)
-                        .map_err(MonomorphizationError::InterpreterError)?
+                let id = global.id;
+                let name = definition.name.clone();
+                if let Some(seen_global) = self.globals.get(&id) {
```
jfecher, line 953: The case for `DefinitionKind::Global` is quite long now, can you break it up into its own function?

compiler/noirc_frontend/src/monomorphization/mod.rs
```diff
@@ -928,16 +948,51 @@ impl<'interner> Monomorphizer<'interner> {
...
+                    } else {
+                        unreachable!("All global values should be resolved at compile time and before monomorphization");
+                    };
 
-                self.expr(expr)?
+                    let expr = self.expr(expr)?;
+
+                    if !is_closure {
+                        let new_id = self.next_global_id();
+                        self.globals.insert(id, new_id);
```
jfecher, line 980: Can you add a comment on why we need this closure check?

compiler/noirc_evaluator/src/ssa/ir/value.rs
```diff
@@ -27,16 +27,27 @@ pub(crate) enum Value {
...
     /// This Value originates from a block parameter. Since function parameters
     /// are also represented as block parameters, this includes function parameters as well.
     ///
     /// position -- the index of this Value in the block parameters list
-    Param { block: BasicBlockId, position: usize, typ: Type },
+    Param {
+        block: BasicBlockId,
+        position: usize,
+        typ: Type,
+    },
```
jfecher, line 44: Odd that the formatter touched this when it hasn't changed? Only a new variant was added. Maybe it is the lack of a doc comment on it? Odd.

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -126,6 +138,11 @@ impl<'a> FunctionContext<'a> {
         } else {
             self.builder.new_function(func.name.clone(), id, func.inline_type);
         }
+
+        for (_, value) in self.shared_context.globals_context.dfg.values_iter() {
+            self.builder.current_function.dfg.make_global(value.get_type().into_owned());
+        }
```
jfecher, line 144: This snippet has popped up here twice, and in normalize_value_ids IIRC. It's non-trivial enough that I think we should move it into a helper function.

test_programs/execution_success/global_var_regression_simple/src/main.nr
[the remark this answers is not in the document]
TomAFrench: Nah, if it's going to show an improvement in a later PR then we can keep it here.

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -1068,3 +1098,212 @@ pub(super) enum LValue {
...
+
+impl GlobalsContext {
+    fn codegen_expression(&mut self, expr: &ast::Expression) -> Result<Values, RuntimeError> {
+        match expr {
+            ast::Expression::Ident(ident) => Ok(self.codegen_ident(ident)),
+            ast::Expression::Literal(literal) => self.codegen_literal(literal),
+            ast::Expression::Block(block) => self.codegen_block(block),
+            ast::Expression::Let(let_expr) => self.codegen_let(let_expr),
+            ast::Expression::Tuple(tuple) => self.codegen_tuple(tuple),
+            _ => {
```
jfecher, line 1118: Really not a fan of duplicating all these codegen functions. Could we create a FunctionBuilder in SharedContext instead and reuse the existing code-gen functions? Would definitely need some setup code since we'd need a full context in that case not just a function builder, and we'd need to not recursively keep filling globals when we're already doing so.

I just see this code getting forgotten about very easily when we're updating any of these codegen functions. Don't want to have to worry about updating both versions each time.

jfecher (changes requested):
Mostly minor comments but the main item is avoiding duplication of codegen functions

compiler/noirc_evaluator/src/ssa/ir/value.rs
vezenovm: Either way I'll add a doc comment to the global variant.

compiler/noirc_evaluator/src/ssa/ir/printer.rs
vezenovm: I'll change to the letter. 

Just noting that in this case our globals are based more on LLVM. Global values in cranelift are _not_ known as compile time https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/docs/ir.md#global-values. In LLVM global variables represent regions of memory allocated at compile time that must be initialized.

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
vezenovm: > Really not a fan of duplicating all these codegen functions

Yeah same here. I just duplicated for simplicity sake at the moment. I wanted to just use a `FunctionBuilder` but as you mentioned we do need a full context. I was planning to do this cleanup in followups but I can do it in this PR.

compiler/noirc_evaluator/src/ssa/ir/printer.rs
jfecher: Yep, I'm aware, just thought we should keep the same naming scheme

compiler/noirc_evaluator/src/ssa/ir/value.rs
vezenovm: Looks like the formatter acted as it did previously with a new doc comment.

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
vezenovm: done

compiler/noirc_evaluator/src/ssa/ssa_gen/program.rs
vezenovm: There is no longer a `GlobalsContext` wrapper struct but I have moved `Display for Ssa` to the printer module.

compiler/noirc_evaluator/src/ssa/ssa_gen/program.rs
vezenovm: I have removed it

compiler/noirc_evaluator/src/ssa/ir/printer.rs
vezenovm: I have updated the PR description with how the SSA looks now.

compiler/noirc_frontend/src/monomorphization/mod.rs
vezenovm: done

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
vezenovm: I have re-used the existing builder flow. The set-up did not end up being that bad actually. Let me know what you think.

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -1022,11 +1044,32 @@ fn convert_operator(op: BinaryOpKind) -> BinaryOp {
...
+            let values = context.codegen_expression(global).unwrap();
+            globals.insert(*id, values);
+        }
+
         Self {
             functions: Default::default(),
             function_queue: Default::default(),
             function_counter: Default::default(),
             program,
+            globals_context: context.builder.current_function.dfg,
```
vezenovm, line 1071: I'm going to switch to having a pseudo `Function` type to make later code gen simpler, but the logic of the PR will essentially remain the same.

vezenovm: Ok I have switched to a `Function`. This PR should be ready for review again @jfecher

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
```diff
@@ -98,20 +98,22 @@ struct InlineContext {
...
+    globals: &'global Function,
 }
 
 /// The per-function inlining context contains information that is only valid for one function.
 /// For example, each function has its own DataFlowGraph, and thus each function needs a translation
 /// layer to translate between BlockId to BlockId for the current function and the function to
 /// inline into. The same goes for ValueIds, InstructionIds, and for storing other data like
 /// parameter to argument mappings.
-struct PerFunctionContext<'function> {
+struct PerFunctionContext<'function, 'global> {
```
TomAFrench, line 110: I'm unsure that we actually need two lifetimes here. Can we make do with one?

vezenovm: Yeah good call. I can pass the globals to the `FunctionContext` directly.

TomAFrench (approved):

TomAFrench:
@jfecher

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -1043,10 +1046,22 @@ fn convert_operator(op: BinaryOpKind) -> BinaryOp {
...
+
+        let globals_id = Program::global_space_id();
+
+        // Queue the function representing the globals space for compilation
+        globals_shared_context.get_or_queue_function(globals_id);
+
+        let mut context = FunctionContext::new(
+            "globals".to_owned(),
+            &vec![],
+            RuntimeType::Brillig(InlineType::default()),
```
jfecher, line 1059: What does making globals' runtime brillig affect?

compiler/noirc_evaluator/src/ssa/ssa_gen/program.rs
```diff
@@ -8,20 +8,16 @@ use serde_with::serde_as;
...
-use super::context::GlobalsContext;
-
 /// Contains the entire SSA representation of the program.
 #[serde_as]
 #[derive(Serialize, Deserialize)]
 pub(crate) struct Ssa {
     #[serde_as(as = "Vec<(_, _)>")]
     pub(crate) functions: BTreeMap<FunctionId, Function>,
-    pub(crate) globals: GlobalsContext,
+    pub(crate) globals: Function,
```
jfecher, line 20: Might still want a type alias or wrapper for this (can wait until a future PR).
That way users wouldn't have to know why this is a Function and we can limit that confusion/detail to its own module.

jfecher (approved):
LGTM. Thanks for the ssa-gen change

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
vezenovm: It doesn't affect anything. I just chose Brillig as all ACIR globals will be inlined anyway so I felt it was clearer to mark it Brillig

PR #6987, opened 2025-01-08
fix: allow multiple trait impls for the same trait as long as one is in scope

asterite:
# Description

## Problem

This code used to give an error saying multiple traits (`Convert` and `Convert`) were in scope:

```noir
struct Foo {
    inner: Field,
}

trait Converter<N> {
    fn convert(self) -> N;
}

impl Converter<Field> for Foo {
    fn convert(self) -> Field {
        self.inner
    }
}

impl Converter<u32> for Foo {
    fn convert(self) -> u32 {
        self.inner as u32
    }
}

fn main() {
    let foo = Foo { inner: 42 };
    let _: u32 = foo.convert();
}
```

## Summary

Now the error is gone. We search if there's only one trait, not one trait method, in scope. In the snippet above there are two but they are for the same trait so it's fine. Later on the type-checker checks that it's only bound to the one for `u32`.

If we remove the `u32` type annotation, we get this error:

```
error: Multiple trait impls match the object type `Foo`
   ┌─ src/main.nr:25:13
   │
25 │     let x = foo.convert();
   │             ----------- Ambiguous impl
   │
   = Candidate 1: `Foo: Converter<Field>`
   = Candidate 2: `Foo: Converter<u32>`
``` 

## Additional Context

To be honest, I'm not sure where's the code that eventually checks what's the impl for `foo.convert()` that matches against `u32`, but seeing it work, and seeing the error we get when there are multiple matching impls, I'm confident it's working fine (it seems to be in `verify_trait_constraint` right before we end analyzing a function)

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/types.rs
+++ b/compiler/noirc_frontend/src/elaborator/types.rs
@@ -1,5 +1,6 @@
 use std::{borrow::Cow, rc::Rc};
 
+use im::HashSet;
 use iter_extended::vecmap;
 use noirc_errors::{Location, Span};
 use rustc_hash::FxHashMap as HashMap;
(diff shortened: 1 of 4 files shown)


jfecher (approved):
LGTM

`lookup_struct_method` has gotten way too large, we should break it up into pieces sometime

asterite:
> lookup_struct_method has gotten way too large, we should break it up into pieces sometime

Makes sense. There are a couple more trait PRs. After those I'll try to simplify things.

TomAFrench:
@asterite This is causing errors in aztec-packages.

asterite:
Oh, I didn't notice. All tests passed. Was it because what you found in the other PR?

asterite:
Where can I see the errors?

TomAFrench:
Nope, we're failing to compile the `private_kernel_reset` package as part of the benchmarking workflow here.

TomAFrench:
https://github.com/noir-lang/noir/actions/runs/12675812872/job/35327773532

asterite:
Strange, I don't understand why it errors. That said, I have a workaround, I'll push that for now.

PR #6988, opened 2025-01-08
chore(ci): fail properly in `external-repo-checks`

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

We're not properly catching when we completely brick an external project as currently we see no difference between this and a project for which all tests are passing.

This PR fixes the CI so that we can distinguish between these two cases.

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
Thanks! I probably relied on the `diff` call for the exit status, then eventually made it show a better error and forgot to exit with -1.

PR #6992, opened 2025-01-08
chore: require safety doc comment for unsafe instead of `//@safety`

asterite:
# Description

## Problem

Experimental: Nico requested having `//@safety` before the `unsafe` block for readability reasons. In the past I suggested using doc comments, so this is what I tried here.

## Summary

The requirement is now that the `unsafe` block has a doc comment that starts with "Safety:" (ignoring case) before it. The doc comment can be exactly before the unsafe expression, or it can be in the statement that holds the unsafe block. I think this gives a bit more flexibility so that the safety comment isn't nested inside other expressions, leading to less readable comments and code.

For example, all of these are valid:

```noir
/// Safety: test
unsafe { 1 }

/// Safety: test
let x = unsafe { 1 };

/// Safety: test
call(unsafe { 1 });
```

## Documentation

Check one:

# PR Checklist


TomAFrench (approved):
Much nicer!

benesjan:
I am getting a few warnings in [my PR](https://github.com/AztecProtocol/aztec-packages/pull/11094) in which it's not clear to me why they are triggered:

image

Typically it does not recognize [this comment ](https://github.com/AztecProtocol/aztec-packages/blob/3e094bc69fe31bffab2ced72cd155bc4ab435b6d/noir-projects/noir-contracts/contracts/nft_contract/src/types/nft_note.nr#L54) as valid.

To reproduce:
1. check out https://github.com/AztecProtocol/aztec-packages/blob/3e094bc69fe31bffab2ced72cd155bc4ab435b6d/noir-projects/noir-contracts/contracts/nft_contract/src/types/nft_note.nr#L54
2. go to `noir-projects/noir-contracts`
3. and compile with `nargo compile`

This is my nargo version:

```
% nargo --version
nargo version = 1.0.0-beta.1
noirc version = 1.0.0-beta.1+56ed6e16b975f91a
(git version hash: 56ed6e16b975f91a, is dirty: false)
```

asterite:
@benesjan Hi! The comment on top of `unsafe` should be `///` (triple slash). I hope this fixes the issue for you.

PR #6994, opened 2025-01-08
fix: Keep format string literal when converting comptime value to HIR expression

### LINKED ISSUE #6990: Cannot print global fmtstr
### Aim

Try and run this function:
```noir
global ONE: Field = 50;
global TWO: Field = 100;
global GLOBAL: fmtstr<18, (Field, Field)> = f"i: {ONE}, j: {TWO}";

fn main() {
    let string = "i: {i}, j: {j}";
    println(string);

    println(GLOBAL);
}
```

### Expected Behavior

I expect to see this output:
```
i: {i}, j: {j}
i: 0x32, j: 0x64
```

### Bug

Instead I get this panic:
```
Message:  assertion `left == right` failed
  left: 4
 right: 1
Location: compiler/noirc_evaluator/src/ssa/opt/inlining.rs:428
```
The SSA produced is the following:
```
After Initial SSA::
brillig(inline) fn main f0 {
  b0(v0: Field, v1: Field):
    v9 = make_array b"i: {i}, j: {j}"
    inc_rc v9
    call f1(v9)
    v14 = make_array b"i: 50, j: 100"
    call f2(v14)
    return
}
brillig(inline) fn println f1 {
  b0(v0: [u8; 14]):
    call f4(u1 1, v0)
    return
}
brillig(inline) fn println f2 {
  b0(v0: [u8; 18], v1: Field, v2: Field, v3: Field):
    call f3(u1 1, v0, v1, v2, v3)
    return
}
brillig(inline) fn print_unconstrained f3 {
  b0(v0: u1, v1: [u8; 18], v2: Field, v3: Field, v4: Field):
    v16 = make_array b"{\"kind\":\"field\"}"
    v17 = make_array b"{\"kind\":\"field\"}"
    call print(v0, v1, v2, v3, v4, v16, v17, u1 1)
    return
}
brillig(inline) fn print_unconstrained f4 {
  b0(v0: u1, v1: [u8; 14]):
    v20 = make_array b"{\"kind\":\"string\",\"length\":14}"
    call print(v0, v1, v20, u1 0)
    return
}
```
As you can see a global fmtstr is lowered into a basic string. This is also noted here https://github.com/noir-lang/noir/blob/101f2d4a134b9c43f11de1653f8ab02574dca312/compiler/noirc_frontend/src/hir/comptime/value.rs#L365

However, the function type still sees a fmtstr so we have a discrepancy in the number of parameters we generate for the monomorphized println and the actual arguments passed to the function. Thus, we fail in inlining where we check that the number of arguments and parameters match.

### To Reproduce

1. Copy the snippet above
2. Run `nargo compile`

### Workaround

Yes

### Workaround Description

Use a local fmtstr. e.g. this function will print accurately:
```noir
global ONE: Field = 50;
global TWO: Field = 100;

    let string = "i: {i}, j: {j}";
    println(string);

    let GLOBAL: fmtstr<18, (Field, Field)> = f"i: {ONE}, j: {TWO}";
    println(GLOBAL);
}
```
Here is an example of valid SSA:
```
After Initial SSA::
brillig(inline) fn main f0 {
  b0():
    v7 = make_array b"i: {i}, j: {j}"
    inc_rc v7
    call f1(v7)
    v14 = make_array b"i: {ONE}, j: {TWO}"
    inc_rc v14
    call f2(v14, u32 2, Field 50, Field 100)
    return
}
brillig(inline) fn println f1 {
  b0(v0: [u8; 14]):
    call f4(u1 1, v0)
    return
}
brillig(inline) fn println f2 {
  b0(v0: [u8; 18], v1: Field, v2: Field, v3: Field):
    call f3(u1 1, v0, v1, v2, v3)
    return
}
brillig(inline) fn print_unconstrained f3 {
  b0(v0: u1, v1: [u8; 18], v2: Field, v3: Field, v4: Field):
    v16 = make_array b"{\"kind\":\"field\"}"
    v17 = make_array b"{\"kind\":\"field\"}"
    call print(v0, v1, v2, v3, v4, v16, v17, u1 1)
    return
}
brillig(inline) fn print_unconstrained f4 {
  b0(v0: u1, v1: [u8; 14]):
    v20 = make_array b"{\"kind\":\"string\",\"length\":14}"
    call print(v0, v1, v20, u1 0)
    return
}
```

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.1 noirc version = 1.0.0-beta.1+77441793fac75cc2e891ec1ee59e4224310ed037 (git version hash: 101f2d4a134b9c43f11de1653f8ab02574dca312, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #6990 

## Summary\*

Go to the issue for the initial context. What I did is a little bit hack-y, but it seemed like the simplest way to prevent this panic without making significant changes to how we monomorphize. 

Format strings are the only comptime type which we lower into a different type (in this case a string literal). I changed the `Value::FormatString` case in `into_hir_expression` to return a `HirLiteral::FmtStr` in the format that the type is expected to be. This required creating a new placeholder expression. In this case I used a string with a single whitespace character as this required minimal changes and does not change the functionality of fmt strings which are entirely used for logging. 

I added some global fmt strings to the `debug_logs` test:
These prints:
```noir
global A: Field = 50;
global B: Field = 100;
global GLOBAL_FMT_STR: fmtstr<14, (Field, Field)> = f"i: {A}, j: {B}";
global GLOBAL_FMT_STR_BIG: fmtstr<22, (Field, Field, Field)> = f"i: {A}, j: {B}, k: {A}";

fn main(..) {
    println(GLOBAL_FMT_STR);
    println(GLOBAL_FMT_STR_BIG);
}
```
Give this output:
```
i: 50, j: 100
i: 50, j: 100, k: 50
```

## Additional Context

We do not currently allow nested format strings. Having this extra whitespace may cause some difficulty if we ever activate the ability to have nested format strings.    

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/value.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/value.rs
@@ -20,7 +20,7 @@ use crate::{
     },
     node_interner::{ExprId, FuncId, NodeInterner, StmtId, StructId, TraitId, TraitImplId},
     parser::{Item, Parser},
-    token::{SpannedToken, Token, Tokens},
+    token::{FmtStrFragment, SpannedToken, Token, Tokens},
     Kind, QuotedType, Shared, Type, TypeBindings,
 };
 use rustc_hash::FxHashMap as HashMap;
@@ -363,8 +363,35 @@ impl Value {
                 HirExpression::Literal(HirLiteral::Str(unwrap_rc(value)))
             }
             // Format strings are lowered as normal strings since they are already interpolated.
-            Value::FormatString(value, _) => {
-                HirExpression::Literal(HirLiteral::Str(unwrap_rc(value)))
+            // We need to mock a FmtStr literal that matches the type expected elsewhere in the program.
+            // If we do not do this, we may end up with mismatched types
+            Value::FormatString(value, fmt_str_type) => {
+                // We include a space here as otherwise the ABI decoder will expect a value that does not exist.
+                let placeholder =
+                    interner.push_expr(HirExpression::Literal(HirLiteral::Str(" ".to_owned())));
+                interner.push_expr_location(placeholder, location.span, location.file);
+
+                let dummy_captures = match fmt_str_type {
+                    Type::FmtString(_, item_types) => {
+                        dbg!(item_types.clone());
+                        let Type::Tuple(fields) = *item_types else {
+                            return Err(InterpreterError::Unimplemented {
+                                item: "Only Type::Tuple is supported for format strings".to_owned(),
+                                location,
+                            });
+                        };
+                        std::iter::repeat(placeholder).take(fields.len()).collect()
+                    }
+                    _ => unreachable!("Expected a Type::FmtString"),
+                };
+
+                let raw_str = unwrap_rc(value);
+                let str_len = raw_str.len();
+                HirExpression::Literal(HirLiteral::FmtStr(
+                    vec![FmtStrFragment::String(raw_str)],
+                    dummy_captures,
+                    str_len as u32,
+                ))
             }
             Value::Function(id, typ, bindings) => {
                 let id = interner.function_definition_id(id);
--- a/test_programs/execution_success/debug_logs/src/main.nr
+++ b/test_programs/execution_success/debug_logs/src/main.nr
@@ -1,6 +1,17 @@
+global A: Field = 50;
+global B: Field = 100;
+global GLOBAL_FMT_STR: fmtstr<14, (Field, Field)> = f"i: {A}, j: {B}";
+global GLOBAL_FMT_STR_BIG: fmtstr<22, (Field, Field, Field)> = f"i: {A}, j: {B}, k: {A}";
+
 fn main(x: Field, y: pub Field) {
     let string = "i: {i}, j: {j}";
     println(string);
+
+    println(GLOBAL_FMT_STR);
+    println(GLOBAL_FMT_STR_BIG);
+
+    let LOCAL_FMT_STR: fmtstr<14, (Field, Field)> = f"i: {A}, j: {B}";
+    println(LOCAL_FMT_STR);
 
     // TODO: fmtstr cannot be printed
     // let fmt_str: fmtstr<14, (Field, Field)> = f"i: {x}, j: {y}";
--- a/test_programs/execution_success/debug_logs/src/main.nr
+++ b/test_programs/execution_success/debug_logs/src/main.nr
@@ -13,7 +13,7 @@ fn main(x: Field, y: pub Field) {
     let LOCAL_FMT_STR: fmtstr<14, (Field, Field)> = f"i: {A}, j: {B}";
     println(LOCAL_FMT_STR);
 
-    // TODO: fmtstr cannot be printed
+    // TODO: fmtstr cannot be printed inside another fmtstr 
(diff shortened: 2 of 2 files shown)


compiler/noirc_frontend/src/hir/comptime/value.rs
```diff
@@ -363,8 +363,35 @@ impl Value {
...
             }
             // Format strings are lowered as normal strings since they are already interpolated.
-            Value::FormatString(value, _) => {
-                HirExpression::Literal(HirLiteral::Str(unwrap_rc(value)))
+            // We need to mock a FmtStr literal that matches the type expected elsewhere in the program.
+            // If we do not do this, we may end up with mismatched types
+            Value::FormatString(value, fmt_str_type) => {
+                // We include a space here as otherwise the ABI decoder will expect a value that does not exist.
+                let placeholder =
+                    interner.push_expr(HirExpression::Literal(HirLiteral::Str(" ".to_owned())));
```
michaeljklein, line 371: Why a space `" "` and not an empty string `""`?

compiler/noirc_frontend/src/hir/comptime/value.rs
```diff
@@ -363,8 +363,35 @@ impl Value {
...
+                interner.push_expr_location(placeholder, location.span, location.file);
+
+                let dummy_captures = match fmt_str_type {
+                    Type::FmtString(_, item_types) => {
+                        dbg!(item_types.clone());
```
michaeljklein, line 376: Leftover `dbg`

test_programs/execution_success/debug_logs/src/main.nr
```diff
@@ -1,7 +1,18 @@
...
     let string = "i: {i}, j: {j}";
     println(string);
 
+    println(GLOBAL_FMT_STR);
+    println(GLOBAL_FMT_STR_BIG);
+
+    let LOCAL_FMT_STR: fmtstr<14, (Field, Field)> = f"i: {A}, j: {B}";
+    println(LOCAL_FMT_STR);
+
     // TODO: fmtstr cannot be printed
```
michaeljklein, line 16:
```suggestion
```

Need to update comment?

michaeljklein (changes requested):

compiler/noirc_frontend/src/hir/comptime/value.rs
vezenovm: removed

test_programs/execution_success/debug_logs/src/main.nr
vezenovm: No this is still the case as it is referring to the fmtstr being used in another fmtstr below. I made it more specific "fmtstr cannot be printed inside another fmtstr"

compiler/noirc_frontend/src/hir/comptime/value.rs
vezenovm: It is due to the ABI decoder expecting a certain number of fields. But your comment made me realize this hack is not going to work in scenarios in types with differing field counts. Will push an update.

vezenovm:
Going to close this PR as the correct solution is to accurately monomorphize the functions based off of the new lowered type. This hack is insufficient anyway for the general use case.

PR #6999, opened 2025-01-09
feat: simplify CFG after `resolve_is_unconstrained` SSA pass

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

We often use the results of `std::runtime::is_unconstrained` in control flow so it makes sense to simplify the CFG based on the results immediately as this can avoid us needing to inline functions into blocks which will eventually be pruned.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
I've changed the external repo reports to take averages as we've reduced the compilation time significantly for much of these circuits.

jfecher (changes requested):
Now that runtimes are separated pre-SSA instead of adding another pass we should remove the resolve_is_unconstrained pass entirely and simplify away the call instruction now that runtime is stored in the dfg as well

TomAFrench:
We'll also need to update `codegen_if` as  well in that case, right? It'll quite happily create if-else structures with constant conditions atm.

jfecher:
Right, odd I thought we checked constant conditions earlier for some reason. I'll make a fresh PR.

TomAFrench:
@jfecher can you approve this as I've repurposed this PR to update CI (and address a failure due to noir_rsa being updated so it compiles again)

jfecher (approved):

PR #7005, opened 2025-01-09
fix: let static_assert fail with the provided message

### LINKED ISSUE #6995: `static_assert` message ignored by `nargo test`
### Aim

Attempted to ensure that a test failed with the message from a `static_assert`:

```noir
fn main() { }

fn foo<let N: u32>() {
    std::static_assert(N == 0, "hi");
}

#[test(should_fail_with = "hi")]
fn test_main() {
    foo::<1>();
}
```

### Expected Behavior

Expected `test_main` to pass

### Bug

It appears that the error message passed to `static_assert` is ignored and `"Argument is false"` is always used:

```bash
❯ ~/.nargo/bin/nargo compile

❯ ~/.nargo/bin/nargo test
[test_static_assert_should_fail_with] Running 1 test function
[test_static_assert_should_fail_with] Testing test_main... FAIL

error: Test failed with the wrong message. 
Expected: hi 
Got: 

error: Argument is false
   ┌─ src/main.nr:10:5
   │
10 │     foo::<1>();
   │     ----------
   │
   = Call stack:
     1. src/main.nr:4:5
     2. src/main.nr:10:5


[test_static_assert_should_fail_with] Failures:
     test_main

[test_static_assert_should_fail_with] 1 test failed
```

With the following:

```noir
#[test(should_fail_with = "Argument is false")]
fn test_main() {
    foo::<1>();
}
```

The test passes:

```bash
❯ ~/.nargo/bin/nargo test
[test_static_assert_should_fail_with] Running 1 test function
[test_static_assert_should_fail_with] Testing test_main... ok
[test_static_assert_should_fail_with] 1 test passed
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.1 noirc version = 1.0.0-beta.1+8bb3908281d531160db7d7898c67fb2647792e6e (git version hash: 8bb3908281d531160db7d7898c67fb2647792e6e, is dirty: false)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #6995

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/ssa/ir/printer.rs
```diff
@@ -274,22 +274,7 @@ pub(crate) fn try_to_extract_string_from_error_payload(
...
-        .map(|fields| {
-            fields
-                .iter()
-                .map(|field| {
-                    let as_u8 = field.try_to_u64().unwrap_or_default() as u8;
-                    as_u8 as char
-                })
-                .collect()
-        })
+    (is_string_type && (values.len() == 1)).then_some(()).and_then(|()| dfg.get_string(values[0]))
```
jfecher, line 277:
```suggestion
    (is_string_type && (values.len() == 1)).then(|| dfg.get_string(values[0]))
```

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -524,6 +524,27 @@ impl DataFlowGraph {
...
+        for value_id in value_ids {
+            let field_value = self.get_numeric_constant(value_id)?;
+            let u64_value = field_value.try_to_u64()?;
+            if u64_value > 255 {
+                return None;
+            };
+            let byte = u64_value as u8;
+            if !byte.is_ascii() {
+                return None;
+            }
```
jfecher, line 542: Is the ascii restriction necessary?
I wonder if we can't collect everything to a `Vec<u8>` first then use a string conversion method on that instead.

asterite: This method is also used when printing an array value in SSA, so I guess it could happen that a random byte sequence ends up being valid UTF-8. But maybe not? I'll change it to `String::from_utf8`

asterite: That said... all our strings are ascii, so maybe this is good enough?

jfecher: I think using `String::from_utf8` should be shorter, still allow ascii, and allow us flexibility in the future when we will eventually want to have more support for utf8 strings

asterite: Done!

TomAFrench (approved):

PR #7006, opened 2025-01-09
chore: Only resolved globals monomorphization

vezenovm:
# Description

## Problem\*

Small change noticed while working on globals

## Summary\*

We should only allow accessing resolved globals in monomorphization after switching to computing all globals at compile time. 

## Documentation\*

Check one:

# PR Checklist\*


jfecher (approved):
I don't remember if this was originally included as part of the change to lazily elaborate globals and whether there were cases with comptime code where we could run into unresolved globals during monomorphization like this or not. Since it passes our tests though, and I agree we ideally should have all globals be resolved by here, it LGTM.

PR #7010, opened 2025-01-09
feat: allow deduplicating `Truncate` and `Not` instructions

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

These two instructions are not sensitive to `EnableSideEffectsIf` instructions so we can safely deduplicate them

## Documentation\*

Check one:
- [x] No documentation needed.
- [x] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [x] I have tested the changes locally.
- [ ] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #7011, opened 2025-01-09
feat: unchecked math operations in SSA

### LINKED ISSUE #6979: Separate overflow checks from arithmetic operations in SSA
As a step towards #6623 we need to have a representation of an overflow check in SSA which is separate from the arithmetic operations.

We have range check instructions already so we should be able to insert these along any arithmetic operations to maintain the current checked behaviour. This would however unlock more opportunities for instruction hoisting etc, as we can hoist the arithmetic while keeping the range checks under a conditional, etc.

asterite:
# Description

## Problem

Resolves #6979

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
```diff
@@ -16,10 +16,16 @@ use super::{
...
+    /// Addition of lhs + rhs without an overflow check.
+    UncheckedAdd,
     /// Subtraction of lhs - rhs.
     Sub,
+    /// Subtraction of lhs - rhs without an overflow check.
+    UncheckedSub,
     /// Multiplication of lhs * rhs.
     Mul,
+    /// Multiplication of lhs * rhs without an overflow check.
+    UncheckedMul,
```
TomAFrench, line 28: I was thinking we could add a boolean field to these variants rather than duplicating.

asterite: That's what I initially did but I ended up going with this option. The reasons are:
1. There are a lot of references to `BinaryOp::Add` (and the others) in SSA tests, so we'd need to change them all (this would be fine, though)
2. Conceptually we could think of them as different operations.
3. Their name would potentially match future methods `unchecked_add`, etc., like in Rust

But I don't have a strong preference, so let me know if you'd prefer a boolean.

asterite: A point in favor of using a boolean: the compiler will force me to check every existing usage, which right now I had to manually do. But it's a one-time thing, so not sure.

TomAFrench: Sure, we can always refactor this later if we decide the other option is better so no strong feelings.

asterite:
I'm taking a look at why `regression_struct_array_conditional` regressed. I think there should be no regressions in this PR.

compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
asterite: Actually, it's not a one-time thing because then every time we issue an `Add` instruction the compiler will have us think if we want to do an overflow check or not, so we'd potentially lose some optimizations opportunities if we didn't use a bool. I'll change it.

asterite: On the other hand we'll have to check every usage in this PR, which might be a good thing or not 😁 

I wonder if pushing to a slice and popping from a slice could use unchecked ops, I don't know if when popping we check if the length is zero somewhere before the subtraction.

jfecher: When popping an element from a slice we can probably use subtraction, it shouldn't go below zero. There is the possibility when pushing that we push more than 2^32 - 1 elements.

compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -2071,6 +2071,7 @@ impl<'a> Context<'a> {
                 if matches!(
                     &dfg[*instruction],
                     Instruction::Binary(Binary { operator: BinaryOp::Sub, .. })
+                        | Instruction::Binary(Binary { operator: BinaryOp::UncheckedSub, .. })
```
TomAFrench:
```suggestion
                    Instruction::Binary(Binary { operator: BinaryOp::Sub | BinaryOp::UncheckedSub, .. })
```

Should be able to do this to keep the patterns small btw.

TomAFrench:
I'm also a little concerned about some of these failures in the external repos. I'm not clear why removing the overflow check would cause issues like that.

compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
```diff
@@ -15,11 +15,11 @@ use super::{
...
-    Add,
+    Add { unchecked: bool },
     /// Subtraction of lhs - rhs.
-    Sub,
+    Sub { unchecked: bool },
     /// Multiplication of lhs * rhs.
-    Mul,
+    Mul { unchecked: bool },
     /// Division of lhs / rhs.
     Div,
```
TomAFrench, line 24: We can leave out of scope for this PR but just noting that we can have this unchecked by removing the zero denominator check.

asterite:
> I'm also a little concerned about some of these failures in the external repos. I'm not clear why removing the overflow check would cause issues like that.

It seems it's the same error you got in https://github.com/noir-lang/noir/pull/7010 so maybe CI in Aztec-Packages is broken.

TomAFrench:
Ah yeah, the protocol circuits team have updated the circuits but not the inputs which is why execution is failing.

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -218,22 +218,18 @@ impl Context<'_> {
...
                 let not_b = self.insert_cast(not_b, NumericType::NativeField);
-                // TODO: should this be unchecked?
-                let r1 = self.insert_binary(a, BinaryOp::Mul { unchecked: false }, b);
-                // TODO: should this be unchecked?
-                let r2 = self.insert_binary(r_squared, BinaryOp::Mul { unchecked: false }, not_b);
-                // TODO: should this be unchecked?
-                r = self.insert_binary(r1, BinaryOp::Add { unchecked: false }, r2);
+                let r1 = self.insert_binary(a, BinaryOp::Mul { unchecked: true }, b);
+                let r2 = self.insert_binary(r_squared, BinaryOp::Mul { unchecked: true }, not_b);
+                r = self.insert_binary(r1, BinaryOp::Add { unchecked: true }, r2);
```
TomAFrench: We originally did all these casts to field elements as a janky method of turning off overflow checks. We may be able to improve this to work on the underlying types.

asterite:
Oops, I didn't realize you were handling some TODOs too (I asked for help but in the end I got to understand most of them 😊). There's only one TODO remaining. I think it shouldn't overflow as it's subtracting one to `half_width`, but I wanted to check with you just in case.

TomAFrench:
Yeah, I think that should be safe. We're subtracting an upcasted boolean from a large constant so it can't underflow afaict.

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -169,14 +177,24 @@ impl Context<'_> {
...
             let shifted_complement = self.insert_binary(one_complement, BinaryOp::Div, pow);
             // Convert back to 2-complement representation if operand is negative
             let lhs_sign_as_int = self.insert_cast(lhs_sign, lhs_typ);
-            let shifted = self.insert_binary(shifted_complement, BinaryOp::Sub, lhs_sign_as_int);
+            // TODO: should this be unchecked?
+            let shifted = self.insert_binary(
+                shifted_complement,
+                BinaryOp::Sub { unchecked: false },
+                lhs_sign_as_int,
+            );
```
asterite, line 197: Oh, I think I did the merge wrong. This is the TODO which I wasn't sure about. This is all inside `insert_wrapping_shift_left` so I guess no overflow should happen here, but I'm not sure.

asterite:
I wonder if for field operations it makes sense to use `unchecked: true`. It just adds noise to the final SSA, and it might make things a bit more confusing ("why is this unchecked_mul over fields if they can't overflow?"). I'm thinking either changing those back to `false` or, given that we already check the operand types when inserting binary instructions, to turn them to `unchecked: false` if the operands are fields. Thoughts?

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
TomAFrench: We should at least be able to create a testcase for this scenario.

asterite:
> I wonder if for field operations it makes sense to use unchecked: true

I just checked and it at least doesn't make a lot of difference in the SSA tests, so for now using `unchecked: true` for fields is probably fine.

TomAFrench:
> why is this unchecked_mul over fields if they can't overflow?

I think `mul` would be more confusing as if it's sprinkled in amongst `unchecked_mul` it gives the impression that we're checking for if we overflow the field modulus, etc.

TomAFrench (approved):
LGTM

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
asterite: Do you want me to capture a follow-up issue for this?

asterite:
> I think mul would be more confusing as if it's sprinkled in amongst unchecked_mul it gives the impression that we're checking for if we overflow the field modulus, etc.

I'll make a change in this PR soon then 👍

asterite:
Cool, I moved the check that checks that binary operands have the same type to when an instruction is created, right before inserting it, and it triggers some bugs in noir-bignum, and I guess noir-rsa too. I'll try to fix those here, it's probably an easy thing to fix.

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -174,6 +175,44 @@ impl DataFlowGraph {
...
+                    operator
+                }
+            } else {
+                operator
+            };
+
+            Instruction::Binary(Binary { operator, ..binary })
+        } else {
+            instruction_data
+        };
```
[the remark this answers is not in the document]
asterite: On the other hand, it's true that simplify will already return some of the operands based on some conditions, and if there's a mistake before that then we wouldn't notice. Hmm... But then `simplify` returns `SimplifyResult` so we can't change the instruction. I'll see how it can be done.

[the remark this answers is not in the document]
TomAFrench: > But then simplify returns SimplifyResult so we can't change the instruction.

We can use `SimplifyResult::SimplifyToInstruction` and pass back an updated instruction to insert.

PR #7017, opened 2025-01-10
chore: disallow inserting ACIR-only instructions into brillig functions

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

We currently disallow inserting brillig only instructions into ACIR so this PR extends this to cover the reverse. I've also added a logging statement so we can see where we're trying to insert these invalid instructions.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -228,6 +238,7 @@ impl DataFlowGraph {
         call_stack: CallStackId,
     ) -> InsertInstructionResult {
         if !self.is_handled_by_runtime(&instruction) {
+            warn!("Attempted to insert instruction not handled by runtime: {instruction:?}");
```
aakoshh, line 241: By the way I inserted the same call [here](https://github.com/noir-lang/noir/blob/33cad9b4543825bb275a936bdfe320e9da804b54/compiler/noirc_evaluator/src/ssa/ir/dfg.rs#L217) when I saw that `insert_instruction_without_simplification` became public again to allow parsing SSA strings as-is, but in retrospect that's exactly why we might _not_ want to remove even these instructions, even if they would cause a panic later on: if the SSA has them, go ahead. So if you agree that this check should only be here, then we could remove the other instance.

jfecher (approved):

TomAFrench:
I'm not seeing those external repo checks failing in master so it may be that these instructions are necessary currently. Would like to investigate a bit more.

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
TomAFrench: I've updated `insert_instruction_and_results_without_simplification` to panic if we hit this condition.

PR #7018, opened 2025-01-10
feat(cli): Add CLI option to filter by contract function name

aakoshh:
# Description

## Problem\*

Resolves having to hack around the code to restrict which contract function to show.

## Summary\*

Adds a `--show-contract-fn` CLI option to restrict the effects of `--show-ssa`, `--show-ssa-pass` and `--print-acir` to a single function in a contract, ie. a single circuit. 

Example:
```
cargo run --release -p nargo_cli -- \
  --program-dir $PROGRAM_DIR compile \
  --package token_contract  \
  --print-acir \
  --show-ssa-pass "Remove IfElse" \
  --show-contract-fn public_dispatch 
```

Note that `--show-ssa-pass-name` has been shortened to `--show-ssa-pass`.

## Additional Context

Often when investigating the size increase of a contract circuit such as the `public_dispatch` function of the `Token` contract in https://github.com/AztecProtocol/aztec-packages/pull/11138#issuecomment-2582218784 I want to see the SSA, but currently running `compile` would show the SSA of 36 different functions, which, even when restricted to a single pass, results in 14MB of text in this example.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_driver/src/lib.rs
+++ b/compiler/noirc_driver/src/lib.rs
@@ -73,7 +73,7 @@ pub struct CompileOptions {
     /// Only show SSA passes whose name contains the provided string.
     /// This setting takes precedence over `show_ssa` if it's not empty.
     #[arg(long, hide = true)]
-    pub show_ssa_pass_name: Option<String>,
+    pub show_ssa_pass: Option<String>,
 
     /// Only show the SSA and ACIR for the contract function with a given name.
     #[arg(long, hide = true)]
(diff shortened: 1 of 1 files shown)


compiler/noirc_driver/src/lib.rs
```diff
@@ -75,6 +75,10 @@ pub struct CompileOptions {
     #[arg(long, hide = true)]
     pub show_ssa_pass_name: Option<String>,
 
+    /// Only show the SSA and ACIR for the contract function with a given name.
+    #[arg(long, hide = true)]
+    pub show_contract_function_name: Option<String>,
```
michaeljklein, line 80: This is a bit long: perhaps `show_contract_fn`? Also, from the name of the argument, I would expect it to show the names of contract functions, not filter by name.

michaeljklein (approved):
Small comment r.e. naming, otherwise LGTM

compiler/noirc_driver/src/lib.rs
aakoshh: I know what you mean, I had the same thought. I called it like this because of the `--show-ssa-pass-name` option.

Do you think we should remove `-name` from both?

aakoshh: Shortened both, cc @asterite

PR #7020, opened 2025-01-10
chore: delete docs for versions which aren't used

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

These aren't visible to end users so we can remove these versions.

## Documentation\*

Check one:

# PR Checklist\*


jzaki:
Do we explicitly want only 0.36.0 in that menu? If 0.39.0 was a sensible release, should 0.36.0 be deleted instead and 0.39.0 linked in the menu?

Versions are configurable via [versions.json](https://docusaurus.io/docs/versioning) (which we currently [ignore](https://github.com/noir-lang/noir/blob/master/docs/.gitignore#L27)), so presumably are manually set when deployed.

jzaki (changes requested):
~Would this help guide which versions to delete? https://github.com/noir-lang/noir/pull/7022~

jzaki (approved):

PR #7021, opened 2025-01-10
feat(brillig): SSA globals code gen

### LINKED ISSUE #6801: SSA: Global variables

vezenovm:
# Description

## Problem\*

Resolves #6801 

## Summary\*

Implements globals from SSA in Brillig gen. We create a new memory space `GlobalSpace` at compile time. Globals are declared in the global space and the linker is then used to connect these with our SSA functions. 

The global brillig variables that are allocated are then passed to every `BrilligBlock`. When converting SSA values for a function's block we check whether we have a global and if so, look for the appropriate register that keys into the appropriate slot in the global space. 

## Additional Context

Some slight regressions are expected. Some reasons for this are laid out here https://github.com/noir-lang/noir/pull/7021#issuecomment-2593932994. 

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
The regressions are most likely due to Value::Global still being treated as a non-constant in the DFG. I will update this separately and keep this a draft for now.

vezenovm:
> The remaining regressions look to be due to dead globals not being removed. Will have a PR shortly.

I tried building off master originally but it was a lot easier to test, building off this branch. The work can be seen here in #7081.

> **diamond_deps_0_inliner_zero**	+2 ❌	**+7.69%**
> **modules_inliner_zero**	+2 ❌	**+7.69%**
> **modules_more_inliner_zero**	+2 ❌	**+7.69%**
> **no_predicates_basic_inliner_zero**	+2 ❌	**+7.69%**
> **traits_in_crates_1_inliner_zero**	+2 ❌	**+7.69%**
> **traits_in_crates_2_inliner_zero**	+2 ❌	**+7.69%**
> **fold_basic_inliner_zero**	+2 ❌	**+7.69%**

The globals DIE in #7081 is also not going to get rid of these extra instructions we see here. `add_globals_init_instruction()` makes an external call to our globals initialization.  Without any globals to declare we will just execute a return instruction. Thus, we see an overhead of 2 instructions, a call and a return. We can get rid of this overhead, but it would require some larger changes that felt not worth holding up this PR on its own and could come in a follow-up.

vezenovm:
I look to be getting failures post https://github.com/AztecProtocol/aztec-packages/pull/10605. Working on resolving it now.

compiler/noirc_evaluator/src/ssa/parser/into_ssa.rs
```diff
@@ -345,6 +347,8 @@ impl Translator {
         // that the SSA we parsed was printed by the `SsaBuilder`, which normalizes
         // before each print.
         ssa.normalize_ids();
+        // Does not matter what ID we sue here.
```
vezenovm, line 350:
```suggestion
        // Does not matter what ID we use here.
```

tooling/nargo/src/errors.rs
```diff
@@ -237,6 +237,7 @@ pub fn try_to_diagnose_runtime_error(
         }
         _ => return None,
     };
+
```
vezenovm, line 240:
```suggestion
```

TomAFrench:
> I look to be getting failures post https://github.com/AztecProtocol/aztec-packages/pull/10605.

I assume that this is from the roots global. This isn't used in brillig code currently so I imagine this would be solved by the die pass PR

That said, as we know all of the globals at compile time we should be able to expand the space allocated as necessary. If so can you create a follow up issue?

vezenovm:
> That said, as we know all of the globals at compile time we should be able to expand the space allocated as necessary. If so can you create a follow up issue?

Yeah I'll make an issue for max global space to be determined by the program rather than a hardcoded value.

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -32,28 +32,38 @@ use super::brillig_fn::FunctionContext;
 use super::constant_allocation::InstructionLocation;
 
 /// Generate the compilation artifacts for compiling a function into brillig bytecode.
-pub(crate) struct BrilligBlock<'block> {
-    pub(crate) function_context: &'block mut FunctionContext,
+pub(crate) struct BrilligBlock<'block, 'global, Registers: RegisterAllocator> {
```
TomAFrench, line 35: Do we need two globals here? I'd expect that the globals would outlive `'block` anyway so we can just reuse that lifetime.

vezenovm: Huh, not sure why I put them on the `FunctionContext`, good point. I think with how I have it now with the globals inside of `FunctionContext` I cannot avoid the second lifetime, but I will move it out of there and put it directly on `BrilligBlock`.

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -24,7 +24,25 @@ impl Ssa {
...
+        // Check which globals are used across all functions
+        for (id, value) in self.globals.dfg.values_rev_iter() {
+            if used_global_values.contains(&id) {
+                if let Value::Instruction { instruction, .. } = &value {
+                    let instruction = &self.globals.dfg[*instruction];
+                    instruction.for_each_value(|value_id| {
+                        used_global_values.insert(value_id);
+                    });
+                }
+            }
```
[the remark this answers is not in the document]
TomAFrench: Ah, never mind. I'm being dumb.

compiler/noirc_evaluator/src/ssa/ir/map.rs
```diff
@@ -194,6 +194,11 @@ impl<T> DenseMap<T> {
         let ids_iter = (0..self.storage.len() as u32).map(|idx| Id::new(idx));
         ids_iter.zip(self.storage.iter())
     }
+
+    pub(crate) fn rev_iter(&self) -> impl ExactSizeIterator<Item = (Id<T>, &T)> {
+        let ids_iter = (0..self.storage.len() as u32).map(|idx| Id::new(idx));
+        ids_iter.zip(self.storage.iter()).rev()
+    }
```
TomAFrench, line 201: Equivalent to the below no?
```suggestion

    pub(crate) fn rev_iter(&self) -> impl ExactSizeIterator<Item = (Id<T>, &T)> {
        self.iter().rev()
    }
```

Which raises the question of do we need this method?

vezenovm: I tried the same, but `DoubleEndedIterator` is not implemented for `ExactSizeIterator` which is needed for `rev`. I will leave a comment

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
vezenovm: Done, I now put the globals directly on the block.

compiler/noirc_evaluator/src/ssa/ir/map.rs
jfecher: Could `self.iter()` and thus DenseMap be changed to return a DoubleEndedIterator instead then?

vezenovm: Yeah I could do that as well. I didn't change the type as I wasn't sure if we wanted to keep `ExactSizeIterator`.

vezenovm: It doesn't look like it changes much, so I can update the type if we would like

vezenovm: I switched to using `DoubleEndedIterator` for code simplicity.

vezenovm:
> That said, as we know all of the globals at compile time we should be able to expand the space allocated as necessary. If so can you create a follow up issue?

https://github.com/noir-lang/noir/issues/7107

TomAFrench (approved):

Issue #6571, opened 2024-11-21
Traits with associated types can't be used in `where` clauses


michaeljklein:
### Aim

Given a simple trait with an associated type:

```noir
trait Foo {
    type Bar;
}
```

I attempted to require `Foo` in a `where` clause in three ways:

1.

```noir
// using "where (): Foo" fails with the same error,
// even if the following impl is present:
// impl Foo for () {
//     type Bar = ();
// }
impl<T> Foo for Option<T> where T: Foo {
    type Bar = T;
}
```

2.

```noir
trait Baz { }

impl<T> Baz for Option<T> where T: Foo { }
```

3.

```noir
fn foo<T>(x: T) where T: Foo { }
```

### Expected Behavior

Expected the above cases to compile successfully

### Bug

Each fails with `'Foo' is missing the associated type 'Bar'`:

1.

```bash
error: `Foo` is missing the associated type `Bar`
  ┌─ src/main.nr:5:36
  │
5 │ impl<T> Foo for Option<T> where T: Foo {
  │                                    ---
  │
```

2.

```bash
error: `Foo` is missing the associated type `Bar`
  ┌─ src/main.nr:7:36
  │
7 │ impl<T> Baz for Option<T> where T: Foo { }
  │                                    ---
  │
```

3.

```bash
error: `Foo` is missing the associated type `Bar`
  ┌─ src/main.nr:5:27
  │
5 │ fn foo<T>(x: T) where (): Foo { }
  │                           ---
  │
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Nargo Version

nargo version = 0.39.0 noirc version = 0.39.0+68c32b4ffd9b069fe4b119327dbf4018c17ab9d4 (git version hash: 68c32b4ffd9b069fe4b119327dbf4018c17ab9d4, is dirty: false)

### Would you like to submit a PR for this Issue?

None

jfecher:
This is expected currently while the feature is incomplete - the only thing we've shipped so far is an explicit version where associated types must always be specified. So I'd expect the following to work:

```noir
fn foo<T, U>(x: T) where T: Foo<Bar = U> { }
```

(we should keep this issue until associated types are feature complete though)

michaeljklein:
Ok: just confirmed that an explicit associated type in the `where` clause works 👍

Issue #7024, opened 2025-01-10
Trait definition where clause doesn't support associated type constraints


jfecher:
### Aim

```noir
trait Baz<T> where
    T: Foo,
    <T as Foo>::Bar: Eq {}

impl Baz<u64> for u32 {}
```

### Expected Behavior

The program to compile without error (assuming `impl Foo for u64` is in scope)

### Bug

The following errors occur when these associated type constraints are used in the where clause of a trait definition specifically:
```
error: The trait bound `u64: Foo<Bar = _>` is not satisfied
   ┌─ src/main.nr:55:8
   │
55 │     T: Foo,
   │        --- required by this bound in `Baz
   ·
58 │ impl Baz<u64> for u32 {}
   │                   --- The trait `Foo<Bar = _>` is not implemented for `u64
   │

error: The trait bound `_: Eq` is not satisfied
   ┌─ src/main.nr:56:22
   │
56 │     <T as Foo>::Bar: Eq {}
   │                      -- required by this bound in `Baz
57 │ 
58 │ impl Baz<u64> for u32 {}
   │                   --- The trait `Eq` is not implemented for `_
   │
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

asterite:
Full failing code:

```noir
trait MyEq {}

impl MyEq for i32 {}

trait Foo {
    type Bar;
}

impl Foo for u64 {
    type Bar = i32;
}

trait Baz<T>
where
    T: Foo,
    <T as Foo>::Bar: MyEq,
{}

impl Baz<u64> for u32 {}
```

The error:

```
error: The trait bound `<T as Foo>::Bar: MyEq` is not satisfied
   ┌─ src/main.nr:16:22
   │
16 │     <T as Foo>::Bar: MyEq,
   │                      ---- required by this bound in `Baz`
   ·
19 │ impl Baz<u64> for u32 {}
   │                   --- The trait `MyEq` is not implemented for `<T as Foo>::Bar`
```

Issue #6811, opened 2024-12-13
Multiple impls of the same trait & object type allowed when the associated types or constants differ


jfecher:
### Aim

```noir
trait Foo {
    let N: u32;

    fn get_array(self) -> [Self; N];
}

impl Foo for Field {
    let N = 2;

    fn get_array(self) -> [Self; Self::N] {
        [1, 2]
    }
}

impl Foo for Field {
    let N = 3;

    fn get_array(self) -> [Self; Self::N] {
        [1, 2, 3]
    }
}

fn main(x: Field) {
    x.get_array();
}
```

### Expected Behavior

Expected an error that there are two `impl Foo for Field`

### Bug

No error - the program compiles without any related warnings

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #7026, opened 2025-01-10
feat: Allow associated types to be ellided from trait constraints

### LINKED ISSUE #7039: Incorrect private warning
### Aim

In the associated_types_implicit example:
```noir
trait Foo {
    type Bar;

    fn foo(self) -> Self::Bar;
}

impl Foo for u64 {
    type Bar = u8;

    fn foo(self) -> Self::Bar {
        self as u8
    }
}

fn main() {
    let three: u64 = 3;
    call_foo(three);

    let x: Option<Option<u64>> = Option::some(Option::some(0));
    let x_foo = x.foo(); // private warning here
    assert_eq(x_foo, x_foo);
    assert_eq(x.foo(), 0 as u8); // private warning here
}

fn call_foo<T>(x: T) where
    T: Foo,
    <T as Foo>::Bar: Eq
{
    let y = x.foo();
    assert_eq(y, y);
}

impl<T> Foo for Option<T> where
    T: Foo,
    <T as Foo>::Bar: Eq
{
    type Bar = <T as Foo>::Bar;

    fn foo(self) -> Self::Bar {
        self.unwrap().foo()
    }
}
```

### Expected Behavior

The program to compile without warnings

### Bug

We get two warnings, one for each `x.foo()` call in `main` that read: `foo is private and not visible from the current module` despite `foo` and `Foo` being defined in the same module `main` is in.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6571 (but does _not_ implement https://github.com/noir-lang/noir/issues/6811)
Resolves #7039

## Summary\*

Allows users to elide associated types from trait constraints. E.g. given a trait `Iterator` with an associated type `Elem`, we now allow constraints of the form `T: Iterator` in addition to the previous `T: Iterator<Elem = U>`.

## Additional Context

Implementing this usually means finding where to insert implicit type variables for the associated types, this location is different for each item (functions, trait impls, trait definitions).

Because of constraints like `where T: Foo, <T as Foo>::Bar: Eq`, we have to introduce these assumed trait impls to scope as we go, otherwise we'll get an error that there's no impl for `T: Foo` when resolving the type `<T as Foo>::Bar`. Since I've chosen to do this in order, constraints like this rely on their ordering so `where <T as Foo>::Bar: Eq, T: Foo` will give an error.

This PR does _not_ implement this feature for trait definitions (see https://github.com/noir-lang/noir/issues/7024). I had trouble getting this working and believe this to be the rarest use case (compared to constraints on functions & trait impls) so I'm opting to leave this for a later PR.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -733,24 +734,98 @@ impl<'context> Elaborator<'context> {
...
             .iter()
-            .filter_map(|constraint| self.resolve_trait_constraint(constraint))
+            .filter_map(|constraint| {
+                self.resolve_trait_constraint(constraint, add_constraints_to_scope)
+            })
+            .collect()
+    }
+
+    /// Expands any traits in a where clause to mention all associated types if they were
+    /// ellided by the user. See `add_missing_named_generics` for more  detail.
```
michaeljklein, line 751: typo
```suggestion
    /// elided by the user. See `add_missing_named_generics` for more  detail.
```

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -733,24 +734,98 @@ impl<'context> Elaborator<'context> {
         None
     }
 
-    /// TODO: This is currently only respected for generic free functions
-    /// there's a bunch of other places where trait constraints can pop up
     fn resolve_trait_constraints(
         &mut self,
         where_clause: &[UnresolvedTraitConstraint],
+        add_constraints_to_scope: bool,
```
asterite, line 740: It seems all calls to this method set this to `true`.

compiler/noirc_frontend/src/elaborator/mod.rs
asterite, line 751:
```suggestion
    /// elided by the user. See `add_missing_named_generics` for more  detail.
```

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -733,24 +734,98 @@ impl<'context> Elaborator<'context> {
...
+    /// `fn foo<T, A, B>() where T: Foo<Bar = A, Baz = B> { ... }`
+    fn add_missing_named_generics(&mut self, bound: &mut TraitBound) -> Vec<ResolvedGeneric> {
+        let mut added_generics = Vec::new();
+
+        let Some(the_trait) = self.lookup_trait_or_error(bound.trait_path.clone()) else {
+            return added_generics;
+        };
+
+        if the_trait.associated_types.len() > bound.trait_generics.named_args.len() {
+            for associated_type in the_trait.associated_types.clone() {
```
michaeljklein, line 784: Would it make sense to drop the `clone` here because `associated_type` is always cloned or used `as_ref` below?

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -733,24 +734,98 @@ impl<'context> Elaborator<'context> {
...
+                if !bound
+                    .trait_generics
+                    .named_args
+                    .iter()
+                    .any(|(name, _)| name.0.contents == *associated_type.name.as_ref())
+                {
+                    // This generic isn't contained in the bound's named arguments,
+                    // so add it by creating a fresh type variable.
+                    let new_generic_id = self.interner.next_type_variable_id();
+                    let type_var = TypeVariable::unbound(new_generic_id, Kind::Normal);
```
michaeljklein, line 794:
```suggestion
                    let type_var = self.interner.type_variable_with_kind(Kind::Normal);
```

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -733,24 +734,98 @@ impl<'context> Elaborator<'context> {
...
+
+                    let span = bound.trait_path.span;
+                    let no_name = Rc::new("_".to_string());
```
michaeljklein, line 797: I'd prefer for this to be more distinct from a potentially user-entered label if possible, e.g.
```suggestion
                    let name = Rc::new(&format("_{}", name.0.contents));
```

michaeljklein (changes requested):

compiler/noirc_frontend/src/elaborator/mod.rs
michaeljklein: duplicate of https://github.com/noir-lang/noir/pull/7026#discussion_r1911117327

compiler/noirc_frontend/src/elaborator/mod.rs
jfecher: I thought showing `T: Foo<Bar = _>` was more readable to the user than `T: Foo<Bar = Bar>` or `T: Foo<Bar = _Bar>` since neither `Bar` nor `_Bar` are actual types in scope.

I haven't tested it much but if this type leaks into other errors I could see it being easier to read if we used `Bar` instead. E.g. if you get an error for `No impl Eq for _` versus `No impl Eq for Bar`. Haven't tested how easy that is to trigger though.

michaeljklein:
> Don't error anymore if the user forgets to specify an associated type in a trait impl (not a constraint, the actual impl)

I think that might be this line? https://github.com/noir-lang/noir/pull/7026/files#diff-ac857943134bba360eee936142f9043704ff127f0a9a188e6feba623ae68d93aL386

Was wondering when I saw it how we could be sure that the error is thrown elsewhere and it sounds similar.

jfecher:
@michaeljklein it was, yeah. That's why I had to add "allow_implicit_named_args." The short of it was that in most cases we do allow them but in trait impls we also use that method for checking the trait being implemented has all its generics and associated types, so in that case we want to disallow them.

With that fixed, there's still one more test to fix which is that we issue the "no trait Foo in scope" error twice now in one case.

compiler/noirc_frontend/src/elaborator/mod.rs
jfecher: associated_types needs to be cloned here since `self.interner` is used mutably in the loop body

compiler/noirc_frontend/src/elaborator/mod.rs
jfecher: That returns a `Type` instead of a `TypeVariable` which is needed for the `NamedGeneric` below

compiler/noirc_frontend/src/elaborator/mod.rs
jfecher: I was able to trigger this with
```noir
fn baz<T>(x: T) where T: Foo {
    let y = x.foo();
    assert_eq(y, y);
}
```
"No matching impl for _: Eq"
so I changed it to use the associated type's name instead which should hopefully be more clear where the type is coming from. It also means users will see something like `Foo<Bar = Bar>` often, but we can try to change the display in the future to leave that out somehow.

jfecher:
@michaeljklein ready for re-review

jfecher:
Currently blocked by https://github.com/noir-lang/noir/issues/7039

michaeljklein (approved):
LGTM

Issue #6956, opened 2025-01-06
Add LSP action to automatically import a trait when using one of its methods


TomAFrench:
In https://github.com/noir-lang/noir/pull/6882 we added a warning for situations where we're using a trait but it's not in scope. For the case where there's only a single candidate trait we can then automatically import it. This would make it much easier to update any legacy codebases (and generally be more ergonomic on new ones).

Rust also allows selecting from multiple candidate traits which would also be nice but can be done separately.

asterite:
Relatedly, when typing `x.` and the completion is a trait method, the trait method is inserted but not the corresponding import (because of how it used to work). So now we'd also need to add the corresponding import. I'll try to do both things in the same PR as they are kind of related.

asterite:
It seems this was eventually implemented.

PR #7027, opened 2025-01-10
fix(lsp): suggest all possible trait methods, but only visible ones

asterite:
# Description

## Problem

Towards #6956

## Summary

I noticed a couple of issues in how we suggest trait methods in LSP:
1. We don't show all of the trait methods if more than one could be called (only one, the first one). That is, if there are multiple traits with the same method name and none is currently imported, all of those should be suggested.
2. Non-visible trait methods were suggested

Solving these two things required some changes on the compiler side, outside of just LSP, and to avoid eventually creating a giant PR that solves all of #6956 I decided to split just this into a smaller PR.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/mod.rs
+++ b/compiler/noirc_frontend/src/elaborator/mod.rs
@@ -1110,10 +1110,19 @@ impl<'context> Elaborator<'context> {
         self.check_parent_traits_are_implemented(&trait_impl);
         self.remove_trait_impl_assumed_trait_implementations(trait_impl.impl_id);
 
+        let trait_id =
+            trait_impl.trait_id.expect("Trait impl trait ID should have been set at this point");
+        let trait_ = self.interner.get_trait(trait_id);
+        let trait_visibility = trait_.visibility;
+
         for (module, function, _) in &trait_impl.methods.functions {
             self.local_module = *module;
             let errors = check_trait_impl_method_matches_declaration(self.interner, *function);
             self.errors.extend(errors.into_iter().map(|error| (error.into(), self.file)));
+
+            // A trait impl method has the same visibility as its trait
+            let modifiers = self.interner.function_modifiers_mut(function);
+            modifiers.visibility = trait_visibility;
         }
 
         self.elaborate_functions(trait_impl.methods);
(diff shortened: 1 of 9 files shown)


compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1098,6 +1103,23 @@ impl<'context> Elaborator<'context> {
         }
     }
 
+    fn set_trait_impl_functions_visibility(&mut self, trait_impl: &UnresolvedTraitImpl) {
+        let trait_visibility = if let Some(trait_id) = trait_impl.trait_id {
+            let trait_ = self.interner.get_trait(trait_id);
+            Some(trait_.visibility)
+        } else {
+            None
+        };
```
jfecher, line 1112:
```suggestion
        let Some(trait_id) = trait_impl.trait_id else { return; };
        let trait_visibility = self.interner.get_trait(trait_id).visibility;
```
Had this as a `map` originally but figured it'd be better to return early instead of go through the next loop if `trait_visibility` is None.

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -346,6 +346,11 @@ impl<'context> Elaborator<'context> {
             &items.module_attributes,
         );
 
+        // Make sure trait impl functions have the same visibility as their traits
+        for trait_impl in &items.trait_impls {
+            self.set_trait_impl_functions_visibility(trait_impl);
+        }
```
jfecher, line 352: Does this need to be a separate loop? Can we move it into `collect_trait_impl`?

compiler/noirc_frontend/src/hir_def/traits.rs
```diff
@@ -148,18 +149,6 @@ impl PartialEq for Trait {
...
-        self.methods = methods;
-    }
-
-    pub fn set_trait_bounds(&mut self, trait_bounds: Vec<ResolvedTraitBound>) {
-        self.trait_bounds = trait_bounds;
-    }
-
-    pub fn set_where_clause(&mut self, where_clause: Vec<TraitConstraint>) {
-        self.where_clause = where_clause;
-    }
```
jfecher, line 161: I don't really mind but why remove these methods?

asterite: Oh, it's because now there's `visibility` so I initially created a mutable method for it but then I noticed it's `pub` (I made it `pub` like the others) and then noticed all of them are pub so I didn't understand why we needed mutable methods if the fields are already mutable. But maybe it's to make it clear they are mutable, not sure. I'll put them back.

compiler/noirc_frontend/src/elaborator/mod.rs
asterite: Done! I didn't realize we had two loops for trait impls.

PR #7007, opened 2025-01-09
chore: Add short circuit in ssa-gen for known if conditions

jfecher:
# Description

## Problem\*

Resolves 

## Summary\*

Follow-up for https://github.com/noir-lang/noir/pull/6999#issuecomment-2580342046

## Documentation\*

Check one:
- [x] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #7030, opened 2025-01-11
feat: skip codegen of zero iteration loops

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

In the vein of #7007, we can skip codegen for any for loops where we know that we're going to eventually simplify it out of the CFG. This helps mostly when numeric generics are used in for loop bounds.

## Documentation\*

Check one:

# PR Checklist\*


aakoshh (approved):
I can't find any unit tests for SSA generation, but it looks correct to me 👍

jfecher:
It could be useful to expand this check to loops with only 1 iteration to skip the loop header and just compile the body (remembering to still define the induction variable).

PR #7031, opened 2025-01-11
feat: avoid generating a new witness when checking if linear expression is zero

TomAFrench:
…sion is zero

# Description

## Problem\*

Resolves 

## Summary\*

We're currently creating a new witness when checking whether linear expressions are equal to zero. This is unnecessary as these will neatly fit into the following constraints (at least for direct scaling of a single witness)

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/acir/generated_acir.rs
+++ b/compiler/noirc_evaluator/src/acir/generated_acir.rs
@@ -535,8 +535,8 @@ impl<F: AcirField> GeneratedAcir<F> {
     /// This is easily observed: `y = 1 - t * 0`
     /// Now since `y` is one, this means that `t` needs to be zero, or else `y * t == 0` will fail.
     fn is_zero(&mut self, t_expr: Expression<F>) -> Witness {
-        // We're checking for equality with zero so we can negate the expression without changing the result.
-        // This is useful as it will sometimes allow us to simplify an expression down to a witness.
+        // We're going to be multiplying this expression by a two different witnesses in a second so we want to
+        // ensure that this expression only contains a single witness. We can tolerate a coefficients and constant terms however.
         let linear_t = if t_expr.is_degree_one_univariate() {
             t_expr
         } else {
--- a/compiler/noirc_evaluator/src/acir/generated_acir.rs
+++ b/compiler/noirc_evaluator/src/acir/generated_acir.rs
@@ -535,8 +535,8 @@ impl<F: AcirField> GeneratedAcir<F> {
     /// This is easily observed: `y = 1 - t * 0`
     /// Now since `y` is one, this means that `t` needs to be zero, or else `y * t == 0` will fail.
     fn is_zero(&mut self, t_expr: Expression<F>) -> Witness {
-        // We're going to be multiplying this expression by a two different witnesses in a second so we want to
-        // ensure that this expression only contains a single witness. We can tolerate a coefficients and constant terms however.
+        // We're going to be multiplying this expression by two different witnesses in a second so we want to
+        // ensure that this expression only contains a single witness. We can tolerate coefficients and constant terms however.
         let linear_t = if t_expr.is_degree_one_univariate() {
             t_expr
         } else {
--- a/compiler/noirc_evaluator/src/acir/generated_acir.rs
+++ b/compiler/noirc_evaluator/src/acir/generated_acir.rs
@@ -479,7 +479,7 @@ impl<F: AcirField> GeneratedAcir<F> {
     pub(crate) fn is_equal(&mut self, lhs: &Expression<F>, rhs: &Expression<F>) -> Witness {
         let t = lhs - rhs;
 
-        self.is_zero(&t)
+        self.is_zero(t)
     }
 
     /// Returns a `Witness` that is constrained to be:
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/acir/generated_acir.rs
```diff
@@ -534,36 +534,32 @@ impl<F: AcirField> GeneratedAcir<F> {
     /// By setting `z` to be `0`, we can make `y` equal to `1`.
     /// This is easily observed: `y = 1 - t * 0`
     /// Now since `y` is one, this means that `t` needs to be zero, or else `y * t == 0` will fail.
-    fn is_zero(&mut self, t_expr: &Expression<F>) -> Witness {
+    fn is_zero(&mut self, t_expr: Expression<F>) -> Witness {
         // We're checking for equality with zero so we can negate the expression without changing the result.
         // This is useful as it will sometimes allow us to simplify an expression down to a witness.
```
TomAFrench, line 539:
```suggestion
        // We're going to be multiplying this expression by a two different witnesses in a second so we want to
        // ensure that this expression only contains a single witness. We can tolerate a coefficients and constant terms however.
```

compiler/noirc_evaluator/src/acir/generated_acir.rs
```diff
@@ -534,36 +534,32 @@ impl<F: AcirField> GeneratedAcir<F> {
...
+    fn is_zero(&mut self, t_expr: Expression<F>) -> Witness {
+        // We're going to be multiplying this expression by two different witnesses in a second so we want to
+        // ensure that this expression only contains a single witness. We can tolerate coefficients and constant terms however.
+        let linear_t = if t_expr.is_degree_one_univariate() {
+            t_expr
         } else {
-            let negated_expr = t_expr * -F::one();
-            self.get_or_create_witness(&negated_expr)
+            Expression::<F>::from(self.get_or_create_witness(&t_expr))
         };
```
aakoshh, line 544: I'm trying to understand where where exactly the generation of a new witness is avoided. 

As I see it `t_expr.to_witness()` just unwrapped the witness from `t_expr` if it was `is_degree_one_univariate`, but if it's not, then now instead of creating a witness for `-t_expr`, we create a witness for `t_expr` itself. So the number of witnesses generated so far is the same, isn't it?

compiler/noirc_evaluator/src/acir/generated_acir.rs
```diff
@@ -534,36 +534,32 @@ impl<F: AcirField> GeneratedAcir<F> {
...
         // Add constraint y == 1 - tz => y + tz - 1 == 0
-        let y_is_boolean_constraint = Expression {
-            mul_terms: vec![(F::one(), t_witness, z)],
-            linear_combinations: vec![(F::one(), y)],
-            q_c: -F::one(),
-        };
+        let mut y_is_boolean_constraint =
+            (&z_expr * &linear_t).expect("multiplying two linear expressions");
+        y_is_boolean_constraint.push_addition_term(F::one(), y);
+        let y_is_boolean_constraint = y_is_boolean_constraint - F::one();
```
aakoshh, line 558: Isn't there a way where this could be written as a single expression, without having to break it down into mutable steps, then a redefinition?

compiler/noirc_evaluator/src/acir/generated_acir.rs
```diff
@@ -534,36 +534,32 @@ impl<F: AcirField> GeneratedAcir<F> {
     /// By setting `z` to be `0`, we can make `y` equal to `1`.
     /// This is easily observed: `y = 1 - t * 0`
     /// Now since `y` is one, this means that `t` needs to be zero, or else `y * t == 0` will fail.
-    fn is_zero(&mut self, t_expr: &Expression<F>) -> Witness {
-        // We're checking for equality with zero so we can negate the expression without changing the result.
-        // This is useful as it will sometimes allow us to simplify an expression down to a witness.
```
aakoshh, line 539: What happened to this negation argument?

compiler/noirc_evaluator/src/acir/generated_acir.rs
TomAFrench: Previously we'd handle the cases `t_expr = x` and `t_expr = -x` only as we were looking to extract the witness `x`. Now we tolerate `t_expr = m*x +c` so we can avoid creating witnesses in more cases.

We can do this as the scaling factor doesn't affect the following constraints and constants in`t_expr` introduce another linear term to the following constraints which fits inside the existing gates.

compiler/noirc_evaluator/src/acir/generated_acir.rs
TomAFrench: I have generalised this further, as mentioned above.

compiler/noirc_evaluator/src/acir/generated_acir.rs
aakoshh: Ah because now this is from `to_witness` is not happening:
```rust
            if coefficient.is_one() && constant.is_zero() {
                return Some(variable);
            }
```

aakoshh (approved):
Nice one, great reduction in bytecode size 👏 

As I mentioned it would be nice to make the `y_is_boolean_constraint ` expression more fluid, but otherwise looks correct 👍

compiler/noirc_evaluator/src/acir/generated_acir.rs
TomAFrench: > (&z_expr * &linear_t).expect("both linear") + y_expr - F::one()

I'd like to avoid this addition of `y_expr` as we're then going to be doing a full `add_mul` when we know that it's going to result in just pushing an extra linear term. This way allows us to just do a simple push onto the linear terms vec.

aakoshh: Yes, `add_mul` does more; perhaps then we could rewrite methods to be able to chain them like this?
```
(&z_expr * &linear_t).expect("both linear")
  .push_addition_term(F::one(), y)
  .sub(F::one())
```
Or make `add_mul` look for more simplification options at the beginning.

PR #7032, opened 2025-01-11
feat: add `ConstrainNotEqual` instruction

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

Currently when asserting that two values are not equal we need to use

```
v2 = eq v0, v1
constrain v2 == u1 0
```

Because we have a full eq instruction being passed to acirgen we need all the handling for the case where the two values ARE equal. Rather than just asserting that an inverse exists for the difference between these two values we need to create two unnecessary witnesses - one which is unconstrained and the other constrained to zedo. This is unnecessary as we want the circuit to just fail in this case.

This PR creates a `ConstrainNotEqual` instruction that allows ACIRgen to just directly assert that an inverse exists.
## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/acir/acir_variable.rs
```diff
@@ -541,6 +541,26 @@ impl<F: AcirField, B: BlackBoxFunctionSolver<F>> AcirContext<F, B> {
...
+    pub(crate) fn assert_neq_var(
+        &mut self,
+        lhs: AcirVar,
+        rhs: AcirVar,
+        assert_message: Option<AssertionPayload<F>>,
+    ) -> Result<(), RuntimeError> {
+        let diff_var = self.sub_var(lhs, rhs)?;
+
+        let one = self.add_constant(F::one());
+        let _ = self.inv_var(diff_var, one)?;
```
aakoshh, line 554: So `one` is the `predicate`, which will be [multiplied with](https://github.com/noir-lang/noir/blob/81a3e8f3c39bf65d191d75eec81b34961904e80e/compiler/noirc_evaluator/src/acir/acir_variable.rs#L372) `(lhs-rhs) * (1/(lhs-rhs))` before we check that `(lhs-rhs) * (1/(lhs-rhs)) * predicate == predicate`. 

Just curious: if you're trying to minimise the number of operations, why not get rid of the multiplication with the predicate, and go straight to `assert_eq_var(should_be_one, one)`? I mean I see that `inv_var` is convenient, but yesterday the `add_mul` would have been correct, but did more than needed, so you preferred to call the push method directly.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -878,6 +902,7 @@ impl Instruction {
                     SimplifiedToInstructionMultiple(constraints)
                 }
             }
+            Instruction::ConstrainNotEqual(..) => None,
```
aakoshh, line 905: Looking at [decompose_constrain](https://github.com/noir-lang/noir/blob/81a3e8f3c39bf65d191d75eec81b34961904e80e/compiler/noirc_evaluator/src/ssa/ir/instruction/constrain.rs#L9), some of the scenarios could be applied to `ConstrainNotEqual`, for example:

```
                        // Replace an explicit two step inequality assertion
                        //
                        // v2 = eq v0, u32 v1
                        // constrain v2 == u1 0
                        //
                        // with a direct assertion of iequality between the two values
                        //
                        // v2 = eq v0, u32 v1
                        // constrain v0 != v1
```

Also [constant.is_one()](https://github.com/noir-lang/noir/blob/81a3e8f3c39bf65d191d75eec81b34961904e80e/compiler/noirc_evaluator/src/ssa/ir/instruction/constrain.rs#L29) could have a `constant.is_zero()` counterpart in `decompose_constrain` itself which returns `ConstrainNotEqual` instead of `Constrain`, perhaps?

compiler/noirc_evaluator/src/ssa/opt/make_constrain_not_equal.rs
```diff
@@ -0,0 +1,74 @@
...
+            // check_instruction can also return an error
+            let instructions = self.dfg[block].instructions().to_vec();
+
+            for instruction in instructions {
+                let constrain_ne: Instruction = match &self.dfg[instruction] {
+                    Instruction::Constrain(lhs, rhs, msg) => {
+                        if self
+                            .dfg
+                            .get_numeric_constant(*rhs)
+                            .map_or(false, |constant| constant.is_zero())
```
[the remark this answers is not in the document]
TomAFrench: I initially went down this route but ran into issues when it came to flattening the CFG. `Instruction::Constrain` can be disabled quite nicely by multiplying both sides by the condition but it's not immediately obvious on how to do this for `Instruction::ConstrainNotEqual`.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
TomAFrench: See below on my reasoning on not having this as part of the simplification step. This also strikes me as a "terminal" optimization where it can't unlock new optimizations easily so having this happen once at the end of SSA optimization seems ok to me.

compiler/noirc_evaluator/src/acir/acir_variable.rs
TomAFrench: > why not get rid of the multiplication with the predicate, and go straight to assert_eq_var(should_be_one, one)

We short circuit on multiplications with a constant value of one so I thought that this would be clearest.

TomAFrench: The difference here in my mind is that we're going through a more general case which short circuits whereas the `add_mul` case didn't.

compiler/noirc_evaluator/src/ssa/opt/make_constrain_not_equal.rs
```diff
@@ -0,0 +1,74 @@
...
+
+impl Function {
+    pub(crate) fn make_constrain_not_equal(&mut self) {
+        if !self.runtime().is_acir() {
+            return;
+        }
+
+        for block in self.reachable_blocks() {
+            // Unfortunately we can't just use instructions.retain(...) here since
+            // check_instruction can also return an error
```
TomAFrench, line 35:
```suggestion
```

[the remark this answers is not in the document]
TomAFrench: Thanks, leftover code.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
aakoshh: Okay, that answers the 2nd part 👍  
And because you only insert `ConstrainNotEqual` as a last step, `simplify` will never encounter it, right? or not with varied conditions at least.

aakoshh (approved):
👍

Issue #5792, opened 2024-08-22
Do not duplicate assertion payloads for Brillig functions


vezenovm:
### Problem

We duplicate assertion messages for all Brillig calls that are inserted from generated brillig into our generated ACIR https://github.com/noir-lang/noir/blob/19ffa2008fc9cbb5972b50d66d14908d5c82ed75/compiler/noirc_evaluator/src/ssa/acir_gen/acir_ir/generated_acir.rs#L593. 

### Happy Case

Similarly to our deduplication of Brillig bytecode with the BrilligCall opcode and [Brillig debug locations](https://github.com/noir-lang/noir/pull/5696) we should also deduplicate the Brillig assertion payloads so to not repeat them as part of ACIR. 

We should note though that assertion payloads are included as part of the `Circuit` struct and resolved using that. This is necessary for ACIR but we should be able to split out unconstrained functionality. Either way, this is a breaking serialization change.

### Workaround

Yes

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #7034, opened 2025-01-13
feat: location tree for debug_info

### LINKED ISSUE #6946: Reduce memory footprint of artifact debug info
We recently [heavily reduced the memory consumption of the callstack representation within the compiler](https://github.com/noir-lang/noir/pull/6791) however once compilation is complete we then use the old inefficient representation of the callstacks.

https://github.com/noir-lang/noir/blob/000799244b6df18f6d3a1335d2e054c983afa0d1/compiler/noirc_errors/src/debug_info.rs#L103-L105

If we pull up the memory flamegraph of executing `rollup-base-public` with a cached artifact, we can then see that 96%  of the peak memory usage comes from just deserializing the program artifact. Of this 1.6GB (52.6%) comes from the debug info as opposed to 1.3GB of actual bytecode (43.6%).

Image

We should perform a similar optimization to the debug data stored in the program artifacts to minimise their memory footprint.

guipublic:
# Description

## Problem\*

Resolves #6946 

## Summary\*
Adds the location tree to debug info and reference call-stacks through their 'index' in the tree (i.e a node of the tree).

## Additional Context
The location tree is merged from ACIR location tree and brillig per function location trees.
N.B In case of several ACIR functions, we get several location trees (one per function). These trees could be merged, but this can be done in a follow-up PR.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Merging in master to compare artifact sizes in aztec-packages.

TomAFrench:
Nice, this drops the artifact size of `rollup-block-root-single-tx` by ~1MB from 20 to 19. Compilation time drops by 50s

compiler/wasm/src/types/noir_artifact.ts
```diff
@@ -145,14 +145,22 @@ export type BrilligFunctionId = number;
 
 export type OpcodeToLocationsMap = Record<OpcodeLocation, SourceCodeLocation[]>;
```
TomAFrench, line 146: Is this type necessary anymore?

compiler/noirc_evaluator/src/acir/acir_variable.rs
```diff
@@ -2088,6 +2092,15 @@ impl<F: AcirField, B: BlackBoxFunctionSolver<F>> AcirContext<F, B> {
...
     }
+
+    // pub(crate) fn set_locations(&mut self, locations: Vec<LocationNode>) {
+    //     self.acir_ir.location_tree = LocationTree {
+    //         locations:     locations.into_iter().map(|node| LocationNodeDebugInfo {
+    //             value: node.value,
+    //             parent: node.parent,
+    //         }).collect(),
+    //     }
+    // }
```
TomAFrench, line 2103:
```suggestion
```

compiler/wasm/src/types/noir_artifact.ts
guipublic: It is used by brillig_locations, so I guess we need it.

compiler/noirc_evaluator/src/acir/acir_variable.rs
guipublic: Done

compiler/noirc_errors/src/debug_info.rs
```diff
@@ -96,13 +98,14 @@ impl ProgramDebugInfo {
...
+        BTreeMap<BrilligFunctionId, BTreeMap<BrilligOpcodeLocation, CallStackId>>,
+    pub location_tree: LocationTree,
     /// Map opcode index of an ACIR circuit into the source code location
     /// Serde does not support mapping keys being enums for json, so we indicate
     /// that they should be serialized to/from strings.
     #[serde_as(as = "BTreeMap<DisplayFromStr, _>")]
-    pub locations: BTreeMap<OpcodeLocation, Vec<Location>>,
-    pub brillig_locations:
-        BTreeMap<BrilligFunctionId, BTreeMap<BrilligOpcodeLocation, Vec<Location>>>,
+    pub location_map: BTreeMap<OpcodeLocation, CallStackId>,
```
TomAFrench, line 108: Can you give an explainer on how exactly these fit together? I'm thinking that we're storing unnecessary data in `location_map` now.

My understanding is that:
- `brillig_locations` stores for each brillig function a mapping from each opcode index to a callstack id
- We can then use this and then look up the callstack id in `location_tree` to get the associate callstack for any brillig opcode.
- `location_map` stores a mapping from `(acir_opcode_index, brillig_index)` to a callstack id which we can use with `location_tree`.

Why do we need to track `(acir_opcode_index, brillig_index)` in `location_map` now? If we fail inside of a brillig call we can tell which brillig function we're executing from the ACIR opcode we've halted on, we then
1. Get the callstack for the ACIR opcode in which we make the call
2. Check the brillig opcode within that call we halted on and get the relevant callstack for the unconstrained function.
3. Smoosh these two together to generate the final callstack.

TomAFrench (commented):
Forgot to hit send on these comments.

Also aztec-packages has their own homebrew method of dealing with debug info in their simulator which is this is definitely going to break. Can you take a look at what this will take to update?

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -45,6 +45,9 @@ pub(crate) struct BrilligBlock<'block> {
 }
 
 impl<'block> BrilligBlock<'block> {
+    // pub(crate) fn get_or_insert_location(&mut self, call_stack: &CallStack) {
+    //     self.brillig_context.get_or_insert_locations(&call_stack);
+    // }
```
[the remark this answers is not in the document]
guipublic: Done

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -991,8 +991,8 @@ fn brillig_bytecode_size(function: &Function) -> usize {
 
     // This is to try to prevent hitting ICE.
     temp.dead_instruction_elimination(false);
-
-    convert_ssa_function(&temp, false).byte_code.len()
+    let mut brillig_temp = Brillig::default();
+    brillig_temp.convert_ssa_function(&temp, false).byte_code.len()
```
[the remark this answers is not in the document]
guipublic: Done

compiler/noirc_errors/src/call_stack.rs
```diff
@@ -122,7 +142,19 @@ impl CallStackHelper {
...
+        self.extend_call_stack(CallStackId::root(), locations)
+    }
+
+    // Clone the locations into a LocationTree
+    pub fn to_location_tree(&self) -> LocationTree {
+        LocationTree {
+            locations: self
+                .locations
+                .clone()
+                .into_iter()
```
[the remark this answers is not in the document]
guipublic: Done

tooling/debugger/src/context.rs
```diff
@@ -450,7 +452,14 @@ impl<'a, B: BlackBoxFunctionSolver<FieldElement>> DebugContext<'a, B> {
                         [debug_location.circuit_id as usize]
                         .brillig_locations
                         .get(&brillig_function_id);
-                    brillig_locations.unwrap().get(&brillig_location).cloned().unwrap_or_default()
+                    let call_stack_id = brillig_locations
+                        .unwrap()
+                        .get(&brillig_location)
+                        .cloned()
```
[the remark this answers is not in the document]
guipublic: Done

compiler/noirc_errors/src/debug_info.rs
guipublic: In theory this makes sense, in practice I don't see where those brillig locations are inserted in the location_map.

compiler/noirc_evaluator/src/brillig/mod.rs
```diff
@@ -49,6 +50,31 @@ impl Brillig {
...
+        for block in function_context.blocks.clone() {
+            BrilligBlock::compile(&mut function_context, &mut brillig_context, block, &func.dfg);
+        }
+        self.call_stacks = brillig_context.call_stacks.clone();
+        let mut artifact = brillig_context.artifact();
+        artifact.name = func.name().to_string();
+        artifact.location_tree = func.dfg.call_stack_data.to_location_tree();
+
+        artifact
+    }
```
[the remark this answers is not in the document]
guipublic: Done

compiler/noirc_errors/src/debug_info.rs
TomAFrench: Hmm, true we seem to have removed the insertion of any `OpcodeLocation::Brillig` a while back.

It still stands that we don't need to keep this enum in the build artifact anymore though as it's currently tech debt. We've kept this enum around just to keep the program serialization format the same but as we're changing it now anyway, now's a good time to clear up this tech debt.
https://github.com/noir-lang/noir/blob/0512ee67cc6e4d963c3c8f34c29de40063aae9b7/acvm-repo/acir/src/circuit/mod.rs#L139-L147

https://github.com/noir-lang/noir/issues/5792

guipublic: Done.
What I did is create a new AcirOpcodeLocation type and use it inside DebugInfo LocationMap, instead of OpcodeLocation, convertir between the 2 when necessary.
I did not touch the other usage of OpcodeLocation, which is used also with BrilligFlavor if a crash occur during ACVM execution.

acvm-repo/acir/src/circuit/mod.rs
```diff
@@ -146,6 +146,37 @@ pub enum OpcodeLocation {
...
+    type Err = OpcodeLocationFromStrError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        fn parse_index(input: &str) -> Result<AcirOpcodeLocation, ParseIntError> {
+            let index = input.parse()?;
+            Ok(AcirOpcodeLocation::new(index))
+        }
+        parse_index(s)
+            .map_err(|_| OpcodeLocationFromStrError::InvalidOpcodeLocationString(s.to_string()))
+    }
+}
```
TomAFrench, line 170: We shouldn't need to do serde through strings anymore.

compiler/wasm/src/types/noir_artifact.ts
TomAFrench, line 146: `OpcodeLocation` can be serialized to a number now that it's just the ACIR index.

acvm-repo/acir/src/circuit/mod.rs
guipublic: I removed the string serialisation

TomAFrench:
Execution memory:

```
[
    {
      name: 'private-kernel-inner',
      value: 207.33,
      unit: 'MB',
      range: undefined,
      extra: undefined
    },
    {
      name: 'private-kernel-reset',
      value: 231.5,
      unit: 'MB',
      range: undefined,
      extra: undefined
    },
    {
      name: 'private-kernel-tail',
      value: 183.61,
      unit: 'MB',
      range: undefined,
      extra: undefined
    },
    {
      name: 'rollup-base-private',
      value: 619.86,
      unit: 'MB',
      range: undefined,
      extra: undefined
    },
    {
      name: 'rollup-base-public',
      value: 718.86,
      unit: 'MB',
      range: undefined,
      extra: undefined
    },
    {
      name: 'rollup-block-merge',
      value: 551.41,
      unit: 'MB',
      range: undefined,
      extra: undefined
    },
    {
      name: 'rollup-block-root',
      value: 1070,
      unit: 'MB',
      range: undefined,
      extra: undefined
    },
    {
      name: 'rollup-merge',
      value: 529.46,
      unit: 'MB',
      range: undefined,
      extra: undefined
    },
    {
      name: 'rollup-root',
      value: 551.39,
      unit: 'MB',
      range: undefined,
      extra: undefined
    }
  ]
```

compiler/noirc_driver/src/debug.rs
```diff
@@ -21,10 +21,14 @@ pub(crate) fn filter_relevant_files(
...
-                .locations
-                .values()
-                .flat_map(|call_stack| call_stack.iter().map(|location| location.file))
+            function_symbols.location_map.values().flat_map(|call_stack_id| {
+                function_symbols
+                    .location_tree
+                    .get_call_stack(*call_stack_id)
+                    .iter()
+                    .map(|location| location.file)
+                    .collect::<Vec<_>>()
```
vezenovm, line 30:
```suggestion
                    .into_iter()
                    .map(|location| location.file)
```
We should be able to do this which is more idiomatic as `get_call_stack` returns `Vec<Location>`

compiler/noirc_driver/src/debug.rs
```diff
@@ -33,9 +37,14 @@ pub(crate) fn filter_relevant_files(
...
             let brillig_location_maps =
                 function_symbols.brillig_locations.values().flat_map(|brillig_location_map| {
-                    brillig_location_map
-                        .values()
-                        .flat_map(|call_stack| call_stack.iter().map(|location| location.file))
+                    brillig_location_map.values().flat_map(|call_stack_id| {
+                        function_symbols
+                            .location_tree
+                            .get_call_stack(*call_stack_id)
+                            .iter()
```
vezenovm, line 44: Same as above

compiler/noirc_errors/src/debug_info.rs
```diff
@@ -93,16 +94,13 @@ impl ProgramDebugInfo {
...
-    /// Serde does not support mapping keys being enums for json, so we indicate
-    /// that they should be serialized to/from strings.
-    #[serde_as(as = "BTreeMap<DisplayFromStr, _>")]
-    pub locations: BTreeMap<OpcodeLocation, Vec<Location>>,
     pub brillig_locations:
-        BTreeMap<BrilligFunctionId, BTreeMap<BrilligOpcodeLocation, Vec<Location>>>,
+        BTreeMap<BrilligFunctionId, BTreeMap<BrilligOpcodeLocation, CallStackId>>,
+    pub location_tree: LocationTree,
+    /// Map opcode index of an ACIR circuit into the source code location
+    pub location_map: BTreeMap<AcirOpcodeLocation, CallStackId>,
```
vezenovm, line 103:
```suggestion
    pub acir_locations: BTreeMap<AcirOpcodeLocation, CallStackId>,
```
nit: What do we think about doing this to distinguish what type of map we have, like `brillig_locations`?

compiler/noirc_evaluator/src/acir/generated_acir.rs
```diff
@@ -582,7 +590,7 @@ impl<F: AcirField> GeneratedAcir<F> {
         if num_bits >= F::max_num_bits() {
             return Err(RuntimeError::InvalidRangeConstraint {
                 num_bits: F::max_num_bits(),
-                call_stack: self.call_stack.clone(),
+                call_stack: self.get_call_stack(),
```
vezenovm, line 593:
```suggestion
                call_stack: self.call_stacks.get_call_stack(self.call_stack_id),
```
Is this not valid and then we can get rid of `call_stack: CallStack`?

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -55,6 +55,7 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
         brillig_context: &'block mut BrilligContext<FieldElement, Registers>,
         block_id: BasicBlockId,
         dfg: &DataFlowGraph,
+        call_stacks: &mut CallStackHelper,
```
vezenovm, line 58: I feel we could put this on the BrilligBlock itself but it is non-blocking

compiler/noirc_evaluator/src/brillig/mod.rs
```diff
@@ -65,6 +68,38 @@ impl Brillig {
             _ => unreachable!("ICE: Expected a function or procedure label"),
         }
     }
+
+    /// Converting an SSA function into Brillig bytecode.
+    pub(crate) fn convert_ssa_function(
```
vezenovm, line 73: nit: this may make more sense inside of `brillig_fn.rs` like `convert_ssa_globals` is in `brillig_globals.rs` but this is non-blocking just noting it.

tooling/debugger/src/source_code_printer.rs
```diff
@@ -266,12 +267,20 @@ mod tests {
...
         // We don't care about opcodes in this context,
         // we just use a dummy to construct debug_symbols
-        let mut opcode_locations = BTreeMap::<OpcodeLocation, Vec<Location>>::new();
-        opcode_locations.insert(OpcodeLocation::Acir(42), vec![loc]);
+        let mut opcode_locations = BTreeMap::<AcirOpcodeLocation, CallStackId>::new();
+        opcode_locations.insert(AcirOpcodeLocation::new(42), CallStackId::new(1));
+        let mut location_tree = LocationTree::default();
+        location_tree
+            .locations
+            .push(LocationNodeDebugInfo { parent: None, value: Location::dummy() });
```
vezenovm, line 275: Just for my understanding why do we push a `None` and dummy location here first instead of just starting with the root() and loc?

tooling/nargo/src/errors.rs
```diff
@@ -165,25 +166,21 @@ fn extract_locations_from_error<F: AcirField>(
         opcode_locations
             .iter()
             .flat_map(|resolved_location| {
+                let call_stack_id = match resolved_location.opcode_location {
+                    OpcodeLocation::Acir(idx) => *debug[resolved_location.acir_function_index]
+                        .location_map
+                        .get(&AcirOpcodeLocation::new(idx))
+                        .unwrap_or(&CallStackId::root()),
+                    // TODO: should we use acir_index here and merge the 2 call stacks?
```
vezenovm, line 174: Are you talking about making one map and using an enum of `AcirOpcodeLocation` and `BrilligOpcodeLocation`?

tooling/nargo/src/errors.rs
```diff
@@ -233,6 +230,7 @@ pub fn try_to_diagnose_runtime_error(
 ) -> Option<FileDiagnostic> {
     let source_locations = match nargo_err {
         NargoError::ExecutionError(execution_error) => {
+            dbg!(&debug);
```
vezenovm, line 233:
```suggestion
```
leftover

tooling/noirc_artifacts/src/debug.rs
```diff
@@ -26,10 +26,14 @@ impl DebugArtifact {
...
-                    .locations
-                    .values()
-                    .flat_map(|call_stack| call_stack.iter().map(|location| location.file))
+                function_symbols.location_map.values().flat_map(|call_stack_id| {
+                    function_symbols
+                        .location_tree
+                        .get_call_stack(*call_stack_id)
+                        .iter()
+                        .map(|location| location.file)
+                        .collect::<Vec<_>>()
```
vezenovm, line 35:
```suggestion
                        .into_iter()
                        .map(|location| location.file)
```
same thing from earlier

tooling/noirc_artifacts/src/debug.rs
```diff
@@ -38,9 +42,14 @@ impl DebugArtifact {
...
-                        brillig_location_map
-                            .values()
-                            .flat_map(|call_stack| call_stack.iter().map(|location| location.file))
+                        brillig_location_map.values().flat_map(|call_stack_id| {
+                            function_symbols
+                                .location_tree
+                                .get_call_stack(*call_stack_id)
+                                .iter()
+                                .map(|location| location.file)
+                                .collect::<Vec<_>>()
```
vezenovm, line 51:
```suggestion
                                .into_iter()
                                .map(|location| location.file)
```
Same thing from earlier

tooling/profiler/src/flamegraph.rs
```diff
@@ -176,29 +176,38 @@ fn find_callsite_labels<'files>(
...
-        {
-            let procedure_locs = debug_symbols.brillig_procedure_locs.get(&brillig_function_id);
-            if let Some(procedure_locs) = procedure_locs {
-                for (procedure, range) in procedure_locs.iter() {
-                    if brillig_location.0 >= range.0 && brillig_location.0 <= range.1 {
-                        procedure_id = Some(*procedure);
-                        break;
+    let source_locations = match opcode_location {
+        OpcodeLocation::Acir(idx) => {
+            debug_symbols.acir_opcode_location(&AcirOpcodeLocation::new(*idx)).unwrap_or_default()
```
vezenovm, line 181:
```suggestion
            debug_symbols.acir_opcode_location(opcode_location).unwrap_or_default()
```
Could we just do this? Looks a bit cleaner

tooling/profiler/src/flamegraph.rs
```diff
@@ -356,32 +369,36 @@ mod tests {
         let baz_whatever_call_location =
             Location::new(find_spans_for(source_code, "whatever()")[2], file_id);
 
-        let mut opcode_locations = BTreeMap::<OpcodeLocation, Vec<Location>>::new();
+        //  let mut opcode_locations = BTreeMap::<OpcodeLocation, Vec<Location>>::new();
```
vezenovm, line 372:
```suggestion
```
leftover

vezenovm (commented):
Mostly looks good, just some nits. Only one main question about removing the `call_stack: CallStack` field on the `GeneratedAcir<F>` struct

acvm-repo/acir/src/circuit/mod.rs
```diff
@@ -146,6 +146,24 @@ pub enum OpcodeLocation {
     Brillig { acir_index: usize, brillig_index: usize },
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
+/// Opcodes are locatable so that callers can
+/// map opcodes to debug information related to their context.
+pub struct AcirOpcodeLocation(usize);
```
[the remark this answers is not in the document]
guipublic: Done

compiler/noirc_driver/src/debug.rs
guipublic: Done

compiler/noirc_driver/src/debug.rs
guipublic: Done

compiler/noirc_errors/src/debug_info.rs
guipublic: Done

tooling/nargo/src/errors.rs
guipublic: Done

tooling/noirc_artifacts/src/debug.rs
guipublic: Done

tooling/profiler/src/flamegraph.rs
guipublic: No, one is OpcodeLocation while the other is AcirOpcodeLocation

tooling/profiler/src/flamegraph.rs
guipublic: Done

tooling/nargo/src/errors.rs
guipublic: I think it was about having only one 'location tree' for all the acir functions, instead of one per acir function.
Not sure if it's worth it.

tooling/debugger/src/source_code_printer.rs
guipublic: The (None, dummy location) is the root, so that's what we do. Root is index 0, we need to have something at index 0 to represent it. We need a root to account for empty call stack.

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
guipublic: The location tree is englobing all the blocks so it needs to be passed around the blocks, which is what we do here.
It seems more straightforward to me.

tooling/nargo/src/errors.rs
vezenovm: Would be nice to consolidate our debug symbols and locations for one program, but yeah not sure how worth it it would be. Let's either add an issue or remove this todo though as this todo is unclear to me on its own.

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
vezenovm: Either way I will not block on this, so no need to update.

compiler/noirc_evaluator/src/acir/generated_acir.rs
guipublic: Yes it is valid, and it does indeed simplify the code a little bit since we can get rid of self.call_stack.
However, it will force to recompute the call stack every time, even if the source location has not changed.
Since this optimisation is very simple, I figure we should keep it.

compiler/noirc_evaluator/src/brillig/mod.rs
guipublic: The call stacks do span several brillig functions, so I think it's better here.

acvm-repo/acir/src/circuit/mod.rs
aakoshh, line 152:
```suggestion
/// Opcodes are locatable so that callers can
/// map opcodes to debug information related to their context.
#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
pub struct AcirOpcodeLocation(usize);
```

Seems to be a more common pattern.

compiler/noirc_evaluator/src/acir/generated_acir.rs
vezenovm: I feel recomputing the call stack should not take much time, especially as this error should be rare. But as this PR has been up very long I am not going to block on this.

vezenovm (approved):

TomAFrench:
@vezenovm Flagging that this PR will break callstack generation in aztec-packages as the acir-simulator would need to be updated. I think it's worth making callstack generation part of noir_js to get around having a broken sync.

tooling/nargo/src/errors.rs
guipublic: I created issue #7517 and removed the TODO.

guipublic:
I merged the branch from #7642 and updated the callstack generation in javascript. It should not fail in AztecPackages anymore.

PR #7036, opened 2025-01-13
feat: avoid inserting `inc_rc` instructions into ACIR

TomAFrench:
# Description

## Problem\*

Followup to #7017 

## Summary\*

This PR removes the remaining locations where we push inc_rc instructions into ACIR. I've then made the compiler panic if it sees these being inserted again.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
Is there really that much of a difference in checking is_brillig before inserting these instructions versus checking it after and returning InstructionRemoved? Worried that going down this route more will complicate the passes.

I see -15% in private-kernel-reset but am not sure why it is such a large change for just moving a conditional earlier.

TomAFrench:
This difference should be explained by just not having to print so many warnings to the log file as there was a _lot_ of these instructions being inserted.

PR #7037, opened 2025-01-13
feat: auto-import traits when suggesting trait methods

asterite:
# Description

## Problem

Part of #6956

## Summary

LSP will now auto-import a trait when completing trait methods:

lsp-auto-import-trait

lsp-auto-import-trait-2

## Additional Context

There's a bug now, where if a trait is private but is exposed by a public import, the trait is imported but it leads to a "private" error:

lsp-auto-import-bug

Rust Analyzer, somehow, will suggest the first public `use` it finds for that alias... we could try to do the same thing (no idea how yet 😅) but I'd leave that to another PR.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/lsp/src/requests/completion.rs
+++ b/tooling/lsp/src/requests/completion.rs
@@ -706,6 +706,7 @@ impl<'a> NodeFinder<'a> {
                     function_completion_kind,
                     function_kind,
                     None, // attribute first type
+                    trait_id,
                     self_prefix,
                 );
                 if !completion_items.is_empty() {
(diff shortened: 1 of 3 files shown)


jfecher (approved):
Very nice!

TomAFrench:
This seems to be causing some problems in aztec-packages.

TomAFrench:
On second thoughts this might just be uncovering an underlying issue... On the sync PR in the bloblib in `blob.nr`, if I try to use the autoimport to import `BigNumTrait`, the import at the top goes from
```
use bigint::BigNum;
```

to

```
use bigint::{BigNum, bignum::BigNumTrait};
```

However when saving the file this then becomes 

```
use bigint::BigNum::{self, BigNumTrait};
```

which is invalid.

asterite:
Interesting! It seems the formatter is considering "BigNum" and "bignum" to be the same. I'll send a fix soon.

PR #7040, opened 2025-01-13
feat(ssa): Treat globals as constant in a function's DFG

vezenovm:
# Description

## Problem\*

Resolves TODO leftover from https://github.com/noir-lang/noir/pull/6985

Also mentioned in this comment https://github.com/noir-lang/noir/pull/7021#issuecomment-2583271915.

## Summary\*

In order to re-use our existing instruction simplification we need to be able to access globals inside of the DFG. In order to do this I have created a new `GlobalsGraph` type that now exists as an `Arc<GlobalsGraph>` in the `DataFlowGraph`. `Arc` is used so that we can share the `GlobalsGraph` across all functions. However, we will have to eat a cloning cost as we now place the globals information in all functions rather than simply placing it in the main `Ssa` object.

## Additional Context

In #6985 we only added a `Value::Global`, but globals support both numeric constants and array constants. This means that we can have overlapping instruction ids in the function graph and the globals graph. To avoid having to add placeholder instructions like we do for values, when checking for a constant array instruction we also check whether the value tied to that instruction is specified as a global. If the instruction is specified as global we look for the instruction in the globals graph.

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
> **bigint**	+12 ❌	**+0.57%**

For some reason it seems inlining global constants immediately causes this test to regress. We saw the same improvement as this regression in PR #6985 https://github.com/noir-lang/noir/pull/6985#issuecomment-2577921642.

compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs
```diff
@@ -23,6 +23,8 @@ use self::{
     value::{Tree, Values},
 };
 
+use super::ir::dfg::GlobalsGraph;
+// use super::ir::dfg::Globals;
```
vezenovm, line 27:
```suggestion
```

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -685,6 +677,20 @@ impl DataFlowGraph {
     pub(crate) fn is_global(&self, value: ValueId) -> bool {
         matches!(self.values[value], Value::Global(_))
     }
+
+    pub(crate) fn get_instruction(&self, value: ValueId) -> Option<&Instruction> {
```
asterite, line 681: I wonder if this should be the implementation of `Index` for `InstructionId`. What do you think? That way we don't have to remember to call `dfg.get_instruction(id)` instead of doing `dfg[id]`.

TomAFrench: `dfg[id]` wouldn't return an `Option` so may introduce panic if we sometimes look up non-instructions here.

vezenovm: Yeah I thought the same thing but this would require some more changes.

From the PR description:
> In #6985 we only added a `Value::Global`, but globals support both numeric constants and array constants. This means that we can have overlapping instruction ids in the function graph and the globals graph. To avoid having to add placeholder instructions like we do for values, when checking for a constant array instruction we also check whether the value tied to that instruction is specified as a global. If the instruction is specified as global we look for the instruction in the globals graph.

This would most likely require some kind of `Instruction::Global` similar to `Value::Global`. If we feel it would be clearer to do this I can make the change.

asterite: Oh, makes sense. This is fine then.

compiler/noirc_evaluator/src/ssa/ir/printer.rs
```diff
@@ -85,7 +85,13 @@ fn value(dfg: &DataFlowGraph, id: ValueId) -> String {
         Value::Function(id) => id.to_string(),
         Value::Intrinsic(intrinsic) => intrinsic.to_string(),
         Value::ForeignFunction(function) => function.clone(),
-        Value::Param { .. } | Value::Instruction { .. } => id.to_string(),
+        Value::Param { .. } | Value::Instruction { .. } => {
+            if dfg.is_global(id) {
+                format!("g{}", id.to_u32())
```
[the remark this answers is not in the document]
vezenovm: We switched `g` https://github.com/noir-lang/noir/pull/6985#discussion_r1909111012

compiler/noirc_evaluator/src/ssa/opt/normalize_value_ids.rs
```diff
@@ -193,14 +197,15 @@ impl IdMaps {
             }
 
             Value::NumericConstant { constant, typ } => {
+                if old_function.dfg.is_global(old_value) {
+                    return old_value;
+                }
```
asterite, line 202: Is this check needed in addition to the one at the top?

vezenovm, line 202:
```suggestion
```

asterite (approved):
Looks good!

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -619,17 +646,21 @@ impl DataFlowGraph {
 
     /// True if the given ValueId refers to a (recursively) constant value
     pub(crate) fn is_constant(&self, argument: ValueId) -> bool {
-        match &self[self.resolve(argument)] {
+        let argument = self.resolve(argument);
+        if self.is_global(argument) {
+            return true;
+        }
```
jfecher, line 652: This is a bit odd when `self.is_global` just matches on the value for a `Value::Global`, I think it'd be clearer to just have the Global match case below return true instead of unreachable

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -642,6 +673,24 @@ impl DataFlowGraph {
             false
         }
     }
+
+    pub(crate) fn is_global(&self, value: ValueId) -> bool {
+        matches!(self.values[value], Value::Global(_))
+    }
+
+    pub(crate) fn get_instruction(&self, value: ValueId) -> Option<&Instruction> {
+        match &self[value] {
```
jfecher, line 682: I think this method needs some clarification in both a doc comment and making the name more verbose to highlight that the returned instruction may not be in this DFG (!) since using the valueids directly from it could result in different underlying values from this DFG. As-is I can see someone using this as a general purpose way to get an instruction in an optimization pass and run into that bug without knowing - and it'd be hard to track down.

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -660,7 +709,11 @@ impl std::ops::IndexMut<InstructionId> for DataFlowGraph {
 impl std::ops::Index<ValueId> for DataFlowGraph {
     type Output = Value;
     fn index(&self, id: ValueId) -> &Self::Output {
-        &self.values[id]
+        let value = &self.values[id];
+        if matches!(value, Value::Global(_)) {
+            return &self.globals[id];
+        }
+        value
```
jfecher, line 716: Ok, if we're checking here then my last comment shouldn't be as much of an issue - but I'll leave it anyway since I think it's an important point to highlight.

For arrays of integers, all those integers should be Value::Globals as well, right?

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -1093,28 +1093,27 @@ fn try_optimize_array_get_from_previous_set(
...
-                    Instruction::ArraySet { array, index, value, .. } => {
-                        if let Some(constant) = dfg.get_numeric_constant(*index) {
-                            if constant == target_index {
-                                return SimplifyResult::SimplifiedTo(*value);
-                            }
-
-                            array_id = *array; // recur
-                        } else {
-                            return SimplifyResult::None;
+        if let Some(instruction) = dfg.get_instruction(array_id) {
```
jfecher, line 1096: Do we expect array_set to be used on global arrays at all? I'd expect/hope this wouldn't be needed here

compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
```diff
@@ -278,8 +282,10 @@ fn create_apply_function(
     function_ids: Vec<FunctionId>,
 ) -> FunctionId {
     assert!(!function_ids.is_empty());
-    ssa.add_fn(|id| {
+    let globals_dfg = std::mem::take(&mut ssa.globals.dfg);
+    let new_id = ssa.add_fn(|id| {
         let mut function_builder = FunctionBuilder::new("apply".to_string(), id);
+        function_builder.set_globals(Arc::new(GlobalsGraph::from_dfg(globals_dfg.clone())));
```
jfecher, line 288: I don't understand why we do the whole `std::mem::take` thing with the globals dfg here. Shouldn't we have a `Arc<GlobalsGraph>` somewhere? If `Ssa` doesn't already store a `Arc<GlobalsGraph>` then it probably should - we should avoid cloning entire DFGs if possible - even small ones like globals is likely expected to be.

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
vezenovm: I chose to separately check as I would need to special case `Value::Instruction`. I can switch to checking inside the match cases though.

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
vezenovm: I'll add a safety comment either way. 

How does the name `get_local_or_global_instruction` sound?

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
vezenovm: It is needed for the `MakeArray` case not the `ArraySet` case. I could assert inside of the DFG method that the instruction must be `MakeArray` for safety.

compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
vezenovm: I switched to using `function_ids` to access a function's globals graph. This is safe due to `assert!(!function_ids.is_empty());`.

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
vezenovm: I have switched it

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
vezenovm: I made the rename to `get_local_or_global_instruction` with an assertion for `MakeArray` when we have a global instruction. I also included a small comment stating that this may come from a separate DFG.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
vezenovm: I added an insertion inside of the dfg method.

vezenovm:
> **6_inliner_zero**	-110 ✅	**-1.40%**
> **regression_4449_inliner_zero**	-3,849 ✅	**-1.67%**

Nice, looks like we already see some benefits with different inliner aggressiveness settings. There looks to be some slight regressions as well. This is most likely due to us not accounting for globals in DIE yet. I will leave this for a follow-up though.

jfecher (approved):

PR #7048, opened 2025-01-13
chore(ci): Unify compilation/execution report jobs that take averages with single runs 

### LINKED ISSUE #7002: Unify compilation/execution reports jobs
https://github.com/noir-lang/noir/blob/724547d8c18cb814342441a7ec83e35bbd931cba/.github/workflows/reports.yml#L364-L378

We have two jobs for each report we want to make which we select between depending on whether we want to take averages of this metric. We should consolidate these and allow passing in the number of runs as a parameter rather than using a boolean flag. 

This will simplify both the workflows and the scripts themselves.

vezenovm:
# Description

## Problem\*

Resolves #7002

## Summary\*

Both compilation and execution reports now accept the number of runs we should do for a command rather than depending on a boolean flag. I also unified the github jobs to pass this flag to the script rather than running a separate job depending on whether we want to take averages.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/.github/workflows/reports.yml
+++ b/.github/workflows/reports.yml
@@ -300,17 +300,17 @@ jobs:
       matrix:
         include:
           # - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-contracts, is_library: true }
-          - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/private-kernel-inner, take_average: true }
-          - project: { repo: AztecProtocol/aztec-packages, path: noir-projects/noir-protocol-circuits/crates/private-kernel-tail, take_average: true  }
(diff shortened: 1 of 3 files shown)


.github/workflows/reports.yml
```diff
@@ -344,41 +344,32 @@ jobs:
...
-          touch parse_time.sh
-          chmod +x parse_time.sh
-          cp /home/runner/work/noir/noir/scripts/test_programs/parse_time.sh ./parse_time.sh
-          ./compilation_report.sh 1 1
+          if [ "${{ matrix.project.take_average }}" == "true" ]; then
+            # Take the average of 5 runs 
+            ./compilation_report.sh 1 5
+          else
+            ./compilation_report.sh 1
+          fi
```
TomAFrench, line 359: we can replace `take_average` with `number_runs` and avoid this if-else.

vezenovm: done

TomAFrench (approved):
LGTM

Issue #7001, opened 2025-01-09
Increase in memory used to compile `rollup-base-public`


aakoshh:
### Aim

Followup for https://github.com/noir-lang/noir/pull/6972 added a new `mem2reg` pass. As a consequence we saw a 100% increase in memory used during the compilation of one of the protocol circuits in aztec-packages.

### Expected Behavior

Didn't expect a significant increase in memory usage.

### Bug

https://github.com/noir-lang/noir/pull/6972#issuecomment-2579872993

### To Reproduce

See [how CI does it](https://github.com/noir-lang/noir/blob/526c84eb8bfa61d9a218c3c0bd623aff8a48f25f/.github/workflows/reports.yml#L505). 

### Workaround

None

### Project Impact

None

### Nargo Version

nargo version = 1.0.0-beta.1 noirc version = 1.0.0-beta.1+bb8dd5ce43f0d89e393bd49f8415008826903652 (git version hash: 13b58717e5a95174fb00d99d51b8b58a26c92740, is dirty: false)

### Would you like to submit a PR for this Issue?

None

TomAFrench:
![Image](https://github.com/user-attachments/assets/df5ad974-ab44-42b4-8252-d7d901e14106)

Memory flamegraph showing that the `blocks` field of the `PerFunctionContext` within mem2reg is holding all the memory.

aakoshh:
This shows the heaviest stack trace in terms of memory allocations:
<img width="1493" alt="Image" src="https://github.com/user-attachments/assets/a1351c72-8ce1-43da-8d5a-33b223fb273f" />

It's also inside `Block::unify` like in the flamegraph above, but it points at the `im::OrdMap` as the culprit, rather than `BTreeMap` as in the flamegraph 🤔

jfecher:
The various maps in `Block` have been a known memory issue in the past. They were changed from hashmaps to OrdMaps since those used less memory in some tests in the past. I think further improvement will require more than just a container change. E.g. sacrificing optimizations by arbitrarily removing known values in a Block after some limit. Or rewriting mem2reg more thoroughly to use a different algorithm.

aakoshh:
With a bit of refactoring I could at least narrow it down to the maintenance of aliases, rather than the other fields that use `OrdMap`:
<img width="1186" alt="Image" src="https://github.com/user-attachments/assets/2414d6be-6e6d-492b-b3ef-eea11ddff5f8" />

aakoshh:
[...]

PR #7053, opened 2025-01-14
fix: Reduce memory usage in mem2reg

aakoshh:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/7001

## Summary\*

Changes `mem2reg::block::Block` to only use the mutable APIs of `OrdMap` if unifying the alias sets would cause something to change. That should save some memory, because doing `self.aliases.entry(...).and_modify(...).or_insert_with(...)` causes `OrdMap::get_mut` to be called, which calls `Arc::make_mut` to prepare the entry for modification, which involves making a clone and a new `Arc` if the `Arc` is shared, which it will be because we cloned a `Block` to get the folding started.

This resulted in the memory required for mem2reg to go down to ~1GB, with later peaks corresponding to `convert_ssa_instruction` during ACIR generation:
Screenshot 2025-01-14 at 14 26 14

Screenshot 2025-01-14 at 14 29 14

With https://github.com/noir-lang/noir/pull/7057 merged the results are even better:

Screenshot 2025-01-14 at 15 02 37
Screenshot 2025-01-14 at 15 00 50

TODO in followup PR:
- See if blocks and aliases can be freed up as we go
- Pre-process functions before inlining them (Tom suggests: inline, unroll and mem2reg)

### Profiling on Mac

I couldn't compile the [heaptrack_gui](https://github.com/KDE/heaptrack?tab=readme-ov-file#compile-heaptrack_gui-on-macos-using-homebrew). Instead tried [cargo-instruments](https://crates.io/crates/cargo-instruments) which requires Xcode to be installed, but it stopped working after a while, and I ended up using [Instruments](https://medium.com/@jpmtech/getting-started-with-instruments-a35485574601) directly (a profiler tool on Mac, part of Xcode). 

Executing the following command:

```shell
./target/release/nargo --program-dir ../aztec-packages/noir-projects/noir-protocol-circuits/crates/rollup-base-public compile  --silence-warnings --skip-underconstrained-check --skip-brillig-constraints-check --force
```

Compiled with:
```shell
cargo build -p nargo_cli --release
```

The following is needed for _Instruments_ to be able to start the executable:
```shell
codesign -s - -v -f --entitlements ../debug.plist target/release/nargo
```
where `debug.plist` has the following content:
```xml
<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "https://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0"><dict><key>com.apple.security.get-task-allow</key><true/></dict></plist>
```

### Memory profile

Original memory profile over time:

Screenshot 2025-01-14 at 01 22 26

EXPERIMENT: Changed `AliasSet` to use `OrdSet` instead of `BTreeSet` to utilise structural sharing.

When I tried this on `aztec-packages` it was promising but ~perhaps I didn't run it long enough~ I accidentally got the 2nd mem2reg commented out. Profiling again locally on `noir` revealed this version uses 21GB of memory 😖 : 

Screenshot 2025-01-14 at 01 27 35

The increase seems to come from the allocation of a large number of `Arc`s:
Screenshot 2025-01-14 at 10 13 52

After having a second look at the size of the `AliasSet`s, they seem to actually only contain a single item, so I was barking up the wrong tree.

Other tweaks I tried:
* Replace `Expression` in `aliases: OrdMap<Expression, AliasSet>` with `u64` calculated by `fxhash::hash64`, just to see the impact of not storing clones of `Expression`s; the result is the exact same memory usage, because as it turns out the memory allocated by `OrdMap::get_mut` call is still for `Arc`s, it has always been. 
* Replace `aliases: OrdMap<Expression, AliasSet>` with `std::HashMap` or `std::BTreeMap`; the memory usage in this case went from 5GB to 11GB; this is because of the increased cost of cloning these maps
* Replace `im` with `im_rc` to see if `Arc` vs `Rc` makes any difference, although it's probably just faster than smaller. Can't because of `rayon`. 

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
image

Nice!

aakoshh:
Now merged https://github.com/noir-lang/noir/pull/7058 the results are very similar 👍

PR #7057, opened 2025-01-14
chore(perf): try making AliasSet an enum

asterite:
# Description

## Problem

Related to #7001

## Summary

While pairing with Akosh on improving compilation memory usage he mentioned that he found that most alias sets had one element in them. I was curious if instead of using a `BTreeSet` of size one using an enum variant for that case would improve memory usage a bit, so that's what this PR does.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/mem2reg/alias_set.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/mem2reg/alias_set.rs
@@ -3,15 +3,14 @@ use std::collections::BTreeSet;
 use crate::ssa::ir::value::ValueId;
 
 /// A set of possible aliases. Each ValueId in this set represents one possible value the reference
-/// holding this AliasSet may be aliased to. This struct wrapper is provided so that when we take
-/// the union of multiple alias sets, the result should be empty if any individual set is empty.
-///
-/// Note that we distinguish between "definitely has no aliases" - `Some(BTreeSet::new())`, and
-/// "unknown which aliases this may refer to" - `None`.
+/// holding this AliasSet may be aliased to. This enum wrapper is provided so that when we take
+/// the union of multiple alias sets, the result should be unknown if any individual set is unknown.
 #[derive(Debug, Default, Clone)]
 pub(super) enum AliasSet {
+    /// Unknown which aliases this may refer to.
     #[default]
     Unknown,
+    /// Definitely has no aliases.
     Empty,
     Known(ValueId),
     KnownMultiple(BTreeSet<ValueId>),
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/mem2reg/alias_set.rs
```diff
@@ -9,71 +9,116 @@ use crate::ssa::ir::value::ValueId;
...
     pub(super) fn known(value: ValueId) -> AliasSet {
-        let mut aliases = BTreeSet::new();
-        aliases.insert(value);
-        Self { aliases: Some(aliases) }
+        AliasSet::Known(value)
     }
 
     pub(super) fn known_multiple(values: BTreeSet<ValueId>) -> AliasSet {
-        Self { aliases: Some(values) }
+        AliasSet::KnownMultiple(values)
```
jfecher, line 30: You should test the length of this set first to see if it actually has multiple ids

asterite: Good catch, done!

TomAFrench:
https://crates.io/crates/vec-collections could be useful here.

asterite:
It's a bit strange because in another recent PR, for example #7056, we get this:

image

Here we get this:

image

The 100% increase doesn't make sense 🤔

TomAFrench:
Those percentages have been a bit suss for a while, I tend to go off the raw figures generally.

I've got plans to move us off of that github action anyway.

TomAFrench:
@asterite Thoughts on https://crates.io/crates/vec-collections which could extend some of the benefits of this PR past a single alias? I haven't dug into the distribution of alias set sizes however.

jfecher:
@TomAFrench if anything wrapping SmallVec in those collections should lead to more memory usage, not less. Unless `size_of::<SmallVec<T>>() <= size_of::<Vec<T>>()` then we'll be using more space for smaller vecs even if they're allocating their data on the stack. I could be wrong but that's my guess.

TomAFrench (approved):
LGTM

aakoshh:
This changes the memory profile two ways:

1) It brings the memory used down to 3.5GB:
Screenshot 2025-01-14 at 14 42 27

2) Shortens the duration of the mem2reg pass:
Screenshot 2025-01-14 at 14 44 43

Instead of a symmetrical pyramid as in the baseline of https://github.com/noir-lang/noir/pull/7053 we get a steeper descent, and the whole thing lasts just seconds compared to 30s originally.

TomAFrench:
I'd need to think about it a little bit @jfecher. Good chance you're right and the fact we'd be putting it inside an enum would risk inflating the other variants.

PR #7058, opened 2025-01-14
chore(perf): try using vec-collections's VecSet in AliasSet

asterite:
# Description

## Problem

Related to #7001

## Summary

Alternative to #7057

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -645,6 +645,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8c3c1a368f70d6cf7302d78f8f7093da241fb8e8807c05cc9e51a125895a6d5b"
 
 [[package]]
+name = "binary-merge"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "597bb81c80a54b6a4381b23faba8d7774b144c94cbd1d6fe3f1329bd776554ab"
+
+[[package]]
 name = "bincode"
 version = "1.3.3"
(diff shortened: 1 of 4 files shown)


TomAFrench:
Looks very similar in performance, up to you about which solution you prefer.

asterite:
Yup, it seems it's exactly the same regarding memory :-)

Compilation times are different (on the left is the other PR):

image

But I guess those numbers vary more so we can't conclude one is better than the other.

I think I prefer this PR because there are less cases to handle. You?

asterite:
Actually, I'll first ask Akosh to profile it locally to see if we get the same results as https://github.com/noir-lang/noir/pull/7057#issuecomment-2590118186

TomAFrench:
Those seem to all be within margin of error.

> I think I prefer this PR because there are less cases to handle. You?

Yeah, I think as long as we're not feeling limited by this library then we can just stick with it. We know that we can easily chuck it if necessary.

aakoshh:
This has virtually the same profile:

Screenshot 2025-01-14 at 15 19 33

asterite:
Yes, soon. Akosh asked me a few minutes to try to profile this together with his other PR.

TomAFrench:
Will let you hit the merge button.

aakoshh:
This is what it looks like with this merged into https://github.com/noir-lang/noir/pull/7053

There is a strange plateuing of the memory, instead of the pronounced peak with the other variant. 

Screenshot 2025-01-14 at 16 04 06

But overall it looks similar:
Screenshot 2025-01-14 at 16 07 59

PR #7059, opened 2025-01-14
chore: Use mimalloc as the default allocator

jfecher:
# Description

## Problem\*

Resolves 

## Summary\*

Testing to see if this improves performance

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/types.rs
+++ b/compiler/noirc_frontend/src/elaborator/types.rs
@@ -638,6 +638,30 @@ impl<'context> Elaborator<'context> {
 
                 let the_trait = self.interner.get_trait(constraint.trait_bound.trait_id);
                 if let Some(method) = the_trait.find_method(path.last_name()) {
+                    let show = |constraint: &TraitConstraint| {
+                        let name = &self.interner.get_trait(constraint.trait_bound.trait_id).name;
+                        let args = vecmap(
+                            &constraint.trait_bound.trait_generics.ordered,
+                            ToString::to_string,
+                        )
+                        .join(", ");
+                        let named = vecmap(&constraint.trait_bound.trait_generics.named, |t| {
+                            format!("{}: {}", t.name, t.typ)
+                        })
+                        .join(", ");
+
+                        if args.is_empty() && named.is_empty() {
+                            name.to_string()
+                        } else if args.is_empty() {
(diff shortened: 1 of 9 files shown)


jfecher:
Some good compilation time improvements here for such a small change but the memory reports are broken

ludamad:
Interesting, thought rust had quite a good allocator

compiler/noirc_frontend/src/elaborator/types.rs
```diff
@@ -638,6 +638,30 @@ impl<'context> Elaborator<'context> {
...
+                        } else if args.is_empty() {
+                            format!("{name}<{named}>")
+                        } else if named.is_empty() {
+                            format!("{name}<{args}>")
+                        } else {
+                            format!("{name}<{args}, {named}>")
+                        }
+                    };
+
+                    eprintln!("Returning assumed constraint: {}", show(&constraint));
```
aakoshh, line 664: Is this supposed to be here?

aakoshh:
FWIW I tried the profiling I do with _Instruments_ in https://github.com/noir-lang/noir/pull/7053 and it seems to have bombed (I'm on Mac):

Screenshot 2025-01-14 at 15 32 54

compiler/noirc_frontend/src/elaborator/types.rs
jfecher: Nope, looks like some debugging for a separate issue leaked in

TomAFrench:
Yeah, [heaptrack](https://milianw.de/blog/heaptrack-a-heap-memory-profiler-for-linux.html) reports that it doesn't work with custom allocators so we may need to replace this tooling for something else if we go forwards with this.

jfecher:
@ludamad Rust just uses the default system allocator

jfecher:
Looks like valgrind's massif tool isn't working either

jfecher:
Profiling with `MIMALLOC_SHOW_STATS=1` I also see the 12GB peak usage. Switching this PR to try out TCMalloc instead

jfecher:
This could be something to continue looking into for the future but I currently can't seem to profile memory usage of jemalloc, and any alternate allocator in general would break our CI so I'm closing this for now despite the possible performance gains.

Issue #6735, opened 2024-12-08
feat(ssa): Update the Brillig constraint SSA check to register descendants coming before the Brillig calls


rkarabut:
### Problem

As mentioned in https://github.com/noir-lang/noir/pull/6658#pullrequestreview-2482103067, currently the check's algorithm of collecting tainted value ids doesn't allow to register the descendant values should they come up _before_ the Brillig call, which could lead to false positives, as in

```
fn main(a: Field, b: Field) {
  let c = a + b;
  let d = unsafe { add(a, b) };
  assert_eq(c, d);
}
```

### Happy Case

Cases like this should be covered, possibly with a lookback pass for the calls found unconstrained by the end of the block.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

Yes

Issue #6713, opened 2024-12-05
feat(ssa): Update the missing Brillig constraint SSA check to process EnableSideEffectsIf


rkarabut:
### Problem

EnableSideEffectsIf instruction isn't handled by the check although it can lead to false positives, leaving Brillig call results descendants affected untainted.

### Happy Case

EnableSideEffectsIf should be handled correctly, marking the results affected by a condition using Brillig output as also using Brillig output.

### Workaround

None

### Additional Context

The issue has been brought up in https://github.com/noir-lang/noir/pull/6658#discussion_r1868449174.

### Project Impact

None

### Would you like to submit a PR for this Issue?

Yes

Issue #6706, opened 2024-12-05
fix(ssa): Update the missing Brillig constraint SSA check to cover chained Brillig calls


rkarabut:
### Problem

At the time, this program would make the pass complain at the `foo(x)` call being insufficently constrained:

```
struct Animal {
    legs: Field,
    eyes: u8,
    tag: Tag,
}

struct Tag {
    no: Field,
}

unconstrained fn foo(x: Field) -> Animal {
    Animal {
        legs: 4,
        eyes: 2,
        tag: Tag { no: x }
    }
}

unconstrained fn bar(x: Animal) -> Animal {
    Animal {
        legs: x.legs,
        eyes: x.eyes,
        tag: Tag { no: x.tag.no + 1 }
    }
}

fn main(x: Field) -> pub Animal {
    let dog = bar(foo(x));
    assert(dog.legs == 4);
    assert(dog.eyes == 2);
    assert(dog.tag.no == x + 1);

    dog
}
```

### Happy Case

Brillig calls chained in this way should be considered one call for the purposes of the check. The results of the final call should be checked against an argument of the initial call. As a result, the above code should not be issuing a bug warning. 

### Workaround

None

### Additional Context

Brought up in https://github.com/noir-lang/noir/pull/6658#discussion_r1868428495

### Project Impact

None

### Would you like to submit a PR for this Issue?

Yes

PR #7060, opened 2025-01-14
feat(ssa): Expand feature set of the Brillig constraint check

### LINKED ISSUE #6698: feat(ssa): Update the missing Brillig constraint SSA check to cover individual array elements
### Problem

The pass implemented in https://github.com/noir-lang/noir/pull/6658 currently does not take into consideration specific array elements, instead accepting a constraint as valid if it involves a relevant array value id descendant at all.

### Happy Case

The pass also should work for individual array elements. E.g., for a Brillig function `foo(x: Field) -> [u8; 3]` every element of the return value array should be later constrained in some way. As it's a compile time check, the array elements would probably have to only be accessed through constant indices in the constraints.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

Yes

### Support Needs

_No response_

rkarabut:
# Description

This PR resolves a number of issues that came up in the initial introduction of the Brillig constraint check, expanding its feature set.

## Problem
#6698, #6706, #6713, #6735 

## Summary

Resolves #6698 by creating and maintaining additional tainted value sets for every result array element,  #6713 by tracking EnableSideEffectsIf condition values, and #6735 along with #6706 by introducing a quick pass gathering information on Brillig calls. New tests for every issue are included.

As fixing #6698 makes the `check_underconstrained_regression` test program (correctly) issue a bug warning now, the program has been updated to properly constrain the involved array elements. 

## Documentation

Check one:

# PR Checklist\*


TomAFrench:
image

This is a pretty brutal increase, we may need to switch off this check by default if this is unavoidable.

rkarabut:
> This is a pretty brutal increase, we may need to switch off this check by default if this is unavoidable.

Indeed, would have to look into it

TomAFrench:
> https://github.com/noir-lang/noir/issues/6698, https://github.com/noir-lang/noir/issues/6706, https://github.com/noir-lang/noir/issues/6713, https://github.com/noir-lang/noir/issues/6735

Can you mark which of these this PR closes?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -99,9 +99,9 @@ struct DependencyContext {
     block_queue: Vec<BasicBlockId>,
     // Map keeping track of values stored at memory locations
     memory_slots: HashMap<ValueId, ValueId>,
-    // Map of values resulting from array get instructions
-    // to the actual array values
-    array_elements: HashMap<ValueId, ValueId>,
+    // Value currently affecting every instruction because
+    // of its involvement in an EnableSideEffectsIf condition
```
michaeljklein, line 103: Please explain a bit more in this comment: I have a decent idea of what you mean from the previous comment `Map of values resulting from array get instructions to the actual array values` and from notes on the PR, but I don't expect this to be clear out of context.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -116,7 +116,9 @@ struct BrilligTaintedIds {
     arguments: HashSet<ValueId>,
     // Results status
     results: Vec<ResultStatus>,
-    // Initial result value ids
+    // Indices of the array elements into the results vector
```
michaeljklein, line 119:
```suggestion
    // Indices of the array elements in the results vector
```

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -128,17 +130,59 @@ enum ResultStatus {
...
 
 impl BrilligTaintedIds {
-    fn new(arguments: &[ValueId], results: &[ValueId]) -> Self {
+    fn new(function: &Function, arguments: &[ValueId], results: &[ValueId]) -> Self {
+        // Exclude numeric constants
+        let arguments: Vec<ValueId> = arguments
+            .iter()
+            .copied()
+            .filter(|value| function.dfg.get_numeric_constant(*value).is_none())
+            .map(|value| function.dfg.resolve(value))
```
michaeljklein, line 139: Can `copied` be called after the `filter` (since you're already dereferencing the `value`) to avoid a few copies? Also in `results` below.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -231,6 +303,40 @@ impl DependencyContext {
...
+                        let mut wrapped_call_found = false;
+                        for (_, tainted_call) in self.tainted.iter_mut() {
+                            if current_tainted.is_wrapper(tainted_call) {
+                                tainted_call.update_results_children(results);
+                                wrapped_call_found = true;
+                                break;
+                            }
+                        }
+
+                        if !wrapped_call_found {
```
michaeljklein, line 330: If I'm reading this correctly, it's equivalent to the following pseudocode:
```rust
if !self.tainted.any(current_tainted.is_wrapper)
```

But wouldn't it make sense to record the current call unless _all_ of the tainted calls were wrapped, i.e. appx.
```rust
if !self.tainted.all(current_tainted.is_wrapper)
```
?

michaeljklein (changes requested):

rkarabut:
> Can you mark which of these this PR closes? I'd like to avoid dangling issues.

Should be all of them

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Done

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: No, it's indeed enough for the current call to be a wrapper over _any_ of the tracked calls to be excluded, e.g. in case `foo(bar(x))` is met `foo()` doesn't need to be tracked, as `bar(x)` tracking would be enough now given that we record `foo(bar(x))` results as descendants of `bar(x)`. Should this be made clearer maybe, or are the comments enough?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Comment updated

michaeljklein (approved):

rkarabut:
@TomAFrench should we merge with the current (greatly reduced but still out there) performance hit? I've dealt with the main culprit which turned out to be an actual bug but the check is still slower now just because of increased complexity. I believe I can bring it further down with some optimization, or otherwise make certain new features optional. The question is, do we need that in this PR or can it be taken care of separately?

rkarabut:
@TomAFrench with the latest fixes and optimizations as well as introducing limiting constraint search depth, only `rollup-base-private` seems to be slightly exceeding the 1.2 compilation time threshold on CI at the time:
изображение

Is this alright? If so I'll ask @michaeljklein to review the changes introduced and merge.

TomAFrench:
Currently testing this on `rollup-block-root`

TomAFrench:
Merging this but we're likely going to have to make this pass opt-in due to it hanging on larger circuits.

rkarabut:
> Merging this but we're likely going to have to make this pass opt-in due to it hanging on larger circuit

Let's see if disabling just the lookback part helps. I'll make an issue for this

PR #7067, opened 2025-01-14
feat!: Handle generic fields in `StructDefinition::fields` and move old functionality to `StructDefinition::fields_as_written`

jfecher:
# Description

## Problem\*

Resolves an issue from a private slack request. Needed for implementing `derive_serialize` recursively so that struct field types have their generics substituted correctly.

## Summary\*

Adds a `generic_args` argument to `StructDefinition::fields` to instantiate the field types with. This means this PR is breaking since this method used to require only `self`.

Since the old functionality of not instantiating can still be useful, I have moved it to a new function `fields_as_written` but wanted to emphasize that if users did need field types, and they don't know what the difference is, that they should prefer to use `fields`.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
@@ -529,7 +529,9 @@ fn struct_def_fields(
 }
 
 /// fn fields_as_written(self) -> [(Quoted, Type)]
-/// Returns (name, type) pairs of each field of this StructDefinition
+/// Returns (name, type) pairs of each field of this StructDefinition.
+///
+/// Note that any generic arguments won't be applied: if you need them to be, use `fields`.
 fn struct_def_fields_as_written(
     interner: &mut NodeInterner,
     arguments: Vec<(Value, Location)>,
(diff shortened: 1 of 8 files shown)


compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -482,12 +485,55 @@ fn struct_def_has_named_attribute(
...
+
+    let typ = Type::Slice(Box::new(Type::Tuple(vec![
+        Type::Quoted(QuotedType::Quoted),
+        Type::Quoted(QuotedType::Type),
+    ])));
+    Ok(Value::Slice(fields, typ))
+}
+
+/// fn fields_as_written(self) -> [(Quoted, Type)]
+/// Returns (name, type) pairs of each field of this StructDefinition
```
michaeljklein, line 532:
```suggestion
/// Returns (name, type) pairs of each field of this StructDefinition.
///
/// Note that any generic arguments won't be applied: if you need them to be, use `fields`.
```
?

michaeljklein (approved):
LGTM, with small note r.e. using `fields` on `fields_as_written`

TomAFrench:
In future please coordinate with me if merging breaking changes which blow up all of our benchmarks. Having this dropped on me when noir syncs are blocked is not fun.

asterite:
Another thing that might have been good to do, given that this is a breaking change, is to also return the visibility of each field (we also have an issue to be able to set those visibilities when setting those fields)

jfecher:
@TomAFrench sorry about that, I mistakenly assumed the sync itself would be quick and it'd just break aztec-noir downstream which could be fixed fairly quickly. I'll definitely coordinate more in the future

PR #7068, opened 2025-01-14
chore: Add benchmarking dashboard

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR adds a new benchmarking dashboard at https://noir-lang.github.io/noir/dev/bench/. Results will be added to this on each push to master.

We'll still be benchmarking on each PR however comments will only be posted if there's a regression above a certain percentage (currently set to a quite aggressive 10% but we can dial that back if necessary)

I'll turn off the comment systems for metrics other than gate counts in a separate PR.

## Documentation\*

Check one:

# PR Checklist\*


test_programs/parse_memory.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+set -e
+
+# This script accepts a string representing the amount of memory allocatedas outputted by `heaptrack`
```
vezenovm, line 4:
```suggestion
# This script accepts a string representing the amount of memory allocated as outputted by `heaptrack`
```

test_programs/parse_memory.sh
vezenovm, line 1: A chore separate from this PR, we should add a scripts package to `test_programs`.

vezenovm (approved):

vezenovm:
For a follow-up it would be nice to have separate pages for our memory/execution memory and compilation/execution timing reports. The dashboard is a bit long with all of our reports. I haven't looked into `github-action-benchmark` deeply though as to whether this is possible.

TomAFrench:
Yeah, I agree that it's not the prettiest but I don't want to invest too much time into it as I'm planning to migrate to Charlie's tool once it's ready.

PR #7069, opened 2025-01-15
chore: Do not make new instruction if it hasn't changed

aakoshh:
# Description

## Problem\*

Followup for https://github.com/noir-lang/noir/pull/7053

## Summary\*

Adds a `Dfg::insert_instruction_and_results_if_simplified` method which receives the existing instruction ID from `constant_folding`, and if the instruction could not be simplified and it still matches the original value, then it's not re-inserted into the `DenseMap`, saving some memory.

In my reference test with `rollup-base-public` it reduced peak memory from 2.7GB to 2.1GB

I suppose a followup could be to try to _remove_ instructions which are no longer needed.

## Additional Context

Noticed that `constant_folding` is involved in the peak memory usage. According to the stack trace it allocates a lot of that memory to make instructions (appears twice, with 32.1% and 13.2%):

Screenshot 2025-01-14 at 23 56 39

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/dfg.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/dfg.rs
@@ -236,6 +236,24 @@ impl DataFlowGraph {
         ctrl_typevars: Option<Vec<Type>>,
         call_stack: CallStackId,
     ) -> InsertInstructionResult {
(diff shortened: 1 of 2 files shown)


jfecher (approved):
Could probably expand this to more passes too

PR #7073, opened 2025-01-15
fix: do not remove memory blocks used as brillig input

### LINKED ISSUE #7062: Panic when casting an array to a slice
### Bug

When I try to compile the following code:

```
fn main(args: [Field; 2]) {
    /// Safety: what?
    unsafe { store(args) };
}

pub unconstrained fn store(_: [Field]) {}
```

I get the following panic:

Image

When I make the function constrained:

```
fn main(args: [Field; 2]) {
    store(args);
}

pub fn store(_: [Field]) {}

```

all compiles fine.

**So the bug seems to be present specifically when casting to a slice which is an argument to an unconstrained function.**

### Nargo Version

nargo version = 1.0.0-beta.1 noirc version = 1.0.0-beta.1+56ed6e16b975f91a (git version hash: 56ed6e16b975f91a, is dirty: false)

guipublic:
# Description

## Problem\*

Resolves #7062 

## Summary\*
The unused memory optimisation was not checking for brillig inputs

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/acvm-repo/acvm/src/compiler/optimizers/unused_memory.rs
+++ b/acvm-repo/acvm/src/compiler/optimizers/unused_memory.rs
@@ -1,4 +1,4 @@
-use acir::circuit::{opcodes::BlockId, Circuit, Opcode};
+use acir::circuit::{brillig::BrilligInputs, opcodes::BlockId, Circuit, Opcode};
 use std::collections::HashSet;
 
 /// `UnusedMemoryOptimizer` will remove initializations of memory blocks which are unused.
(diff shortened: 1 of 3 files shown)


TomAFrench (approved):
Strange that we would be generating a memory block at all in this case. Can you open an issue to see if we can get it so that we're just passing in witnesses to this brillig function?

vezenovm:
> Strange that we would be generating a memory block at all in this case.

Anytime we initialize an array in ACIR we do call a `MemoryInit` opcode.

PR #7074, opened 2025-01-15
fix: allow calling trait impl method from struct if multiple impls exist

asterite:
# Description

## Problem

Makes [this](https://github.com/noir-lang/noir-bignum/pull/95#discussion_r1916374270) work

## Summary

Edit: this one is very had to fix. The main issue is that:
1. We have to solve `U60Repr::from2()`
2. That's a call, so we have to solve what the call is pointing to, that is, solve `U60Repr::from2`
3. That's a path lookup. In our code a path lookup, for functions, end up returning a single `FuncId`. In this case it's a trait method so there are potentially multiple `FuncId`, so we'd need to represent the result in a different way here. And once we do that, change the code to support this new result variant.

---

It seems we had logic in place to remove a method definition if it gives an error while adding it. When multiple trait impl methods existed for the same struct (in different impls), the second definition would erase both, though only from the `ModuleData` (I think the definitions still exist somewhere else so that's why using `Trait::` worked).

I'm not sure why that was done, but I changed it to not do that. All tests seem to pass, so either that was unneeded or there is a scenario that breaks with this change but it's untested.

## Additional Context

I also introduced refactors/renames to make the code a bit clearer.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/mod.rs
+++ b/compiler/noirc_frontend/src/elaborator/mod.rs
@@ -1508,23 +1508,15 @@ impl<'context> Elaborator<'context> {
                 // object types in each method overlap or not. If they do, we issue an error.
                 // If not, that is specialization which is allowed.
                 let name = method.name_ident().clone();
-                let result = if let Some(trait_id) = trait_id {
-                    module.declare_trait_function(name, *method_id, trait_id)
+                if let Some(trait_id) = trait_id {
+                    // We don't check the result here because for trait and trait impl functions
+                    // we already checked if a duplicate definition exists
+                    let _ = module.declare_trait_function(name, *method_id, trait_id);
                 } else {
-                    module.declare_function(name, method.def.visibility, *method_id)
+                    // For non-trait methods we'll check if there are duplicates below
+                    let _ =
+                        module.declare_function(name.clone(), method.def.visibility, *method_id);
                 };
-                if result.is_err() {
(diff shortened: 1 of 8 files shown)


asterite:
Closing as this PR is only part of the fix but doing the full fix is very hard.

TomAFrench:
@asterite can you make a new PR with tests from this one with `#[should_panic]` attributes? These should link off to an issue on this repo tracking that we need to fix these.

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1508,34 +1513,31 @@ impl<'context> Elaborator<'context> {
...
-                if result.is_err() {
-                    let existing = module.find_func_with_name(method.name_ident()).expect(
-                        "declare_function should only error if there is an existing function",
-                    );
-
-                    // Only remove the existing function from scope if it is from a trait impl as
-                    // well. If it is from a non-trait impl that should override trait impl methods
-                    // anyway so that Foo::bar always resolves to the non-trait impl version.
-                    if self.interner.function_meta(&existing).trait_impl.is_some() {
-                        module.remove_function(method.name_ident());
```
jfecher, line 1525: IIRC without this check the compiler would still call one of the trait methods on error that wasn't removed instead of erroring that there were two matching candidates or something. Could be wrong, this has been here a while

Issue #7075, opened 2025-01-15
Failed to find trait implementation


sirasistant:
### Aim

Implement an automatically derivable serialize trait.
I have outputted what the derive is generating, and it's not finding a trait and I don't know why:
```noir
pub trait Deserialize<let N: u32> {
    fn deserialize(serialized: [Field; N]) -> Self;
}
impl Deserialize<1> for Field {
    fn deserialize(serialized: [Field; 1]) -> Self {
        serialized[0]
    }
}
impl Deserialize<1> for u32 {
    fn deserialize(serialized: [Field; 1]) -> Self {
        serialized[0] as u32
    }
}
impl<T, let K: u32> Deserialize<K + 1> for Option<T>
where
    T: Deserialize<K>,
{
    fn deserialize(serialized: [Field; K + 1]) -> Self {
        let is_some = serialized[0] == 1;
        if is_some {
            let mut deserialized = [0; K];
            for i in 0..K {
                deserialized[i] = serialized[i + 1];
            }
            Self::some(T::deserialize(deserialized))
        } else {
            Self::none()
        }
    }
}
impl<T, let N: u32, let K: u32> Deserialize<K * N> for [T; N]
where
    T: Deserialize<K>,
{
    fn deserialize(serialized: [Field; K * N]) -> Self {
        let mut result = [std::mem::zeroed(); N];
        let mut offset = 0;
        for i in 0..N {
            let mut deserialized = [0; K];
            for j in 0..K {
                deserialized[j] = serialized[offset + j];
            }
            result[i] = T::deserialize(deserialized);
            offset += K;
        }
        result
    }
}
#[derive(Eq)]
struct Bar {
    a: [Field; 1],
    b: Option<Field>,
}

// Autogenerated implementation
impl<let a_serialized_length: u32, let b_serialized_length: u32> Deserialize<a_serialized_length + b_serialized_length> for Bar
where
    [Field; 1]: Deserialize<a_serialized_length>,
    Option<Field>: Deserialize<b_serialized_length>,
{
    fn deserialize(serialized: [Field; a_serialized_length + b_serialized_length]) -> Self {
        let mut offset = 0;
        let mut a_serialized = [0; a_serialized_length];
        for i in 0..a_serialized_length {
            a_serialized[i] = serialized[offset + i];
        }
        offset += a_serialized_length;
        let a = Deserialize::deserialize(a_serialized);
        let mut b_serialized = [0; b_serialized_length];
[...]

PR #7076, opened 2025-01-15
fix!: Include kind in `StructDefinition::generics` and fix derivation of Eq in structs with numeric generics

sirasistant:
# Description

## Problem\*

Eq derivation didn't work in structs with numeric generics

## Summary\*

This PR changes the implementation of `struct_def_generics` to return tuples of the type and the numeric type, if the type is of numeric kind.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
@@ -185,7 +185,9 @@ impl<'local, 'context> Interpreter<'local, 'context> {
             "struct_def_fields_as_written" => {
                 struct_def_fields_as_written(interner, arguments, location)
             }
-            "struct_def_generics" => struct_def_generics(interner, arguments, location),
+            "struct_def_generics" => {
+                struct_def_generics(interner, arguments, return_type, location)
+            }
             "struct_def_has_named_attribute" => {
                 struct_def_has_named_attribute(interner, arguments, location)
             }
@@ -445,10 +447,11 @@ fn struct_def_as_type(
     Ok(Value::Type(Type::Struct(struct_def_rc, generics)))
 }
 
-/// fn generics(self) -> [Type]
+/// fn generics(self) -> [(Type, Option<Type>)]
 fn struct_def_generics(
     interner: &NodeInterner,
     arguments: Vec<(Value, Location)>,
+    return_type: Type,
     location: Location,
 ) -> IResult<Value> {
     let argument = check_one_argument(arguments, location)?;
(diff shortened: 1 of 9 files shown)


noir_stdlib/src/meta/mod.nr
```diff
@@ -97,8 +97,28 @@ pub comptime fn make_trait_impl<Env1, Env2>(
...
+        let typ = f"{typ}".quoted_contents();
+
+        if numeric_type.is_some() {
+            let numeric_type = numeric_type.unwrap_unchecked();
+            // Numeric generics expect the name to be an ident, not a type
+            impl_generics = impl_generics.push_back(quote { let $typ : $numeric_type });
+        } else {
+            impl_generics = impl_generics.push_back(quote { $typ });
+            where_clause = where_clause.push_back(quote { $typ: $trait_name });
+        }
```
jfecher, line 117: Not sure this is the correct solution actually. Looking at what we had previously:
```noir
    let impl_generics = s.generics().map(|g| quote { $g }).join(quote {,});
```
We directly quoted the `g: Type` generics before despite them being types, not identifiers. So we already handle the case within the compiler that the type is an existing, already quoted type. I think the right fix here would actually be to investigate why the compiler has an existing quoted type but doesn't get the Kind right in your original issue.

We could still use the Kind returned with each generic (maybe as an alternate method?) but I don't think it's necessary to solve this issue. I can investigate why we're losing the old Kind of the quoted type later but am still trying to figure out https://github.com/noir-lang/noir/issues/7075 currently.

jfecher (changes requested):
Not sure this

noir_stdlib/src/meta/mod.nr
jfecher: I say this in part because I think going forward with this solution would complicate all metaprogramming using generics to have to handle unintuitive conversions between String and Type representations

sirasistant: No no you understood right, the problem is the two representations. It's fine to block this until we can find a clean solution

jfecher: I mean before we did have
```noir
let where_clause = s.generics().map(|name| quote { $name: $trait_name }).join(quote {,});
```
Which I don't see working if it generates things like `N: Eq`

sirasistant: Let me see if it works without generating the numeric generics with let, but keeping the removal of the trait constraints for the numeric

sirasistant: Ah yes, I think you were right! We don't need to generate the numeric generics from idents, the type already contains the kind, so no `let N: u32` structure is necessary. It just works by checking wether it's generic or not and skipping the numerics in the where clause that way :)

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -445,22 +447,50 @@ fn struct_def_as_type(
...
+    ])));
+
+    let actual = return_type.clone();
+
+    let slice_item_type = match return_type {
+        Type::Slice(item_type) => *item_type,
+        _ => return Err(InterpreterError::TypeMismatch { expected, actual, location }),
+    };
+
+    let option_typ = match &slice_item_type {
```
sirasistant, line 474: Regarding the rust implementation of the builtin, I needed to fetch the option_typ from the return_type. This is the only way right? I didn't know how to construct the option type otherwise

jfecher: This is correct, we can't do a lookup since there may be many types with that name and we don't know what scope this is called from. Our other functions returning option just ended up being simpler than this one (didn't have to unwrap the return type as much).

jfecher (approved):
LGTM. I double checked that all our derive_vias in the stdlib use `make_trait_impl` internally so they're all fixed by this change, not just Eq.

PR #7079, opened 2025-01-15
feat(LSP): auto-import trait reexport if trait is not visible

asterite:
# Description

## Problem

Part of #6956

## Summary

A trait might be private in a module but publicly reexported from a parent module. That trait can be used but referring to it via the module it's in, rather than the module it's reexported from, is an error.

This PR makes LSP suggest a visible reexport if a trait is not visible where it's defined. For example, there's `std::ops::arith::Add` in the standard library, which is actually hidden but reexported as `std::ops::Add`. Here's how it works now when autocompleting `add`:

lsp-auto-import-trait-via-reexport

And the same thing is done in the code action that suggests importing a trait:

lsp-code-action-import-trait-via-reexport

## Additional Context

Rust Analyzer works the same way. I initially thought we could hardcode it just for the stdlib ops, but Rust Analyzer managed to do it with any traits I defined so I figured they track reexports and suggest those.

## Documentation

Check one:

# PR Checklist


TomAFrench (approved):
LGTM

PR #7081, opened 2025-01-15
feat(ssa): Globals DIE

vezenovm:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/pull/7021#issuecomment-2593646460

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
+++ b/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
@@ -86,8 +86,15 @@ impl<'block, 'global, Registers: RegisterAllocator> BrilligBlock<'block, 'global
         brillig_block.convert_block(dfg);
     }
 
-    pub(crate) fn compile_globals(&mut self, globals: &DataFlowGraph) {
+    pub(crate) fn compile_globals(
+        &mut self,
+        globals: &DataFlowGraph,
+        used_globals: &HashSet<ValueId>,
+    ) {
         for (id, value) in globals.values_iter() {
+            if !used_globals.contains(&id) {
+                continue;
+            }
             match value {
                 Value::NumericConstant { .. } => {
                     self.convert_ssa_value(id, globals);
--- a/compiler/noirc_evaluator/src/ssa/opt/die.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/die.rs
@@ -24,7 +24,26 @@ impl Ssa {
     /// unused results.
     #[tracing::instrument(level = "trace", skip(self))]
     pub(crate) fn dead_instruction_elimination(mut self) -> Ssa {
-        self.functions.par_iter_mut().for_each(|(_, func)| func.dead_instruction_elimination(true));
+        let mut used_global_values: HashSet<_> = self
+            .functions
+            .par_iter_mut()
+            .map(|(_, func)| func.dead_instruction_elimination(true))
+            .flatten()
+            .collect();
+
+        // Check which globals are used across all functions
+        for (id, value) in self.globals.dfg.values_rev_iter() {
+            if used_global_values.contains(&id) {
+                if let Value::Instruction { instruction, .. } = &value {
+                    let instruction = &self.globals.dfg[*instruction];
+                    instruction.for_each_value(|value_id| {
+                        dbg!(value_id);
+                        used_global_values.insert(value_id);
+                    });
+                }
+            }
+        }
+        self.used_global_values = used_global_values;
 
         self
     }
@@ -37,7 +56,10 @@ impl Function {
     /// instructions that reference results from an instruction in another block are evaluated first.
     /// If we did not iterate blocks in this order we could not safely say whether or not the results
     /// of its instructions are needed elsewhere.
-    pub(crate) fn dead_instruction_elimination(&mut self, insert_out_of_bounds_checks: bool) {
+    pub(crate) fn dead_instruction_elimination(
+        &mut self,
+        insert_out_of_bounds_checks: bool,
+    ) -> HashSet<ValueId> {
         let mut context = Context::default();
         for call_data in &self.dfg.data_bus.call_data {
             context.mark_used_instruction_results(&self.dfg, call_data.array_id);
@@ -58,11 +80,12 @@ impl Function {
         // instructions (we don't want to remove those checks, or instructions that are
         // dependencies of those checks)
         if inserted_out_of_bounds_checks {
-            self.dead_instruction_elimination(false);
-            return;
+            return self.dead_instruction_elimination(false);
         }
 
         context.remove_rc_instructions(&mut self.dfg);
+
+        context.used_values.into_iter().filter(|value| self.dfg.is_global(*value)).collect()
     }
 }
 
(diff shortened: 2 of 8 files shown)


compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -24,7 +24,26 @@ impl Ssa {
...
+            .flatten()
+            .collect();
+
+        // Check which globals are used across all functions
+        for (id, value) in self.globals.dfg.values_rev_iter() {
+            if used_global_values.contains(&id) {
+                if let Value::Instruction { instruction, .. } = &value {
+                    let instruction = &self.globals.dfg[*instruction];
+                    instruction.for_each_value(|value_id| {
+                        dbg!(value_id);
```
vezenovm, line 40:
```suggestion
```

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -37,7 +55,10 @@ impl Function {
     /// instructions that reference results from an instruction in another block are evaluated first.
     /// If we did not iterate blocks in this order we could not safely say whether or not the results
     /// of its instructions are needed elsewhere.
-    pub(crate) fn dead_instruction_elimination(&mut self, insert_out_of_bounds_checks: bool) {
+    pub(crate) fn dead_instruction_elimination(
+        &mut self,
+        insert_out_of_bounds_checks: bool,
+    ) -> HashSet<ValueId> {
```
aakoshh, line 61: Can you add some explanation to the docstrings about the new return value? Are the globals the only reasonable thing to return here?

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -24,7 +24,25 @@ impl Ssa {
...
     pub(crate) fn dead_instruction_elimination(mut self) -> Ssa {
-        self.functions.par_iter_mut().for_each(|(_, func)| func.dead_instruction_elimination(true));
+        let mut used_global_values: HashSet<_> = self
+            .functions
+            .par_iter_mut()
+            .flat_map(|(_, func)| func.dead_instruction_elimination(true))
+            .collect();
+
+        // Check which globals are used across all functions
+        for (id, value) in self.globals.dfg.values_rev_iter() {
```
aakoshh, line 34: Why does it matter if it iterates in reverse or forward?

compiler/noirc_evaluator/src/ssa/opt/die.rs
vezenovm: I will add some comments. We do not need anything but the used globals as all other values are going to be per function.

compiler/noirc_evaluator/src/ssa/opt/die.rs
vezenovm: Let's take this example for used globals here:
```
g0 = Field 1
g1 = make_array [Field 1, Field 1] : [Field; 2]
g2 = Field 0
g3 = make_array [Field 0, Field 0] : [Field; 2]
g4 = make_array [g1, g3] : [[Field; 2]; 2]
g5 = 1000000
```
In this case g0 is only used in g1, g2 is only used in g3, while g1 and g3 are only used in g4. Only g4 is used in our functions. If I start at g0 I won't mark correctly that it is used as I gate checking value with `used_global_values.contains(&id)`. I would then go to g1, which also is not marked as used so I would skip over it. This would continue until g4. I would correctly then mark g1 and g3 as being used, but the internals of g1 and g3 would not be marked as used. This same logic applies to the main DIE pass itself. When parsing a block's instructions we iterate in reverse due to what I laid out above.

compiler/noirc_evaluator/src/ssa/opt/die.rs
vezenovm: Added a comment

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -86,8 +86,15 @@ impl<'block, 'global, Registers: RegisterAllocator> BrilligBlock<'block, 'global
...
     }
 
-    pub(crate) fn compile_globals(&mut self, globals: &DataFlowGraph) {
+    pub(crate) fn compile_globals(
+        &mut self,
+        globals: &DataFlowGraph,
+        used_globals: &HashSet<ValueId>,
+    ) {
         for (id, value) in globals.values_iter() {
+            if !used_globals.contains(&id) {
```
TomAFrench, line 95: Seems like we could iterate through used_globals instead which would avoid the need for filtering.

TomAFrench: That said, this would be a root for indeterminism it seems

compiler/noirc_evaluator/src/ssa/ssa_gen/program.rs
```diff
@@ -11,13 +12,16 @@ use crate::ssa::ir::{
...
+use super::ValueId;
+
 /// Contains the entire SSA representation of the program.
 #[serde_as]
 #[derive(Serialize, Deserialize)]
 pub(crate) struct Ssa {
     #[serde_as(as = "Vec<(_, _)>")]
     pub(crate) functions: BTreeMap<FunctionId, Function>,
     pub(crate) globals: Function,
+    pub(crate) used_global_values: HashSet<ValueId>,
```
TomAFrench, line 24: Could we not mutate the globals Function to remove all unused values?

TomAFrench: It's not clear to me why we need to continue to track these unused globals

TomAFrench:
Generally looks good to me, some nits. I think it's worth merging this into the other SSA globals PR due to the compilation speed regressions in there that are solved by this one

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
TomAFrench: This comment is mooted if we implement the one below

compiler/noirc_evaluator/src/ssa/ssa_gen/program.rs
vezenovm: Yeah I wanted to do the same. But globals iterate over values (as we want to convert constants to Brillig opcodes not just instructions). Removing these would require a new map structure for globals: https://github.com/noir-lang/noir/blob/ad5a9808ab7c2a89beadb6f95c664441ee27a5f8/compiler/noirc_evaluator/src/ssa/ir/map.rs#L171

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
vezenovm: True I could iterate over `used_globals`. I could change it to a `BTreeSet` to avoid indeterminism. Let me know what you prefer.

vezenovm:
> I think it's worth merging this into the other SSA globals PR due to the compilation speed regressions in there that are solved by this one

Yeah agreed. I will wait for all comments to be resolved before merging.

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
TomAFrench: Eh, it's not causing significant slowdown for now. We can clean up later if necessary

PR #7088, opened 2025-01-16
feat(experimental): try to infer lambda argument types inside calls

asterite:
# Description

## Problem

No issue, just a limitation of the language before this PR, which is that lambdas in calls always require their parameters to have type annotations (unlike Rust).

## Summar

I've been thinking for days how we could have lambda parameter types be "inferred" from the call they are being passed to.

The first thing that came to my mind is that lambdas are most commonly passed as callbacks after invoking a method on `self`, where either `self` or a part of `self` is given to the lambda (like `Option::map`, `BoundedVec::map`, etc.). So the first thing that I tried in this PR is to eagerly unify a method call's function type with the object type. Then, when elaborating a lambda as a method call argument we pass the potential parameter types of a function type that is in that call position:

```noir
fn map<U>(&self, f: fn(T) -> U) { ... }
                    ^
                  this

self.map(|x| ...)
          ^
    has to unify with this
```

Then these types are unified, without erroring because later we'll check the type of the lambda against the argument anyway.

And... that worked! And that already covers a lot of cases.

Then I did the same thing for function calls, except that there's no `self`, but at least it now works if a callback has a concrete type, like if it's:

```noir
fn foo(f: fn(Foo) -> ...
```

And that worked too! Though I'm not sure there are many uses of that...

BUT: it didn't work in the code Nico shared in Slack, because it's a function call where the first argument given is like a self type, except that it's not a method all:

```noir
for_each_in_bounded_vec(
    discovered_notes,
    |discovered_note, _| {
    })
```

So the final thing I did was to eagerly try to unify argument types as we elaborate them against the target function type. And that made that example work!

It won't work if the lambda comes before the argument (which works in Rust) but I think that pattern is uncommon (though we could try to make it work in the future).

## Additional Context

I don't know if this is the right way to approach this.

I also don't know if unifying eagerly would cause any issues. One thing that's not done here is using "unify_with_coercions", but given that we don't issue the errors that happens in these eager checks, maybe it's fine (maybe it won't work in cases where an array is automatically converted to a slice, though I guess we could make it work in the future).

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/expressions.rs
+++ b/compiler/noirc_frontend/src/elaborator/expressions.rs
@@ -388,6 +388,8 @@ impl<'context> Elaborator<'context> {
     fn elaborate_call(&mut self, call: CallExpression, span: Span) -> (HirExpression, Type) {
         let (func, func_type) = self.elaborate_expression(*call.func);
 
+        let any_argument_is_lambda = call.arguments.iter().any(|arg| arg.kind.is_lambda());
+
         let mut arguments = Vec::with_capacity(call.arguments.len());
         let args: Vec<_> = call
             .arguments
@@ -403,6 +405,16 @@ impl<'context> Elaborator<'context> {
                 } else {
                     self.elaborate_call_argument_expression(arg, arg_index, &func_type)
                 };
+
+                if any_argument_is_lambda {
+                    // Try to unify this argument type against the function's argument type
+                    // so that a potential lambda following this argument can have more concrete types.
+                    if let Type::Function(func_args, _, _, _) = &func_type {
+                        if let Some(func_arg_type) = func_args.get(arg_index) {
+                            let _ = func_arg_type.unify(&typ);
+                        }
+                    }
+                }
 
                 arguments.push(arg);
                 (typ, arg, span)
@@ -479,14 +491,19 @@ impl<'context> Elaborator<'context> {
                     self.type_check_variable(function_name.clone(), function_id, generics.clone());
                 self.interner.push_expr_type(function_id, func_type.clone());
 
-                // Try to unify the object type with the first argument of the function.
-                // The reason to do this is that many methods that take a lambda will yield `self` or part of `self`
-                // as a parameter. By unifying `self` with the first argument we'll potentially get more
-                // concrete types in the arguments that are function types, which will later be passed as
-                // lambda parameter hints.
-                if let Type::Function(args, _, _, _) = &func_type {
-                    if !args.is_empty() {
-                        let _ = args[0].unify(&object_type);
+                let any_argument_is_lambda =
+                    method_call.arguments.iter().any(|arg| arg.kind.is_lambda());
+
+                if any_argument_is_lambda {
+                    // Try to unify the object type with the first argument of the function.
+                    // The reason to do this is that many methods that take a lambda will yield `self` or part of `self`
+                    // as a parameter. By unifying `self` with the first argument we'll potentially get more
+                    // concrete types in the arguments that are function types, which will later be passed as
+                    // lambda parameter hints.
+                    if let Type::Function(args, _, _, _) = &func_type {
+                        if !args.is_empty() {
+                            let _ = args[0].unify(&object_type);
+                        }
                     }
                 }
 
@@ -501,6 +518,17 @@ impl<'context> Elaborator<'context> {
                     let span = arg.span;
                     let (arg, typ) =
                         self.elaborate_call_argument_expression(arg, arg_index + 1, &func_type);
+
+                    if any_argument_is_lambda {
+                        // Try to unify this argument type against the function's argument type
+                        // so that a potential lambda following this argument can have more concrete types.
+                        if let Type::Function(func_args, _, _, _) = &func_type {
+                            if let Some(func_arg_type) = func_args.get(arg_index + 1) {
+                                let _ = func_arg_type.unify(&typ);
+                            }
+                        }
+                    }
+
                     arguments.push(arg);
                     function_args.push((typ, arg, span));
                 }
--- a/compiler/noirc_frontend/src/elaborator/expressions.rs
+++ b/compiler/noirc_frontend/src/elaborator/expressions.rs
@@ -52,7 +52,7 @@ impl<'context> Elaborator<'context> {
             ExpressionKind::If(if_) => self.elaborate_if(*if_),
             ExpressionKind::Variable(variable) => return self.elaborate_variable(variable),
             ExpressionKind::Tuple(tuple) => self.elaborate_tuple(tuple),
-            ExpressionKind::Lambda(lambda) => self.elaborate_lambda(*lambda),
+            ExpressionKind::Lambda(lambda) => self.elaborate_lambda(*lambda, None),
             ExpressionKind::Parenthesized(expr) => return self.elaborate_expression(*expr),
             ExpressionKind::Quote(quote) => self.elaborate_quote(quote, expr.span),
             ExpressionKind::Comptime(comptime, _) => {
@@ -472,6 +472,17 @@ impl<'context> Elaborator<'context> {
                     self.type_check_variable(function_name.clone(), function_id, generics.clone());
                 self.interner.push_expr_type(function_id, func_type.clone());
 
+                // Try to unify the object type with the first argument of the function.
+                // The reason to do this is that many methods that take a lambda will yield `self` or part of `self`
+                // as a parameter. By unifying `self` with the first argument we'll potentially get more
+                // concrete types in the arguments that are function types, which will later be passed as
+                // lambda parameter hints.
+                if let Type::Function(args, _, _, _) = &func_type {
+                    if !args.is_empty() {
+                        let _ = args[0].unify(&object_type);
+                    }
+                }
+
                 // These arguments will be given to the desugared function call.
                 // Compared to the method arguments, they also contain the object.
                 let mut function_args = Vec::with_capacity(method_call.arguments.len() + 1);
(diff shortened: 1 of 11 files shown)


asterite:
I didn't expect this to have any repercussion in SSA 😮 

Lambdas that involve math operations somehow compile before this change without type annotations:

```noir
let myarray: [i32; 3] = [1, 2, 3];

// Compiles fine without saying `n: i32`
assert(myarray.any(|n| n > 2));
```

I think it's because of the `>`.

But in the end `n` ends up being `i32` so I don't understand how it could change... I'll investigate it later today. Though I guess if it ends up in more optimizations that's good, though maybe it's suspicious and this introduced a bug... 🤔

asterite:
Also compilation memory and execution memory went up 🤔

asterite:
It seems with this PR some functions get inlined right from the beginning, while in master some are not. I'm not sure why... but if this only happens for higher-order functions (maybe uncommon?) and if it leads to an optimization, maybe it's good.

compiler/noirc_frontend/src/elaborator/expressions.rs
```diff
@@ -520,6 +586,40 @@ impl<'context> Elaborator<'context> {
...
     }
 
+    /// Elaborates an expression taking into account that it's a call argument in a function
+    /// that has the given type, and `arg_index` is the index of that argument in that function type.
+    fn elaborate_call_argument_expression(
+        &mut self,
+        arg: Expression,
+        arg_index: usize,
+        func_type: &Type,
+        any_argument_has_lambda_without_type_annotations: bool,
```
jfecher, line 596: > any_argument_has_lambda_without_type_annotations

We can remove this flag from the PR (not just this method), it shouldn't be needed if we push down the type instead of unify afterward

compiler/noirc_frontend/src/elaborator/expressions.rs
```diff
@@ -846,19 +946,44 @@ impl<'context> Elaborator<'context> {
...
+        let parameters: Vec<_> = lambda
+            .parameters
+            .into_iter()
+            .enumerate()
+            .map(|(index, (pattern, typ))| {
+                let parameter = DefinitionKind::Local(None);
+                let is_unspecified = matches!(typ.typ, UnresolvedTypeData::Unspecified);
+                let typ = self.resolve_inferred_type(typ);
+
+                if is_unspecified {
```
jfecher, line 972: Shouldn't need this check here, we can resolve the inferred type which will return a type variable if it was unspecified. Then we can just unify the result unconditionally.

jfecher (changes requested):
Instead of unifying afterward we can push the type down and that should be sufficient. This is what Rust does as well since it uses a different algorithm called "bidirectional type inference" which splits up type checking functions into `fn infer(T) -> Type` which we have, and `fn check(T, Type)` which we don't have. Certain constructs like literals are always inferred, and others like lambdas are usually checked.

For our purposes though we can just use your existing check arg type function but pass down the expected type instead of the entire function type and argument index.

asterite:
> Instead of unifying afterward we can push the type down and that should be sufficient

Some weeks ago I tried something similar to what I did here but I can't remember what (it didn't work so I deleted the branch). I thought I was assigning types instead of unifying, and getting errors like "can't find method foo for T" and that's why I tried unifying here. But it's probably the case that I had a bug or something else and that's why it wasn't working. I'll try pushing the type down and using it when the type is unspecified 👍

asterite:
Hm, now we get a failure on code like this:

```noir
struct U60Repr<let N: u32, let NumSegments: u32> {}

impl<let N: u32, let NumSegments: u32> U60Repr<N, NumSegments> {
    fn new<let NumFieldSegments: u32>(_: [Field; N * NumFieldSegments]) -> Self {
        U60Repr {}
    }
}

fn main() {
    let input: [Field; 6] = [0; 6];
    let _: U60Repr<3, 6> = U60Repr::new(input);
}
```

The error is this:

```
error: Expected type [Field; ((6 / _) * _)], found type [Field; 6]

let u60repr: U60Repr<3, 6> = unsafe { U60Repr::new(input) };
                                                   -----
```

Looking at the types that are compares, these are the ones before this PR:

```
arg = [Field; (6: numeric u32)]
param = [Field; (Numeric(Shared(RefCell { value: Unbound('8, Numeric(u32)) }): u32) * Numeric(Shared(RefCell { value: Unbound('10, Numeric(u32)) }): u32))]
```

And these are the ones in this PR:

```
arg = [Field; (6: numeric u32)]
param = [Field; ((6 / _) * Numeric(Shared(RefCell { value: Unbound('10, Numeric(u32)) }): u32))]
```

It seems the compiler is trying to solve the equation and that's why there's a division. I don't know if this is a bug in the unification code or if it's a bug introduced in this PR.

jfecher:
@asterite we can't remove the occurs check, that'd break the check for infinite types like `a = (a, a)`. I think the error is stemming from our inability to solve `6 = ? * ?` because `N` and `NumSegments` are not also being pushed down, since they're only known from the let statement. My guess is before this change on arguments we did `?a = ?b * ?c` instead and could just set the value of `?a` to equal `?b * ?c` in that case. Then when they're constrained further later it'd be all fine.

asterite:
I see, thanks.

> I think the error is stemming from our inability to solve 6 = ? * ?

In the code before this PR, `[Field; 6]` was checked against `[Field; ?a * ?b]` and that somehow didn't seem to break. Maybe it's that it solved `?a` to be `6 / ?b`? I didn't check.

I think now it's checking `6 = (6 / ?a) * ?b`, maybe we don't have that case covered. But that equation could be turned into `6 / ?b = 6 / ?a` so `?a` and `?b` should be the same... not sure! I jumped into working on the `loop` statement, but I'll come back to this after that.

jfecher:
> // Handle cases like `4 = a + b` by trying to solve to `a = 4 - b`

I don't see anything wrong with that case aside from the fact the last `Ok(())` is unnecessary as you mentioned. I don't think it's the cause of the issue here at least.

> I think now it's checking 6 = (6 / ?a) * ?b, maybe we don't have that case covered.

This looks incorrect to me since as you mentioned it'd require `a = b` which shouldn't follow from the original equation of `6 = a * b`. So the error probably lies in how that constraint comes about

asterite:
Coming back to this, I found that it's trying to unify a `Constant` against an `InfixExpr` with these values:

```
# Infix
lhs = Numeric(Shared(RefCell { value: Unbound('10, Numeric(u32)) }): u32)
op = Multiplication
rhs = ((6: numeric u32) / Numeric(Shared(RefCell { value: Unbound('10, Numeric(u32)) }): u32))

# Constant
value = 6
```

So it's unifying `6` against `(6 / x) * x`. And effectively the `x` is the same in both cases.

What ends up happening, because of

```
// Handle cases like `4 = a + b` by trying to solve to `a = 4 - b`
```

is trying to unify `x` against `(6 / (6 / x))`, and that's where the error happens.

Now I'm trying to see whre that `(6 / x) * x` is coming from, because I think that could be just `6`.

jfecher:
The 6 / x * x is definitely a result of trying to solve the equation previously. It's unfortunate we can't just optimize that to `6` anymore like we used to before we considered integer division.

I wonder if we added a version of division that tells the compiler to ignore rounding if that'd fix this. E.g. assuming the original is `x = a * b` and the translation is `x / b = a` then we already know x should be divisible by b so we should be able to optimize a theoretical `x / b * b` later on.

asterite:
> It's unfortunate we can't just optimize that to 6 anymore like we used to before we considered integer division.

Yeah... that's what I tried at first but found that a test about that failed.

I wonder where that `(6 / x) * x` came from, because in the code there's only `N * NumFieldSegments`. Like, here there aren't two type variables that are the same that came from the user, it's the compiler duplicating them and putting them in the equation...

I tried gating `Type::InfixExpr(..)` behind a method to see where's that coming from, but all I can see it's coming from `canonicalize` which already takes an `InfixExpr` like that, and I have no idea why I can't find where it all starts 😕 

Some commits ago this PR worked, I guess because we didn't type-check as much as we do now (previously we'd only do it if there were lambdas). We could go back to that version, though I guess there would still be code that failed because of this if it was similar to the snippet that fails but also has a lambda argument in it (maybe uncommon). That is, I wonder if this bugs exists regardless of this feature and we could procede with this feature by only applying these changes if there are lambda arguments.

asterite:
That code is currently erroring on `type_check_call` because `func_type` changed after we unified its types with the argument types.

I wanted to try to call `type_check_call` on a fresh `func_type` without all the type information we added from arguments... but I didn't know how. I cloned `func_type` before type-checking the arguments and passed but it didn't work.

asterite:
I think I found a way to fix this, in a way that might be the "correct" way to do it.

asterite:
@jfecher I pushed a commit with a fix that feels intuitive to me but I'm not totally sure it's correct.

Essentially, at one point the compiler is trying to unify `6 = a * b` so it translates that to `a = 6 / b`. Later on it tries to unify `6 = b * (6 / b)` and it fails because it can't simplify `b * (6 / b)` to `b` (as we know that's not always possible).

However, that `6 / b` doesn't exist in code, it came from transforming `6 = a * b` to `a = 6 / b`. So I changed the compiler to keep track of this fact (it's carried in `InfixExpr`).

Then, when the `InfixExpr` `b * (6 / b)` is created, if we know `6 / b` was created as an "inversion" of another expression, we can be sure that `b * (6 / b)` is `6`, because there was never a division in the first place (this is the part I'm slightly hesitant about, but feels intuitively correct to me).

jfecher:
@asterite yep, that's the solution I was hinting at with:

> I wonder if we added a version of division that tells the compiler to ignore rounding if that'd fix this. E.g. assuming the original is x = a * b and the translation is x / b = a then we already know x should be divisible by b so we should be able to optimize a theoretical x / b * b later on.

I'm not sure if we need it for more than just division but it should do the same thing either way so whichever approach is cleaner.

jfecher (approved):
PR looks good.

While reviewing it I was wondering if it'd be cleaner to represent the inverse flag as an additional division operator instead like my earlier comment. I noticed a few things would be different:
- The unification check checks that `op_a == op_b` so we'd return false there if we ever had a `a / b` unified against the inversed division `a // b`. It seems like it'd be better to me if they were treated as equal? Not sure if this'd ever lead to any hard to find bugs with arithmetic types.
- We don't check the inverse flag on Type::eq or hash. Again probably fine since this is an internal flag only used for simplification of these types.
- Notably existing checks in `inverse` could return true that `//` is an inverse of `*` for canonicalizations that care about this.

And another point unrelated to the above
- This isn't needed for this PR (so I won't block on it) but I think the check to eliminate inversed types should be along with the other checks in `canonicalize`. The reason for this is that types can change when they're unified so what was `a // b` may become `(b * c) // b` which we'd want to simplify. This scenario may not be possible currently though since I think we only create these inversed operators during unify when we know we have `constant = a <op> b` rather than `constant = a` where a could later unify to an infix expression itself.

jfecher:
> I started implementing it but I'm not sure... there's inverse() and approx_inverse(). For Multiplication in approx_inverse() now we have two operations that could undo that: both types of divisions. So we can't check anymore if an op is the inverse of just another one, we'd have to check all of them, or maybe have methods is_approx_inverse() and is_inverse().

Yeah, we'd probably have to change the method so it accepts an operator and returns a boolean instead so it could match on multiple. The main concern was the existing canonicalizations that call `inverse` won't see an inversed division as the inverted form of a multiplication still.

asterite:
@jfecher Would you prefer me to try the new division operator in this PR, or leave it to a follow-up PR? I'm fine either way but, as you said, it's unrelated to the original PR so I don't know if this approach is good-enough for now and we can later improve it.

jfecher:
@asterite I'm not sure it'd be an improvement or not given my past reasoning. We can leave it to a later PR

PR #7096, opened 2025-01-16
feat: `loop` keyword in runtime and comptime code

asterite:
# Description

## Problem

No tracking issue, but implements `loop` in unconstrained and comptime code

## Additional Context

Note that right now this doesn't work if there's no `break` inside a loop:
- `Function::returns()` in SSA expect a return to be there
- Inlining breaks because it can't find a return terminator

I also had to change some unrolling code that assumed there's an induction variable or a single instruction in a loop header, because now that's not true.

I'm not sure how to handle the inlining code... should we abandon the assumption that functions always have a return terminator? If so, how does the inlining code logic needs to change to handle that?

## Documentation

Check one:

# PR Checklist


asterite:
It seems that if you have a loop like this:

```noir
loop {
  if false {
    break;
  }
}
```

then it leads to the same crashes. So detecting whether a loop has at least one `break` in the frontend isn't enough.

Maybe for now we can track this down as an issue. I tried allowing infinite loops but it eventually crashes with this:

```
Message:  ICE: Value not found in cache v13
Location: compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block_variables.rs:46
```

and I don't understand why it happens (that variable indeed doesn't exist in the final SSA so I don't know how it got there in the first place).

Another issue is that if you have an infinite loop in comptime code then it will make LSP hang because LSP, via nargo, will try to interpret that. I'll try to at least fix that by stopping at some iterations (maybe 10K), only if LSP is active.

asterite:
This is ready for review. I tracked the infinite loop bug here: https://github.com/noir-lang/noir/issues/7103

jfecher:
> It seems that if you have a loop like this:
```rust
loop {
  if false {
    break;
  }
}
```

We'll have to do break detection in SSA then since `if false { break }` will just be removed entirely in ssa-gen. I don't think we should merge `loop {}` with a panic like this that I think would be relatively common for users to accidentally write. If we still warn for `loop {}` as an experimental feature though then it is fine

asterite:
Sounds good, I'll make it an error if `loop` doesn't have at least one `break`.

I'll try to find infinite loops in SSA. One simple thing is seeing if a function doesn't have returns (I think it crashes during inlining, so we could place the check there).

However, this code loops forever:

```noir
fn main() {
    unsafe {
        foo(false);
    }
}

unconstrained fn foo(x: bool) {
    let mut i = 1;
    loop {
        println(i);
        i += 1;
        if x {
            break;
        }
    }
}
```

and it doesn't crash the compiler but I don't understand why yet. So there might be ways for users to craft infinite loops, but I don't know if the worry is about the compiler panicking, not actually creating infinite loops.

jfecher:
@asterite yeah, infinite loops are fine and shouldn't crash the compiler. The compiler is panicking because it expects the SSA to be in a certain format with a return, and it won't be in that format if the infinite loop has any possibly reachable breaks (even if that break relies on a condition that is always false - just as long as its not optimized out).

docs/docs/noir/concepts/control_flow.md
```diff
@@ -46,9 +45,26 @@ Alternatively, `start..=end` can be used for a range that is inclusive on both e
...
 The index for loops is of type `u64`.
 
+## Loops
+
+In unconstrained code, `loop` is allowed for loops that end after a certain condition is met
+(or to loop forever):
+
+```rust
+let mut i = 10
+loop {
```
jfecher, line 55: Maybe this should be introduced after the `Break and Continue` section since it requires break at least

compiler/noirc_frontend/src/elaborator/statements.rs
```diff
@@ -271,11 +271,25 @@ impl<'context> Elaborator<'context> {
...
+        self.push_scope();
+
+        let (block, _block_type) = self.elaborate_expression(block);
+
+        self.pop_scope();
+        self.nested_loops -= 1;
+
+        let statement = HirStatement::Loop(block);
+
+        (statement, Type::Unit)
```
jfecher, line 292: We should have loops return the value that `break` was used with if possible. We can save this for a later PR though.

compiler/noirc_frontend/src/hir/comptime/errors.rs
```diff
@@ -683,6 +687,13 @@ impl<'a> From<&'a InterpreterError> for CustomDiagnostic {
                 let secondary = String::new();
                 CustomDiagnostic::simple_error(msg, secondary, location.span)
             }
+            InterpreterError::LoopHaltedForUiResponsiveness { location } => {
+                let msg = "This loop took too much time to execute so it was halted for UI responsiveness"
+                    .to_string();
+                let secondary =
+                    "This error doesn't happen in normal executions of `nargo`".to_string();
+                CustomDiagnostic::simple_error(msg, secondary, location.span)
```
jfecher, line 695: This should be a warning I think since its not something actually wrong with the program

compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs
```diff
@@ -587,6 +588,38 @@ impl<'a> FunctionContext<'a> {
...
+    /// Codegens a loop, creating three new blocks in the process.
+    /// The return value of a loop is always a unit literal.
+    ///
+    /// For example, the loop `loop { body }` is codegen'd as:
+    ///
+    /// ```text
+    ///   br loop_body()
+    /// loop_body():
+    ///   v3 = ... codegen body ...
+    ///   br loop_body(v4)
```
jfecher, line 600:
```suggestion
    ///   br loop_body(v3)
```

jfecher (approved):

asterite:
> If you just want to keep marking this as experimental though it'd make the PRs easier to review

How do we mark that this is experimental?

docs/docs/noir/concepts/control_flow.md
asterite: Done! I also added the same explanation as `break` and `continue` about why `loop` can't be used in constrained code.

compiler/noirc_frontend/src/elaborator/statements.rs
asterite: Good point. I think `break` doesn't allow expressions yet. And if we make this change, maybe `for` should also allow that.

compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs
asterite: Good catch! I changed it to not pass any value.

jfecher:
@asterite emit an ExperimentalFeature warning in the parser whenever the loop keyword is used

compiler/noirc_frontend/src/elaborator/statements.rs
jfecher: `for` can't also allow that since like in rust it may exit without a break, so there'd be no value to return. We could at most return a `Option<T>` but that gets awkward - we should just keep following Rust here.

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
```diff
@@ -1741,6 +1742,34 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
...
+        let in_lsp = self.elaborator.interner.is_in_lsp_mode();
+        let mut counter = 0;
+
+        loop {
+            self.push_scope();
+
+            match self.evaluate(expr) {
+                Ok(_) => (),
+                Err(InterpreterError::Break) => break,
+                Err(InterpreterError::Continue) => continue,
```
aakoshh, line 1756: Shouldn't `pop_scope` be called before we `continue`?

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
```diff
@@ -1741,6 +1742,34 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
...
+        let was_in_loop = std::mem::replace(&mut self.in_loop, true);
+        let in_lsp = self.elaborator.interner.is_in_lsp_mode();
+        let mut counter = 0;
+
+        loop {
+            self.push_scope();
+
+            match self.evaluate(expr) {
+                Ok(_) => (),
+                Err(InterpreterError::Break) => break,
```
aakoshh, line 1755: Shouldn't `pop_scope` be called before we `break`? Otherwise we exit `evaluate_loop` and leave an extra entry in `comptime_scopes`.

Issue #6963, opened 2025-01-06
Add feature flags to Noir


michaeljklein:
### Problem

- It's ergonomic to include assertions and debugging helpers within the functions/methods to test, but this approach currently either requires duplication or ad-hoc abstractions to enable/disable code.
- Most dead code isn't eliminated until fairly late in compilation (?)

See description of how [features work in Rust](https://doc.rust-lang.org/cargo/reference/features.html).

### Happy Case

Minimal useful version:
- Support declaring features in `Nargo.toml`
  + `[features]` section
  + `feature_name = ["included_feature1", "included_feature2", ..]`
- Support `#[cfg(feature = "feature_name")]` on top-level declarations and filter out unused declarations after parsing
- Support `let _: bool = cfg!(feature = "feature_name");` or similar for use inside declarations
- Support `nargo` CLI option: `--features "feat1 feat2"`

Outside original PR:
- Support [`all/any/not`](https://doc.rust-lang.org/reference/conditional-compilation.html) in `cfg`
- `default` feature
- Support other CLI options: `--all-features`, `--no-default-features`

### Workaround

Yes

### Workaround Description

Maintain multiple versions of declarations that need different features.

### Project Impact

Nice-to-have

### Would you like to submit a PR for this Issue?

None

jfecher:
Wonder if this will be compatible with our current use of attributes & macros. Since we collect & typecheck functions before we apply attributes to them, if a function is `#[cfg(...)]`'d out of the program, we'd still issue errors for it. E.g. "duplicate function with name foo", or feature-specific functions not being available within its body, etc.

PR #7097, opened 2025-01-16
feat: support cfg(feature = "name") attributes

michaeljklein:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6963

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/lexer/token.rs
+++ b/compiler/noirc_frontend/src/lexer/token.rs
@@ -904,6 +904,9 @@ pub enum SecondaryAttribute {
 
     /// Allow chosen warnings to happen so they are silenced.
     Allow(String),
+
+    // A #[cfg(..)] attribute
+    Cfg(CfgAttribute),
 }
 
 impl SecondaryAttribute {
@@ -921,6 +924,7 @@ impl SecondaryAttribute {
             SecondaryAttribute::Varargs => Some("varargs".to_string()),
             SecondaryAttribute::UseCallersScope => Some("use_callers_scope".to_string()),
             SecondaryAttribute::Allow(_) => Some("allow".to_string()),
+            SecondaryAttribute::Cfg(_) => Some("cfg".to_string()),
         }
     }
 
@@ -950,6 +954,7 @@ impl SecondaryAttribute {
             SecondaryAttribute::Varargs => "varargs".to_string(),
             SecondaryAttribute::UseCallersScope => "use_callers_scope".to_string(),
             SecondaryAttribute::Allow(ref k) => format!("allow({k})"),
+            SecondaryAttribute::Cfg(ref cfg_attribute) => format!("cfg({cfg_attribute})"),
         }
     }
 }
@@ -975,6 +980,25 @@ impl Display for MetaAttribute {
             let args =
                 self.arguments.iter().map(ToString::to_string).collect::<Vec<_>>().join(", ");
             write!(f, "{}({})", self.name, args)
+        }
+    }
+}
+
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub enum CfgAttribute {
+    // feature = "{name}"
+    Feature {
+        name: String,
+        span: Span,
+    },
+}
+
+impl Display for CfgAttribute {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            CfgAttribute::Feature { name, span: _ } => {
+                write!(f, "feature = {:?}", name)
+            }
         }
     }
 }
@@ -1010,6 +1034,7 @@ pub enum Keyword {
     Bool,
     Break,
     CallData,
+    Cfg,
     Char,
     Comptime,
     Constrain,
@@ -1021,6 +1046,7 @@ pub enum Keyword {
     Else,
     Enum,
     Expr,
+    Feature,
     Field,
     Fn,
     For,
@@ -1069,6 +1095,7 @@ impl fmt::Display for Keyword {
             Keyword::AssertEq => write!(f, "assert_eq"),
             Keyword::Bool => write!(f, "bool"),
             Keyword::Break => write!(f, "break"),
+            Keyword::Cfg => write!(f, "cfg"),
             Keyword::Char => write!(f, "char"),
             Keyword::CallData => write!(f, "call_data"),
             Keyword::Comptime => write!(f, "comptime"),
@@ -1081,6 +1108,7 @@ impl fmt::Display for Keyword {
             Keyword::Else => write!(f, "else"),
             Keyword::Enum => write!(f, "enum"),
             Keyword::Expr => write!(f, "Expr"),
+            Keyword::Feature => write!(f, "feature"),
             Keyword::Field => write!(f, "Field"),
             Keyword::Fn => write!(f, "fn"),
             Keyword::For => write!(f, "for"),
@@ -1133,6 +1161,7 @@ impl Keyword {
             "bool" => Keyword::Bool,
             "break" => Keyword::Break,
             "call_data" => Keyword::CallData,
+            "cfg" => Keyword::Cfg,
             "char" => Keyword::Char,
             "comptime" => Keyword::Comptime,
             "constrain" => Keyword::Constrain,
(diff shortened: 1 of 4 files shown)


compiler/noirc_frontend/src/parser/mod.rs
```diff
@@ -132,6 +136,7 @@ pub struct Item {
     pub kind: ItemKind,
     pub location: Location,
     pub doc_comments: Vec<String>,
+    pub cfg_feature_disabled: bool,
```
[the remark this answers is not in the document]
asterite: (that is, it could be done together with https://github.com/noir-lang/noir/issues/7574 )

compiler/noirc_frontend/src/parser/parser/attributes.rs
```diff
@@ -633,6 +680,46 @@ mod tests {
...
+        assert!(parse_attribute_result.is_some());
+        assert_eq!(parser.errors.len(), 1);
+        assert!(matches!(
+            parser.errors[0].reason(),
+            Some(ParserErrorReason::DisallowedCfgAttributeContents { .. })
+        ));
+    }
+
+    #[test]
+    fn parsing_cfg_feature_requires_nonempty_string() {
```
asterite, line 710: Should this be "allows" instead of "requires" or is there a missing error here?

compiler/noirc_frontend/src/parser/parser/statement.rs
```diff
@@ -7,26 +7,32 @@ use crate::{
...
+            )
         }
     }
 
     /// Statement = Attributes StatementKind ';'?
-    pub(crate) fn parse_statement(&mut self) -> Option<(Statement, (Option<Token>, Location))> {
+    #[allow(clippy::type_complexity)]
+    pub(crate) fn parse_statement(
+        &mut self,
+    ) -> Option<(Statement, (Option<Token>, Location), Option<CfgAttribute>)> {
```
asterite, line 35: Here too, we could have `Statement` hold a list of `CfgAttribute`. Then we'd also filter those at some point (not sure when, maybe during elaboration).

compiler/noirc_frontend/src/parser/errors.rs
```diff
@@ -78,6 +78,12 @@ pub enum ParserErrorReason {
     TraitImplVisibilityIgnored,
     #[error("comptime keyword is deprecated")]
     ComptimeDeprecated,
+    #[error(
+        "Only one 'cfg' attribute is allowed, but found {cfg_attribute} and {second_cfg_attribute}"
+    )]
+    MultipleCfgAttributesFound { cfg_attribute: CfgAttribute, second_cfg_attribute: CfgAttribute },
```
asterite, line 84: I think Rust allows multiple config attributes

compiler/noirc_frontend/src/parser/parser/statement.rs
```diff
@@ -455,6 +461,32 @@ impl Parser<'_> {
             is_global_let: false,
         })
     }
+
+    /// Return the unique `CfgAttribute` in the given `attributes` list along
+    /// with its `Span`, if it exists.
```
asterite, line 466:
```suggestion
    /// with its `Location`, if it exists.
```

compiler/noirc_frontend/src/tests/cfg.rs
```diff
@@ -0,0 +1,506 @@
+#![cfg(test)]
```
asterite, line 1: The tests here could use the latest way to test errors that use "^^^" and "~~~".

tooling/nargo_fmt/src/formatter/attribute.rs
```diff
@@ -165,6 +172,35 @@ impl Formatter<'_> {
         self.write_right_bracket();
     }
 
+    fn format_cfg_attribute(&mut self, cfg_attribute: CfgAttribute) {
+        self.write_current_token_and_bump(); // #[
+        self.skip_comments_and_whitespace();
+        self.format_path(Path::from_single("cfg".to_string(), cfg_attribute.location()));
+        self.skip_comments_and_whitespace();
+        if self.is_at(Token::LeftParen) {
```
asterite, line 180: Is there a case where a left paren wouldn't be at this position? (I think otherwise the parser will error and with errors the formatter doesn't run)

compiler/noirc_frontend/src/lexer/token.rs
```diff
@@ -998,6 +1011,22 @@ impl SecondaryAttribute {
         matches!(self, SecondaryAttribute::Abi(_))
     }
 
+    pub(crate) fn is_disabled_cfg(&self) -> bool {
+        match self {
```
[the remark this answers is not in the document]
jfecher: Or just a catch-all case in the match.

compiler/noirc_frontend/src/lexer/token.rs
```diff
@@ -1042,6 +1072,49 @@ impl Display for MetaAttribute {
...
+    }
+}
+
+impl CfgAttribute {
+    // TODO(https://github.com/noir-lang/noir/issues/7574): enable more features once working
+    pub(crate) fn is_enabled(&self) -> bool {
+        match self {
+            CfgAttribute::Feature { name, .. } => name == "default",
+        }
+    }
```
jfecher, line 1111: Seems a bit odd to have this in lexer/token.rs

compiler/noirc_frontend/src/parser/parser/expression.rs
```diff
@@ -855,8 +855,11 @@ impl Parser<'_> {
     }
 
     fn parse_statement_in_block(&mut self) -> Option<(Statement, (Option<Token>, Location))> {
-        if let Some(statement) = self.parse_statement() {
-            Some(statement)
+        if let Some((statement, semicolon, opt_cfg_attribute)) = self.parse_statement() {
+            if !self.is_enabled_cfg(opt_cfg_attribute) {
+                return None;
+            }
+            Some((statement, semicolon))
```
jfecher, line 862: I'm curious why the parser has knowledge of cfg attributes - why not handle this logic in the elaborator for example?

compiler/noirc_frontend/src/parser/parser/item.rs
```diff
@@ -134,52 +140,52 @@ impl<'a> Parser<'a> {
             self.comptime_mutable_and_unconstrained_not_applicable(modifiers);
 
             let use_tree = self.parse_use_tree();
-            return vec![ItemKind::Import(use_tree, modifiers.visibility)];
+            return vec![(ItemKind::Import(use_tree, modifiers.visibility), cfg_feature_disabled)];
```
jfecher, line 143: I wonder if it'd be more clear to have these as their own item kinds that wrap existing item kinds? That way we can limit touching external logic. We would likely need a little pass to go through and handle these though before e.g. imports or global items are processed. Perhaps it'd be cleaner to move the filtering of these to a small preprocessing pass.

compiler/noirc_frontend/src/parser/parser/statement.rs
jfecher: I think this is another place having CfgAttribute wrap an existing item would be cleaner instead of making it a separate tuple element. Doesn't touch existing logic but also forces it to be checked since you can't get the data inside otherwise.

compiler/noirc_frontend/src/parser/errors.rs
michaeljklein: Yes, Rust allows e.g. duplicate config attributes:
```rust
#[cfg(feature="default")]
#[cfg(feature="default")]
fn main() {
    println!("Hello, world!");
}
```

but my current plan is to support this using `any`/`all`/etc., so it could be addressed in a follow up.

Do you have a use case for multiple config attributes?

asterite: No use case but I saw it working in Rust, so... For example you could have a `cfg(test)` and `cfg(feature = "foo")` on top of a test, though if there's going to be `all` and `any` then it's not necessary.

compiler/noirc_frontend/src/ast/cfg_attribute.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl<T> From<T> for CfgAttributed<T> {
+    fn from(input: T) -> CfgAttributed<T> {
+        CfgAttributed { cfg_attribute: None, inner: input }
+    }
+}
+
+impl<T> CfgAttributed<T> {
+    pub(crate) fn is_enabled(&self) -> bool {
+        self.cfg_attribute.as_ref().map(|cfg_attribute| cfg_attribute.is_enabled()).unwrap_or(false)
```
jfecher, line 19: Shouldn't this be unwrap_or true so they're enabled if there's no attr?

Issue #7104, opened 2025-01-17
Dead Instruction Elimination doesn't consider `&mut` parameters


aakoshh:
### Aim

Discovered in https://github.com/noir-lang/noir/pull/7072

During preprocessing functions we wanted to call the DIE pass on them to remove leftover instructions after doing mem2reg, but in one example described in the PR description it eliminated the entire function body, causing execution to fail. 

### Expected Behavior

Adding the DIE pass to the preprocessing of functions should not cause integration test failures.

Maxim said:

> Looks like we need to pass some extra context to the can_eliminate_if_unused method on Instruction [noir/compiler/noirc_evaluator/src/ssa/ir/instruction.rs](https://github.com/noir-lang/noir/blob/ad5a9808ab7c2a89beadb6f95c664441ee27a5f8/compiler/noirc_evaluator/src/ssa/ir/instruction.rs#L534)

### Bug

Here's the SSA before and after the preprocessing pass, when the DIE was called on every function:

Before:
```
After Removing Paired rc_inc & rc_decs:
acir(inline) fn main f0 {
  b0(v0: Field, v1: Field):
    v2 = allocate -> &mut Field
    store v0 at v2
    call f1(v2)
    v4 = load v2 -> Field
    v5 = eq v4, v1
    constrain v4 == v1
    return
}
acir(inline) fn Add10 f1 {
  b0(v0: &mut Field):
    v1 = load v0 -> Field
    v2 = load v0 -> Field
    v4 = add v2, Field 10
    store v4 at v0
    return
}
```

After:
```
After Preprocessing Functions:
acir(inline) fn main f0 {
  b0(v0: Field, v1: Field):
    v2 = allocate -> &mut Field
    store v0 at v2
    call f1(v2)
    v4 = load v2 -> Field
    v5 = eq v4, v1
    constrain v4 == v1
    return
}
acir(inline) fn Add10 f1 {
  b0(v0: &mut Field):
    return                              # Whole function body got eliminated!
}
```

It saw `store v4 at v0` as a dead instruction, not realising that `v0` can be used by the caller.

### To Reproduce

```console
$ cde test_programs/execution_success/traits_in_crates_1
$ cargo run -q -p nargo_cli -- --program-dir . execute --force --silence-warnings
error: Failed constraint
  ┌─ /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/traits_in_crates_1/src/main.nr:4:12
  │
4 │     assert(V.Q == y);
  │            --------
[...]

PR #7072, opened 2025-01-15
feat(ssa): Pass to preprocess functions

aakoshh:
# Description

## Problem\*

Followup for https://github.com/noir-lang/noir/issues/7001#issuecomment-2588717106

## Summary\*

Adds a `Ssa::preprocess_fns` pass that does a subset of SSA passes on functions in a bottom-up order, so that by the time they are inlined into other functions, they have a chance to be already simplified. The changes made are cumulative, ie. if A calls B, then B gets optimised first, and then A inlines the optimised B before it gets optimised itself.

To avoid going all the way to the top and optimising the functions near the entry point, the weight of the functions (derived from their instruction count) is accumulated to the callers, and the average of the accumulated weights serves as a cutoff for running the preprocessing, unless most of the weight comes from the functions own body.

TODO:
- [x] Fix the inlining logic by separating expensive-or-recursive vs ACIR-to-Brillig related info (allows Brillig-to-Brillig calls to be inlined even if ACIR-to-Brillig marks them as Brillig entry points)
- [x] Fix defunctionalization to inherit the runtime onto the callee
- [x] Fix the removal of unreachable functions to remove never-called Brillig functions 
- [x] Create issue for fixing the DIE for mutable references https://github.com/noir-lang/noir/issues/7104
- [ ] Look at the SSA in Tom's comments and create issues as required

### Idea 
We noticed in https://github.com/noir-lang/noir/issues/7001#issuecomment-2588414651 that `main` gets bloated to 65K blocks after the _unrolling_ pass, because everything it calls has already been inlined into it, and then we have to simplify down this large function, which came with a big memory footprint. 

@TomAFrench suggested switching to a bottom-up scenario, where we optimise functions that get called by others before inlining them, and to start as follows:

> We can validate this by hacking the inlining pass so that we fully inline into, unroll and mem2reg  functions called `serialize`. Note I'm thinking having this as a separate pass and we'd run the full inlining pass afterwards rather than trying to fit everything inside of the inlining pass.

### Testing

```shell
cargo run -p nargo_cli -- --program-dir ../aztec-packages/noir-projects/noir-protocol-circuits/crates/rollup-base-public compile  --silence-warnings --skip-underconstrained-check --skip-brillig-constraints-check --force --show-ssa-pass "Preprocessing"
```

### Result

Unfortunately this hasn't significantly changed the number of blocks after unrolling in the `main`, at least not yet:

```
AFTER Unrolling: functions=33
    FUNCTION main: blocks=62495
```

[...]

PR #7106, opened 2025-01-17
feat(ssa): Add flag to DIE pass to be able to keep `store` instructions

aakoshh:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/7104

## Summary\*

Adds a `flattened` flag to `Function::dead_instruction_elimination` which is passed to `Instruction::can_eliminate_if_unused` to prevent [this assumption](https://github.com/noir-lang/noir/blob/a6685befe3b103b236997843f69afc07b6ea0a8c/compiler/noirc_evaluator/src/ssa/ir/instruction.rs#L534-L544) from removing the `store` instruction before the inlining, the flattening of the CFG, and mem2reg passes have happened. Thanks @vezenovm for the pointer!

## Additional Context

Originally I thought I would have to track `&mut` block parameters as "use", but the assumptions listed in `can_eliminate_in_unused` suggested that `store` _must_ be removed from ACIR, and to stop that from happening it's enough to give it more information, so it doesn't infer whether other passes happened already based on the number of blocks in the function.

It looked like tracking `&mut` parameters as "use" would be a bigger change, since only [instruction results](https://github.com/noir-lang/noir/blob/a6685befe3b103b236997843f69afc07b6ea0a8c/compiler/noirc_evaluator/src/ssa/opt/die.rs#L177) are checked for use at the moment. Maybe there are other cases which would better be handled that way?

For the use case of https://github.com/noir-lang/noir/pull/7072 I thought we'd just call the function with `false` and be done with it.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/instruction.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/instruction.rs
@@ -10,7 +10,7 @@ use fxhash::FxHasher64;
 use iter_extended::vecmap;
 use noirc_frontend::hir_def::types::Type as HirType;
 
-use crate::ssa::{ir::function::RuntimeType, opt::flatten_cfg::value_merger::ValueMerger};
+use crate::ssa::opt::flatten_cfg::value_merger::ValueMerger;
 
 use super::{
     basic_block::BasicBlockId,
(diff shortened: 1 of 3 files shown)


TomAFrench (approved):
LGTM

vezenovm (approved):

Issue #988, opened 2023-03-15
Add enumerations (tagged unions)


jfecher:
### Problem

It would be useful in Noir to support types like `Option` and `Result` which hold different values based upon a tag value. Optional values in particular would be useful for future Noir contracts.

### Proposed solution

Add rust-style enumerations into noir, pattern matching and all.

The main concern is how to represent enumerations given we cannot reinterpret-cast data into another type. I propose representing an arbitrary enumeration:
```rs
enum Foo {
    A,
    B(BData),
    C(CData, Other),
    ...,
}
```
As a struct:
```rs
struct Foo {
    tag: Field, // Set internally, 0=A, 1=B, 2=C, etc.
    a_data: (),
    b_data: BData,
    c_data: (CData, Other),
    ...,
}
```
Now, given a constructor call such as `Foo::B(my_b_data)`, the compiler can translate this into a struct initialization. There is the question of what to do for the fields of other variants like (), CData, and Other. Since these values would be guaranteed to be inaccessible by the compiler when the tag value is B, and since all values in Noir are comprised of Fields, we can initialize all these values to 0 without changing behavior:
```rs
let b = Foo::B(my_b_data);

// translated to:
let b = Foo {
    tag: 1, // B is the second variant
    a_data: dep::std::unsafe::zeroed(),
    b_data: my_b_data,
    c_data: dep::std::unsafe::zeroed(),
};
```
---
A concrete example

`Option` in this scheme could be typed as:
```rs
enum Option<T> {
    None,
    Some(T),
}

fn main(x: Option<Field>) {
    let y = Some(2);

    let x_value = match x {
        None => 0,
        Some(value) => value,
    };

    constrain y == Some(x_value);
}
```
Which would be translated internally as:
```rs
struct Option<T> {
    tag: Field,
    some_value: T,
}

fn main(x: Option<Field>) {
    let y = Option { tag: 1, some_value: 2 };

    let x_value = 
        if x.tag == 0 { 0 }
        else { x.some_value }; // compiler guarantees that x != 0 implies x == 1

    constrain y == Option { tag: 1, some_value: x_value };
}
```

### Alternatives considered

As a slight variant of the above translation of `match`, we could consider an alternate translation that explicitly checks each tag and fails if an impossible case is somehow reached:
```rs
let x_value = 
    if x.tag == 0 { 0 }
    else if x.tag == 1 { x.some_value }
    else { panic("Invalid tag for Option enum") }
```
Where `panic` would be defined as:
[...]

PR #7110, opened 2025-01-17
feat: Parser and formatter support for `enum`s

jfecher:
# Description

## Problem\*

Step towards #988 

## Summary\*

Adds parser & formatter support for enumerations. Elaboration & comptime support is left as a `todo!()`.

## Additional Context

Enums are currently gated behind an experimental feature:

```
warning: Enums are experimental and aren't fully supported yet
  ┌─ src/main.nr:3:1
  │
3 │ enum Foo {
  │ ----
  │
```

Although they're behind a `todo!()`, it won't panic the compiler since they're elaborated to begin with.

## Documentation\*

Check one:
  - Waiting until the feature is finished

# PR Checklist\*


compiler/noirc_frontend/src/parser/parser/enums.rs
```diff
@@ -0,0 +1,281 @@
...
+                Vec::new(),
+                start_span,
+            );
+        };
+
+        let generics = self.parse_generics();
+
+        if self.eat_semicolons() {
+            return self.empty_enum(name, attributes, visibility, generics, start_span);
+        }
```
asterite, line 43: Should this be removed? This would allow `enum Foo;` which I thought was maybe valid in Rust because of this code but apparently it's not.

jfecher: We could allow it and treat it as the same as `enum Foo {}` but I'll just remove it for consistency.

We could leave it as a separate question whether to support `enum Void {}` for the Void type

jfecher:
> Performance Alert
> rollup-block-root-single-tx 	152 s 	138 s

Does seem like a large difference for a PR which shouldn't impact anything

compiler/noirc_frontend/src/parser/parser/enums.rs
asterite: Sounds good. I then tried `enum Foo {}` in Rust and it worked so I was wondering why Rust then doesn't allow `enum Foo;`. It would be consistent with structs so maybe us supporting that would be good. So up to you to leave it or remove it.

asterite (approved):
It's happening! 🤩

asterite:
Hm, seeing that huge difference, maybe allowing `enum Foo;` is good then.

Just kidding, there's probably something wrong in CI? The compilation time change would also be explained by that.

vezenovm:
> Hm, seeing that huge difference, maybe allowing `enum Foo;` is good then.
> 
> Just kidding, there's probably something wrong in CI? The compilation time change would also be explained by that.

Globals brought this +2 overhead to all programs, probably just need to sync with master.

PR #7115, opened 2025-01-17
feat: Resolve enums & prepare type system

jfecher:
# Description

## Problem\*

Working towards #988 

## Summary\*

- Implements `def_collector::collect_enums`
- Lays down some prep work in the type system for enums. The most significant of which (and the majority of the diff) is changing `StructType` to `DataType` to handle structs and enums. Similarly `StructId` is now `TypeId`, although I'm leaving metaprogramming changes for a future date (we can abstract these changes there or impose the same merger).
- Implements some missing language server integration

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
Currently failing due to the `NestedSlice` error now failing because it's now catching that we're trying to get the fields of a type which still hasn't had it's fields set yet. This was the case before too but we previously just returned an empty vec of fields.

compiler/noirc_frontend/src/hir/def_collector/dc_mod.rs
```diff
@@ -1081,16 +1085,97 @@ pub fn collect_struct(
...
+    let unresolved = UnresolvedEnum { file_id, module_id, enum_def };
+
+    let resolved_generics = Context::resolve_generics(
+        interner,
+        &unresolved.enum_def.generics,
+        definition_errors,
+        file_id,
+    );
+
+    // Create the corresponding module for the struct namespace
```
asterite, line 1113: Really minor, but...

```suggestion
    // Create the corresponding module for the enum namespace
```

compiler/noirc_frontend/src/hir/def_collector/dc_mod.rs
```diff
@@ -1081,16 +1085,97 @@ pub fn collect_struct(
...
+            let local_id = module_id.local_id;
+            interner.new_type(name, span, attributes, resolved_generics, krate, local_id, file_id)
+        }
+        Err(error) => {
+            definition_errors.push((error.into(), file_id));
+            return None;
+        }
+    };
+
+    interner.set_doc_comments(ReferenceId::Struct(id), doc_comments);
```
asterite, line 1141:
```suggestion
    interner.set_doc_comments(ReferenceId::Enum(id), doc_comments);
```

compiler/noirc_frontend/src/hir/def_collector/dc_mod.rs
```diff
@@ -1081,16 +1085,97 @@ pub fn collect_struct(
...
+    for (index, variant) in unresolved.enum_def.variants.iter().enumerate() {
+        if !variant.doc_comments.is_empty() {
+            let id = ReferenceId::EnumVariant(id, index);
+            interner.set_doc_comments(id, variant.doc_comments.clone());
+        }
+    }
+
+    // Add the enum to scope so its path can be looked up later
+    let visibility = unresolved.enum_def.visibility;
+    let result = def_map.modules[module_id.0].declare_struct(name.clone(), visibility, id);
```
asterite, line 1152: Maybe this should be `declare_type`?

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -388,100 +403,130 @@ enum FunctionCoercionResult {
...
     /// created. Therefore, this method is used to set the fields once they
     /// become known.
     pub fn set_fields(&mut self, fields: Vec<StructField>) {
-        self.fields = fields;
+        self.body = TypeBody::Struct(fields);
     }
 
+    /// Returns the number of fields this struct has. Returns 0 if this type
+    /// is not a struct.
     pub fn num_fields(&self) -> usize {
```
asterite, line 447: It seems this method is unused.

tooling/lsp/src/requests/inlay_hint.rs
```diff
@@ -95,13 +95,21 @@ impl<'a> InlayHintCollector<'a> {
...
                         let field = struct_type.field_at(field_index);
                         self.push_type_hint(lsp_location, &field.typ, false);
                     }
+                    ReferenceId::EnumVariant(type_id, variant_index) => {
+                        let typ = self.interner.get_type(type_id);
+                        let shared_type = typ.clone();
+                        let typ = typ.borrow();
+                        let variant_type = typ.variant_function_type(variant_index, shared_type);
+                        self.push_type_hint(lsp_location, &variant_type, false);
+                    }
```
asterite, line 109: I think the `StructMember` case is when an identifier refers to a struct field, like here:

image

I don't know if an identifier would point to an enum variant where that identifier would have an inlay hint. But I also think enums aren't type-checked yet so I couldn't try code with enums.

asterite (approved):
Looks good! I left a few minor comments.

compiler/noirc_frontend/src/hir/def_collector/dc_mod.rs
jfecher: Really minor, but objectively correct!

compiler/noirc_frontend/src/hir/def_collector/dc_mod.rs
jfecher: Good catch. I renamed each struct method I found but it's very likely I've missed quite a few since I've just been renaming them as I go. I'll rename this one too.

tooling/lsp/src/requests/inlay_hint.rs
jfecher: I wasn't sure if we'd want to give an inlay hint for them either but I thought it was a possibility so I figured I might as well support it. I'd be open to removing it in the future if we don't want it.

Issue #7111, opened 2025-01-17
Remove comment system for time/memory benchmarks


TomAFrench:
After #7068, it's not necessary to print these comments on every PR as well be alerted to any regressions. We can then remove any usage of noir-bench-diff from this repo

PR #7119, opened 2025-01-20
feat: unconstrained optimizations for BoundedVec

### LINKED ISSUE #7094: `BoundedVec.any` (and similar) should stop iterating at `self.len()` in brillig runtimes
@nventuro is currently doing work which requires iterating over long `BoundedVec`s in brillig code and so is paying a lot of costs associated with unnecessary looping.

https://github.com/noir-lang/noir/blob/c172880ae47ec4906cda662801bd4b7866c9586b/noir_stdlib/src/collections/bounded_vec.nr#L390-L400

This can be solved by us conditionally dispatching using `std::runtime::is_unconstrained()` to runtime specific versions such as the below:

```nr
 pub fn any<Env>(self, predicate: fn[Env](T) -> bool) -> bool { 
     for i in 0..self.len { 
         if predicate(self.storage[i]) { 
             return true
         } 
     } 
     false 
 } 
```

This change should not change the interface of the stdlib.

asterite:
# Description

## Problem

Resolves #7094

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/noir_stdlib/src/collections/bounded_vec.nr
+++ b/noir_stdlib/src/collections/bounded_vec.nr
@@ -1,4 +1,4 @@
-use crate::{cmp::Eq, convert::From};
+use crate::{cmp::Eq, convert::From, runtime::is_unconstrained};
 
 /// A `BoundedVec<T, MaxLen>` is a growable storage similar to a `Vec<T>` except that it
 /// is bounded with a maximum possible length. Unlike `Vec`, `BoundedVec` is not implemented
@@ -320,11 +320,17 @@ impl<T, let MaxLen: u32> BoundedVec<T, MaxLen> {
         let new_len = self.len + append_len;
         assert(new_len <= MaxLen, "extend_from_bounded_vec out of bounds");
 
-        let mut exceeded_len = false;
-        for i in 0..Len {
-            exceeded_len |= i == append_len;
-            if !exceeded_len {
+        if is_unconstrained() {
+            for i in 0..append_len {
                 self.storage[self.len + i] = vec.get_unchecked(i);
+            }
+        } else {
+            let mut exceeded_len = false;
+            for i in 0..Len {
+                exceeded_len |= i == append_len;
+                if !exceeded_len {
+                    self.storage[self.len + i] = vec.get_unchecked(i);
+                }
             }
         }
         self.len = new_len;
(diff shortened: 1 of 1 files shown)


asterite:
I was wondering if we could have something like this in the standard library:

```noir
use std::runtime::is_unconstrained;

fn main() {
    for_loop::<5>(0, 3, |i| { println(f"{i}"); })
}

fn for_loop<let Max: u32>(start: u32, end: u32, f: fn(u32) -> ()) {
    if is_unconstrained() {
        if end < Max {
            for i in start..end {
                f(i);
            }
        }
    } else {
        for i in start..Max {
            if i < end {
                f(i);
            }
        }
    }
}
```

Using `--show-ssa` it seems the compiler optimizes this well, but I don't know if it would make compilation slower, or if it would make the code less clear (it could at least be a `pub(crate)` function)

noir_stdlib/src/collections/bounded_vec.nr
```diff
@@ -413,11 +426,19 @@ impl<T, let MaxLen: u32> BoundedVec<T, MaxLen> {
...
-        for i in 0..MaxLen {
-            if i < self.len() {
+
+        if is_unconstrained() {
+            for i in 0..self.len() {
                 ret.storage[i] = f(self.get_unchecked(i));
             }
+        } else {
+            for i in 0..MaxLen {
+                if i < self.len() {
```
aakoshh, line 436: I wonder why it's okay to use `if i < self.len()` here, but not in `any`.

asterite: Maybe both are fine and `exceeded_len` is an optimization? I didn't check if changing this makes a difference.

aakoshh:
> I was wondering if we could have something like this in the standard library:

Wouldn't it make sense to have an `assert(end <= Max)` instead of `if end < Max {`?

noir_stdlib/src/collections/bounded_vec.nr
aakoshh: Yes, I have no doubt both are equivalent in terms of what they do, just didn't understand why the same optimisation isn't applied here, whether it should 🤷

aakoshh:
@TomAFrench so is there a way now to see the output of https://github.com/noir-lang/noir/actions/runs/12869320334/job/35878119440?pr=7119 ?

TomAFrench:
Circuit sizes are still posted as comments as they're not prone to fluctuations and so are only posted when there's a meaningful change.

aakoshh (approved):
Looks correct. Would be nice to see the impact.

aakoshh:
Ah ok, because I can't see them at https://noir-lang.github.io/noir/dev/bench/
So basically we don't see anything because it didn't affect the circuits we have in the integration tests.

TomAFrench:
Yes that's correct. If these metrics are unchanged then no comment is posted, this is completely unchanged by #7111 as that only affected time and memory benchmarks.

noir_stdlib/src/collections/bounded_vec.nr
TomAFrench: Yeah, `exceeded_len` is an optimization as it's easier to check for equality than do a comparison.

TomAFrench:
> I was wondering if we could have something like this in the standard library

Yeah that looks sensible and would be efficient to compile (is_unconstrained is resolved at ssa_gen time so the unused branch is never even generated). I don't think it's something that we want to expose from the stdlib but if it's something we'd find ourselves doing over and over then let's use it internally.

PR #7124, opened 2025-01-20
fix: allow calling trait impl method from struct if multiple impls exist

### LINKED ISSUE #7090: Cannot invoke trait function on struct if multiple impls exist
### Aim

This code:

```noir
trait From2<T> {
    fn from2(input: T) -> Self;
}
struct U60Repr {}
impl From2<[Field; 3]> for U60Repr {
    fn from2(_: [Field; 3]) -> Self {
        U60Repr {}
    }
}
impl From2<Field> for U60Repr {
    fn from2(_: Field) -> Self {
        U60Repr {}
    }
}
fn main() {
    let _ = U60Repr::from2([1, 2, 3]);
    let _ = U60Repr::from(1);
}
```

### Expected Behavior

Expected the code above to compile.

### Bug

It doesn't compile:

```
error: Could not resolve 'from2' in path
   ┌─ src/main.nr:16:22
   │
16 │     let _ = U60Repr::from2([1, 2, 3]);
   │                      -----
   │

error: Could not resolve 'from' in path
   ┌─ src/main.nr:17:22
   │
17 │     let _ = U60Repr::from(1);
   │                      ----
   │

Aborting due to 2 previous errors
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

See also https://github.com/noir-lang/noir/pull/7074 as an attempt to fix this

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #7090

## Summary

Try to solve `Struct::method` in a special way like we solve `Trait::method` in a special way.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/types.rs
+++ b/compiler/noirc_frontend/src/elaborator/types.rs
@@ -37,7 +37,11 @@ use crate::{
     Generics, Kind, ResolvedGeneric, Type, TypeBinding, TypeBindings, UnificationError,
 };
 
-use super::{lints, path_resolution::PathResolutionItem, Elaborator, UnsafeBlockStatus};
+use super::{
+    lints,
+    path_resolution::{PathResolutionItem, Turbofish},
+    Elaborator, UnsafeBlockStatus,
+};
 
 pub const SELF_TYPE_NAME: &str = "Self";
 
@@ -692,7 +696,8 @@ impl<'context> Elaborator<'context> {
 
         let mut path = path.clone();
         let span = path.span();
-        let last_path_segment = path.pop();
+        let last_segment = path.pop();
+        let before_last_segment = path.last_segment();
 
         let path_resolution = self.resolve_path(path).ok()?;
         let PathResolutionItem::Struct(struct_id) = path_resolution.item else {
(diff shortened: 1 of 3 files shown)


asterite:
This makes cases that previously didn't compile to now compile, but there are still some cases that don't compile. For example:

```noir
trait From2<T> {
    fn from2(input: T) -> Self;
}
struct U60Repr<let N: u32> {}
impl<let N: u32> From2<[Field; N]> for U60Repr<N> {
    fn from2(_: [Field; N]) -> Self {
        U60Repr {}
    }
}
impl<let N: u32> From2<Field> for U60Repr<1> {
    fn from2(_: Field) -> Self {
        U60Repr {}
    }
}
fn main() {
    test_eq::<3>()
}

fn test_eq<let N: u32>() {
    let w = array();
    let _: U60Repr<N> = U60Repr::from2(w);
}

fn array<let N: u32>() -> [Field; N] {
    [0; N]
}
```

This one fails with this:

```
error: Could not determine array length `_`, encountered error: `Expected a constant, but found `_``
   ┌─ src/main.nr:20:13
   │
20 │     let w = array();
   │             -----
   │
   = Call stack:
     1. src/main.nr:20:13
```

And this works fine if you use `From2::from2(w)`.

The difference seems to be that when we solve `From2::from2(w)`, the trait constraint object type we use is a type variable. In the case of `U60Repr::from2(w)` I made the trait constraint object to be the struct type `U60Repr` with all unbound type variables. But for some reason that fails in this case. If I make the trait constraint object to be a type variable (not the struct type) then it works, but the other tests in this PR fail because it's as if we wrote `From2::from2(...)` so the compiler can't know it's actually for `U60Repr`.

I don't know how to solve this, but maybe someone else realizes what's missing in this PR.

TomAFrench (approved):

PR #7125, opened 2025-01-20
fix: proper cleanup when breaking from comptime loop on error

asterite:
# Description

## Problem

Resolves https://github.com/noir-lang/noir/pull/7096#discussion_r1922597178

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
@@ -1730,17 +1730,21 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
             self.push_scope();
             self.current_scope_mut().insert(for_.identifier.id, make_value(i));
 
-            match self.evaluate(for_.block) {
-                Ok(_) => (),
-                Err(InterpreterError::Break) => break,
-                Err(InterpreterError::Continue) => continue,
+            let must_break = match self.evaluate(for_.block) {
+                Ok(_) => false,
+                Err(InterpreterError::Break) => true,
+                Err(InterpreterError::Continue) => false,
                 Err(error) => {
                     result = Err(error);
-                    break;
+                    true
                 }
-            }
+            };
 
             self.pop_scope();
+
+            if must_break {
+                break;
+            }
         }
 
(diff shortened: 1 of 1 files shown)


compiler/noirc_frontend/src/hir/comptime/interpreter.rs
```diff
@@ -1732,20 +1734,24 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
...
                 Err(InterpreterError::Break) => break,
                 Err(InterpreterError::Continue) => continue,
-                Err(other) => return Err(other),
+                Err(error) => {
+                    result = Err(error);
+                    break;
+                }
             }
 
             self.pop_scope();
```
aakoshh, line 1743: These still don't look good to me: 
* if we `break` out of the `for` then `self.pop_scope()` will not be called for the last loop, leaving an extra entry that `self.push_scope()` created.
* if we `continue` the with the next iteration, then `self.pop_scope()` will not be called, and another `self.push_scope()` will be created on top of it

aakoshh: Maybe a unit test could be added after [these](https://github.com/noir-lang/noir/blob/3153058a780444622c5628caf4c18f50999c0e7b/compiler/noirc_frontend/src/hir/comptime/tests.rs#L193), something like:
```
let x = 1;
for i in 0 .. 10 {
  let x = 2;
  break; 
}
println("x = {x}"); // Is it going to find it in `current_context`, which is a leftover from the loop?
```

asterite: Oops, you are right! I think I got confused with `break` and breaking out of `match`, or I don't know what I was thinking.

I tried a different approach.

aakoshh (approved):
Thanks, that does what I thought it should be!

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
asterite: I'll try to add a unit test before merging this.

Issue #7103, opened 2025-01-17
Infinite `loop`s crash the compiler


asterite:
### Aim

This code:

```noir
fn main() {
    /// Safety:
    unsafe {
        some_loop();
    }
}

unconstrained fn some_loop() {
    let mut i = 0;
    loop {
        println(i);
        i += 1;
    }
}
```

### Expected Behavior

Expected the code to compile and execute well.

### Bug

It leads to a crash:

```
The application panicked (crashed).
Message:  Expected a return instruction, as function construction is finished
Location: compiler/noirc_evaluator/src/ssa/ir/function.rs:180
```

If we change the failing function to handle empty return values, then we must do more things:
- The SSA inliner code will crash now in a place it expects at least one return. This can be changed to return `Vec::new()` in that case.
- There's an SSA optimization that tries to evaluate brillig calls with all constants. It must be changed so that if that function has no returns, don't try to evaluate it (it hangs the compiler)
- There's a similar optimization when generating brillig from SSA when we attempt to evaluate brillig, so a similar change must be done there

But even then the compiler crashes with this:

```
The application panicked (crashed).
Message:  ICE: Value not found in cache v13
Location: compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block_variables.rs:46
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

aakoshh:
Perhaps it could also be handled similarly to https://github.com/noir-lang/noir/pull/6292

jfecher:
SSA in general isn't really equipped to deal with alternate looping constructs. Not only do we assume every function has exactly one return, there are also assumptions on a loop's block structure in the loop unrolling pass (pre-header, header, body blocks). This block structure has been relaxed over time to return optionals but I wouldn't be surprised if that pass finds a block before the infinite loop which happens to start with a `<` instruction and identifies it as the header of the loop.

This is fixable though. I'd definitely be fine with erroring if the loop doesn't contain at least 1 break as well. Identifying whether the break is reachable or not correctly in all cases is impossible so we don't need to try especially for a first pass at this. We could add a warning/error later if the break is known unreachable though. We'll probably know this somewhere in SSA.

PR #7126, opened 2025-01-20
feat: `loop` must have at least one `break`

asterite:
# Description

## Problem

Related to #7103

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/statements.rs
+++ b/compiler/noirc_frontend/src/elaborator/statements.rs
@@ -289,12 +289,10 @@ impl<'context> Elaborator<'context> {
 
         self.pop_scope();
 
-        if !in_constrained_function {
-            let last_loop =
-                std::mem::replace(&mut self.current_loop, old_loop).expect("Expected a loop");
-            if !last_loop.has_break {
-                self.push_err(ResolverError::LoopWithoutBreak { span });
-            }
+        let last_loop =
+            std::mem::replace(&mut self.current_loop, old_loop).expect("Expected a loop");
+        if !last_loop.has_break {
+            self.push_err(ResolverError::LoopWithoutBreak { span });
         }
 
         let statement = HirStatement::Loop(block);
--- a/compiler/noirc_frontend/src/elaborator/statements.rs
+++ b/compiler/noirc_frontend/src/elaborator/statements.rs
@@ -33,7 +33,7 @@ impl<'context> Elaborator<'context> {
             StatementKind::Constrain(constrain) => self.elaborate_constrain(constrain),
             StatementKind::Assign(assign) => self.elaborate_assign(assign),
             StatementKind::For(for_stmt) => self.elaborate_for(for_stmt),
-            StatementKind::Loop(block) => self.elaborate_loop(block, statement.span),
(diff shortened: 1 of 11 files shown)


compiler/noirc_frontend/src/tests.rs
```diff
@@ -4073,3 +4073,25 @@ fn regression_7088() {
...
+    let src = r#"
+    fn main() {
+        /// Safety: test
+        unsafe {
+            foo()
+        }
+    }
+
+    unconstrained fn foo() {
+        loop {}
```
michaeljklein, line 4088: Please add 1-2 non-trivial cases, e.g.
```noir
loop {
  x += 1;
  bar(x);
}
```
or a case where there are 2+ nested loops where only only nesting level is missing a `break`, to test the `self.nested_loops`/`self.current_loop` logic.

michaeljklein (changes requested):

compiler/noirc_frontend/src/tests.rs
asterite: Done!

compiler/noirc_frontend/src/elaborator/statements.rs
```diff
@@ -279,13 +281,21 @@ impl<'context> Elaborator<'context> {
...
         self.pop_scope();
-        self.nested_loops -= 1;
+
+        if !in_constrained_function {
+            let last_loop =
+                std::mem::replace(&mut self.current_loop, old_loop).expect("Expected a loop");
+            if !last_loop.has_break {
+                self.push_err(ResolverError::LoopWithoutBreak { span });
+            }
+        }
```
jfecher, line 298: Why guard this with a `if !in_constrained_function`?
Even if the user writes a `loop {}` in a function they accidentally let be constrained I don't see the harm in issuing a loop without break error. I also think because the `std::mem::replace` is still here it opens up more chances for it to get out of sync if we're in a constrained function.

michaeljklein (approved):

jfecher (approved):
LGTM.

I imagine we still need to fix SSA so it doesn't crash if the break is optimized out entirely. E.g.
```noir
loop {
    if false { break; }
}
```

compiler/noirc_frontend/src/elaborator/statements.rs
asterite: My thinking was that the user already got an error for the `loop`, so giving another error might be odd (if you can't have a loop here, does it matter whether it has a break or not?). But I'm fine either way, so let me know.

jfecher: I'd prefer the second error because there are still two things wrong with the code. I tend to disagree with omitting any errors in general since it makes fixing code more difficult (need to fix the first issue, then save and run the compiler again to see the second). In this case I think it could be the case the user wanted a `loop` but forgot to make their function unconstrained first so they'll need to fix both errors anyway.

asterite: Done! Also the thing you said about `std::mem::replace` is true, I meant for `last_loop` to be out of that `if` but I ended up putting it inside it.

PR #7130, opened 2025-01-21
chore: relax threshold for reporting regressions

TomAFrench:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):

asterite:
I do wonder if it should be a percentage or maybe absolute values... though it's tricky. Because if something takes 1 second and it changed to 1.2, maybe it's just a fluke, but going from 130 seconds to 150 seconds might mean something really changed.

TomAFrench:
If there's pain in future then someone can make this more restrictive but personally right now.. :shrug:

PR #7136, opened 2025-01-21
feat: add `noir-inspector`

### LINKED ISSUE #7131: Write tool to inspect build artifacts
Currently we have no way to inspect a build artifact after it's built. This would be useful for:
- being able to print the ACIR and inspect it to compare against different build artifacts
- being able to count the number of brillig opcodes in the program
- etc.

asterite:
# Description

## Problem

Resolves #7131

## Summary

noir-inspector

## Documentation

Check one:

# PR Checklist


TomAFrench (changes requested):
I'd like to separate this tool from nargo ideally

asterite:
Sounds good! Any name for the tool?

tooling/inspector/src/main.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// Specify the backend expression width that should be assumed
+    #[arg(long, value_parser = parse_expression_width)]
+    expression_width: Option<ExpressionWidth>,
+
+    /// Display the ACIR for compiled circuit
+    #[arg(long)]
+    print_acir: bool,
+}
+
+pub const DEFAULT_EXPRESSION_WIDTH: ExpressionWidth = ExpressionWidth::Bounded { width: 4 };
```
asterite, line 26: `DEFAULT_EXPRESSION_WIDTH` is in `noirc_driver` but I didn't want to depend on that, and I wasn't sure if it was okay to move it to the same crate `ExpressionWidth` is.

TomAFrench: I'm inclined to omit this information if we're relying on the user to provide it for us.

tooling/inspector/src/main.rs
```diff
@@ -0,0 +1,77 @@
+use std::{path::PathBuf, process::exit};
+
+use acir::circuit::ExpressionWidth;
+use clap::Parser;
+use noirc_artifacts::program::ProgramArtifact;
+use noirc_artifacts_info::{count_opcodes_and_gates_in_program, show_info_report, InfoReport};
+
+#[derive(Debug, Clone, Parser)]
+pub(crate) struct Args {
```
TomAFrench, line 9: Can you make this have be a `noir-inspector info ./target/program.json` command rather than just `noir-inspector ./target/program.json` so we can more easily add extra commands in future?

TomAFrench (commented):
I'm good with the `noir-inspector` name for now. We can always rename it in future.

Something which we should do as a followup to this PR is have `noirup` install this tool in parallel. We can also put some thought on how we can add this binary into releases.

tooling/inspector/src/main.rs
```diff
@@ -0,0 +1,77 @@
...
+        Err(error) => {
+            println!("Cannot deserialize artifact: {}", error);
+            exit(1);
+        }
+    };
+
+    if args.print_acir {
+        println!("Compiled ACIR for main:");
+        println!("{}", artifact.bytecode);
+    }
```
TomAFrench, line 49: printing the ACIR could come under a different command than the info cmd.

tooling/inspector/src/main.rs
asterite: Done!

tooling/inspector/src/main.rs
asterite: In the end I found that the expression width was only used to print it in the final report... so I made it optional (in the code) and now there's no command line argument for it.

TomAFrench (approved):

PR #7138, opened 2025-01-21
feat(stdlib): Make Hasher trait generic over input type

crStiv:
~~Makes the Hasher trait generic with type parameter T to support different input types beyond Field.~~ This change:
- ~~Adds type parameter T to Hasher trait~~
- ~~Updates BuildHasher and BuildHasherDefault accordingly~~
- ~~Maintains backwards compatibility with existing Field implementations~~
- Removes TODO comment about generic types


TomAFrench (changes requested):
This is missing necessary trait bounds. Will review later.

jfecher (commented):
What is the use case for this? Generally, even if we need to cast to a field that should be a 0-cost operation

PR #7140, opened 2025-01-21
fix(ssa): Use post order when mapping instructions in loop invariant pass

### LINKED ISSUE #7128: Program exposes unsound and incomplete behavior
### Aim

Hello!

The following program seems to reveal a bug that can manifest in different bug-forms:

**main.nr**
~~~rust
fn main(in0: Field) -> pub Field {
    let mut out0 : Field = 0;
    let mut tmp1 : Field = 0;

    if ((out0 == out0) & true) // <== changing out0 to in0 or removing
    {                          //     the comparison changes the result
        let in0_as_bytes : [u8; 32] = in0.to_be_bytes();
        let mut result : [u8; 32] = 0.to_be_bytes();
        for i in 0..32 {
            result[i] = (in0_as_bytes[i] ^ result[i]);
        }
        tmp1 = std::field::bytes32_to_field(result);
    }
    else { } // <== if this is removed the compiler crashes

    let mut tmp2 : Field = 0; // <== moving this to the top of main,
    if (0.lt(in0))            //     changes the result
    {
        tmp2 = 1;
    }
    else { } // <== if this is removed the compiler crashes

    out0 = (tmp2 - tmp1);

    // assert(out0 == 0, "completeness violation"); <== uncomment to see effect
    // assert(out0 != 0, "soundness violation");    <== uncomment to see effect

    out0
}
~~~

**Prover.toml**
~~~toml
in0="1"
~~~

See the comments inside of the code or the bug section for more details.

I tried to minimize it but got stuck at the current form. I would be very interested in the cause and fix of this bug.

### Expected Behavior

 1. For the presented program the return value for `out0` should be `0` in any case.
 2. Removing the `else` blocks should not trigger a compiler panic

### Bug

The first bug, looks like a compiler error. By removing any of the **empty** `else` blocks, the command `nargo execute` stops the execution with following error:

~~~
Message:  internal error: entered unreachable code: All Value::Instructions should already be known during inlining after creating the original inlined instruction. Unknown value v293 = Instruction { instruction: Id(240), position: 0, typ: Reference(Numeric(NativeField)) }
Location: compiler/noirc_evaluator/src/ssa/opt/inlining.rs:504
~~~

If the program is executed as-is the returned value for `out0` is `Field(-1)`. This value is unexpected and wrong as the correct value should be `0`. However, this can be changing by applying the following three semantic equivalent changes to the code:

 1. replacing the **valid** condition part `(out0 == out0)` by `(in0 == in0)`, or
    `(tmp1 == tmp1)`
 2. removing the condition part `(out0 == out0) &&`
 3. moving the declaration `let mut tmp2 : Field = 0;` to the top of the program.

By applying *ANY* of these changes the returned value for `out0` changes to the expected `Field(0)`.

While the compiler crash can be seen as simple crash or completeness issue, the second part of the bug is much more dangerous. It exposes not only a completeness  but also a soundness problem. This can be seen by uncommenting the assertions `assert(out0 == 0, "completeness violation");` or `assert(out0 != 0, "soundness violation");`. Their description state the problem in the current program. The assertion `assert(out0 == 0, "completeness violation");` should pass, but due to the evaluation error it fails. Even worse, the expression `assert(out0 != 0, "soundness violation");` should fail, however, for the current structure it passes for `in0="1"`, allowing for unexpected input.

### To Reproduce

 1. create a new project with `nargo init`
 2. change the content of `src/main.nr` with the provided code
 3. create a `Prover.toml`file with the content `in0 = "1"`
 4. execute `nargo execute`

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.1 noirc version = 1.0.0-beta.1+dd7084545dfd93a07599fc10676b6c8ec1e3d458 (git version hash: dd7084545dfd93a07599fc10676b6c8ec1e3d458, is dirty: false)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #7128 

## Summary\*

Ended up being a small fix as it was a silly oversight for how we should be mapping instructions. I switched away from using a BTreeSet of blocks (the return value from `reachable_blocks()`) to using the reverse post order when mapping functions.

Mapping instructions by the blocks consecutive ordering is incorrect as we may have a situation like such:
```text
acir(inline) fn main f0 {
  b0(v0: Field):
    jmp b3(v0)
  b3(v1: Field):
    v2 = allocate -> &mut Field
    store Field 0 at v2
    v4 = eq v1, Field 0
    jmpif v4 then: b1, else: b2
  b1():
    v5 = load v2 -> Field
    return
  b2():
    return
}
```
We want to map the instructions from `b3` first, not b1 and b2. In the erroneous way we are mapping on master we could potentially be keeping the same `load v2` in `b1` while `v2 = allocate` is updated to a new value. The panics from the missing `else` block in the issue were occurring due to this same bug. 

I noticed this as the cause of the bug when printing the SSA would get rid of all the bugs listed in the issue. This was due to the normalization pass which would make the reverse post order of the blocks line up with their consecutive numerical ordering. 

## Additional Context

This ended up being fixed inadvertently by the preprocessing PR. The last failing nightly was `nightly-2025-01-20`. I found the fix on the failing nightly and brought it over to master even though the test is technically passing on master already. The bug most likely did not come up previously due to mem2reg successfully removing references which may be shared across blocks and the preprocess PR further helped clear out these references which may be in numerically smaller children blocks.  

I wanted to add a minor unit test to make sure we are mapping in the correct block order but this would require some extra changes to the SSA parser (putting a flag on ID normalization would not be enough as the parser also inserts at fresh IDs). 

Perhaps we should add some extra safety checks or comments for using `reachable_blocks`.  

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench (approved):
I've reduced the regression tests a bit and confirmed that I get panics when compiling them with the old block ordering.

PR #7144, opened 2025-01-22
fix: preserve types when reading from calldata arrays

### LINKED ISSUE #7143: Types are not preserved when reading dynamic indices from calldata
This was found by @aakoshh 

Currently when reading from calldata databus arrays it's assumed that the base numeric types are all `Field`.

https://github.com/noir-lang/noir/blob/a1cf830b3cdf17a9265b8bdbf366d65c253f0ca4/compiler/noirc_evaluator/src/acir/mod.rs#L1337

This results in further errors where a bitwise not which is intended to be applied to an integer type is instead applied to a `Field` type which is invalid.

TomAFrench:
# Description

## Problem\*

Resolves #7143

## Summary\*

This pulls out a fix that @aakoshh made in the sync PR so we have a regression test for it.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/acir/mod.rs
+++ b/compiler/noirc_evaluator/src/acir/mod.rs
@@ -1334,7 +1334,7 @@ impl<'a> Context<'a> {
         typ: &Type,
     ) -> Result<AcirValue, RuntimeError> {
         match typ {
-            Type::Numeric(_) => self.array_get_value(&Type::field(), call_data_block, offset),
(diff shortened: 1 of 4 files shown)


aakoshh (approved):
Thanks Tom, nice test! I wasn't sure whether a cast had to happen, but also didn't see any other reason why `typ` was even passed to the function, so I thought it was a typo.

PR #7145, opened 2025-01-22
chore: turn on overflow checks in CI rust tests

TomAFrench:
# Description

## Summary\*

This PR runs all of our rust tests in debug mode so that we can catch overflows such as in #7143 which appear in tests.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/.github/workflows/test-rust-workspace.yml
+++ b/.github/workflows/test-rust-workspace.yml
@@ -29,7 +29,7 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
         with:
-          key: x86_64-unknown-linux-gnu
+          key: x86_64-unknown-linux-gnu-debug
           cache-on-failure: true
           save-if: ${{ github.event_name != 'merge_group' }}
 
--- a/.github/workflows/test-rust-workspace.yml
+++ b/.github/workflows/test-rust-workspace.yml
@@ -39,7 +39,7 @@ jobs:
           tool: nextest@0.9.67
 
       - name: Build and archive tests
-        run: cargo nextest archive --workspace --release --archive-file nextest-archive.tar.zst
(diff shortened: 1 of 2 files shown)


TomAFrench:
We may need to split the incremental build caches as a result of this.

TomAFrench:
Confirming that this would catch #7143 and the slowdown in testing is ~1min so let's move forward on this.

.github/workflows/test-rust-workspace.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           tool: nextest@0.9.67
 
       - name: Build and archive tests
-        run: cargo nextest archive --workspace --release --archive-file nextest-archive.tar.zst
+        run: cargo nextest archive --workspace --archive-file nextest-archive.tar.zst
```
aakoshh, line 42: What does this archiving do?

aakoshh: Never mind, I see it's used below. Didn't know about this feature.

aakoshh:
I see the slowdown as 5 minutes in the build phase:
* it took almost 11 minutes now: https://github.com/noir-lang/noir/actions/runs/12906993835/job/35989513371?pr=7145
* while it took just less than 6 in another PR: https://github.com/noir-lang/noir/actions/runs/12905835556/job/35985880783

aakoshh (commented):
Would it make sense to for example to let the `msrv` test to do it in `debug` mode, while the other workflow does it in `release`, so we cover both?

TomAFrench:
Only the difference in timing on actually running the tests are relevant here.

aakoshh (approved):
Okay. I'd still be more confident if we covered both compilation modes, just in case `--release` does something unexpected, but happy to hear the performance impact is acceptable.

PR #7147, opened 2025-01-22
feat: optimize FieldElement::num_bits

asterite:
# Description

## Problem

No issue. While sampling a compilation it landed on `FieldElement::num_bits` so I was wondering if optimizing it would have any visible effect.

## Summary

Before this PR `num_bits` would allocate the bytes in `Vec`, now I think there would be no allocations.

I didn't try it locally but it was a simple optimization so trying it here to see what CI says.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -31,6 +31,7 @@ dependencies = [
  "ark-bls12-381",
  "ark-bn254 0.5.0",
  "ark-ff 0.5.0",
+ "ark-std 0.5.0",
  "cfg-if",
  "hex",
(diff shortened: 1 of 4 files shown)


TomAFrench:
I don't think would show up in CI benchmarks as all of the allocations would be temporary and the time to count the numbers of bits in a field element would be insignificant when it comes to compilation.

Running heaptrack locally would show the difference, also you could add a criterion benchmark to track this.

TomAFrench (approved):

PR #7149, opened 2025-01-22
chore: let `Function::inlined` take a `should_inline_call` function

### LINKED ISSUE #7148: Expose `should_inline_call` from `Function.inlined` to allow more fine grained inlining control
In #7083 we allowed passing custom closures into the inliner to determine whether we should inline a call or not. We currently still have a single closure which defines standard logic for all of the inlining passes.

https://github.com/noir-lang/noir/blob/02056d6a3ca2932a0062552859195a4e3b11f9dc/compiler/noirc_evaluator/src/ssa/opt/inlining.rs#L81-L122

Which pass is being handled is determined by the `inline_no_predicates_functions` and `inline_infos` args. The first determines whether we're in the "Inlining (1st)" or "Inlining (2nd)" pass. The second arg, `inline_infos`, contains the functions which we want to start inlining from and is  overloaded by the pass added in #7072 to add a "weight" parameter to functions to determine whether we should inline sooner or delay until the full inlining pass.

We've then got a lot of flexibility on choosing how we inline calls but we're not making best use of it currently. For instance, we're currently not inlining:
- functions which are empty and just return inputs.
- functions which contain a single instruction and return the inputs.
- functions which are only called once in the entire program.

These are optimization cases which I can see are not being handled currently in the protocol circuits (at least before the main inlining pass).

This should then allow us to more aggressively inline these calls without needing to manipulate the weight score which is being applied to functions currently.

asterite:
# Description

## Problem

Part of #7148

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/inlining.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/inlining.rs
@@ -68,8 +68,16 @@ impl Ssa {
         // instead of inlining the "leaf" functions, moving up towards the entry point.
         self.functions = btree_map(inline_targets, |entry_point| {
             let function = &self.functions[&entry_point];
-            let new_function =
-                function.inlined(&self, inline_no_predicates_functions, inline_infos);
+            let should_inline_call =
+                |_context: &PerFunctionContext, ssa: &Ssa, called_func_id: FunctionId| -> bool {
+                    function.should_inline_call(
+                        ssa,
+                        called_func_id,
+                        inline_no_predicates_functions,
+                        inline_infos,
+                    )
+                };
+            let new_function = function.inlined(&self, &should_inline_call);
             (entry_point, new_function)
         });
         self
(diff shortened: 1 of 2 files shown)


TomAFrench:
If we're marking this as resolving #7148, can you break out issues for all the extra inlining conditions it specifies?

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
```diff
@@ -68,8 +68,16 @@ impl Ssa {
         // instead of inlining the "leaf" functions, moving up towards the entry point.
         self.functions = btree_map(inline_targets, |entry_point| {
             let function = &self.functions[&entry_point];
-            let new_function =
-                function.inlined(&self, inline_no_predicates_functions, inline_infos);
+            let should_inline_call =
+                |_context: &PerFunctionContext, ssa: &Ssa, called_func_id: FunctionId| -> bool {
```
TomAFrench, line 72: Can you remove the `_context` arg if this is forcing us to expose `PerFunctionContext` beyond the inlining pass?

asterite:
I changed it to be "part of".

Regarding the extra inlining conditions, are those to be done in `preprocess_fns.rs`?

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
```diff
@@ -81,47 +89,47 @@ impl Function {
...
-                    return false;
-                }
-                let callee = &ssa.functions[&called_func_id];
+    ) -> bool {
+        // Do not inline self-recursive functions on the top level.
+        // Inlining a self-recursive function works when there is something to inline into
+        // by importing all the recursive blocks, but for the entry function there is no wrapper.
+        if called_func_id == self.id() {
+            return false;
+        }
```
TomAFrench, line 110: Note that these checks should be enforced even if we're passing a different inlining condition in as the closure as theyre important safety checks. Can you pull them out of this function and ensure they're always enforced?

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
```diff
@@ -81,47 +89,47 @@ impl Function {
...
-                            .get(&called_func_id)
-                            .map(|info| info.should_inline)
-                            .unwrap_or_default()
-                    }
+                !inline_type.is_entry_point() && !preserve_function
+            }
+            RuntimeType::Brillig(_) => {
+                if self.runtime().is_acir() {
+                    // We never inline a brillig function into an ACIR function.
+                    return false;
```
TomAFrench, line 127: ditto here and for inlining ACIR into brillig.

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
```diff
@@ -81,47 +89,47 @@ impl Function {
     pub(super) fn inlined(
         &self,
         ssa: &Ssa,
+        should_inline_call: &impl Fn(&PerFunctionContext, &Ssa, FunctionId) -> bool,
+    ) -> Function {
+        InlineContext::new(ssa, self.id()).inline_all(ssa, &should_inline_call)
+    }
+
+    /// Generic function that determines whether a function should inline a call.
+    pub(super) fn should_inline_call(
```
TomAFrench, line 98: We don't want to enshrine this as the only way of determining whether a function call should be inlined or not (quite the opposite) so I don't think it should be defined as a method on `Function`.

TomAFrench:
> at least the first two which seem simpler than the last one

This should be able to be calculated using `compute_times_called`.

TomAFrench:
> Regarding the extra inlining conditions, are those to be done in preprocess_fns.rs?

I don't think so as `preprocess_fns` is at risk of becoming a mega-pass which does too much. These early inlining passes can be done prior to this.

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
asterite: > and for inlining ACIR into brillig

Is that a new check that should be added?

asterite:
I addressed all the comments (I think). To avoid a double lookup in `ssa.functions` I changed the callback to receive `&Function`, and now it doesn't need to take an `&Ssa` anymore (and for the future optimizations I think that'll also remain like that, as we can check what's in a function without going through `Ssa`). Let me know if that's okay, otherwise we can revert the last commit.

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
```diff
@@ -882,14 +872,46 @@ impl<'function> PerFunctionContext<'function> {
...
+                    return None;
+                }
+            }
+            RuntimeType::Brillig(_) => {
+                if self.source_function.runtime().is_acir() {
+                    // We never inline a brillig function into an ACIR function.
+                    return None;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
asterite: Done! I couldn't easily check that the two runtime types match (because of `InlineType`) so I compared the bools of `is_acir()`.

asterite:
There's clearly something wrong but I can't tell what. I'll continue debugging this tomorrow.

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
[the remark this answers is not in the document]
asterite: This change broke things. It seems we so inline acir functions into brillig so I reverted that last commit.

asterite:
It's working! The performance alter was likely due to things not properly inlining anymore.

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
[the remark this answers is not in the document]
TomAFrench: Huh, can you open an issue to investigate this as it's not immediately obvious why this should happen after the SSA gen split?

My money would be on some special-cased functions such as built-ins / blackbox functions not being duplicated in the same way that regular noir functions are handled.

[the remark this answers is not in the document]
asterite: I just checked. One program that was failing was `array_sort`, which I think doesn't have built-ins or blackbox functions. Printing the conflicting functions, it was trying to inline an acir lambda into a brillig function. Maybe lambdas aren't separated?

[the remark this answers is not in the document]
asterite: Here's a small program that reproduces the issue:

```noir
fn main() -> pub Field {
    foo(|x| x + 1)
}

fn foo(f: fn(Field) -> Field) -> Field {
    unsafe {
        bar(f)
    }
}

unconstrained fn bar(f: fn(Field) -> Field) -> Field {
    f(1)
}
```

The SSA is this:

```
acir(inline) fn main f0 {
  b0():
    v2 = call f2(f1) -> Field
    return v2
}
acir(inline) fn lambda f1 {
  b0(v0: Field):
    v2 = add v0, Field 1
    return v2
}
acir(inline) fn foo f2 {
  b0(v0: function):
    v2 = call f3(v0) -> Field
    return v2
}
brillig(inline) fn bar f3 {
  b0(v0: function):
    v2 = call v0(Field 1) -> Field
    return v2
}
```

What's the SSA pass that separates runtimes?

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
```diff
@@ -205,7 +185,7 @@ pub(super) struct InlineInfo {
     is_brillig_entry_point: bool,
     is_acir_entry_point: bool,
     is_recursive: bool,
-    should_inline: bool,
+    pub(super) should_inline: bool,
```
TomAFrench, line 188:
```suggestion
    should_inline: bool,
```

TomAFrench (approved):
LGTM

Issue #7109, opened 2025-01-17
Hoist constants shared across functions into the global space


vezenovm:
### Problem

Post https://github.com/noir-lang/noir/pull/7021 we will have globals from SSA fully implemented in Brillig. 

However, we do not account for shared local  constants across functions. For example let's take this simplified version of `global_var_regression_simple`
```noir
global EXPONENTIATE: [[Field; 2]; 2] = [[1, 1], [0, 0]];

fn main(x: Field, y: pub Field) {
    let mut acc: Field = 0;
    for i in 0..2 {
        for j in 0..2 {
            acc += EXPONENTIATE[i][j];
        }
    }
    assert(x != y);
    dummy_again(x, y);
}
fn dummy_again(x: Field, y: Field) {
    let mut acc: Field = 0;
    for i in 0..2 {
        for j in 0..2 {
            acc += EXPONENTIATE[i][j];
        }
    }
    assert(x != y);
}
```
Running `nargo info --force-brillig --inliner-aggressiveness -9223372036854775808 --profile-execution` will give you 178 opcodes executed.

If we made a `global ZERO: Field = 0` and used that to initialize `acc` in both functions the opcodes executed would go down to 176. This is a pretty small example, but for larger programs with lots of constants performing this hoisting could be quite beneficial. 

### Happy Case

We should be able to automatically hoist constants shared across functions into the global space. We should not have any repeated initialization of constants. 

### Workaround

Yes

### Workaround Description

Manually writing globals that are shared across functions 

### Project Impact

Nice-to-have

### Blocker Context

This provides execution benefits, so even though it is not a true blocker, it is high priority. 

### Would you like to submit a PR for this Issue?

None

PR #7153, opened 2025-01-22
feat(ssa): Reuse constants from the globals graph when making constants in a function DFG

vezenovm:
# Description

## Problem\*

Does not resolve any issues, just a small optimization I noticed while working on #7109. This PR is much simpler than what #7109 requires as it checks for constants which have been already declared as globals by the program. 

## Summary\*

I keep the `constants` context map from the DFG in the `GlobalsGraph`. Now when checking whether to reuse a value id when making a constant we also check whether that constants exists in the globals space.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench (approved):

TomAFrench:
@vezenovm Note that this PR has caused a measurable regression in execution speed of `rollup-block-root` https://noir-lang.github.io/noir/dev/bench/

PR #7156, opened 2025-01-22
chore(refactor): Remove globals field on Ssa object and use only the shared globals graph

vezenovm:
# Description

## Problem\*

No issue, but just a general refactor. We have repeated declaration of globals (the shared `globals: Arc<GlobalsGraph>` field on `DataFlowGraph` and the `globals: Function` on `Ssa`). We need the globals field on the DFG for resolving constants https://github.com/noir-lang/noir/pull/7040. However, the globals field on the `Ssa` object is only used by the SSA printer and the brillig codegen. Which in both case could use the globals graph from any function.

## Summary\*

I removed `globals: Function` from `Ssa` and use the globals graph from a function in the places where the `Ssa` field was being used. 

Certain interfaces such as the printer and Brillig codegen still utilize the `DataFlowGraph` type so some conversions from a `GlobalsGraph` to a `DataFlowGraph` were necessary to avoid larger changes. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/brillig/mod.rs
+++ b/compiler/noirc_evaluator/src/brillig/mod.rs
@@ -89,10 +89,7 @@ impl Ssa {
             return brillig;
         }
 
-        let func_id = brillig_reachable_function_ids
-            .first()
-            .expect("ICE: Should have a reachable function at this point");
-        let globals = (*self.functions[func_id].dfg.globals).clone();
+        let globals = (*self.functions[&self.main_id].dfg.globals).clone();
         let (artifact, brillig_globals, globals_size) =
             convert_ssa_globals(enable_debug_trace, globals, &self.used_global_values);
         brillig.globals = artifact;
--- a/compiler/noirc_evaluator/src/brillig/mod.rs
+++ b/compiler/noirc_evaluator/src/brillig/mod.rs
@@ -89,6 +89,8 @@ impl Ssa {
             return brillig;
         }
 
+        // Globals are computed once at compile time and shared across all functions,
+        // thus we can just fetch globals from the main function.
         let globals = (*self.functions[&self.main_id].dfg.globals).clone();
         let (artifact, brillig_globals, globals_size) =
             convert_ssa_globals(enable_debug_trace, globals, &self.used_global_values);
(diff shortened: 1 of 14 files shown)


compiler/noirc_evaluator/src/brillig/mod.rs
```diff
@@ -85,8 +85,16 @@ impl Ssa {
 
         let mut brillig = Brillig::default();
 
+        if brillig_reachable_function_ids.is_empty() {
+            return brillig;
+        }
+
+        let func_id = brillig_reachable_function_ids
```
michaeljklein, line 92: What is this `func_id`? Is it [this](https://github.com/noir-lang/noir/blob/c8d5ce5e72f3171f24e4fe2b94ac098ba6d99ef6/compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs#L76C1-L80C42) `Function`?

compiler/noirc_evaluator/src/ssa/opt/normalize_value_ids.rs
```diff
@@ -25,7 +26,8 @@ impl Ssa {
         let mut context = Context::default();
         context.populate_functions(&self.functions);
         for function in self.functions.values_mut() {
-            context.normalize_ids(function, &self.globals);
+            let globals = (*function.dfg.globals).clone();
```
michaeljklein, line 29: Is this `clone` before passing a reference avoidable?

michaeljklein (changes requested):

compiler/noirc_evaluator/src/brillig/mod.rs
vezenovm: No it is just the first function from `brillig_reachable_function_ids`. The globals graph is computed a single time and shared among all functions https://github.com/noir-lang/noir/pull/7040 so we can use any func id here. I switched to using `self.main_id` to be more explicit.

compiler/noirc_evaluator/src/ssa/opt/normalize_value_ids.rs
vezenovm: It is not as `GlobalsGraph` cannot be moved out of `Arc` as it does not implement `Copy`. I have moved it out of the loop though.

jfecher (approved):
LGTM

michaeljklein (approved):

PR #7159, opened 2025-01-23
feat: add array range and reverse functions

DeVikingMark:
## Description:
This PR adds two utility functions to the array module:
- `range(size: Field)` - creates an array of numbers from 0 to size-1
- `reverse()` - returns a new array with elements in reverse order

Resolves TODO comments from test_programs/noir_test_success/brillig_oracle/src/main.nr

**Both functions are fully tested and documented with examples.**

### DIFF
--- a/noir_stdlib/src/slice.nr
+++ b/noir_stdlib/src/slice.nr
@@ -67,6 +67,34 @@ impl<T> [T] {
         ret
     }
 
+    // Apply a function to each element of the slice with its index, returning a
+    // new slice containing the mapped elements.
+    pub fn mapi<U, Env>(self, f: fn[Env](u32, T) -> U) -> [U] {
+        let mut ret = &[];
+        let mut index = 0;
+        for elem in self {
+            ret = ret.push_back(f(index, elem));
+            index += 1;
+        }
+        ret
+    }
+
+    // Apply a function to each element of the slice
+    pub fn for_each<Env>(self, f: fn[Env](T) -> ()) {
+        for elem in self {
+            f(elem);
+        }
+    }
+
+    // Apply a function to each element of the slice with its index
+    pub fn for_eachi<Env>(self, f: fn[Env](u32, T) -> ()) {
+        let mut index = 0;
+        for elem in self {
+            f(index, elem);
+            index += 1;
+        }
+    }
+
     // Apply a function to each element of the slice and an accumulator value,
     // returning the final accumulated value. This function is also sometimes
     // called `foldl`, `fold_left`, `reduce`, or `inject`.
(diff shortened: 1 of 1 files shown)


noir_stdlib/src/array/mod.nr
```diff
@@ -5,6 +5,24 @@ use crate::runtime::is_unconstrained;
...
+/// This is a utility function to create arrays from ranges.
+///
+/// Example:
+/// ```noir
+/// fn main() {
+///     let arr = std::array::range(5);
+///     assert(arr == [0, 1, 2, 3, 4]);
+/// }
+/// ```
+pub fn range<T>(size: Field) -> [Field] {
```
aakoshh, line 18: The `T` generic parameter is not used, and technically this returns a [slice](https://github.com/noir-lang/noir/blob/1cb40881b0de2d9fd5ac1700cbb78cead6e49097/noir_stdlib/src/slice.nr), rather than an array. 
`size` is usually expressed as `u32`.
Returning `Field` seems a bit specific; the library so far is completely generic.

Perhaps something like [std::array::from_fn](https://doc.rust-lang.org/nightly/core/array/fn.from_fn.html) would be a more flexible alternative to make an actual array of a given size, with content created for each index given to a lambda as `u32` between 0 and `N`. You could use it to cast to `Field`, or map it to a range other than 0..N, e.g. by an offset.

DeVikingMark:
@aakoshh hey Akosh, I've made an update, just curious what are ur thoughts abt it?

aakoshh:
@DeVikingMark sorry for the late reply. Thanks for the changes, it's looking better! 

All that's left is this part:

> and technically this returns a [slice](https://github.com/noir-lang/noir/blob/1cb40881b0de2d9fd5ac1700cbb78cead6e49097/noir_stdlib/src/slice.nr), rather than an array.

To returned an array (which are fixed size), the functions signatures should look like this:

```rust
pub fn from_fn<T, let N: u32>(f: fn[Env](u32) -> T) -> [T; N];
pub fn range<let N: u32>(size: u32) -> [u32; N];
```

It will figure out the `N` from the type you're trying to assign to.

aakoshh:
Another remark is that these functions could be added to `BoundedVec` as well.

noir_stdlib/src/array/mod.nr
```diff
@@ -5,6 +5,74 @@ use crate::runtime::is_unconstrained;
...
+/// This is a utility function to create arrays with custom initialization.
+///
+/// Example:
+/// ```noir
+/// fn main() {
+///     let arr = std::array::from_fn(5, |i| i * 2);
+///     assert(arr == [0, 2, 4, 6, 8]);
+/// }
+/// ```
+pub fn from_fn<T>(size: u32, f: fn(u32) -> T) -> [T] {
```
aakoshh, line 18: This function does not return an array, they return a slice. I suggest we remove it, and rename the `from_fn_fixed` to `from_fn` instead.

noir_stdlib/src/array/mod.nr
```diff
@@ -5,6 +5,74 @@ use crate::runtime::is_unconstrained;
...
+/// This is a utility function to create arrays from ranges.
+///
+/// Example:
+/// ```noir
+/// fn main() {
+///     let arr = std::array::range(5);
+///     assert(arr == [0, 1, 2, 3, 4]);
+/// }
+/// ```
+pub fn range(size: u32) -> [u32] {
```
aakoshh, line 36: Same as above: it returns a slice with a dynamic length. Should be removed in favour of renaming `range_fixed` to `range`.

noir_stdlib/src/array/mod.nr
```diff
@@ -5,6 +5,74 @@ use crate::runtime::is_unconstrained;
...
+///     assert(arr == [0, 1, 2, 3, 4]);
+/// }
+/// ```
+pub fn from_fn_fixed<T, let N: u32>(size: u32, f: fn(u32) -> T) -> [T; N] {
+    assert(size <= N, "Size cannot exceed array capacity N");
+    
+    let mut result = [crate::mem::zeroed(); N];
+    for i in 0..size {
+        result[i] = f(i);
+    }
```
aakoshh, line 57:
```suggestion
/// fn main() {
///     let arr: [Field; 5] = std::array::from_fn(|i| i as Field);
///     assert(arr == [0, 1, 2, 3, 4]);
/// }
/// ```
pub fn from_fn<T, let N: u32>(f: fn(u32) -> T) -> [T; N] {
    let mut result = [crate::mem::zeroed(); N];
    for i in 0..N {
        result[i] = f(i);
    }
```
It's simpler this way, I don't think the partial fill was a good suggestion on my part, sorry.

noir_stdlib/src/array/mod.nr
```diff
@@ -5,6 +5,74 @@ use crate::runtime::is_unconstrained;
...
+/// Example:
+/// ```noir
+/// fn main() {
+///     let arr: [u32; 5] = std::array::range_fixed(5);
+///     assert(arr == [0, 1, 2, 3, 4]);
+/// }
+/// ```
+pub fn range_fixed<let N: u32>(size: u32) -> [u32; N] {
+    from_fn_fixed(size, |i| i)
+}
```
aakoshh, line 74:
```suggestion
///
/// Example:
/// ```noir
/// fn main() {
///     let arr: [u32; 5] = std::array::range();
///     assert(arr == [0, 1, 2, 3, 4]);
/// }
/// ```
pub fn range<let N: u32>() -> [u32; N] {
    from_fn(|i| i)
}
```

noir_stdlib/src/array/mod.nr
```diff
@@ -5,6 +5,74 @@ use crate::runtime::is_unconstrained;
...
+///     assert(arr == [0, 1, 2, 3, 4]);
+/// }
+/// ```
+pub fn range(size: u32) -> [u32] {
+    from_fn(size, |i| i)
+}
+
+/// Creates a new fixed-size array by applying a function to each index.
+/// This is a utility function to create arrays with custom initialization.
+/// The size parameter can be less than N to partially fill the array.
```
aakoshh, line 42:
```suggestion
```

PR #7160, opened 2025-01-23
feat: inline simple functions

### LINKED ISSUE #7148: Expose `should_inline_call` from `Function.inlined` to allow more fine grained inlining control

asterite:
# Description

## Problem

Part of #7148

## Summary

Pending: add a few tests for this.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa.rs
+++ b/compiler/noirc_evaluator/src/ssa.rs
@@ -153,6 +153,7 @@ fn optimize_all(builder: SsaBuilder, options: &SsaEvaluatorOptions) -> Result<Ss
         .run_pass(Ssa::remove_unreachable_functions, "Removing Unreachable Functions (1st)")
         .run_pass(Ssa::defunctionalize, "Defunctionalization")
         .run_pass(Ssa::inline_simple_functions, "Inlining simple functions")
+        .run_pass(Ssa::mem2reg, "Mem2Reg (1st)")
         .run_pass(Ssa::remove_paired_rc, "Removing Paired rc_inc & rc_decs")
         .run_pass(
             |ssa| ssa.preprocess_functions(options.inliner_aggressiveness),
@@ -160,7 +161,7 @@ fn optimize_all(builder: SsaBuilder, options: &SsaEvaluatorOptions) -> Result<Ss
         )
         .run_pass(|ssa| ssa.inline_functions(options.inliner_aggressiveness), "Inlining (1st)")
         // Run mem2reg with the CFG separated into blocks
-        .run_pass(Ssa::mem2reg, "Mem2Reg (1st)")
+        .run_pass(Ssa::mem2reg, "Mem2Reg (2nd)")
         .run_pass(Ssa::simplify_cfg, "Simplifying (1st)")
         .run_pass(Ssa::as_slice_optimization, "`as_slice` optimization")
         .run_pass(Ssa::remove_unreachable_functions, "Removing Unreachable Functions (2nd)")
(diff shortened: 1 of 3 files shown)


TomAFrench:
Very nice! There's some regressions on keccak256 when using a zero aggressiveness, can you take a look (non blocker but it would need an issue)

asterite:
Sure! I'll take a look at it before merging this.

asterite:
@TomAFrench I was wondering if the `--inliner-aggressiveness` argument should be taken into account by this new pass or not. It's maybe strange that `--inliner-aggressiveness=0` would still inline these, but I'm not sure.

(so far the only difference I found was that a `to_le_radix` call is inlined now, and that somehow makes things work with zero inlining)

asterite:
I think I'll merge this PR and then look at that issue because it seems to need an extra fix.

asterite:
Regarding keccak256, what I found was that `std::hash::keccak256` calls the one in another module:

https://github.com/noir-lang/noir/blob/2da0a4f129bbd2183e87a7e5dd07127527fc2195/noir_stdlib/src/hash/mod.nr#L118-L122

And that function is `no_predicates`. So what's end up happening is that the call to the `no_predicates` function ends up in the caller... but maybe that's fine? After all the wrapper function doesn't have `no_predicates`. From there I still couldn't understand why it changes the final SSA.

vezenovm:
> `rollup-block-root-single-tx`	`16210` MB	`6250` MB	`2.59`

Looks like we have a really big regression here as well reversing the gains made by https://github.com/noir-lang/noir/pull/7072.

TomAFrench:
@asterite Chucking a mem2reg pass afterwards should help

asterite:
> @asterite Chucking a mem2reg pass afterwards should help

Done.

Regarding `keccak256`, I think now that there are some inlined functions before the preoprocessing, the weight of functions changes and so that pass behaves differently... in this case maybe it makes it into a regression. It's a bit hard to debug or understand 😊

TomAFrench:
Looks like the regression is fixed now but there's a test failure :thinking:

TomAFrench:
Yeah, inlining with weight scores is a little bit of a pain for that reason. Let's just record it and once we've done all of these optimizations which definitely should reduce bytecode sizes then we can see if there's ways to improve these cases.

vezenovm:
> Looks like the regression is fixed now but there's a test failure 🤔

I ran into the same thing in https://github.com/noir-lang/noir/pull/7165 and fixed it there. I'll extract it.

vezenovm:
Merging in https://github.com/noir-lang/noir/pull/7169 should fix those panics in https://github.com/noir-lang/noir/actions/runs/12935521653/job/36079229284?pr=7160

vezenovm:
I guess linking the comment we want to resolve will close the PR 😅

TomAFrench (approved):

asterite:
> Looks like the regression is fixed

Where can this be seen?

PR #7163, opened 2025-01-23
fix: Keep `inc_rc` for array inputs during preprocessing 

aakoshh:
# Description

## Problem\*

Backport of https://github.com/AztecProtocol/aztec-packages/pull/11448 

## Summary\*

See original PR.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Those increases in circuit sizes are pretty suspect. What's going on there?

test_programs/execution_success/regression_11294/Prover.json
TomAFrench, line 1: why not just run this through a json to toml converter?

aakoshh: This was captured from the integration tests in aztec-packages. You're right if we were adamant that we only handle TOML files then we could transform it, but I personally find nested TOML almost impossible to read compared to JSON, so I thought it would be a good idea not to have to involve yet another step where things can potentially go wrong, and run the tool in the output that is most convenient to capture. Especially since we have both formats, but no way to instruct it about which one to parse.

vezenovm: > I personally find nested TOML almost impossible to read compared to JSON

I do agree with this as well

TomAFrench: Can we keep that separate from this PR though? There's knock on effects of accepting new input formats which this PR isn't dealing with.

aakoshh: I can certainly open another PR, although I thought this was a good test that the input is actually accepted.

aakoshh: https://github.com/noir-lang/noir/pull/7170

TomAFrench (approved):

TomAFrench:
Ok. I'm happy for you to apply whichever fix is felt most appropriate as long as we don't block that on adding half of a new feature to avoid going to https://transform.tools/json-to-toml

test_programs/execution_success/regression_11294/Prover.json
aakoshh: Changed to TOML to unblock.

aakoshh:
> Those increases in circuit sizes are pretty suspect. What's going on there?

I looked yesterday at `brillig_fns_as_values`:
* I changed the rule of what to inline to remove the cutoff at the mean weight to make it easier (🤞) to match the stats of a production circuit in an integration test. For example in `private_kernel_inner` the `main` is bigger than what it is in `regression_11249`, and because of that the mean weight in the test was so much smaller that what got preprocessed in production was skipped as "too heavy" in the test. Instead the new rule relies only on what a function calls below it, and if it's mostly calling others, it's not preprocessed, thinking that what it calls are the utility functions, and itself is more like an entry function. 
* IIRC the `wrapper` functions were considered heavy in this sense because they don't do much on their own, all their weight comes from who they call. Previously they were processed as their weight was lower than average. 
* OTOH on `master` there was the new `Inline Simple Functions` pass, which inlined these `wrapper` functions into `apply` because they were a single line. Then, master consisted of only `main` and `apply` as reachable functions.

After merging `master` the sizes went back because of the extra inlining before preprocessing.

PR #7165, opened 2025-01-23
fix(ssa): Check arrays used as a terminator argument in the RC tracker 

vezenovm:
# Description

## Problem\*

Resolves failures in https://github.com/AztecProtocol/aztec-packages/pull/11294 from the preprocessing pass. https://github.com/noir-lang/noir/pull/7163 is a bit too conservative with marking that we should keep the inc rcs on any parameters. Doing so happens to also fix the test case, but marking arrays passed to terminators is the more accurate fix for how the RC tracker should be looking at mutated arrays. 

## Summary\*

We want to make sure we do not remove inc rcs on arrays that are potentially mutated elsewhere. To do so, we check whether an array was passed to a store instruction or a call. We used to only run DIE (which removes these RCs) at the end of the SSA optimizations where we were aggressively inlining. 

Now that we are starting to process functions in isolation we triggered a bug as outlined here https://github.com/AztecProtocol/aztec-packages/pull/11294#issuecomment-2608216917. We need to account for whether an array is used in a terminator argument such as a return as that returned array may be mutated later.

## Additional Context

Will need to add a new test making sure we do not remove an inc rc on an array type used in a terminator argument.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/die.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/die.rs
@@ -965,6 +965,36 @@ mod test {
     }
 
     #[test]
+    fn do_not_remove_inc_rcs_for_arrays_in_terminator() {
+        let src = "
+        brillig(inline) fn main f0 {
+          b0(v0: [Field; 2]):
+            inc_rc v0
+            inc_rc v0
+            inc_rc v0
+            v2 = array_get v0, index u32 0 -> Field
+            inc_rc v0
+            return v0, v2
+        }
+        ";
+
+        let ssa = Ssa::from_str(src).unwrap();
+
+        let expected = "
+        brillig(inline) fn main f0 {
+          b0(v0: [Field; 2]):
+            inc_rc v0
+            v2 = array_get v0, index u32 0 -> Field
+            inc_rc v0
+            return v0, v2
+        }
+        ";
+
+        let ssa = ssa.dead_instruction_elimination();
+        assert_normalized_ssa_equals(ssa, expected);
+    }
+
+    #[test]
     fn do_not_remove_inc_rc_if_used_as_call_arg() {
         // We do not want to remove inc_rc instructions on values
         // that are passed as call arguments.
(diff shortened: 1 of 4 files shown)


compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -688,7 +672,7 @@ impl RcTracker {
                 // Treat any array-type arguments to calls as possible sources of mutation.
                 for arg in arguments {
                     let typ = function.dfg.type_of_value(*arg);
-                    if matches!(&typ, Type::Array(..) | Type::Slice(..)) {
+                    if matches!(&typ, Type::Array(_, _) | Type::Slice(_)) {
```
vezenovm, line 675:
```suggestion
                    if matches!(&typ, Type::Array(..) | Type::Slice(..)) {
```

aakoshh (approved):
Awesome, thanks!

vezenovm:
> Looking at the increase in the opcodes executed, one difference between this and mine is that I only did the special treatment during the preprocessing, but when the functions were fully inlined it was removed. Do you reckon something like that for the final pass (maybe piggybacking on the `flattened` flag) should be used?

Ah yes good call out. I think we can do that as a follow-up optimization as we want will want to have some better tests around under what circumstances we do not check the terminator. For now, let's just get this fix in.

aakoshh:
I tried commenting out preprocessing and the terminator marking separately and together and they did not make a difference in the output of this:

```
❯ cargo run -q -p nargo_cli -- --program-dir test_programs/execution_success/brillig_fns_as_values info --force --silence-warnings --skip-underconstrained-check --skip-brillig-constraints-check --force-brillig --inliner-aggressiveness 0

+-----------------------+----------+----------------------+--------------+-----------------+
| Package               | Function | Expression Width     | ACIR Opcodes | Brillig Opcodes |
+-----------------------+----------+----------------------+--------------+-----------------+
| brillig_fns_as_values | main     | Bounded { width: 4 } | 1            | 170             |
+-----------------------+----------+----------------------+--------------+-----------------+
| brillig_fns_as_values | main     | N/A                  | N/A          | 170             |
+-----------------------+----------+----------------------+--------------+-----------------+
```

aakoshh:
It's probably down to the change in the change in rules in which functions to preprocess.

PR #7170, opened 2025-01-23
feat(cli): Try Prover.json if Prover.toml is not present

aakoshh:
# Description

## Problem\*

Extracted from https://github.com/AztecProtocol/aztec-packages/pull/11294

## Summary\*

The `debug` and `execute` CLI commands accept `--program-dir` and a `--prover-name` argument to specify where to look for the inputs for the circuit. The prover name is `Prover` by default, but we can't specify e.g. `Prover.json` to make it take a JSON file, it always tries to use `Format::Toml` to parse it, which determines the `.toml` extension. 

The PR changes commands that read `Prover.toml` to try both `Format::Toml` and `Format::Json` (in that order), so that if `Prover.toml` is not present but `Prover.json` is, then we use the latter. 

## Additional Context

Ostensibly we could have a `--format` argument, but I found this useful when capturing the input witnesses from the  TypeScript integration tests [like here](https://github.com/AztecProtocol/aztec-packages/pull/11294#issuecomment-2604803423), and wanted to use it as-is to replicate the error [like this](https://github.com/AztecProtocol/aztec-packages/pull/11294#issuecomment-2607011177), without having to go through more hoops.

## Documentation\*

Check one:

# PR Checklist\*


vezenovm (commented):
Let's add the change for the `info` command as well. We need it for profiling Brillig execution.

vezenovm (approved):

TomAFrench (changes requested):
This conflicts with `nargo check` which will ignore json files.

TomAFrench:
If we're accepting this as a first class input format we then also need to support it in the profiler also, etc. etc.

aakoshh:
I see that the `check` command can create an `Prover.toml` template file, but I don't see it reading the inputs, other than to verify they exist before accidentally overwriting them 👀  

If that's all it does, then I don't think it necessarily conflicts, it just doesn't create a JSON template at the moment.

TomAFrench:
I think it conflicts if we're expecting users to have a `Prover.json` file but that then prevents them from running `nargo check` without their inputs being soft-overwritten by an empty template file.

aakoshh:
Maybe I'm misunderstanding something, but the `check` command at the moment is hardcoded to write a `Prover.toml` file, nothing else, it will ignore any JSON, won't write over it.

TomAFrench:
Yeah but if you then try and execute your program afterwards then `nargo` will pick up the empty toml file over the json file you've been using previously.

aakoshh:
Ah, by "soft overwrite" you meant shadowing, gotcha.

Thanks for pointing out the profiler, I updated that to handle both formats.

aakoshh:
Updated `check_cmd` to check that any prover file with a format other than `Format::Toml` exists, and if so then ask for `--overwrite` to be used, even though it will not be deleted, just shadowed by the TOML template.

aakoshh:
To highlight some of the differences with working with the different formats, I transformed [this JSON](https://github.com/AztecProtocol/aztec-packages/pull/11294#issuecomment-2605283346) witness to TOML with https://transform.tools/json-to-toml ; here's a part of it as it's opened in my editor:

Here's what a heavily hierarchical part looks like in JSON:
Screenshot 2025-01-24 at 09 45 37

The same in TOML as it's formatted by the tool:
Screenshot 2025-01-24 at 09 47 17

Luckily the tool indents the content to follow the nesting, but it's inconsistent: `[previous_kernel_public_inputs.constants]` is indented more than `[previous_kernel_public_inputs.constants.historical_header.state.l1_to_l2_message_tree]` below it. 

When I save the TOML in VSCode it gets rid of all indentation. Tried a setting in the plugin but it didn't make a difference, but I don't expect a uniform setting for everyone.

Another thing I found work better in the JSON editing is the collapse of arrays. The TOML editor can collapse structs like the`key_validation_requests_and_generators` here:

Screenshot 2025-01-24 at 09 55 22

But there is no wrapper for the `note_hashes` array below it, so those entries can only be collapsed individually, not as a whole group. 

In JSON this is easy, so helps to explore the document and delete regions that aren't interesting for the test:
Screenshot 2025-01-24 at 10 01 55

To TOML's credit a collapsed record still shows its full path which is more useful than seeing those `{}` above `note_hashes`.

aakoshh:
I think this is no longer interesting, the `artifact_cli` handles both IIRC.

Issue #7173, opened 2025-01-23
Recursive enums cause stack overflow


jfecher:
### Aim

```noir
fn main() {
    let c = Foo::C(Foo::C(Foo::C(3)));
}

enum Foo<T> {
    C(T),
    D,
}

```

### Expected Behavior

The program to compile with c's type being `Foo<Foo<Foo<Field>>>`

### Bug

fatal runtime error: stack overflow

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

Issue #7172, opened 2025-01-23
Enum variants with no fields must still be called like functions


jfecher:
### Aim

```noir
fn main() {
    let _foo: Foo = Foo::Bar;
}

enum Foo {
    Bar
}
```

### Expected Behavior

The program to compile

### Bug

```
error: Expected type Foo, found type fn() -> Foo
  ┌─ src/main.nr:2:21
  │
2 │     let _foo: Foo = Foo::Bar;
  │                     --------
  │
```
The compiler treats `Foo::Bar` as a function despite it being declared as `Bar` instead of `Bar()`.

Fixing this requires a new type of definition in Noir since `Foo::Bar` would normally be a global constant with its tag value, but this does not work for generic types since Noir doesn't support generic globals.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #7174, opened 2025-01-23
feat(experimental): Construct enum variants in expressions

jfecher:
# Description

## Problem\*

Working towards #988 
Remaining work:
- https://github.com/noir-lang/noir/issues/7173
- https://github.com/noir-lang/noir/issues/7172
- Match expressions
- Enums in the comptime interpreter

## Summary\*

Resolves an enum constructor expression like `Foo::Bar(1, 2, 3)` in an expression context. Enum values are handled all the way through monomorphization now and you can compile programs with them. You just can't do anything useful with them since you can't match on them yet.

Enum types are currently always represented as a tuple of `(tag, variant1_fields, variant2_fields, ..., variantN_fields)` with the unused variants being zeroed out. We can relax this constraint in the future e.g. with the user-defined representations given in #988 but that isn't needed for this initial work.

Enum variants like `Foo::Bar` above are first-class functions which return an enum value and can be mapped or passed around like any other first class function.

## Additional Context

One downside mentioned in https://github.com/noir-lang/noir/issues/7172 is that we don't have a way to express a global definition that is generic but not a function currently so enum variants with no fields must still be called as if they were functions. We can't represent e.g. `Option::None` as a global since it'd need to be generic but we can't have a global be generic over any types.

## Documentation\*

Check one:

# PR Checklist\*


asterite:
> Enum types are currently always represented as a tuple of (tag, variant1_fields, variant2_fields, ..., variantN_fields) with the unused variants being zeroed out

I might have asked this before but could enums be represented as `(tag, variant_with_the_maximum_length)`? Not a blocker for this PR.

compiler/noirc_frontend/src/elaborator/expressions.rs
```diff
@@ -75,7 +75,10 @@ impl<'context> Elaborator<'context> {
                 self.push_err(ResolverError::UnquoteUsedOutsideQuote { span: expr.span });
                 (HirExpression::Error, Type::Error)
             }
-            ExpressionKind::AsTraitPath(_) => todo!("Implement AsTraitPath"),
+            ExpressionKind::AsTraitPath(_) => {
+                self.push_err(ResolverError::UnquoteUsedOutsideQuote { span: expr.span });
+                (HirExpression::Error, Type::Error)
```
asterite, line 80: Should a different error be issued here?

compiler/noirc_frontend/src/hir/comptime/hir_to_display_ast.rs
```diff
@@ -212,6 +212,17 @@ impl HirExpression {
 
             // A macro was evaluated here: return the quoted result
             HirExpression::Unquote(block) => ExpressionKind::Quote(block.clone()),
+
+            // Convert this back into a function call `Enum::Foo(args)`
+            HirExpression::EnumConstructor(constructor) => {
+                let typ = constructor.r#type.borrow();
+                let variant = &typ.variant_at(constructor.variant_index);
+                let path = Path::from_single(variant.name.to_string(), span);
```
asterite, line 220: Should this path have both segments? Or it doesn't matter?

compiler/noirc_frontend/src/hir/def_collector/dc_mod.rs
```diff
@@ -324,6 +326,33 @@ impl<'a> ModCollector<'a> {
...
+        for struct_definition in types {
+            if let Some((id, the_enum)) = collect_enum(
+                &mut context.def_interner,
+                &mut self.def_collector.def_map,
+                &mut context.usage_tracker,
+                struct_definition,
+                self.file_id,
+                self.module_id,
+                krate,
+                &mut definition_errors,
```
asterite, line 348:
```suggestion
        for enum_definition in types {
            if let Some((id, the_enum)) = collect_enum(
                &mut context.def_interner,
                &mut self.def_collector.def_map,
                &mut context.usage_tracker,
                enum_definition,
                self.file_id,
                self.module_id,
                krate,
                &mut definition_errors,
```

compiler/noirc_frontend/src/hir/def_map/module_data.rs
```diff
@@ -32,7 +32,7 @@ pub struct ModuleData {
     pub is_contract: bool,
 
     /// True if this module is actually a struct
```
asterite, line 34:
```suggestion
    /// True if this module is actually a type
```

compiler/noirc_frontend/src/hir/def_map/module_data.rs
```diff
@@ -57,7 +57,7 @@ impl ModuleData {
             definitions: ItemScope::default(),
             location,
             is_contract,
-            is_struct,
+            is_type: is_struct,
```
asterite, line 60: nit: use `is_type` instead of `is_struct`

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1809,6 +1814,45 @@ impl<'context> Elaborator<'context> {
...
+
+            let self_type = Type::DataType(datatype.clone(), generics);
+            let self_type_id = self.interner.push_quoted_type(self_type.clone());
+            let unresolved = UnresolvedType {
+                typ: UnresolvedTypeData::Resolved(self_type_id),
+                span: typ.enum_def.span,
+            };
+
+            for (i, variant) in typ.enum_def.variants.iter().enumerate() {
+                self.interner.add_definition_location(ReferenceId::EnumVariant(*type_id, i), None);
```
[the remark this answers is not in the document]
asterite: Captured here: https://github.com/noir-lang/noir/issues/7177

asterite (approved):
Looks good! I left a few small suggestions.

compiler/noirc_frontend/src/elaborator/expressions.rs
jfecher: Yep, this should be reverted, it's remaining from debugging code

compiler/noirc_frontend/src/hir/comptime/hir_to_display_ast.rs
jfecher: Added

jfecher:
> I might have asked this before but could enums be represented as (tag, variant_with_the_maximum_length)? Not a blocker for this PR.

Good question, I have this in a comment somewhere in the PR as well but no we can't have them share space as easily. In the future we can try to share fields but arrays, etc are harder to handle. To help with this, there are some plans in #988 on how to let users choose the representation of the enum then provide conversion functions to and from that struct for each variant.

compiler/noirc_frontend/src/elaborator/expressions.rs
asterite: Sounds good. I think this panic makes LSP crash, so even if you don't want to write an `AsTraitPath` it might happen if it's an incomplete expression. It would be nice to at least change this panic to a user-error (in a separate PR, of course).

jfecher: That's how this was originally triggering while I was debugging as well.

When I was working on the type alias example in slack:

```noir
type Double<let N: u32> = N * 2;

fn main() {
    let _x: [Field; Double<2>] = [1, 2, 3, 4];
}
```
We were getting a syntax error on Double<2> and automatically correcting it to an AsTraitPath

PR #7179, opened 2025-01-24
chore(LSP): several improvements for enums and enum variants

asterite:
# Description

## Problem

Part of #988

## Summary

Now enum variants show up as they do in Rust Analyzer.

Completion:

image

Signature help:

image

Hover:

image

## Documentation

Check one:
- [x] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #7180, opened 2025-01-24
chore(experimental): Prevent enum panics by returning Options where possible instead of panicking

jfecher:
# Description

## Problem\*

Working towards #988

## Summary\*

The original enum PRs changed the representation of DataType to have a TypeBody which only conditionally contains struct fields. So the existing `get_fields` methods and similar were changed to panic if the type was not a struct for fewer changes. This PR changes these methods to return Options instead so that we can find all the sites that need to be updated to handle enums.

## Additional Context

Currently this PR contains the necessary changes in the frontend but not the language server. https://github.com/noir-lang/noir/pull/7179/files is expected to work on those changes so ideally these two PRs should be merged to update all cases in the compiler.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
This PR is ready for review now.
Should only contain fixes to stop code using enums from panicking where a struct was expected. Unsure what the circuit size changes are about - this should be up to date with master.

asterite (approved):

PR #7189, opened 2025-01-24
feat: simplify subtraction from self to return zero

### LINKED ISSUE #7187: Optimization: replace `sub vN, vN` with zero
While looking at the SSA for this code:

```noir
use std::hash::sha256::sha256;

fn main(input: [u8; 32]) -> pub [u8; 32] {
    sha256(input)
}
```

I found this instruction:

```
    v172 = sub v41, v41
```

I think that could be replaced with the constant zero. There are two occurrences of that in the final SSA. Maybe it could also help with constant propagation, etc.

I tried implementing this with this patch:

```diff
diff --git a/compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs b/compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
index df1e8f537d..14cddb7f41 100644
--- a/compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
@@ -103,6 +103,11 @@ impl Binary {
             );
         }

+        if self.lhs == self.rhs {
+            let zero = dfg.make_constant(FieldElement::zero(), lhs_type);
+            return SimplifyResult::SimplifiedTo(zero);
+        }
+
         let operator = if lhs_type == NumericType::NativeField {
             // Unchecked operations between fields or bools don't make sense, so we convert those to non-unchecked
             // to reduce noise and confusion in the generated SSA.
```

It works for a simple program like this one:

```noir
fn main(input: Field) -> pub Field {
    input - input
}
```

(it now returns `Field 0` instead of doing the subtraction)

but the compiler enters a loop in the sha256 case and I don't understand why. It seems it happens during unrolling.

TomAFrench:
# Description

## Problem\*

Resolves #7187

## Summary\*

This PR simplifies subtracting a value from itself to a constant value of zero.

@asterite, you ran into issues with this as you didn't place this optimiation inside the match arm for subtraction so it was being applied to all operations.

## Documentation\*

Check one:

# PR Checklist\*


vezenovm (approved):

vezenovm:
> `rollup-block-root`	`122` s	`84.3` s	`1.45`

This is quite the jump...

jfecher (approved):
LGTM. I'm suspicious of the rollup-block-root change. Could that really be from this one change?

I'm assuming it's not but if it is we really need to look at the efficiency of resolving values in general.

compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
```diff
@@ -155,6 +155,11 @@ impl Binary {
                 }
             }
             BinaryOp::Sub { .. } => {
+                if dfg.resolve(self.lhs) == dfg.resolve(self.rhs) {
```
jfecher, line 158:
```suggestion
                if lhs == rhs {
```
The rest of this function has been updated since the PR was opened to resolve these beforehand

TomAFrench:
I've reworked the failing regression test so that we continue to trigger the expected error. I've also opened #7240

PR #7190, opened 2025-01-25
fix(ssa): Make the lookback feature opt-in

### LINKED ISSUE #7171: Flip the SSA safety checks to be opt-in
As mentioned in #7060, these safety checks are taking up a significant portion of compilation time and prevent some of the protocol circuits from being able to be compiled at all.

We don't need to run these checks on every compilation, instead they should be opt-in when the developer is compiling their final circuit and wants to spend time to catch potential bugs which may still be lurking.

rkarabut:
# Description

## Problem

The Brillig constraints check can get slow on large rollout functions, with the recently added lookback feature, while allowing to catch some rare false positives, looking to be the main culprit. 

## Summary

This PR disables lookback by default and allows the developer to employ it at their discretion. Also, repeated code locations of brillig calls are filtered out, improving performance of the pass.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Even prior to the addition of this feature, this pass was causing compilation to hang on the protocol circuits so I'm in favour of making the entire pass opt in.

rkarabut:
> Even prior to the addition of this feature, this pass was causing compilation to hang on the protocol circuits so I'm in favour of making the entire pass opt in.

Alright. Gotta put some additional research into this, as in doing test runs working on the initial check protocol circuits worked fine for me (although I've had word of the hangups later). Removed "resolves #7171" from the description. I think the feature should be opt-in any case though.

rkarabut:
Added filtering for already visited code locations of the brillig calls, which seems to help a lot with huge unrolled functions. Probably could switch the check back as the default.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -9,13 +9,14 @@ use crate::ssa::ir::instruction::{Hint, Instruction, InstructionId, Intrinsic};
...
 use im::HashMap;
+use noirc_errors::Location;
 use rayon::prelude::*;
 use std::collections::{BTreeMap, BTreeSet, HashSet};
 use tracing::trace;
 
 /// The number of instructions that have to be passed to stop
 /// following a Brillig call, with assumption it wouldn't get constrained
-const BRILLIG_CONSTRAINT_SEARCH_DEPTH: usize = 1000;
+const BRILLIG_CONSTRAINT_SEARCH_DEPTH: usize = 10_000_000;
```
aakoshh, line 19: Is this 1000, now 10 million, the same as the one mentioned [in the error](https://github.com/noir-lang/noir/pull/7190/files#diff-79227328f8c2f51f153be0e0732254f9c1a731d14477e7d94d184fd5c47578c5R125)? 

Maybe there is a way to keep the two in sync? 
Is the 10M the actual value you want to use, or was it used for debugging?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -42,7 +43,10 @@ impl Ssa {
 
     /// Detect Brillig calls left unconstrained with manual asserts
     /// and return a vector of bug reports if any have been found
-    pub(crate) fn check_for_missing_brillig_constraints(&mut self) -> Vec<SsaReport> {
+    pub(crate) fn check_for_missing_brillig_constraints(
+        &mut self,
+        enable_lookback: bool,
```
aakoshh, line 48: I wonder if it would make sense to make the search depth configurable?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Yeah it's temporary, there seems to be no need to limit search depth anymore with the location filtering added actually.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Could do that, but see above

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -263,12 +276,11 @@ impl BrilligTaintedIds {
...
-        // (skipped if there were no arguments, or if an actual result and not a
-        // descendant has been constrained _alone_, e.g. against a constant)
+        // (skipped if there were no arguments, or if a result descendant
+        // has been constrained _alone_, e.g. against a constant)
         if !results_involved.is_empty()
             && (self.arguments.is_empty()
-                || (constrained_values.len() == 1
-                    && self.root_results.intersection(constrained_values).next().is_some())
+                || (constrained_values.len() == 1)
                 || self.arguments.intersection(constrained_values).next().is_some())
```
aakoshh, line 284: Why is it no longer required for the arguments to contain the single constrained value?

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
aakoshh: By "location filtering", do you mean this?

```
    // Code locations of brillig calls already visited (we don't
    // need to recheck calls happening in the same unrolled functions)
    visited_locations: HashSet<Location>,
```

rkarabut: Yes

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
aakoshh: Ah, missed the `root_results`, I thought `arguments` were detached 🙈 

(Maybe a helper function for `this_set.intersection(that_set).next().is_some()` would be helpful, it appears 6 (now 5) times).

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
aakoshh: Awesome, then I suppose that means the error can be changed back to not mention 1000 either?

rkarabut: Yes, done that already

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -323,40 +335,56 @@ impl DependencyContext {
...
-                                }
+                        if all_functions[&callee].runtime().is_brillig() {
+                            // Skip already visited locations (happens often in unrolled functions)
+                            let call_stack = function.dfg.get_instruction_call_stack(*instruction);
+                            let location = call_stack.last();
+
+                            // If there is no call stack (happens for tests), consider unvisited
+                            let mut visited = false;
+                            if let Some(location) = location {
+                                visited = self.visited_locations.contains(location);
```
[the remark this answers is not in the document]
aakoshh: Thanks for the update!

On a different note, it looks like a lot of nesting/indent could be avoided with a pattern like this:

```rust
                let Instruction::Call { func, arguments } = &function.dfg[*instruction] else {
                    return;
                }
                let Value::Function(callee) = &function.dfg[*func] else {
                    return;
                }
                if !all_functions[callee].runtime().is_brillig() {
                    return;
                }

                if let Some(location) = location {
                    if !self.visited_locations.insert(location) {
                        return;
                    }
                }
                ...
```

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -323,40 +335,55 @@ impl DependencyContext {
...
+                                            .entry(*value)
+                                            .or_default()
+                                            .push(*instruction);
+                                    });
+
+                                    // Set the constraint search limit for the call
+                                    self.search_limits.insert(
+                                        block_index + BRILLIG_CONSTRAINT_SEARCH_DEPTH,
+                                        *instruction,
+                                    );
```
aakoshh, line 381: I'm trying to understand this. The docstring of `DependencyContext` says:
```
    // Map of block indices to Brillig call ids that should not be
    // followed after meeting them
    search_limits: HashMap<usize, InstructionId>,
```
and `block_index` comes from:
```
       function.dfg[block].instructions().iter().enumerate().for_each(
            |(block_index, instruction)| {
```
and we have a single `DependencyContext` per function.

Just looking at these, it would seem like instructions from multiple blocks will end up being mixed into the same `search_limits` collection, sometimes overriding each other, like the 5th instruction of every block will land in the same slot. 

Looking at how this is used below, I see we're iterating over the same instructions again, so the entries are not going to be immediately overwritten, and if we have one block with 1000+ (or 1M+) instructions then the fading out of earlier instructions can happen as expected, but since this data structure is not cleared between calls to `process_instructions`, it's not entirely clear how or if this clearing should work between blocks.

rkarabut: Actually I think that is true, but you know what as search limit doesn't seem to be required any longer let me just cut this entirely

rkarabut: Done, limiting removed

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -393,14 +409,6 @@ impl DependencyContext {
                 self.tracking_count += 1;
             }
 
-            // Stop tracking calls when their search limit is hit
-            if let Some(call) = self.search_limits.get(&block_index) {
-                if let Some(tainted_ids) = self.tainted.get_mut(call) {
-                    tainted_ids.tracking = false;
-                    self.tracking_count -= 1;
```
aakoshh, line 400: Now that `search_limits` is removed, is there something else that could be used as a signal to decrease `tracking_count`, so that we can skip the checks until we have something interesting to look at? For example when a constraint check on a tainted call is satisfied, could that mean we can decrease by 1?

rkarabut: Oh yeah, good catch

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -375,15 +389,17 @@ impl DependencyContext {
...
-                if let Some(calls) = self.call_arguments.get(argument) {
-                    for call in calls {
-                        if let Some(tainted_ids) = self.tainted.get_mut(call) {
-                            tainted_ids.tracking = true;
-                            self.tracking_count += 1;
+            // If the lookback feature is enabled, start tracking calls when
+            // their argument value ids first appear, or when their
+            // instruction id comes up (in case there were no non-constant arguments)
+            if self.enable_lookback {
+                for argument in &arguments {
```
aakoshh, line 394: I had to look up this comment I made with the illustration to remind myself what this was about: https://github.com/noir-lang/noir/pull/6658#issuecomment-2522699849 and the https://github.com/noir-lang/noir/issues/6735 issue we made. 

So we start tracking when the values which are going to be direct arguments to a tainted call first appear in an instruction. 

* What does it mean in simple terms to "start tracking calls"? 
* Is it okay to only do this when direct arguments are encountered, as opposed to any ancestor of an argument?
* Is it okay that the `arguments` does not contain the _results_ of the instruction? `for_each_value` only seems to visit the inputs.

rkarabut: * "start tracking" means begin to fill out the descendancy graph and partial constraints for these calls, otherwise we waste way too much time on unneeded checks
* yes, starting with ancestors doesn't make sense as the only valid argument descendants would come from directly involved values
* yes, outputs are not important

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
aakoshh: Isn't it true though that it gets decremented once when a trained thingy is successfully constrained, but it gets incremented more than once if `enable_lookback` is `true`?

rkarabut: Oh right. Fixed

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -319,11 +324,20 @@ impl DependencyContext {
...
+            if let Instruction::Call { func, arguments } = &function.dfg[*instruction] {
+                if let Value::Function(callee) = &function.dfg[*func] {
+                    if all_functions[&callee].runtime().is_brillig() {
+                        // Skip already visited locations (happens often in unrolled functions)
+                        let call_stack = function.dfg.get_instruction_call_stack(*instruction);
+                        let location = call_stack.last();
+
+                        // If there is no call stack (happens for tests), consider unvisited
+                        let visited = location
+                            .map(|loc| self.visited_locations.contains(loc))
```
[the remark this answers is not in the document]
rkarabut: After checking that it indeed doesn't make things worse on rollup-block-root, bringing this in. Done

Issue #7196, opened 2025-01-27
Add purity analysis optimization to deduplicate pure functions


jfecher:
### Problem

Function calls are treated as black boxes by most function passes. We don't have information on what they may do so we have to assume the worst. This means we can never e.g. deduplicate brillig function calls in acir which can lead to duplicate constraints if the called function is pure. This can also increase compile times due to additional work done even if these instructions are later optimized out.

### Happy Case

Add a purity analysis pass so that future SSA passes can query the purity of each function.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #7188, opened 2025-01-24
fix(brillig): Globals entry point reachability analysis 

vezenovm:
# Description

## Problem\*

Resolves #7158 

Was not sure whether to label this as a feat or a fix. It fixes repeated global initialization across multiple Brillig entry points, however, it is a pretty meaningful change in how we generate globals so I thought about labelling it a feat. As I labelled the issue a bug I have used the fix prefix. 

## Summary\*

We now track globals Brillig artifacts by their entry point rather than having a single globals artifact for all entry points. 

A few changes have been made:
1. We now track the `used_globals` by function rather than over the entire program.
2. We made a new `BrilligGlobals` context structure to simplify code gen for globals.
3. When generating `BrilligGlobals` we create some initial state from the function call graph.
```
    /// Both `used_globals` and `brillig_entry_points` need to be built
    /// from a function call graph.
    /// 
    /// Maps a Brillig function to the globals used in that function.
    /// This includes all globals used in functions called internally.
    used_globals: HashMap<FunctionId, HashSet<ValueId>>,
    /// Maps a Brillig entry point to all functions called in that entry point.
    /// This includes any nested calls as well, as we want to be able to associate
    /// any Brillig function with the appropriate global allocations.
    brillig_entry_points: HashMap<FunctionId, HashSet<FunctionId>>,
```
4. We then declare the globals for each entry point by looping over `brillig_entry_points`. 
5. We construct a new mapping `inner_call_to_entry_point`
```
    /// Maps an inner call to its Brillig entry point
    /// This is simply used to simplify fetching global allocations when compiling
    /// individual Brillig functions.
    inner_call_to_entry_point: HashMap<FunctionId, Vec<FunctionId>>,
```
6. After compiling the globals we instantiate a map of entry points to global allocations.
7. The `GlobalInit` label has been updated to be associated with a function id representing an entry point. We now have `GlobalInit(entry_point: FunctionId)`. 

## Additional Context

Most of the line additions are from the large global array included in `test_programs/execution_success/global_var_regression_entry_points/src/consts.nr` file. This change is most meaningful when tested with a large amount of globals across many entry points. 

I compiled `global_var_regression_entry_points` with `nargo compile --force --inliner-aggressiveness -100000000000000`.
Running `NOIR_LOG=trace nargo execute` looking at `nargo::ops::execute: close time.busy` for multiple runs:
On master:
~35ms
This PR:
~7ms

About a 5x improvement in this example.

[...]

PR #7197, opened 2025-01-27
feat(optimization): Add purity analysis to SSA

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/7196

## Summary\*

Adds a purity analysis pass to approximate the purity of each function (we only fail for mutually recursive functions) and use its results in other SSA passes.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
Pending some tests but the pass itself is finished

compiler/noirc_evaluator/src/ssa/opt/pure.rs
```diff
@@ -0,0 +1,269 @@
...
+                                dependencies.insert(*function_id);
+                            }
+                            Value::Intrinsic(intrinsic) => match intrinsic.purity() {
+                                Purity::Pure => (),
+                                Purity::PureWithPredicate => result = Purity::PureWithPredicate,
+                                Purity::Impure => return (Purity::Impure, BTreeSet::new()),
+                            },
+                            // The function we're calling is unknown in the remaining cases,
+                            // so just assume the worst.
+                            Value::ForeignFunction(_)
```
TomAFrench, line 165: Nit but I think it's worth separating out `ForeignFunction` to call out that foreign calls are 100% impure whereas the other cases are unreachable off of the top of my head.

compiler/noirc_evaluator/src/ssa/opt/pure.rs
```diff
@@ -0,0 +1,269 @@
...
+        let function_index = ids_to_indices[&function];
+
+        for dependency in dependencies {
+            let dependency_index = ids_to_indices[&dependency];
+            graph.add_edge(function_index, dependency_index, ());
+        }
+    }
+
+    (graph, ids_to_indices, indices_to_ids)
+}
```
TomAFrench, line 269: We have some code in the inlining pass which builds up a callgraph so there's potential for code reuse here.

jfecher:
Moved the purity pass, and added another closer to the end of SSA.
I've found the results can change as different calls get optimized out so I opted to move the call closer to:
1. The loop invariant code motion pass
2. The constant folding pass

compiler/noirc_evaluator/src/ssa/opt/pure.rs
jfecher: Done

compiler/noirc_evaluator/src/ssa/opt/pure.rs
jfecher: I see compute_bottom_up_order but I'll leave them as separate for now since there's enough logic there having to do with weights that I think they are reasonably different. We probably should have a pass agnostic function to create a call graph eventually though. Perhaps this one can be reorganized in a later refactor.

Another thing I ran into was that I tried to account for mutually recursive functions in my first approach but would have needed a call graph which grouped SCCs into the same node which proved difficult. Having a call graph variant that did that (more expensive) as well would be nice.

jfecher:
This PR still doesn't seem to be deduplicating
```
fn main(x: Field) {
    unsafe {
        let x1 = foo(x);
        let x2 = foo(x);
        assert_eq(x1, 0);
        assert_eq(x2, 0);
    }
}

unconstrained fn foo(x: Field) -> Field {
    x
}
```

TomAFrench:
We don't currently do any deduplication of brillig calls so I think we should add a test to ensure that we don't end up deduplicating a brillig call such that an incorrect predicate is applied.

```nr
fn main(x: Field, active: bool) {
    let mut x1 = false;
    unsafe {
        if active {
            x1 = foo(x);        
        }
    
        let x2 = foo(x);
        assert_eq(x2);
    }
}

unconstrained fn foo(x: Field) -> bool {
    true
}
```

jfecher:
I've also found the reason we weren't deduplicating was that we've been introducing different ValueIds for the same function:
```
  b0(v0: Field):
    v2 = call (v1) f1(v0) -> Field
    v4 = call (v3) f1(v0) -> Field
    constrain v2 == Field 0
    constrain v4 == Field 0
    return
}
```

compiler/noirc_evaluator/src/ssa/opt/pure.rs
vezenovm: I'm also building a call graph in https://github.com/noir-lang/noir/pull/7188, but with slightly different data than in the inlining pass and this pass. I think it would be good to keep each PR as they are and then refactor later to create a pass agnostic call graph.

jfecher:
Had to rewrite a good chunk of the defunctionalize pass because it depended on ValueIds for Value::Function _not_ being unique so that it could map value ids for higher order functions while keeping value ids that were called directly. PR should be ready for review now though it's a bit heftier.

vezenovm:
> Looks like (I'm guessing) my defunctionalize changes come with a hefty compile-time setback though I've yet to test what specifically it is from.

Yeah looking at the size of the regressions it would be good to discern where exactly it is coming form. Perhaps we should pull out the defunctionalization rework into a separate PR.

jfecher:
I think the performance alert here may be bugged. I tested sha256_regression locally and found that this PR compiles  it slightly faster, not several times slower. I also opened https://github.com/noir-lang/noir/pull/7227 to run the CI on a fresh PR and it shows no performance regression comment there.

TomAFrench:
I think it's basically correct but not the best at flagging that a regression has been resolved.

jfecher:
Since there are no perf regressions this is ready to review again at least

TomAFrench (approved):
LGTM. Only followup improvement (which I expect won't catch much that this PR doesn't catch already)

compiler/noirc_evaluator/src/ssa/opt/pure.rs
```diff
@@ -0,0 +1,367 @@
...
+        // from called functions to their callers. Resultingly, an initial "Pure"
+        // result here could be overridden by one of these dependencies being impure.
+        let mut dependencies = BTreeSet::new();
+
+        let mut result = if self.runtime().is_acir() {
+            Purity::Pure
+        } else {
+            // Because we return bogus values when a brillig function is called from acir
+            // in a disabled predicate, brillig functions can never be truly pure unfortunately.
+            Purity::PureWithPredicate
```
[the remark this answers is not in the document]
jfecher: Made an issue here: https://github.com/noir-lang/noir/issues/7229

PR #7199, opened 2025-01-27
feat: array concat method

iAmMichaelConnor:
# Description

Add a `concat` method to the stdlib for the array type.

Where are the array tests? Where should I put a test for this method? Edit: I've put the test in the same file (despite not seeing tests for the other array methods there)

## Documentation\*

Check one:
- [ ] No documentation needed.
- [x] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #7200, opened 2025-01-27
feat: concat3

iAmMichaelConnor:
# Description

This builds off #7199 (hence the diff also including the `concat` method. I haven't learned graphite.

I might've gone a bit far with this one: an ugly method called `concat3` which concatenates 3 arrays. The question then becomes: where do you stop? Where does the madness end? 4? 5? Maybe extra ones should be created through some metaprogramming magic?
Anyway, I would put concat3 to good use immediately, but I could also just use `concat`. If you think that instead I should just use `concat` and accept the slowdown due to the unnecessary `for` loops when doing `arr1.concat(arr2.concat(arr3))`, I guess I could...

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -60,6 +60,7 @@
         "combinators",
         "compinit",
         "comptime",
+        "concat",
         "cpus",
         "cranelift",
         "critesjosh",
--- a/docs/docs/noir/concepts/data_types/arrays.md
+++ b/docs/docs/noir/concepts/data_types/arrays.md
@@ -255,6 +255,23 @@ fn main() {
 }
 ```
 
+### concat
+
+Concatenates this array with another array.
+
+```rust
+fn concat<let M: u32>(self, array2: [T; M]) -> [T; N + M]
+```
+
+```rust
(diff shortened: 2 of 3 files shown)


jfecher:
> The question then becomes: where do you stop? Where does the madness end? 4? 5? Maybe extra ones should be created through some metaprogramming magic?

Right, my initial reaction would be that concat3 seems like it'd be something that may be better fit for a library.

iAmMichaelConnor:
Fair enough. I'll close this one, and leave the `concat` one open

TomAFrench:
To be honest I don't see any reason why the naive method of two calls to the concat function should be less efficient than concat3.

Currently in ACIR, we optimise out the intermediate array. This may not be the case currently but as long as the compiler realizes that the intermediate array is not used then it should be the same.

jfecher:
@TomAFrench I'd expect it to be roughly the same at least in acir but I would expect concat3 to be faster in brillig

PR #7205, opened 2025-01-28
feat: simplify simple conditionals for brillig

### LINKED ISSUE #6394: Flatten simple `if-else` statements in brillig to avoid unnecessary jump statements
Consider the program

```noir
unconstrained fn main(active: bool) -> pub Field {
    if active { 2 } else { 0 }
}
```

This compiles to the SSA

```
brillig(inline) fn main f0 {
  b0(v0: u1):
    jmpif v0 then: b2, else: b1
  b2():
    jmp b3(Field 2)
  b3(v1: Field):
    return v1
  b1():
    jmp b3(Field 0)
}
```

This is equivalent to the program

```noir
unconstrained fn main(active: bool) -> pub Field {
    2 * (active as Field)
}
```

which compiles down to the program

```
brillig(inline) fn main f0 {
  b0(v0: u1):
    v1 = cast v0 as Field
    v3 = mul Field 2, v1
    return v3
}
```

This is much smaller in terms of bytecode size and more amenable to further optimizations. We should then attempt to find simple if-else blocks similar to this in SSA and flatten them even in the case of brillig functions (we'll likely want to apply the changes made in https://github.com/noir-lang/noir/pull/6073 as part of this).

This may seem a bit of a contrived example but we make heavy use of small if statements such as these in the sha256 implementation such that these jumpifs make a non-trivial portion of the overall program.

guipublic:
# Description

## Problem\*

Resolves #6394 

## Summary\*
Add a pass which simplify simple if statements in brillig functions

## Additional Context
The PR is working fine now, the memory consumption alert does not seem to be up-to-date.
The report from the last commit is correct: https://github.com/noir-lang/noir/actions/runs/13119725684/artifacts/2528887358

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -853,8 +853,64 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
...
+                            else_address,
+                            result.extract_single_addr(),
+                        );
+                    }
+                    (
+                        BrilligVariable::BrilligArray(then_array),
+                        BrilligVariable::BrilligArray(else_array),
+                    ) => {
+                        // Pointer to the array which result from the if-else
+                        let pointer = self.brillig_context.allocate_register();
```
vezenovm, line 883: Do we need to allocate another register here? Can we just use the result from above and `result.extract_register()`?

compiler/noirc_evaluator/src/ssa/ir/function_inserter.rs
```diff
@@ -79,6 +79,12 @@ impl<'f> FunctionInserter<'f> {
         (instruction, self.function.dfg.get_instruction_call_stack_id(id))
     }
 
+    /// Get an instruction, map all its values, and replace it with the resolved instruction.
+    pub(crate) fn map_instruction_in_place(&mut self, id: InstructionId) {
+        let mut instruction = self.function.dfg[id].clone();
+        instruction.map_values_mut(|id| self.resolve(id));
+        self.function.dfg.set_instruction(id, instruction);
+    }
```
vezenovm, line 87:
```suggestion
    }
    
```
nit just to separate the `}` from the next comment

compiler/noirc_evaluator/src/ssa/ir/function_inserter.rs
```diff
@@ -251,4 +257,22 @@ impl<'f> FunctionInserter<'f> {
...
     }
+
+    /// Merge the internal mapping into the given mapping
+    /// The merge is guaranteed to be coherent because ambiguous cases are prevented
+    pub(crate) fn extract_mapping(&self, mapping: &mut HashMap<ValueId, ValueId>) {
+        for (k, v) in &self.values {
+            if mapping.contains_key(k) {
+                unreachable!("cannot merge key");
+            }
+            if mapping.contains_key(v) {
```
vezenovm, line 268: We want to check `contains_key` for the value here?

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,379 @@
...
+                    BinaryOp::Add { unchecked }
+                    | BinaryOp::Sub { unchecked }
+                    | BinaryOp::Mul { unchecked } => if unchecked { 1 } else { return u32::MAX },
+                    BinaryOp::Div
+                    | BinaryOp::Mod => return u32::MAX,
+                    BinaryOp::Eq => 2,
+                    BinaryOp::Lt => 5,
+                    BinaryOp::And
+                    | BinaryOp::Or
+                    | BinaryOp::Xor => 1,   //todo
```
vezenovm, line 176: This TODO doesn't have a description and there are a couple TODOs in this function. Could you make issues to handle them if we are not going to in this PR?

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,379 @@
...
+            | Instruction::Call { .. }  //TODO support calls with no-predicate set to true
+            |      Instruction::Load { .. }
+            | Instruction::Store { .. }
+            | Instruction::ArraySet { .. } => return u32::MAX,
+
+            Instruction::ArrayGet { array, index  } => {
+                // check if index is in bound
+                let index = dfg.get_numeric_constant(*index);
+                let mut ok_bound = false;
+                if let Some(index) = index {
```
vezenovm, line 197:
```suggestion
                let mut ok_bound = false;
                if let (Some(index), Some(len)) = (dfg.get_numeric_constant(*index), dfg.try_get_array_length(*array)) {
```
I think something like this would be more ergonomic and reduces nesting

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,379 @@
...
+                    }
+                }
+                if ok_bound {
+                    1
+                } else {
+                    return u32::MAX;
+                }
+            },
+
+            Instruction::Allocate => 0,
```
vezenovm, line 210: nit: we can combine these zero cases into one match statement

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,379 @@
...
+fn block_cost(block: BasicBlockId, dfg: &DataFlowGraph) -> u32 {
+    let mut cost: u32 = 0;
+
+    for instruction in dfg[block].instructions() {
+        let instruction_cost = match &dfg[*instruction] {
+            Instruction::Binary(binary) => {
+                match binary.operator {
+                    BinaryOp::Add { unchecked }
+                    | BinaryOp::Sub { unchecked }
+                    | BinaryOp::Mul { unchecked } => if unchecked { 1 } else { return u32::MAX },
```
vezenovm, line 169: Probably want a TODO for checked operations right? e.g. the extra instructions from an `Add` overflow check are known. Is the cost we are estimating purely number of instructions?

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,379 @@
+use std::collections::{HashSet, VecDeque};
+
+use acvm::AcirField;
+use fxhash::FxHashMap;
```
vezenovm, line 4:
```suggestion
use fxhash::FxHashMap as HashMap;
```
nit: we do this throughout the codebase, would be good to do here for consistency

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,379 @@
...
+        };
+        cost += instruction_cost;
+    }
+    cost
+}
+
+/// Identifies all simple conditionals in the function and flattens them
+fn flatten_function(function: &mut Function, no_predicates: &mut FxHashMap<FunctionId, bool>) {
+    // This pass is dedicated to brillig functions
+    if !matches!(function.runtime(), RuntimeType::Brillig(_)) {
```
vezenovm, line 226:
```suggestion
    if !function.runtime().is_brillig() {
```

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,379 @@
...
+            branch_ends.insert(conditional.block_entry, conditional.block_exit);
+            let mut context = Context::new(function, cfg, branch_ends, cfg_root);
+            context.flatten_single_conditional(conditional, no_predicates);
+            // extract the mapping into 'mapping
+            context.inserter.extract_mapping(&mut mapping);
+        }
+        // 2. re-map the full program for values that may been simplified.
+        if !mapping.is_empty() {
+            let po = post_order.as_slice();
+            for block in po {
```
vezenovm, line 374:
```suggestion
            for block in post_order.as_slice() {
```

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,379 @@
...
+
+// Returns the blocks of the simple conditional sub-graph whose input block is the entry.
+// Returns None if the input block is not the entry block of a simple conditional.
+fn is_conditional(
+    block: BasicBlockId,
+    cfg: &ControlFlowGraph,
+    function: &Function,
+) -> Option<BasicConditional> {
+    // jump overhead is the cost for doing the conditional and jump around the blocks
+    // I use 10 as a rough estimate, real cost is less.
```
vezenovm, line 53:
```suggestion
    // We use 10 as a rough estimate, the real cost is less.
```

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,379 @@
...
+    let jump_overhead = 10;
+    let mut successors = cfg.successors(block);
+    let mut result = None;
+    // a conditional must have 2 branches
+    if successors.len() == 2 {
```
vezenovm, line 58: Let's just early return with `None` if `successors.len() != 2` to reduce this nesting

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,379 @@
...
+            let cost = block_cost(left, &function.dfg);
+            if cost < cost / 2 + jump_overhead {
+                if let Some(TerminatorInstruction::JmpIf {
+                    condition: _,
+                    then_destination,
+                    else_destination,
+                    call_stack: _,
+                }) = function.dfg[block].terminator()
+                {
+                    if left == *then_destination {
```
vezenovm, line 106: I think this would be more readable if we did something like:
```rust
let (block_then, block_else) = if left == *then_destination {
   (Some(left), None)
} else if left == *else_destination {
   (None, Some(left))
};
result = Some(BasicConditional {
       block_entry: block,
       block_then,
       block_else,
       block_exit: right,
});
```

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,379 @@
...
+            let cost = block_cost(right, &function.dfg);
+            if cost < cost / 2 + jump_overhead {
+                if let Some(TerminatorInstruction::JmpIf {
+                    condition: _,
+                    then_destination,
+                    else_destination,
+                    call_stack: _,
+                }) = function.dfg[block].terminator()
+                {
+                    if right == *else_destination {
```
vezenovm, line 135: Same thing as above for the `else if left_successors_len == 1 && next_left == Some(right) {` case

vezenovm (commented):
Did an initial scan. Do you know why we are regressing in `brillig_conditional`?

TomAFrench:
>  Do you know why we are regressing in brillig_conditional?

Agreed that it would be good to have more clarity on this. We've got some serious benefits when conditionally mutating large structs but there's quite a few regressions.

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Happy to address these in this PR. The point is to provide realistic brillig cost for each opcode. The idea is to improve execution speed so the cost should be the runtime cost in whatever unit as along as it is the same for all opcodes. My numbers can certainly be improved.
I will add a warning instead of a todo at the beginning of the function to indicate that the numbers are estimates and can be improved.

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: checked operations can fail, so they are excluded from the simplification (u32::MAX)

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/ir/function_inserter.rs
guipublic: No, if the two mapping both map to the same value, I think that's fine.
However, if the value of one is the key of the other, then one mapping is somehow overwriting what the other mapping is doing, and I disallow this. Especially because mappings are constructed in reverse order, so a mapping overwriting the 'next one' is probably bad.

guipublic:
> > Do you know why we are regressing in brillig_conditional?
> 
> Agreed that it would be good to have more clarity on this. We've got some serious benefits when conditionally mutating large structs but there's quite a few regressions.

I am not sure where these numbers come exactly, doing nargo info on brillig_conditional, I get 31 opcodes with the simplification VS 32 without:  brillig_conditional | conditional      | N/A                  | N/A          | 32              |
WIth binary_operator_overloading, I get the same number, with or without.

TomAFrench:
The regression is in the number of opcodes executed (i.e. the execution trace length). This can be shown by `nargo info --profile-execution`.

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
guipublic: Pointing to the array (in arguments) directly is very tempting, but I don't believe it is safe. The semantic of the operation is a copy, isn't it?

guipublic:
> The regression is in the number of opcodes executed (i.e. the execution trace length). This can be shown by `nargo info --profile-execution`.

Oh right, then this is expected, since the optimisation execute both branches, it is expected to get more opcode executed, it's just that we avoid jumping around. For a number of cases, the numbers are better because of handling of arrays

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,380 @@
...
+                    return u32::MAX;
+                }
+                1
+            },
+
+            Instruction::Allocate
+            | Instruction::EnableSideEffectsIf { .. }
+            | Instruction::IncrementRc { .. }
+            | Instruction::DecrementRc { .. }
+            | Instruction::MakeArray { .. }
```
vezenovm, line 216: We want to get a more accurate instruction count here. MakeArray is either several store and add instructions or a runtime loop for large non-nested arrays https://github.com/noir-lang/noir/blob/32f05f4e42ce25ad48d53274be682af104ef0104/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs#L1714.

guipublic: I put 10 as the cost then, since it seems to be the maximum cost

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
vezenovm: Discussed in person, we can copy for now to be safe.

compiler/noirc_evaluator/src/brillig/brillig_ir/debug_show.rs
```diff
@@ -126,6 +126,24 @@ impl DebugShow {
...
+        source_b: MemoryAddress,
+        condition: MemoryAddress,
+    ) {
+        debug_println!(
+            self.enable_debug_trace,
+            "  {} = CONDITIONAL MOV {} then {}, else {}",
+            destination,
+            condition,
+            source_a,
+            source_b
```
aakoshh, line 143:
```suggestion
        source_then: MemoryAddress,
        source_else: MemoryAddress,
        condition: MemoryAddress,
    ) {
        debug_println!(
            self.enable_debug_trace,
            "  {} = CONDITIONAL MOV {} then {}, else {}",
            destination,
            condition,
            source_then,
            source_else
```
I'd consider putting an `if` in there as well. Perhaps `MOV_IF` like `JUMP_IF`?

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -213,6 +216,10 @@ struct Context<'f> {
     /// us from unnecessarily inserting extra instructions, and keeps ids unique which
     /// helps simplifications.
     not_instructions: HashMap<ValueId, ValueId>,
+
+    /// Flag to tell the context to not issue 'enable_side_effect' instructions during flattening.
+    /// This should set to true only by flatten_single(), when none instruction is known to fail.
```
aakoshh, line 221:
```suggestion
    /// This should be set to true only by `flatten_single()`, when no instruction is known to fail.
```

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,522 @@
...
+struct BasicConditional {
+    block_entry: BasicBlockId,
+    block_then: Option<BasicBlockId>,
+    block_else: Option<BasicBlockId>,
+    block_exit: BasicBlockId,
+}
+
+impl Ssa {
+    #[tracing::instrument(level = "trace", skip(self))]
+    pub(crate) fn flatten_basic_conditionals(mut self) -> Ssa {
```
aakoshh, line 32: It would be helpful to add a doctoring with a basic example of what this pass achieves. For example when we walk someone through the passes, it's a quick way to explain what we mean without having to look up unit tests. Maybe examples are not realistic in more complex cases, and having them here but not there is just irregular, but at sentence or two of what it's looking for and what tradeoffs it's considering would be great IMO.

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,522 @@
...
+        }
+        for function in self.functions.values_mut() {
+            flatten_function(function, &mut no_predicates);
+        }
+        self
+    }
+}
+
+// Returns the blocks of the simple conditional sub-graph whose input block is the entry.
+// Returns None if the input block is not the entry block of a simple conditional.
```
aakoshh, line 46:
```suggestion
/// Returns the blocks of the simple conditional sub-graph whose input block is the entry.
/// Returns None if the input block is not the entry block of a simple conditional.
```

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,522 @@
...
+                result = Some(BasicConditional {
+                    block_entry: block,
+                    block_then,
+                    block_else,
+                    block_exit: right,
+                });
+            }
+        }
+    } else if right_successors_len == 1 && next_right == Some(left) {
+        // Right branch joins the right branch, it is a if/else statement with no then
```
aakoshh, line 125:
```suggestion
        // Right branch joins the left branch, it is a if/else statement with no then
```

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,522 @@
...
+                condition: _,
+                then_destination,
+                else_destination,
+                call_stack: _,
+            }) = function.dfg[block].terminator()
+            {
+                let (block_then, block_else) = if left == *then_destination {
+                    (Some(left), None)
+                } else if left == *else_destination {
+                    (None, Some(left))
```
aakoshh, line 111: It's a bit confusing that the comment above says "it is a if/then statement with no else" but then here we can have an else and no then. (I'm not sure what that means tbh).

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,522 @@
...
+    fn flatten_single_conditional(
+        &mut self,
+        conditional: &BasicConditional,
+        no_predicates: &mut HashMap<FunctionId, bool>,
+    ) {
+        // Manually inline 'then', 'else' and 'exit' into the entry block
+        //0. initialize the context for flattening a 'single conditional'
+        let mut queue = vec![];
+        self.target_block = conditional.block_entry;
+        self.no_predicate = true;
```
aakoshh, line 273: I don't see this being reset anywhere. Ostensibly the `Context` could outlive the call to `flatten_single_conditional`, in which case this should go back to where it was before the call, no?

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,522 @@
...
+        } else {
+            assert_eq!(next, Some(conditional.block_exit));
+        }
+
+        //3. process 'exit' block
+        self.inline_block(conditional.block_exit, no_predicates);
+        // Manually set the terminator of the entry block to the one of the exit block
+        let terminator =
+            self.inserter.function.dfg[conditional.block_exit].terminator().unwrap().clone();
+        let mut next_blocks = VecDeque::new();
```
aakoshh, line 310: I may be blind, but where is this consumed?

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,522 @@
...
+
+        //2. process 'else' branch, in case there is no 'then'
+        let next = queue.pop();
+        if next == conditional.block_else {
+            let next = next.unwrap();
+            self.inline_block(next, no_predicates);
+            let to_process = self.handle_terminator(next, &queue);
+            for incoming_block in to_process {
+                if !queue.contains(&incoming_block) {
+                    queue.push(incoming_block);
```
aakoshh, line 298: I don't see see these values being popped: the 3rd section handles the exit block, and there is no looping 👀

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,522 @@
...
+    if left_successors_len == 1 && right_successors_len == 1 && next_left == next_right {
+        // The branches join on one block so it is a non-nested conditional
+        let cost_left = block_cost(left, &function.dfg);
+        let cost_right = block_cost(right, &function.dfg);
+        // For the flattening to be valuable, we compare the cost of the flattened code with the average cost of the 2 branches,
+        // including an overhead to take into account the jumps between the blocks.
+        let cost = cost_right.saturating_add(cost_left);
+        if cost < cost / 2 + jump_overhead {
+            if let Some(TerminatorInstruction::JmpIf {
+                condition: _,
```
aakoshh, line 82: What happens to the `condition`? Based on the ticket description I thought it would be come a multiplying factor in the merging of the two blocks somehow.

aakoshh: I see from the tests that the multiplication indeed happens, I guess somehow facilitated by the `Context`. I suppose one has to be super familiar with the `flatten_cfg` to understand where the magic happens.

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: It is just to be safe and handle all the sub-cases.
When there is no 'else' or no 'then' we have 3 blocks: 
exit <- entry -> branch->exit
Now several cases can lead to this structure, whether which arrow is left or right...
I guess that the language cannot generate all cases (for instance 'else' without a 'then'), but may be after some optimisations we could have weird cases, so it is better to handle all of them.
What matters it to identify on which side the 'branch' is relative to the JUMP_IF instruction (then or else).

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Done

compiler/noirc_evaluator/src/brillig/brillig_ir/debug_show.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: That's precisely the point, I use the same context for multiple conditional, but I added a few fields that need to be reset for each conditional, which is done here.

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Good point, this part is copy/paste from the flatten_cfg pass but slightly adapted, hence the comment 'manually inline'. The 'next blocks to process' are not needed here because there is no nested if nor loops.

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Same as the previous conversation, I have removed this code that is indeed not needed here.

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
aakoshh: > I use the same context for multiple conditional

I see `Context` it used like this:
```
    for conditional in conditionals {
            ...
            let mut context = Context::new(function, cfg, branch_ends, cfg_root);
            context.flatten_single_conditional(conditional, no_predicates);
            // extract the mapping into 'mapping
            context.inserter.extract_mapping(&mut mapping);
            ...
     }
```
Seemingly it only handles a single conditional, although it might recursively handle more. 

What I was trying to say is that the type is a `flatten_cfg::Context`, and you are adding more methods to it; ostensibly it could happen like this:
```
let mut context = Context::new(...);
context.foo(...); // no_predicates is false
context.flatten_single_conditional(...); // sets no_predicates to true
context.foo(...); // no_predicates is still true
```
By _reset_ I meant that perhaps you should restore the original value of `self.no_predicates` at the end of the call to `flatten_single_conditional`, so that subsequent calls to other methods on the `Context` are not affected.

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
aakoshh: Right, I didn't know it was based on an existing pass, but then can we remove it from here?

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: I agree this is confusing, rather than taking the bet with the unreachable, I updated the comment.

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Thank you for the clarification, indeed I did comment explicitely: _using a new context per conditional_ ...
I added the code to restore the context.

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: I think I forgot to push the commit, should be ok now.

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
vezenovm: I think we can differentiate between side-effectual instructions and instructions that just rely on a predicate like we do in other passes (e.g. constant folding). But this can come later, for now we can just mark a max cost. We should track wanting to re-assess this block cost estimation in an issue.

vezenovm (commented):
~~You had added some unit tests in this commit https://github.com/noir-lang/noir/pull/7205/commits/f5b4076765299cb78ff59082650e5806426ebd1c. I don't see those tests anymore? Looks like they may have been reverted.~~

EDIT: I was looking at an out of date diff ignore my comments about seemingly reverted code.

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,526 @@
...
+                    block_entry: block,
+                    block_then: Some(*then_destination),
+                    block_else: Some(*else_destination),
+                    block_exit: next_left.unwrap(),
+                });
+            }
+        }
+    } else if left_successors_len == 1 && next_left == Some(right) {
+        // Left branch joins the right branch, e.g if/then statement with no else
+        // I am not sure whether this case can happen, but it is not difficult to handle it
```
vezenovm, line 105: nit: I prefer not have first person language like "I am not sure". Let's change to something like "We may never code gen this case, but it is..." or just remove the line entirely

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -213,6 +216,10 @@ struct Context<'f> {
     /// us from unnecessarily inserting extra instructions, and keeps ids unique which
     /// helps simplifications.
     not_instructions: HashMap<ValueId, ValueId>,
+
+    /// Flag to tell the context to not issue 'enable_side_effect' instructions during flattening.
+    /// This should be set to true only by flatten_single(), when none instruction is known to fail.
```
vezenovm, line 221:
```suggestion
    /// This should be set to true only by flatten_single(), when no instruction is known to fail.
```
?

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,526 @@
...
+        let mut inserter = FunctionInserter::new(func);
+        inserter.set_mapping(mapping);
+        let instructions = inserter.function.dfg[block].instructions().to_vec();
+        for instruction in instructions {
+            inserter.map_instruction_in_place(instruction);
+        }
+        inserter.map_terminator_in_place(block);
+    }
+
+    fn flatten_multiple(
```
vezenovm, line 360: nit: I think this would be cleaner outside of the `Context` impl, but I don't feel super strongly about this.

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,526 @@
...
+            }
+        }
+    }
+    // A conditional exit would have exactly 2 predecessors
+    result.filter(|result| cfg.predecessors(result.block_exit).len() == 2)
+}
+
+/// Computes a cost estimate of a basic block
+/// returns u32::MAX if the block has side-effect instructions
+/// WARNING: the estimates are estimate of the runtime cost of each instructions,
```
vezenovm, line 165:
```suggestion
/// WARNING: these are estimates of the runtime cost of each instruction,
```

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
```diff
@@ -0,0 +1,526 @@
...
+/// 1 being the cost of the simplest instruction. These numbers can be improved.
```
vezenovm, line 166: Let's make an issue for tracking purposes

vezenovm (approved):
I have some nits, but they are all non-blocking so I am approving. Let's just make sure to capture an issue for improving the instruction cost estimates.

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs
guipublic: Done here: #7543

PR #7209, opened 2025-01-28
chore(ci): Add test for global vars entry points regression

vezenovm:
# Description

## Problem\*

Test for Brillig program with multiple entry points and various usage of globals in CI. Should see the affects in https://github.com/noir-lang/noir/pull/7188 and afterwards it will act as a regression test.

## Summary\*

I made a program with a few Brillig entry points.

1. Entry point that uses a lot of globals
2. Entry point that uses a single global constants
3. Entry point with no globals
4. Entry point that uses a lot of globals but only in a nested call

`global_var_regression_entry_points` has now been added to the CI reports.

## Additional Context

I decided to keep both `regression_4709` and `global_var_regression_entry_points` as they test different things. I have the upper loop bound in `global_var_regression_entry_points` as we want to test globals initialization rather than the regression from keeping `inc_rc` instructions during loop unrolling. 

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
Do we want to merge this before https://github.com/noir-lang/noir/pull/7188 as to see the affects more clearly?

TomAFrench (approved):

PR #7216, opened 2025-01-28
feat(brillig): Hoist shared constants across functions to the global space

### LINKED ISSUE #7109: Hoist constants shared across functions into the global space

vezenovm:
# Description

## Problem\*

Resolves #7109 

## Summary\*

Before compiling Brillig globals we run an analysis to see which constants are shared across multiple functions. Those values are allocated along with the user-defined globals and then special cased when compiling regular functions. Any repeat constants across functions should be a part of the global read-only memory space and will live throughout the program.

1. When building our `BrilligGlobals` context structure we now maintain a new map:
```rust
/// Maps entry point to constants 
hoisted_global_constants: HashMap<FunctionId, ConstantCounterMap>,
/// Mapping of a constant value and the number of functions in which it occurs
pub(crate) type ConstantCounterMap = HashMap<(FieldElement, NumericType), usize>;
```
2. NB: I re-computed `ConstantAllocation` to determine which constant appears in both functions. This was technically recomputed as we calculate a `ConstantAllocation` when creating a `FunctionContext`. For now, I decided to recompute it as it has a minor affect on compilation, but just noting  this is a place of optimization in the future.
3. When compiling globals, we fetch the constants which appear in more than one function. 
4. After compiling the user specified globals, we compile the hoisted global constants.
5. Maintain a separate map for entry points to their hoisted globals:
```rust
/// Final map that associates an entry point with any local function constants
/// that are shared and were hoisted to the global space.
/// This map is kept separate from `entry_point_globals_map` to clearly distinguish
/// the two types of globals.
entry_point_hoisted_globals_map: HashMap<FunctionId, HoistedConstantsToBrilligGlobals>,
pub(crate) type HoistedConstantsToBrilligGlobals =
    HashMap<(FieldElement, NumericType), BrilligVariable>;
```
6. When compiling regular Brillig blocks we gate liveness analysis on checking whether a variable is actually a hoisted constant, not only on `dfg.is_global`.
7. When converting a variable we escape early if it is a hoisted constant. 

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/brillig/mod.rs
```diff
@@ -104,6 +107,7 @@ impl Ssa {
         let mut brillig_globals =
             BrilligGlobals::new(&self.functions, used_globals_map, self.main_id);
 
+        // brillig_globals.fetch_global_constants_to_hoist(&self.functions);
```
vezenovm, line 110:
```suggestion
```

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -94,7 +103,10 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
         &mut self,
         globals: &DataFlowGraph,
         used_globals: &HashSet<ValueId>,
-    ) {
+        hoisted_global_consts: &HashSet<(FieldElement, NumericType)>,
```
vezenovm, line 106:
```suggestion
        hoisted_global_constants: &HashSet<(FieldElement, NumericType)>,
```

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -113,6 +125,17 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
                 }
             }
         }
+
+        for (constant, typ) in hoisted_global_consts {
```
vezenovm, line 129:
```suggestion
        for (constant, typ) in hoisted_global_constants {
```

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block_variables.rs
```diff
@@ -21,7 +21,7 @@ use super::brillig_fn::FunctionContext;
 
 #[derive(Debug, Default)]
 pub(crate) struct BlockVariables {
-    available_variables: HashSet<ValueId>,
+    pub(crate) available_variables: HashSet<ValueId>,
```
vezenovm, line 24:
```suggestion
    available_variables: HashSet<ValueId>,
```

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_globals.rs
```diff
@@ -195,26 +265,43 @@ impl BrilligGlobals {
...
+    HashMap<ValueId, BrilligVariable>,
+    usize,
+    HashMap<(FieldElement, NumericType), BrilligVariable>,
+);
+
 pub(crate) fn convert_ssa_globals(
     enable_debug_trace: bool,
     globals_dfg: &DataFlowGraph,
     used_globals: &HashSet<ValueId>,
+    hoisted_global_consts: &HashSet<(FieldElement, NumericType)>,
```
vezenovm, line 302:
```suggestion
    hoisted_global_constants: &HashSet<(FieldElement, NumericType)>,
```

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_globals.rs
```diff
@@ -231,17 +318,19 @@ pub(crate) fn convert_ssa_globals(
         variables: Default::default(),
         last_uses: HashMap::default(),
         globals: &empty_globals,
+        hoisted_global_constants: &HashMap::default(),
         building_globals: true,
     };
 
-    brillig_block.compile_globals(globals_dfg, used_globals);
+    let hoisted_global_constants =
+        brillig_block.compile_globals(globals_dfg, used_globals, hoisted_global_consts);
```
vezenovm, line 326:
```suggestion
        brillig_block.compile_globals(globals_dfg, used_globals, hoisted_global_constants);
```

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_globals.rs
```diff
@@ -195,26 +265,43 @@ impl BrilligGlobals {
             for map in entry_point_allocations {
                 globals_allocations.extend(map);
             }
+
+            // The global memory space is not shared across entry points.
+            // Thus, if a call can be used by two entry points we should not use the hoisted constant allocations.
+            if entry_points.len() == 1 {
+                let hoisted_allocations = self.entry_point_hoisted_globals_map.get(&entry_points[0]).unwrap_or_else(|| panic!("ICE: Expected hoisted constant allocations to be set for function {brillig_function_id}"));
```
michaeljklein, line 272:
```suggestion
                let hoisted_allocations = self
                    .entry_point_hoisted_globals_map
                    .get(&entry_points[0])
                    .unwrap_or_else(|| panic!("ICE: Expected hoisted constant allocations to be set for function {brillig_function_id}"));
```

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_globals.rs
```diff
@@ -195,26 +265,43 @@ impl BrilligGlobals {
...
+                hoisted_constants_allocations.extend(hoisted_allocations);
+            }
         } else if let Some(globals) = self.entry_point_globals_map.get(&brillig_function_id) {
             // If there is no mapping from an inner call to an entry point, that means `brillig_function_id`
             // is itself an entry point and we can fetch the global allocations directly from `self.entry_point_globals_map`.
-            // vec![globals]
             globals_allocations.extend(globals);
+
+            let hoisted_allocations = self.entry_point_hoisted_globals_map.get(&brillig_function_id).unwrap_or_else(|| panic!("ICE: Expected hoisted constant allocations to be set for function {brillig_function_id}"));
```
michaeljklein, line 280:
```suggestion
            let hoisted_allocations = self
                .entry_point_hoisted_globals_map
                .get(&brillig_function_id)
                .unwrap_or_else(|| panic!("ICE: Expected hoisted constant allocations to be set for function {brillig_function_id}"));
```

michaeljklein (changes requested):

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_globals.rs
vezenovm: This is what fmt does. I could try and split up the `get` and the `unwrap_or_else` if we would like it split on multiple lines.

michaeljklein: Sure. I'm surprised that `fmt` makes a 200+ char line.

test_programs/execution_success/global_var_regression_simple/src/main.nr
```diff
@@ -14,12 +14,15 @@ fn main(x: Field, y: pub Field) {
...
+    assert(x != y);
+
     let mut acc: Field = 0;
     for i in 0..2 {
         for j in 0..2 {
             acc += EXPONENTIATE[i][j];
         }
     }
     assert(!acc.lt(x));
-    assert(x != y);
```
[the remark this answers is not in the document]
vezenovm: The original regression is still tested, this would just test an additional regression from this PR. Moving the assert up just makes sure that any shared constants (e.g. the indices of the loops) have different SSA value ids as the identical functions would otherwise produce the same value ids. We would hoist if we had the same ids as well, I just wanted to make sure we hoisted with different ids. I'll instead make a unit test for this logic and move the assertion back as to avoid copying the entire test.

TomAFrench:
Can you explain a bit more on why we don't place these hoisted constants into the main globals map? We should be able to treat them in exactly the same way.

vezenovm:
> Can you explain a bit more on why we don't place these hoisted constants into the main globals map? We should be able to treat them in exactly the same way.

I decided to do it during Brillig for a couple reasons. 
1. To avoid having to alter the globals graph and for every function and the cloning cost that comes with it. 
2. I also wanted to re-use some of the pre-existing Brillig constant analysis tools. 
3. Doing this during our existing Brillig globals generation we could avoid another SSA pass. 
4. This is a Brillig unique operation. 

Due to the above reasons, I felt we could do this as part of Brillig gen rather than having to alter the pre-existing SSA.

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_globals.rs
```diff
@@ -195,26 +265,43 @@ impl BrilligGlobals {
...
             unreachable!(
                 "ICE: Expected global allocation to be set for function {brillig_function_id}"
             );
         }
-        globals_allocations
+        (globals_allocations, hoisted_constants_allocations)
     }
 }
 
+pub(crate) type BrilligGlobalsArtifact = (
```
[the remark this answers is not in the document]
vezenovm: I added comments as it is only really used directly above in this same file.

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_globals.rs
vezenovm: Yeah I was surprised by it as well. I split up the `get` and `unwrap_or_else`

michaeljklein (approved):
LGTM

TomAFrench (changes requested):
blocking this waiting for #7277

TomAFrench:
Just needs conflicts to be addressed now.

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -57,11 +60,17 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
...
         dfg: &DataFlowGraph,
         globals: &'block HashMap<ValueId, BrilligVariable>,
+        hoisted_global_constants: &'block HoistedConstantsToBrilligGlobals,
     ) {
         let live_in = function_context.liveness.get_live_in(&block_id);
 
         let mut live_in_no_globals = HashSet::default();
         for value in live_in {
+            if let Value::NumericConstant { constant, typ } = dfg[*value] {
+                if hoisted_global_constants.get(&(constant, typ)).is_some() {
```
TomAFrench, line 70:
```suggestion
                if hoisted_global_constants.contains_ke((constant, typ)) {
```

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -95,7 +105,10 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
         &mut self,
         globals: &DataFlowGraph,
         used_globals: &HashSet<ValueId>,
-    ) {
+        hoisted_global_constants: &BTreeSet<(FieldElement, NumericType)>,
+    ) -> HashMap<(FieldElement, NumericType), BrilligVariable> {
+        let mut new_hoisted_constants = HashMap::default();
```
TomAFrench, line 110: Can you move this down to where it's being used?

vezenovm: done

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -114,6 +127,17 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
                 }
             }
         }
+
+        for (constant, typ) in hoisted_global_constants {
+            let constant = *constant;
+            let typ = *typ;
```
TomAFrench, line 132:
```suggestion
        for (constant, typ) in hoisted_global_constants.iter().copied() {
```

TomAFrench (approved):

Issue #7207, opened 2025-01-28
Add regression testing to track that callstacks are correct


TomAFrench:
We've recently had a regression where some callstacks were partially reversed. We currently don't have any tracking for the correctness of these so we should add some tests which compile a program, purposefully trigger a revert and then checks the callstack is correct.

Issue #7195, opened 2025-01-27
Stack trace order in `nargo debug` is broken


smanilov:
### Aim

Use `nargo debug` to inspect the stack trace of a simple program.

### Expected Behavior

Before #7163, `nargo debug` used to work in the following way.

Consider the program:

```noir
fn bar(y: Field) {
    assert(y != 0);
}

fn foo(x: Field) {
    bar(x)
}

fn main(x: Field, y: pub Field) {
    foo(x);
    foo(y);
    assert(x != y);
}
```

When running `nargo debug` on it, and stepping until the `assert(y != 0);` line, the printed trace is the following:

```
> next
At opcode 0:0.25 :: Const { destination: Relative(8), bit_size: Integer(U32), value: 0 }
At /home/aristotelis/code/repos/clone_fast/tmp_sync/test_programs/plonky2_trace/2_function_calls/src/main.nr:10:5
  5    ...
  6        bar(x)
  7    }
  8    
  9    fn main(x: Field, y: pub Field) {
 10 ->     foo(x);
 11        foo(y);
 12        assert(x != y);
 13    }
At /home/aristotelis/code/repos/clone_fast/tmp_sync/test_programs/plonky2_trace/2_function_calls/src/main.nr:6:5
  1    fn bar(y: Field) {
  2        assert(y != 0);
  3    }
  4    
  5    fn foo(x: Field) {
  6 ->     bar(x)
  7    }
  8    
  9    fn main(x: Field, y: pub Field) {
 10        foo(x);
 11    ...
At /home/aristotelis/code/repos/clone_fast/tmp_sync/test_programs/plonky2_trace/2_function_calls/src/main.nr:1:18
  1 -> fn bar(y: Field) {
  2        assert(y != 0);
  3    }
  4    
  5    fn foo(x: Field) {
  6        bar(x)
  7    }
  8    ...
> q
```

Note that the order of the stack frames here is correct. First is `main`, then `foo`, then `bar`. This is chronological. This happens after about 6 steps into the program (six times calling `next` in the debugger CLI).

### Bug

#7163 seems to break `nargo debug` in the following way.

With the same program, `nargo debug` prints the following stack trace after a few `next` commands:

```
> next
At opcode 0:0.25 :: Const { destination: Relative(8), bit_size: Integer(U32), value: 0 }
At /home/aristotelis/code/repos/clone_fast/tmp_sync/test_programs/plonky2_trace/2_function_calls/src/main.nr:10:5
  5    ...
[...]

PR #7233, opened 2025-01-29
chore: regression test #7195

michaeljklein:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/7207

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
We'll also want to make some brillig calls to make sure that we stitch the ACIR and brillig callstacks together properly.

tooling/debugger/src/repl.rs
```diff
@@ -32,6 +32,9 @@ pub struct ReplDebugger<'a, B: BlackBoxFunctionSolver<FieldElement>> {
 
     // Brillig functions referenced from the ACIR circuits above
     unconstrained_functions: &'a [BrilligBytecode<FieldElement>],
+
+    // whether to print the source without highlighting / pretty-printing
+    raw_source_printing: bool,
```
asterite, line 37: It seems this option also removes line numbers, right? I don't know if that's intentional, but it should probably be documented too.

tooling/debugger/tests/debug.rs
```diff
@@ -49,4 +49,122 @@ mod tests {
...
+            // after running "next" indicates that the
+            // debugger has not panicked for this step.
+            dbg_session
+                .send_line("next")
+                .expect("Debugger panicked while attempting to step through program.");
+            dbg_session
+                .exp_string(">")
+                .expect("Failed while waiting for debugger to step through program.");
+
+            if step_number == num_steps {
```
asterite, line 91: Would it work the same if the body of this `if` is put after the for loop?

asterite (approved):
Looks good! I left a couple of questions.

tooling/debugger/src/repl.rs
michaeljklein: Yes, it intentionally removes line numbers. I've updated the comment.

tooling/nargo_cli/src/cli/debug_cmd.rs
```diff
@@ -50,6 +50,10 @@ pub(crate) struct DebugCommand {
     /// Disable vars debug instrumentation (enabled by default)
     #[clap(long)]
     skip_instrumentation: Option<bool>,
+
+    /// Raw string printing of source for testing
+    #[clap(long, hide = true, default_value = "false")]
+    raw_source_printing: Option<bool>,
```
aakoshh, line 56: I didn't think a flag had to be optional if it had a default value 🤔

tooling/debugger/tests/debug.rs
```diff
@@ -49,4 +49,115 @@ mod tests {
...
+            }
+        }
+
+        let lines_expected_to_contain: Vec<&str> = vec![
+            "> next",
+            "    let x = unsafe { baz(x) };",
+            "unconstrained fn baz(x: Field) -> Field {",
+            "> next",
+            "    let x = unsafe { baz(x) };",
+            "unconstrained fn baz(x: Field) -> Field {",
```
TomAFrench, line 105: I think that this ordering of the stacktrace is pretty confusing and backwards from as most word expect. Rather than starting at the most specific location and then showing the callstack reaching this point below it (as is done in the majority of programming languages afaik) we're always starting at the furthest away location.

tooling/debugger/tests/debug.rs
```diff
@@ -49,4 +49,115 @@ mod tests {
...
+            "    foo(x);",
+            "    let y = unsafe { baz(x) };",
+            "unconstrained fn baz(x: Field) -> Field {",
+            "> next",
+            "    foo(x);",
+            "    let y = unsafe { baz(x) };",
+            "}",
+            "> next",
+            "    foo(x);",
+            "    let y = unsafe { baz(x) };",
```
aakoshh, line 131: Just curious: why do we have seem to have everything appear twice?

michaeljklein: It's skipping the highlighting moving from e.g. all of `foo(x)` to just `x`.

tooling/debugger/tests/debug.rs
michaeljklein: I think this might be the difference between the order from printing unwinding a stack after a panic and the order from stepping through execution with a debugger:
- Unwinding: `assert(..)\n..[inside foo]\n..foo()\n..main()`
- Line-by-line debugger: `main()\n..foo()\n..[inside foo]\n..assert(..)`

vezenovm (approved):

PR #7236, opened 2025-01-30
feat(ssa): More aggressive Brillig loop unrolling

### LINKED ISSUE #7099: Loops in brillig aren't being unrolled aggressively enough.
Consider the program below which looks suspiciously similar to [`__compute_partial_sums`](https://github.com/AztecProtocol/aztec-packages/blob/0b2a61901d183ffe6389a6543a379acfef25c5e5/noir-projects/noir-protocol-circuits/crates/blob/src/blob.nr#L405-L450)

```nr
global ROOTS: [Field; 16] = [1, 2, 3, 4, 5, 6, 7,8, 9, 10, 11, 12, 13, 14, 15, 16];

unconstrained fn main(
    fracs: [Field; 16],
) -> pub [Field; 2] {
    let mut partial_sums: [Field; 2] = std::mem::zeroed();

    let mut partial_sum: Field = 0;
    for i in 0..8 {
        let summand = ROOTS[i] * fracs[i];
        partial_sum = partial_sum + summand;
    }
    partial_sums[0] = partial_sum;

    let NUM_PARTIAL_SUMS = 2;
    for i in 1..NUM_PARTIAL_SUMS {
        let mut partial_sum: Field = partial_sums[i - 1];
        for j in 0..8 {
            let k = i * 8 + j;
            let summand = ROOTS[k]* fracs[k];
            partial_sum = partial_sum + summand;
        }
        partial_sums[i] = partial_sum;
    }

    partial_sums
}
```

This should be able to be compiled down to a stack of adds and muls but instead we get

```
g0 = Field 1
g1 = Field 2
g2 = Field 3
g3 = Field 4
g4 = Field 5
g5 = Field 6
g6 = Field 7
g7 = Field 8
g8 = Field 9
g9 = Field 10
g10 = Field 11
g11 = Field 12
g12 = Field 13
g13 = Field 14
g14 = Field 15
g15 = Field 16
g16 = make_array [Field 1, Field 2, Field 3, Field 4, Field 5, Field 6, Field 7, Field 8, Field 9, Field 10, Field 11, Field 12, Field 13, Field 14, Field 15, Field 16] : [Field; 16]

brillig(inline) fn main f0 {
  b0(v17: [Field; 16]):
    v21 = make_array [Field 0, Field 0] : [Field; 2]
    v22 = allocate -> &mut [Field; 2]
    store v21 at v22
    v23 = allocate -> &mut Field
    store Field 0 at v23
    jmp b1(u32 0)
  b1(v18: u32):
    v26 = lt v18, u32 8
    jmpif v26 then: b6, else: b2
  b2():
    v27 = load v22 -> [Field; 2]
    v28 = load v23 -> Field
    v29 = array_set v27, index u32 0, value v28
    store v29 at v22
    v30 = allocate -> &mut Field
    store v28 at v30
    jmp b3(u32 0)
  b3(v19: u32):
    v31 = lt v19, u32 8
    jmpif v31 then: b5, else: b4
  b4():
    v32 = load v22 -> [Field; 2]
    v33 = load v30 -> Field
    v35 = array_set v32, index u32 1, value v33
    store v35 at v22
    return v35
  b5():
    v36 = add u32 8, v19
    v37 = array_get g16, index v36 -> Field
    v38 = array_get v17, index v36 -> Field
    v39 = mul v37, v38
    v40 = load v30 -> Field
    v41 = add v40, v39
    store v41 at v30
    v42 = unchecked_add v19, u32 1
    jmp b3(v42)
  b6():
    v43 = array_get g16, index v18 -> Field
    v44 = array_get v17, index v18 -> Field
    v45 = mul v43, v44
    v46 = load v23 -> Field
    v47 = add v46, v45
    store v47 at v23
    v48 = unchecked_add v18, u32 1
    jmp b1(v48)
}

```

Compare this with the equivalent ACIR, it's marginally more instructions but much faster execution.

```
acir(inline) fn main f0 {
  b0(v17: [Field; 16]):
    v19 = array_get v17, index u32 0 -> Field
    v21 = array_get v17, index u32 1 -> Field
    v23 = mul Field 2, v21
    v24 = add v19, v23
    v26 = array_get v17, index u32 2 -> Field
    v28 = mul Field 3, v26
    v29 = add v24, v28
    v31 = array_get v17, index u32 3 -> Field
    v33 = mul Field 4, v31
    v34 = add v29, v33
    v36 = array_get v17, index u32 4 -> Field
    v38 = mul Field 5, v36
    v39 = add v34, v38
    v41 = array_get v17, index u32 5 -> Field
    v43 = mul Field 6, v41
    v44 = add v39, v43
    v46 = array_get v17, index u32 6 -> Field
    v48 = mul Field 7, v46
    v49 = add v44, v48
    v51 = array_get v17, index u32 7 -> Field
    v53 = mul Field 8, v51
    v54 = add v49, v53
    v56 = array_get v17, index u32 8 -> Field
    v58 = mul Field 9, v56
    v59 = add v54, v58
    v61 = array_get v17, index u32 9 -> Field
    v63 = mul Field 10, v61
    v64 = add v59, v63
    v66 = array_get v17, index u32 10 -> Field
    v68 = mul Field 11, v66
    v69 = add v64, v68
    v71 = array_get v17, index u32 11 -> Field
    v73 = mul Field 12, v71
    v74 = add v69, v73
    v76 = array_get v17, index u32 12 -> Field
    v78 = mul Field 13, v76
    v79 = add v74, v78
    v81 = array_get v17, index u32 13 -> Field
    v83 = mul Field 14, v81
    v84 = add v79, v83
    v86 = array_get v17, index u32 14 -> Field
    v88 = mul Field 15, v86
    v89 = add v84, v88
    v91 = array_get v17, index u32 15 -> Field
    v93 = mul Field 16, v91
    v94 = add v89, v93
    v95 = make_array [v54, v94] : [Field; 2]
    return v95
}

```

vezenovm:
# Description

## Problem\*

Resolves #7099 

## Summary\*

Still a draft, needs lots of cleanup and tests, but I want to see the reports on CI.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Just a small improvement in execution speed then :laughing:

tooling/nargo_cli/build.rs
```diff
@@ -198,8 +198,8 @@ fn test_{test_name}(force_brillig: ForceBrillig, inliner_aggressiveness: Inliner
         nargo.arg("--force-brillig");
 
         // Set the maximum increase so that part of the optimization is exercised (it might fail).
-        nargo.arg("--max-bytecode-increase-percent");
-        nargo.arg("50");
+        // nargo.arg("--max-bytecode-increase-percent");
```
vezenovm, line 201: Noting that as we unroll more loops, this is potentially going to lead to panics. When we have a non-aggressive inliner the compiler is not going to have yet resolved certain intrinsics which we expect to be entirely known at compile-time (e.g. DerivePedersenGenerators).

jfecher:
This'll still be configurable via `--inliner-aggressiveness`, right?
Wonder what our default value we should target for that is. I wonder if the changes here result in too-large bytecode for example.

vezenovm:
> This'll still be configurable via `--inliner-aggressiveness`, right?
> Wonder what our default value we should target for that is. I wonder if the changes here result in too-large bytecode for example.

Do you mean `--max-bytecode-increase-percent`? This flag determines how large of a bytecode size we are willing to tolerate from unrolling Brillig loops. 

Yes this will still be configurable and we will have to experiment a bit with what a good default target should be. Right now we default to this flag not being set, although we do test for it in `nargo_cli/build.rs`. I currently have it commented as it can cause failures (we are attempting to compile to Brillig before certain compile-time intrinsics can be simplified out), and am working on resolving how we can improve this heuristic.

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -54,14 +53,14 @@ impl Ssa {
...
-        max_bytecode_increase_percent: Option<i32>,
+        max_bytecode_increase_percent: i32,
     ) -> Result<Ssa, RuntimeError> {
         for function in self.functions.values_mut() {
             let is_brillig = function.runtime().is_brillig();
 
             // Take a snapshot in case we have to restore it.
             let orig_function =
-                (max_bytecode_increase_percent.is_some() && is_brillig).then(|| function.clone());
+                (max_bytecode_increase_percent == i32::MAX && is_brillig).then(|| function.clone());
```
vezenovm, line 63:
```suggestion
                (max_bytecode_increase_percent < i32::MAX && is_brillig).then(|| function.clone());
```

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -790,7 +798,35 @@ impl Loop {
...
+        // AND every such store has a constant value.
+        //
+        // We must exclude the loop's own blocks from this scan: for nested loops,
+        // `find_blocks_in_loop(entry, pre_header)` traverses backward through the
+        // outer loop's back-edge and re-enters the inner loop blocks. Without this
+        // filter, stores *inside* the loop body (which are not initial values) would
+        // incorrectly prevent the ref from being recognized as constant-initial.
+        let mut has_store: HashSet<ValueId> = HashSet::default();
+        let mut has_non_constant_store: HashSet<ValueId> = HashSet::default();
+        for block in blocks.iter().filter(|b| !self.blocks.contains(b)) {
```
vezenovm, line 948: The root cause of this filter was that `find_blocks_in_loop(entry, pre_header)` traverses backward through the outer loop's back-edge and re-enters the inner loop blocks. This caused stores inside the inner loop body to be mistakenly treated as pre-header stores, preventing the reference from being recognized as constant-initial.
                                                                                                                                                                                                     A dominator-based analysis would avoid this structurally (inner loop blocks don't dominate the pre-header), but the filter is a simpler fix. Worth revisiting if more nested-loop edge cases come up.

jfecher:
> Constant-initial reference propagation: When a pre-header reference (e.g., an accumulator let mut acc = 0) has only constant stores before the loop, the load result is added to constant_after_unroll. This lets downstream instructions (e.g., acc += LOOKUP_TABLE[i]) cascade as useless.

I haven't looked through the PR yet but reading this concerns me a bit that unrolling may accumulate more features over time and become increasingly buggy and difficult to fix because it tries to do so much. In other words, exactly what happened with mem2reg.

vezenovm:
> unrolling may accumulate more features over time and become increasingly buggy and difficult to fix because it tries to do so much

Yeah I understand that fear, I have it too. I am a bit more ok with it in the case of unrolling as the worst thing we could do is unroll a little too aggressively (e.g., we are not structurally changing how we unroll vs. mem2reg where optimizations require additional correctness checks). 

We also already try to attempt to unroll Brillig loops, so this felt like an improvement of existing functionality rather than something entirely new.

jfecher:
From the quoted comment it sounded like we were attempting to add logic dealing with references and catching cases where they're not expected to change. This makes me question if we covered possible aliases of these allocations too for example and we get deeper into duplicating mem2reg logic. Did you also check for the case it is passed to a block terminator for example?

I'd feel safer without such an optimization if my reading of it is right - we can rely on one of the mem2reg's to do it later on. Either mem2reg after unrolling or mem2reg_simple once that is merged will be able to optimize across loops.

vezenovm:
> From the quoted comment it sounded like we were attempting to add logic dealing with references and catching cases where they're not expected to change. This makes me question if we covered possible aliases of these allocations too for example and we get deeper into duplicating mem2reg logic. Did you also check for the case it is passed to a block terminator for example?

I will add the case with block terminators to see how it is affected by this optimization. It does look like this cost estimate is missing to account for terminators. 

However, I still think this PR is fundamentally doing something different than mem2reg. This is purely a cost estimation to determine whether we want to unroll vs. mem2reg which actually modifies elides loads/stores. If this analysis were being used for eliding loads/stores I definitely agree that this would just lead to a duplication of mem2reg logic. 

We are just expanding our existing check of "will unrolling this loop fold away more instructions?". If it is incorrect, the consequence is an unideal unroll decision (we either unroll to aggressively or miss an optimization opportunity). 

This optimization is also quite conservative and should generally underestimate foldable instructions. It only marks a reference as constant-inital if every pre-header store is constant. If there are three constant stores and one non-constant store to a ref, that ref will not be counted as constant-initial. If a pre-header ref comes from a block parameter those loads will not be counted as constant foldable loads. 

In fact, it looks like LLVM does something very similar. They have an [LoopUnrollAnalyzer](https://llvm.org/doxygen/LoopUnrollAnalyzer_8cpp_source.html) which builds a `SimplifiedValues` map. It uses this map to propagate constant values to binary instructions as well as folding loads. 

Also see `analyzeLoopUnrollCost` in the unroll pass itself https://llvm.org/doxygen/LoopUnrollPass_8cpp_source.html. Per the comments on that method:
```
/// Figure out if the loop is worth full unrolling.
///
/// Complete loop unrolling can make some loads constant, and we need to know
/// if that would expose any further optimization opportunities.  This routine
/// estimates this optimization.  It computes cost of unrolled loop
/// (UnrolledCost) and dynamic cost of the original loop (RolledDynamicCost). By
/// dynamic cost we mean that we won't count costs of blocks that are known not
/// to be executed (i.e. if we have a branch in the loop and we know that at the
/// given iteration its condition would be resolved to true, we won't add up the
/// cost of the 'false'-block).
```

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -19,6 +19,9 @@
 //!   - Brillig functions only have small loops unrolled, where a small loop is defined as a loop
 //!     which, when unrolled, is estimated to have the same or fewer total instructions as it
 //!     has when not unrolled.
+//!     This cost estimation is analogous to LLVM's `analyzeLoopUnrollCost` which estimates
+//!     which loads become constant after unrolling. See:
+//!     https://llvm.org/doxygen/LoopUnrollPass_8cpp_source.html
```
vezenovm:
```suggestion
//!     <https://llvm.org/doxygen/LoopUnrollPass_8cpp_source.html>
```

PR #7242, opened 2025-01-30
fix(ssa): Use number of SSA instructions for the Brillig unrolling bytecode size limit

vezenovm:
# Description

## Problem\*

Pulls a change out of https://github.com/noir-lang/noir/pull/7236 that was causing panics with less aggressive inliner settings. Comment for reference: https://github.com/noir-lang/noir/pull/7236#discussion_r1935634486.

I labelled this PR a fix rather than a chore as it does prevent possible panics. 

## Summary\*

After performing an unroll in the Brillig runtime we have a setting that lets us set an upper bound on the bytecode size increase percentage. However, we have certain intrinsics (e.g. DerivePedersenGenerators) that we expect to entirely compile-time and fully resolved by Brillig gen. 

Another possible fix would be to set any method which calls `derive_generators` (such as `pedersen_hash_with_separator`) as `#[inline_always]`. However, we still do not have a guarantee that the `separator` has been resolved to a constant. Let's say we know `derive_generators` is called by `pedersen_hash_with_separator`. Even if  `pedersen_hash_with_separator` is always inlined into its parent we cannot guarantee that the input for the separator comes from the parent's function parameter. This issue will exist with any other intrinsics we expect to be fully resolved by compile-time. 

By using the number of SSA instructions we can also avoid the extra compilation to Brillig as well as having to expose extra Brillig gen logic to the SSA passes. This gives us a better separation of concerns. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/function.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/function.rs
@@ -211,6 +211,16 @@ impl Function {
 
         unreachable!("SSA Function {} has no reachable return instruction!", self.id())
     }
+
+    pub(crate) fn num_instructions(&self) -> usize {
+        self.reachable_blocks()
+            .iter()
+            .map(|block| {
+                let block = &self.dfg[*block];
+                block.instructions().len() + block.terminator().is_some() as usize
+            })
+            .sum()
+    }
 }
 
 impl Clone for Function {
--- a/compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
@@ -24,10 +24,6 @@ use acvm::{acir::AcirField, FieldElement};
 use im::HashSet;
 
 use crate::{
-    brillig::{
-        brillig_gen::{brillig_globals::convert_ssa_globals, convert_ssa_function},
-        brillig_ir::brillig_variable::BrilligVariable,
-    },
     errors::RuntimeError,
     ssa::{
         ir::{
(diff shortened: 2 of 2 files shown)


compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -60,7 +56,7 @@ impl Ssa {
         mut self,
         max_bytecode_increase_percent: Option<i32>,
     ) -> Result<Ssa, RuntimeError> {
-        let mut global_cache = None;
+        // let mut global_cache = None;
```
TomAFrench, line 59:
```suggestion
```

TomAFrench (approved):

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -60,7 +56,6 @@ impl Ssa {
         mut self,
         max_bytecode_increase_percent: Option<i32>,
     ) -> Result<Ssa, RuntimeError> {
-        let mut global_cache = None;
```
TomAFrench, line 59:
```suggestion
```

compiler/noirc_evaluator/src/ssa/ir/function.rs
```diff
@@ -211,6 +211,16 @@ impl Function {
 
         unreachable!("SSA Function {} has no reachable return instruction!", self.id())
     }
+
+    pub(crate) fn num_instructions(&self) -> usize {
+        self.reachable_blocks()
+            .iter()
+            .map(|block| {
+                let block = &self.dfg[*block];
+                block.instructions().len() + block.terminator().is_some() as usize
```
jfecher, line 220: Do we need to count the terminator here? It should always be Some after the Ssa is initially constructed so even if we do want to count it, maybe we should just add a constant 1?

jfecher (approved):
This seems much simpler than going off of brillig instructions

compiler/noirc_evaluator/src/ssa/ir/function.rs
vezenovm: I do prefer this to a magic number `1`

PR #7261, opened 2025-02-03
feat(LSP): suggest enum variants without parameters

asterite:
# Description

## Problem

Now that enum variants without parameters can be referenced without `()`, those variants weren't offered in LSP completion.

## Summary

Let's LSP suggest enum variants without parameters.

Also changes these completions to use the `EnumMember` kind so that they end up with a different icon than the one for functions. Finally, also changes the icon for enum types themselves (previously they would use the icon for structs).

image

image

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/lsp/src/requests/completion.rs
+++ b/tooling/lsp/src/requests/completion.rs
@@ -804,18 +804,14 @@ impl<'a> NodeFinder<'a> {
         };
 
         for (index, variant) in variants.iter().enumerate() {
+            // Variants with parameters are represented as functions and are suggested in `complete_type_methods`
             if variant.is_function || !name_matches(&variant.name.0.contents, prefix) {
                 continue;
             }
 
-            if !variant.is_function {
-                let item = self.enum_variant_completion_item(
-                    variant.name.to_string(),
-                    data_type.id,
-                    index,
-                );
-                self.completion_items.push(item);
-            }
+            let item =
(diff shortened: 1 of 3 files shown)


tooling/lsp/src/requests/completion.rs
```diff
@@ -1900,13 +1914,13 @@ fn get_array_element_type(typ: Type) -> Option<Type> {
     }
 }
 
-fn get_type_struct_id(typ: &Type) -> Option<TypeId> {
+fn get_type_type_id(typ: &Type) -> Option<TypeId> {
```
jfecher, line 1917: Should probably use `.follow_bindings_shallow` here to handle type variables as well. 

You will need to add the Alias case to that function though.

jfecher (approved):
LGTM. Found a case where we should probably be using `follow_bindings_shallow` although we'd need to add the Alias case to that method.

tooling/lsp/src/requests/completion.rs
asterite: Done!

Oh, I was about to say that `Alias` is already handled by this function, though I guess it must be handled by `follow_bindings_shallow`, right? I'm not sure I'd like to make that change in this PR because that's also used in the compiler and maybe it breaks something. I'll create a follow-up PR to try it, though!

PR #7263, opened 2025-02-03
chore: early check type equality in try_unify

### LINKED ISSUE #7256: Example of `compile` hanging forever with no output
### Aim

This PR in Aztec is hanging forever on `compile` and `test`. CI gives up after 40 minutes.

It's interesting, because it's only a ~200 line change vs `master`.

At Maxim's suggestion, I ran `nargo check` and that actually started outputting warnings after 4 minutes. But... it's been 15 minutes and that `nargo check` command is still slowly thinking and hasn't finished yet.

https://github.com/AztecProtocol/aztec-packages/pull/11554

```sh
git checkout mc/address-nico-comments
./bootstrap.sh fast
```

I guess that bootstrap might fail, because of the issue being described here. Regardless, do this afterwards:

```sh
cd noir-projects/aztec-nr
```

Using the version of Noir that aztec-packages uses:

```sh
../../noir/noir-repo/target/release/nargo check
```

```sh
../../noir/noir-repo/target/release/nargo test
```

```sh
../../noir/noir-repo/target/release/nargo compile
```

One thing of note in this PR is a rather complex arithmetic on generics in the changed `note.nr` file. Now, there was already some complex arithmetic on generics in that file in `master`; I just made some changes, and perhaps those changes are confusing the compiler now.

This code did work 3 or 4 days ago on master. I've since merged master and it's broken.

asterite:
# Description

## Problem

Resolves #7256

## Summary

While debugging this I realized that `try_unify` was being called with `self` and `other` being complex infix expressions that were actually the **same** expression (same type variables, etc.). When that happened canonicalizing and trying to unify both would involve a lot of computations (it ends up trying to unify lots of `u32` for a reason that I still didn't sit down to investigate). I tried doing an early equality check in `try_unify` and, magically, the time to compile the relevant program went from 10 minutes to 3 seconds.

## Additional Context

I _think_ checking for equality here is fine because it means there's nothing to unify, but let me know if this is not okay.

The second commit is just a micro-optimization: when doing `if let (_, _) = (a, b)`, `a` and `b` are evaluated and then pattern-matched. Because `evaluate_to_field_element` might be an expensive call this just ends up doing one of those calls if it ends up returning `Err`.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir_def/types.rs
+++ b/compiler/noirc_frontend/src/hir_def/types.rs
@@ -1742,6 +1742,13 @@ impl Type {
     ) -> Result<(), UnificationError> {
         use Type::*;
 
+        // If the two types are exactly the same then they trivially unify.
+        // This check avoids potentially unifying very complex types (usually infix
+        // expressions) when they are the same.
+        if self == other {
+            return Ok(());
+        }
+
         let lhs = self.follow_bindings_shallow();
         let rhs = other.follow_bindings_shallow();
 
--- a/compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
+++ b/compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
@@ -63,12 +63,15 @@ impl Type {
                 let dummy_span = Span::default();
                 // evaluate_to_field_element also calls canonicalize so if we just called
                 // `self.evaluate_to_field_element(..)` we'd get infinite recursion.
-                if let (Ok(lhs_value), Ok(rhs_value)) = (
-                    lhs.evaluate_to_field_element_helper(&kind, dummy_span, run_simplifications),
-                    rhs.evaluate_to_field_element_helper(&kind, dummy_span, run_simplifications),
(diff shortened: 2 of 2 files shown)


TomAFrench:
Awesome, can we create a regression test out of that branch? I'm not sure how much we can just pull out the changes made in that branch vs the full contract.

TomAFrench (approved):

asterite:
I tried. Or, well, at least I tried with a program that had `get_arr_of_size__log_bytes_padding__from_plaintext` and all the necessary functions, called it in a similar way done there... but it ends up being fast. I still see a lot more work is done before this PR in that case, but I couldn't find a program that would take a lot of time. I'll try it in a follow-up PR.

As a side note, I was curious if this generally improved compile times. In Aztec-Packages `master` running `nargo compile --force --skip-brillig-constraints-check --silence-warnings` takes 46 seconds before this PR, and 19 seconds with this PR 🎉

PR #7264, opened 2025-02-03
fix: prevent panic within `remove_possibly_mutated_cached_make_arrays`

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This adds a quick fix for the panic in https://github.com/AztecProtocol/aztec-packages/pull/11696 that @vezenovm raised. 

I (or someone) needs to look at this more closely as we should potentially distinguish between local and global arrays here. For the time being it fixes the panic however.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
@@ -734,7 +734,6 @@ impl<'brillig> Context<'brillig> {
                 _ => return,
             };
 
-
             if matches!(instruction, Instruction::MakeArray { .. }) {
                 self.cached_instruction_results.remove(instruction);
             }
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -725,6 +725,10 @@ impl<'brillig> Context<'brillig> {
 
         // Should we consider calls to slice_push_back and similar to be mutating operations as well?
         if let Store { value: array, .. } | ArraySet { array, .. } = instruction {
+            if function.dfg.is_global(*array) {
```
vezenovm, line 728: Should we add a brief comment just that globals are expected to be immutable?

vezenovm (approved):

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -725,6 +725,10 @@ impl<'brillig> Context<'brillig> {
...
+                return;
```
TomAFrench, line 729:
```suggestion
                // Early return as we expect globals to be immutable.
                return;
```

PR #7265, opened 2025-02-03
fix(ssa): Unused functions removals post folding constant Brillig calls

vezenovm:
# Description

## Problem\*

Silly bug leftover from https://github.com/noir-lang/noir/pull/7260, but also fixes this removal pass to be more accurate. We are not checking the new call graph of unused functions.

e.g. for the following SSA:
```
        g0 = Field 2

        acir(inline) fn main f0 {
          b0():
            v1 = call f1() -> Field
            return v1
        }

        brillig(inline) fn entry_point f1 {
          b0():
            v1 = call f2() -> Field
            return v1
        }

        brillig(inline) fn one f2 {
          b0():
            v1 = add g0, Field 3
            return v1
        }
```
After constant folding we would still have:
```
        brillig(inline) fn one f2 {
          b0():
            v1 = add g0, Field 3
            return v1
        }
```
even though it would be unused. This is because we would only inline Brillig calls for ACIR entry points. Thus, we would still have a call to f2, even though f1 is now unused, thus f2 is also unused.

## Summary\*

I simply moved to using our pre-existing `remove_unreachable_functions` pass. I felt this kept a better separation of concerns for the passes and was a more complete removal of unreachable functions.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa.rs
+++ b/compiler/noirc_evaluator/src/ssa.rs
@@ -136,6 +136,9 @@ pub(crate) fn optimize_into_acir(
         print_codegen_timings: options.print_codegen_timings,
     }
     .run_pass(|ssa| ssa.fold_constants_with_brillig(&brillig), "Inlining Brillig Calls Inlining")
+    // It could happen that we inlined all calls to a given brillig function.
+    // In that case it's unused so we can remove it. This is what we check next.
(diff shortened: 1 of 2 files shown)


asterite (approved):
Makes sense and looks good!

Issue #2128, opened 2023-08-02
Debug compilation for Noir programs


vezenovm:
### Problem

When working with `nargo test`, everything supplied to a function is a constant. This leads to issues as outlined in this PR: https://github.com/noir-lang/noir/pull/2116, where `println` does not work in tests with failing constraints. Essentially a failed constraint is resolved during compilation before we ever reach execution (where the `println` foreign call is handled). 

### Happy Case

We should offer a `debug` compile option that does not enforce constrain instructions at compile-time. This compile option will be enabled during `nargo test` to allow `println` to execute even with compile time constraint failures. 

### Alternatives Considered

Otherwise `println` just will not work during failing tests which leads to inconsistent functionality with `nargo execute`. No realistic alternatives.

### Would you like to submit a PR for this Issue?

No

TomAFrench:
There are a number of other optimisations which we could turn off in order to improve debugging here, e.g. the redundant range optimisation can result in an overflow not being displayed due to the fact that it's being caught later.

kevaundray:
An alternative approach is for tests to make the input be seen as non-constant, so that you can reproduce the exact environment one gets when a program is executed (with optimizations)

vezenovm:
Linking this comment (https://github.com/noir-lang/noir/pull/2446#discussion_r1307701028) to note another optimization that can be removed in a debug compilation mode

TomAFrench:
In the short term (and complementary to debug mode in the future), we could add something similar to Rust's `blackbox` function. This would just be an identity function which is listed as having side-effects, which would block any SSA optimizations from being applied to the outputs.

PR #7277, opened 2025-02-04
fix(brillig): Brillig entry point analysis and function specialization through duplication

vezenovm:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/7306.

The description in this PR repeats a lot of the comments and tests from the new SSA pass, but I have copied things here for visibility. 

Take this program:
```noir
global ONE: Field = 1;
global TWO: Field = 2;
global THREE: Field = 3;
    /// Safety: testing context
    unsafe {
        entry_point_one(x, y);
        entry_point_two(x, y);
    }
}
unconstrained fn entry_point_one(x: Field, y: Field) {
    let z = ONE + x + y;
    assert(z == 2);
    inner_func(x, y);
}
unconstrained fn entry_point_two(x: Field, y: Field) {
    let z = TWO + x + y;
    assert(z == 3);
    inner_func(x, y);
}
unconstrained fn inner_func(x: Field, y: Field) {
    let z = THREE + x + y;
    assert(z == 4);
}
```
We have this SSA before Brillig gen:
```
g0 = Field 1
g1 = Field 2
g2 = Field 3

  b0(v3: Field, v4: Field):
    call f1(v3, v4)
    call f2(v3, v4)
    return
}
brillig(inline) predicate_pure fn entry_point_one f1 {
  b0(v3: Field, v4: Field):
    v5 = add Field 1, v3
    v6 = add v5, v4
    constrain v6 == Field 2
    call f3(v3, v4)
    return
}
brillig(inline) predicate_pure fn entry_point_two f2 {
  b0(v3: Field, v4: Field):
    v5 = add Field 2, v3
    v6 = add v5, v4
    constrain v6 == Field 3
    call f3(v3, v4)
    return
}
brillig(inline) predicate_pure fn inner_func f3 {
  b0(v3: Field, v4: Field):
    v5 = add Field 3, v3
    v6 = add v5, v4
    constrain v6 == Field 4
    return
}
```
For each entry point we will generate the follows artifacts for globals initialization:
```
GlobalInit(Id(1)):
  CONST M32835 = 1
  CONST M32836 = 2
  CONST M32837 = 3
  RETURN
GlobalInit(Id(2)):
  CONST M32835 = 2
  CONST M32836 = 3
  RETURN
```
It is then not clear to `inner_func` which map of SSA value id -> Brillig global initialization we should use. 
Without the hack from #7306 we get the following bytecode for f3:
```
Function(Id(3), Some(Id(0))):
  S3 = M32836 + S1
  S1 = S3 + S2
  CONST S2 = 4
  S3 = S1 == S2
  JUMP_IF S3 TO Function(Id(3), Some(Id(0))) - 1
  CONST S4 = 0
  TRAP FreeMem[0..S4]
  RETURN
```
This works when f3 is called from the f2 entry point. But it will break when called from the f1 entry point. We need to specialize f3 to account for the potentially conflicting global allocation maps. 

## Summary\*

We now specialize functions per entry point. 

A few things were done in this PR:
1. A new pass the specializes functions per entry point
2. I run the new pass following DIE as to reduce the amount of code we may have to duplicate. However, this also required re-running DIE following this pass as the `used_globals` map is set during DIE. Perhaps we do not have to re-run this though as the global value IDs should not change across SSA passes. I could foresee the CFG simplification after DIE leading to some terminators that use globals being removed and needing to recompute the used globals map. Perhaps we could recompute this during CFG simplification. I feel this can be done in a followup PR though. 
3. I have been testing against aztec-packages directly and added a few simplified regression tests to our integration tests as well as some unit tests for the new SSA pass. 

## Additional Context

This works further re-iterates the need for a debug/release mode as laid out in https://github.com/noir-lang/noir/issues/2128. This duplication can lead to code bloat. During debug mode we could potentially not perform any entry point analysis and just have the same globals across all functions if this work is shown to seriously degrade compilation. Then during release mode we can perform the full entry point analysis to optimize for runtime. 

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/brillig_entry_points.rs
```diff
@@ -0,0 +1,612 @@
...
+            }
+        }
+    }
+}
+
+/// Returns a map of Brillig entry points to all functions called in that entry point.
+/// This includes any nested calls as well, as we want to be able to associate
+/// any Brillig function with the appropriate global allocations.
+/// TODO: return a BTreeMap
+pub(crate) fn get_brillig_entry_points(
```
TomAFrench, line 202: This function must exist in like 3 other places at this point.

compiler/noirc_evaluator/src/ssa/opt/brillig_entry_points.rs
```diff
@@ -0,0 +1,612 @@
...
+            for entry_point in inner_call_entry_points {
+                add_function_to_clone(entry_point, inner_call);
+            }
+        } else if entry_points.contains(&inner_call) {
+            add_function_to_clone(inner_call_entry_points[0], inner_call);
+        }
+    }
+
+    functions_to_clone_map
+}
```
TomAFrench, line 145: It's not entirely clear what the conditions to choose to clone a function are without reading the entire function. This could do with a renaming or documentation.

compiler/noirc_evaluator/src/ssa/opt/brillig_entry_points.rs
```diff
@@ -0,0 +1,615 @@
...
+fn build_functions_to_clone(
+    functions: &BTreeMap<FunctionId, Function>,
+    inner_call_to_entry_point: HashMap<FunctionId, Vec<FunctionId>>,
+    entry_points: HashSet<FunctionId>,
+) -> HashMap<FunctionId, Vec<(Function, FunctionId)>> {
+    let mut functions_to_clone_map: HashMap<FunctionId, Vec<(Function, FunctionId)>> =
+        HashMap::default();
+
+    let mut add_function_to_clone = |entry_point: FunctionId, inner_call: FunctionId| {
+        let cloned_function = Function::clone_no_id(&functions[&inner_call]);
```
TomAFrench, line 130: I don't think that we need to actually clone the function here. This should be able to be done inside of `build_calls_to_update`, which would make this function much easier to test.

TomAFrench: We can then just make use of `clone_with_id` inside of the closure there.

compiler/noirc_evaluator/src/ssa/opt/brillig_entry_points.rs
```diff
@@ -0,0 +1,615 @@
...
+            explored_functions.clone(),
+        );
+    }
+}
+
+/// Builds a mapping from a [`FunctionId`] to the set of [`FunctionId`s][`FunctionId`] of all the brillig entrypoints
+/// from which this function is reachable.
+pub(crate) fn build_inner_call_to_entry_points(
+    brillig_entry_points: &BTreeMap<FunctionId, BTreeSet<FunctionId>>,
+) -> HashMap<FunctionId, Vec<FunctionId>> {
```
TomAFrench, line 288: We likely want to replace this to return a set of function ids rather than a vec.

compiler/noirc_evaluator/src/ssa/opt/brillig_entry_points.rs
TomAFrench: It would be nice to have some tests on this to sanity check this function once the above check is made.

compiler/noirc_evaluator/src/ssa/opt/brillig_entry_points.rs
vezenovm: I did deduplicate this logic from `brillig_globals.rs`, but that should be the only place?

TomAFrench:
Feel free to bump the execution time limit a little bit here.

compiler/noirc_evaluator/src/ssa/opt/brillig_entry_points.rs
vezenovm: I have switched to using `clone_with_id`. This does require an extra clone of the function to fetch the function for cloning inside of the `add_fn` closure. I agree the code is easier to follow now though.

compiler/noirc_evaluator/src/ssa/opt/brillig_entry_points.rs
vezenovm: done

compiler/noirc_evaluator/src/ssa/opt/brillig_entry_points.rs
vezenovm: > It would be nice to have some tests on this to sanity check this function once the above check is made.

Are you referring to `build_functions_to_clone` or `build_calls_to_update`? Do you not feel this logic is pretty much covered by the existing unit tests or are you thinking of a more explicit test on one of those methods?

TomAFrench: This was for `build_functions_to_clone` but the new implementation is so simple that :shrug:

compiler/noirc_evaluator/src/ssa/opt/brillig_entry_points.rs
TomAFrench: This was mostly based on a hunch as we've been building a lot of callgraph generating functions recently. No concrete example.

Feel free to resolve.

compiler/noirc_evaluator/src/ssa/opt/brillig_entry_points.rs
```diff
@@ -0,0 +1,610 @@
...
+            let func_value = &function.dfg[func_id];
+            let Value::Function(func_id) = func_value else { continue };
+            let Some(new_id) = calls_to_update.get(&(entry_point, *func_id)) else {
+                continue;
+            };
+
+            new_functions_map.insert(*func_id, *new_id);
+            let new_function_value_id = function.dfg.import_function(*new_id);
+            function.dfg[instruction_id] =
+                Instruction::Call { func: new_function_value_id, arguments: arguments.clone() };
```
TomAFrench, line 188:
```suggestion
                Instruction::Call { func: new_function_value_id, arguments };
```

TomAFrench (approved):
LGTM

Issue #7247, opened 2025-01-31
Sync temp fix 2


TomAFrench:
https://github.com/noir-lang/noir/pull/7241#discussion_r1936190245

aakoshh:
See the original comment with how to reproduce at https://github.com/AztecProtocol/aztec-packages/pull/11294#issuecomment-2624379102

I spent some time debugging this today. I checked out `master` on `aztec-packages`, uncommented the `panic!`, but it doesn't occur any more with the following:

```shell
cargo run -q -p nargo_cli -- --program-dir ../../noir-projects/noir-contracts compile --package token_contract --silence-warnings --inliner-aggressiveness 0 --force
```

I checked out the original branch (available as `af/sync-noir-11249` on `aztec-packages`), where I am still able to reproduce the issue. I used the following command to limit the SSA to the contract function where the panic occurs:

```shell
cargo run -q -p nargo_cli -- --program-dir ../../noir-projects/noir-contracts compile --package token_contract --silence-warnings --inliner-aggressiveness 0 --show-ssa --show-contract-fn setup_refund --force
```

I noticed by accident that if I use `--show-ssa --show-contract-fn setup_refund` on `master` I get no output at all, made me realise that the reason it doesn't fail on `master` is because the [setup_refund](https://github.com/AztecProtocol/aztec-packages/blob/feca4ed10f4978832b51d39d3c055a972073826f/noir-projects/noir-contracts/contracts/token_contract/src/main.nr#L609) method has been removed. So this bug is still relevant, probably.

Looking at the SSA I think the problem is with the defunctionalization. We start off having these methods:

```
After Initial SSA:
...
acir(inline) fn derive_aes_symmetric_key_and_iv_from_ecdh_shared_secret_using_sha256 f77 {
  b0(v45: Field, v46: Field, v47: u1):
    v50, v51 = call f109(v45, v46, v47, f110) -> ([u8; 16], [u8; 16])
    return v50, v51
}
acir(inline) fn derive_aes_symmetric_key_and_iv_from_ecdh_shared_secret f109 {
  b0(v45: Field, v46: Field, v47: u1, v48: function):
...
brillig(inline) fn derive_aes_symmetric_key_and_iv_from_ecdh_shared_secret_using_sha256 f182 {
  b0(v45: Field, v46: Field, v47: u1):
    v50, v51 = call f196(v45, v46, v47, f197) -> ([u8; 16], [u8; 16])
    return v50, v51
}
[...]

PR #7279, opened 2025-02-04
Try to turn acir lambdas into brillig

aakoshh:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/7247

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/monomorphization/mod.rs
```diff
@@ -500,7 +500,13 @@ impl<'interner> Monomorphizer<'interner> {
             },
             HirExpression::Literal(HirLiteral::Unit) => ast::Expression::Block(vec![]),
             HirExpression::Block(block) => self.block(block.statements)?,
-            HirExpression::Unsafe(block) => self.block(block.statements)?,
+            HirExpression::Unsafe(block) => {
+                let was_in_unconstrained_function = self.in_unconstrained_function;
+                self.in_unconstrained_function = true;
+                let res = self.block(block.statements);
+                self.in_unconstrained_function = was_in_unconstrained_function;
```
TomAFrench, line 507: This seems like we're treating the contents of an `unsafe` block as unconstrained. This isn't the case however as it's still constrained but the only difference is that we can call into unconstrained functions.

compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
```diff
@@ -226,19 +226,22 @@ fn find_variants(ssa: &Ssa) -> Variants {
...
     }
 
     let mut variants: Variants = BTreeMap::new();
 
     for (dispatch_signature, caller_runtime) in dynamic_dispatches {
-        let target_fns =
-            signature_to_functions_as_value.get(&dispatch_signature).cloned().unwrap_or_default();
-        variants.insert((dispatch_signature, caller_runtime), target_fns);
+        let key = (dispatch_signature, caller_runtime);
+        let target_fns = signature_to_functions_as_value.get(&key).cloned().unwrap_or_default();
```
aakoshh, line 242: With this change we end up in a situation where if the lambda is created in constrained code, and subsequently is passed to unconstrained code, then this list is empty and the compiler crashes further down.

vezenovm:
I think this can be closed post #7289 being resolved and #9412 cc @aakoshh

PR #7280, opened 2025-02-04
feat: while statement

asterite:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs
@@ -153,6 +153,7 @@ impl<'a> FunctionContext<'a> {
             Expression::Cast(cast) => self.codegen_cast(cast),
             Expression::For(for_expr) => self.codegen_for(for_expr),
             Expression::Loop(block) => self.codegen_loop(block),
+            Expression::While(condition, block) => self.codegen_while(condition, block),
             Expression::If(if_expr) => self.codegen_if(if_expr),
             Expression::Tuple(tuple) => self.codegen_tuple(tuple),
             Expression::ExtractTupleField(tuple, index) => {
@@ -588,7 +589,7 @@ impl<'a> FunctionContext<'a> {
         Ok(Self::unit_value())
     }
 
-    /// Codegens a loop, creating three new blocks in the process.
+    /// Codegens a loop, creating two new blocks in the process.
     /// The return value of a loop is always a unit literal.
     ///
     /// For example, the loop `loop { body }` is codegen'd as:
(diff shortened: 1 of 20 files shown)


compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs
```diff
@@ -620,6 +621,47 @@ impl<'a> FunctionContext<'a> {
...
+    /// while_entry:
+    ///   v0 = ... codegen cond ...
+    ///   jmpif v0, then: while_body, else: while_end  
+    /// while_body():
+    ///   v3 = ... codegen body ...
+    ///   br while_entry()
+    /// while_end():
+    ///   ... This is the current insert point after codegen_for finishes ...
+    /// ```
+    fn codegen_while(&mut self, while_: &While) -> Result<Values, RuntimeError> {
```
TomAFrench, line 640: Would be interesting if we could desugar for-loops into this. I would need to think on whether it makes the code cleaner or more messy (with the need for an induction variable)

asterite: I can try, but I'm not sure. I think the only "duplicated" code is `codegen_for` and `codgen_while`, but they are a bit different because of the loop var. And incrementing the loop var currently does an `add_unchecked` and I'm not sure we have an AST node to do that.

tooling/nargo_fmt/src/formatter/statement.rs
```diff
@@ -800,6 +833,29 @@ mod tests {
...
+    }
+
+    #[test]
+    fn format_non_empty_while() {
+        let src = " fn foo() {  while  condition  {  1 ; 2  }  } ";
+        let expected = "fn foo() {
+    while condition {
+        1;
+        2
+    }
```
michaeljklein, line 857: Please add a test for the `If there's a trailing semicolon, remove it` case, e.g.

```suggestion
    }
    
    #[test]
    fn format_while_with_semicolon() {
        let src = " fn foo() {  while  condition  {  1 ; 2  };  } ";
        let expected = "fn foo() {
    while condition {
        1;
        2
    }
```

michaeljklein (changes requested):
LGTM besides the formatter test

compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs
```diff
@@ -620,6 +621,47 @@ impl<'a> FunctionContext<'a> {
...
+    /// ```text
+    ///   jmp while_entry()
+    /// while_entry:
+    ///   v0 = ... codegen cond ...
+    ///   jmpif v0, then: while_body, else: while_end  
+    /// while_body():
+    ///   v3 = ... codegen body ...
+    ///   br while_entry()
+    /// while_end():
+    ///   ... This is the current insert point after codegen_for finishes ...
```
aakoshh, line 638:
```suggestion
    ///   jmp while_entry()
    /// while_end():
    ///   ... This is the current insert point after codegen_while finishes ...
```

asterite: Thanks! That `br` is from having worked with LLVM 😅

tooling/nargo_fmt/src/formatter/statement.rs
```diff
@@ -758,6 +791,42 @@ mod tests {
...
+
+    #[test]
+    fn format_while_with_semicolon() {
+        let src = " fn foo() {  while  condition  {  1 ; 2  };  } ";
+        let expected = "fn foo() {
+    while condition {
+        1;
+        2
+    }
```
TomAFrench, line 828: This looks like it should be a syntax error. Missing the semicolon on `2` implies that we're returning this value whereas a while loop doesn't have a return value.

asterite: Good point! Well, not a syntax error but maybe a semantic error (Rust says that it expected type `()` but got `i32`).

I just checked and we don't error in `for` nor in `loop`. If it's fine with you I'll capture an issue to fix all of those in one go, separately from this PR.

asterite: https://github.com/noir-lang/noir/issues/7408

PR #7282, opened 2025-02-05
chore: replace action with reusable workflows

TomAFrench:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/.github/workflows/publish-nightly.yml
+++ b/.github/workflows/publish-nightly.yml
@@ -6,17 +6,16 @@ on:
     - cron: "0 2 * * *"
 
 jobs:
-  dispatch-publish-es:
-    runs-on: ubuntu-22.04
-    steps:
-      - name: Dispatch to publish-nargo
-        uses: ./.github/workflows/publish-nargo.yml
-        with:
-          # Omitting a tag results in a nightly 
-          publish: true
+  public-nargo:
+    name: Publish Nargo binary
+    uses: ./.github/workflows/publish-nargo.yml
+    with:
+      # Omitting a tag results in a nightly 
+      publish: true
 
-      - name: Dispatch to publish-es-packages
-        uses: ./.github/workflows/publish-es-packages.yml
-        with:
-          noir-ref: ${{ env.GITHUB_REF }}
-          npm-tag: nightly
+  public-js-packages:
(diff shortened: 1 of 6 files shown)


.github/workflows/publish-nightly.yml
```diff
@@ -6,21 +6,16 @@ on:
...
+      publish: true
 
-
-      - name: Dispatch to publish-es-packages
-        uses: benc-uk/workflow-dispatch@v1
-        with:
-          workflow: publish-es-packages.yml
-          token: ${{ secrets.NOIR_REPO_TOKEN }}
-          inputs: "{ \"noir-ref\": \"${{ env.GITHUB_REF }}\", \"npm-tag\": \"nightly\" }"
+  public-js-packages:
```
aakoshh, line 16:
```suggestion
  publish-js-packages:
```

.github/workflows/release.yml
```diff
@@ -145,42 +145,24 @@ jobs:
...
-    steps:
-      - name: Dispatch to publish workflow
-        uses: benc-uk/workflow-dispatch@v1
-        with:
-          workflow: publish-nargo.yml
-          repo: noir-lang/noir
-          ref: master
-          token: ${{ secrets.GITHUB_TOKEN }}
-          inputs: '{ "tag": "${{ needs.release-please.outputs.tag-name }}", "publish": true }'
+    uses: noir-lang/noir/.github/workflows/publish-nargo.yml@master
```
aakoshh, line 148: Why is this pinned to `master`?

aakoshh: I might be misunderstanding it: it looks like the `release` workflow [only runs on master](https://github.com/noir-lang/noir/blob/2711b8414d1e5a071bb01ad11a2d0b176e366176/.github/workflows/release.yml#L6), and while previously we dispatched via the API, which required to tell which branch to run the publish workflow on, now we call it inside the workflow, so it will run on the same branch as the release workflow itself. 

With `@master`, would it just pull the workflow definition from `master` and apply it to whatever branch this workflow is running on, or would `master` apply to the jobs in the `publish-nargo.yml` workflow itself?

PR #7284, opened 2025-02-05
chore: do not differentiate fixed-length vs variable-length in Poseidon2

guipublic:
# Description

## Problem\*

We do not differentiate anymore fixed-length and variable-length poseidon2 hashes.

## Summary\*
Simply remove the terminator that was added at the end of the input.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/noir_stdlib/src/hash/poseidon2.nr
+++ b/noir_stdlib/src/hash/poseidon2.nr
@@ -13,7 +13,7 @@ pub struct Poseidon2 {
 impl Poseidon2 {
     #[no_predicates]
     pub fn hash<let N: u32>(input: [Field; N], message_size: u32) -> Field {
-        Poseidon2::hash_internal(input, message_size, message_size != N)
+        Poseidon2::hash_internal(input, message_size)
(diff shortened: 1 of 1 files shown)


TomAFrench:
I think we can hold off from making this change until we have Poseidon2 out in an external library

guipublic:
> I think we can hold off from making this change until we have Poseidon2 out in an external library

I don't see any benefit in holding this off?

saleel:
@TomAFrench someone in discord was complaining about this issue.
Isn't it better to have this fixed now to not have a discrepancy when its moved to a separate lib anyway?

lucasxia01 (commented):
I thought we were keeping a variable length implementation? like with message_size being a witness value

guipublic:
> I thought we were keeping a variable length implementation? like with message_size being a witness value

Yes we still do this, it's just the 0/1 at the end that we don't do anymore.

lucasxia01 (approved):
looks good to me

PR #7291, opened 2025-02-05
feat: simplify consecutive arithemtic

### LINKED ISSUE #5707: Optimize arithmetic across multiple instructions

asterite:
# Description

## Problem

Resolves #5707

## Summary

For now this just simplifies consecutive operations with the same operation (so "add 3", "sub 1" doesn't get simplified to "add 2" yet) but I wanted to see if just this change makes a different (I think it will, at least in the sha256 test program I noticed consecutive multiplications)

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
@@ -156,7 +156,7 @@ impl Binary {
                     return SimplifyResult::SimplifiedTo(lhs);
                 }
                 if let Some(instruction) =
-                    self.simplify_consecutive(lhs, rhs, lhs_const, rhs_const, lhs_type, dfg)
+                    self.simplify_consecutive(lhs, lhs_const, rhs_const, lhs_type, dfg)
                 {
                     return SimplifyResult::SimplifiedToInstruction(instruction);
                 }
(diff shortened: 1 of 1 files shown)


asterite:
Oooh, I didn't know you looked into this before. And your explanation makes sense. Probably making this after some other pass would be good.

I wonder about the reports that count number of instructions, etc. In my tests the number of SSA instructions for the sha256 program went down, but no changes were reported in this PR. Do the reports only count something else that isn't necessarily affected by the number of SSA instructions?

asterite:
Or maybe it's because it's a draft. I'll change to "Ready for review" to try that.

TomAFrench:
> In my tests the number of SSA instructions for the sha256 program went down, but no changes were reported in this PR. Do the reports only count something else that isn't necessarily affected by the number of SSA instructions?

The reports only count the number of acir and brillig opcodes we generate. ACIRgen lays down constraints lazily so you can keep adding constants into an expression and it will just increment the constant term rather than creating a new opcode.

I'm not sure if brillig does the same.

PR #7295, opened 2025-02-05
feat: require safety comments instead of safety doc comments

### LINKED ISSUE #7273: Suggest using a doc comment for the "Safety: " notice when a regular comment is used
### Problem

When getting the `warning: Unsafe block must have a safety doc comment above it`, I expect that a regular comment will work because doc comments usually belong at the top of declarations.

```noir
/**
 * Given an input array of type T, return a sorted array.
 * Type `T` must satisfy the Ord and Eq trait
 * The Eq function is used within an unconstrained function so its constraint-efficiently is not relevant
 * Note: sort_extended is likely more efficient as constraining `x < y` can typically be described
 *       more efficiently than returning a boolean that describes whether `x < y`
 **/
pub fn sort<T, let N: u32>(input: [T; N]) -> [T; N]
where
    T: std::cmp::Ord + std::cmp::Eq,
{
    // Safety: validated using check_shuffle..
    let sorted = unsafe { quicksort(input) };
```

### Happy Case

Either:
1. Support regular comments as `"Safety: "` comments or
2. Check whether the user accidentally used a regular comment and suggest "// -> ///" in that case

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #7273

## Summary

Instead of requiring `/// Safety:` on top of `unsafe`, with this PR we require `// Safety:`. This matches how Rust works (for that lint).

In order to avoid introducing a big breaking change, this PR:
- still allows doc comments on top of `unsafe` and statements, but **warns** when that happens
- in general all misplaced doc comments will now produce a warning instead of an error. This is just temporary so that the formatter can work for this upgrade. We can turn it back to how it was after a week or two.
- the formatter will turn safety doc comments into comments, so all you have to do to fix the warnings because of this change is just run the formatter

## Documentation

Check one:

# PR Checklist


tooling/nargo_fmt/src/formatter/doc_comments.rs
```diff
@@ -36,6 +36,48 @@ impl<'a> Formatter<'a> {
...
     }
+
+    /// Formats outer doc comments, turning them into regular comments if they start with "Safety:"
+    pub(super) fn format_outer_doc_comments_checking_safety(&mut self) {
+        self.skip_comments_and_whitespace();
+
+        let is_safety = match &self.token {
+            Token::LineComment(comment, Some(DocStyle::Outer))
+            | Token::BlockComment(comment, Some(DocStyle::Outer)) => {
+                comment.trim().to_lowercase().starts_with("safety:")
```
michaeljklein, line 47: Nit: this currently matches/preserves `sAfEtY` or `sAFeTY`. It may make sense to normalize to `[sS]afety` when formatting.

michaeljklein (approved):
LGTM with small nit r.e. `sAfEtY`

tooling/nargo_fmt/src/formatter/doc_comments.rs
asterite: Just to be clear, this code is for when there's `/// Safety` and it's changed into `// Safety`. That means that there's no specific code to turn `// sAfEtY` into `// Safety`. So maybe both should be done at the same time... but I don't feel strongly about this.

michaeljklein: All good :+1:

TomAFrench:
I'm fine with changing this but we should stick to this format going forwads.

asterite:
Captured https://github.com/noir-lang/noir/issues/7394

TomAFrench:
I've fixed some formatting issues with new tests in master. I've also expanded the formatter check to run on more of the test programs.

PR #7296, opened 2025-02-05
fix: allows for infinite brillig loops

### LINKED ISSUE #7103: Infinite `loop`s crash the compiler

guipublic:
# Description

## Problem\*

Resolves #7103 

## Summary\*
Infinite loops do not return, so it was needed to allow functions to return nothing during inlining, and then to fix the variable liveness analysis to handle infinite loops (with no header).

## Documentation\*

Check one:

# PR Checklist\*


test_programs/compile_success_with_bug/regression_7103/src/main.nr
TomAFrench, line 1: This shouldn't be passing afaik. The loop is causing a warning to be emitted but this folder implies that we should be looking for `bug:` in the stderr.

TomAFrench: Ah, this is because of the check being opt-in now.

TomAFrench (approved):
Looks good to me.

compiler/noirc_evaluator/src/ssa/ir/function.rs
```diff
@@ -164,7 +164,7 @@ impl Function {
     }
 
     /// Returns the return types of this function.
-    pub(crate) fn returns(&self) -> &[ValueId] {
+    pub(crate) fn returns(&self) -> Vec<ValueId> {
```
jfecher, line 167: Can you change this back to returning a `&[ValueId]` so we can avoid the allocation here?

We should be able to return `&[]` in the default case

compiler/noirc_evaluator/src/ssa/ir/function.rs
```diff
@@ -174,8 +174,7 @@ impl Function {
                 break;
             }
         }
-        function_return_values
-            .expect("Expected a return instruction, as function construction is finished")
+        function_return_values.unwrap_or(&Vec::new()).clone()
```
jfecher, line 177: Can you also update this function to directly return instead of using `function_return_values` then breaking? If we reach after the loop we could then return `&[]` without needing an expect or anything.

jfecher (commented):
Looks good, one nitpick

compiler/noirc_evaluator/src/ssa/ir/function.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/ir/function.rs
guipublic: Done

PR #7299, opened 2025-02-05
fix: always normalize ssa when priting at least one pass

### LINKED ISSUE #7235: Internal Consistency Evaluators Errors: aes128_encrypt
### Aim

Trying to print the SSA occasionally results in an error around the aes128_encrypt function. It only seems to happen when I try to limit printing to a specific pass name with `--show-ssa-pass`. If I print the full SSA it doesn't come up.

This has been happening since https://github.com/noir-lang/noir/pull/6973

### Expected Behavior

The SSA of the selected pass should be printed without errors. It seems to print it, and then die later during ACIR generation. 

### Bug

```console
$ cargo run -q -p nargo_cli -- --program-dir ../../noir-projects/noir-contracts compile --package stateful_test_contract --silence-warnings --show-ssa --show-contract-fn destroy_and_create_no_init_check --show-ssa-pass "Dead Instruction Elimination (2nd)"
After Dead Instruction Elimination (2nd):
g0 = u32 44
g1 = Field 18446744073709551616
...
  b5():
    v101 = unchecked_add v43, u32 1
    jmp b1(v101)
}
error: Internal Consistency Evaluators Errors: 

                    This is likely a bug. Consider opening an issue at https://github.com/noir-lang/noir/issues
  ┌─ std/aes128.nr:1:1
  │
1 │ #[foreign(aes128_encrypt)]
  │ - ICE: "Cannot find linked fn Function(Id(18), None)"
  │

```

### To Reproduce

It doesn't happen all the time, but try limiting the SSA to a single pass.

### Workaround

None

### Workaround Description

The SSA is printed, so there is nothing to work around, except figuring out why it can't find this particular function.

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.1 noirc version = 1.0.0-beta.1+4608c490032f2467cead6b76f4e81f9a4bf9ab50 (git version hash: 4608c490032f2467cead6b76f4e81f9a4bf9ab50, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

If you used `--show-ssa-pass` it would sometimes crash depending on the program.

Resolves #7235

## Summary

I'm not sure why that is the case, but it works fine with `--show-ssa` which normalized the SSA after every pass. As a quick fix, we now also always normalize the SSA after every pass, but we just print the relevant pass(es) if we are told to do so.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa.rs
+++ b/compiler/noirc_evaluator/src/ssa.rs
@@ -494,6 +494,11 @@ impl SsaBuilder {
     }
 
     fn print(mut self, msg: &str) -> Self {
+        // Always normalize if we are going to print at least one of the passes
+        if !matches!(self.ssa_logging, SsaLogging::None) {
+            self.ssa.normalize_ids();
+        }
+
         let print_ssa_pass = match &self.ssa_logging {
             SsaLogging::None => false,
             SsaLogging::All => true,
@@ -505,7 +510,6 @@ impl SsaBuilder {
             }
         };
         if print_ssa_pass {
-            self.ssa.normalize_ids();
             println!("After {msg}:\n{}", self.ssa);
         }
         self


aakoshh:
Is this related to https://github.com/noir-lang/noir/issues/7235 in any way?

aakoshh (approved):
Looks like a benign change to me 👍 
It would certainly explain why I also experienced problems when printing one pass, but not when all of them are printed, good thinking!

asterite:
It does seem to resolve #7235, though to be honest I don't know why. Maybe there's a bug when normalizing... though I don't understand why it only happens if we normalize once and not always.

In any case this makes debugging easier for the time being...

PR #7301, opened 2025-02-05
feat: add native `u128` type

TomAFrench:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
@asterite Can you take over this PR? I'm not sure that I can devote enough time to it atm.

Remaining things to do include:
- Block creation of the `i128` type (we can potentially add this in future but I don't think we should add this in this PR)
- Add an SSA pass which lowers multiplication and division of u128s into being implemented in smaller types for which we can perform these operations safely
    - We have implementations for the `U128` struct here which can be used as inspiration: https://github.com/noir-lang/noir/blob/582f56e6b6ea43ab79b08aacfe7f1ba67a097f26/noir_stdlib/src/uint128.nr#L198-L211
    - addition (and potentially subtraction) should be safe to leave as is.
    - This pass should only affect ACIR functions as brillig has full support for u128 types.
- Add more selective version of this check: https://github.com/noir-lang/noir/pull/7301/files#diff-7c64342380b96ff80bc78d976a826185fbf2f08c0a10d77318dde72f697cb3b7 which only blocks unsafe operations.
- A bunch of tests

compiler/noirc_evaluator/src/ssa/ir/types.rs
```diff
@@ -65,7 +65,8 @@ impl NumericType {
     ) -> Option<String> {
         match self {
             NumericType::Unsigned { bit_size } => {
-                let max = 2u128.pow(bit_size) - 1;
+                // The math here is a bit convoluted to avoid overflowing for u128
```
TomAFrench, line 68: We've generally used an if statement to just directly return `u128::MAX` in other cases.

asterite: Ah, I considered doing that but I wasn't sure which way was better. I'll change it 👍

asterite:
>  Add an SSA pass which lowers multiplication and division of u128s into being implemented in smaller types for which we can perform these operations safely

Is there an operation that currently fails for this? I'm mainly trying to understand how it fails in order to understand how to implement this.

> Add more selective version of this check

Why is that check needed? I know it says "for the truncation technique" but where is that, and why every math operation needs this check?

TomAFrench:
> Is there an operation that currently fails for this? I'm mainly trying to understand how it fails in order to understand how to implement this.

The issue is that the bn254 field can only represent numbers up to ~2^253 safely as otherwise we overflow the field modulus. If we then multiply 2 128 bit numbers together it's possible that we'll wrap around the field modulus and get a "small" result, this will then pass the overflow check as it fits within 128 bits. This isn't an issue for the other types as they're small enough that we'll never overflow the field.

We then want to make sure that there's no operations which we do in ACIR gen which can overflow the field modulus. These should then be lowered to be implemented in terms of the smaller types to avoid the potential for an uncaught overflow.

TomAFrench:
We then need to consider the maximum sizes of values generated by different operations, e.g multiplying 2 u128s together can give a 256 bit number so it's unsafe, adding 2 u128s can give a 129 bit number so it's safe, etc. I haven't thought through fully about comparisons, etc.

TomAFrench:
>>Add more selective version of this check
>
>Why is that check needed? I know it says "for the truncation technique" but where is that, and why every math operation needs this check?

We need this as a check in case we later on allow operations of an unsafe bitsize creep into acirgen. We can then fail compilation rather than generating an insecure circuit.

TomAFrench:
Yeah that makes sense to me. We'd still catch all potential overflows between this pre-check and the regular overflow check while being more efficient than the regular U128 algo.

asterite:
Do we still need [this check](https://github.com/noir-lang/noir/pull/7301/files#diff-7c64342380b96ff80bc78d976a826185fbf2f08c0a10d77318dde72f697cb3b7) now that we check that multiplication doesn't overflow?

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -435,6 +447,68 @@ impl<'a> FunctionContext<'a> {
...
+            //
+            // lhs_masked = lhs & mask
+            // rhs_masked = rhs & mask
+            // res = lhs_masked * rhs_masked
+            // assert_eq(res, 0, "attempt to multiply with overflow")
+            //
+            // which is equivalent to checking that either of them is less than 2^64
+
+            let lhs_masked = self.builder.insert_binary(lhs, BinaryOp::And, mask);
+            let rhs_masked = self.builder.insert_binary(rhs, BinaryOp::And, mask);
```
TomAFrench, line 499: Euclidean division should be more efficient here rather than bitwise operations.

asterite: Done! Let me know if I got this right.

asterite:
I think other operations are correctly constrained. That is, we always check for unsigned overflow in acir, and the only operation that could overflow the field element max is multiplication.

I thought shifting to the left could overflow and I found a [bug](https://github.com/noir-lang/noir/issues/7428), but it seems in general when you shift too much you just get a zero instead of an overflow so I assumed that's expected behavior.

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -355,7 +356,10 @@ impl<'a> FunctionContext<'a> {
 
     /// Insert constraints ensuring that the operation does not overflow the bit size of the result
     ///
-    /// If the result is unsigned, overflow will be checked during acir-gen (cf. issue #4456), except for bit-shifts, because we will convert them to field multiplication
+    /// If the result is unsigned, overflow will be checked during acir-gen (cf. issue #4456), except for:
+    /// - bit-shifts, because we will convert them to field multiplication
+    /// - when multiplying two u128: the overflow will be checking during acir-gen, but we also need to do an
+    ///   extra check because the result might wrap around the field maximum value and end up in a valid u128 value
```
TomAFrench, line 362:
```suggestion
    /// - when multiplying two u128: the overflow will be checked during acir-gen, but we also need to do an
    ///   extra check because the result might wrap around the field maximum value and end up in a valid u128 value
```

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -408,10 +412,19 @@ impl<'a> FunctionContext<'a> {
...
                     }
+                    BinaryOpKind::Multiply => {
+                        // Part of the overflow check is done here for u128
+                        if matches!(result_type, NumericType::Unsigned { bit_size: 128 }) {
+                            self.check_u128_mul_overflow(lhs, rhs);
+                        }
+
+                        // Overflow check is also done in acir-gen
+                        return result;
+                    }
```
TomAFrench, line 427: I just noticed that we're doing this up in SSA gen. We want to avoid this happening in brillig functions as the overflow checks in brillig gen should work fine.

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -435,6 +448,58 @@ impl<'a> FunctionContext<'a> {
...
+            let rhs = self.builder.insert_binary(rhs, BinaryOp::Div, two_pow_64);
+            self.builder.insert_binary(lhs, BinaryOp::Mul { unchecked: true }, rhs)
+        };
+
+        let zero = self.builder.numeric_constant(FieldElement::zero(), u128);
+        self.builder.insert_constrain(
+            res,
+            zero,
+            Some(ConstrainError::StaticString("attempt to multiply with overflow".to_string())),
+        );
```
TomAFrench, line 500: We've also got some type mismatches going on here where we can end up constraining a boolean against a u128.

asterite:
> but it seems in general when you shift too much you just get a zero instead of an overflow so I assumed that's expected behavior

Oh, it seems that if you shift by a value that will surely overflow then you get an assertion failure. But if you do it by steps then you don't get a failure:

```noir
let x: u8 = 1;
let _ = x << 8; // produces an assertion failure

let x: u8 = 1;
let x = x << 4;
let x = x << 4; // no error here
```

Should I capture that as a bug?

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
TomAFrench: I think a small SSA pass which just searches for multiplications and then inserts the instructions which are generated by `check_u128_mul_overflow` afterwards would be best. We can run this towards the end of the SSA pipeline which avoids the situation which we have currently where the overflow checks could prevent a multiplication from being optimized out.

TomAFrench: Eh, I could tolerate this being done in ssa gen if we have an issue to explore moving it further down the SSA pipeline. We should be able to distinguish between runtimes up here at least.

TomAFrench:
> Should I capture that as a bug?

Yeah the behaviour of bit shifts in acir is kind screwy in general though: https://github.com/noir-lang/noir/issues/7412

TomAFrench:
Looks good to me. I've spun off #7432 which I noticed from looking at the SSA from porting `1_mul` to use `u128`.

@kashbrti can you give a quick sanity check wrt to mul being the only operation which we need to worry about here?
Otherwise we just need to update documentation, etc.

TomAFrench:
This looks like it will cause an overflow when performing a bitwise not on u128 values.

https://github.com/noir-lang/noir/blob/21f3e64f9f9fa857a7f0c10cb8645be46d2af2e8/compiler/noirc_evaluator/src/acir/acir_variable.rs#L778-L780

@asterite can you add some test programs which perform all potential operations on u128 types to catch all of these?

asterite:
(that is: the code looks pretty complex so I have to sit down and understand it before making it work, but first I wanted to know if it is possible to make it work without too much effort because I don't know if that `unreachable` means there's no easy way to do it for `u128`)

TomAFrench:
It should be doable. The issue is that when we're dividing by a non-constant denominator then when checking that `a = q * b + r` both `q` and `b` can be potentially any u128 value. We then need to ensure that we're not overflowing the field when we multiply them together again.

compiler/noirc_evaluator/src/acir/acir_variable.rs
```diff
@@ -944,6 +935,20 @@ impl<F: AcirField, B: BlackBoxFunctionSolver<F>> AcirContext<F, B> {
                     predicate,
                     rhs_const.num_bits(),
                 )?;
+            } else if bit_size == 128 {
+                // q and b are u128 and q*b could overflow so we check that either q or b are less than 2^64
+                let two_pow_64 = F::from(1_u128 << 64);
```
TomAFrench, line 940: We have a helper function to get powers of 2 in a safe way.

test_programs/execution_success/u128_type/src/main.nr
```diff
@@ -0,0 +1,21 @@
...
+    assert_eq(x + y, const_x + const_y);
+    assert_eq(y - x, const_y - const_x);
+    assert_eq(x * y, const_x * const_y);
+    assert_eq(y / x, const_y / const_x);
+    assert_eq(y % x, const_y % const_x);
+    assert_eq(!x, !const_x);
+    assert_eq(x ^ y, const_x ^ const_y);
+    assert_eq(x & y, const_x & const_y);
+    assert_eq(x | y, const_x | const_y);
+    assert_eq(x >> 2, const_x >> 2);
```
TomAFrench, line 14: Having some bit shifts by unknown sizes would be good too. This is optimized down to a simple division ATM.

TomAFrench:
LGTM

asterite (approved):

PR #7310, opened 2025-02-06
chore: split acirgen into multiple modules

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This file is huge so I'm wanting to break out the functionality related the black boxes and brillig calls as they have a large number of private helper functions which we don't need to worry about most of the time.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/acir/black_box.rs
```diff
@@ -0,0 +1,309 @@
...
+
+                (vec![proof_type_constant], Vec::new())
+            }
+            _ => (vec![], vec![]),
+        };
+        let inputs = self.prepare_inputs_for_black_box_func(inputs, name)?;
+        // Call Black box with `FunctionInput`
+        let mut results = vecmap(&constant_outputs, |c| self.add_constant(*c));
+
+        let output_vars = vecmap(0..output_count, |_| self.add_variable());
```
aakoshh, line 185: This does look cleaner than the way it was broken up before 👍

aakoshh (approved):
LGTM!

PR #7312, opened 2025-02-06
feat(experimental): Compile match expressions

jfecher:
# Description

## Problem\*

Working towards #988 

## Summary\*

Adds support for compiling match expressions. Specifically:
- During elaboration, convert match expressions to decision trees where:
  - No tag value check is repeated
  - Nested enums, tuples, and integer types are supported
  - NO errors or warnings are issued, they're all stubbed out for future PRs (just write perfect code)
  - The match compiler supports more pattern types than we currently parse. In addition to basic constructors & variable patterns it also supports:
    - Range patterns for ranges of integers
    - Pattern guards (`_ if .. => ...`)
    - We do not support matching on strings, arrays, or slices but could consider this for future work.
- Support for lowering decision trees into SSA via many `jmpif`s
  - We'll want to add a `TerminatorInstruction::Switch` eventually but I left this for future work

So you can now execute & prove code containing match expressions.

## Additional Context

Quite a few parts of match expressions are still left as TODOs for future PRs to keep this PR "small":
- Redundancy checking
- Missing case analysis
- Errors in general (e.g. `error: type '&mut T' can't be matched on`)
- There is an odd stack overflow when using variants which reference themselves in their generics. E.g. `MyOption<MyOption<Field>>` will cause a stack overflow in `Monomorphizer::convert_type`.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/elaborator/enums.rs
```diff
@@ -243,4 +253,672 @@ impl Elaborator<'_> {
...
+            this.unify(actual, expected_type, || TypeCheckError::TypeMismatch {
+                expected_typ: expected_type.to_string(),
+                expr_typ: actual.to_string(),
+                expr_span,
+            });
+        };
+
+        match expression.kind {
+            ExpressionKind::Literal(Literal::Integer(x, negative)) => {
+                let Some(mut x): Option<i64> = x.try_to_u64().and_then(|x| x.try_into().ok())
```
asterite, line 306: Is there a reason why `Pattern::Int` takes an i64 instead of `(FieldElement, bool /* negative */)`? For example this code triggers the panic below:

```noir
fn main() {
    let x = 1;
    match x {
        18446744073709551615 => {
            println("yes")
        }
        _ => println("no")
    };
}
```

asterite: (not blocking, though, as this scenario would be rare, but a follow-up issue could be captured)

compiler/noirc_frontend/src/elaborator/enums.rs
```diff
@@ -243,4 +253,672 @@ impl Elaborator<'_> {
...
+        let unify_with_expected_type = |this: &mut Self, actual| {
+            this.unify(actual, expected_type, || TypeCheckError::TypeMismatch {
+                expected_typ: expected_type.to_string(),
+                expr_typ: actual.to_string(),
+                expr_span,
+            });
+        };
+
+        match expression.kind {
+            ExpressionKind::Literal(Literal::Integer(x, negative)) => {
```
asterite, line 305: Would it be easy to also support bools? Of course it could be a follow-up thing.

compiler/noirc_frontend/src/elaborator/enums.rs
```diff
@@ -243,4 +253,672 @@ impl Elaborator<'_> {
...
+                // - in `(a, b)`, both variables may be free and should be defined, or
+                //   may refer to an enum variant named `a` or `b` in scope.
+                // - Possible diagnostics improvement: warn if `a` is defined as a variable
+                //   when there is a matching enum variant with name `Foo::a` which can
+                //   be imported. The user likely intended to reference the enum variant.
+                let path_len = path.segments.len();
+                let location = Location::new(path.span(), self.file);
+                let last_ident = path.last_ident();
+
+                match self.resolve_path(path) {
```
asterite, line 329: Should probably use `resolve_path_or_error`. That will report any warnings, for example if a type was private but still exists.

(the name `resolve_path_or_error` is not ideal but I don't know what other name would be better)

compiler/noirc_frontend/src/elaborator/enums.rs
```diff
@@ -243,4 +253,672 @@ impl Elaborator<'_> {
...
+        name: Expression,
+        args: Vec<Expression>,
+        expected_type: &Type,
+    ) -> Pattern {
+        match name.kind {
+            ExpressionKind::Variable(path) => {
+                let span = path.span();
+                let location = Location::new(span, self.file);
+
+                match self.resolve_path(path) {
```
asterite, line 420: Here it should also be `resolve_path_or_error`

compiler/noirc_frontend/src/elaborator/enums.rs
```diff
@@ -243,4 +253,672 @@ impl Elaborator<'_> {
...
+                    GlobalValue::Resolved(Value::Enum(tag, ..)) => tag,
+                    _ => todo!("Value is not an enum constant"),
+                };
+
+                let global_type = self.interner.definition_type(global.definition_id);
+                let actual_type = global_type.instantiate(self.interner).0;
+                (actual_type, Vec::new(), variant_index)
+            }
+            PathResolutionItem::Method(_type_id, _turbofish, func_id) => {
+                // TODO: Is this turbofish on the type or on the method?
```
asterite, line 471: It's on the type.

compiler/noirc_frontend/src/elaborator/enums.rs
jfecher: Ah, yes. I meant to add a test case for bools today - seems like if I did it would've ran into me forgetting the bool case here. I'll add it - we should already support bools in the match compiler.

compiler/noirc_frontend/src/elaborator/enums.rs
jfecher: This is leftover from when I was calling `resolve_path` twice, calling it again in the helper function below and erroring only the second time. Seems I changed that but did not update here. I'll change it.

compiler/noirc_frontend/src/elaborator/enums.rs
jfecher: I didn't think I could get the range constructor working with the boolean. I tried it again though and it wasn't too bad - just had to create a wrapper type for `SignedField` and use that. This also fixed a bug with representation of signed fields in an enum tag value.

compiler/noirc_frontend/src/elaborator/enums.rs
```diff
@@ -243,4 +252,670 @@ impl Elaborator<'_> {
...
+                        resolution,
+                        Vec::new(),
+                        expected_type,
+                        location.span,
+                    ),
+                    Err(_) if path_len == 1 => {
+                        // Define the variable
+                        let kind = DefinitionKind::Local(None);
+                        // TODO: `allow_shadowing` is false while I'm too lazy to add a check that we
+                        // don't define the same name multiple times in one pattern.
```
asterite, line 336: Is there an issue captured for this?

asterite: Or maybe not need to capture anything... I see a bunch of `todo!` below. I guess the idea is to fix them in follow-up PRs, right?

compiler/noirc_frontend/src/hir/resolution/errors.rs
```diff
@@ -60,6 +60,8 @@ pub enum ResolverError {
     InvalidArrayLengthExpr { span: Span },
     #[error("Integer too large to be evaluated in an array length context")]
     IntegerTooLarge { span: Span },
+    #[error("Integer too large to be evaluated in an array length context")]
+    IntegerPatternTooLarge { span: Span },
```
asterite, line 64: I think this ended up being unused.

compiler/noirc_frontend/src/hir_def/expr.rs
```diff
@@ -354,3 +355,156 @@ pub struct HirLambda {
...
+}
+
+impl std::fmt::Display for Constructor {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Constructor::True => write!(f, "true"),
+            Constructor::False => write!(f, "false"),
+            Constructor::Unit => write!(f, "()"),
+            Constructor::Int(x) => write!(f, "{x}"),
+            Constructor::Tuple(_) => Ok(()), // args are printed after already
```
asterite, line 495: What does this mean?

compiler/noirc_frontend/src/elaborator/enums.rs
jfecher: > Or maybe not need to capture anything... I see a bunch of todo! below. I guess the idea is to fix them in follow-up PRs, right?

Right. I made an issue anyway though since this was a smaller item that'd be easier to forget.

compiler/noirc_frontend/src/hir_def/expr.rs
jfecher: This is just the constructor itself, so the function value. We print the arguments of the constructor e.g. `(x, y)` afterwards anyway and in that case it already matches a tuples syntax so we don't want to print an extra `()` or similar before it since that wouldn't be correct. I'll clarify the comment.

compiler/noirc_frontend/src/hir/resolution/errors.rs
jfecher: Removed

compiler/noirc_frontend/src/elaborator/enums.rs
```diff
@@ -243,4 +252,670 @@ impl Elaborator<'_> {
...
+        let block = HirExpression::Block(HirBlockExpression { statements: vec![let_, body] });
+        let block = self.interner.push_expr(block);
+        self.interner.push_expr_type(block, body_type);
+        self.interner.push_expr_location(block, location.span, location.file);
+        block
+    }
+}
+
+/// Patterns are represented as resolved expressions currently.
+/// This type alias just makes code involving them more clear.
```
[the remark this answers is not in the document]
jfecher: Yes, updated now

compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs
```diff
@@ -752,6 +754,155 @@ impl<'a> FunctionContext<'a> {
...
+            self.builder.switch_to_block(else_block);
+        }
+
+        if let Some(branch) = &match_expr.default_case {
+            let results = self.codegen_expression(branch)?.into_value_list(self);
+            self.builder.terminate_with_jmp(end_block, results);
+        } else {
+            // If there is no default case, assume we saved the last case from the
+            // last_case optimization above
+            let case = match_expr.cases.last().unwrap();
```
vezenovm, line 804: We allow empty types right? Can this potentially panic if we had an empty match statement?

compiler/noirc_frontend/src/elaborator/enums.rs
```diff
@@ -243,4 +252,670 @@ impl Elaborator<'_> {
...
+        // We must unify the actual type before `expected_arg_types` are used since those
+        // are instantiated and rely on this already being unified.
+        self.unify(&actual_type, expected_type, || TypeCheckError::TypeMismatch {
+            expected_typ: expected_type.to_string(),
+            expr_typ: actual_type.to_string(),
+            expr_span: span,
+        });
+
+        if args.len() != expected_arg_types.len() {
+            // error expected N args, found M?
```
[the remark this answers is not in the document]
jfecher: Yep, just leaving errors for a future PR

compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs
jfecher: That code works as expected - you can even illegally create such a type via `std::mem::zeroed::<Empty>()` and call `my_match`. When this happens the match contains one branch which is just `assert(false);`

vezenovm: I think it would be a good test case still, but can be done in follow-ups as this is experimental

vezenovm (approved):
Approving, just have those few comments for tracking follow-ups

compiler/noirc_frontend/src/monomorphization/mod.rs
```diff
@@ -1987,6 +1989,72 @@ impl<'interner> Monomorphizer<'interner> {
...
+            HirMatch::Guard { cond, body, otherwise } => {
+                let condition = Box::new(self.expr(cond)?);
+                let consequence = Box::new(self.expr(body)?);
+                let alternative = Some(Box::new(self.match_expr(*otherwise, expr_id)?));
+
+                Ok(ast::Expression::If(ast::If {
+                    condition,
+                    consequence,
+                    alternative,
+                    typ: ast::Type::Unit, // TODO
```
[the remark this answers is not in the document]
jfecher: Ended up just filling it in

compiler/noirc_frontend/src/elaborator/enums.rs
```diff
@@ -243,4 +252,670 @@ impl Elaborator<'_> {
...
+                    GlobalValue::Resolved(Value::Enum(tag, ..)) => tag,
+                    _ => todo!("Value is not an enum constant"),
+                };
+
+                let global_type = self.interner.definition_type(global.definition_id);
+                let actual_type = global_type.instantiate(self.interner).0;
+                (actual_type, Vec::new(), variant_index)
+            }
+            PathResolutionItem::Method(_type_id, _type_turbofish, func_id) => {
+                // TODO: Take type_turbofish into account when instantiating the function's type
```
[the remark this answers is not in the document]
jfecher: Updated to include issue number

compiler/noirc_frontend/src/elaborator/enums.rs
```diff
@@ -243,4 +252,670 @@ impl Elaborator<'_> {
...
+                        // TODO: `allow_shadowing` is false while I'm too lazy to add a check that we
+                        // don't define the same name multiple times in one pattern.
+                        let id = self.add_variable_decl(last_ident, false, false, true, kind).id;
+                        self.interner.push_definition_type(id, expected_type.clone());
+                        Pattern::Binding(id)
+                    }
+                    Err(error) => {
+                        self.push_err(error);
+                        // Foo::Bar pattern of length at least 2 not found...
+                        // default to defining a variable of the same name?
```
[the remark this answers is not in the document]
jfecher: This is for a resolution error where the name doesn't exist, the question mark is more so meant to be because it isn't clear what pattern to return when we failed to resolve the name so I defaulted to just a variable pattern. I reworded the comment. In the future maybe we could have `Pattern::Error` but I'll keep it this way for now.

PR #7316, opened 2025-02-06
fix!: check abi integer input is within signed range

### LINKED ISSUE #7304: 8-bit signed integer doesn't have correct range when passed as a `main` parameter
### Aim

The compiler should error when an integer outside of the range `-128`..`127` is passed as an argument to the following function:
```noir
fn main(x: i8) {
  println(x);
}
```
With a `Prover.toml`:
```
x = "200"
```

### Expected Behavior

This program should fail to execute with the following error:
```
The value passed for parameter `x` does not match the specified type:
Value Field(200) does not fall within range of allowable values for a Integer { sign: Signed, width: 8 }
```

### Bug

The program successfully executes and it prints "-56".

### To Reproduce

1. `nargo check`
2. Write `x = "200"` in the `Prover.toml`
3. `nargo execute` 

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.1 noirc version = 1.0.0-beta.1+3a42eb5c68f9616f0ebe367c894f0376ba41e0ef

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #7304

## Summary

I accidentally started fixing this when parsing strings, but I think that's also good to have. (edit: it turns out the original issue was about strings 😅)

Then I added a similar check for integers... though here we only need to check the maximum value because input values are always unsigned (~is there a reason for this?~ -> this PR now also allows negative integers in at least Prover.toml (I hope that's fine!)).

Finally, I noticed that the error message produced in this case was very barebones: it didn't mention which parameter name errored. So now that's shown, and also if the value is inside an array, that index is shown (a similar thing is done for tables where we show which field we are in).

The error shown for the program in #7304 is now this:

```
The value passed for parameter `x` is invalid:
Value 200 exceeds maximum value of 127
```

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/noirc_abi/src/errors.rs
+++ b/tooling/noirc_abi/src/errors.rs
@@ -10,17 +10,17 @@ use thiserror::Error;
 pub enum InputParserError {
     #[error("input file is badly formed, could not parse, {0}")]
     ParseInputMap(String),
-    #[error("Expected witness values to be integers, provided value causes `{0}` error")]
-    ParseStr(String),
+    #[error("Expected witness values to be integers, provided value causes `{value}` error")]
+    ParseStr { value: String },
     #[error("Input {value} exceeds maximum value of {max}")]
     InputExceedsMaximum { value: u64, max: u64 },
     #[error("Input {value} outside of valid range. Value must fall within [{min}, {max}]")]
     InputOutsideOfRange { value: BigInt, min: BigInt, max: BigInt },
     #[error(
-        "Input {0} exceeds field modulus. Values must fall within [0, {})",
+        "Input {value} exceeds field modulus. Values must fall within [0, {})",
         FieldElement::modulus()
     )]
-    InputExceedsFieldModulus(String),
+    InputExceedsFieldModulus { value: String },
     #[error("cannot parse value into {0:?}")]
     AbiTypeMismatch(AbiType),
     #[error("Expected argument `{0}`, but none was found")]
--- a/tooling/noirc_abi/src/errors.rs
+++ b/tooling/noirc_abi/src/errors.rs
@@ -10,17 +10,21 @@ use thiserror::Error;
 pub enum InputParserError {
     #[error("input file is badly formed, could not parse, {0}")]
     ParseInputMap(String),
-    #[error("Expected witness values to be integers, provided value causes `{value}` error")]
-    ParseStr { value: String },
-    #[error("Input {value} exceeds maximum value of {max}")]
-    InputExceedsMaximum { value: u64, max: u64 },
-    #[error("Input {value} outside of valid range. Value must fall within [{min}, {max}]")]
-    InputOutsideOfRange { value: BigInt, min: BigInt, max: BigInt },
     #[error(
-        "Input {value} exceeds field modulus. Values must fall within [0, {})",
+        "The value passed for parameter `{arg_name}` has an error:\nExpected witness values to be integers, provided value causes `{value}` error"
+    )]
+    ParseStr { arg_name: String, value: String },
+    #[error("The value passed for parameter `{arg_name}` has an error:\nValue {value} exceeds maximum value of {max}")]
+    InputExceedsMaximum { arg_name: String, value: u64, max: u64 },
+    #[error(
+        "The value passed for parameter `{arg_name}` has an error:\nValue {value} outside of valid range. Value must fall within [{min}, {max}]"
+    )]
+    InputOutsideOfRange { arg_name: String, value: BigInt, min: BigInt, max: BigInt },
+    #[error(
+        "The value passed for parameter `{arg_name}` has an error:\nValue {value} exceeds field modulus. Values must fall within [0, {})",
         FieldElement::modulus()
     )]
-    InputExceedsFieldModulus { value: String },
+    InputExceedsFieldModulus { arg_name: String, value: String },
     #[error("cannot parse value into {0:?}")]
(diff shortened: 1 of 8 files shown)


asterite:
Hm, it seems the behavior before this PR was maybe expected:

https://github.com/noir-lang/noir/tree/master/test_programs/execution_success/signed_cmp

Maybe because there's no way to pass negative integers otherwise (well, you could use a string).

tooling/noirc_abi/src/errors.rs
```diff
@@ -2,17 +2,31 @@ use crate::{
...
 #[derive(Debug, Error)]
 pub enum InputParserError {
     #[error("input file is badly formed, could not parse, {0}")]
     ParseInputMap(String),
-    #[error("Expected witness values to be integers, provided value causes `{0}` error")]
-    ParseStr(String),
-    #[error("Could not parse hex value {0}")]
-    ParseHexStr(String),
+    #[error(
+        "The value passed for parameter `{arg_name}` is invalid:\nExpected witness values to be integers, provided value causes `{value}` error"
```
michaeljklein, line 14: - Is `value` an error?

- Would it make sense to have both the value string and the error string here, e.g.

```suggestion
        "The value passed for parameter `{arg_name}` is invalid:\nExpected witness values to be integers, but `{value}` failed with `{error_str}`"
```
?

tooling/noirc_abi/src/errors.rs
```diff
@@ -2,17 +2,31 @@ use crate::{
...
+    )]
+    ParseStr { arg_name: String, value: String },
+    #[error("The value passed for parameter `{arg_name}` is invalid:\nValue {value} is less than minimum allowed value of {min}")]
+    InputExceedsMinimum { arg_name: String, value: String, min: String },
+    #[error("The value passed for parameter `{arg_name}` is invalid:\nValue {value} exceeds maximum allowed value of {max}")]
+    InputExceedsMaximum { arg_name: String, value: String, max: String },
+    #[error(
+        "The value passed for parameter `{arg_name}` is invalid:\nValue {value} outside of valid range. Value must fall within [{min}, {max}]"
+    )]
+    InputOutsideOfRange { arg_name: String, value: BigInt, min: BigInt, max: BigInt },
```
michaeljklein, line 24: `InputOutsideOfRange` seems redundant with the more specific `InputExceedsMinimum`/`InputExceedsMaximum`. Could we exclusively use those more specific errors?

tooling/noirc_abi/src/input_parser/mod.rs
```diff
@@ -405,31 +459,41 @@ mod test {
...
 
-        let value = parse_str_to_signed("-1", 8).unwrap();
+        let value = parse_str_to_signed("-1", 8, "arg_name").unwrap();
         assert_eq!(value, FieldElement::from(255_u128));
 
-        let value = parse_str_to_signed("-1", 16).unwrap();
+        let value = parse_str_to_signed("-1", 16, "arg_name").unwrap();
         assert_eq!(value, FieldElement::from(65535_u128));
+
+        assert!(parse_str_to_signed("127", 8, "arg_name").is_ok());
```
michaeljklein, line 488:
```suggestion
        assert_eq!(parse_str_to_signed("127", 8, "arg_name"), Ok(FieldElement::from(127_i128)));
```

tooling/noirc_abi/src/input_parser/mod.rs
```diff
@@ -405,31 +459,41 @@ mod test {
...
+        assert!(parse_str_to_signed("128", 8, "arg_name").is_err());
+        assert!(parse_str_to_signed("-128", 8, "arg_name").is_ok());
```
michaeljklein, line 490:
```suggestion
        assert_eq!(parse_str_to_signed("-128", 8, "arg_name"), Ok(FieldElement::from(-128_i128)));
```

tooling/noirc_abi/src/input_parser/mod.rs
```diff
@@ -405,31 +459,41 @@ mod test {
...
+        assert!(parse_str_to_signed("-129", 8, "arg_name").is_err());
+
+        assert!(parse_str_to_signed("32767", 16, "arg_name").is_ok());
```
michaeljklein, line 493:
```suggestion
        assert_eq!(parse_str_to_signed("32767", 8, "arg_name"), Ok(FieldElement::from(32767_i128)));
```

tooling/noirc_abi/src/input_parser/mod.rs
```diff
@@ -405,31 +459,41 @@ mod test {
...
+        assert!(parse_str_to_signed("32768", 16, "arg_name").is_err());
+        assert!(parse_str_to_signed("-32768", 16, "arg_name").is_ok());
```
michaeljklein, line 495:
```suggestion
        assert_eq!(parse_str_to_signed("-32768", 16, "arg_name"), Ok(FieldElement::from(-32768_i128)));
```

michaeljklein (changes requested):

michaeljklein (approved):
LGTM, though just noticed that `InputExceedsMinimum` sounds like `min < input` instead of the intended `input < min` (not blocking).

PR #7327, opened 2025-02-09
fix: lock git dependencies folder when resolving workspace

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR adds a simple file lock on the `~/nargo` directory to prevent two instances of nargo attempting to download dependencies into the same folder at the same time.

This can't cause deadlock with the nargo.toml file lock.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/nargo_toml/src/flock.rs
+++ b/tooling/nargo_toml/src/flock.rs
@@ -1,0 +1,33 @@
+use fs2::FileExt;
+use std::{
+    fs::{File, OpenOptions},
+    path::Path,
+};
+
+// TODO: move this to some utils crate.
+
+pub(crate) struct FileLock {
+    file: File,
+}
+
+impl FileLock {
+    pub(crate) fn new(file_path: &Path, lock_name: &str) -> std::io::Result<Self> {
+        std::fs::create_dir_all(file_path.parent().expect("can't create lock on filesystem root"))?;
+        let file = OpenOptions::new().create(true).truncate(true).write(true).open(file_path)?;
+        if file.try_lock_exclusive().is_err() {
+            eprintln!("Waiting for lock on {lock_name}...");
+        }
+
+        file.lock_exclusive()?;
+
+        Ok(Self { file })
+    }
+}
+
+impl Drop for FileLock {
+    fn drop(&mut self) {
+        if let Err(e) = self.file.unlock() {
+            tracing::warn!("failed to release lock: {e:?}");
+        }
+    }
+}
--- a/tooling/nargo_toml/src/flock.rs
+++ b/tooling/nargo_toml/src/flock.rs
@@ -13,7 +13,7 @@ pub(crate) struct FileLock {
 impl FileLock {
     pub(crate) fn new(file_path: &Path, lock_name: &str) -> std::io::Result<Self> {
         std::fs::create_dir_all(file_path.parent().expect("can't create lock on filesystem root"))?;
-        let file = OpenOptions::new().create(true).truncate(true).write(true).open(file_path)?;
+        let file = OpenOptions::new().create(true).truncate(false).write(true).open(file_path)?;
         if file.try_lock_exclusive().is_err() {
             eprintln!("Waiting for lock on {lock_name}...");
         }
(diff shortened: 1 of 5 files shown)


tooling/nargo_toml/src/lib.rs
```diff
@@ -518,7 +519,10 @@ pub fn resolve_workspace_from_toml(
     current_compiler_version: Option<String>,
 ) -> Result<Workspace, ManifestError> {
     let nargo_toml = read_toml(toml_path)?;
+    let lock = lock_git_deps().expect("Failed to lock git dependencies cache");
     let workspace = toml_to_workspace(nargo_toml, package_selection)?;
+    lock.unlock().expect("Failed to unlock git dependencies cache");
```
aakoshh, line 524: I reckon it would be prudent to unlock the file before returning an error if `toml_to_workspace` fails.

TomAFrench: Ah good point. I'm being lazy here. Will fix

tooling/nargo_toml/src/lib.rs
```diff
@@ -518,7 +519,10 @@ pub fn resolve_workspace_from_toml(
     current_compiler_version: Option<String>,
 ) -> Result<Workspace, ManifestError> {
     let nargo_toml = read_toml(toml_path)?;
+    let lock = lock_git_deps().expect("Failed to lock git dependencies cache");
     let workspace = toml_to_workspace(nargo_toml, package_selection)?;
```
aakoshh, line 523: It would be nice to add some docs to `toml_to_workspace` and the stuff it calls, because it's not obvious that it can result in git dependencies being cloned.

tooling/nargo_toml/src/flock.rs
```diff
@@ -0,0 +1,33 @@
...
+pub(crate) struct FileLock {
+    file: File,
+}
+
+impl FileLock {
+    pub(crate) fn new(file_path: &Path, lock_name: &str) -> std::io::Result<Self> {
+        std::fs::create_dir_all(file_path.parent().expect("can't create lock on filesystem root"))?;
+        let file = OpenOptions::new().create(true).truncate(true).write(true).open(file_path)?;
+        if file.try_lock_exclusive().is_err() {
+            eprintln!("Waiting for lock on {lock_name}...");
```
aakoshh, line 18: I think printing the full path here would be fine, or as I mentioned just the name of the lock file.

tooling/nargo_toml/src/flock.rs
```diff
@@ -0,0 +1,33 @@
...
+// TODO: move this to some utils crate.
+
+pub(crate) struct FileLock {
+    file: File,
+}
+
+impl FileLock {
+    pub(crate) fn new(file_path: &Path, lock_name: &str) -> std::io::Result<Self> {
+        std::fs::create_dir_all(file_path.parent().expect("can't create lock on filesystem root"))?;
+        let file = OpenOptions::new().create(true).truncate(true).write(true).open(file_path)?;
```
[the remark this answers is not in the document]
aakoshh: Ostensibly there could even be two levels of locks:
1. Take a read lock 
2. Check if the dependency exists
3. If it exists unlock and return
4. If it doesn't exist, promote to write lock, download, unlock and return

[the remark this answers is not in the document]
TomAFrench: I've removed the truncation just in case someone uses this on a meaningful file. I think the other changes are overkill for the current situation and we should focus on allowing multiple processes to be able to work on the same workspace together cleanly before optimising.

Keep in mind that Cargo has a single lock on the packages cache also so this solution should also work for us.

tooling/nargo_toml/src/flock.rs
TomAFrench: I think it would be more user-friendly to have a custom name used here for when we want a directory lock.

aakoshh: Okay, I suppose at least very cryptic file names which only work in the full path context can be avoided this way.

aakoshh (approved):

tooling/nargo_toml/src/lib.rs
TomAFrench: I agree but I think that we should address the code issues here which are tightly coupling downloading and resolving of toml files.

aakoshh: You mean we should restrict ourselves to just the code, as opposed to improving docs? Fair enough. 

I'm not sure we have addressed any tight coupling though.

PR #7329, opened 2025-02-09
chore: remove misleading output from `nargo check`

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

The output of `Constraint system successfully built` is misleading as `nargo check` doesn't build any sort of constraint system and instead is just doing type checking.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/nargo_cli/src/cli/check_cmd.rs
+++ b/tooling/nargo_cli/src/cli/check_cmd.rs
@@ -66,16 +66,13 @@ pub(crate) fn run(args: CheckCommand, config: NargoConfig) -> Result<(), CliErro
             continue;
         }
 
-        let any_file_written = check_package(
+        check_package(
             &workspace_file_manager,
             &parsed_files,
             package,
             &args.compile_options,
             args.allow_overwrite,
         )?;
-        if any_file_written {
-            println!("[{}] Constraint system successfully built!", package.name);
-        }
     }
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


TomAFrench:
I've updated the documentation to remove references to constraint systems.

tooling/nargo_cli/src/cli/check_cmd.rs
```diff
@@ -110,9 +107,7 @@ fn check_package(
                 eprintln!("Note: Prover.toml already exists. Use --overwrite to force overwrite.");
             }
 
-            let any_file_written = should_write_prover;
-
-            Ok(any_file_written)
+            Ok(())
```
aakoshh, line 110: Based on https://github.com/noir-lang/noir/pull/7170/files#diff-fe0ef4227fae4e8c5894bbc6bdc9ddaa5b43a5f8a7022e75244b525f817de47c I thought this was somehow an indication for the fact that the `Prover.toml` file has been created. Should there be a dedicate feedback about it in `check_package` instead, now that `any_file_written` is no longer returned?

TomAFrench: I don't think that we need to display any feedback for this file being created tbh.

aakoshh: Okay. I wasn't sure if it was common knowledge that `check` creates this template, and thought it would be good to give the user a hint in case they were wondering where a new file came from, but I suppose they will know once it starts warning them about the `--overwrite` flag.

aakoshh (approved):

Issue #6965, opened 2025-01-06
Deduplicate compiler warnings based on their source location


TomAFrench:
After #6926 we still get a number of cases where multiple instances of the same compiler warning/error is emitted pointing at the same line of source code. This tends to just make the output harder to read for very little benefit so I'm inclined to deduplicate these so that we only print them once.

jfecher:
What are some examples of this?

TomAFrench:
I think the case where this comes up most frequently is on comptime functions (specifically attributes being applied to functions). If you run `nargo check` on aztec-nr then you should see it. For instance I get the below as part of the warnings emitted.


```
warning: index is private and not visible from the current module
   ┌─ value-note/src/value_note.nr:21:1
   │
21 │ #[note]
   │ ------- index is private
   │

warning: offset is private and not visible from the current module
   ┌─ value-note/src/value_note.nr:21:1
   │
21 │ #[note]
   │ ------- offset is private
   │

warning: length is private and not visible from the current module
   ┌─ value-note/src/value_note.nr:21:1
   │
21 │ #[note]
   │ ------- length is private
   │

warning: index is private and not visible from the current module
   ┌─ value-note/src/value_note.nr:21:1
   │
21 │ #[note]
   │ ------- index is private
   │

warning: offset is private and not visible from the current module
   ┌─ value-note/src/value_note.nr:21:1
   │
21 │ #[note]
   │ ------- offset is private
   │

warning: length is private and not visible from the current module
   ┌─ value-note/src/value_note.nr:21:1
   │
21 │ #[note]
   │ ------- length is private
   │

warning: index is private and not visible from the current module
   ┌─ value-note/src/value_note.nr:21:1
   │
21 │ #[note]
   │ ------- index is private
   │

warning: offset is private and not visible from the current module
   ┌─ value-note/src/value_note.nr:21:1
   │
21 │ #[note]
   │ ------- offset is private
   │

warning: length is private and not visible from the current module
   ┌─ value-note/src/value_note.nr:21:1
   │
21 │ #[note]
   │ ------- length is private
   │

warning: trait `traits::Empty` which provides `empty` is implemented but not in scope, please import it
   ┌─ value-note/src/value_note.nr:21:1
   │
21 │ #[note]
   │ -------
   │

warning: trait `note::note_interface::NoteInterface` which provides `get_note_type_id` is implemented but not in scope, please import it
   ┌─ value-note/src/value_note.nr:21:1
   │
21 │ #[note]
[...]

PR #7333, opened 2025-02-10
feat(experimental): show macro errors where they happen

### LINKED ISSUE #6013: Location information is lost after macro expansion
### Aim

Given this code:

```noir
use std::meta::unquote;

fn main() {
    comptime
    {
        let q = quote { let _x: i32 = "hello"; };
        unquote!(q);
    }
}
```

We get this error:

```
error: Expected type i32, found type str<5>
   ┌─ src/main.nr:92:9
   │
92 │         unquote!(q);
   │         -----------
```

Like that, it's impossible to understand how to fix the error.

### Expected Behavior

Ideally the error points at exactly the "source location" of the error. I'm not sure how we can do that. But for example consider macros in Crystal:

```crystal
macro my_macro(name)
    {{name}} = 1
    {{name}}.length()
end

my_macro(x)
```

When we call `my_macro` that generates this code:

```crystal
name = 1
name.length()
```

The error Crystal gives is this:

```
There was a problem expanding macro 'my_macro'

Code in foo.cr:6:1

 6 | my_macro(x)
     ^
Called macro defined in foo.cr:1:1

 1 | macro my_macro(name)

Which expanded to:

 > 1 | x = 1
 > 2 | x.length()
         ^-----
Error: undefined method 'length' for Int32
```

The way Crystal does this is by associating the expanded macro location to a virtual location (and storing the virtual source). Then when an error happens in a virtual location its get the virtual source, shows it and shows the error.

I don't know if we could do the same, but it's just an idea.

### Bug

The error cannot be understood without being able to see the generated source code.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem\*

Related to #6965
Resolves #6013

## Summary

This PR makes macro errors show up in the code that produces the error. For example:

```noir
#[foo] // Error used to be notified here (and still is, kind of)
comptime fn foo(_: FunctionDefinition) -> Quoted {
    quote {
        fn bar() {
            1 + "a"; // Error is notified here
        }
    }
}
```

For that to work several changes were needed:
1. `Value::Quoted` in comptime code used to be a list of bare tokens, without spans. The first change was to change that to `SpannedToken`. With that, the snippet above gives the error in the "correct" location.
2. But what if the quoted value is in another file? Just spans aren't enough: every token needs to hold a `Location`, and that `Location` needs to be propagated to AST nodes. That's what the rest of the PR does.

For example in a program like this:

```noir
// main.nr
mod other;
use other::expr;

#[foo]
comptime fn foo(_: FunctionDefinition) -> Quoted {
    let expr = expr();
    quote {
        fn bar() {
            $expr;
        }
    }
}

// other.nr
pub comptime fn expr() -> Quote {
    quote { 1 + "a" }
}
```

The resulting program will have a function `bar` located at "main.nr", with an expression that's located in "other.nr". When the error is found, it's found in "other.nr".

Going back to the first snippet:

```noir
#[foo]
comptime fn foo(_: FunctionDefinition) -> Quoted {
    quote {
        fn bar() {
            1 + "a";
        }
    }
}
```

This is what the error looks like now:

image

That is:
- the error primarily shows up in the macro source code that triggered it (which hopefully is understandable)
- the error mentions that it happened while running a function attribute

This solves the concern about only showing the error where it happens but without that being clear that it was because of a macro call or similarly.

To implement this, the Elaborator keeps a stack of "elaborate reasons": before elaborating items produced by a macro we record that we are elaborating because of this macro (and track that location). Then, if an error happens, it's wrapped so that it also mentions that location.

## Additional Context

This PR is massive, but most of the diff is:
1. Changing `Span` to `Location` in **a lot** of places
2. A new file, `with_file.rs`, whose task is to transform a `ParsedModule` into another one where all the `FileID`s that exist in it are replaced with a different `FileID` (this is so that LSP parser cache can still be used, which is much faster than not caching)

There's another thing I noticed. In this snippet:

```noir
mod other;

fn main() {
    comptime {
        let expr = quote { 1 + "a" }.as_expr().unwrap();
        let resolved = expr.resolve(Option::none());
    }
}
```

We get this error:

```
error: Types in a binary operation should match, but found Field and str<1>
  ┌─ src/main.nr:6:24
  │
6 │         let resolved = expr.resolve(Option::none());
  │                        ----
```

Without going back to the code and checking what's `expr` it's hard to understand the error. I think the error should be (like the rest of this PR) first in `1 + "a"` then a mention that it happened while doing `expr.resolve`. The issue is that comptime `Value`s that wrap expressions and statements only hold an `ExpressionKind` or `StatementKind`, without a `Location`. To improve this we could change that to storing `Expression` and `Statement`, and using those locations in errors. But I'll leave that to a separate PR because this was is already huge.

## Documentation

Check one:

# PR Checklist


TomAFrench:
Hmm those regressions are interesting. Maybe false positives as protocol circuits don't rely on macros much.

(Haven't reviewed this PR deeply)

asterite:
One explanation could be that many nodes, and tokens, now carry `Location` inside of a `Span`, which is one `usize` bigger. But for example `Expression` is 320 bytes, so it would be strange if a minimal increase in memory would have that impact in compilation speed. I still have some more commits to push here so we'll see...

compiler/noirc_frontend/src/ast/visitor.rs
```diff
@@ -1339,73 +1342,84 @@ impl UnresolvedType {
...
-            UnresolvedTypeData::FieldElement => visitor.visit_field_element_type(self.span),
+            UnresolvedTypeData::String(expr) => visitor.visit_string_type(expr, self.location.span),
+            UnresolvedTypeData::Unspecified => visitor.visit_unspecified_type(self.location.span),
+            UnresolvedTypeData::Quoted(typ) => visitor.visit_quoted_type(typ, self.location.span),
+            UnresolvedTypeData::FieldElement => {
+                visitor.visit_field_element_type(self.location.span);
+            }
             UnresolvedTypeData::Integer(signdness, size) => {
-                visitor.visit_integer_type(*signdness, *size, self.span);
+                visitor.visit_integer_type(*signdness, *size, self.location.span);
```
michaeljklein, line 1412: typo: `signdness` -> `signedness`
```suggestion
                visitor.visit_integer_type(*signedness, *size, self.location.span);
```

compiler/noirc_frontend/src/elaborator/comptime.rs
```diff
@@ -653,4 +665,41 @@ impl<'context> Elaborator<'context> {
                 _ => false,
             }
     }
+
+    /// Pushes an ElaborateReason but also takes the current errors and returns them.
+    pub(crate) fn push_elaborate_reason(
```
michaeljklein, line 670: Please rename to make it clear that this resets `self.errors`

compiler/noirc_frontend/src/elaborator/enums.rs
```diff
@@ -518,7 +526,7 @@ impl Elaborator<'_> {
     /// which is an implementation of https://julesjacobs.com/notes/patternmatching/patternmatching.pdf
     pub(super) fn elaborate_match_rows(&mut self, rows: Vec<Row>) -> HirMatch {
         self.compile_rows(rows).unwrap_or_else(|error| {
-            self.push_err(error);
+            self.push_err(error, self.file); // TODO: fix this
```
michaeljklein, line 529: Please make a follow-up issue for this TODO

compiler/noirc_frontend/src/lexer/token.rs
```diff
@@ -320,7 +321,7 @@ pub fn token_to_borrowed_token(token: &Token) -> BorrowedToken<'_> {
 #[derive(Clone, PartialEq, Eq, Hash, Debug, PartialOrd, Ord)]
 pub enum FmtStrFragment {
     String(String),
-    Interpolation(String, Span),
+    Interpolation(String, Span, FileId),
```
michaeljklein, line 324: Why not directly store a `Location` instead of separate `Span` and `FileId`?

compiler/noirc_frontend/src/elaborator/trait_impls.rs
```diff
@@ -59,7 +62,9 @@ impl<'context> Elaborator<'context> {
 
                     let func_id = self.interner.push_empty_fn();
                     let module = self.module_id();
-                    let location = Location::new(default_impl.def.span, trait_impl.file_id);
+                    // TODO: check this, it doesn't make sense to mix locations from two different files
```
michaeljklein, line 65: Please make a follow-up issue for this TODO

compiler/noirc_frontend/src/elaborator/types.rs
```diff
@@ -1096,9 +1156,11 @@ impl<'context> Elaborator<'context> {
         lhs_type: &Type,
         op: &HirBinaryOp,
         rhs_type: &Type,
-        span: Span,
+        location: Location,
     ) -> bool {
-        self.unify(lhs_type, rhs_type, || TypeCheckError::TypeMismatchWithSource {
+        let span = location.span;
+
+        self.unify(lhs_type, rhs_type, location.file, || TypeCheckError::TypeMismatchWithSource {
```
michaeljklein, line 1163: Definitely something to leave for a follow-up issue, but I think that we should try to move all these `file`'s to the error enums so that `push_err` and `unify` don't need to keep track of them.

compiler/noirc_frontend/src/lexer/token.rs
```diff
@@ -352,6 +353,63 @@ pub enum DocStyle {
...
+    fn from(spt: &'a LocatedToken) -> Self {
+        &spt.0.contents
+    }
+}
+
+impl LocatedToken {
+    pub fn new(token: Token, location: Location) -> LocatedToken {
+        LocatedToken(Located::from(location, token))
+    }
+    pub fn to_location(&self) -> Location {
```
michaeljklein, line 386: Why `to_location` and not `location` as used elsewhere? Same for `to_span`

compiler/noirc_frontend/src/parser/errors.rs
```diff
@@ -258,46 +270,46 @@ impl<'a> From<&'a ParserError> for Diagnostic {
                             .collect::<Vec<_>>()
                             .join(", ")
                     ),
-                    error.span,
+                    error.span(),
                 ),
                 ParserErrorReason::ExperimentalFeature(_) => {
-                    Diagnostic::simple_warning(reason.to_string(), "".into(), error.span)
+                    Diagnostic::simple_warning(reason.to_string(), "".into(), error.span())
```
michaeljklein, line 276: Would it make sense for `Diagnostic::simple_warning` to directly take a `Location`, to reduce the amount of `Location -> Span` conversions?

compiler/noirc_frontend/src/parser/parser.rs
```diff
@@ -479,17 +491,35 @@ impl<'a> Parser<'a> {
...
             start_span
         } else {
-            let end_span = self.previous_token_span;
-            Span::from(start_span.start()..end_span.end())
-        }
+            let end_span = self.previous_token_location.span;
+            if start_span.start() <= end_span.end() {
+                Span::from(start_span.start()..end_span.end())
+            } else {
+                // TODO: workaround for now
```
michaeljklein, line 509: Please add a follow-up issue for this TODO

michaeljklein (changes requested):
This looks great! I have a handful of small comments, but it looks good overall.

I know we don't test error messages much, but I came up with a way to test this feature that should be stable, even if we update the error messages:

1. Pick a key location where errors could occur
2. Use a macro to generate a definition that uses a unique variable/parameter/struct-field name where the error occurs, e.g.
```noir
fn foo() {
    ..
    let unique_var_name = ERROR_CAUSE_HERE;
    ..
}
```
3. In the test, we can ensure that the error references the correct line by asserting that the `error_str.contains("unique_var_name")`.

Please add a few tests or make a follow up issue to test this.

compiler/noirc_frontend/src/lexer/token.rs
asterite: I forgot why so I tried it again and remembered: `Location` isn't `PartialOrd`/`Ord` and I wasn't sure if this was intentional or not. If there's no issue with that I can make the change.

compiler/noirc_frontend/src/lexer/token.rs
asterite: Oh, it was called `to_span` in `SpannedToken` so I called it `to_location` in `LocatedToken`, just to follow a tradition. I'll change the names.

compiler/noirc_frontend/src/elaborator/types.rs
asterite: Definitely! It's something I wanted to do, and I actually considered doing as part of this PR, but I found that not doing it led to a smaller PR. I captured https://github.com/noir-lang/noir/issues/7452 for this.

compiler/noirc_frontend/src/lexer/token.rs
asterite: Actually, let me know if you'd prefer me to do this change now or in a follow-up PR.

compiler/noirc_frontend/src/parser/errors.rs
asterite: I can try it. I tried to make the minimal changes in this PR to get it working, mainly because it's already huge. Let me know if you'd prefer me  to do it in this PR or a follow-up.

compiler/noirc_frontend/src/parser/parser.rs
asterite: Captured in https://github.com/noir-lang/noir/issues/7453

compiler/noirc_frontend/src/elaborator/trait_impls.rs
asterite: Done: https://github.com/noir-lang/noir/issues/7454

compiler/noirc_frontend/src/elaborator/enums.rs
asterite: I forgot about this, thanks! It's fixed now.

compiler/noirc_frontend/src/elaborator/comptime.rs
asterite: Done!

compiler/noirc_frontend/src/ast/visitor.rs
asterite: Done, thanks!

asterite:
I made a follow-up issue to test error locations: https://github.com/noir-lang/noir/issues/7455

compiler/noirc_frontend/src/lexer/token.rs
jfecher: I think it was just never added because it wasn't needed before. Should be fine to add, although I'll add that it'd make more sense when comparing locations that the FileId is compared before the Span.

compiler/noirc_frontend/src/elaborator/types.rs
asterite: Also to clarify: I started doing that but before that I had almost one thousand errors and that change brought it to 1K+ so the number of errors was unknown (at least in VSCode) and it demotivated me a bit to continue without seeing the number of errors going down. Then I realized requiring a file in `push_err` would work "the same way" so I took that approach.

compiler/noirc_errors/src/position.rs
```diff
@@ -8,6 +8,61 @@ use std::{
...
 pub type Position = u32;
 
+#[derive(Eq, Debug, Clone)]
+pub struct Located<T> {
+    pub contents: T,
+    location: Location,
+}
+
+/// This is important for tests. Two Located objects are equal if their content is equal
+/// They may not have the same location. Use into_location to test for Location being equal specifically
```
jfecher, line 18:
```suggestion
/// They may not have the same location. Use `.location()` to test for Location being equal specifically
```

compiler/noirc_errors/src/position.rs
```diff
@@ -8,6 +8,61 @@ use std::{
...
+impl<T: Hash> Hash for Located<T> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.contents.hash(state);
+    }
+}
+
+impl<T> Located<T> {
+    pub const fn from(location: Location, contents: T) -> Located<T> {
+        Located { location, contents }
+    }
```
jfecher, line 55: Do we require this to be const currently? Or is there an advantage in it being const if we're not using it in a const context?

jfecher (commented):
LGTM. I only got through the first 50 files or so though

compiler/noirc_frontend/src/lexer/token.rs
asterite: Done!

compiler/noirc_frontend/src/lexer/token.rs
asterite: Done!

compiler/noirc_frontend/src/elaborator/comptime.rs
```diff
@@ -552,20 +559,27 @@ impl<'context> Elaborator<'context> {
...
 
             if !generated_items.is_empty() {
+                let previous_errors = self.push_elaborate_reason_and_take_errors(
+                    ElaborateReason::RunningAttribute,
+                    location,
+                );
+
                 self.elaborate_items(generated_items);
+
+                self.pop_elaborate_reason(previous_errors);
```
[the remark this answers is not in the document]
asterite: It's also used in another part of the code when adding items: 

https://github.com/noir-lang/noir/pull/7333/files#diff-5fed0da906822475da14cdd715d74976cd54c2593372f292d7f5d7772bb5e76fR356-R363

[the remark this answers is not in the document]
jfecher: I was thinking the method calls could be inlined into that function, not that the method calls weren't needed anywhere else. I was thinking the other calls like `elaborate_item_from_function` should also be included for similar errors.

Thinking about it some more though, the other elaborate_item functions like `elaborate_item_from_function` are called when we e.g. call a macro from the interpreter. Not sure if we want to include that in the error trace - possibly? I think this is fine for now but it could be another avenue to explore in the future.

compiler/noirc_errors/src/position.rs
jfecher: Sure, lets remove it. It probably doesn't do anything but when I see it it makes me think "wait, do we use this in a const context anywhere?"

jfecher (approved):
LGTM

michaeljklein (approved):

Aristotelis2002:
This PR seems to lead to an unexpected behavior in `nargo debug`: 
image
Note the step in the seemingly random file. Program is:
```rust
fn bar(y: Field) {
    assert(y != 0);
}
fn foo(x: Field) {
    bar(x)
}
    foo(x);
    foo(y);
    assert(x != y);
}
```

smanilov:
> This PR seems to lead to an unexpected behavior in `nargo debug`: image Note the step in the seemingly random file. Program is:
> 
> ```rust
> fn bar(y: Field) {
>     assert(y != 0);
> }
> fn foo(x: Field) {
>     bar(x)
> }
> fn main(x: Field, y: pub Field) {
>     foo(x);
>     foo(y);
>     assert(x != y);
> }
> ```

Filed separately as https://github.com/noir-lang/noir/issues/7479.

PR #7334, opened 2025-02-10
fix: perform SSA constraints check on final SSA

### LINKED ISSUE #7292: "Separate subgraph" bug misfires on indirect Brillig call with return value
### Aim

Attempted to compile a program that calls an unconstrained function and asserts that the result is zero:

```noir
unconstrained fn foo<T>(_x: T) -> Field {
    0
}

fn bar<T>(x: T) -> Field {
    /// Safety: test
    unsafe { foo(x) }
}

fn main() {
    let x: Field = 0;
    assert(bar(x) == 0);
}
```

### Expected Behavior

Expected the program to compile without errors or `bug`'s.

### Bug

`nargo compile` returns:

```bash
bug: Input to Brillig function is in a separate subgraph to output
  ┌─ src/main.nr:7:14
  │
7 │     unsafe { foo(x) }
  │              ------ There is no path from the output of this Brillig call to either return values or inputs of the circuit, which creates an independent subgraph. This is quite likely a soundness vulnerability
  │
  = Call stack:
    1. src/main.nr:12:12
    2. src/main.nr:7:14
```

However, the Brillig call is completely constrained:
- All arguments to `foo` are constrained (i.e. constant)
- `foo`'s result is constrained to be zero

### To Reproduce

1.
2.
3.
4.

### Workaround

Yes

### Workaround Description

Ignore `bug` warning

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.1 noirc version = 1.0.0-beta.1+3a42eb5c68f9616f0ebe367c894f0376ba41e0ef (git version hash: 3a42eb5c68f9616f0ebe367c894f0376ba41e0ef, is dirty: false)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

TomAFrench:
# Description

## Problem\*

Resolves #7292 

## Summary\*

We're currently performing the safety checks on the SSA before we do constant folding with brillig calls. This prevents bugs being shown for unconstrained calls which are optimized out. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa.rs
+++ b/compiler/noirc_evaluator/src/ssa.rs
@@ -104,6 +104,28 @@ pub(crate) fn optimize_into_acir(
 
     let mut ssa_level_warnings = vec![];
 
+    drop(ssa_gen_span_guard);
+
+    let used_globals_map = std::mem::take(&mut ssa.used_globals);
+    let brillig = time("SSA to Brillig", options.print_codegen_timings, || {
+        ssa.to_brillig_with_globals(options.enable_brillig_logging, used_globals_map)
+    });
+
+    let ssa_gen_span = span!(Level::TRACE, "ssa_generation");
+    let ssa_gen_span_guard = ssa_gen_span.enter();
+
+    let mut ssa = SsaBuilder {
+        ssa,
+        ssa_logging: options.ssa_logging.clone(),
+        print_codegen_timings: options.print_codegen_timings,
+    }
+    .run_pass(|ssa| ssa.fold_constants_with_brillig(&brillig), "Inlining Brillig Calls Inlining")
+    // It could happen that we inlined all calls to a given brillig function.
+    // In that case it's unused so we can remove it. This is what we check next.
+    .run_pass(Ssa::remove_unreachable_functions, "Removing Unreachable Functions (3rd)")
+    .run_pass(Ssa::dead_instruction_elimination, "Dead Instruction Elimination (2nd)")
+    .finish();
+
     if !options.skip_underconstrained_check {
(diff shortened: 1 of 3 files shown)


compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -130,7 +114,7 @@ pub(crate) fn optimize_into_acir(
     let ssa_gen_span = span!(Level::TRACE, "ssa_generation");
     let ssa_gen_span_guard = ssa_gen_span.enter();
 
-    let ssa = SsaBuilder {
+    let mut ssa = SsaBuilder {
```
aakoshh, line 117: Strange that using `&mut` somehow makes the `.par_bridge` used in the checks work, even though it shouldn't actually need mutation.

aakoshh (approved):
Just want to clarify the comments you left:

>  This prevents bugs being shown for unconstrained calls which are optimized out.

It sounds like you want to show bugs for calls which have been optimised away, which is what happened before this PR, so I assume you mean the opposite, perhaps "this PR prevents" or "This allows bugs...". 

I suppose there is no way to tell when exactly a function is removed, so running the checks on the final state is the right thing to do, which will only show bugs that are still present.

TomAFrench:
Yeah, I mean "this PR" here

compiler/noirc_evaluator/src/ssa.rs
TomAFrench: Yeah it's some interior mutability issue which having a mutable reference allows us to sidestep.

PR #7344, opened 2025-02-10
feat(performance): Use unchecked ops based upon known induction variables

### LINKED ISSUE #7343: Optimization: Check whether binary ops can be unchecked based upon the induction variable
### Problem

When performing binary operations we default to checked operations. However, when we know we have a known upper loop bound we can determine whether there will be an overflow at compilation time.
For example:
```noir
fn main(six_as_u32: u32) {
    for i in 0..64 {
        let z: u32 = i * 64;
        assert(z != six_as_u32);
    }
}
``` 
We know the maximum loop bound is going to be 63 and the 63*64 will fit into a `u32`. This operation should be unchecked.

However, let's look at the output from `noir-profiler execution-opcodes`:
Image

45.3% (448 samples) of the executed opcodes comes from `i * 64`, while only 6.47% (64 samples as expected) are actual Brillig operations. The rest of the operations for `i * 64` are from overflow checks.

### Happy Case

Get rid of unnecessary overflow checks when we know from the induction variable that the operation will never overflow. 
e.g. The above example should only see 64 opcode samples from `i * 64`. 

This PR https://github.com/noir-lang/noir/pull/6910 is of similar style but only checks induction variables from outer loops and hoists binary ops into the outer loop based upon the induction variable upper bound.

### Workaround

None

### Workaround Description

There is no ability to workaround this at the moment, this requires compiler changes. 

### Additional Context

_No response_

### Project Impact

Blocker

### Blocker Context

I am marking this as a blocker as improving unconstrained execution is of general benefit to all Noir devs and this this specific optimization is only possible through compile changesr.

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #7343 

## Summary\*

In a similar theme to https://github.com/noir-lang/noir/pull/6910 we can check whether a given binary op will overflow when we have a constant and a known upper loop bound. In this PR, we now check every binary add/mul for whether it can be converted to an unchecked operation.  

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -697,4 +747,53 @@ mod test {
         let ssa = ssa.loop_invariant_code_motion();
         assert_normalized_ssa_equals(ssa, expected);
     }
+
+    #[test]
+    fn transform_safe_ops_to_unchecked_during_code_motion() {
+        // This test is identical to `simple_loop_invariant_code_motion`, except this test
+        // has an induction variable with a known upper loop bound.
```
vezenovm, line 754:
```suggestion
        // uses a checked add in `b3`.
```

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -87,7 +89,8 @@ struct LoopInvariantContext<'f> {
     inserter: FunctionInserter<'f>,
     defined_in_loop: HashSet<ValueId>,
     loop_invariants: HashSet<ValueId>,
-    // Maps induction variable -> fixed upper loop bound
+    current_induction_variables: HashMap<ValueId, FieldElement>,
```
vezenovm, line 92:
```suggestion
    // Maps current loop induction variable -> fixed upper loop bound
    current_induction_variables: HashMap<ValueId, FieldElement>,
```

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -236,34 +253,68 @@ impl<'f> LoopInvariantContext<'f> {
...
-            _ => false,
         }
     }
 
+    fn eval_binary_op_with_induction_vars(
+        &self,
+        binary: &Binary,
+        induction_vars: &HashMap<ValueId, FieldElement>,
+    ) -> bool {
+        if !matches!(binary.operator, BinaryOp::Add { .. } | BinaryOp::Mul { .. }) {
```
asterite, line 296: Out of curiosity: is there a reason why `Sub` isn't here? (I know it was like that before this PR )

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -236,34 +253,68 @@ impl<'f> LoopInvariantContext<'f> {
...
+    /// Binary operations can take advantage of that our induction variable has a fixed maximum,
+    /// to be transformed from a checked operation to an unchecked operation.
+    /// Checked operations require more bytecode and thus we aim to minimize their usage wherever possible.
+    ///
+    /// If one side of a binary operation is a constant and the other is an induction variable
+    /// with a known upper bound, we know whether that binary operation will ever overflow.
+    /// If we determine that an overflow is not possible we can convert the checked operation to unchecked.
+    fn transform_to_unchecked_from_upper_bound(&mut self, instruction_id: InstructionId) {
+        if let Instruction::Binary(binary) = &self.inserter.function.dfg[instruction_id] {
+            let new_instruction = if !binary.operator.is_unchecked()
```
asterite, line 271: (optional) Just a tiny thing, but when I saw `!binary.operator.is_unchecked()` I was wondering if that would also include operations that don't have the `unchecked` bool at all. Looking at `fn is_unchecked()` they won't because that returns `true` and `!true` is `false`, but I wonder if having an `fn is_checked()` method instead would make this a bit clearer. Also because this is the only occurrence of this method call and it's negated.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -697,4 +748,53 @@ mod test {
...
+        }
+        ";
+
+        let ssa = Ssa::from_str(src).unwrap();
+        let main = ssa.main();
+
+        let instructions = main.dfg[main.entry_block()].instructions();
+        assert_eq!(instructions.len(), 0); // The final return is not counted
+
+        // `v8 = add v2, i32 1` in b3 should now be `v9 = unchecked_add v2, i32 1` in b3
```
asterite, line 779: The SSA parser supports comments (and trailing comments) so, if it's clearer, you could put this comment right next to the relevant line (or not! up to you)

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -216,6 +215,24 @@ impl<'f> LoopInvariantContext<'f> {
...
+    fn set_induction_var_bounds(&mut self, loop_: &Loop, current_loop: bool) {
+        let upper_bound = loop_.get_const_upper_bound(self.inserter.function);
+        if let Some(upper_bound) = upper_bound {
+            let induction_variable = loop_.get_induction_variable(self.inserter.function);
+            let induction_variable = self.inserter.resolve(induction_variable);
+            if current_loop {
+                self.current_induction_variables.insert(induction_variable, upper_bound);
+            } else {
+                self.outer_induction_variables.insert(induction_variable, upper_bound);
+            }
```
TomAFrench, line 232: Seems like these two maps will only differ by the current loop's induction variable (we never remove anything from `current_induction_variables`).

We could then simplify this to just have `induction_variables` and then track the current induction variable if we need to special case something.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -236,34 +253,68 @@ impl<'f> LoopInvariantContext<'f> {
...
+            rhs_const,
+            induction_vars.get(&binary.lhs),
+            induction_vars.get(&binary.rhs),
+        ) {
+            (Some((lhs, _)), None, None, Some(upper_bound)) => (lhs, *upper_bound),
+            (None, Some((rhs, _)), Some(upper_bound), None) => (*upper_bound, rhs),
+            _ => return false,
+        };
+
+        eval_constant_binary_op(lhs, rhs, binary.operator, operand_type).is_some()
```
TomAFrench, line 315: This could do with some documentation on what this call tells us, i.e. 

> We evaluate this expression using the upper bounds of its inputs to check whether it will ever overflow. If so, this will cause `eval_constant_binary_op` to return `None`. Therefore a `Some` value shows that this operation is safe.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -236,34 +253,68 @@ impl<'f> LoopInvariantContext<'f> {
...
 
-                eval_constant_binary_op(lhs, rhs, binary.operator, operand_type).is_some()
+            if let Some(new_instruction) = new_instruction {
+                self.inserter.function.dfg[instruction_id] = new_instruction;
             }
-            _ => false,
         }
     }
 
+    fn eval_binary_op_with_induction_vars(
```
TomAFrench, line 291: I think this naming is a bit weird as it's telling us how it does what it does, rather than _what_ it's doing.

Maybe `induction_var_bounds_prevent_overflow`?

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
TomAFrench, line 271: We can remove a bunch of nesting with early returns here.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: It is just out of simplicity for now. Will add `Sub` in a follow-up. I have made an issue here https://github.com/noir-lang/noir/issues/7353.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: I preferred the comment here as it is colored with green text rather than the same text of the SSA.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: done, the method is much simpler now

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: I went with `can_evaluate_binary_op` as I felt it was simpler. I left a doc comment as well. Let me know what you think.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: done

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: I ended up switching how this condition is checked so we now early exit the method on `binary.operator.is_unchecked() || !self.can_evaluate_binary_op(binary, &self.current_induction_variables)`

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: Yeah I had thought about combining them as well, but was unsure how much it would simplify things. Tracking the current induction var would require also special casing this when checking whether we can evaluate a binary operation for hoisting rather than simply passing `self.outer_induction_variables`. I ultimately just went with an extra map as there shouldn't be very many induction variables. If we feel one `induction_variables` map would be simpler though I can switch to this.

TomAFrench: I think at the very least we want to have some documentation near the definitions of these two maps making this relationship between them very explicit and why we've chosen to do this. Merging them can be left as a TODO.

TomAFrench (approved):
LGTM

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
TomAFrench: Eh :shrug: , that's fine in combo with the comment.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: I switched to clearing the `current_induction_variables` map for every loop. And have added some comments.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -236,34 +253,70 @@ impl<'f> LoopInvariantContext<'f> {
...
+
+        if binary.operator.is_unchecked()
+            || !self.can_evaluate_binary_op(binary, &self.current_induction_variables)
+        {
+            return;
+        }
+
+        if let Instruction::Binary(binary) = &mut self.inserter.function.dfg[instruction_id] {
+            binary.operator = binary.operator.into_unchecked();
+        };
```
TomAFrench, line 283:
```suggestion
        binary.operator = binary.operator.into_unchecked();
```

This check looks duplicated.

TomAFrench: Ah no, you have different mutabilities. Can we not make the first `binary` a mutable reference?

vezenovm: `self.can_evaluate_binary_op` borrows immutably so we cannot.

Issue #7335, opened 2025-02-10
Improve usage of file locks to allow better parallel use of nargo


TomAFrench:
We've currently got a bit of a sledgehammer approach to file locking where ony a single instance of nargo can access a workspace at any given time. This is causing some pain for users of nargo who want to build a workspace in parallel and kick off processing of the artifacts without waiting for the full workspace to build.

We should then restrict the file locking to cover only the directories which multiple nargo instances may clash on. These are the git dependency cache and the target directory afaik. I think a sensible plan is then:

```[tasklist]
### Tasks
- [ ] https://github.com/noir-lang/noir/pull/7327 
- [ ] https://github.com/noir-lang/noir/issues/7542
- [ ] https://github.com/noir-lang/noir/issues/7349
- [ ] Remove Nargo.toml lock
```

aakoshh:
Regarding these two:
* Add file lock on target directory to prevent multiple commands using the same output dir concurrently.
* Remove Nargo.toml lock

I did it differently in #7345: I still take out locks on the `Nargo.toml` file, but it's the package specific version, rather than the workspace one. This is because if we use the workspace level output directory for all members, then locking that directory itself would again prevent parallelism, unless we lock some package specific output (e.g. the JSON artefact, the witness file or a special marker); I thought locking the already existing `Nargo.toml` in the input directory should be sufficient. 

If we had the ability to specify the target directory, then we would still face some challenges:
* If they user doesn't override it, then we are still writing to the same workspace directory, so we can't just lock it if we want to support using `parallel` in the shell to speed things up. 
* If the user overrides it but only to a member specific directory, then we still have to worry about using different compilation options for the same package, which is what caused the issue with the tests in the first place: while we were executing with e.g. inliner aggressiveness 0 and forcing brillig, another test was overwriting the artefacts with a different aggressiveness. We would need to lock some compilation flag specific lock file, and save separate artefacts for each configuration.

Issue #7340, opened 2025-02-10
LSP formatter always uses the default configuration


asterite:
### Aim

If there's a `noirfmt.toml` file in a package or its workspace, that configuration should be picked up when a file is formatted via the LSP.

### Expected Behavior

`noirfmt.toml` is picked up from the package the file is in or, if it doesn't exist, from the parent workspace.

### Bug

The default formatter configuration is currently used, making it impossible to customize formatting while also using an IDE.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #7345, opened 2025-02-11
fix(cli): Only lock the packages selected in the workspace

aakoshh:
# Description

## Problem\*

Related to #7335 

## Summary\*

Changes the `nargo` CLI to take out individual locks on the packages affected by the commands, rather than the workspace level `Nargo.toml` file, so that we can execute commands in parallel. 

Also fixes the issue where the `target` changed depending on whether we use `--package` to select a member in the workspace or whether we `cd` into the directory of the member itself. We always want it to be the workspace level target. 

## Additional Context

https://github.com/noir-lang/noir/pull/6752 changed the CLI to look for the first package on the ancestor path when neither `--workspace` nor `--package` options were present, and treat that as the `Workspace`, which resulted in the `target` directory being different in the two cases. Now we always look for the top level `Nargo.toml`, ie. the workspace, so that we use the correct `target`. We look for both the top and bottom level workspaces; if they differ, then we consider the bottom one as the selected member.

https://github.com/noir-lang/noir/pull/6941 moved the locks from the integration tests into `nargo` itself, where it was easier to manage, however it interfered with the CI scripts using `parallel` to run CLI commands concurrently because the lock was taken out on the entire workspace.

### Testing

Here's a session demonstrating that indeed on `master` if we execute a member the artefacts are saved in its own `target` directory, whereas on this branch they go into the workspace level one (note how different packages write to different output files in the common directory, so they don't conflict):

```console
$ cd test_programs/execution_success/workspace
$ ls
Nargo.toml  crates/     target/
$ ls crates
a/ b/
$ rm -rf target
$ cd crates/a 
$ ls
Nargo.toml   Prover.toml  src/
$ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
$ cargo run -q -p nargo_cli -- --program-dir . execute 
[a] Circuit witness successfully solved
[a] Witness saved to /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/workspace/crates/a/target/a.gz
$ ls target
a.gz    a.json
$ rm -rf target
$ git checkout fix-cli-locking
Switched to branch 'fix-cli-locking'
Your branch is up to date with 'origin/fix-cli-locking'.
$ cargo run -q -p nargo_cli -- --program-dir . execute
[a] Circuit witness successfully solved
[a] Witness saved to /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/workspace/target/a.gz
$ ls
Nargo.toml   Prover.toml  src/
$ ls ../../target
a.gz    a.json
$ cargo run -q -p nargo_cli -- execute
[a] Circuit witness successfully solved
[a] Witness saved to /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/workspace/target/a.gz
$ cargo run -q -p nargo_cli -- execute --workspace
[a] Circuit witness successfully solved
[a] Witness saved to /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/workspace/target/a.gz
[b] Circuit witness successfully solved
[b] Witness saved to /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/workspace/target/b.gz
$ cargo run -q -p nargo_cli -- execute --package b
[b] Circuit witness successfully solved
[b] Witness saved to /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/workspace/target/b.gz
```

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/acvm-repo/acvm/src/pwg/brillig.rs
+++ b/acvm-repo/acvm/src/pwg/brillig.rs
@@ -133,7 +133,7 @@ impl<'b, B: BlackBoxFunctionSolver<F>, F: AcirField> BrilligSolver<'b, F, B> {
 
         // Instantiate a Brillig VM given the solved calldata
         // along with the Brillig bytecode.
-        let vm = VM::new(calldata, brillig_bytecode, vec![], bb_solver, profiling_active);
+        let vm = VM::new(calldata, brillig_bytecode, bb_solver, profiling_active);
         Ok(vm)
     }
 
--- a/acvm-repo/brillig_vm/src/lib.rs
+++ b/acvm-repo/brillig_vm/src/lib.rs
@@ -107,7 +107,6 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>> VM<'a, F, B> {
     pub fn new(
         calldata: Vec<F>,
         bytecode: &'a [Opcode<F>],
-        foreign_call_results: Vec<ForeignCallResult<F>>,
         black_box_solver: &'a B,
         profiling_active: bool,
     ) -> Self {
@@ -117,7 +116,7 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>> VM<'a, F, B> {
             calldata,
             program_counter: 0,
             foreign_call_counter: 0,
-            foreign_call_results,
+            foreign_call_results: Vec::new(),
             bytecode,
             status: VMStatus::InProgress,
             memory: Memory::default(),
@@ -857,7 +856,7 @@ mod tests {
 
         // Start VM
         let solver = StubbedBlackBoxSolver::default();
-        let mut vm = VM::new(calldata, &opcodes, vec![], &solver, false);
+        let mut vm = VM::new(calldata, &opcodes, &solver, false);
 
         let status = vm.process_opcode();
         assert_eq!(status, VMStatus::Finished { return_data_offset: 0, return_data_size: 0 });
@@ -908,7 +907,7 @@ mod tests {
         ];
 
         let solver = StubbedBlackBoxSolver::default();
-        let mut vm = VM::new(calldata, &opcodes, vec![], &solver, false);
+        let mut vm = VM::new(calldata, &opcodes, &solver, false);
 
         let status = vm.process_opcode();
         assert_eq!(status, VMStatus::InProgress);
@@ -977,7 +976,7 @@ mod tests {
         ];
 
         let solver = StubbedBlackBoxSolver::default();
-        let mut vm = VM::new(calldata, &opcodes, vec![], &solver, false);
+        let mut vm = VM::new(calldata, &opcodes, &solver, false);
 
         let status = vm.process_opcode();
         assert_eq!(status, VMStatus::InProgress);
@@ -1050,7 +1049,7 @@ mod tests {
             },
         ];
         let solver = StubbedBlackBoxSolver::default();
-        let mut vm = VM::new(calldata, opcodes, vec![], &solver, false);
+        let mut vm = VM::new(calldata, opcodes, &solver, false);
 
         let status = vm.process_opcode();
         assert_eq!(status, VMStatus::InProgress);
@@ -1111,7 +1110,7 @@ mod tests {
             },
         ];
         let solver = StubbedBlackBoxSolver::default();
-        let mut vm = VM::new(calldata, opcodes, vec![], &solver, false);
+        let mut vm = VM::new(calldata, opcodes, &solver, false);
 
         let status = vm.process_opcode();
         assert_eq!(status, VMStatus::InProgress);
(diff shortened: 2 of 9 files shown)


tooling/nargo_cli/src/cli/check_cmd.rs
```diff
@@ -39,15 +37,18 @@ pub(crate) struct CheckCommand {
...
-        Some(NOIR_ARTIFACT_VERSION_STRING.to_string()),
-    )?;
+impl WorkspaceCommand for CheckCommand {
+    fn package_selection(&self) -> PackageSelection {
+        self.package_options.package_selection()
+    }
+    fn lock_type(&self) -> LockType {
+        // Creates a Nargo.toml template if it doesn't exist, otherwise only writes if `allow_overwrite` is true,
+        // so it shouldn't lead to any conflicts. Doesn't produce compilation artifacts.
+        LockType::None
```
asterite, line 47: If `allow_overwrites` is true, wouldn't this need to lock the thing is writes?

asterite: Ah, nevermind, I guess it writes the target and that wouldn't lead to any conflicts. I'm thinking if two check commands run in parallel, or if this command runs in parallel with another one that reads the target (I don't know if there's such command). So closing as resolved.

tooling/nargo_cli/src/cli/fmt_cmd.rs
```diff
@@ -22,24 +23,26 @@ pub(crate) struct FormatCommand {
...
+    }
+}
 
+pub(crate) fn run(args: FormatCommand, workspace: Workspace) -> Result<(), CliError> {
+    let check_mode = args.check;
     let mut workspace_file_manager = workspace.new_file_manager();
     insert_all_files_for_workspace_into_file_manager(&workspace, &mut workspace_file_manager);
 
-    let config = nargo_fmt::Config::read(&config.program_dir)
+    let config = nargo_fmt::Config::read(&workspace.root_dir)
```
asterite, line 45: I don't know if this is right. I think the closest `noirfmt.toml` file should be picked up so that you can customize formatting for a project inside a workspace differently from the entire workspace.

tooling/nargo_cli/src/cli/mod.rs
```diff
@@ -115,14 +107,28 @@ enum NargoCommand {
     GenerateCompletionScript(generate_completion_script_cmd::GenerateCompletionScriptCommand),
 }
 
+/// Commands that can execute on the workspace level, or be limited to a selected package.
+trait WorkspaceCommand {
+    fn package_selection(&self) -> PackageSelection;
+    fn lock_type(&self) -> LockType;
```
asterite, line 113: Would it be okay to document that the lock type is related to artifacts?

asterite: Oh, I see it's documented in `LockType`, so probably no need to do it here too.

tooling/nargo_cli/src/cli/fmt_cmd.rs
aakoshh: Thanks for highlighting that, I thought about whether people want to do that too. The reason I thought this should be fine is because in my earlier PR I actually changed the `config.program_dir` based on whether `--workspace` or `--package` was present, which meant you had to `cd` into the package for the package specific configuration to be effective. If we want this to work reliably we need to parse the config in the loop, with a fallback to the workspace if it doesn't exist. 

I also thought that for anyone that uses a workspace, why would they actually not want all their codebase to be consistently formatted 🤷

tooling/nargo_cli/src/cli/mod.rs
aakoshh: Added them anyway, maybe helps when looking at the `_cmd.rs` files 👍

tooling/nargo_cli/src/cli/fmt_cmd.rs
asterite: Sounds good.

Yesterday I found out that [formatting via LSP always uses the default configuration](https://github.com/noir-lang/noir/issues/7340) so I thought about fixing that first by passing the directory the file is in, then by recursively checking parent directories until a `noirfmt.toml` file is found. That's [how rustfmt works](https://github.com/rust-lang/rustfmt?tab=readme-ov-file#configuring-rustfmt). So when I pick up that issue I might modify this again (or, first I'll check what rustfmt does when you have a config file in the project and in the workspace).

aakoshh: Cool! There's also the fact that unlike every other command, `fmt` changes `PackageSelection::DefaultOrAll` to `PackageSelection::All`, because we want to format the everything unless specifically told otherwise, but that also means what if we're in a subdirectory and don't select any `--package` or the `--workspace`, it will still be effectively `--workspace`. So yeah, I think what you said about looking for the closest `noirfmt.toml` visible in every package is the right way.

asterite (approved):
Great refactor, looks good!

PR #7355, opened 2025-02-12
fix: let LSP read `noirfmt.toml` for formatting files

### LINKED ISSUE #7340: LSP formatter always uses the default configuration

asterite:
# Description

## Problem

Resolves #7340

## Summary

Also changes the logic of `nargo fmt` to try to find the configuration file in the current directory or parent directories.

## Additional Context

It seems `rustfmt` also allows the config file to be called ".rustfmt.toml", or for it to be placed in the global config directory (`.config/rustfmt/`) but those could be done in follow-up PRs.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/lsp/src/requests/mod.rs
+++ b/tooling/lsp/src/requests/mod.rs
@@ -301,6 +301,7 @@ fn on_formatting_inner(
     state: &LspState,
     params: lsp_types::DocumentFormattingParams,
 ) -> Result<Option<Vec<lsp_types::TextEdit>>, ResponseError> {
+    let file_path = params.text_document.uri.to_file_path();
     let path = params.text_document.uri.to_string();
 
     if let Some(source) = state.input_files.get(&path) {
@@ -310,7 +311,12 @@ fn on_formatting_inner(
             return Ok(None);
         }
 
-        let new_text = nargo_fmt::format(source, module, &Config::default());
+        let config = match file_path {
+            Ok(file_path) => Config::read(&file_path).unwrap_or_default(),
+            Err(_) => Config::default(),
+        };
+
+        let new_text = nargo_fmt::format(source, module, &config);
 
         let start_position = Position { line: 0, character: 0 };
         let end_position = Position {
--- a/tooling/lsp/src/requests/mod.rs
+++ b/tooling/lsp/src/requests/mod.rs
@@ -1,5 +1,5 @@
 use std::collections::BTreeMap;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::{collections::HashMap, future::Future};
 
 use crate::{insert_all_files_for_workspace_into_file_manager, parse_diff, PackageCacheData};
@@ -303,6 +303,7 @@ fn on_formatting_inner(
 ) -> Result<Option<Vec<lsp_types::TextEdit>>, ResponseError> {
     // The file_path might be Err/None if the action runs against an unsaved file
     let file_path = params.text_document.uri.to_file_path().ok();
+    let directory_path = file_path.as_ref().and_then(|path| path.parent());
 
     let path = params.text_document.uri.to_string();
 
@@ -313,7 +314,7 @@ fn on_formatting_inner(
             return Ok(None);
         }
 
-        let config = read_config(file_path);
+        let config = read_config(directory_path);
         let new_text = nargo_fmt::format(source, module, &config);
 
         let start_position = Position { line: 0, character: 0 };
@@ -331,9 +332,9 @@ fn on_formatting_inner(
     }
 }
 
-fn read_config(file_path: Option<PathBuf>) -> Config {
+fn read_config(file_path: Option<&Path>) -> Config {
     match file_path {
-        Some(file_path) => match Config::read(&file_path) {
+        Some(file_path) => match Config::read(file_path) {
             Ok(config) => config,
             Err(err) => {
                 eprintln!("Failed to parse noirfmt.toml: {}", err);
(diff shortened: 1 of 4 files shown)


tooling/lsp/src/requests/mod.rs
```diff
@@ -301,6 +301,7 @@ fn on_formatting_inner(
     state: &LspState,
     params: lsp_types::DocumentFormattingParams,
 ) -> Result<Option<Vec<lsp_types::TextEdit>>, ResponseError> {
+    let file_path = params.text_document.uri.to_file_path();
```
aakoshh, line 304: Are you deliberately passing the path of the document being edited to `Config::read`, rather than the directory containing the document? I suppose it works, e.g. `/foo/bar/baz.nr` will try `/foo/bar/baz.nr/noirfmt.toml` first, then `/foo/bar/noirfmt.toml`, but it's a bit confusing.

tooling/nargo_fmt/src/config.rs
```diff
@@ -56,16 +56,24 @@ config! {
...
-    pub fn read(path: &Path) -> Result<Self, ConfigError> {
-        let config_path = path.join("noirfmt.toml");
+    /// Reads a Config starting at the given path and going through the path parents
+    /// until a `noirfmt.toml` file is found in one of them or the root is reached.
+    pub fn read(mut path: &Path) -> Result<Self, ConfigError> {
+        loop {
+            let config_path = path.join("noirfmt.toml");
+            if config_path.exists() {
+                match std::fs::read_to_string(&config_path) {
+                    Ok(input) => return Self::of(&input),
```
aakoshh, line 66: Conceptual question: should the format files be merged all the way from the bottom to the top, so you an place ` noirfmt.toml` file in a subdirectory to override a subset of the settings, but still apply the defaults from the workspace, similarly to how `.gitignore` works? I didn't see the docs of `rustfmt` mention this, but just thinking it might be a nice feature.

aakoshh: For example I can imagine someone wanted to override some of the defaults for their entire project, but in a particular module with lots of array literals they find themselves unhappy with the default width. Should they now have to copy their workspace level config into that subdirectory so as not to revert back to the noir defaults?

Instead it could be something like this:

```
let mut toml_configs = Vec::new();
let mut path = document_file_path;
loop {
  path = match path.parent() {
    Some(path) => path,
    None => break,
  };
  match read_config_toml(path) { 
    Ok(toml) => toml_configs.push(toml_config),
    Err(Missing) => {},
    Err(e) => eprintln!("failed to parse {path}: {e}"), 
  }
}
toml_configs.reverse();
let config = toml_configs.fold(Config::default(), |mut cfg, toml| { cfg.fill_from_toml(toml); cfg });
```

tooling/lsp/src/requests/mod.rs
asterite: Ah, good point. I initially did it to make sure it checked parent paths but it makes sense to start with the parent.

tooling/nargo_fmt/src/config.rs
asterite: I like that idea! But I don't know if we should do it in this PR.

There's also a chance that a workspace configures the maximum width to 80, but a sub-project wants the default. They put an empty config file but it doesn't work because it's merged with the parent values. So maybe being explicit about inheritance might be desired (with an `inherit = true` or similar).

tooling/lsp/src/requests/mod.rs
asterite: Done!

tooling/lsp/src/requests/mod.rs
```diff
@@ -327,6 +332,19 @@ fn on_formatting_inner(
     }
 }
 
+fn read_config(file_path: Option<&Path>) -> Config {
+    match file_path {
+        Some(file_path) => match Config::read(file_path) {
+            Ok(config) => config,
+            Err(err) => {
+                eprintln!("Failed to parse noirfmt.toml: {}", err);
```
aakoshh, line 340: I'd print the full path, otherwise if there are multiple the user wouldn't know which one failed exactly.

aakoshh: Ah never mind, that is part of the error 👍

asterite: Ah, I thought about it but we don't always get the path in the error 😅 I'll change it to include it.

aakoshh (approved):
Great, let's consider the overrides in a follow up!

tooling/lsp/src/requests/mod.rs
asterite: Done! I change it to just print `err`, which now always includes the full path.

aakoshh: Okay! A consequence here is that if there is an error in a deeply nested config, it reverts all of it to defaults, even if there is a non-default in one of the ancestors.

asterite: I guess we could try again starting from the grandparent directory of the file that errored... but maybe for now this is good enough, if there's an error it might be noticeable and the user would fix it before continuing formatting everything.

PR #7356, opened 2025-02-12
feat(performance): Check sub operations against induction variables

### LINKED ISSUE #7353: Optimize Sub operations based upon the loop induction variable
### Problem

In https://github.com/noir-lang/noir/pull/6910 we started hoisting add/mul operations based upon the outer loop's induction variable. While in https://github.com/noir-lang/noir/pull/7344 we are adding further optimizations to transform checked operations to unchecked based upon the induction variable. 

For simplicity in those initial PRs, sub operations were not included in the above optimizations. Sub overflows depend on if the induction variable is on the lhs or the rhs (e.g. if it is the rhs we should check the induction variable's upper bound but if it is the lhs we should check the lower bound).

### Happy Case

`Sub` operations that do not overflow when using a given induction variable should be optimized either through hoisting or being transformed to an unchecked operation. 

### Workaround

None

### Workaround Description

The compiler needs to add this optimization for it to be applied.

### Additional Context

_No response_

### Project Impact

Blocker

### Blocker Context

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #7353 

## Summary\*

This is an expansion of the optimization from https://github.com/noir-lang/noir/pull/7344 and https://github.com/noir-lang/noir/pull/6910. In those PRs we simply handled add/mul operations as they only needed to be checked against the upper bound of a loop. This PR now tracks the lower bounds of our induction variables and special cases for when the induction variable is on the LHS during a sub operation. 

I have added a basic unit test for when the induction var is on the LHS of a sub operation with a lower bound that would trigger an overflow.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
@@ -90,12 +90,12 @@ struct LoopInvariantContext<'f> {
     inserter: FunctionInserter<'f>,
     defined_in_loop: HashSet<ValueId>,
     loop_invariants: HashSet<ValueId>,
-    // Maps current loop induction variable -> fixed upper loop bound
+    // Maps current loop induction variable -> fixed lower and upper loop bound
     // This map is expected to only ever contain a singular value.
     // However, we store it in a map in order to match the definition of
     // `outer_induction_variables` as both maps share checks for evaluating binary operations.
     current_induction_variables: HashMap<ValueId, (FieldElement, FieldElement)>,
-    // Maps outer loop induction variable -> fixed upper loop bound
+    // Maps outer loop induction variable -> fixed lower and upper loop bound
     // This will be used by inner loops to determine whether they
     // have safe operations reliant upon an outer loop's maximum induction variable.
     outer_induction_variables: HashMap<ValueId, (FieldElement, FieldElement)>,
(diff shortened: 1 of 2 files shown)


compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -93,11 +94,15 @@ struct LoopInvariantContext<'f> {
     // This map is expected to only ever contain a singular value.
     // However, we store it in a map in order to match the definition of
     // `outer_induction_variables` as both maps share checks for evaluating binary operations.
-    current_induction_variables: HashMap<ValueId, FieldElement>,
+    current_induction_variables: HashMap<ValueId, (FieldElement, FieldElement)>,
     // Maps outer loop induction variable -> fixed upper loop bound
     // This will be used by inner loops to determine whether they
     // have safe operations reliant upon an outer loop's maximum induction variable.
-    outer_induction_variables: HashMap<ValueId, FieldElement>,
+    outer_induction_variables: HashMap<ValueId, (FieldElement, FieldElement)>,
```
asterite, line 101: The docs for both of these should be updated ("-> fixed lower and upper loop bound")

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -804,4 +827,32 @@ mod test {
         let ssa = ssa.loop_invariant_code_motion();
         assert_normalized_ssa_equals(ssa, expected);
     }
+
+    #[test]
+    fn do_not_transform_unsafe_sub_to_unchecked() {
```
asterite, line 832: Should a similar test be added where `sub` is transformed to `unchecked_sub`?

vezenovm: I will add one

vezenovm: done

asterite (approved):
Looks great!

Issue #7324, opened 2025-02-07
Refactor frontend to use `SignedField` instead of `(FieldElement, bool)` pair


jfecher:
### Problem

(FieldElement, bool) is a clunky abstraction which always requires comments to remember the `is_negative` parameter

### Happy Case

https://github.com/noir-lang/noir/pull/7312 adds SignedField which is a better abstraction - lets use it

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

TomAFrench:
Agreed, we have a few places where we just completely ignore the sign value which I saw while addressing #7433. Granted these were in display functions but it does seem like we've considered that the negative value is not relevant for field types so forcing us to deal with these is good.

PR #7368, opened 2025-02-13
feat: LSP hover for integer literals

asterite:
# Description

## Problem

No existing issue, just something Rust Analyzer does that I thought would be nice to have.

## Summary

(it's better to review this commit by commit as the first one just moves things around)

Now hovering over an integer literal will show its inferred type together with its decimal value and hexadecimal value (Rust Analyzer also shows the binary (`0b...`) literal but I omitted that because Noir doesn't have such literals).

lsp-hover-integer

## Additional Context

This PR also makes it easier for later PRs to provide hover over types and other AST nodes, to for example explain what `Expr` is, what `i32` is, etc. (also similar to Rust Analyzer, which I think is nice because you can learn a lot of the language through the editor hover).

## Documentation

Check one:

# PR Checklist


michaeljklein (changes requested):
Please add a few sanity-check tests for this since we could change a representation and end up showing the wrong hex/decimal value (I'm thinking of the upcoming https://github.com/noir-lang/noir/issues/7324 changes).

Otherwise LGTM

asterite:
I think when we change to `SignedField` we'll have to unpack that to pass it to the visitor, or make the visitor use `SignedField`. I'm not sure it could break without first stopping to compile 🤔 

What would the test be like?

tooling/lsp/src/requests/hover/from_visitor.rs
```diff
@@ -0,0 +1,78 @@
...
+    args: &ProcessRequestCallbackArgs,
+) -> Option<Hover> {
+    let file_id = file_id?;
+    let file = args.files.get_file(file_id)?;
+    let source = file.source();
+    let (parsed_moduled, _errors) = parse_program(source);
+    let byte_index = utils::position_to_byte_index(args.files, file_id, &position)?;
+
+    let mut finder = HoverFinder::new(args.files, file_id, args.interner, byte_index);
+    parsed_moduled.accept(&mut finder);
```
TomAFrench, line 27:
```suggestion
    let (parsed_modules, _errors) = parse_program(source);
    let byte_index = utils::position_to_byte_index(args.files, file_id, &position)?;

    let mut finder = HoverFinder::new(args.files, file_id, args.interner, byte_index);
    parsed_modules.accept(&mut finder);
```

tooling/lsp/src/requests/hover/from_visitor.rs
```diff
@@ -0,0 +1,78 @@
...
+
+        let value_base_10 = value.to_string();
+
+        // For simplicity we parse the value as a BigInt to convert it to hex
+        // because `FieldElement::to_hex` will include many leading zeros.
+        let value_big_int = BigInt::from_str(&value_base_10).unwrap();
+        let negative = if negative { "-" } else { "" };
+
+        let value =
+            format!("    {typ}\n---\nvalue of literal: `{negative}{value_base_10} ({negative}0x{value_big_int:x})`");
```
TomAFrench, line 74:
```suggestion
            format!("    {typ}\n---\nvalue of literal: `{negative}{value_base_10} ({negative}0x{value_big_int:02x})`");
```

Currently we'll print `-0x8` for `-8` but I think that we should generally keep to printing even length hex strings.

TomAFrench (approved):
LGTM, one nit on odd-length hexadecimal.

Agree that if we're changing how we're representing signed integers then this will break rather than failing silently, I don't think there's any way around that that isn't a full e2e test, which is overkill. I think the most we'd want to test for this is a couple of sanity checks that we get the expected formatted string given a `(FieldElement, bool)` input (to catch stuff like odd-length hex strings, etc.)

tooling/lsp/src/requests/hover/from_visitor.rs
asterite: Done! (there was another occurrence of this so I changed them both in one go)

tooling/lsp/src/requests/hover/from_visitor.rs
asterite: Done!

Any reason to print even-length hex strings? For example Rust will show it as `-0x8` (just a question, either way is fine)

asterite:
> I think the most we'd want to test for this is a couple of sanity checks that we get the expected formatted string given a (FieldElement, bool)

Any numbers in particular that we should test? The logic mostly relies on format strings.

tooling/lsp/src/requests/hover/from_visitor.rs
TomAFrench: I know that a decent number of implementation of hex string parsing will fail on odd-length strings so I think it's a good idea to pad them like this in general.

PR #7371, opened 2025-02-13
feat: add `wrap_comments` and `comment_width` options to `nargo fmt`

### LINKED ISSUE #7332: Add a `wrap_comments` option to nargo fmt
### Problem

Similar to [rustfmt](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#wrap_comments), we should add a `wrap_comments` option to `nargo fmt`. When enabled, comments are wrapped without breaking whole words or URLs.

### Happy Case

There's a new option for this.

### Workaround

None

### Workaround Description

Manually inserting newlines.

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #7332

## Summary

This is an opt-in feature, just like in `rustfmt`.

## Additional Context

I tried to make sure this covers most scenarios though I'm sure I missed some. We could enhance things as we find them.

I also thought about enabling comment wrapping in the stdlib and test programs (things become easier to read) but I'll leave that for an (optional) PR.

Question: should the default `comment_width` be 100 or 80? It's 80 in `rustfmt`, which is different than the default `max_width` of 100 and I don't know why. Maybe comments are easier to read if they are shorter (compared to code)? 

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/nargo_fmt/src/chunks.rs
+++ b/tooling/nargo_fmt/src/chunks.rs
@@ -42,7 +42,7 @@ pub(crate) enum Chunk {
     /// (for example for a call we'll add a trailing comma to the last argument).
     TrailingComma,
     /// A trailing comment (happens at the end of a line, and always after something else have been written).
-    TrailingComment(TextChunk),
+    TrailingComment(TextChunk, bool /* indent */),
     /// A leading comment. Happens at the beginning of a line.
     LeadingComment(TextChunk),
     /// A group of chunks.
@@ -67,7 +67,7 @@ impl Chunk {
         match self {
             Chunk::Text(chunk)
             | Chunk::Verbatim(chunk)
-            | Chunk::TrailingComment(chunk)
+            | Chunk::TrailingComment(chunk, _)
             | Chunk::LeadingComment(chunk) => chunk.width,
             Chunk::Group(group) => group.width(),
             Chunk::SpaceOrLine => 1,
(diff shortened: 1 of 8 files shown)


TomAFrench:
How would you expect 

```noir
// This is a long comment that's going to be wrapped.
// Short
```

to be formatted?

```noir
// This is a long comment
// that's going to be
// wrapped.
// Short

or 

// This is a long comment
// that's going to be
// wrapped. Short
```

TomAFrench:
We also don't push short comments onto a single line, e.g. 

```noir
// This
// is
// a
// long
```

into

```noir
// This is a long
```

asterite:
Right... I thought about those two cases. I actually wanted to see what rustfmt does but that's unstable and I was lazy to install an unsafe version of it to try it.

My guess is that this:

```noir
// This is a long comment that's going to be wrapped.
// Short
```

is formatted like this:

```noir
// This is a long comment
// that's going to be
// wrapped.
// Short
```

The main reason is that otherwise if you really want a newline after "wrapped." then the formatter would undo it (your second comment "We also don't push short comments onto a single line, e.g.")

This is especially tricky if code in comments is being formatted. If you write this:

```noir
// If we have a loop like this:
// 
// for i in 0..10 {
//   println("hello!");
// }
```

it would be kind of bad if that `println` always moves up because there's space.

I know it ends up a bit annoying that a wrapped comment might not look good, but my thought of this feature is that it makes sure you don't exceed the maximum, and the comment might end up not looking very nice and you'll fix it, always in a way that doesn't exceed the maximum.

One other thing: I think the feature in rustfmt is unstable because it has many bugs. I don't know if those bugs are because it's not clear how comments should be wrapped (this discussion).

nventuro:
It'd save me lots of time if this was implemented. Pretty please.

PR #7373, opened 2025-02-13
chore: box `ParserError`s in `InterpreterError`

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

`InterpreterError` is currently 488 bytes (and so all of `IResults` are as well). Most of that size comes from it holding a `ParserError` inside one of its variants so I've boxed that which reduces the size of an `InterpreterError` to 192 bytes.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/errors.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/errors.rs
@@ -150,7 +150,7 @@ pub enum InterpreterError {
         location: Location,
     },
     FailedToParseMacro {
-        error: ParserError,
+        error: Box<ParserError>,
         tokens: String,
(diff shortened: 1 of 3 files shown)


asterite (approved):
Looks good!

Is this noticeable somewhere? I'm asking mainly because I'm thinking of reviving https://github.com/noir-lang/noir/pull/7147 : it's probably not noticeable in a big compilation, but it is faster (milliseconds faster) and that always helps if it's seen from an LSP perspective (where every millisecond matters).

TomAFrench:
> Value is already 888 bytes. So IResult's size doesn't change.

Oof I didn't see that. Will make a followup

PR #7374, opened 2025-02-13
feat: cache argless comptime function values

asterite:
# Description

## Problem

I noticed that doing `nargo check` on "rollup-base-private" would take about 1.5 seconds. Profiling it turned out that a lot of that time (around 900ms) is spent in the comptime interpreter.

I know a tree-walk interpreter is slow but I wanted to know where the slowness happened so I printed expressions that took more than 10ms to evaluate. It turns out that there were many calls to `get_vk_merkle_tree` which builds a massive tree. Those are mainly used in tests so my first reaction was to change those `comptime`s to non-comptime because it would work the same.

## Summary

A second thought was that we could cache function return values if they have no arguments, which is a more general solution, so that's what this PR does.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/comptime.rs
+++ b/compiler/noirc_frontend/src/elaborator/comptime.rs
@@ -98,6 +98,7 @@ impl<'context> Elaborator<'context> {
             self.debug_comptime_in_file,
             self.interpreter_call_stack.clone(),
             self.pedantic_solving,
+            self.cached_function_values,
         );
 
         elaborator.function_context.push(FunctionContext::default());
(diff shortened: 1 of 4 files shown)


asterite:
Aaah... almost. It turns out non-comptime functions can refer to comptime globals. I guess that makes sense if we want to build things at comptime, then use them at runtime. Then we'd have to determine if we can cache a function's value if it doesn't use comptime globals... but that's more complex than this PR.

Maybe at least those `comptime { get_vk_merkle_tree() }` calls could be removed on aztec-packages tests so that compilation is faster.

PR #7391, opened 2025-02-14
7381 use artifact fs

### LINKED ISSUE #7381: Use the artifact CLI file system libraries in other CLI tools
Some other CLIs have duplicated the `fs` utilities:
* `profiler`
* `inspector`
* criterion benchmarks

Those can all import the `noir_artifact_cli` crate as a library to use the `fs` module.

aakoshh:
# Description

## Problem\*

Resolves #7381

## Summary\*

Changes other CLI tools to use the `noir_artifact_cli::fs` functions to read artefacts.

- [x] `inspector`
- [x] `profiler`
- [ ] `nargo`

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_driver/src/contract.rs
```diff
@@ -41,6 +41,8 @@ pub struct CompiledContract {
 pub struct ContractFunction {
     pub name: String,
 
+    pub hash: u64,
+
```
TomAFrench, line 45: These shouldn't have hashes as we don't use them.

aakoshh: In the next PR I got rid of the `Circuit` type in the `execute_cmd` of `noir-execute` in favour of mapping both programs and contract functions back to a `CompiledProgram` [here](https://github.com/noir-lang/noir/pull/7406/files#diff-efbce6b0629ea34dbdf888d453bc5c1371704981a7c2af869a56d39a329e56c8R80). 

I understand we don't use them at the moment in `nargo`, however the `ContractFunction` comes from a `CompiledProgram` and at that point it does have `hash`. I don't see why, if programs and functions come from the same source, we have to lose this information, when it can be handy if for nothing else than providing some extra metadata about a failed contract execution.

aakoshh: I can imagine that a contract fails, we have a dump from Nico, and yes we can execute the artifact based on its path, but with the `hash` we can also compile the source from `nargo` and ascertain that we are dealing with the same code, before starting to modify it for debugging. 

I’m sure other hashes like the git hash of the contract repo can be added somewhere by the relevant teams to make this a non-issue, but also don’t see the downside of having it.

aakoshh: I'd also argue that adding these fields, and the ability to deal with both kinds of artefacts in our CLIs with the minimum amount of special cases is a nice way to minimise the need to [do the same in Javascript](https://github.com/AztecProtocol/aztec-packages/blob/2e83891b1b1d080febe5f3eac5dfe51e5844cf06/noir-projects/noir-contracts/scripts/extractFunctionAsNoirArtifact.js#L27-L36). For example that script has to do what I did in this PR by setting `hash` to 0 and `names` to some non-empty array to be able to pass it to the profiler. If we do it ourselves, it's less likely that the profiler or the inspector will be surprised at the input.

compiler/wasm/src/types/noir_artifact.ts
```diff
@@ -58,6 +58,8 @@ export interface ABIVariable {
 export interface NoirFunctionEntry {
   /** The name of the function. */
   name: string;
+  /** The hash of the circuit. */
+  hash?: number | string;
```
TomAFrench, line 62: I think we can break backwards compatibility here without much issues. The cases where we have artifacts used across nargo builds currently is very small.

tooling/profiler/src/cli/mod.rs
```diff
@@ -1,5 +1,5 @@
 use clap::{Parser, Subcommand};
-use color_eyre::eyre;
+use color_eyre::eyre::{self};
```
TomAFrench, line 2: Looks odd

compiler/wasm/src/compile.rs
```diff
@@ -128,16 +128,21 @@ impl JsCompileContractResult {
...
+/// This is for all intents and purposes the file system that the compiler will use to resolve/compile
+/// files in the crate being compiled and its dependencies.
+///
+/// Using a `BTreeMap` to add files to the [FileManager] in a deterministic order,
+/// which affects the `FileId` in the `Location`s in the AST on which the `hash` is based.
+/// Note that we cannot expect to match the IDs assigned by the `FileManager` used by `nargo`,
+/// because there the order is determined by the dependency graph as well as the file name.
 #[derive(Deserialize, Default)]
-pub struct PathToFileSourceMap(pub(crate) HashMap<std::path::PathBuf, String>);
+pub struct PathToFileSourceMap(pub(crate) BTreeMap<std::path::PathBuf, String>);
```
TomAFrench, line 145: Can you move the `#[wasm_bindgen]` down below the doc comment to avoid it getting lost?

compiler/noirc_driver/src/contract.rs
TomAFrench: That's fair, there's no harm in keeping them.

TomAFrench (commented):
Some nits, generally looks good.

TomAFrench (approved):

PR #7395, opened 2025-02-14
feat: allow unquoting TraitConstraint in trait impl position

### LINKED ISSUE #7389: Allow unquoting TraitConstraint in trait positions
### Problem

The following code doesn't compile:

```noir
struct Struct {}

trait Trait {}

fn main() {}

#[foo]
comptime fn foo(_: FunctionDefinition) -> Quoted {
    let tr = quote { Trait }.as_trait_constraint();
    let st = quote { Struct }.as_type();
    quote {
        impl $tr for $st {

        }
    }
}
```

It gives this error:

```
error: Cannot inline values of type `TraitConstraint` into this position
   ┌─ src/main.nr:11:5
   │
11 │ ╭     quote {
12 │ │         impl $tr for $st {
13 │ │
14 │ │         }
15 │ │     }
   │ ╰─────' Cannot inline value `Trait`
```

### Happy Case

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #7389

## Summary

Before this PR we parsed `impl Trait for Type` as requiring the `Trait` part to be parsed as a `NamedType`, because that's the only way you can refer a trait. Because now we need to allow a "resolved" type in that position, resulting from unquoting a `TraitConstraint` into a `Type::TraitAsType` type, this PR updates the AST node for a trait impl to hold any `UnresolvedType`, together with other structs that made that assumption. Then when solving that type we check that it's either a `NamedType` or a resolved type that points to `Type::TraitAsType` and proceed from there. 

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1983,7 +2007,31 @@ impl<'context> Elaborator<'context> {
             self.file = trait_impl.file_id;
             self.local_module = trait_impl.module_id;
 
-            let trait_id = self.resolve_trait_by_path(trait_impl.trait_path.clone());
+            let trait_type = match &trait_impl.r#trait.typ {
+                UnresolvedTypeData::Named(trait_path, trait_generics, _) => {
+                    TraitType::Named(trait_path.clone(), trait_generics.clone())
+                }
+                UnresolvedTypeData::Resolved(quoted_type_id) => {
```
jfecher, line 2014: Do we need to define this new TraitType enum or could we just extract the trait_id, path_span, and trait_generics here instead?

compiler/noirc_frontend/src/hir/comptime/value.rs
```diff
@@ -495,6 +495,11 @@ impl Value {
             Value::UnresolvedType(typ) => {
                 Token::InternedUnresolvedTypeData(interner.push_unresolved_type_data(typ))
             }
+            Value::TraitConstraint(trait_id, generics) => {
+                let name = Rc::new(interner.get_trait(trait_id).name.0.contents.clone());
+                let typ = Type::TraitAsType(trait_id, name, generics);
+                Token::QuotedType(interner.push_quoted_type(typ))
```
jfecher, line 501: Lowering this as a type made me a bit uncomfortable at first but I suppose it is fine since no data is lost. It does mean you could also use a `quote { impl Foo }.as_type()` where a trait constraint is expected which I suppose is fine.

compiler/noirc_frontend/src/hir/resolution/errors.rs
```diff
@@ -678,6 +680,13 @@ impl<'a> From<&'a ResolverError> for Diagnostic {
                     *span)
 
             }
+            ResolverError::ExpectedTrait { span  } => {
+                Diagnostic::simple_error(
+                    "Expected a trait".to_string(),
```
jfecher, line 685: I think this is a bit too general - I could imagine being confused by this as a user. Can you add the "but found ___" part back to this error?

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1983,7 +2007,31 @@ impl<'context> Elaborator<'context> {
...
+                    let typ = self.interner.get_quoted_type(*quoted_type_id);
+                    if let Type::TraitAsType(trait_id, _, trait_generics) = typ {
+                        TraitType::Trait(*trait_id, trait_generics.clone())
+                    } else {
+                        self.push_err(ResolverError::ExpectedTrait {
+                            span: trait_impl.r#trait.span,
+                        });
```
jfecher, line 2021: Looks like we could convert `typ` to a string here to keep the "but found" part of the error

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1983,7 +2007,31 @@ impl<'context> Elaborator<'context> {
...
+                        continue;
+                    }
+                }
+                _ => {
+                    self.push_err(ResolverError::ExpectedTrait { span: trait_impl.r#trait.span });
+                    continue;
```
jfecher, line 2027: Same here

jfecher (commented):
LGTM, just one comment about a clearer error message

compiler/noirc_frontend/src/hir/comptime/value.rs
asterite: Ah, good point. I actually started doing it differently, having a new "interned" quote until I saw we have `Type::TraitAsType` which has `TraitId` and generics so I thought that was it. When printing it, it shows up as `impl Foo<..>` as you mentioned, so inside a quoted value in an `impl $constraint for ...` it looks odd, like "impl impl Foo for". So I think this concern is valid.

Would it make sense to introduce a different type here? Or we could add a boolean that says whether it's a constraint or not so we print it better but also disallow it for `quote { impl Foo }.as_type()`.

compiler/noirc_frontend/src/hir/resolution/errors.rs
asterite: I think the "but found {found}" was showing the token, but I'll craft a message here based on the type.

asterite: Done!

compiler/noirc_frontend/src/elaborator/mod.rs
asterite: Done!

compiler/noirc_frontend/src/elaborator/mod.rs
asterite: Done!

compiler/noirc_frontend/src/elaborator/mod.rs
asterite: Done, thanks! It ends being much clearer and shorter.

compiler/noirc_frontend/src/hir/comptime/value.rs
jfecher: I think keeping it the way it is would use the least code at least and is somewhat convenient for users. I don't think "impl Foo" being printed out is too confusing.

jfecher (approved):

PR #7396, opened 2025-02-14
fix(ssa): Accurately mark binary ops for hoisting and check Div/Mod against induction variable lower bound

vezenovm:
# Description

## Problem\*

No issue but just a bug I found while working towards other tasks.

## Summary\*

We currently mark an invariant as being allowed for hoisting with the following:
```
        let can_be_deduplicated = instruction.can_be_deduplicated(self.inserter.function, false)
            || matches!(instruction, Instruction::MakeArray { .. })
            || matches!(instruction, Instruction::Binary(_))
            || self.can_be_deduplicated_from_loop_bound(&instruction);
```
This means we are currently allowing any Binary op to be hoisting if it is an invariant. We need to remove the `|| matches!(instruction, Instruction::Binary(_))` and rely on `instruction.can_be_deduplicated` and  `self.can_be_deduplicated_from_loop_bound`. 

This is an issue because if we were to have an operation that can potentially error (such as from overflow or div by zero) that was under a predicate, we would hoist that instruction out of the loop. If that predicate turns off the error (e.g. by protecting against a div by zero in user code) a user would still hit an error. 

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
> **loop_invariant_regression_inliner_max**	+73 ❌	**+7.61%**

To get these regressions back, we would need some way to determine that the predicate that the instruction is under is the same as the predicate of the loop pre-header. This is a bit simpler for ACIR after flattening and unrolling, but invariant code motion occurs before unrolling and is mostly needed for Brillig where we still have a CFG.

vezenovm:
> **ram_blowup_regression_inliner_max**	+43,008 ❌	**+6.31%**
> **brillig_cow_regression_inliner_max**	+28,236 ❌	**+6.22%**

Bug was introduced in this PR https://github.com/noir-lang/noir/pull/6910. We look to have kept about half of the gains with this correction fix:
https://github.com/noir-lang/noir/pull/6910#issuecomment-2557920237

vezenovm:
> This is an issue because if we were to have an operation that can potentially error (such as from overflow or div by zero) that was under a predicate, we would hoist that instruction out of the loop. If that predicate turns off the error (e.g. by protecting against a div by zero in user code) a user would still hit an error.

Will add some more tests showing this explicitly.

vezenovm:
> # Changes to Brillig bytecode sizes
> > Generated at commit: [b86b465b135eda14235af49815aeb03a3d5c5fa0](/noir-lang/noir/commit/b86b465b135eda14235af49815aeb03a3d5c5fa0), compared to commit: [38eeee39a98a62747dcca3b31b409151761d4ef1](/noir-lang/noir/commit/38eeee39a98a62747dcca3b31b409151761d4ef1)
> 
> ### 🧾 Summary (10% most significant diffs)
> **loop_invariant_regression_inliner_max**	+34 ❌	**+29.31%**
> **loop_invariant_regression_inliner_zero**	+34 ❌	**+29.31%**
> **loop_invariant_regression_inliner_min**	+39 ❌	**+26.90%**

This is due to a new method to the `loop_invariant_regression` test.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -228,7 +228,6 @@ impl<'f> LoopInvariantContext<'f> {
 
         let can_be_deduplicated = instruction.can_be_deduplicated(self.inserter.function, false)
             || matches!(instruction, Instruction::MakeArray { .. })
-            || matches!(instruction, Instruction::Binary(_))
```
TomAFrench, line 231: We can still allow any unchecked operations here though.

TomAFrench: Ah, this should be covered by `can_be_deduplicated`.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -315,7 +314,11 @@ impl<'f> LoopInvariantContext<'f> {
...
             binary.operator,
-            BinaryOp::Add { .. } | BinaryOp::Mul { .. } | BinaryOp::Sub { .. }
+            BinaryOp::Add { .. }
+                | BinaryOp::Mul { .. }
+                | BinaryOp::Sub { .. }
+                | BinaryOp::Div
+                | BinaryOp::Mod
         ) {
             return false;
         }
```
TomAFrench, line 324: I think we may want to make this a positive match rather than a negated one at this point. Also a quick comment explaining why we're rejecting these operators would be good.

TomAFrench (approved):
LGTM, small nit.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -330,7 +333,15 @@ impl<'f> LoopInvariantContext<'f> {
             induction_vars.get(&binary.lhs),
             induction_vars.get(&binary.rhs),
         ) {
-            (Some((lhs, _)), None, None, Some((_, upper_bound))) => (lhs, *upper_bound),
+            (Some((lhs, _)), None, None, Some((lower_bound, upper_bound))) => {
+                if matches!(binary.operator, BinaryOp::Div | BinaryOp::Mod) {
+                    // If we have a Div/Mod operation we want to make sure that the
+                    // lower bound is not zero.
+                    (lhs, *lower_bound)
```
jfecher, line 340: Can edit the comment below on line 357 too.

"We evaluate this expression using the upper bounds (or lower in the case of div/mod) of its inputs to check..."

jfecher (approved):

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: I decided to just remove the match.

PR #7211, opened 2025-01-28
chore(ssa): Flip the SSA Brillig constraint check to off by default

rkarabut:
# Description

## Problem

Resolves #7171

## Additional Context

In light of the pass currently making protocol-cicuits unable to finish compilation, switching it off by default.

## Documentation

Check one:
- [x] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #7404, opened 2025-02-17
chore(ssa): Turn the Brillig constraints check back on by default

rkarabut:
# Description

As the reasons of the major compile time performance hit on certain large unrolled functions (e.g. rollup-block-root) have been figured out and eliminated in #7190 after temporarily switching the Brillig constraints check off by default in #7211, turn the check back on.

## Summary

Reverts #7211

## Documentation

Check one:

# PR Checklist\*


rkarabut:
@aakoshh, also changed tracking to employ a HashSet<(FunctionId, InstructionId)> instead of the count, as that one _did_ in fact lead to some strange miscounts.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -110,11 +110,11 @@ struct DependencyContext {
...
     // Map of Brillig call ids to sets of the value ids descending
     // from their arguments and results
-    tainted: BTreeMap<InstructionId, BrilligTaintedIds>,
+    tainted: BTreeMap<(FunctionId, InstructionId), BrilligTaintedIds>,
     // Map of argument value ids to the Brillig call ids employing them
     call_arguments: HashMap<ValueId, Vec<InstructionId>>,
-    // Maintains count of calls being tracked
-    tracking_count: usize,
+    // The set of calls currently being tracked
+    tracking: HashSet<(FunctionId, InstructionId)>,
```
aakoshh, line 117: I don't think `FunctionId` should be needed here. 

It looks like a [separate DependencyContext is created](https://github.com/noir-lang/noir/blob/5935381183e2457d304ba76ae66cba8640afe8ef/compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs#L51-L61) for each function, and the [key inserted](https://github.com/noir-lang/noir/blob/5935381183e2457d304ba76ae66cba8640afe8ef/compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs#L395) is always with the ID of the `function` passed to `build` and `process_instructions`, which is the same for the entire lifetime of `DependencyContext`. 

Another way to look at it is that if this wasn't the case then the [bookkeeping of blocks](https://github.com/noir-lang/noir/blob/5935381183e2457d304ba76ae66cba8640afe8ef/compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs#L303) would also have to have this key, because block IDs are not unique across functions.

rkarabut: Oh yeah, you're right. In fact it would be better to just make a `function` field in the context and get rid of the redundant `function` argument in a bunch of inner calls. That's probably a remnant of the initial implementation.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -394,23 +391,19 @@ impl DependencyContext {
                 for argument in &arguments {
                     if let Some(calls) = self.call_arguments.get(argument) {
                         for call in calls {
-                            if let Some(tainted_ids) = self.tainted.get_mut(call) {
-                                tainted_ids.tracking = true;
-                                self.tracking_count += 1;
+                            if self.tainted.get(call).is_some() {
```
aakoshh, line 394:
```suggestion
                            if self.tainted.contains_key(call) {
```

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -394,23 +391,19 @@ impl DependencyContext {
...
                         }
                     }
                 }
             }
-            if let Some(tainted_ids) = self.tainted.get_mut(instruction) {
-                if !tainted_ids.tracking {
-                    tainted_ids.tracking = true;
-                    self.tracking_count += 1;
-                }
+            if self.tainted.get(instruction).is_some() {
```
aakoshh, line 401:
```suggestion
            if self.tainted.contains_key(instruction) {
```

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -582,8 +579,8 @@ impl DependencyContext {
         self.side_effects_condition.map(|v| parents.insert(v));
 
         // Don't update sets for the calls not yet being tracked
-        for (_, tainted_ids) in self.tainted.iter_mut() {
-            if tainted_ids.tracking {
+        for (call, tainted_ids) in self.tainted.iter_mut() {
+            if self.tracking.contains(call) {
```
aakoshh, line 583: I would swap these around: loop over the items in `self.tracking` and look them up in `self.tainted`. That's because as far as I can see we only insert into `self.tracking` items that are in `self.tainted` and we encounter their key when going over the instructions, that is, `self.tracking` is a subset of `self.tainted.keys()`, and so it should always contain at most as many items as `self.tainted`.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -600,15 +597,15 @@ impl DependencyContext {
             .collect();
 
         // Skip untracked calls
-        for (_, tainted_ids) in self.tainted.iter_mut() {
-            if tainted_ids.tracking {
+        for (call, tainted_ids) in self.tainted.iter_mut() {
+            if self.tracking.contains(call) {
```
aakoshh, line 601: Same here.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
```diff
@@ -619,19 +616,19 @@ impl DependencyContext {
...
         use acvm::acir::AcirField;
 
         // Only allow numeric constant indices
         if let Some(value) = function.dfg.get_numeric_constant(index) {
             if let Some(index) = value.try_to_u32() {
                 // Skip untracked calls
-                for (_, tainted_ids) in self.tainted.iter_mut() {
-                    if tainted_ids.tracking {
+                for (call, tainted_ids) in self.tainted.iter_mut() {
+                    if self.tracking.contains(call) {
```
aakoshh, line 631: And here.

compiler/noirc_evaluator/src/ssa/checks/check_for_underconstrained_values.rs
rkarabut: Done

aakoshh (approved):
👍

rkarabut:
@TomAFrench is this okay to merge?

TomAFrench:
Sure, I'm fine with this

PR #7406, opened 2025-02-17
chore(cli): Forward `nargo execute` to `noir_artifact_cli`

### LINKED ISSUE #7380: Call the artifact CLI from `nargo execute`
Redirect `nargo execute` to `noir-artifact execute`, șo we get integration test coverage and remove the duplicated code between them.

Need to figure out how this "shelling out" should work exactly: perhaps the `noir_artifact_cli` crate can be built from the `build.rs` of `nargo_cli`, but should we assume that it's installed with `noirup.sh` for example, on CI, etc? Or maybe just turn more things into library functions.

aakoshh:
# Description

## Problem\*

Resolves #7380

Builds on #7391

## Summary\*

Changed `nargo_cli::execute_cmd::run` to create and run an instance of `noir_artifact_cli::commands::execute_cmd::ExecuteCommand` to completely outsource program execution to `noir_artifact_cli`.

## Additional Context

Up to now the `ExecuteCommand` in `noir_artifact_cli` was part of the binary package; now it's exposed as part of the library through the `commands` module, so we can forward calls from `nargo` by using it as a dependency and just mapping command parameters. 
 
Other options would be to:
* "shell out" to the `noir-artifact execute` as a sub-process 
* use the functions in `noir_artifact_cli::executions` 

The reason I thought creating and running `ExecuteCommand` would be good are:
* With the sub-process option:
  * Aztec packages would need to install both binaries and make `nargo` aware of where `noir-execute` can be found, which requires changing a bunch of scripts that currently use `nargo`
  * We would lose the static typing checks around parameter passing, instead we would need to pass on the subset of CLI options that `noir-execute` expects, which is a bit more error prone
  * We wouldn't be able to pass on custom foreign executors generically, although once we can handle oracle transcripts we might want `nargo execute` to work with them as well, and there wouldn't be a need to differentiate. 
  * The locking handled by `nargo` has to span the compilation and the execution phase; if they are in one process there cannot be any clashes, although at the moment `noir-execute` doesn't do any locking
* With the `execution` functions option there is still some duplication of the workflow, and we're a bit more removed from the conceptual "shelling out"

I mainly wanted to leave aztec-packages untouched for now, but with this change at least it would be simpler as `nargo` does very little on its own.

## Documentation\*

Check one:

# PR Checklist\*

- [ ] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo aartifact` on default settings.


aakoshh:
Hm, I don't understand why the compilation memory would be affected by this PR 🤔 
The execution memory as well, I'm still calling the same things from nargo ops.

TomAFrench (approved):
A little concerned that we're building up a number of redundant versions of `ProgramArtifact` but we can deal with that later.

aakoshh:
> we're building up a number of redundant versions of `ProgramArtifact`

Which one do you mean? Getting rid of redundancy was my motivation for getting rid of the `Circuit` in this crate and forcing things back to be `CompiledProgram`.

TomAFrench:
Sorry, my comment isn't specific to this PR. Just something I'm thinking about related to the CLI in general recently, I think we're on the same page.

PR #7409, opened 2025-02-17
feat(LSP): auto-import via visible reexport

### LINKED ISSUE #7378: LSP auto-imports do not correctly handle re-exports
In Aztec contracts we generally have protocol types exposed as a re-export from Aztec.nr and the contracts do not import that crate directly. When I use LSP to fix a missing import it does not use the re-export:

Image

### Nargo Version

nargo version = 1.0.0-beta.1 noirc version = 1.0.0-beta.1+3f8376928465907d (git version hash: 3f8376928465907d, is dirty: false)

asterite:
# Description

## Problem

Resolves #7378

## Summary

This is a new feature but also a bug fix.

The bug was that if a type was directly visible from the current module it was offered for autocompletion. However, that didn't take into account if the type was actually in a crate that's in one of the current crate's dependencies. That's why in the original issue `protocol_types::traits::Packable` was offered: if it were a dependency it's good to suggest that... but it's not!

Then the feature is that we now track all of an item's reexports (not just traits). Then, if the item isn't directly visible we search for that item's parent module's reexports. If there's one visible, we can suggest importing the item via the parent module's reexport. This is done recursively so that if the parent doesn't have a reexport but the grandparent module has one, that's used instead.

Here's how `Packable` is imported now:

lsp-import-via-reexport

## Additional Context

The changes here are bigger than just the above things because I also took the time to refactor and remove some code duplication that existed, and that now would have gotten even bigger. I tried to document the code as much as possible because the logic is getting more and more complex by the minute.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/def_collector/dc_crate.rs
+++ b/compiler/noirc_frontend/src/hir/def_collector/dc_crate.rs
@@ -3,7 +3,7 @@ use super::errors::{DefCollectorErrorKind, DuplicateType};
 use crate::elaborator::Elaborator;
 use crate::graph::CrateId;
 use crate::hir::comptime::InterpreterError;
-use crate::hir::def_map::{CrateDefMap, LocalModuleId, ModuleDefId, ModuleId};
+use crate::hir::def_map::{CrateDefMap, LocalModuleId, ModuleId};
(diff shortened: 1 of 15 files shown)


jfecher (approved):
:+1:

PR #7415, opened 2025-02-18
chore: bump aztec-packages commit

TomAFrench:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/scripts/bump-aztec-packges-commit.sh
+++ b/scripts/bump-aztec-packges-commit.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+AZTEC_COMMIT=$(git ls-remote https://github.com/AztecProtocol/aztec-packages.git HEAD | grep -oE '^\b[0-9a-f]{40}\b')
+
+function bump_commit() {
+    FILE=$1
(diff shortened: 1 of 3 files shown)


scripts/bump-aztec-packges-commit.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+AZTEC_COMMIT=$(git ls-remote https://github.com/AztecProtocol/aztec-packages.git HEAD | grep -oE '^\b[0-9a-f]{40}\b')
+
+function bump_commit() {
+    FILE=$1
+    sed -E -i.bak "s/(^define: &AZ_COMMIT) .*/\1 $AZTEC_COMMIT/" $FILE  && rm $FILE.bak
```
aakoshh, line 7: This looks a bit fragile. At first glance `define:` sounds like something under which could be multiple variables, but it's just a commit. Someone not familiar with the implications could change it like this:

```
define:
  az_commit: &AZ_COMMIT abc123
  default_timeout: &DEFAULT_TIMEOUT 100
```
and then they wouldn't know that this script would (silently?) stop doing anything. 

Since we have structured configuration files, you could do the above now and then use `yq` to [update the value](https://mikefarah.gitbook.io/yq/operators/assign-update#update-string-value-via-or)

aakoshh (approved):

Issue #7382, opened 2025-02-14
Add an example to test the artifact CLI on CI


aakoshh:
Create a new module under https://github.com/noir-lang/noir/tree/master/examples that shows how to use the artefact CLI to execute oracle transcripts:
* Create a contract that uses the mock functionality to make oracle calls in tests
* Somehow capture the returned values from the mocks and save them as an `Oracle.toml` transcript file
* Add a `test.sh` script to execute the following workflow:
  * Test the contract with `nargo test`
  * Compile the contract with `nargo compile`
  * Execute the compiled artefact with the `Prover.toml` and `Oracle.toml` files using `noir-artifact execute`
* Call `test.sh` on CI

PR #7417, opened 2025-02-18
feat(cli): Log and replay oracle transcript

### LINKED ISSUE #7379: Execute artefact with oracle transcript
Read an oracle transcript from a file and return the items as responses to foreign calls. This should allow us to replay contract calls captured on aztec packages, once Nico's team adds the feature to dump a session.

aakoshh:
# Description

## Problem\*

Resolves #7379 and #7382 

## Summary\*

* Added `NARGO_TEST_FOREIGN_CALL_LOG` to `nargo test` which can be used to log the foreign calls made by a test either to `stdout` or a file. It only makes sense with a single test. The format of the log is JSON Line with a `call` and `response` as it appears in the JSON-RPC.
* Added a `LoggingForeignCallExecutor` that prints foreign calls to a transcript log.
* Added a `ReplayForeignCallExecutor` that deserialises the transcript, and implements the handler by popping off the next call, comparing the logged function name and inputs, rejecting any unexpected call to avoid confusion, then returns the logged response.
* Added `--oracle-file` to `noir-execute` and `nargo execute` to be able to point at the transcript
* Changed `noir-execute` to use the `ReplayForeignCallExecutor` as the final layer in the foreign call executor stack, if the file is present.
* Added `examples/oracle_transcript` based on `test_programs/noir_test_success/mock_oracle`, which calls `nargo test` to capture a transcript and verify the expected input/output, then `nargo execute` with the transcript file to verify the expected `return` value in `Prover.toml`. The oracle file is checked in for reference.
* Repeats the same call with `noir-execute` to demonstrate how it's used.
* Build the `noir-execute` artifact on CI to be able to use it in the test script.

## Additional Context

Here's a link to show the new example passing on CI: https://github.com/noir-lang/noir/actions/runs/13394747019/job/37411091682

The format in this PR, or something similar, will be the one used in https://github.com/AztecProtocol/aztec-packages/issues/11855 

## Documentation\*

Check one:

# PR Checklist\*


asterite:
> Added NARGO_TEST_FOREIGN_CALL_LOG to nargo test which can be used to log the foreign calls made by a test either to stdout or a file. It only makes sense with a single test.

I'm about to start reviewing this PR, but why does it only make sense with a single test? I know oracle calls from multiple tests might lead to different results, but I was wondering if we could key them under the test name.

asterite (approved):
Looks good! I guess, if it's possible, we could support running multiple tests in a future PR (is it an error right now if you run it with multiple tests?)

aakoshh:
> I'm about to start reviewing this PR, but why does it only make sense with a single test?

Only because I didn't spend any effort to delineate them:
* if we set `NARGO_TEST_FOREIGN_CALL_LOG=stdout` then it will print things as it goes
* if we set `NARGO_TEST_FOREIGN_CALL_LOG=path/to/file.jsonl` then the last test will overwrite the previous ones
* there is no option to log to a string and print it to the stdout as one unit like you are doing it with the test result reporter for `println` collected during execution

aakoshh:
I didn't want to change the signature of `nargo::ops::run_test` to add another `PrintOutput` for these, and also didn't want to have to copy paste from the terminal into a file, but if we think this is useful, or should be displayed along with the `println` output of the test, I'm happy to revisit. Currently it was just an idea for getting something to work with.

aakoshh:
The TypeScript side of the transaction dump capturing the transcript can be seen in https://github.com/AztecProtocol/aztec-packages/pull/12148

PR #7432, opened 2025-02-18
feat: simplify assertions that squared values are equal to zero

TomAFrench:
# Description

## Problem\*

Something I noticed when reviewing #7301 

## Summary\*

We currently don't simplify 

```
 v1 = mul v0, v0
constrain v1 == u1 0
```

into

```
constrain v0 == u1 0
```

I've then added this as an optimization when decomposing constraints.

This looks like it should be safe to me due to us using a prime field so there's no nonzero solutions to `v0 * v0 == 0`.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/instruction/constrain.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/instruction/constrain.rs
@@ -177,3 +177,31 @@ pub(super) fn decompose_constrain(
         }
     }
 }
+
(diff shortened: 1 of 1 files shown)


asterite:
I was going to suggest adding a test but I wasn't sure it was possible, but it is with `from_str_simplifying`. Given that I already wrote the test at that point, I pushed it 😅

asterite (approved):

PR #7434, opened 2025-02-18
fix(performance): Remove redundant slice access check from brillig

vezenovm:
# Description

## Problem\*

No issue, small optimization I found while exploring other Brillig gen optimizations.

## Summary\*

For every slice access, we code gen a slice access check against its dynamic index: https://github.com/noir-lang/noir/blob/988adae231c1aca09fd4d08d5e7bf5e72deecf9d/compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs#L475

In Brillig gen we use `!dfg.is_safe_index` to determine whether to code gen an array index. This always returns false for slice types as we cannot check against their length. However, we already have generated a check the precedes the array access. We should generate this validation code for array types in Brillig gen.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench (approved):

TomAFrench:
Actually, looks like we should match on `array_variable` inside of `validate_array_index` and early return on vectors to encapsulate the check.

Issue #6410, opened 2024-10-30
Document the noir-profiler


vezenovm:
### Problem

A few months ago we added a profiler under `tooling/profiler`. The profiler has been used almost exclusively internally. This profiler allows devs generate flamegraphs of the opcodes and backend constraints laid down by a given program. Just recently the profiler was expanded to also provide information on which part of the program generated the most steps in the Brillig VM. 

### Happy Case

We should document the profiler inside the Noir repo and externally to allow developers to start to utilize this very useful tool. 

### Workaround

None

### Project Impact

Nice-to-have

### Would you like to submit a PR for this Issue?

None

Savio-Sou:
Does the profiler also analyze witness generation?

Savio-Sou:
From [HashCloak's noir-mpc-ml](https://github.com/hashcloak/noir-mpc-ml) project report, if useful as a documentation starter:
> You can find the profiler tool [here](https://github.com/noir-lang/noir/tree/master/tooling/profiler). These are the steps we took to make it work:
> 
> 1. Git clone the Noir repo
> 2. Checkout the tag of the nargo version you intend to use. Make sure the installed `bb` version aligns with it.
> 3. Set `PATH_TO_ARTIFACT` to where the `json` of your compiled circuit is located, for example `target/test.json`.
> 4. Set `PATH_TO_OUTPUT` to where the `svg` file will be saved, for example `/output`.
> 5. Set `PATH_TO_BB` to where `bb` is installed.
> 6. Obtain gates flamegraph:
> ```sh
> cargo run --package noir_profiler --bin noir-profiler  gates-flamegraph --artifact-path PATH_TO_ARTIFACT --output PATH_TO_OUTPUT --backend-path PATH_TO_BB
> ```
> 7. Obtain opcodes flamegraph
> ```sh
> cargo run --package noir_profiler --bin noir-profiler  opcodes-flamegraph --artifact-path PATH_TO_ARTIFACT --output PATH_TO_OUTPUT
> ```

vezenovm:
> Does the profiler also analyze execution / witness generation?

It will analyze the number of Brillig opcodes executed but only when the main is force compiled to Brillig. It does not analyze ACVM witness generation and Brillig execution within an ACIR program.

PR #7436, opened 2025-02-19
chore(profiler): Docs on profiler command and more complete error reporting

vezenovm:
# Description

## Problem\*

Towards #6410 but only on the CLI commands themselves

## Summary\*

Add documentation on commands for main help command:
```
Profiler for noir circuits

Usage: noir-profiler <COMMAND>

Commands:
  gates              Generates a flamegraph mapping backend opcodes to their associated locations in the source code
  opcodes            Generates a flamegraph mapping ACIR opcodes to their associated locations in the source code
  execution-opcodes  Generates a flamegraph mapping unconstrained Noir execution to source code
  help               Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help
  -V, --version  Print version
```
All the commands previously had no help information. 

I also added some initial error reporting for the CLI. 

Previously we would panic if someone supplied a program artifact to `execution-opcodes` that was not compiled with an unconstrained main function. 
```
Executing
Executed
Collecting 0 samples
Generating flamegraph with 0 samples
2025-02-18T23:42:43.252759Z ERROR inferno::flamegraph: No stack counts found    
I/O error: No stack counts found

Caused by:
    No stack counts found

Location:
    tooling/profiler/src/flamegraph.rs:122:9
```
This provides little information to the user and also opens up the internals of the profiler. 
We now report the following:
Screenshot 2025-02-18 at 7 50 50 PM

I also added reporting in case of an execution error. We would previously just fail with "Failed to solve Brillig function". Now we utilize the same reporting from the execute command.
Here is an example failure:
Screenshot 2025-02-18 at 7 51 57 PM

## Additional Context

I held off on improving error reporting for reading artifacts as that can be done as part of https://github.com/noir-lang/noir/issues/7381.

## Documentation\*

Check one:

# PR Checklist\*


tooling/profiler/src/cli/execution_flamegraph_cmd.rs
```diff
@@ -104,3 +132,22 @@ fn run_with_generator(
 
     Ok(())
 }
+
+fn is_brillig_entry_point(artifact: &ProgramArtifact) -> Result<(), CliError> {
```
michaeljklein, line 136: From the name, I expected it to return a `bool`:
```suggestion
fn ensure_brillig_entry_point(artifact: &ProgramArtifact) -> Result<(), CliError> {
```

tooling/profiler/src/errors.rs
```diff
@@ -0,0 +1,15 @@
...
+
+#[derive(Debug, Error)]
+pub(crate) enum CliError {
+    #[error("Failed to run profiler command")]
+    Generic,
+}
+
+pub(crate) fn report_error(message: String) -> Result<(), CliError> {
+    let error = CustomDiagnostic::simple_error(message.clone(), String::new(), Span::default());
+    noirc_errors::reporter::report(&FileMap::default(), &error, None, false);
```
michaeljklein, line 13: Does the `FileMap::default()` work on errors from dependencies in multi-crate projects? I would expect it to need to be initialized for accurate reporting, as is done [here](https://github.com/noir-lang/noir/blob/1d04f8ba0292e493e4e64cf8caf6df15c51b3346/tooling/nargo_cli/src/cli/execute_cmd.rs#L172C27-L172C61).

michaeljklein (changes requested):

tooling/profiler/src/cli/execution_flamegraph_cmd.rs
```diff
@@ -54,17 +59,40 @@ fn run_with_generator(
     let program =
         read_program_from_file(artifact_path).context("Error reading program from file")?;
 
+    is_brillig_entry_point(&program)?;
```
vezenovm, line 62:
```suggestion
    ensure_brillig_entry_point(&program)?;
```

tooling/profiler/src/errors.rs
vezenovm: The profiler uses binary artifacts, so a user is going to have to point to the correct artifact if they are working within a workspace. I did `FileMap::default()` as this allows us to report in cases where we do not have a stack trace and avoiding what is done [here](https://github.com/noir-lang/noir/blob/1d04f8ba0292e493e4e64cf8caf6df15c51b3346/tooling/nargo_cli/src/cli/execute_cmd.rs#L172C27-L172C61).

I'll add a comment.

tooling/profiler/src/errors.rs
```diff
@@ -0,0 +1,15 @@
...
+use noirc_errors::{CustomDiagnostic, Span};
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub(crate) enum CliError {
+    #[error("Failed to run profiler command")]
+    Generic,
+}
+
+pub(crate) fn report_error(message: String) -> Result<(), CliError> {
```
vezenovm, line 11:
```suggestion
/// Report an error from the CLI that is not reliant on a stack trace. 
pub(crate) fn report_error(message: String) -> Result<(), CliError> {
```

TomAFrench (approved):
LGTM, let's get this in before 1.0.0-beta.3

TomAFrench:
@michaeljklein can you remove your block?

michaeljklein (approved):

PR #7437, opened 2025-02-19
fix: require loop/for/while body to be unit

### LINKED ISSUE #7408: bug: `loop`, `for` and `while` don't require their body to be `()`
### Aim

Given this code:

```noir
fn main() {
    for _ in 0..10 {
        10
    }
}
```

### Expected Behavior

The compiler should give an error saying that the `for` body's type should be `()`, not `Field`.

### Bug

The code gives no error (unlike Rust).

The same applies to `loop` and `while`.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #7408

## Summary

In addition to that:
1. It fixes an issue where the error span when the "then" and "else" branches of an "if" didn't match was wrong (I introduced that bug some weeks ago)
2. A new feature where some type errors will now point to more precise spans

For point 2, here's the error you got when the branches of an "if" didn't match:

image

And this is how it looks now:

image

That is, for a given expression we now point the type mismatch in the innermost expression that gives it its type.

The same is done for loops:

image

I made this change in places where a block usually happens. We could eventually do it in other places (prefix expression, index expressions, call arguments, etc.) as we find them, though using blocks there is pretty rare.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/ast/expression.rs
+++ b/compiler/noirc_frontend/src/ast/expression.rs
@@ -245,6 +245,45 @@ impl Expression {
     pub fn new(kind: ExpressionKind, span: Span) -> Expression {
         Expression { kind, span }
     }
+
+    /// Returns the innermost span that gives this expression its type.
+    pub fn type_span(&self) -> Span {
+        match &self.kind {
+            ExpressionKind::Block(block_expression)
+            | ExpressionKind::Comptime(block_expression, _)
+            | ExpressionKind::Unsafe(block_expression, _) => {
+                if let Some(statement) = block_expression.statements.last() {
+                    statement.type_span()
+                } else {
+                    self.span
+                }
+            }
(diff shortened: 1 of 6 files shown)


TomAFrench:
Can you bump that compilation timeout to 110?

TomAFrench (approved):
LGTM

asterite:
Ah, sorry. Today I have a meeting in my kid's school, I'll continue working a bit later. I'll send a separate PR.

TomAFrench:
Ah nah, I bumped it in #7441 so no worries.

PR #7444, opened 2025-02-19
fix: don't let nargo fmt produce multiple trailing newlines

asterite:
# Description

## Problem

Just a tiny thing: I noticed `nargo fmt` would end up with two trailing newlines at the end of the file if you have two or more newlines.

## Summary

I think the correct thing is to end up with just one newline, so this is what this PR does.

## Documentation

Check one:

# PR Checklist


jfecher (approved):
:+1:

Issue #7423, opened 2025-02-18
Add the ability to have unstable features


jfecher:
### Problem

Nargo currently only supports what we call "experimental features" where when used they will always issue an experimental feature warning but never stop compilation. We'd like to make this process opt-in so that even if a user uses a crate as a library which uses one of these features and they are ignoring warnings, they will still be alerted their library uses an experimental feature and would need to decide whether to opt-in or not.

Without this feature it is significantly more difficult to extend the language post 1.0.

### Happy Case

Gating unstable features could function similarly to unstable features in cargo detailed here: https://doc.rust-lang.org/cargo/reference/unstable.html. One choice we could make is whether to merge this cli flag approach with the `#[feature(name)]` attribute Rust uses. The command-line flag would be a more broadly applicable approach, but controlling this in source code could be more clear for users coming from Rust. Note that `#[feature(..)]` is distinct from `#[cfg(feature(...))]` which has its own issue [here](https://github.com/noir-lang/noir/issues/6963).

We should test this out using enums before they are stabilized.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #7449, opened 2025-02-19
feat(cli)!: Add `--unstable-features` to gate unstable features

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/7423

## Summary\*

Adds the CLI flag `--unstable-features` (or `-Z`) to nargo for all compile sub-commands which accepts a list of unstable features to enable. Without this, usage of these features is now a hard error (previously a warning), making this a breaking change. The list of features to accept can be separated by commas: `-Zenums,array-ownership` or by specifying the same flag multiple times: `-Zenums -Zarray-ownership`.

Currently the only unstable features are:
- Enums
- Array ownership (reserved for future experimentation with `&` and move semantics for arrays)

## Additional Context

I'm using the name "unstable features" to match rust a bit more closely although previously we used "experimental features." I'm open to naming debates since this is a public facing flag we shouldn't change once it is stabilized.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
Hmm looks like CI is broken for some reason because I added a new workspace package

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -194,8 +193,8 @@ pub struct Elaborator<'context> {
     /// that comptime value and any visibility errors were already reported.
     silence_field_visibility_errors: usize,
 
-    /// Use pedantic ACVM solving
-    pedantic_solving: bool,
+    /// Options from the nargo cli
+    cli_options: ElaboratorOptions<'context>,
```
aakoshh, line 197: I always think about CLIs as the outer layer about which the core should not know about. It was already a bit strange that `CompileOptions` had `clap` annotations, although useful for defaults, but to see actual `cli_options` and `nargo` referred to in the `noirc_*` modules is strange.

aakoshh: Do we really need a separate crate for them though? Could they not be defined next to `CompileOptions`?

aakoshh: (By the way the `utils/cli_args` crate seems to be missing from the commits).

Cargo.toml
```diff
@@ -109,6 +110,7 @@ ark-std = { version = "^0.5.0", default-features = false }
 
 # Misc utils crates
 iter-extended = { path = "utils/iter-extended" }
+cli-args = { path = "utils/cli-args" }
```
vezenovm, line 113: I don't see a `cli-args` package pushed on this PR

compiler/noirc_frontend/src/elaborator/mod.rs
jfecher: I spent far too long trying to figure out what this meant only to realize it was never added to git... :man_facepalming:.
Should be added now.

As for the reason why I needed a new crate: it's because `UnstableFeature` is shared between the driver and the frontend so it could only be defined in the frontend since the frontend can't depend on the driver. However this type is also a CLI input and requires clap to derive arguments so I thought it'd be cleaner as a new crate rather than adding clap and CLI functionality to the frontend crate.

aakoshh: Sorry, bad phrasing 😄

Right, I see you have the common UnstableFeature type there. I think we could have an UnstableFeature enum for the compiler front end, and other crates would have their own, so conceptually the front end doesn’t need to know about every potentially feature across the repo. And then the CLI can handle the a union and the mappings. 

I share your feelings about clap encroaching on core libraries. I saw people use I think strum to map from strings to enums, which was to support CLIs. In any case depending on a `cli_args` crate is perhaps even more pronounced, but maybe that’s just me.

utils/cli-args/src/lib.rs
```diff
@@ -0,0 +1,40 @@
+#[derive(Copy, Clone, Debug, clap::ValueEnum, PartialEq, Eq)]
+pub enum UnstableFeature {
+    Enums,
+    ArrayOwnership,
+}
+
+impl std::fmt::Display for UnstableFeature {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        // Note: clap::ValueEnum uses kebab-case by default so
```
aakoshh, line 9: https://difftastic.wilfred.me.uk/rustdoc/strum/derive.EnumString.html seems to allow a lot of control, could be an alternative to clap?

compiler/noirc_frontend/src/parser/parser/enums.rs
```diff
@@ -23,8 +23,6 @@ impl<'a> Parser<'a> {
     ) -> NoirEnumeration {
         let attributes = self.validate_secondary_attributes(attributes);
 
-        self.push_error(ParserErrorReason::ExperimentalFeature("Enums"), start_span);
```
michaeljklein, line 26: Was this moved to the elaborator because the experimental features list is more accessible from there?
It makes a small gap where a user could avoid realizing that enums are experimental and spend some time fixing parser errors before getting the "at your own risk" warning.

utils/cli-args/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+    /// that are not in this list result in an error when used.
+    pub enabled_unstable_features: &'a [UnstableFeature],
+}
+
+impl<'a> FrontendOptions<'a> {
+    pub fn test_default() -> FrontendOptions<'static> {
+        FrontendOptions {
+            debug_comptime_in_file: None,
+            pedantic_solving: true,
+            enabled_unstable_features: &[UnstableFeature::Enums],
```
michaeljklein, line 37: Please add tests for enum definitions and match statements where the `UnstableFeature::Enums` is disabled

michaeljklein (changes requested):

utils/cli-args/src/lib.rs
jfecher: Not sure how it'd be an alternative to clap if it doesn't handle command line arg parsing. Clap would still need a value_parser

compiler/noirc_frontend/src/hir/def_collector/dc_crate.rs
```diff
@@ -254,6 +255,31 @@ impl From<UnsupportedNumericGenericType> for CompilationError {
...
+    /// that are not in this list result in an error when used.
+    pub enabled_unstable_features: &'a [UnstableFeature],
+}
+
+impl<'a> ElaboratorOptions<'a> {
+    pub fn test_default() -> ElaboratorOptions<'static> {
+        let options = FrontendOptions::test_default();
+        ElaboratorOptions {
+            debug_comptime_in_file: None,
+            pedantic_solving: options.pedantic_solving,
```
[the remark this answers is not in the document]
jfecher: It is on, this just forwards the defaults from `FrontendOptions::test_default` so you don't have to edit 2 places for the same defaults

utils/cli-args/src/lib.rs
aakoshh: What I mean is that you can decide how strings are parsed/displayed here, and we can elsewhere, in the actual CLI, have another construct dedicated where we specify the `value_parser` to be based on strum.

compiler/noirc_frontend/src/parser/parser/enums.rs
jfecher: I added the options struct to the parser originally but it ended up really poluting the codebase since we instantiate parsers in many places (e.g. comptime code) and the elaborator ended up needing to carry around these options anyway to forward them. I figured it was simpler to move it to the elaborator.

For errors I don't think it is too large a concern since in principle we should try to recover on error and the user will still see it soon after if we don't. The use case I was considering when writing this was the formatter since that wouldn't error if formatting code with experimental features versus it would before when these warnings were in the parser. I think this is fine though since we're mostly concerned about the security of executing code with half-baked features in them - formatting such code should still be safe.

utils/cli-args/src/lib.rs
jfecher: Added one

utils/cli-args/src/lib.rs
aakoshh: https://github.com/clap-rs/clap/discussions/4264 Shows a somewhat verbose example

michaeljklein (approved):
LGTM

compiler/noirc_frontend/src/elaborator/mod.rs
jfecher: Ok, I managed to reorganize the crates a bit to remove the args-cli crate. We can implement `FromStr` on the type in the frontend and use `value_parser!` on clap in the driver to use that impl and keep them separate.

This took quite a bit longer than I expected though since I was trying to use `FromStr::Error = ()` but clap emits a very cryptic error and it turns out that the unit type is invalid there. I switched it to String and it is now compiling. I'm keeping the same UnstableFeature enum in the CLI but in the future we could create a wrapper enum combining other enums when needed. We'd just need to manually implement ValueParser at that point.

PR #7456, opened 2025-02-20
fix: prevent incorrect ACIRgen caused by noop truncations

### LINKED ISSUE #7451: Wrong evaluation of `&` on `u8` values
### Aim

Wrong result for `y` and different output for `nargo test` and `nargo execute`.

**main.nr**
~~~noir
use dep::std;

pub fn main(x : Field) {
    let mut y : Field = 0;
    let mut tmp : Field = 0;

    if 1.lt(2) {
        tmp = 0xff;
    }
    else {}

    let tmp_bytes : [u8; 32] = tmp.to_be_bytes();
    let x_bytes : [u8; 32] = x.to_be_bytes();
    let mut y_bytes : [u8; 32] = 0.to_be_bytes();

    for i in 0..32 {
        y_bytes[i] = (tmp_bytes[i] & x_bytes[i]);
    }

    y = std::field::bytes32_to_field(y_bytes);
    print("y => ");
    println(y);
}

#[test]
fn test_main() {
    let _ = main(1);
}
~~~

**Prover.toml**
~~~toml
x="1"
~~~

It seems that a wrong optimization is causing the problem. If I remove the `if` statement, the code executes as expected.

### Expected Behavior

I expected `y` to be `0x01` and the test and execute output to match.

### Bug

`nargo execute` results in `y => 0x00`
`nargo test --show-output` results in `y => 0x01`

### To Reproduce

1. create a project with the given files
2. `nargo execute` 
3. `nargo test --show-output`
4. compare the outputs

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.2 noirc version = 1.0.0-beta.2+c63cc73313690dfd3ab0e4b190c5b6013dd6af03 (git version hash: c63cc73313690dfd3ab0e4b190c5b6013dd6af03, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

TomAFrench:
# Description

## Problem\*

Resolves #7451 

## Summary\*

This PR fixes an issue where doing a bitwise AND which covers the full width of its type will result in a truncation instruction which results in invalid ACIR.

I've fixed this by avoiding creating this bad instruction but also added checks in acirgen so that we'll reject any invalid truncations and also any other divisions by constant values which are greater than the allowed type.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -2094,6 +2094,12 @@ impl<'a> Context<'a> {
         max_bit_size: u32,
         dfg: &DataFlowGraph,
     ) -> Result<AcirVar, RuntimeError> {
+        assert_ne!(bit_size, max_bit_size, "Attempted to generate a noop truncation");
+        assert!(
+            bit_size < max_bit_size,
+            "Attempted to generate a truncation into large size than max input"
```
TomAFrench, line 2100:
```suggestion
            "Attempted to generate a truncation into size larger than max input"
```

jfecher (approved):
LGTM

PR #7470, opened 2025-02-20
chore(profiler): Use brillig names for outputted flamegraphs 

vezenovm:
# Description

## Problem\*

No issue, just a QOL improvement I found while doing other documentation.

## Summary\*

We are using the brillig function index when writing a flamegraph file. 
e.g. Running `noir-profiler opcodes` on this program:
```noir
fn main(ptr: pub u32, mut array: [u32; 5]) -> pub [u32; 5] {
    // Safety: Sets all elements after `ptr` in `array` to zero.
    let new_array = unsafe { write_to_array(ptr, array) };
    for i in 0..5 {
        if i > ptr {
            assert_eq(new_array[i], 0);
        } else {
            assert_eq(new_array[i], array[i]);
        }
    }
    new_array
}
unconstrained fn write_to_array(ptr: u32, mut array: [u32; 5]) -> [u32; 5] {
    for i in 0..5 {
        if i > ptr {
            array[i] = 0;
        }
    }
    array
}
```
Will result in the following flamegraph files: `main_opcodes.svg`, `0_brillig_opcodes.svg`, `1_brillig_opcodes.svg`, `2_brillig_opcodes.svg`. The indices are pretty meaningless without a good amount extra context that is not reasonable to place on users.

Here is how the output directory looks after this PR:
Screenshot 2025-02-20 at 3 50 19 PM

I also changed the flamegraph title:
Screenshot 2025-02-20 at 3 51 24 PM

Before this PR the title would have just been `./target/program.json-brillig_0`.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/profiler/src/cli/gates_flamegraph_cmd.rs
+++ b/tooling/profiler/src/cli/gates_flamegraph_cmd.rs
@@ -4,7 +4,6 @@ use acir::circuit::OpcodeLocation;
 use clap::Args;
 use color_eyre::eyre::{self, Context};
 
-use fxhash::FxHashMap as HashMap;
 use noirc_artifacts::debug::DebugArtifact;
 
 use crate::flamegraph::{CompilationSample, FlamegraphGenerator, InfernoFlamegraphGenerator};
@@ -73,22 +72,17 @@ fn run_with_provider<Provider: GatesProvider, Generator: FlamegraphGenerator>(
 
     let debug_artifact: DebugArtifact = program.into();
 
-    // We can have repeated names if there are functions with the same name in different
-    // modules or functions that use generics.
-    let mut seen_names: HashMap<String, usize> = HashMap::default();
-    for (func_idx, (func_gates, bytecode)) in
+    let num_functions = bytecode.functions.len();
+    for (func_idx, (func_gates, circuit)) in
         backend_gates_response.functions.into_iter().zip(bytecode.functions).enumerate()
     {
-        let original_name = function_names[func_idx].as_str();
-        let mut function_name = original_name.to_owned();
-
-        let mut counter = seen_names.get(original_name).copied().unwrap_or(0);
-        // Ensure uniqueness by checking existing names
-        while seen_names.contains_key(&function_name) {
-            counter += 1;
-            function_name = format!("{}_{}", original_name, counter);
-        }
-        seen_names.insert(function_name.clone(), counter);
+        // We can have repeated names if there are functions with the same name in different
+        // modules or functions that use generics. Thus, add the unique function index as a suffix.
+        let function_name = if num_functions > 1 {
+            format!("{}_{}", function_names[func_idx].as_str(), func_idx)
+        } else {
+            function_names[func_idx].to_owned()
+        };
 
         println!(
             "Opcode count: {}, Total gates by opcodes: {}, Circuit size: {}",
@@ -100,7 +94,7 @@ fn run_with_provider<Provider: GatesProvider, Generator: FlamegraphGenerator>(
         let samples = func_gates
             .gates_per_opcode
             .into_iter()
-            .zip(bytecode.opcodes)
+            .zip(circuit.opcodes)
             .enumerate()
             .map(|(index, (gates, opcode))| CompilationSample {
                 opcode: Some(format_acir_opcode(&opcode)),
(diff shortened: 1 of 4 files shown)


TomAFrench:
How does this handle generic functions or functions with the same name in different modules?

vezenovm:
> How does this handle generic functions or functions with the same name in different modules?

This is going to override repeated names, however, this is already an issue with the ACIR flamegraph (although much more rare as it requires multiple ACIR functions) which uses the function name.

Similarly `nargo info` just repeats the name:
Screenshot 2025-02-20 at 4 18 59 PM

I'll look at a clean way of representing these function names, although we do lose a lot of the information by the time we get to Brillig/ACIR gen. 

For now, I have just pushed a bit of a hack that checks for seen names and generates a suffix to attach to the previously seen function name. 

Here is some example output:
Screenshot 2025-02-20 at 4 44 37 PM

jfecher:
> For now, I have just pushed a bit of a hack that checks for seen names and generates a suffix to attach to the previously seen function name.

Do we check if the suffix already exists as well? E.g. if we not only have two duplicate `write_to_array` functions, but also have a `write_to_array1` function defined?

vezenovm:
> Do we check if the suffix already exists as well? E.g. if we not only have two duplicate `write_to_array` functions, but also have a `write_to_array1` function defined?
> 
> >

The hack tracks a map of seen names -> suffixes. So it will see multiple duplicates and apply the appropriate suffix. Another example where I use a third duplicate `write_to_array`:
Screenshot 2025-02-21 at 10 17 25 AM

> Koality of life? 🐨

🤣 I was definitely typing too fast. Edited the description

jfecher:
> Another example where I use a third duplicate write_to_array:

I don't mean a third duplicate, I mean still two `write_to_array` and one function called in source as `write_to_array_1` already such that it collides with the default disambiguation strategy. This can potentially happen N times

vezenovm:
> I don't mean a third duplicate, I mean still two `write_to_array` and one function called in source as `write_to_array_1` already such that it collides with the default disambiguation strategy. This can potentially happen N times

Ah I see I read that wrong. Yeah good call out. I'll modify to cover that case.

tooling/profiler/src/cli/gates_flamegraph_cmd.rs
```diff
@@ -75,15 +75,20 @@ fn run_with_provider<Provider: GatesProvider, Generator: FlamegraphGenerator>(
...
-            if *count == 0 { func_name.to_owned() } else { format!("{}_{}", func_name, count) };
+        let original_name = function_names[func_idx].as_str();
+        let mut function_name = original_name.to_owned();
+
+        let mut counter = seen_names.get(original_name).copied().unwrap_or(0);
+        // Ensure uniqueness by checking existing names
+        while seen_names.contains_key(&function_name) {
+            counter += 1;
+            function_name = format!("{}_{}", original_name, counter);
+        }
```
jfecher, line 90: We could also avoid the loop entirely if we separate with a suffix normal names can't use. E.g. `write_to_file.1.svg`

jfecher (approved):
LGTM. The while loop could conceptually slow down code but only on inputs where users would go out of their way to name most functions with a numbering scheme. That shouldn't happen in practice so this is fine for a quick tooling feature.

vezenovm:
> LGTM. The while loop could conceptually slow down code but only on inputs where users would go out of their way to name most functions with a numbering scheme. That shouldn't happen in practice so this is fine for a quick tooling feature.

Yeah I figured this case would be quite rare.

tooling/profiler/src/cli/gates_flamegraph_cmd.rs
vezenovm: I also looked making the names unique by just associating the brillig fn index with the name automatically, which also would get rid of the loop. So in the situation you laid our earlier we would end up with `write_to_array_0_brillig`, `write_to_array_1_brillig`, `write_to_array_1_2_brillig`.

vezenovm: I think I may switch to that to avoid the loop. We can then refactor to switch to placing the function id in path's suffix later.

tooling/profiler/src/cli/opcodes_flamegraph_cmd.rs
```diff
@@ -245,11 +274,17 @@ mod tests {
         super::run_with_generator(&artifact_path, &flamegraph_generator, temp_dir.path(), false)
             .expect("should run without errors");
 
-        // Check that the output files ware written to
+        // Check that the output files that were written
```
vezenovm, line 277:
```suggestion
        // Check that the output files were written
```

PR #7482, opened 2025-02-21
fix: don't use dummy location when inserting debug code

### LINKED ISSUE #7479: `nargo debug` jumps to a random file
### Aim

I am trying to use `nargo debug` to debug programs. The bug always occurs no matter the source program, so let's take a look at the minimal possible example:
```rust
fn main() {}
```

### Expected Behavior

When we run `nargo debug` the expected behavior is:
1. With first `next` we step in the body of the function
2. With the second `next` we exit the body of the function
3. With the third `next` we should get the message `Finished execution`  

### Bug

Currently the behavior is as follows:
1. With the first `next` we step in the body of the function
2. With the second `next` we jump into some random function from a random file. Check the provided image for more details.
3. With the third `next` we get the message `Finished execution`

Image for more details:
Image

### To Reproduce

See above.

### Workaround

Yes

### Workaround Description

This is a regression after https://github.com/noir-lang/noir/pull/7333

Reverting it solves the issue.

### Additional Context

_No response_

### Project Impact

Blocker

### Blocker Context

This impeeds the development of our project that is based on `nargo debug`.

### Nargo Version

nargo version = 1.0.0-beta.3 noirc version = 1.0.0-beta.3+c254c3ca07c65c7845d5affb17cb37114c5a3051

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #7479

## Summary

I thought about adding a test for this but the infrastructure for that is in https://github.com/noir-lang/noir/pull/7233 which is not yet merged.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/debug/mod.rs
+++ b/compiler/noirc_frontend/src/debug/mod.rs
@@ -123,7 +123,7 @@ impl DebugInstrumenter {
         let func_body = &mut func.body.statements;
         let mut statements = take(func_body);
 
-        self.walk_scope(&mut statements, func.location.span);
+        self.walk_scope(&mut statements, func.location);
 
         // walk_scope ensures that the last statement is the return value of the function
         let last_stmt = statements.pop().expect("at least one statement after walk_scope");
@@ -139,7 +139,7 @@ impl DebugInstrumenter {
 
     // Modify a vector of statements in-place, adding instrumentation for sets and drops.
     // This function will consume a scope level.
-    fn walk_scope(&mut self, statements: &mut Vec<ast::Statement>, span: Span) {
+    fn walk_scope(&mut self, statements: &mut Vec<ast::Statement>, location: Location) {
         statements.iter_mut().for_each(|stmt| self.walk_statement(stmt));
 
         // extract and save the return value from the scope if there is one
@@ -166,8 +166,8 @@ impl DebugInstrumenter {
             }
         };
 
-        let span = Span::empty(span.end());
-        let location = Location::new(span, FileId::dummy());
+        let span = Span::empty(location.span.end());
+        let location = Location::new(span, location.file);
 
         // drop scope variables
         let scope_vars = self.scope.pop().unwrap_or_default();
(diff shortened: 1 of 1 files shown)


compiler/noirc_frontend/src/debug/mod.rs
```diff
@@ -166,8 +166,8 @@ impl DebugInstrumenter {
             }
         };
 
-        let span = Span::empty(span.end());
-        let location = Location::new(span, FileId::dummy());
+        let span = Span::empty(location.span.end());
```
jfecher, line 169: Why do we use an empty span here instead of taking the original location wholesale?

jfecher (approved):
Approving since we used an empty span before as well. Just curious on the why

compiler/noirc_frontend/src/debug/mod.rs
asterite: Apparently these are inserted for debug calls like "__debug_var_drop".

It looks like this:

```
At opcode 0:0 :: BRILLIG CALL func 0: inputs: [], outputs: []
> next
At opcode 0:0.13 :: Const { destination: Relative(1), bit_size: Integer(U32), value: 0 }
At /Users/asterite/Sandbox/one/src/main.nr:1:11
  1 -> fn main() {
  2        let x = 1;
  3    }
> next
At opcode 0:0.15 :: Const { destination: Relative(2), bit_size: Field, value: 1 }
At /Users/asterite/Sandbox/one/src/main.nr:2:9
  1    fn main() {
  2 ->     let x = 1;
  3    }
> next
At opcode 0:0.18 :: ForeignCall { function: "__debug_var_drop", destinations: [], destination_value_types: [], inputs: [MemoryAddress(Relative(3))], input_value_types: [Simple(Field)] }
At /Users/asterite/Sandbox/one/src/main.nr:3:2
  1    fn main() {
  2        let x = 1;
  3 -> }
```

Or like this:

```
At opcode 0:0 :: BRILLIG CALL func 0: inputs: [], outputs: []
> next
At opcode 0:0.13 :: Const { destination: Relative(1), bit_size: Integer(U32), value: 0 }
At /Users/asterite/Sandbox/one/src/main.nr:1:11
  1 -> fn main() {}
> next
At opcode 0:0.15 :: ForeignCall { function: "__debug_fn_exit", destinations: [], destination_value_types: [], inputs: [MemoryAddress(Relative(1))], input_value_types: [Simple(Integer(U32))] }
At /Users/asterite/Sandbox/one/src/main.nr:1:13
  1 -> fn main() {}
```

I do wonder if calls that don't actually have a source representation should even be shown here. Like maybe after stepping through `let x = 1;` it should exit the program... though maybe it's nice to have a change to see the state right before exiting, not sure.

At least this PR brings back the old functionality.

asterite: We could improve this in a follow-up PR.

PR #7486, opened 2025-02-21
feat: let all compiler errors carry a Location instead of a Span

### LINKED ISSUE #7452: Let compiler errors always carry a Location instead of a Span
Once https://github.com/noir-lang/noir/pull/7333 is merged it would be nice if errors carried `Location`s instead of `Span`s. That way `Elaborator::push_err` would only need to take an error, not a file.

Then, some errors currently hold multiple spans, for example on duplicate errors (a duplication function) and that duplication could happen in different files so moving to `Location` instead of `Span` will also produce more correct error messages in some cases.

Additionally, let `CustomDiagnostic` API work with `Location` instead of `Span`, and make `CustomLabel` hold a `Location` instead of a `Span` and an optional `FileId`.

### LINKED ISSUE #7453: Handle a TODO in the parser when the previous span starts after the current one
This can happen if the previous token location is in a different file. In that case it probably makes sense to not merge the location, but this issue is captured to avoid growing the original PR.

asterite:
# Description

## Problem

Resolves #7452
Resolves #7453

## Summary

Another huge PR, but this one is simpler than the other one (it's more mechanical).

Now that most of the compiler code works with `Location` instead of `Span`, this PR makes all compiler errors carry a `Location` instead of a `Span` (when they carried spans).

With this, a lot is simplified:
1. When pushing errors we no longer need to specify in which file they happen, because each error knows its file
2. Consequently, the errors we track are just `Vec<CompilationError>` instead of `Vec<(CompilationError, FileId)>` because all errors have a location and so have a file.
3. In fact, the field `file` from `Elaborator` is no longer needed, so bugs that would happen because it wasn't updated can't happen anymore
4. Errors that mentioned multiple spans from potentially different files now mention multiple locations so their files can naturally be different

An example of 4 is this code:

```noir
// main.nr
mod other;

use other::Foo;

impl Foo {
    fn foo(self) {}
}

fn main() {}

// other.nr
pub struct Foo {}

impl Foo {
    pub fn foo(self) {}
}
```

The old error is:

```
error: duplicate definitions of foo found
  ┌─ src/other.nr:4:12
  │
4 │     pub fn foo(self) {}
  │            ---   --- first definition found here
  │            │
  │            second definition found here
```

The new error is:

```
error: duplicate definitions of foo found
  ┌─ src/main.nr:6:8
  │
6 │     fn foo(self) {}
  │        --- first definition found here
  │
  ┌─ src/other.nr:4:12
  │
4 │     pub fn foo(self) {}
  │            --- second definition found here
```

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1554,29 +1520,24 @@ impl<'context> Elaborator<'context> {
...
-                        span: self_type_location.span,
-                    },
-                    self_type_location.file,
-                );
+                self.push_err(DefCollectorErrorKind::OverlappingImpl {
+                    typ: self_type.clone(),
+                    location: self_type_location,
+                });
 
+                // TODO: this comment isn't accurate anymore, maybe the two errors can be merged now?
```
michaeljklein, line 1535: Please add a follow-up issue for this TODO

michaeljklein (changes requested):
Remaining TODO but otherwise LGTM

Optionally, the two `Location`'s in `ExpressionKind::Unsafe` could get mixed up so you could:
- Move one of the `Location`'s to `BlockExpression`
- Make `ExpressionKind::Unsafe` a struct variant

compiler/noirc_frontend/src/elaborator/mod.rs
asterite: I ended up fixing it in this PR.

asterite:
@michaeljklein I created an `UnsafeExpression` struct. In the end only one of the locations ended up being used, but I think with the struct it's a bit clearer so I left it.

compiler/noirc_driver/src/lib.rs
```diff
@@ -352,9 +352,10 @@ pub fn check_crate(
     let crate_files = context.crate_files(&crate_id);
     let warnings_and_errors: Vec<FileDiagnostic> = diagnostics
         .into_iter()
-        .map(|(error, file_id)| {
+        .map(|error| {
+            let location = error.location();
             let diagnostic = CustomDiagnostic::from(&error);
-            diagnostic.in_file(file_id)
+            diagnostic.in_file(location.file)
```
jfecher, line 358: Seems like we can get rid of `FileDiagnostic` now, no?
We get the file from the location here than attach the file to the outer diagnostic as well - seems unnecessary.

compiler/noirc_frontend/src/elaborator/types.rs
```diff
@@ -555,24 +521,24 @@ impl<'context> Elaborator<'context> {
...
-                                    ResolverError::BinaryOpError { lhs, op, rhs, err, span },
-                                    location.file,
-                                );
+                                self.push_err(ResolverError::BinaryOpError {
+                                    lhs,
+                                    op,
+                                    rhs,
+                                    err,
+                                    location,
+                                });
```
jfecher, line 541:
```suggestion
                                let error =
                                    ResolverError::BinaryOpError { lhs, op, rhs, err, location };
                                self.push_err(error);
```
fmt nitpick

jfecher (approved):
LGTM - this change has been a long time coming.

Looks like we'll need to bump the maximum memory limit for private-kernel-tail

compiler/noirc_driver/src/lib.rs
asterite: The main issue is that `CustomDiagnostic` doesn't have a main file. All the files are in `secondaries` and that can be empty. There's `CustomDiagnostic::from_message` that ends up with a file, though in almost all places we immediately turn that into a `FileDiagnostic` with a file.

We can add a `file: FileId` field to `CustomDiagnostic` and make sure one is there whenever it's created. What do you think? I'm fine doing it in this PR or in a follow-up one.

jfecher: Ah, ok. It's not as straightforward as I thought then so lets leave it for a separate PR since this one is already rather large.

michaeljklein (approved):
LGTM

PR #7487, opened 2025-02-21
feat(experimental): Add invalid pattern syntax error

jfecher:
# Description

## Problem\*

Resolves some todos in the match compiler

## Summary\*

Replaces some `todo!()`s with an explicit error.

Since we currently parse all expressions in a pattern position and only later try to convert them into a `Pattern`, we have to  issue an error when something unexpected like a `if a { b } else { c } => (),` pattern rule is parsed.

Note that the todo for the constructor case still remains. That's because we should allow matching on struct constructors rather than issue an error for them. That is work for a separate PR though since it requires special handling of them.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/hir/resolution/errors.rs
```diff
@@ -182,6 +182,8 @@ pub enum ResolverError {
     LoopNotYetSupported { span: Span },
     #[error("Expected a trait but found {found}")]
     ExpectedTrait { found: String, span: Span },
+    #[error("Invalid syntax in match pattern")]
+    InvalidSyntaxInPattern { span: Span },
```
vezenovm, line 186: Would be nice to have a unit test for this, but as this is experimental work I won't block

vezenovm (approved):

PR #7488, opened 2025-02-21
fix(frontend)!: Restrict capturing mutable variable in lambdas

### LINKED ISSUE #4795: Closure environments are treated as mutable when they actually aren't
### Aim

Change mutable variables given in environment to lambda.
Mutable variable is created and then given as environment to lambda function. Lambda function changes this variable's value.

### Expected Behavior

Mutable variables change their values outside of lambda function when they are in this function's environment.

### Bug

Variables don't change their value.

### To Reproduce

1. Use repository: https://github.com/vlayer-xyz/repro-derived-constant
2. Run `nargo execute`
3. See the result 

### Project Impact

Nice-to-have

### Impact Context

_No response_

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Installation Method

Binary (`noirup` default)

### Nargo Version

nargo version = 0.26.0 noirc version = 0.26.0+c46b164ce56e6a8f81255fb17eb6539bd040f336

### NoirJS Version

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #4795 

## Summary\*

We now restrict capturing mutable variables that are not mutable references in lambdas. 
This is now banned:
```noir
fn main() {
    let mut x = 3;

    let f = || {
        x += 2;
    };

    f();
    assert(x == 5);
}
```
A mutable reference must be used for `x` instead. 

We can still capture a mutable variable if it is only ever used immutably. 
e.g. this code is still allowed:
```noir
fn main() {
        let mut x = 3;
        let f = || x;
        let _x2 = f();
        assert(x == 3);
   }
```

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/elaborator/expressions.rs
```diff
@@ -1220,7 +1220,20 @@ impl<'context> Elaborator<'context> {
...
+            if let Some(definition) = self.interner.try_definition(capture.ident.id) {
+                if definition.mutable && !typ.is_mutable_ref() {
+                    let location = definition.location;
+                    self.push_err(
+                        TypeCheckError::MutableCaptureWithoutRef {
+                            name: definition.name.clone(),
+                            span: location.span,
+                        },
+                        location.file,
+                    );
```
jfecher, line 1233: Is it necessary to issue an error here? The original intent for lambda captures was not to prevent capturing a mutable variable completely, but rather that any mutable variables should be captured by value and thus be immutable within the lambda. E.g. the equivalent of `let mut x = Foo();   bar(x)` where `fn bar(y: Foo) { ... }`

vezenovm: It is not. This is what we currently allow. Taking the example from https://github.com/noir-lang/noir/issues/4795#issuecomment-2086646284:
```noir
fn main() {
    let mut x = 3;

    let f = || {
        x += 2;
    };

    f();
    assert(x == 5);
}
```
This example will fail on master. `assert(x == 3);` passes.

vezenovm: I thought we desired to have an error, but I could just have this be a warning.

vezenovm: If we wanted to capture a mutable variable immutably it would require devs to reassign their variable before capturing in a lambda. This is a bit cumbersome in my opinion.

vezenovm: Ah I see. I misunderstood as I thought from the original issue you desired this restrictive of a check. I'll update and add this test.

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -84,6 +84,7 @@ pub struct ResolverMeta {
 
 type ScopeForest = GenericScopeForest<String, ResolverMeta>;
 
+#[derive(Clone, Debug)]
```
vezenovm, line 87:
```suggestion
```

vezenovm:
Looks like the dispatch aztec-nr macro is going to need updating.

compiler/noirc_frontend/src/elaborator/statements.rs
```diff
@@ -151,8 +151,24 @@ impl Elaborator<'_> {
...
+            let (id, name, location) = self.get_lvalue_error_info(&lvalue);
+            let typ = self.interner.definition_type(id);
+            for capture_id in capture_ids {
+                if capture_id == id && !typ.is_mutable_ref() {
+                    self.push_err(TypeCheckError::MutableCaptureWithoutRef {
+                        name: name.clone(),
+                        location,
+                    });
+                }
+            }
```
asterite, line 171: I didn't try it but I thought of a way to avoid collecting the capture IDs into a Vec:
```suggestion
            let (id, name, location) = self.get_lvalue_error_info(&lvalue);
            let typ = self.interner.definition_type(id);
            if !typ.is_mutable_ref() && lambda_context.captures.iter().any(|var| var.ident.id == id)
            {
                let name = name.clone();
                self.push_err(TypeCheckError::MutableCaptureWithoutRef { name, location });
            }
```

asterite:
I think this doesn't completely solve the issue, though it solves one case of it.

For example I think this code should also be banned:

```noir
fn main() {
    let mut x = 3;

    let f = || mutate(&mut x);

    f();
    assert(x == 3);
}

fn mutate(x: &mut Field) {
    *x = 5;
}
```

And this one:

```noir
struct Foo {
    value: Field,
}

impl Foo {
    fn mutate(&mut self) {
        self.value = 2;
    }
}

fn main() {
    let mut foo = Foo { value: 1 };

    let f = || foo.mutate();

    f();

    println(foo);
}
```

(no need to fix those in this PR, though!)

compiler/noirc_frontend/src/elaborator/statements.rs
```diff
@@ -151,8 +151,24 @@ impl Elaborator<'_> {
...
+            let capture_ids =
+                lambda_context.captures.iter().map(|var| var.ident.id).collect::<Vec<_>>();
+            let (id, name, location) = self.get_lvalue_error_info(&lvalue);
+            let typ = self.interner.definition_type(id);
+            for capture_id in capture_ids {
+                if capture_id == id && !typ.is_mutable_ref() {
+                    self.push_err(TypeCheckError::MutableCaptureWithoutRef {
+                        name: name.clone(),
+                        location,
+                    });
```
jfecher, line 169: I was thinking instead of checking on assignment or mutable-ref creation or method calls we could go to the source of where a lambda capture is introduced into a lambda's scope and change that variable to be immutable. Is that possible?

jfecher: Ok, I looked over the lambda capture code some more and we don't create any new definition ids.

So instead of changing the definition (which we can't since it is not a copy and would change the original to be immutable), we should have a single function to verify whether something is mutable and have the check only in that function. I looked across the codebase and found that we actually have already accidentally duplicated this code. `&mut x` calls `check_can_mutate` while `obj.mutating_method()` calls `verify_mutable_reference`. Assignment uses elaborate_lvalue` which returns a mutable boolean, although that method operates on an LValue rather than an ExprId so it is a bit different.

We should unify these methods into one (probably keeping `elaborate_lvalue`) and add the mutable capture check on the Ident case for this shared method. `elaborate_lvalue` should also call a new helper method to check the capture mutability on its Ident case as well.

vezenovm: Ok I will update to do this. 

> obj.mutating_method() calls verify_mutable_reference

I'm not sure where this is? I searched for `mutating_method()` but could not find it. Do you mean `try_add_mutable_reference_to_object`? This is the only place I see `verify_mutable_reference` being called.

jfecher: I'm saying when we use the syntax `obj.mutating_method()` in code we veryify `obj` is mutable within the compiler by calling `verify_mutable_reference`.

Likewise when a user uses `&mut x` in code we use `check_can_mutate` within the compiler to verify it is mutable. So there is code duplication going on. There's `elaborate_lvalue` as well but at least that works on different types.

vezenovm: I have moved the mutable capture check inside of `check_can_mutate` and `verify_mutable_reference `. I also unified some of the errors, as we had repeated errors specified as resolver errors and type check errors. I moved to these all being type check errors.

vezenovm:
> (no need to fix those in this PR, though!)

With PR updates I just pushed, these are also banned now. I have added tests for these cases as well as nested lambdas.

jfecher (approved):
LGTM. Thank you!

Issue #7431, opened 2025-02-18
`path_resolution_to_constructor` panics on struct constructors and error cases


jfecher:
### Aim

We shouldn't panic for incorrect code

### Expected Behavior

We should issue an error for cases we cannot match on or implement the missing features for what we can (matching on structs)

### Bug

We do panic, due to lelftover `todo!()`s from https://github.com/noir-lang/noir/pull/7312

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #7489, opened 2025-02-21
feat(experimental): Support struct constructors in match patterns

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/7431

## Summary\*

Implements structs in match patterns

## Additional Context

There's some mostly unavoidable duplication involving checking for missing/duplicate fields when converting constructor expressions to a pattern. I think abstracting this to work with the existing already abstracted code to do so converting constructors to HirExpressions would result in less readable code.

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
Looks good! It would be nice to have tests for each of the errors that happen here (duplicate field, no such field, etc.)

jfecher:
Added tests along with a separate quick fix that I found while writing the tests.

We were erroring with "variable already defined in pattern" when a pattern used `_` multiple times, e.g. `Foo { x: _, y: _ }`. I added a quick check to allow this and it is tested for by the new tests.

PR #7500, opened 2025-02-24
feat: use resolved type instead of needing Constructor.struct_type

asterite:
# Description

## Problem

Something I noticed while reviewing https://github.com/noir-lang/noir/pull/7489

## Summary

`ConstructorExpression` has a field, `struct_type`, that stores a resolved type from macro expansion. Instead of it we can use `Type::Resolved` for the actual type inside an unresolved type. This works the same way but there's one less field needed.

Additionally, this actually allows more valid programs to compile. For example this one:

```noir
use std::meta::unquote;

struct Foo<T> {}

fn main() {
    let x = comptime {
        let x = quote { Foo::<i32> };
        let q = quote { $x {} };
        unquote!(q)
    };
}
```

Previously when unquoting `Foo::<i32>` as `$x` the generics were lost, so the above program errored with "Type annotation needed". Now the above program compiles fine.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/ast/expression.rs
+++ b/compiler/noirc_frontend/src/ast/expression.rs
@@ -7,9 +7,7 @@ use crate::ast::{
     Ident, ItemVisibility, Path, Pattern, Statement, StatementKind, UnresolvedTraitConstraint,
     UnresolvedType, UnresolvedTypeData, Visibility,
 };
-use crate::node_interner::{
-    ExprId, InternedExpressionKind, InternedStatementKind, QuotedTypeId, TypeId,
-};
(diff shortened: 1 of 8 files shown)


jfecher (approved):
Good catch, this annoyed me on my last PR you mentioned.
For context, we only needed `struct_type: Option<TypeId>` previously because we used to have a `Path` for the constructor type before your PR to change it to an UnresolvedType.

asterite:
> For context, we only needed struct_type: Option<TypeId> previously because we used to have a Path for the constructor type before your PR to change it to an UnresolvedType.

Oooh... I forgot about that. That explains the comment about "path" that I didn't understand.

PR #7509, opened 2025-02-24
fix: shift right overflow in ACIR with unknown var now returns zero

### LINKED ISSUE #7412: result of shr differs in acir and brillig
### Aim

I'm trying to find differences between Brillig and ACIR execution results using fuzzing.

### Expected Behavior

### Bug

Noir program
```
fn main(x: u64, y: u8) -> pub u64 {
  (x >> y)
}
```

with `Prover.toml`: 
```
"x" = "16746359216597577687"
"y" = "65"
```
`nargo execute` ->  `Failed to solve program: 'Cannot satisfy constraint'`

 `nargo debug`:
```
[noir] Circuit witness successfully solved
[noir] Circuit output: Field(0)
```

### To Reproduce

1. noir_version = `1.0.0-beta.1+bd33bebb6ed6c0192cd2a44e6a25e7b52ead04e1`
2. nargo project with provided code and prover
3. in acir `nargo execute`
4. in brillig `nargo debug` and press `n` until `Execution finished`

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

1.0.0-beta.1+bd33bebb6ed6c0192cd2a44e6a25e7b52ead04e1

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #7412

## Summary

The explanation of the bug is here: https://github.com/noir-lang/noir/issues/7412#issuecomment-2679487304

This is one way to solve this. I wonder if there's a better one (less ops).

## Documentation

Check one:

# PR Checklist


jfecher:
Looks like Rust panics when this happens. Are we sure we want to return 0 in acir then instead of changing brillig to also halt?

asterite:
In a chat with @TomAFrench on Slack he said that ideally it's zero. But it's a good opportunity to revisit this as I also tend to prefer matching Rust's behavior.

jfecher:
@TomAFrench can you elaborate here (or on the linked issue) why zero is preferable to erroring?

TomAFrench:
I suggested this as currently the program 

```nr
fn main(x: u8) -> pub u8 {
    x >> 9
}
```

returns 0 when given an input of 1 so we have precedent. (we should create an error for this case if we're wanting to reject larger shifts)

I was also thinking of [these lines in the sha256 impl](https://github.com/noir-lang/sha256/blob/8f2256780e1946430818ac9d2fc0669d9bc0036a/src/sha256.nr#L390-L394) where when we're bitshifting by a value which is only known at runtime and need to add a guard to avoid this error case.

jfecher (approved):
The regressions seem fairly small

PR #7512, opened 2025-02-25
fix: don't crash on broken impl syntax

asterite:
# Description

## Problem

I noticed that LSP would crash whenever starting to write an impl's generics. It was because the compiler actually crashed.

## Summary

This PR prevents this crash.

## Additional Context

I'll try to fix LSP crashes as soon as I find them because I heard LSP crashes often.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/test_programs/compile_failure/broken_impl/src/main.nr
+++ b/test_programs/compile_failure/broken_impl/src/main.nr
@@ -1,0 +1,2 @@
+// This used to crash the compiler
+impl< Foo for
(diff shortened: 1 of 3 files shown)


test_programs/compile_failure/broken_impl/src/main.nr
TomAFrench, line 1: This could be a frontend test right?

asterite: I tried it but because it doesn't parse well I think it's not elaborated (in these tests)

Issue #7511, opened 2025-02-25
Backwards compatible `Program` serialisation


aakoshh:
Currently the `ProgramArtifact` and `ContractFunctionArtifact` contain [base64 encoded bytecode](https://github.com/noir-lang/noir/blob/3bc52e40c86251b3a07cc7e17b5b96bea967d5a5/tooling/noirc_artifacts/src/program.rs#L24-L28) of a `Program`, where the [serialisation](https://github.com/noir-lang/noir/blob/3bc52e40c86251b3a07cc7e17b5b96bea967d5a5/acvm-repo/acir/src/circuit/mod.rs#L241-L247) is done by `bincode` encoding the `Program` and `gzip` compressing the bytes. 

The Noir repo [generates](https://github.com/noir-lang/noir/blob/c254c3ca07c65c7845d5affb17cb37114c5a3051/acvm-repo/acir/src/lib.rs#L52-L137) C++ code for the program and the witness stack using `serde` machinery, which is [used](https://github.com/AztecProtocol/aztec-packages/blob/master/barretenberg/cpp/src/barretenberg/dsl/acir_format/serde/acir.hpp) in Berratenberg to [deserialize](https://github.com/AztecProtocol/aztec-packages/blob/8835b31d76b2f7c45416eaf67a748d8df9dbc753/barretenberg/cpp/src/barretenberg/dsl/acir_format/acir_to_constraint_buf.cpp#L850) the artifact. 

The `Program` has [two fields](https://github.com/noir-lang/noir/blob/3bc52e40c86251b3a07cc7e17b5b96bea967d5a5/acvm-repo/acir/src/circuit/mod.rs#L41-L42): a list of ACIR circuits, and a list of Brillig bytecodes. Out of these two Berratenberg (`bb`) only needs the ACIR circuits, it never touches the unconstrained Brillig part. However, it still deserialises it, and therefore any change in the Brillig opcodes is a breaking change, as `bincode` provides no backwards compatibility. For example a new opcode won't be recognised, resulting in an error, and a deleted opcode (unless it's the last one) ends up shifting the index of all subsequent enum members, which older readers do not expect.

Ideally we want to be able to:
* Add/remove Brillig opcodes without breaking `bb`
* Ignore the Brillig part in `bb` completely
* Remove ACIR opcodes without breaking `bb` (obviously we cannot add new opcodes it can't handle, it should error in that case)

We can consider the formats mentioned in https://github.com/noir-lang/noir/issues/1125 

[...]

PR #7513, opened 2025-02-25
feat: Compare bincode to CBOR, FlexBuffers and Protobuf - implement best

aakoshh:
# Description

## Problem\*

Related to https://github.com/noir-lang/noir/issues/7511

## Summary\*

The comments under the PR contain charts showing the impact the impact on bytecode size of switching to a different serialisation format. Protobuf turned out to be better than `bincode`, so this PR implements the boilerplate to serialise `Program` and `WitnessStack` using protobuf. We can't switch to using it yet, because the Barretenberg wouldn't be able to handle it, and the integration tests use the latest `bb` to parse bytecode and witness files as well. 

Once `bb` can handle both formats (try one, fall back to the other, or take a format parameter), then we can update the integration tests to use the new format and sunset the other. We can potentially also just keep both available, since we will most likely keep using `serde-reflect` as a convenient way to generate the domain objects for C++, into which we can deserialise from protobuf.

## Additional Context

CBOR was relatively good for larger contracts, not much worse than bincode. But on very small contracts it has substantial overhead, so hopefully the extra maintenance of protobuf will be worth it. 

The PR contains scripts to produce the comparisons in the comments (see below).

I added the `GITHUB_TOKEN` to the workflow jobs that build nargo so that don't get rate limited on CI when downloading pre-built `protoc` binaries, but there are other ways we could procure it: I'm sure there is a GitHub action already, but also there is a crate which would build it from source. I wanted it to be low maintenance but quick, which was the appeal of the crate I used that downloads it on the fly.

### Compiling contracts

Run these commands to compile noir protocol circuits and contracts in aztec-packages after rebuilding `nargo`:

```shell
cargo build -p nargo_cli --release
./target/release/nargo --program-dir ../aztec-packages/noir-projects/noir-protocol-circuits compile --force --silence-warnings --skip-underconstrained-check
./target/release/nargo --program-dir ../aztec-packages/noir-projects/noir-contracts compile --force --silence-warnings --skip-underconstrained-check
```

### Baseline 

Recorded the bytecode size with `bincode` before switching to other implementations:
```shell
./scripts/bytecode-sizes/print-bytecode-size.sh ../aztec-packages > ./scripts/bytecode-sizes/bincode.jsonl
```

### Comparing bytecode

Run these commands to record a new measurement (e.g. `flexbuffers` here) and compare against `bincode`.

```shell
./scripts/bytecode-sizes/print-bytecode-size.sh ../aztec-packages > ./scripts/bytecode-sizes/flexbuffers.jsonl
./scripts/bytecode-sizes/compare-bytecode-size.sh ./scripts/bytecode-sizes/bincode.jsonl ./scripts/bytecode-sizes/flexbuffers.jsonl > ./scripts/bytecode-sizes/bincode-vs-flexbuffers.jsonl
./scripts/bytecode-sizes/plot-bytecode-size.sh ./scripts/bytecode-sizes/bincode-vs-flexbuffers.jsonl
```

## Documentation\*

Check one:

# PR Checklist\*


aakoshh:
CBOR shows a stable 5-10% overhead for circuits over 10KB in size; below that the overhead goes up to 170% percent.
bincode-vs-cbor

aakoshh:
FlexBuffers has a higher base overhead and doesn't show any trend of converging to a lower value for larger circuits, going from +200% to +600%:
bincode-vs-flexbuffers

acvm-repo/acir/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![forbid(unsafe_code)]
+#![cfg_attr(not(test), forbid(unsafe_code))]
```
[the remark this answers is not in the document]
aakoshh: Okay, added a comment. I guess it would also be easy for someone to remove it and try to compile, to see why it's there, if that was preferred over the comment.

asterite (approved):
Looks great!

acvm-repo/acir/Cargo.toml
```diff
@@ -18,26 +18,42 @@ workspace = true
...
 [dev-dependencies]
 serde_json = "1.0"
 serde-reflection = "0.3.6"
 serde-generate = "0.25.1"
 fxhash.workspace = true
 criterion.workspace = true
 pprof.workspace = true
+num-bigint.workspace = true
+
+acir = { path = ".", features = ["arb"] } # Self to turn on `arb`.
```
TomAFrench, line 50: Why are we having this crate depend on itself to turn on a feature flag? If we need to have this feature on all the time like this then we should just not have a feature flag.

acvm-repo/acir/src/circuit/opcodes.rs
```diff
@@ -30,7 +31,8 @@ impl BlockType {
 
 #[allow(clippy::large_enum_variant)]
 #[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
-pub enum Opcode<F> {
+#[cfg_attr(feature = "arb", derive(proptest_derive::Arbitrary))]
+pub enum Opcode<F: AcirField> {
```
TomAFrench, line 35: :grimacing: Is this trait bound really necessary?

acvm-repo/acir/Cargo.toml
TomAFrench: Ah, it's just for testing.

acvm-repo/acir/src/circuit/opcodes.rs
TomAFrench: :face_exhaling: seems like a really leaky solution to just needing a `Debug` implementation for `Arbitrary`

acvm-repo/acir/Cargo.toml
aakoshh: It is for testing, yes, but I somehow again encountered the problem that if I just enabled this feature then `noir-wasm` didn't compile. I think it happened when it was like `acir = { workspace = true, features = ["arb"] }` instead of using `path`. 

Originally I tried to make it work like this:
```
cfg_attr(any(test, feature = "arb"), derive(proptest_derive::Arbitrary))]
```
The problem is that half of the stuff that needs `Arbitrary` live in the `brillig` crate, so for that to kick in I either need to add a `dev-dependency` on that crate with `arb`, which can trigger the `wasm` problem, or I can rely on the `arb` feature of `acir` turning on `brillig/arb`.

acvm-repo/acir/src/circuit/opcodes.rs
aakoshh: Yes, it's annoying. I thought I'm not adding any bound that won't be ultimately satisfied, so it's just making what would inevitably be true official, but it's certainly viral.

Issue #7506, opened 2025-02-24
INT_MIN for signed integer types gives out of range error in comptime code


jfecher:
### Aim

```noir
global G: i8 = -128;

fn main() {
    let _x = G;
}
```

### Expected Behavior

No error.

### Bug

```
error: 128 is outside the range of the i8 type
```
This error occurs since we try to evaluate all globals as comptime values.

~~This is another case for #7324 since we forget to check the negative flag in comptime code resulting in this error.~~

Edit: This is related but is not a matter of a missed check. Just one that doesn't account for `abs(INT_MIN) > INT_MAX` for each signed type.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #7519, opened 2025-02-25
fix: No longer error on INT_MIN globals

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/7506
Resolves https://github.com/noir-lang/noir/issues/7324

This PR also possibly improves some error messages which previously displayed negative integer values as a negated field value (a large number) - these will now show an actual negative sign and integer value. I don't expect this to be a noticeable change however since there's a check on `FieldElement::display` to just display the value as negative if that value is smaller (in terms of string characters).

## Summary\*

Allow INT_MIN globals by properly handling negatives by refactoring the frontend to use `SignedField` instead of `FieldElement, bool` pairs.

## Additional Context

One advantage I noticed while using SignedField was that I was able to refactor several separate conversion functions we had (mostly in the interpreter) from SignedField to various signed and unsigned integer types into one place. Before we had written these conversions several times and some handled INT_MIN properly while others did not. Now there are conversion methods on `SignedField` itself.

Some future improvements we could do that I did not tackle in this PR:
- `Type::Constant` still holds a `FIeldElement` rather than a `SignedField`. I think refactoring it to handle negatives was beyond the scope of this PR, although theoretically it should be easier in the future once we have the following point.
- We have several repeated methods for checking if a field value fits in a desired type. We could refactor these to just one method. (There is at least one in SSA and one for `Type::Constant` although the later is only on Fields). This feels like something that should be on SignedField itself (sans error reporting).

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/signed_field.rs
+++ b/compiler/noirc_frontend/src/signed_field.rs
@@ -18,6 +18,7 @@ impl SignedField {
     /// Convert a signed integer to a SignedField, carefully handling
     /// INT_MIN in the process. Note that to convert an unsigned integer
     /// you can call `SignedField::positive`.
+    #[inline]
     pub fn from_signed<T>(value: T) -> Self
     where
         T: num_traits::Signed + AbsU128,
@@ -29,6 +30,7 @@ impl SignedField {
 
     /// Convert a SignedField into an unsigned integer type (up to u128),
     /// returning None if the value does not fit (e.g. if it is negative).
+    #[inline]
     pub fn try_to_unsigned<T: TryFrom<u128>>(self) -> Option<T> {
         if self.is_negative {
             return None;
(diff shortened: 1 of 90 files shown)


jfecher:
Looks like this is still hitting the 2s CI timeout

compiler/noirc_frontend/src/signed_field.rs
```diff
@@ -0,0 +1,160 @@
...
+impl Ord for SignedField {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        if self.is_negative != other.is_negative {
+            if self.is_negative {
+                std::cmp::Ordering::Less
+            } else {
+                std::cmp::Ordering::Greater
+            }
+        } else {
+            self.field.cmp(&other.field)
```
asterite, line 93: If both are negative values should the comparison be the other way around? Otherwise I think `-1` will be less than `-2`.

asterite: (if this was the case, it might be good to add a small test for this)

asterite:
Sorry, scrolling through the page or something I clicked "Update branch" though it wasn't needed...

compiler/noirc_frontend/src/signed_field.rs
jfecher: Great catch! Fixed

PR #7520, opened 2025-02-25
feat!: automatically validate inputs

### LINKED ISSUE #4218: Document how structs returned from unconstrained functions can be invalid and need to be checked
See note on https://github.com/noir-lang/noir/pull/4217

> Note that we cannot enforce safety for all types which have validity conditions outside of the type system. e.g. If an unconstrained function returns a U128 then this will not be constrained as its limbs are made up of Fields so no constraints will be applied allowing a potentially invalid value to be returned.

We should ensure that the documentation explicitly calls out the need for users to add these constraints themselves.

asterite:
# Description

## Problem

Resolves #4218

## Summary

This PR transforms a program like this:

```noir
fn main(x: u8, y: u8) {
    assert_eq(x, y);
}
```

into this one:

```noir
fn main(x: u8, y: u8) {
    if !std::runtime::is_unconstrained() {
        std::validation::AssertsIsValidInput::assert_is_valid_input(x);
        std::validation::AssertsIsValidInput::assert_is_valid_input(y);
    }
    assert_eq(x, y);
}
```

The idea is that all inputs now must implement the `AssertsIsValidInput` trait.

The `if !std::runtime::is_unconstrained()` check is so that these checks aren't done for brillig (I thought `--force-brillig` would be passed to the Elaborator to avoid having this check but this wasn't the case).

Pending:
- [ ] Validate return values from unconstrained functions called from constrained functions
- [ ] Make sure the name of the trait and its method are okay
- [ ] Docs

## Documentation

Check one:

# PR Checklist


aakoshh:
@asterite do you remember what happened to this PR?

asterite:
@aakoshh I think it was just an idea Tom had that we wanted to try out, but then... maybe because of the regressions or because it wasn't clear if this was the direction we wanted to go, it was kind of abandoned. That said, I always keep remembering this PR, there are some audit issues about having for example `BoundedVec` in the ABI boundary and how that would be validated, etc. I think it's a good idea. We should probably ask Tom.

asterite:
Closing for now: we decided we need to spec this a bit more before jumping into an implementation.

PR #7522, opened 2025-02-25
feat(performance): Avoid extra Brillig array offsetting for constant indices 

### LINKED ISSUE #7403: Avoid code gen of unnecessary Brillig array offsetting
### Problem

There are various places in Brillig gen where we seem to be unnecessarily having to increment the array or vector items pointer.   

When we have an array get we need to [generate the pointer to its items](https://github.com/noir-lang/noir/blob/b7ace682af1ab8a43308457302f08b151af342db/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs#L753). Arrays are laid out as [RC, ...items], so every time we want to access the array items we have to increment by one https://github.com/noir-lang/noir/blob/b7ace682af1ab8a43308457302f08b151af342db/compiler/noirc_evaluator/src/brillig/brillig_ir/codegen_memory.rs#L388
A vector pointer requires incrementing by three. 

So if we have an array get like such: `array_get v1, index u32 1`, we will first always increment the pointer specified with `v1` by one, and only then will we apply the user specified offset. The extra increment is completely unnecessary instructions as we already have a constant index. We should be able to just increment by `array_get v1, index u32 1 (user index) + 1 (RC offset)` -> `array_get v1, index u32 2` to fetch the correct value.

### Happy Case

Avoid any unnecessary increments to get the items pointer to an array. The only places this is done is for array get, array set, and make array. I only really looked at array get, would have to see if it is possible to remove this extra RC offset instruction in all cases.
 
Possible options:
1. Handling this in an SSA pass right before Brillig. If we see any constant array gets we should transform them by incrementing by one right away. 
2. SSA code gen different array index handling for the Brillig runtime. 

There might be some extra challenges elsewhere in the compilation pipeline when changing array indexing. For example, arrays are handled the same for ACIR/Brillig in the instruction insertion simplification, so this change would lead to a divergence. It may be simpler to simply convert array indices right before Brillig gen.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #7403

## Summary\*

See the issue for full context of the motivation behind this PR.

Adds a Brillig specific SSA pass to reduce unnecessary operations for offsetting during array accesses. 

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/brillig_array_gets.rs
```diff
@@ -0,0 +1,140 @@
...
+
+        for (instruction_id, _) in instructions_to_update {
+            let new_instruction = match self.dfg[instruction_id] {
+                Instruction::ArrayGet { array, index } => {
+                    let index_constant =
+                        self.dfg.get_numeric_constant(index).expect("ICE: Expected constant index");
+                    let offset = if matches!(self.dfg.type_of_value(array), Type::Array(..)) {
+                        1u128
+                    } else {
+                        3u128
```
asterite, line 64: I guess this is when the type is a slice? If so, why is it three? Also maybe a test for this might be good to have.

asterite: Ah, yes, I see it now, sorry I missed it.

vezenovm: I added a note in the code itself and pushed a test.

asterite (approved):

Issue #7461, opened 2025-02-20
Attach a human-readable code for each error variant.


TomAFrench:
Rust has a snake case code for each lint e.g. `unreachable_pub` which allows developers to easily refer to classes of lints on the CLI, in code, or in config files. We should add somthing similar for all of our diagnostics.

PR #7526, opened 2025-02-25
feat: test error spans and messages

### LINKED ISSUE #7455: Improve testing error locations
Right now in the compiler's frontend tests we just check that expected errors happen. However, we don't check **where** they happen.

One way to do this would be in a way similar to how it's done in parser tests:

https://github.com/noir-lang/noir/blob/8783e480a39efb47c7783c68b0e09c1c65b1f33a/compiler/noirc_frontend/src/parser/parser/statement.rs#L711-L724

That is, allow squiggles bellow a line, pre-process that input and extract the span where the error would happen.

asterite:
# Description

## Problem

Resolves #7455

## Summary

Introduces a `check_errors` helper so we can to be used in frontend tests. Here's an example:

```rust
let src = "
fn main() -> pub i32 {
                 ^^^ expected i32 because of return type
    true        
    ~~~~ bool returned here
}
";
check_errors(src);
```

The idea is that "^^^" lines mark primary errors and "~~~" mark secondary errors.

Pros:
1. (in my opinion) It becomes easier to write these tests as we no longer need a long sequence of code to test thngs
2. The tests also become easier to understand as they include the error messages and their locations
3. We get to test error locations and make sure that's where we want them to be
4. New tests can be written, and be compiled, before the expected error is implemented
5. Multiple errors can be tested at once
6. The tests become more resilient to code refactors: if we change the structure of errors it would just affect the `check_errors` function
7. Tests are shorter (check the diff: +864, -2010)

Cons:
1. We no longer check the underlying type of the error, which might have been desired (not sure!)
2. We can no longer copy the test source code to a file and compile that without first removing the error squiggles (though this wasn't always possible before this PR, for example if a `trait Default` was introduced in a test it would conflict with the existing `Default` trait, etc.)

## Additional Context

I decided to translate all tests here because after I started working on this I wanted to see it through, but also because I believe it would make things easier from now on, and we can also move faster.

As I translated tests I put some "TODO"s about small things to improve. Once this PR merges I'll create a follow-up issue to tackle those TODOs (here I decided not to change the compiler's code except in a couple of places where testing like this was impossible because of small bugs or unpredictability of the error messages).

## Documentation

Check one:

# PR Checklist


asterite:
(re-requested a review because previously it was a draft, now it's ready)

jfecher:
Seeing this I can't help but wonder if we should have just started with [golden tests](https://github.com/jfecher/golden-tests/) all along

asterite:
Ah, yes, using golden tests would also work. I guess we'd need to put frontend tests in a way similar to the ones in `test_programs`, right? Then, it wouldn't be as easy to write a test before introducing a new error, but it's not a big deal.

Another thing is that in these tests one can easily see the code and the errors at once. With golden tests I guess the code will be in one file and the error output will be in another one... which isn't a big deal either, though.

jfecher (approved):
LGTM. 
This is moving us closer to a golden-tests style testing which is definitely more ergonomic to write tests for.

jfecher:
> Ah, yes, using golden tests would also work. I guess we'd need to put frontend tests in a way similar to the ones in test_programs, right? Then, it wouldn't be as easy to write a test before introducing a new error, but it's not a big deal.

Yeah, they'd probably need to be in a different folder to be more discoverable. You should still be able to write the test before the fix though. The program would just error in a different way (presumably) similar to running the frontend test today before the feature is implemented.

> Another thing is that in these tests one can easily see the code and the errors at once. With golden tests I guess the code will be in one file and the error output will be in another one... which isn't a big deal either, though.

No, the code and output would be in the same file (see the examples of that repository I linked)

TomAFrench:
> We no longer check the underlying type of the error, which might have been desired (not sure!)

We can go some part of the way towards this with #7461 (but granted, we still lose full pattern matching)

TomAFrench (approved):
LGTM

asterite:
> We can go some part of the way towards this with https://github.com/noir-lang/noir/issues/7461 (but granted, we still lose full pattern matching)

I guess once we add error codes we can include them like "^^^ [E1234] some message" and parse that (another big refactor, but a much simpler one)

TomAFrench:
Yep, exactly

PR #7536, opened 2025-02-26
fix(experimental): Replace most remaining match panics with errors

jfecher:
# Description

## Problem\*

Resolves 

## Summary\*

Replaces most remaining panic/println cases in the match compiler with friendly user errors.

The only two remaining errors are what I consider to be the meatier ones: redundant cases & unhandled cases since these require more analysis of the match tree itself. The later also typically requires constructing an example of a value that wouldn't be matched as part of the error message.

## Additional Context

I've moved the frontend tests which deal with enums & matches to a new `tests/enums.rs` file since the frontend tests file is quite large. I'll add a comment noting where the new tests in this file are - the old ones are unchanged.

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
Looks great!

PR #7538, opened 2025-02-26
feat: lock on target directories not `Nargo.toml`

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR moves the locks which are currently applied to `Nargo.toml` files onto the target directory. This allows users to run commands in parallel more easily without any potential for clashing by just specifying a separate build directory.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/nargo_cli/src/cli/mod.rs
+++ b/tooling/nargo_cli/src/cli/mod.rs
@@ -6,7 +6,7 @@ use nargo_toml::{
 };
 use noirc_driver::{CrateName, NOIR_ARTIFACT_VERSION_STRING};
 use std::{
-    fs::File,
+    fs::{File, OpenOptions},
     path::{Path, PathBuf},
 };
 
@@ -205,7 +205,9 @@ where
     // Lock manifests if the command needs it.
     let _locks = match cmd.lock_type() {
         LockType::None => None,
-        typ => Some(lock_workspace(&workspace, typ == LockType::Exclusive)?),
+        typ => {
+            Some(lock_target_dir(&workspace.target_directory_path(), typ == LockType::Exclusive)?)
+        }
     };
     run(cmd, workspace)
 }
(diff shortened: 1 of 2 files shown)


tooling/nargo_cli/src/cli/mod.rs
```diff
@@ -223,31 +222,32 @@ fn lock_workspace(
...
-            fs2::FileExt::lock_exclusive(&file)
-                .unwrap_or_else(|e| panic!("Failed to lock {path_display}: {e}"));
-        } else {
-            if fs2::FileExt::try_lock_shared(&file).is_err() {
-                eprintln!("Waiting for lock on {path_display}...",);
-            }
-            fs2::FileExt::lock_shared(&file)
-                .unwrap_or_else(|e| panic!("Failed to lock {path_display}: {e}"));
-        }
+    let lock_path = target_dir.join(".nargo-lock");
```
aakoshh, line 225: I would suggest locking on a package specific lock file in the output directory, to allow the compilations in aztec-packages to go concurrently without all waiting on this single `.nargo-lock` to be free.

tooling/nargo_cli/src/cli/mod.rs
```diff
@@ -223,31 +222,32 @@ fn lock_workspace(
...
+    } else {
+        if fs2::FileExt::try_lock_shared(&file).is_err() {
+            eprintln!("Waiting for lock on {path_display}...",);
+        }
+        fs2::FileExt::lock_shared(&file)
+            .unwrap_or_else(|e| panic!("Failed to lock {path_display}: {e}"));
     }
-    Ok(locks)
+
+    Ok(LockedFile(file))
```
aakoshh, line 250: I would also suggest removing the lock files to not leave around hundreds of them, but maybe that's not important. Or maybe they can all be in a `target/locks` directory, so nobody is bothered.

Issue #4539, opened 2024-03-12
Support `<Type as Trait>` syntax


jfecher:
### Problem

`<Type as Trait>` can sometimes be necessary for referring to trait items within a type. For example, once Noir has associated types, users should be able to do `<Type as Trait>::MyAssociatedType`.

### Happy Case

`<Type as Trait>` syntax is added and usable in a path expression.

### Project Impact

None

### Workaround

None

### Would you like to submit a PR for this Issue?

None

PR #7551, opened 2025-02-27
feat: Support `<Type as Trait>::method` in expressions

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/4539

## Summary\*

Implements 'as trait' syntax in an expression context. It was previously only allowed for associated types.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/expressions.rs
+++ b/compiler/noirc_frontend/src/elaborator/expressions.rs
@@ -34,7 +34,7 @@ use crate::{
         DefinitionId, DefinitionKind, ExprId, FuncId, InternedStatementKind, StmtId, TraitMethodId,
     },
     token::{FmtStrFragment, Tokens},
-    DataType, Kind, QuotedType, Shared, Type, TypeBindings,
+    DataType, Kind, QuotedType, Shared, Type,
 };
 
 use super::{Elaborator, LambdaContext, UnsafeBlockStatus};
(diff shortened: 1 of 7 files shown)


compiler/noirc_frontend/src/elaborator/patterns.rs
```diff
@@ -791,7 +791,7 @@ impl Elaborator<'_> {
         typ
     }
 
-    fn instantiate(
+    pub(super) fn instantiate(
```
asterite, line 794: Is this change needed?

compiler/noirc_frontend/src/elaborator/expressions.rs
```diff
@@ -94,11 +94,8 @@ impl Elaborator<'_> {
                 self.push_err(ResolverError::UnquoteUsedOutsideQuote { location: expr.location });
                 (HirExpression::Error, Type::Error)
             }
-            ExpressionKind::AsTraitPath(_) => {
-                self.push_err(ResolverError::AsTraitPathNotYetImplemented {
```
asterite, line 98: I think this error can be removed now

asterite (approved):
Looks good! I left a couple of comments.

compiler/noirc_frontend/src/elaborator/patterns.rs
jfecher: Ah, not any longer. An earlier version of this PR shared more code which I later figured out could be replaced with `type_check_variable`.

Issue #7540, opened 2025-02-26
Address TODO's in frontend tests


asterite:
The frontend tests have a bunch of TODOs for improving some error messages and locations. We should address those.

Also, there are no TODOs for this, but some errors seem to have a redundant secondary message. For example for unused items we say "unused struct Foo" then "unused struct" as a secondary message, but that doesn't add more information (and Rust doesn't produce a secondary message either).

PR #7554, opened 2025-02-28
chore: address some frontend tests TODOs

asterite:
# Description

## Problem

Part of #7540

## Summary

Addresses the TODOs that are straight-forward to solve. I left others which are a bit more complex.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/type_check/errors.rs
+++ b/compiler/noirc_frontend/src/hir/type_check/errors.rs
@@ -384,7 +384,13 @@ impl<'a> From<&'a TypeCheckError> for Diagnostic {
                             format!("the constant is not of type `{typ}`"),
                             *expr_location,
                         )
-
+                    }
+                    (Kind::Numeric(expected_type), Kind::Numeric(found_type)) => {
+                        Diagnostic::simple_error(
+                            format!("The constant is not of type `{expected_type}`"),
+                            format!("expected `{expected_type}`, found `{found_type}`"),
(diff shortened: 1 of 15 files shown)


compiler/noirc_frontend/src/hir/type_check/errors.rs
```diff
@@ -369,11 +369,37 @@ impl<'a> From<&'a TypeCheckError> for Diagnostic {
...
-                Diagnostic::simple_error(
-                    format!("Expected kind {expected_kind}, found kind {expr_kind}"),
-                    String::new(),
-                    *expr_location,
-                )
+                // Try to improve the error message for some kind combinations
+                match (expected_kind, expr_kind) {
+                    (Kind::Normal, Kind::Numeric(_)) => {
+                        Diagnostic::simple_error(
+                            "Expected type, found let parameter".into(),
```
jfecher, line 376: I think this may be less confusing if we use the term "numeric generic" since that is what the docs use. I think the cases under this should use that term as well instead of referring to them as constants, which they may not be.

asterite: Thanks! I copied Rust in most cases but I wasn't sure about that name either. I'll change it.

asterite: Done!

jfecher (approved):
:+1:

PR #7556, opened 2025-02-28
feat(experimental): Issue errors for unreachable match branches

jfecher:
# Description

## Problem\*

Resolves 

## Summary\*

Implements "unreachable case" errors for patterns in match expressions that are redundant with other patterns.

## Additional Context

This took more time than expected. While the initial error was easy it was surprisingly difficult to get it to stop erroring if there were also any other errors in constructing the match rows. E.g. a program like

```noir
match Some(3) {
    Some(x) => (),
    3 => (),
    None => (),
    None => (),
}
```
Will give a type error on `3 => (),` but will also give a separate error saying that it is unreachable. This extra error broke existing match tests (which mostly only expect 1 error) and was difficult to remove since it is technically correct. In the end I opted to not compile the decision tree of a match at all if there were any errors in constructing the pattern rows. This is what Ante does and it seems to match Rust as well. Neither of those languages issue errors for obviously redundant cases like the second `None => (),` if there was a type error or a name resolution error in constructing the match tree.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/elaborator/enums.rs
```diff
@@ -20,13 +20,87 @@ use crate::{
...
+
+impl Row {
+    fn remove_column(&mut self, variable: DefinitionId) -> Option<Column> {
+        self.columns
+            .iter()
+            .position(|c| c.variable_to_match == variable)
+            .map(|idx| self.columns.remove(idx))
+    }
+}
+
```
jfecher, line 103: All this was just moved from the bottom of the file to the top.

The only change is the addition of `original_body: RowBody` and `location: Location` in `Row`

compiler/noirc_frontend/src/elaborator/enums.rs
```diff
@@ -722,16 +794,38 @@ impl Elaborator<'_> {
...
+            has_missing_cases: false,
+            unreachable_cases: rows.iter().map(|row| (row.body, row.location)).collect(),
+        };
+
+        let hir_match = compiler.compile_rows(rows).unwrap_or_else(|error| {
+            compiler.elaborator.push_err(error);
+            HirMatch::Failure { missing_case: false }
+        });
+
+        if compiler.has_missing_cases {
```
vezenovm, line 814:
```suggestion
        if let HirMatch::Failure { missing_case: true } = hir_match {
            compiler.issue_missing_cases_error(&hir_match);
        }
```
nit: This is the only place I see `has_missing_cases` is read. We could remove this field entirely by just check the match status.

vezenovm (approved):

compiler/noirc_frontend/src/elaborator/enums.rs
jfecher: It's possible the missing case is nested within `hir_match` though. This is the case when you have a match expression like `match x { Some(None) => (), None => (), }` and we're missing a `Some(Some(_))` case

PR #7559, opened 2025-02-28
feat: add pass which inlines constant arguments into brillig functions

### LINKED ISSUE #6812: Add pass which inlines constant arguments into brillig functions
In the https://github.com/noir-lang/noir-bignum library it's common for programs to pass the same constant `Params` object to all of the unconstrained functions used as this contains config values for the field. We should be able to recognise this and push this `params` constant inside of the brillig function's SSA and remove the argument from the function entirely. This would allow us to perform more constant folding optimizations within the brillig function itself as currently we're unable to use the fact that we know the field in use.

asterite:
# Description

## Problem

Resolves #6812

## Summary

Pending:
- [ ] Check what happens if the brillig call is recursive
- [ ] Determine if the >= 30% heuristic is good
- [ ] Add more tests
- [ ] Code cleanup

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/lib.rs
+++ b/compiler/noirc_evaluator/src/lib.rs
@@ -20,10 +20,20 @@ pub(crate) fn trim_leading_whitespace_from_lines(src: &str) -> String {
     while first_line.is_empty() {
         first_line = lines.next().unwrap();
     }
+    let first_line_original_length = first_line.len();
     let mut result = first_line.trim_start().to_string();
+    let first_line_trimmed_length = result.len();
+
+    // Try to see how many spaces we chopped off the first line
+    let difference = first_line_original_length - first_line_trimmed_length;
     for line in lines {
         result.push('\n');
-        result.push_str(line.trim_start());
+        // Try to remove just `difference` spaces to preserve indents
+        if line.len() - line.trim_start().len() >= difference {
+            result.push_str(&line.chars().skip(difference).collect::<String>())
+        } else {
+            result.push_str(line.trim_start());
+        }
     }
     result
 }
--- a/compiler/noirc_evaluator/src/ssa.rs
+++ b/compiler/noirc_evaluator/src/ssa.rs
@@ -211,6 +211,10 @@ fn optimize_all(builder: SsaBuilder, options: &SsaEvaluatorOptions) -> Result<Ss
         .run_pass(Ssa::remove_if_else, "Remove IfElse")
         .run_pass(Ssa::purity_analysis, "Purity Analysis (2nd)")
         .run_pass(Ssa::fold_constants, "Constant Folding")
+        .run_pass(
+            Ssa::inline_constants_into_brillig_functions,
+            "Inline constants into brillig functions",
+        )
         .run_pass(Ssa::flatten_basic_conditionals, "Simplify conditionals for unconstrained")
         .run_pass(Ssa::remove_enable_side_effects, "EnableSideEffectsIf removal")
         .run_pass(Ssa::fold_constants_using_constraints, "Constraint Folding")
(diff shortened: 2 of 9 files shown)


asterite:
I don't understand why these regressions are happening (and less why the "stack too deep" error happens). I'm looking at the final SSA before and after this PR, and the new SSA is shorter. In the case of program "6" the main function is exactly the same, except that some calls have less arguments because they have been inlined. `nargo info` tells me it's more ACIR opcodes with this PR, but it's less brillig opcodes, but that's somehow not reflected in the reports. With `--profile-execution` the executed brillig opcodes are the same, though.

asterite:
I'm also trying it with this program:

```noir
use bignum::BigNum;
use bignum::BigNumTrait;
use bignum::fields::U256::U256Params;

// Define (compile-time) BigNum type
// number of limbs, number of bits of modulus, parameter set
type U256 = BigNum<3, 257, U256Params>;

fn main(x: [Field; 3], y: [Field; 3]) -> pub [Field; 3] {
    let one = U256::from_slice(x);
    let two = U256::from_slice(y);
    // (one + two).limbs
    (one + two).limbs
}
```

This is the output of `nargo info` in master:

```
+---------+------------------+----------------------+--------------+-----------------+
| Package | Function         | Expression Width     | ACIR Opcodes | Brillig Opcodes |
+---------+------------------+----------------------+--------------+-----------------+
| one     | main             | Bounded { width: 4 } | 17           | 783             |
+---------+------------------+----------------------+--------------+-----------------+
| one     | __add_with_flags | N/A                  | N/A          | 783             |
+---------+------------------+----------------------+--------------+-----------------+

(with --profile-execution)

+---------+----------+----------------------+--------------+-----------------+
| Package | Function | Expression Width     | ACIR Opcodes | Brillig Opcodes |
+---------+----------+----------------------+--------------+-----------------+
| one     | main     | Bounded { width: 4 } | 0            | 1617            |
+---------+----------+----------------------+--------------+-----------------+
| one     | main     | N/A                  | N/A          | 1617            |
+---------+----------+----------------------+--------------+-----------------+
```

And this is the output with this PR:

```
+---------+------------------+----------------------+--------------+-----------------+
| Package | Function         | Expression Width     | ACIR Opcodes | Brillig Opcodes |
+---------+------------------+----------------------+--------------+-----------------+
| one     | main             | Bounded { width: 4 } | 17           | 719             |
+---------+------------------+----------------------+--------------+-----------------+
| one     | __add_with_flags | N/A                  | N/A          | 719             |
+---------+------------------+----------------------+--------------+-----------------+

(with --profile-execution)

+---------+----------+----------------------+--------------+-----------------+
| Package | Function | Expression Width     | ACIR Opcodes | Brillig Opcodes |
+---------+----------+----------------------+--------------+-----------------+
| one     | main     | Bounded { width: 4 } | 0            | 1617            |
+---------+----------+----------------------+--------------+-----------------+
| one     | main     | N/A                  | N/A          | 1617            |
+---------+----------+----------------------+--------------+-----------------+
```

So it seems the number of brillig opcodes decreases, but the same number of opcodes is executed.

Maybe we are missing some optimizations done on the functions where constants have been inlined (for example there's no constant folding happening there, though I tried it and it didn't change anything...)

vezenovm:
> (and less why the "stack too deep" error happens)

The rollup block root programs utilizes massive array constants. I thought we had moved all of them to globals, looks like we still pass these globals: https://github.com/AztecProtocol/aztec-packages/blob/f72ccc28641c40dc30ff63929842cb87c3c2a6d4/noir-projects/noir-protocol-circuits/crates/blob/src/blob.nr#L156

We should be able to move to using the `ROOTS` global being passed as an `unconstrained_roots` param now that globals are stable (noted here https://github.com/AztecProtocol/aztec-packages/pull/10605#issuecomment-2595262863 as well). 

However, we should still account for massive constants that may overflow the stack through normal Brillig compilation. In globals compilation, the memory space available to the stack is adjusted at compile-time to account for user specified constants.

PR #7560, opened 2025-02-28
feat(experimental): Issue missing case error for match expressions

jfecher:
# Description

## Problem\*

Resolves 

## Summary\*

Issues a missing case(s) error for match expressions missing any cases. This includes constructing an example of a value which fails to match.

## Additional Context

~~Limitation: if the match doesn't match on any cases such as `match foo {}` we don't issue a helpful error like `Missing cases Foo, Bar, Baz` and instead issue `Missing cases: _`. This could be fixed but it is the end of the week so I wanted to get this work out.~~ This has been implemented along with ranges for integers (see tests).

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/tests/enums.rs
+++ b/compiler/noirc_frontend/src/tests/enums.rs
@@ -356,3 +356,14 @@ fn missing_integer_cases_with_empty_match() {
     ",
     );
 }
+
+#[test]
+fn match_on_empty_enum() {
+    check_errors(
+        "
+        pub fn foo(v: Void) {
+            match v {}
+        }
+        pub enum Void {}",
+    );
+}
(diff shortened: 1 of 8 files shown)


compiler/noirc_frontend/src/tests/enums.rs
```diff
@@ -250,3 +250,109 @@ fn match_reachability_errors_ignored_when_there_is_a_type_error() {
...
+    ",
+    );
+}
+
+#[test]
+fn missing_cases_with_empty_match() {
+    check_errors(
+        "
+        fn main() {
+            match Abc::A {}
```
vezenovm, line 335: Do we have a test for an emtpy match on an enum with no variants? I remember you saying it compiles, but we may as well add a test.

vezenovm (approved):

PR #7562, opened 2025-03-03
chore: add underscore parameter documentation

### LINKED ISSUE #6917: Document support of `_` as function parameters

crStiv:
# Description

## Problem*
Resolves #6917

## Summary*
Added documentation for the underscore `_` parameter feature in functions. This documents an existing feature where parameters can be marked with an underscore when they are not used in the function body, which is particularly useful for implementing traits or interfaces where not all parameters are needed.

## Additional Context
The documentation was added to the Functions concept guide, specifically in the parameters section. The addition includes:
- Clear explanation of the underscore parameter feature
- Practical example showing its usage
- Follows the existing documentation style and formatting

## Documentation*

Check one:

# PR Checklist*


docs/docs/noir/concepts/functions.md
```diff
@@ -34,6 +34,15 @@ is pre-pended with a colon and the parameter type. Multiple parameters are separ
 fn foo(x : Field, y : Field){}
 ```
 
+You can use an underscore `_` as a parameter name when you don't need to use the parameter in the function body. This is useful when you need to satisfy a function signature but don't need to use all the parameters:
+
+```rust
+fn foo(_ : Field, y : Field) {
+    // Only using y parameter
+    y + 1
+}
```
TomAFrench, line 43: This function isn't valid noir as it's returning a value while the return type of the function is `()`.

docs/docs/noir/concepts/functions.md
```diff
@@ -34,6 +34,15 @@ is pre-pended with a colon and the parameter type. Multiple parameters are separ
 fn foo(x : Field, y : Field){}
 ```
 
+You can use an underscore `_` as a parameter name when you don't need to use the parameter in the function body. This is useful when you need to satisfy a function signature but don't need to use all the parameters:
```
TomAFrench, line 37: Noir actually ignores any parameter with a leading underscore so `_x` would also be accepted. This is preferred in general as it self-documents what the unused parameter is.

TomAFrench (changes requested):
Thanks for this PR. A couple of notes

TomAFrench (approved):

TomAFrench:
Thanks for the updates :)

crStiv:
> Thanks for the updates :)

does it need any improvements or corrections?

PR #7570, opened 2025-03-03
fix(experimental): Fix execution of match expressions with multiple branches

### LINKED ISSUE #7323: Match expressions panic during SSA flattening if they're on a runtime value with > 2 branches
### Aim

```rs
fn main(x: Field) {
    match x {
        1 => println(1),
        2 => println(1),
        3 => println(3),
        x => println(x),
    }
}
```

### Expected Behavior

The value of `x` to be printed

### Bug

```
Message:  internal error: entered unreachable code: return encountered before a join point was found. This can only happen if early-return was added to the language without implementing it by jmping to a join block first
Location: compiler/noirc_evaluator/src/ssa/opt/flatten_cfg/branch_analysis.rs:105
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

jfecher:
# Description

## Problem\*

Resolves #7323

## Summary\*

Our SSA flattening pass still doesn't support > 2 predecessors on a block so lower a match by jumping to a new block each time instead. Each end block is merged LIFO like nested if statements. This is less efficient than jumping to the end directly or merging the blocks tournament-bracket style but flattening supports neither of those options.

## Documentation\*

Check one:

# PR Checklist\*


tooling/debugger/ignored-tests.txt
```diff
@@ -5,5 +5,6 @@ macros
 reference_counts
 references
 regression_4709
+regression_7323
```
vezenovm, line 8: Would be good for us to track better what functionality the debugger cannot handle. Just noting this as I the additional ignored test made me think of it.

vezenovm (approved):

vezenovm:
The circuit size report failure is strange. I could call `bb info` locally with that new test.

vezenovm:
> The circuit size report failure is strange. I could call `bb info` locally with that new test.

The error actually looks to not have to do with `bb`. It was failing on the `rebuild.sh` script which simply executes the circuit. We fail with the following:
Screenshot 2025-03-05 at 5 00 11 PM

I didn't expect this as I thought the nargo_cli tests would fail but I see we are already gating for the enums feature in those tests. 

I simply excluded this test from the rebuild script.

PR #7572, opened 2025-03-03
feat: add eq, gte, and lte functions to BN254 library

partylikeits1983:
# Description

I was writing a circuit in noir the other day, and I really wished the `eq`, `gte`, and `lte` functions existed on the `bn254` field library as this would make my circuit cleaner.

## Problem\*

The `bn254` field library does not have `eq`, `gte`, nor `lte` functions. It is possible to compare equality of fields with `==` however, I believe the following syntax is useful: 
```rust
let a: Field = 1;
let b: Field = 1;
let is_equal = a.eq(b);
```

Also the syntax: 

```rust
let a: Field = 0;
let b: Field = 1;
let is_lte = a.lte(b);
```

```rust
let a: Field = 1;
let b: Field = 0;
let is_gte = a.gte(b);
```
would be super useful to have.

## Summary\*
Implemented these functions and added tests.

## Additional Context
I believe this simple PR will make circuit writing slightly easier. These abstractions are super useful.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/noir_stdlib/src/field/bn254.nr
+++ b/noir_stdlib/src/field/bn254.nr
@@ -122,9 +122,123 @@ pub fn lt(a: Field, b: Field) -> bool {
     gt(b, a)
 }
 
+pub fn eq(a: Field, b: Field) -> bool {
+    a == b
+}
+
+pub fn lte(a: Field, b: Field) -> bool {
+    if is_unconstrained() {
+        unsafe {
+            lte_hint(a, b)
+        }
+    } else {
+        if a == b {
+            true
+        } else {
+            unsafe {
+                if field_less_than(a, b) {
+                    assert_gt(b, a);
+                    true
+                } else {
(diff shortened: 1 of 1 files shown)


TomAFrench (changes requested):
Hey @partylikeits1983, this code within the stdlib is currently tech debt and if we were to open up more comparisons between Field types we'd likely go in a different direction.

I'm going to close this PR now and in the meantime I'd recommend implementing this logic in your own code.

partylikeits1983:
@TomAFrench Could you explain the "noir" way of implementing this? Also what is the technical debt that needs to be addressed before useful abstractions can be added?

partylikeits1983:
@critesjosh Opened an issue to discuss how this basic functionality can be added: https://github.com/noir-lang/noir/issues/7580

PR #7578, opened 2025-03-04
chore(profiler): Add option to only get the total sample count for the `execution-opcodes` command

### LINKED ISSUE #7499: Add `nargo info --profile-execution` functionality to `noir-profiler`
We currently can only track the execution trace length of brilig programs using `nargo`. It would be much more ergonomic to have this as part of the `noir-profiler` tool so that it can act on exported noir build artifacts.

Ideally we can feed anything we need for library benchmarking into `noir-profiler` and get JSON output out to be manipulated in CI.

vezenovm:
# Description

## Problem\*

Resolves #7499 

## Summary\*

1. I added a `--sample-count` command. This command returns a single number to stdout of the total number of opcodes executed. I went with this name as "sample" felt clearer to re-iterate that we are profiling over something like `--opcode-count`. I initially was going to do JSON output, but as the profiler works with a single artifact I felt it was easier for consumers to simply work with a single number.
2. I added a `--verbose` flag to hide logging that is outputted by `execution-opcodes`. This flag could be expanded to all the profiler commands. I only did it here as it is convenient for consumers of the `noir-profiler` to fetch output from stdout.
3. We must have updated versions as I was getting errors from the reporter that did not exist before about missing files. I switched the `CustomDiagnostic` constructor to one that does not use secondary labels. Before this PR we were passing dummy information about the secondary labels. Now error reporting acts as expected.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/profiler/src/cli/execution_flamegraph_cmd.rs
+++ b/tooling/profiler/src/cli/execution_flamegraph_cmd.rs
@@ -30,13 +30,22 @@ pub(crate) struct ExecutionFlamegraphCommand {
 
     /// The output folder for the flamegraph svg files
     #[clap(long, short)]
-    output: PathBuf,
+    output: Option<PathBuf>,
 
     /// Use pedantic ACVM solving, i.e. double-check some black-box function
     /// assumptions when solving.
     /// This is disabled by default.
     #[clap(long, default_value = "false")]
     pedantic_solving: bool,
+
+    /// A single number representing the total opcodes executed.
+    /// Outputs to stdout and skips generating a flamegraph.
+    #[clap(long, default_value = "false")]
+    sample_count: bool,
+
+    /// Enables additional logging
+    #[clap(long, default_value = "false")]
+    verbose: bool,
 }
 
 pub(crate) fn run(args: ExecutionFlamegraphCommand) -> eyre::Result<()> {
@@ -46,6 +55,8 @@ pub(crate) fn run(args: ExecutionFlamegraphCommand) -> eyre::Result<()> {
         &InfernoFlamegraphGenerator { count_name: "samples".to_string() },
         &args.output,
         args.pedantic_solving,
+        args.sample_count,
+        args.verbose,
     )
 }
 
@@ -53,20 +64,29 @@ fn run_with_generator(
     artifact_path: &Path,
     prover_toml_path: &Path,
     flamegraph_generator: &impl FlamegraphGenerator,
-    output_path: &Path,
+    output_path: &Option<PathBuf>,
     pedantic_solving: bool,
+    print_sample_count: bool,
+    verbose: bool,
 ) -> eyre::Result<()> {
     let program =
         read_program_from_file(artifact_path).context("Error reading program from file")?;
 
     ensure_brillig_entry_point(&program)?;
 
+    if !print_sample_count && output_path.is_none() {
+        return report_error("Missing --output <OUTPUT> argument for when building a flamegraph")
+            .map_err(Into::into);
+    }
+
     let (inputs_map, _) =
         read_inputs_from_file(&prover_toml_path.with_extension("toml"), &program.abi)?;
 
     let initial_witness = program.abi.encode(&inputs_map, None)?;
 
-    println!("Executing...");
+    if verbose {
+        println!("Executing...");
+    }
 
     let solved_witness_stack_err = nargo::ops::execute_program_with_profiling(
         &program.bytecode,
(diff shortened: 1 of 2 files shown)


tooling/profiler/src/errors.rs
```diff
@@ -9,8 +9,8 @@ pub(crate) enum CliError {
 }
 
 /// Report an error from the CLI that is not reliant on a stack trace.
-pub(crate) fn report_error(message: String) -> Result<(), CliError> {
-    let error = CustomDiagnostic::simple_error(message.clone(), String::new(), Location::dummy());
+pub(crate) fn report_error(message: &str) -> Result<(), CliError> {
+    let error = CustomDiagnostic::from_message(message, FileId::dummy());
```
vezenovm, line 13: This had to be changed as we were creating dummy secondary labels. This was not failing when initially created (look at the description of https://github.com/noir-lang/noir/pull/7436), but was on the most up-to-date master.
I was getting errors like this:
```
% RUST_BACKTRACE=1 noir-profiler execution-opcodes -a ./target/1_mul.json -p Prover.toml --sample-count               ~/noir/test_programs/execution_success/1_mul master + maxim-box

thread 'main' panicked at /mnt/user-data/maxim/noir/compiler/noirc_errors/src/reporter.rs:232:65:
called `Result::unwrap()` on an `Err` value: FileMissing
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: noirc_errors::reporter::report
   4: noir_profiler::errors::report_error
   5: noir_profiler::cli::start_cli
   6: noir_profiler::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```
I think this is due to a version update as the error comes from `term::emit(..)`.

TomAFrench: Yeah, we bumped a lot of the dependencies when we bumped the MSRV. Can you add a quick regression test for this? We could do this with just a mocked out `ProgramArtifact` with two functions.

vezenovm: done

tooling/profiler/src/cli/execution_flamegraph_cmd.rs
```diff
@@ -174,6 +174,7 @@ fn ensure_brillig_entry_point(artifact: &ProgramArtifact) -> Result<(), CliError
     let err_msg = "Command only supports fully unconstrained Noir programs e.g. `unconstrained fn main() { .. }";
     let program = &artifact.bytecode;
     if program.functions.len() != 1 || program.unconstrained_functions.len() != 1 {
+        dbg!("got here");
```
TomAFrench:
```suggestion
```

vezenovm: I think you reviewed before [6f00ae0](https://github.com/noir-lang/noir/pull/7578/commits/6f00ae0ef16efd4970f7c4e260f8764dd376889d), removed this debug in that commit.

aakoshh:
> I added a --verbose flag to hide logging that is outputted by execution-opcodes. This flag could be expanded to all the profiler commands. I only did it here as it is convenient for consumers of the noir-profiler to fetch output from stdout.

Maybe I'm misunderstanding how it works, but I would suggest logging to `stderr` instead of `stdout`, so `stdout` is always reserved for actual program output. This can be achieved using `.with_writer(io::stderr)` in `main`.

tooling/profiler/src/cli/execution_flamegraph_cmd.rs
```diff
@@ -152,3 +188,82 @@ fn ensure_brillig_entry_point(artifact: &ProgramArtifact) -> Result<(), CliError
...
+            },
+            debug_symbols: ProgramDebugInfo { debug_infos: vec![DebugInfo::default()] },
+            file_map: BTreeMap::default(),
+            names: vec!["main".to_string()],
+            brillig_names: Vec::new(),
+        };
+
+        // Write the artifact to a file
+        let artifact_file = std::fs::File::create(&artifact_path).unwrap();
+        serde_json::to_writer(artifact_file, &artifact).unwrap();
```
aakoshh, line 252: You can use:
```
let artifact_path = noir_artifact::fs::artifact::save_program_to_file(&artifact, CrateName("test".into()), temp_dir.path()).unwrap();
```

tooling/profiler/src/cli/execution_flamegraph_cmd.rs
```diff
@@ -46,27 +55,38 @@ pub(crate) fn run(args: ExecutionFlamegraphCommand) -> eyre::Result<()> {
...
+    }
+
     let (inputs_map, _) =
         read_inputs_from_file(&prover_toml_path.with_extension("toml"), &program.abi)?;
 
     let initial_witness = program.abi.encode(&inputs_map, None)?;
 
-    println!("Executing...");
+    if verbose {
+        println!("Executing...");
```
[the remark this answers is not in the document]
aakoshh: Combined with https://github.com/noir-lang/noir/pull/7585 this would mean you don't need a `--verbose` flag; instead you can use the `NOIR_LOG` env var to set the log level.

TomAFrench (approved):
LGTM, agree on handling the logging change in a followup.

PR #7579, opened 2025-03-04
fix: Error on infinitely recursive types

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/7173

## Summary\*

Adds a check during monomorphization that we never create any recursive types which are likely to be infinitely sized.

## Additional Context

I also added a file for monomorphization tests so we don't have to keep creating a full `test_program` for these monomorphization errors.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/monomorphization/tests.rs
```diff
@@ -0,0 +1,37 @@
...
+fn recursive_type_errors() {
+    let src = "
+        fn main() {
+            let _tree: Tree<Tree<Tree<()>>> = Tree::Branch(
+                Tree::Branch(Tree::Leaf, Tree::Leaf),
+                Tree::Branch(Tree::Leaf, Tree::Leaf),
+            );
+        }
+
+        enum Tree<T> {
```
vezenovm, line 30: Will we get errors if we just have a recursive type defined but not instantiated anywhere? 
e.g. 
```noir
fn main() { ... do other stuff ... }
enum Tree {
   Branch(Tree),
   Leaf
}
```
Based off the PR I don't think we will.

compiler/noirc_frontend/src/monomorphization/tests.rs
```diff
@@ -0,0 +1,37 @@
...
+        .get_main_function(crate_id)
+        .unwrap_or_else(|| panic!("get_monomorphized: test program contains no 'main' function"));
+    monomorphize(main, &mut context.def_interner, false)
+}
+
+#[test]
+fn recursive_type_errors() {
+    let src = "
+        fn main() {
+            let _tree: Tree<Tree<Tree<()>>> = Tree::Branch(
```
vezenovm, line 24: This type may be recursive but it is not infinite. Should we not allow this? This code is legal in Rust.

compiler/noirc_frontend/src/monomorphization/tests.rs
vezenovm: I also expect the error from enums to be the same as the one from structs.

I think we could unify this error with [ResolverError::SelfReferentialType](https://github.com/noir-lang/noir/blob/a5b47df4d7c1da746f5377f9c1d35b5a06ea5648/compiler/noirc_frontend/src/hir/resolution/errors.rs#L122). 
e.g. this struct
```noir
struct Tree {
    branch: Tree,
}
```
should trigger the same error about an infinitely sized type as the enum above

jfecher: > Will we get errors if we just have a recursive type defined but not instantiated anywhere?

We wouldn't. I'd have to add a separate check for it. The reason we couldn't only have it when the enum is defined is that we still need to catch types like `Tree<T>` in the original test you commented on which only become recursive when instantiated with `T = Tree<_>`.

compiler/noirc_frontend/src/monomorphization/tests.rs
jfecher: I wanted to allow it but we panic when trying to monomorphize it. I figured we should fix the panic now then could look into relaxing the error and fixing this case later. I'll add a comment.

compiler/noirc_frontend/src/monomorphization/tests.rs
vezenovm: > We wouldn't. I'd have to add a separate check for it. 

Probably could have that check be a part of the checks that trigger [ResolverError::SelfReferentialType](https://github.com/noir-lang/noir/blob/a5b47df4d7c1da746f5377f9c1d35b5a06ea5648/compiler/noirc_frontend/src/hir/resolution/errors.rs#L122). I have not looked at the code though.

compiler/noirc_frontend/src/monomorphization/tests.rs
vezenovm: Ok fair enough. Let's capture that we want to relax the error in an issue though.

compiler/noirc_frontend/src/monomorphization/tests.rs
```diff
@@ -0,0 +1,37 @@
+#![cfg(test)]
+use crate::tests::get_program;
+
+use super::{ast::Program, errors::MonomorphizationError, monomorphize};
+
+fn get_monomorphized(src: &str) -> Result<Program, MonomorphizationError> {
```
asterite, line 6: In the top-level frontend tests there's `monomorphize_program` and `get_monomorphization_error` which could be reused here.

(we could also have another type of `check_errors` calls `check_errors_with_monomorphization` to use squiggles, but that could be done in a separate PR)

jfecher: I ended up removing the one in the frontend since it didn't assert that there were no errors.

I also thought about moving the arithmetic generics tests that called monomorphization functions to the monomorphization tests but thought it'd be better if we kept features together rather than splitting them. I suppose there's an argument for not having a separate monomorphization tests module at all then and just doing it by feature. I think it's fine as long as it is miscellaneous things / single tests though.

michaeljklein (changes requested):
What about a test for `Foo<Bar<()>>` where both `Foo` and `Bar` are enums? Or if that case works, what about if `Bar` is an alias or struct-wrapper of `Foo`?

Naming nit: are these recursive types or simply nested enums? I'd expect the error to be more similar to the nested slices error.

jfecher:
@michaeljklein Added tests

michaeljklein (approved):

PR #7582, opened 2025-03-04
chore!: remove merkle module from stdlib

### LINKED ISSUE #4858: Deprecate `merkle` module in stdlib and create external library
We don't want the functionality in the `merkle` module within the stdlib to be part of it at 1.0. We should then deprecate this module and create a new library from which users can install it.

TomAFrench:
# Description

## Problem\*

Resolves #4858 

## Summary\*

This PR removes the merkle module as set out in the deprecation notice. I've neglected creating an external library as there are other merkle libraries out in the wild already,

## Documentation\*

Check one:

# PR Checklist\*


aakoshh (approved):

asterite:
I pushed a commit. There's a small code duplication but I think trying to refactor it to pass a lambda or something will make things a bit harder to understand. I also think `find_location_index` isn't the right way to solve this but a more correct way is harder and will take more time, so maybe this is good enough for now.

PR #7583, opened 2025-03-05
feat: type alias for numeric generics

### LINKED ISSUE #7272: Type aliases for numeric types
### Problem

We don't currently allow type aliases for numeric generics in Noir:

```noir
type Double<let N: u32>: u32 = N * 2;
```

(syntax up for debate)

### Happy Case

We should support type aliases for numeric types to simplify larger arithmetic generic expressions.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #7272 

## Summary\*
Allow type alias to reference numeric generic expression.

## Additional Context
Changes are minimal, more use cases need to be tested. In particular there is no documentation. It can be added as a subsequent PR (or even in this one).
The PR is done in order to validate the approach taken.

## Documentation\*

Check one:

# PR Checklist\*


guipublic:
I am not sure why I had to update some unit tests.

compiler/noirc_frontend/src/parser/parser/type_alias.rs
```diff
@@ -16,39 +19,49 @@ impl Parser<'_> {
...
+        let mut num_typ = None;
         let generics = self.parse_generics();
-
-        if !self.eat_assign() {
+        if self.eat_colon() {
+            // To specify a type alias on a numeric generic expression, we need to specify the type of the expression
+            // It must be a numeric type
+            num_typ = Some(self.parse_type_or_error());
+        }
+        let location = self.location_since(start_location);
```
asterite, line 38: This location should be created right before finishing parsing the type alias (I think this is why you get a difference in the tests). It seems previously this was right before the semicolons.

compiler/noirc_frontend/src/ast/type_alias.rs
```diff
@@ -4,18 +4,89 @@ use noirc_errors::Location;
...
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let generics = vecmap(&self.generics, |generic| generic.to_string());
         write!(f, "type {}<{}> = {}", self.name, generics.join(", "), self.typ)
     }
 }
+
+#[derive(Clone, Debug)]
+pub struct NumericTypeAlias {
+    pub type_alias: NormalTypeAlias,
+    pub numeric_type: UnresolvedType,
```
asterite, line 41: If the only difference between the two aliases is this numeric_type, maybe it could be turned into `numeric_type: Option<UnresolvedType>` in `NoirTypeAlias` to avoid an enum. I think parsing will be simplified too, and maybe the elaborator side too. The only downside I can see is that we need to remember to check this field during elaboration, but it's a one-time case and with a test we'll have it covered.

compiler/noirc_frontend/src/ast/type_alias.rs
```diff
@@ -4,18 +4,89 @@ use noirc_errors::Location;
...
+}
+
+impl Display for NumericTypeAlias {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let generics = vecmap(&self.type_alias.generics, |generic| generic.to_string());
+        write!(
+            f,
+            "numeric ({}) type {}<{}> = {}",
```
asterite, line 49: I don't know if we use this strings for compilation, but right now that output doesn't match the syntax that we use for parsing it.

compiler/noirc_frontend/src/tests.rs
```diff
@@ -1572,6 +1593,22 @@ fn numeric_generic_as_struct_field_type_fails() {
...
 
+#[test]
+fn type_alias_to_numeric_generic() {
+    let src = r#"
+    type Double<let N: u32>: u32 = N * 2;
+    fn main() {
+        let b: [u8; 6] = foo();
+        assert(b[0] == 0);
+    }
+    fn foo<let N:u32>() -> [u8;Double::<N>] {
```
jfecher, line 1604: Can we parse `Double::<N>` instead as `Double<N>`? That way it'd fit more in line with generics on other non-numeric type aliases

compiler/noirc_frontend/src/tests.rs
```diff
@@ -203,21 +208,33 @@ fn check_errors_with_options(src: &str, allow_parser_errors: bool, options: Fron
             continue;
         }
 
+        if let Some((span, message)) =
+            get_error_line_span_and_message(line, '-', byte, last_line_length)
+        {
+            custom_spans_with_errors.push((span, message));
+            continue;
+        }
```
jfecher, line 216: Unsure why we'd need to change this either. @asterite any insights?

compiler/noirc_frontend/src/tests.rs
```diff
@@ -203,21 +208,33 @@ fn check_errors_with_options(src: &str, allow_parser_errors: bool, options: Fron
...
+
         code_lines.push(line);
 
         byte += line.len() + 1; // For '\n'
         last_line_length = line.len();
     }
-
+    dbg!(&custom_spans_with_errors);
```
jfecher, line 223: A few leftover `dbg` in this function

compiler/noirc_frontend/src/ast/type_alias.rs
jfecher: I also think the `Option<UnresolvedType>` approach may be simpler

compiler/noirc_frontend/src/tests.rs
jfecher: Hmm, I see the problem though. We're parsing an expression here not a type so we run into something like `D < N > 2` actually being a valid expression...

compiler/noirc_frontend/src/tests.rs
asterite: I think the overall location of the type alias was changed so these tests broke (though I'm not sure why we'd need custom spans here).

compiler/noirc_frontend/src/parser/parser/type_alias.rs
guipublic: I changed the location to be parsed before the semicolons, it fixed on unit test but not the ones in test/visibility.rs

compiler/noirc_frontend/src/tests.rs
guipublic: done

compiler/noirc_frontend/src/ast/type_alias.rs
guipublic: Done

compiler/noirc_frontend/src/ast/type_alias.rs
guipublic: I guess it is now irrelevant after the switch to the option instead of NoirTypeAlias enum

compiler/noirc_frontend/src/tests.rs
```diff
@@ -1523,6 +1541,25 @@ fn numeric_generic_as_struct_field_type_fails() {
...
+    fn foo<let N:u32>() -> [u32;Double::<N>] {
+        let mut a = [0;Double::<N>];
+        for i in 0..Double::<N> {
+            a[i] = i;
+        }
+        a
+    }
+    "#;
+    assert_no_errors(src);
+}
```
TomAFrench, line 1561: Can we test composing these? 

```noir
type Quadruple<let N; u32>: u32 = Double<Double<N>>;
```

compiler/noirc_frontend/src/tests.rs
guipublic: Done, cf: type_alias_to_numeric_as_generic()

compiler/noirc_frontend/src/tests.rs
guipublic: I could not make it work in a loop bound, so I disallow it. It can be worked out in another PR.

compiler/noirc_frontend/src/tests.rs
guipublic: I resolve the conversation because this has been reverted (after location change and update from master)

compiler/noirc_frontend/src/parser/parser/type_alias.rs
guipublic: These unit tests are now working after an update from master.

compiler/noirc_frontend/src/tests/aliases.rs
```diff
@@ -75,3 +75,74 @@ fn double_generic_alias_in_path() {
...
     assert_no_errors(src);
 }
+
+#[test]
+fn identity_numeric_type_alias_works() {
+    let src = r#"
+    pub type Identity<let N: u32>: u32 = N;
+    "#;
+    assert_no_errors(src);
+}
```
TomAFrench, line 85: I've moved the tests here into the `aliases` module to scope them a bit more.

I've also added this test which checks that we can have a type alias which just returns its generic parameter which seems to fail.

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1645,17 +1645,34 @@ impl<'context> Elaborator<'context> {
...
+        let location = alias.type_alias_def.location;
 
         let generics = self.add_generics(&alias.type_alias_def.generics);
         self.current_item = Some(DependencyId::Alias(alias_id));
-        let typ = self.resolve_type(alias.type_alias_def.typ);
+
+        let (typ, num_expr) = if let Some(num_type) = alias.type_alias_def.numeric_type {
+            let num_type = self.resolve_type(num_type);
+            let kind = Kind::numeric(num_type);
+            let num_expr = if let UnresolvedTypeData::Expression(expr) =
```
TomAFrench, line 1656: This looks to be the issue with the identity alias. This is a pretty fragile check it seems as we can avoid it by just ensuring that the last step is an expression and it doesn't look deeper?

`type Quadruple<let N: u32>: u32 = Double<N> + Double<N>;`

compiler/noirc_frontend/src/tests.rs
```diff
@@ -148,6 +148,7 @@ fn assert_no_errors(src: &str) {
 /// where:
 /// - lines with "^^^" are primary errors
 /// - lines with "~~~" are secondary errors
+/// - lines with "---" are custom errors
```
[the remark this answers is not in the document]
guipublic: I added it to workaround some unit test, but I do not need it anymore and I removed it from the PR, but I forgot the comment.
Using '---' was a bad idea anyways as it conflicted with the minus operator.
I will remove this line.

TomAFrench:
I'm a little sus by how loose we're being with the various kinds of type aliases now.

e.g. `Foo` is now considered an arithmetic generic type alias by the compiler.

```noir
type Foo: u32 = u32;
```

This errors with 

```
error: Cannot use a type alias inside a type alias
  ┌─ src/main.nr:6:1
  │
6 │ type Foo: Bar = u32;
  │ -------------------
  │
```
 but really we should be rejecting it as an invalid type alias off the bat.

TomAFrench:
Can you add some unit tests for the cases where composing numeric type aliases results in issues? Ideally we'd have tests that we can use to determine when we can remove the block on nested numeric type aliases

compiler/noirc_frontend/src/ast/mod.rs
```diff
@@ -530,6 +547,38 @@ impl UnresolvedTypeExpression {
...
+                    rhs: Expression { kind: rhs.to_expression_kind(), location: *location },
+                }))
+            }
+            UnresolvedTypeExpression::AsTraitPath(path) => {
+                ExpressionKind::AsTraitPath(*path.clone())
+            }
+        }
+    }
+
+    fn operator_to_binary_op_kind_helper(op: &BinaryTypeOperator) -> BinaryOpKind {
```
jfecher, line 572: This should be a method on BinaryTypeOperator

compiler/noirc_frontend/src/ast/mod.rs
```diff
@@ -553,6 +602,17 @@ impl UnresolvedTypeExpression {
             }
         }
     }
+
+    pub(crate) fn is_valid_expression(&self) -> bool {
+        match self {
+            UnresolvedTypeExpression::Variable(path) => path.is_ident(),
```
jfecher, line 608: The comment in elaborator/mod.rs mentions this function should allow globals & constants. We can have multiple segment paths that still resolve to constants or globals. E.g. `crate::CONSTANT` or `foo::bar::my_global`, but this method disallows these.

jfecher (commented):
A couple more things I found before we merge.

tooling/lsp/src/with_file.rs
```diff
@@ -131,16 +131,23 @@ fn pattern_with_file(pattern: Pattern, file: FileId) -> Pattern {
...
+        location: location_with_file(type_alias.location, file),
+        numeric_type: type_alias
+            .numeric_type
+            .map(|num_type| unresolved_type_with_file(num_type, file)),
     }
 }
 
+fn noir_type_alias_with_file(noir_type_alias: TypeAlias, file: FileId) -> TypeAlias {
+    type_alias_with_file(noir_type_alias, file)
+}
```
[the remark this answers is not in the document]
guipublic: Done

compiler/noirc_frontend/src/tests/aliases.rs
```diff
@@ -75,3 +75,130 @@ fn double_generic_alias_in_path() {
...
+            a[i] = i + n;
+        }
+        a
+    }
+    "#;
+    check_errors(src);
+}
+
+#[test]
+fn disallows_numeric_type_aliases_to_expression_with_alias_1() {
```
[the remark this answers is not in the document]
guipublic: Done

compiler/noirc_frontend/src/hir/resolution/errors.rs
```diff
@@ -288,515 +292,529 @@ impl<'a> From<&'a ResolverError> for Diagnostic {
...
-                    "second definition found here".to_string(),
-                    *second_location,
-                );
-                diag.add_secondary("first definition found here".to_string(), *first_location);
-                diag
-            }
+                                let mut diag = Diagnostic::simple_error(
+                                    format!("duplicate definitions of {name} found"),
+                                    "second definition found here".to_string(),
+                                    *second_location,
```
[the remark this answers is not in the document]
guipublic: Done

compiler/noirc_frontend/src/elaborator/types.rs
```diff
@@ -61,6 +61,15 @@ impl Elaborator<'_> {
...
 
+    pub(crate) fn resolve_type_with_kind(&mut self, typ: UnresolvedType, kind: &Kind) -> Type {
+        let location = typ.location;
+        let resolved_type = self.resolve_type_inner(typ, kind);
+        if resolved_type.is_nested_slice() {
+            self.push_err(ResolverError::NestedSlices { location });
+        }
+        resolved_type
+    }
+
```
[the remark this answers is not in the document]
guipublic: No, this function is still there as far as I can tell. I refactored it a bit to use resolve_type_inner()

compiler/noirc_frontend/src/elaborator/patterns.rs
```diff
@@ -537,11 +537,24 @@ impl Elaborator<'_> {
 
     pub(super) fn elaborate_variable(&mut self, variable: Path) -> (ExprId, Type) {
         let unresolved_turbofish = variable.segments.last().unwrap().generics.clone();
-
         let location = variable.location;
         let (expr, item) = self.resolve_variable(variable);
         let definition_id = expr.id;
 
+        if expr.id == DefinitionId::dummy_id() {
+            if let Some(PathResolutionItem::TypeAlias(alias)) = item {
```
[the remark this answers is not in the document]
guipublic: Done

compiler/noirc_frontend/src/ast/mod.rs
guipublic: Done

compiler/noirc_frontend/src/ast/mod.rs
guipublic: Done, I simply disallow the usage of generics instead.

compiler/noirc_frontend/src/tests/visibility.rs
```diff
@@ -22,7 +22,7 @@ fn errors_if_type_alias_aliases_more_private_type() {
     let src = r#"
     struct Foo {}
     pub type Bar = Foo;
-                   ^^^ Type `Foo` is more private than item `Bar`
+    ^^^^^^^^^^^^^^^^^^ Type `Foo` is more private than item `Bar`
```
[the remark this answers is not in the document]
guipublic: Done

compiler/noirc_frontend/src/tests/aliases.rs
```diff
@@ -75,3 +75,130 @@ fn double_generic_alias_in_path() {
...
+    let src = r#"
+    type Double<let N: u32>: u32 = N * 2;
+    type Quadruple<let N: u32>: u32 = N*(Double::<N>+3);
+    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^  Cannot use a type alias inside a type alias
```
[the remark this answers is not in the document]
guipublic: Done

compiler/noirc_frontend/src/elaborator/scope.rs
```diff
@@ -88,14 +88,31 @@ impl Elaborator<'_> {
...
+                let global = self.interner.get_global(global);
+                Ok((global.definition_id, item))
+            }
+            PathResolutionItem::TypeAlias(type_alias_id) => {
+                let type_alias = self.interner.get_type_alias(type_alias_id);
+
+                if type_alias.borrow().numeric_expr.is_some() {
+                    // Type alias to numeric generics are aliases to some global value
+                    // Therefore we allow this case although we cannot provide the value yet
+                    return Ok((DefinitionId::dummy_id(), item));
```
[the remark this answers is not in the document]
guipublic: I don't think the dummy ids get filled for type aliases. Type aliases are replaced during elaboration by the expression they alias for. This dummy is done in order to avoid a failure here for type aliases.
I tried to create a definition id for them but then it has a lot of implications because it requires a new DefinitionKind for type aliases. It is possible to ignore this new kind, but then it seems to me that creating a DefinitionId and not doing anything with it is worse than using a dummy id (which is more explicit that we are not going to use it).

asterite:
@guipublic CI is now passing but I see a lot of error locations shifted here when I think the previous locations were better, so that might need to be fixed.

guipublic:
> @guipublic CI is now passing but I see a lot of error locations shifted here when I think the previous locations were better, so that might need to be fixed.

Thank you!
I improved the location, let me know if there are some cases that I miss.

jfecher:
@guipublic looks like there are still some location differences, e.g. on test_noirc_frontend_tests_aliases_disallows_numeric_type_aliases_to_expression_with_alias

guipublic:
> @guipublic looks like there are still some location differences, e.g. on test_noirc_frontend_tests_aliases_disallows_numeric_type_aliases_to_expression_with_alias

There was indeed a snapshot issue (now solved).

compiler/noirc_frontend/src/elaborator/patterns.rs
```diff
@@ -568,6 +568,7 @@ impl Elaborator<'_> {
     }
 
     pub(super) fn elaborate_variable(&mut self, variable: Path) -> (ExprId, Type) {
+        // let unresolved_turbofish = variable.segments.last().unwrap().generics.clone();
```
asterite: Extra comment

compiler/noirc_frontend/src/elaborator/patterns.rs
```diff
@@ -581,6 +582,19 @@ impl Elaborator<'_> {
         let (expr, item) = self.resolve_variable(variable);
         let definition_id = expr.id;
 
+        if let Some(PathResolutionItem::TypeAlias(alias)) = item {
+            // A type alias to a numeric generics is considered like a variable
+            // but it is not a real variable so it does not resolve to a valid Identifier
+            // In order to handle this, we retrieve the numeric generics expression that the type aliases to
+            let type_alias = self.interner.get_type_alias(alias);
+            if let Some(expr) = type_alias.borrow().numeric_expr.clone() {
```
asterite:
```suggestion
            if let Some(expr) = type_alias.borrow().numeric_expr {
```

compiler/noirc_frontend/src/elaborator/patterns.rs
```diff
@@ -581,6 +582,19 @@ impl Elaborator<'_> {
...
+                let expr = UnresolvedTypeExpression::to_expression_kind(&expr);
+                let expr = Expression::new(expr, type_alias.borrow().location);
+                return self.elaborate_expression(expr);
+            }
+        }
+
+        //    let type_generics = item.map(|item| self.resolve_item_turbofish(item)).unwrap_or_default();
```
asterite: Extra comment

compiler/noirc_frontend/src/tests/visibility.rs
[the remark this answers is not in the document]
asterite: I think the location is still changed compare to master.

tooling/nargo_fmt/src/formatter/alias.rs
```diff
@@ -1,18 +1,23 @@
...
 impl Formatter<'_> {
-    pub(super) fn format_type_alias(&mut self, type_alias: NoirTypeAlias) {
+    pub(super) fn format_type_alias(&mut self, type_alias: TypeAlias) {
         self.write_indentation();
         self.format_item_visibility(type_alias.visibility);
         self.write_keyword(Keyword::Type);
         self.write_space();
         self.write_identifier(type_alias.name);
         self.format_generics(type_alias.generics);
+        if let Some(num_type) = type_alias.numeric_type {
```
asterite, line 16: Could you add a formatter test for this?

guipublic: Done

compiler/noirc_frontend/src/elaborator/patterns.rs
guipublic: done

compiler/noirc_frontend/src/elaborator/patterns.rs
guipublic: done

compiler/noirc_frontend/src/elaborator/patterns.rs
guipublic: done

jfecher (approved):

PR #7584, opened 2025-03-05
fix: Display causes but not stack trace in CLI error report

aakoshh:
# Description

## Problem\*

Resolves something I noticed while reviewing https://github.com/noir-lang/noir/pull/7578

## Summary\*

Some of the CLI used `eprintln!("{report}")` to display errors, while others use `eprintln!("{report:?}")`. 
Using the former with `eyre` means we only see the outermost `.context(msg)`, which can hide a lot useful details about the original error. See https://docs.rs/eyre/latest/eyre/struct.Report.html#display-representations

## Additional Context

For example if I point the profiler to a non-existing artifact, all we see it this:
```console
$ ./target/debug/noir-profiler opcodes --artifact-path  ./test_programs/execution_success/blah.json --output test.json
Error reading program from file
```

With `eprintln!("{report:?}");` we get the nested contextual messages like a stack, and the stack trace:

```console
$ ...
Error reading program from file

Caused by:
    No such file or directory (os error 2)

Location:
    tooling/profiler/src/cli/opcodes_flamegraph_cmd.rs:46:47
```

With `eprintln!("{report:#}")` we get a more compact output that omits source code locations (which could be considered an implementation detail that should not concern the user, but I'd only recommend it if we used `.context` more vigorously to provide a human readable trace):

```console
$ ...
Error reading program from file: No such file or directory (os error 2)
```

Since `noir-inspector` and `noir-execute` use `:?` I thought we can stick to that for now.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/.github/benchmark_projects.yml
+++ b/.github/benchmark_projects.yml
@@ -1,4 +1,4 @@
-define: &AZ_COMMIT cca90e5e655ed9a2d2bb969f034e42ac15f87439
+define: &AZ_COMMIT 5f141167d16f1450348a4bbe55d94a462b5b62eb
 projects:
   private-kernel-inner:
     repo: AztecProtocol/aztec-packages
@@ -65,7 +65,7 @@ projects:
     cannot_execute: true
     num_runs: 1
     timeout: 60
-    compilation-timeout: 110
+    compilation-timeout: 135
     compilation-memory-limit: 8000
   rollup-block-root:
     repo: AztecProtocol/aztec-packages
@@ -73,7 +73,7 @@ projects:
     path: noir-projects/noir-protocol-circuits/crates/rollup-block-root
     num_runs: 1
     timeout: 60
-    compilation-timeout: 120
+    compilation-timeout: 135
     execution-timeout: 40
     compilation-memory-limit: 8000
     execution-memory-limit: 1500
--- a/.github/workflows/deny.yml
+++ b/.github/workflows/deny.yml
@@ -21,6 +21,6 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
-      - uses: EmbarkStudios/cargo-deny-action@v1
+      - uses: EmbarkStudios/cargo-deny-action@8d73959fce1cdc8989f23fdf03bec6ae6a6576ef
         with:
           command: check all
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2828,12 +2828,12 @@ dependencies = [
 
 [[package]]
 name = "light-poseidon"
-version = "0.2.0"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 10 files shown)


vezenovm (changes requested):
I had removed `eprintln!("{report:?}");` for `eprintln!("{report}");` in the profiler as I didn't want to expose internal details to the user for CLI errors. 

For example now in `nargo` let's say I forget to include a `Prover.toml`:
```
Cannot find input file '/mnt/user-data/maxim/noir/test_programs/execution_success/1_mul/Prover.toml'

Location:
    tooling/nargo_cli/src/cli/mod.rs:137:5
```
I don't think we should be showing the CLI locations to the user. I think it is actually more confusing. 

Perhaps we can have something similar to `RUST_BACKTRACE=1` for debugging mode. However, this is normally used for Rust panics. So we can run `nargo` with this flag already if we have an internal panic, and for Noir execution, we already provide full call stacks.

aakoshh:
> I had removed eprintln!("{report:?}"); for eprintln!("{report}"); in the profiler as I didn't want to expose internal details to the user for CLI errors.

Okay, like I mentioned in the description we can use `eprintln!("{report:#}")` to display as much of the context of what went wrong as possible, including the original error message, and all the contextual messages that got attached to it. I'll change the value accordingly to all CLIs. Like I said it's just less likely to tell us exactly what you want to know, unless we consistently provide good error messages and contexts. For example some `std::io::Error` will say that "file not found" but won't say which one, unless the thing that called it used `.context` or `.wrap_err` to give hints, or maybe `.with_context` to even attach the path, or if we used a custom error message which captured the arguments we had problem with. Even then, if somebody used `.context` on top of our custom error, the message of the `.context` would be the only one shown with `{report}` which is usually very frustrating.

aakoshh:
I suppose we could have an env var to influence the report format, like if `RUST_BACKTRACE` itself is set we can use debug format to help in cases where the error is nondescript.

vezenovm:
> eprintln!("{report:#}")

Yeah I think this is better than `{report:?}`. 

> Like I said it's just less likely to tell us exactly what you want to know, unless we consistently provide good error messages and contexts.

I understand your desire for more information in the error reports. I do think for public facing binaries we should consistently provide good error messages and contexts through our `noirc_errors` reporter. 

This [method](https://github.com/noir-lang/noir/blob/33284ac0641343981ac2a209d9d46fdb8ec0ed84/compiler/noirc_errors/src/reporter.rs#L220) is used in `nargo_cli` for reporting execution failures and I recently started using it in the profiler as well when I switched `{report:?}` to `{report}`.

vezenovm (approved):

vezenovm:
> Yes, for the errors generated about Noir that's fantastic 👍 This is more for just the bits and bobs that can go wrong in the CLI, opening and decoding files, etc.

Yeah I understand, I just didn't want to muddle user errors. If `{report:#}` I think we should look at an env var to influence extra debug info.

PR #7585, opened 2025-03-05
fix: Log to `stderr`

aakoshh:
# Description

## Problem\*

Resolves something I noticed while reviewing https://github.com/noir-lang/noir/pull/7578

## Summary\*

The `profiler` to `stdout`, while also aims to display information that the users should be able to pick up in scripting, which is made difficult if log messages and program output are mixed. The solution is to send all logs to `stderr`, in which case piping the output to another program ignores it unless it's redirected with `2>&1`. 

## Additional Context

I tested that the default behaviour of `tracing_subscriber` was indeed to log to `stdout` by adding a `tracing::info!(<msg>)` call and looking for the message with `| grep <msg>`, which was highlighted with red. If we apply the changes in this PR, it is no longer highlighted, as it appears on `stderr`.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/acvm_cli/src/main.rs
+++ b/tooling/acvm_cli/src/main.rs
@@ -25,6 +25,7 @@ fn main() {
     } else {
         tracing_subscriber::fmt()
             .with_span_events(FmtSpan::ACTIVE)
+            .with_writer(std::io::stderr)
             .with_ansi(true)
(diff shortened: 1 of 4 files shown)


TomAFrench (approved):
LGTM

PR #7587, opened 2025-03-05
fix: add u128 overflow checks after DIE

### LINKED ISSUE #7555: optimization of acir and brillig differs
### Aim

I'm trying to find differences between Brillig and ACIR execution results using fuzzing

### Expected Behavior

### Bug

Prover.toml
```
"a" = "3170535237180456974"
"b" = "3170535237180456974"
"c" = "3170535237180456974"
"d" = "3170535237180456974"
```

main.nr
```
pub fn main(a: u128, b: u128, c: u128, d: u128) -> pub u128 {
  let k = a * b;
  let t = k * c;
  let p = t * d;
  a + a
}
```

nargo execute --show-ssa
```
.
.
.
  b0(v0: u128, v1: u128, v2: u128, v3: u128):
    v4 = mul v0, v1
    v6 = div v0, u128 18446744073709551616
    v7 = div v1, u128 18446744073709551616
    v8 = unchecked_mul v6, v7
    constrain v8 == u128 0, "attempt to multiply with overflow"
    v10 = mul v4, v2
    v11 = div v4, u128 18446744073709551616
    v12 = div v2, u128 18446744073709551616
    v13 = unchecked_mul v11, v12
    constrain v13 == u128 0, "attempt to multiply with overflow"
    v14 = div v10, u128 18446744073709551616
    v15 = div v3, u128 18446744073709551616
    v16 = unchecked_mul v14, v15
    constrain v16 == u128 0, "attempt to multiply with overflow"
    v17 = add v0, v0
    return v17
}

public_parameter_witnesses: {}
private_parameters: {Witness(0), Witness(1), Witness(2), Witness(3)}
warning: unused variable p
  ┌─ src/main.nr:4:7
  │
4 │   let p = t * d;
  │       - unused variable
  │

  ┌─ /home/defkit/heck/src/main.nr:3:11
  │
3 │   let t = k * c;
  │           -----
  │
  = Call stack:
    1. /home/defkit/heck/src/main.nr:3:11

Failed assertion
```

nargo execute --force-brillig --show-ssa
```
.
.
.
  b0(v0: u128, v1: u128, v2: u128, v3: u128):
    v4 = add v0, v0
    return v4
}

public_parameter_witnesses: {}
private_parameters: {Witness(0), Witness(1), Witness(2), Witness(3)}
warning: unused variable p
  ┌─ src/main.nr:4:7
  │
4 │   let p = t * d;
  │       - unused variable
  │

[heck] Circuit witness successfully solved
[heck] Circuit output: Field(6341070474360913948)
[heck] Witness saved to /home/defkit/heck/target/heck.gz
```

### To Reproduce

1.   With files above
2.   `nargo execute --force-brillig --show-ssa`
3.   `nargo execute --show-ssa`
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

noirc version = 1.0.0-beta.3+50193ed9f85f10f7b815b3288eb20de9f9a3613d

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #7555

## Summary

The compiler generally doesn't issue overflow errors for expressions that end up being unused. The overflow checks for u128 were added before the first DIE pass and contain some `div` and `constrain` instructions that DIE won't remove. Adding these checks after the first DIE fixes the issue as the multiplications are removed by DIE so no checks are added.

## Additional Context

It's not clear if overflows for unused expressions should be removed or not, but this is what the compiler does in other cases so at least this makes the compiler work the same way in all cases.

## Documentation

Check one:

# PR Checklist


TomAFrench:
Seems like we could have this easily reoccur in some cases if we were to add more passes in future which would result in more u128 ops being revealed as unused in later passes. :thinking:

asterite:
I also checked what happened with signed integers: it always fails with overflow regardless of ACIR and brillig, because the checks are inserted during SSA.

One way to solve this particular case is to always check for u128 overflow in SSA, even for brillig. However, for other unsigned types it will still be different (no overflow on unused expressions).

So maybe we need to answer this question first: do we want overflows for unused expressions or not?

TomAFrench:
Hmm, I'd be open to always performing overflow checks for unused expressions. It'll simplify us not having to jump through hoops to ensure that we never output overflow checks for these operations and I can imagine situations where people expect overflow checks on some arithmetic which they don't use of the result of to constrain a certain result.

asterite:
Would that require changes to the brillig vm to not check for overflows anymore?

TomAFrench:
I'm not sure I follow, how would this affect the brillig VM?

asterite:
If we always want to give an error on overflow then we must add those checks, which will end up being some `constraint` in the SSA. Then the logic to detect overflow in Brillig might not be needed (though to be honest I don't know if there's any logic for this, or maybe it's fine to have this be redundant).

That said, I think right now the checks are deferred to Brillig for performance reasons, so moving those checks back to SSA might introduce some undesired regressions.

asterite:
> If we stop removing arithmetic operations for which the result is unused

Ah, this is what I was missing. I thought we'd add the checks in SSA, then DIE wouldn't remove those instructions, but in the end not removing the arithmetic operations that could overflow is the same, and simpler.

PR #7590, opened 2025-03-05
feat: C++ bindings for protobuf

aakoshh:
# Description

## Problem\*

Part of #7511 

## Summary\*

Generates C++ bindings for the Protobuf schemas, to be used by Barretenberg.

I originally thought this would be convenient since we already have the `acir.cpp` and `witness.cpp` files generated, which contain the models necessary to deserialise the `bincode` encoded `Program` and `WitnessMap`, and at this point we already have `protoc` which we can invoke right there in the same step, always keeping the two in sync.

On second thought, this is probably best left to be done on the `aztec-packages` side: as long as the `.proto` schemas are available to that build, we can decide there which ones need to be built, and also make sure they compile (ie. that we have not forgotten anything, because unlike `prost`, `protoc` only generates source for what we tell it to do explicitly). It is also more conceptually sound because then the consumer decides what bindings to generate, rather we having to know exactly what `bb` can ignore (that is, we don't generate C++ `program.proto` and `brillig.proto`).

I'm opening the PR for the record, but I think we should close it and wait until we have synced the schemas over to `aztec-packages` and do it there.

We still have to generate `acir.cpp` and `witness.cpp` on the Noir side because it works with reflection, and in `aztec-packages` our job is to update e.g. `acir.hpp`, generate fresh Protobuf mappings, then manually do the mapping between them whenever something changes that we want to use on that side.

## Documentation\*

Check one:

# PR Checklist\*


aakoshh:
@TomAFrench do you agree that we close this and do it in `aztec-packages` instead?

TomAFrench:
Hey, yeah I agree that once we have the proto file then we can leave it up to aztec-packages to generate their bindings.

PR #7595, opened 2025-03-05
feat(ssa): Post dominator tree

### LINKED ISSUE #7571: Post-Dominator Tree
### Problem

As we expand our Brillig optimizations we need to more easily identify control flow. 

In a dominator tree, we say that a node _n_ is dominated by a node _d_ if every path from the entry node to _n_ must pass through _d_. 

In a post-dominator tree (PDT) we want to look at the exit node. A node _n_ post-dominates a node _d_ if all paths to the exit node (node with no successors) from node _d_ must pass through _n_. 

PDTs can greatly simplify control flow analysis. For example, as part of https://github.com/noir-lang/noir/issues/7569 we want to identify whether an instruction in a loop is under the same condition as the loop itself so that the instruction can be hoisted. 

### Happy Case

We should be able to construct a post-dominator tree in our SSA similar to the current dominator tree by reversing the CFG. 

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

Yes

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #7571 

## Summary\*

This PR is mostly contained to tests as it will be the basis for further control dependence analysis.

1. Logic for reversing the CFG
2. Switched post order computation to use the CFG rather than the function definition
3. Add a test to show using the the reversed CFG and updated post order provides an accurate post-dominator tree

## Additional Context

The new `reverse` and `with_function_post_dom` methods are only used in tests. The `#[cfg(test)]` attribute will be removed as these methods are used by later work.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/cfg.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/cfg.rs
@@ -125,6 +125,7 @@ impl ControlFlowGraph {
     }
 
     /// Reverse the control flow graph, swapping predecessors and successors.
+    #[cfg(test)]
     pub(crate) fn reverse(&self) -> Self {
         let mut reversed_cfg = ControlFlowGraph::default();
 
@@ -147,7 +148,8 @@ impl ControlFlowGraph {
 #[cfg(test)]
 mod tests {
     use crate::ssa::ir::{
-        call_stack::CallStackId, instruction::TerminatorInstruction, map::Id, types::Type,
+        basic_block::BasicBlockId, call_stack::CallStackId, instruction::TerminatorInstruction,
+        map::Id, types::Type,
     };
 
     use super::{super::function::Function, ControlFlowGraph};
@@ -165,8 +167,7 @@ mod tests {
         ControlFlowGraph::with_function(&func);
     }
 
-    #[test]
-    fn jumps() {
+    fn build_test_function() -> (Function, BasicBlockId, BasicBlockId, BasicBlockId) {
         // Build function of form
         // fn func {
         //   block0(cond: u1):
@@ -200,6 +201,50 @@ mod tests {
             call_stack: CallStackId::root(),
         });
 
+        (func, block0_id, block1_id, block2_id)
+    }
+
+    fn modify_test_function(
+        func: &mut Function,
+        block0_id: BasicBlockId,
+        block1_id: BasicBlockId,
+        block2_id: BasicBlockId,
+    ) -> BasicBlockId {
+        // Modify function to form:
+        // fn func {
+        //   block0(cond: u1):
+        //     jmpif cond, then: block1, else: ret_block
+        //   block1():
+        //     jmpif cond, then: block1, else: block2
+        //   block2():
+        //     jmp ret_block()
+        //   ret_block():
+        //     return ()
+        // }
+        let ret_block_id = func.dfg.make_block();
+        func.dfg[ret_block_id].set_terminator(TerminatorInstruction::Return {
+            return_values: vec![],
+            call_stack: CallStackId::root(),
+        });
+        func.dfg[block2_id].set_terminator(TerminatorInstruction::Jmp {
+            destination: ret_block_id,
+            arguments: vec![],
+            call_stack: CallStackId::root(),
+        });
+        let cond = func.dfg[block0_id].parameters()[0];
+        func.dfg[block0_id].set_terminator(TerminatorInstruction::JmpIf {
+            condition: cond,
+            then_destination: block1_id,
+            else_destination: ret_block_id,
+            call_stack: CallStackId::root(),
+        });
+        ret_block_id
+    }
+
+    #[test]
+    fn jumps() {
+        let (mut func, block0_id, block1_id, block2_id) = build_test_function();
+
         let mut cfg = ControlFlowGraph::with_function(&func);
 
         #[allow(clippy::needless_collect)]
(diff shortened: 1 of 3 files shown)


compiler/noirc_evaluator/src/ssa/ir/cfg.rs
```diff
@@ -115,20 +115,42 @@ impl ControlFlowGraph {
...
+                reversed_cfg.add_edge(successor, *block_id);
+            }
+        }
+
+        reversed_cfg
+    }
+
+    /// Returns the entry blocks for a CFG. This is all nodes without any predecessors.
+    pub(crate) fn compute_entry_blocks(&self) -> Vec<BasicBlockId> {
+        self.data.keys().filter(|&&block| self.predecessors(block).len() == 0).cloned().collect()
```
vezenovm, line 145:
```suggestion
        self.data.keys().filter(|&&block| self.predecessors(block).len() == 0).copied().collect()
```

vezenovm:
> `rollup-block-root`	`162` s	`122` s	`1.33`

This alert is out of date. From the bench show:
```
rollup-block-root | 125 s | 123 s | 1.02
```

compiler/noirc_evaluator/src/ssa/ir/dom.rs
```diff
@@ -159,6 +162,18 @@ impl DominatorTree {
         Self::with_cfg_and_post_order(&cfg, &post_order)
     }
 
+    /// Allocate and compute a post-dominator tree for the given function.
+    ///
+    /// This approach computes the reversed control flow graph and post-order internally and then
+    /// discards them. If either should be retained reuse it is better to instead pre-compute them
```
michaeljklein, line 168:
```suggestion
    /// discards them. If either should be retained for reuse, it is better to instead pre-compute them
```

compiler/noirc_evaluator/src/ssa/opt/normalize_value_ids.rs
```diff
@@ -76,13 +76,14 @@ impl Context {
             new_function.dfg.make_global(value.get_type().into_owned());
         }
 
-        let mut reachable_blocks = PostOrder::with_function(old_function).into_vec();
-        reachable_blocks.reverse();
-
+        let reachable_blocks = old_function.reachable_blocks().into_iter().collect::<Vec<_>>();
```
michaeljklein, line 79: Can `populate_blocks` accept an iterator to avoid `collect`ing into ` a `Vec` here?

compiler/noirc_evaluator/src/ssa/opt/normalize_value_ids.rs
```diff
@@ -76,13 +76,14 @@ impl Context {
...
         self.new_ids.populate_blocks(&reachable_blocks, old_function, new_function);
 
+        let mut rpo = PostOrder::with_function(old_function).into_vec();
```
michaeljklein, line 82: Please rename or comment that `rpo` is Reverse Post Order

compiler/noirc_evaluator/src/ssa/parser/tests.rs
```diff
@@ -139,6 +139,18 @@ fn test_multiple_blocks_and_jmp() {
...
+        acir(inline) fn main f0 {
+          b0(v0: Field):
+            jmpif v0 then: b1, else: b2
+          b1():
+            return
+          b2():
+            return
+        }
+        ";
+    assert_ssa_roundtrip(src);
```
michaeljklein, line 152: Please add a more complex `assert_ssa_roundtrip` along these lines

michaeljklein (changes requested):

compiler/noirc_evaluator/src/ssa/opt/normalize_value_ids.rs
vezenovm: yeah I think it can I'll try that

compiler/noirc_evaluator/src/ssa/opt/normalize_value_ids.rs
vezenovm: I found the code to be pretty self-documenting but I will rename

compiler/noirc_evaluator/src/ssa/parser/tests.rs
vezenovm: What kind of test were you thinking? In a way we do have many SSA parser tests through our SSA pass tests which utilize the parser.

compiler/noirc_evaluator/src/ssa/opt/normalize_value_ids.rs
vezenovm: `populate_blocks` accepts an `impl IntoIterator` now

compiler/noirc_evaluator/src/ssa/parser/tests.rs
michaeljklein: Because this involves the ordering of jumps, I was thinking of a program that jumps in a few different "directions" to ensure the ordering is correct, e.g.

```
b -> c, d
foo -> a, b
c -> e, d
d, e -> return
a -> c, b
```

where `A -> B, C` is expanded to `fn A(x) { jmpif x then B(x) else C(x)`.

Because most of our SSA tests define programs in a dependency-respecting order, e.g.

```
a -> b, c
b -> c, d
..
```

vezenovm: > where A -> B, C is expanded to fn A(x) { jmpif x then B(x) else C(x)

I am a bit confused by this notation as it diverges from our SSA and looks to be mixing functions and blocks. Either way I'll add another more complex test.

michaeljklein: Something like:

```noir
        acir(inline) fn main f0 {
          b0(v0: Field):
            v1 = call f1(v0) -> Field
            v2 = call f1(v0) -> Field
            jmpif v0 then: b0; else: b1;
          b1(v0: Field):
            return v1
          b0(v0: Field):
            return v2
        }
        ..
```

vezenovm: I added a test with two jmpifs and 3 different blocks we may jump to

michaeljklein (approved):

PR #7598, opened 2025-03-05
fix(LSP): references/rename only when underlying span has the correct…

### LINKED ISSUE #7557: LSP rename sometimes incorrectly renames comptime code
### Aim

Image

### Expected Behavior

The quote code shouldn't be changed.

### Bug

The quote code is changed.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
… name

# Description

## Problem

Resolves #7557

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/lsp/src/requests/mod.rs
+++ b/tooling/lsp/src/requests/mod.rs
@@ -19,6 +19,7 @@ use nargo_fmt::Config;
 use noirc_frontend::ast::Ident;
 use noirc_frontend::graph::CrateId;
 use noirc_frontend::hir::def_map::{CrateDefMap, ModuleId};
+use noirc_frontend::node_interner::ReferenceId;
 use noirc_frontend::parser::ParserError;
 use noirc_frontend::usage_tracker::UsageTracker;
 use noirc_frontend::{graph::Dependency, node_interner::NodeInterner};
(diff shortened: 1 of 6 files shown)


tooling/lsp/src/requests/mod.rs
```diff
@@ -590,68 +591,104 @@ pub(crate) fn find_all_references_in_workspace(
...
-        .map(|locations| {
-            locations
-                .iter()
-                .filter_map(|location| to_lsp_location(files, location.file, location.span))
-                .collect()
-        })
         .unwrap_or_default()
 }
 
+fn get_reference_name(reference: ReferenceId, interner: &NodeInterner) -> Option<String> {
```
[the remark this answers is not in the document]
asterite: Since it's only used here and it doesn't require anything internal to `interner` I thought of putting it here. Otherwise the method on `interner` needs to be public and if we later change how this is implemented we might forget to remove that method.

tooling/lsp/src/requests/references.rs
```diff
@@ -170,4 +169,65 @@ mod references_tests {
...
+        let (mut state, noir_text_document) = test_utils::init_lsp_server("document_symbol").await;
+
+        let (line, column) = src
+            .lines()
+            .enumerate()
+            .filter_map(|(line_index, line)| {
+                line.find(">|<").map(|char_index| (line_index, char_index))
+            })
+            .next()
+            .expect("Expected to find one >|< in the source code");
```
aakoshh, line 200: This pattern of finding the location of `">|<"` appears quite a few times now, along with the `apply_text_edits(&src.replace(">|<", ""), &item.additional_text_edits.unwrap());` replacement. Perhaps two helper functions wouldn't be a bad idea?

asterite: Good point, I was being lazy here. I thought it's only used in completion tests but even then there's no excuse for duplicating this. I'll update it.

aakoshh (approved):
I think I understand what's happening in the gif, the fix makes sense to me, although I'm not that familiar with the interner and its reference graph.

tooling/lsp/src/requests/references.rs
asterite: Done!

TomAFrench (approved):

PR #7602, opened 2025-03-05
fix: compare Quoted by expanding interned values

### LINKED ISSUE #5979: Comparing Quoted values doesn't give correct result for interned values
### Aim

If a Quoted value has interned tokens, those are compared by ID. If two identical values were interned, their comparison would fail because their ID is different.

### Expected Behavior

When comparing quoted values, interned tokens should be resolved, and those resolved values should be compared.

### Bug

Here's an example:

```noir
let expr = quote { foo::bar::Baz::<i32, i64> { a: 1, b: 2 } }.as_expr().unwrap();
let expr = expr.modify(times_two);
let (typ, fields) = expr.as_constructor().unwrap();
assert_eq(typ, quote { foo::bar::Baz::<i32, i64> }); // This line fails
assert_eq(fields.len(), 2);
assert_eq(fields[0].0, quote { a });
assert_eq(fields[0].1.as_integer().unwrap(), (2, false));
assert_eq(fields[1].0, quote { b });
assert_eq(fields[1].1.as_integer().unwrap(), (4, false));
```

The first `assert_eq` fails, but it shouldn't.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #5979

## Summary

Compares two Quoted values by turning them into strings, then comparing the strings.

Another way to do this would be to expand interned nodes to the nodes they represent, then turning those nodes into tokens, then finally comparing all tokens. However, that involves more code (or at least code that we currently don't have) while comparing strings works as well and might have more or less the same performance (we might need to allocate tokens with the other approach).

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
@@ -28,6 +28,7 @@ use crate::{
     hir::{
         comptime::{
             InterpreterError, Value,
+            display::tokens_to_string,
             errors::IResult,
             value::{ExprValue, TypedExpr},
         },
@@ -166,7 +167,7 @@ impl Interpreter<'_, '_> {
             "quoted_as_module" => quoted_as_module(self, arguments, return_type, location),
             "quoted_as_trait_constraint" => quoted_as_trait_constraint(self, arguments, location),
             "quoted_as_type" => quoted_as_type(self, arguments, location),
-            "quoted_eq" => quoted_eq(arguments, location),
+            "quoted_eq" => quoted_eq(self.elaborator.interner, arguments, location),
             "quoted_hash" => quoted_hash(arguments, location),
             "quoted_tokens" => quoted_tokens(arguments, location),
             "slice_insert" => slice_insert(interner, arguments, location),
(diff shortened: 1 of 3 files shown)


compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -2913,10 +2914,24 @@ fn modulus_num_bits(arguments: Vec<(Value, Location)>, location: Location) -> IR
...
+    let (self_arg, other_arg) = check_two_arguments(arguments, location)?;
+    let self_arg = get_quoted(self_arg)?;
+    let other_arg = get_quoted(other_arg)?;
+
+    // Comparing tokens one against each other doesn't work in the general case because tokens
+    // might be refer to interned expressions/statements/etc. We'd need to convert those nodes
+    // to tokens and compare the final result, but comparing their string representation works
+    // equally well and, for simplicity, that's what we do here.
+    let self_string = tokens_to_string(&self_arg, interner);
+    let other_string = tokens_to_string(&other_arg, interner);
```
jfecher, line 2931: Should we change the Quoted Eq implementation as well for consistency?

Or keep a separation between the type within Noir and what we expose to the comptime API?

asterite: Ah, good point. The issue is that we need an `interner` in order to compare them so I'm not sure how that can be done in `Eq`.

jfecher: Ah, ok. It's probably fine to have them separate like this.

jfecher (approved):
LGTM. Not sure the cause of the failure in aztec packages though.

asterite:
It seems to be a timeout.

jfecher:
> It seems to be a timeout.

Whoops, got my PRs mixed up :sweat_smile:

Issue #7287, opened 2025-02-05
Rework `inc_rc`, `dec_rc` and how arrays are handled in brillig in general


jfecher:
### Problem

`inc_rc` and `dec_rc` are a common source of bugs in SSA code generated by the compiler. They require special handling since they are effectively implicit variables attached to each array in unconstrained code.

### Happy Case

Now that runtimes are separated before SSA we could represent arrays differently in acir & brillig. In Acir they can continue to be represented as normal arrays, while in brillig we could lower them as a tuple of `(&mut Field, [T; N])` where the first element of the tuple is the reference count represented as a mutable field value, and the second is the array itself. If we do this, reference counts now need much fewer special handling or optimizations since they'll be handled by mem2reg along with other references.

This could make it more difficult for optimizations like deduplication to insert extra `inc_rc` equivalents since they'd have to find the integer that represents the array's reference count instead of it always being attached to the array.

Uncertain how much code improvement this will bring or if it will improve optimizations so I labeled this as "experiment" with rather than a "lets definitely do this."

EDIT: See comments below for some alternative approaches.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

aakoshh:
Just to make sure I understand, you're not suggesting that `inc_rc` will disappear from the code, but rather that instead of figuring out which array it needs to act on, with the ref count being an internal part of the array in the Brillig VM, it will instead be like the length of a BoundedVec, a number passed around in the SSA together with the array used for storage. But where the  `inc_rc` and `dec_rc` are inserted into the SSA will remain a source of speculation about how the array is going to be used, as it is now. 

I'm asking because I have difficulty understanding some of the SSA generated from the test that fails in #7297 

For example `copy_mut` seems to insert an `inc_rc` because it receives a `mut array`, but it doesn't seem to account for the possibility that the array might not be mutated, and the count stays higher after the call no matter what. 

Take this example:

<details>
<summary>maybe_mut</summary>

```rust
use std::mem::array_refcount;

fn main() {
    let mut array = [0, 1, 2];
    maybe_mut(array, false);
    maybe_mut(array, true);
    println(array[0]);
    println(array_refcount(array));
}

fn maybe_mut(mut array: [Field; 3], should_mutate: bool) {
    if should_mutate {
        array[0] = 4;
    }
    println(array[0]);
}
```
It prints 3 as the final refcount in `main`, because of how `maybe_mut` looks like:

```rust
[...]

PR #7603, opened 2025-03-05
feat(experimental): Enable ownership syntax

jfecher:
# Description

## Problem\*

Working towards https://github.com/noir-lang/noir/issues/7287

## Summary\*

Implements just the syntax for (and type system components) for ownership:
- `&Type`
- `&expr`
- Renames `UnstableFeature::ArrayOwnership` to `UnstableFeature::Ownership` since if arrays are owned anything containing them must be too.

## Additional Context

No runtime differences are expected from this PR

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/ast/visitor.rs
+++ b/compiler/noirc_frontend/src/ast/visitor.rs
@@ -397,10 +397,6 @@ pub trait Visitor {
         true
     }
 
-    fn visit_mutable_reference_type(&mut self, _: &UnresolvedType, _: Span) -> bool {
-        true
-    }
-
     fn visit_tuple_type(&mut self, _: &[UnresolvedType], _: Span) -> bool {
         true
     }
--- a/compiler/noirc_frontend/src/ast/visitor.rs
+++ b/compiler/noirc_frontend/src/ast/visitor.rs
@@ -393,6 +393,10 @@ pub trait Visitor {
         true
     }
 
+    fn visit_reference_type(&mut self, _: &UnresolvedType, _mutable: bool, _: Span) -> bool {
+        true
+    }
+
     fn visit_mutable_reference_type(&mut self, _: &UnresolvedType, _: Span) -> bool {
         true
     }
@@ -1382,8 +1386,8 @@ impl UnresolvedType {
                     generic_type_args.accept(visitor);
                 }
             }
-            UnresolvedTypeData::MutableReference(unresolved_type) => {
-                if visitor.visit_mutable_reference_type(unresolved_type, self.location.span) {
+            UnresolvedTypeData::Reference(unresolved_type, mutable) => {
+                if visitor.visit_reference_type(unresolved_type, *mutable, self.location.span) {
                     unresolved_type.accept(visitor);
                 }
             }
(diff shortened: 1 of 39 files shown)


compiler/noirc_frontend/src/ast/visitor.rs
```diff
@@ -393,6 +393,10 @@ pub trait Visitor {
         true
     }
 
+    fn visit_reference_type(&mut self, _: &UnresolvedType, _mutable: bool, _: Span) -> bool {
+        true
+    }
+
     fn visit_mutable_reference_type(&mut self, _: &UnresolvedType, _: Span) -> bool {
```
asterite, line 400: I think this one can be removed now.

compiler/noirc_frontend/src/parser/parser/types.rs
```diff
@@ -370,15 +370,15 @@ impl Parser<'_> {
     }
 
     fn parses_mutable_reference_type(&mut self) -> Option<UnresolvedTypeData> {
-        if self.eat(Token::Ampersand) {
-            if !self.eat_keyword(Keyword::Mut) {
-                self.expected_mut_after_ampersand();
-            }
+        // The `&` may be lexed as a slice start if this is an array or slice type
+        if self.eat(Token::Ampersand) || self.eat(Token::SliceStart) {
```
asterite, line 374: I don't understand this. If we get "&[" then we also try to eat "mut" next, so we'd get "&[ mut" which isn't valid syntax.

tooling/lsp/src/requests/signature_help.rs
```diff
@@ -146,7 +146,7 @@ impl<'a> SignatureFinder<'a> {
             }
 
             if has_self && index == 0 {
-                if let Type::MutableReference(..) = typ {
+                if let Type::Reference(..) = typ {
                     label.push_str("&mut self");
```
asterite, line 150: Maybe check `mutable` here too.

compiler/noirc_frontend/src/lexer/token.rs
```diff
@@ -522,6 +527,7 @@ impl fmt::Display for Token {
             Token::Slash => write!(f, "/"),
             Token::Percent => write!(f, "%"),
             Token::Ampersand => write!(f, "&"),
+            Token::SliceStart => write!(f, "&"),
```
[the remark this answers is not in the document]
jfecher: SliceStart is still just the leading `&` token, we don't lex `&[` as a double character token. It just simplified parsing a bit this way.

compiler/noirc_frontend/src/parser/parser/types.rs
jfecher: `&[` would be `SliceStart` followed by a `LeftBracket`

compiler/noirc_frontend/src/lexer/token.rs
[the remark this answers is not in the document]
asterite: Oooh, I see. I didn't understand why it was documented as a hack, but now I do :-). All good, I think it's okay.

asterite (approved):
Looks good! (though apparently something broke)

PR #7605, opened 2025-03-06
feat: Rename noir-execute to noir-executor

aakoshh:
# Description

## Problem\*

I want to add more commands related to the execution of artefacts without increasing the number of binaries.

## Summary\*

Renames `noir-execute <args>` to `noir-executor execute <args>`, so that we can add new commands related to artifact execution in the future. `noir-execute` is not installed yet with `noirup` so this shouldn't break anything at the moment.

## Additional Context

https://github.com/AztecProtocol/aztec-packages/pull/12148 shows what a "dump" from TypeScript will look like. Notably the inputs to the circuit are encoded similar to how the ACVM CLI takes them: as a witness map, as opposed to the human readable ABI encoded data structure we use in `Prover.toml` with `nargo execute` and `noir-execute`. 

We can have another command to deal with this input format as it is, e.g. _replay_, which we could invoke as `noir-replay`, but I also thought we might want to take the witness map from the recording and write it into a `Prover.toml` file which we could use in integration tests. 

This is something I did [during debugging](https://github.com/AztecProtocol/aztec-packages/pull/11294#issuecomment-2604803423) by logging the TypeScript DTOs to the console in JSON format, however that is not easy to capture in the dump due to where it happens: it's just a transient value in a conversion function.

We could call such a command e.g. `noir-abi decode` for example, with a potential `noir-abi encode` to do the opposite. So that's already two more executables to install along `noir-execute`, `noir-profiler`, `noir-inspector`, etc.

Originally I proposed to have a `noir-artifact` command to bundle artifact related commands. This PR is an alternative to that where we could group execution related commands: `noir-executor execute`, `noir-executor replay`, `noir-executor abi decode` for example. 

An alternative is to introduce an umbrella `noir` command at some point could would transparently forward `noir <cmd>` to `noir-<cmd>`, which what @asterite suggested `git` does. 

Even then, I think having a separate `execute` and `replay` might be better than `execute` taking either a dump or separate args for inputs and the oracle transcript (which is in itself optional, it could come from RPC, or nowhere).

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
> An alternative is to introduce an umbrella noir command at some point could would transparently forward noir <cmd> to noir-<cmd>, which what @asterite suggested git does.

I would be in support of doing this at some point.

TomAFrench:
I'll have a proper think about the knock on changes you mention in the OP today.

michaeljklein (approved):

TomAFrench (changes requested):
Short block while we discuss the CLI structure as mentioned.

TomAFrench:
> We can have another command to deal with this input format as it is, e.g. replay, which we could invoke as `noir-replay`,

I'm not clear on why this needs to be a separate command to be honest. We already have a CLI option where a user can specify an oracle file so this should be covered by `noir-execute --oracle-file=/path/to/dump/file`

https://github.com/noir-lang/noir/blob/637deea91a01a636148448125d6c263d48cba548/tooling/artifact_cli/src/commands/execute_cmd.rs#L46-L53

My understanding is that this CLI option was created to support the transcripts that Nico's team were creating for us? If not, then why?

> Originally I proposed to have a noir-artifact command to bundle artifact related commands. This PR is an alternative to that where we could group execution related commands: noir-executor execute, noir-executor replay, noir-executor abi decode for example.
>
>An alternative is to introduce an umbrella noir command at some point, that would transparently forward noir <cmd> to noir-<cmd>, which what @asterite suggested git does.

My preference is that we have a suite of small tools which are composable through bash, likely looking similar to git as Ary  suggests but we can avoid having to shell out to other binaries for now.

We _should_ add `noir-abi` imo. Having a tool which allows us to interact with witness maps more cleanly on the CLI is very useful (I can definitely see there being benefits for recursive proof workflows) but those don't fit into a `noir-executor` tool.

Part of why I pushed back against `noir_artifact_cli` is that we should explicitly avoid creating "everything" tools. `nargo` was this previously and we've been gradually taking chunks out of it so that the noir ecosystem is more modular and composable and `noir_artifact_cli` went against this.

PR #7609, opened 2025-03-06
feat: relate errors to macro built-ins errors

### LINKED ISSUE #7459: Relate errors to macro built-ins errors
In https://github.com/noir-lang/noir/pull/7333 we now show macro errors where they happen, together with why they happened (for example: a function attribute was running). However, this still misses a few cases, essentially when `elaborate_in_function` is used.

For example this program:

```noir
fn main() {
    comptime {
        let x = quote { Hello };
        let _ = x.as_type();
    }
}
```

gives this error:

```
error: Could not resolve 'Hello' in path
   ┌─ src/main.nr:11:25
   │
11 │         let x = quote { Hello };
   │                         -----
```

but it doesn't mention that it's because of the `as_type()` call.

We should also take this as an opportunity to address this comment: https://github.com/noir-lang/noir/pull/7333#discussion_r1963987955

asterite:
# Description

## Problem

Resolves #7459

## Summary

Now this code:

```noir
fn main() {
    comptime {
        let x = quote { Hello };
        let _ = x.as_type();
    }
}
```

Gives this error:

```
error: Could not resolve 'Hello' in path
  ┌─ src/main.nr:3:25
  │
3 │         let x = quote { Hello };
  │                         -----
4 │         let _ = x.as_type();
  │                 ----------- While evaluating `Quoted::as_type`
```

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/comptime.rs
+++ b/compiler/noirc_frontend/src/elaborator/comptime.rs
@@ -49,9 +49,10 @@ impl<'context> Elaborator<'context> {
     pub fn elaborate_item_from_comptime_in_function<'a, T>(
         &'a mut self,
         current_function: Option<FuncId>,
+        reason: Option<ElaborateReason>,
         f: impl FnOnce(&mut Elaborator<'a>) -> T,
     ) -> T {
-        self.elaborate_item_from_comptime(f, |elaborator| {
+        self.elaborate_item_from_comptime(reason, f, |elaborator| {
             if let Some(function) = current_function {
                 let meta = elaborator.interner.function_meta(&function);
                 elaborator.current_item = Some(DependencyId::Function(function));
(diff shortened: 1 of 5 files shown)


compiler/noirc_frontend/src/elaborator/comptime.rs
```diff
@@ -104,7 +107,14 @@ impl<'context> Elaborator<'context> {
         let result = f(&mut elaborator);
         elaborator.check_and_pop_function_context();
 
-        self.errors.append(&mut elaborator.errors);
+        let mut errors = std::mem::take(&mut elaborator.errors);
+        if let Some(reason) = reason {
+            errors = vecmap(errors, |error| {
+                CompilationError::ComptimeError(reason.to_macro_error(error))
+            });
```
jfecher, line 114: I was going to comment on allocating a new Vec possibly being unnecessary here although it seems rust may actually optimize this case to reuse the same Vec.

jfecher (approved):
Neat!

compiler/noirc_frontend/src/elaborator/comptime.rs
asterite: Ah, yes, I thought about it but I couldn't find a method to do it. Or, well, there's `iter_mut()` but I imagined it would make the code a bit longer. I'll try to send a small PR tomorrow to do this.

jfecher: It's alright - seems Rust already optimizes this case after all. You won't be able to use `.iter_mut` without cloning the error itself, and I don't think you'll be able to use `drain` either since it'd borrow the vector while it is draining and you wouldn't be able to reassign to the same vector while it is borrowed without collecting to another Vec anyway first.

PR #7629, opened 2025-03-07
fix: Experiment with `DecrementRc`

jfecher:
# Description

## Problem\*

Resolves 

## Summary\*

This is an experimental branch where I'm hoping to fix the underlying issues with our usage of `dec_rc`. Currently this PR:
- Reverts `dec_rc value original_value` back to a single valued `dec_rc value`
- Changes `dec_rc` on parameters to always be issued on the original load value rather than creating a new load. This should fix the original underflow issue we changed dec_rc for.
- Splits the `reference_counts` into a separate test for each inliner setting: min, 0, and max. The inliner `0` setting was actually seeing different optimization characteristics compared to the other two just due to optimization ordering. The behavior is still correct though.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
Opening this to review because I think this solves the original issue which https://github.com/noir-lang/noir/pull/7297 patched. As far as I can see performance with this PR is similar to before it, although ref counts are slightly different depending on inliner settings.

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -999,11 +999,15 @@ impl<'a> FunctionContext<'a> {
         mut incremented_params: Vec<(ValueId, ValueId)>,
         terminator_args: &[ValueId],
     ) {
+        // TODO: This check likely leads to unsoundness.
```
vezenovm, line 1002: Should we link to our existing issue about reworking RCs?

tooling/nargo_cli/build.rs
```diff
@@ -60,10 +60,20 @@ const IGNORED_BRILLIG_TESTS: [&str; 10] = [
...
+    ("reference_counts_inliner_0", 0),
+    ("reference_counts_inliner_min", i64::MIN),
+    ("reference_counts_inliner_max", i64::MAX),
+];
+
+/// Tests which aren't expected to work with the default maximum inliner cases.
+const INLINER_MAX_OVERRIDES: [(&str, i64); 3] = [
+    ("reference_counts_inliner_0", 0),
+    ("reference_counts_inliner_min", i64::MIN),
+    ("reference_counts_inliner_max", i64::MAX),
```
vezenovm, line 76:
```suggestion
```
Why do we need this here? If we don't find a test `max_inliner` will just be set to `i64::MAX`, same with the min case above?

test_programs/execution_success/reference_counts_inliner_0/src/main.nr
vezenovm, line 1: Could you elaborate on why we needed to separate the tests? I see this from the PR description:

> The inliner 0 setting was actually seeing different optimization characteristics compared to the other two just due > to optimization ordering

What do you mean by optimization ordering?

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -872,65 +871,40 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
...
+                    self.brillig_context.memory_op_instruction(
+                        min_addr,
+                        rc_register,
+                        condition.address,
+                        BrilligBinaryOp::LessThan,
+                    );
+                    self.brillig_context.codegen_constrain(
                         condition,
-                        BrilligBinaryOp::Equals,
+                        Some("array ref-count underflowed below zero".to_owned()),
```
vezenovm, line 896:
```suggestion
                        Some("array ref-count underflow detected".to_owned()),
```
or add `underflowed` to cspell

vezenovm (commented):
LGTM just curious about the splitting up of `reference_counts`

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
jfecher: No, I think this unsoundness issue is separate. I made an issue here: https://github.com/noir-lang/noir/issues/7737. I should have a fix for it today.

tooling/nargo_cli/build.rs
jfecher: I just thought it was clearer to specify all of these tests since they're only meant to be run on that one setting. I can remove them though if you'd prefer

test_programs/execution_success/reference_counts_inliner_0/src/main.nr
jfecher: It's a bit complex unfortunately. I mean the ordering optimizations are applied in SSA depends on the inliner setting since whether something is inlined affects whether other optimizations are performed. If we inline everything early for example we may immediately see some inc/dec pairs that can be optimized out. If we inline only a few functions or only inline in the later inlining step then this may not happen or may not happen only in the non-inlined functions. Similarly for if we do not inline at all.

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
jfecher: I may need to change this check - it looks like it is triggering on a fresh array that gets decremented: `v0 = make_array; dec_rc v0` even though that should just change it to an rc of 0, not below 0. Any ideas?

test_programs/execution_success/reference_counts_inliner_0/src/main.nr
vezenovm: Ah ok, I wasn't sure if you were referring to something other than how optimizations happen with different inliner settings. I do find it clearer to have a single test as we then already handle multiple inliner settings in our CLI tests. I would rather stick to the paradigm of relying on the test matrix for different settings.

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
jfecher: Not sure - I checked and it looks like we do initialize to 1. I was assuming it must be my changed check that was wrong but I don't see why that'd be either.

test_programs/execution_success/reference_counts_inliner_0/src/main.nr
jfecher: > I do find it clearer to have a single test as we then already handle multiple inliner settings in our CLI tests

The problem is the result of this test changes depending on the inliner settings since we're asserting on the reference counts at each step, which may be optimized out in some inliner settings.

jfecher:
Looks like merging master broke the tests

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
vezenovm: The changed check looks correct. 

How were you able to trigger this code `v0 = make_array; dec_rc v0`? I'm struggling to reproduce.

test_programs/execution_success/reference_counts_inliner_0/src/main.nr
vezenovm: > The problem is the result of this test changes depending on the inliner settings

Yeah I understand that. But why is that a problem? We expect different results for different inliner settings?

jfecher: Yes, the actual reference count changes and that test is littered with `assert_refcount(_, expected_rc)` where expected_rc changes depending on the optimization setting

vezenovm: alright, let's just keep the separate tests then.

tooling/nargo_cli/build.rs
vezenovm: Let's just comment that, as I was confused to see `reference_counts_inliner_max` here.

jfecher: I thought that would be too messy for such a small thing so I went to remove the max/min cases from inliner max/min excludes instead but then figured I may as well make a separate list entirely for tests which should only be run on exactly 1 inliner setting.

vezenovm: Yeah I like this more

vezenovm (approved):

PR #7641, opened 2025-03-10
feat: ssa fuzzer

jewelofchaos9:
# Description

## Summary\*
This PR introduces SSA fuzzer for testing ACIR and Brillig execution of ssa built from scratch. 

## Documentation\*

Check one:

# PR Checklist\*


tooling/nargo/src/foreign_calls/default.rs
```diff
@@ -121,7 +121,6 @@ pub type DefaultForeignCallLayers<'a, B, F> = Layer<
 >;
 
 /// Convenience constructor for code that used to create the executor this way.
-#[cfg(feature = "rpc")]
 pub struct DefaultForeignCallExecutor;
```
Rumata888, line 124: Why have you removed this? Please bring it back

tooling/nargo/src/foreign_calls/mod.rs
```diff
@@ -9,7 +9,6 @@ pub mod default;
 #[cfg(feature = "rpc")]
 pub mod rpc;
 pub use default::DefaultForeignCallBuilder;
-#[cfg(feature = "rpc")]
 pub use default::DefaultForeignCallExecutor;
```
Rumata888, line 12: Why have you removed it?

tooling/ssa_fuzzer/fuzzer/README.md
```diff
@@ -0,0 +1,17 @@
+## setup
```
Rumata888, line 1: I would appreciate a proper readme with explanation of what the tool is doing, why we need to delete the rus toolchain, etc

tooling/ssa_fuzzer/fuzzer/fuzz_targets/uint.rs
```diff
@@ -0,0 +1,419 @@
+#![no_main]
+
+use acvm::acir::native_types::Witness;
```
Rumata888, line 3: Please add comments throughout the file

tooling/ssa_fuzzer/fuzzer/fuzz_targets/field.rs
```diff
@@ -0,0 +1,454 @@
...
+fn index_presented(
+    index: u32,
+    acir_witnesses_indeces: &mut Vec<u32>,
+    brillig_witnesses_indeces: &mut Vec<u32>,
+) -> bool {
+    acir_witnesses_indeces.contains(&index) && brillig_witnesses_indeces.contains(&index)
+}
+
+/// Checks if two indices exist in both ACIR and Brillig witness vectors
+fn both_indeces_presented(
```
Rumata888, line 78: indices

Rumata888: Rename to "index_exists_in_both_witness_vectors"

tooling/ssa_fuzzer/fuzzer/fuzz_targets/field.rs
```diff
@@ -0,0 +1,454 @@
...
+    /// Array element access
+    ArrayGet { array: u32, index: u32 },
+    /// Array element assignment
+    ArraySet { array: u32, index: u32, value: u32 },
+    /// Array creation
+    MakeArray { elements: Vec<u32> },
+}
+
+/// Checks if an index exists in both ACIR and Brillig witness vectors
+fn index_presented(
```
Rumata888, line 69: index_exists

tooling/ssa_fuzzer/fuzzer/fuzz_targets/field.rs
```diff
@@ -0,0 +1,454 @@
...
+        let brillig_len = brillig_values_ids.len();
+        let acir_array = self.acir_builder.insert_make_array(acir_values_ids);
+        let brillig_array = self.brillig_builder.insert_make_array(brillig_values_ids);
+        self.acir_arrays.push(Array { id: acir_array, length: acir_len as u32 });
+        self.brillig_arrays.push(Array { id: brillig_array, length: brillig_len as u32 });
+    }
+
+    /// Gets an element from an array at the given index
+    fn insert_array_get(&mut self, array_idx: u32, index: u32) {
+        if array_idx >= self.acir_arrays.len() as u32
```
Rumata888, line 161: It is better to just use indices modulo lengths. Otherwise the majority of testcases will skip instructions instead of creating them

tooling/ssa_fuzzer/fuzzer/fuzz_targets/field.rs
```diff
@@ -0,0 +1,454 @@
...
+        let acir_result = self.acir_builder.insert_array_get(acir_array, acir_id);
+        let brillig_result = self.brillig_builder.insert_array_get(brillig_array, brillig_id);
+        self.acir_ids.push(id_to_int(acir_result));
+        self.brillig_ids.push(id_to_int(brillig_result));
+    }
+
+    /// Sets an element in an array at the given index
+    fn insert_array_set(&mut self, array_idx: u32, index: u32, value: u32) {
+        if array_idx >= self.acir_arrays.len() as u32
+            || array_idx >= self.brillig_arrays.len() as u32
```
Rumata888, line 185: Same here

tooling/ssa_fuzzer/fuzzer/fuzz_targets/uint.rs
```diff
@@ -0,0 +1,498 @@
...
+    /// Array element access
+    ArrayGet { array: u32, index: u32 },
+    /// Array element assignment
+    ArraySet { array: u32, index: u32, value: u32 },
+    /// Array creation
+    MakeArray { elements: Vec<u32> },
+}
+
+/// Checks if an index exists in both ACIR and Brillig witness vectors
+fn index_presented(
```
Rumata888, line 78: It seems like you are duplicating functionality from field.rs here

Rumata888 (changes requested):

tooling/nargo/src/ops/mod.rs
```diff
@@ -11,7 +11,7 @@ pub use self::test::{TestStatus, run_test};
 
 mod check;
 mod compile;
-mod execute;
+pub mod execute;
```
TomAFrench, line 14:
```suggestion
mod execute;
```

jewelofchaos9: I need it [here](https://github.com/noir-lang/noir/blob/399b0f47a394ad33102e375e26675aa049f35708/tooling/ssa_fuzzer/src/runner.rs#L12)

TomAFrench: We reexport that function a couple of lines up.

tooling/ssa_fuzzer/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+            let mut acir_builder = FuzzerBuilder::new_acir();
+            let mut brillig_builder = FuzzerBuilder::new_brillig();
+            acir_builder.insert_variables(typ.clone());
+            brillig_builder.insert_variables(typ.clone());
+            Self { acir_builder, brillig_builder }
+        }
+
+        fn insert_instruction_double_arg(
+            &mut self,
+            instruction: fn(&mut FuzzerBuilder, Id<Value>, Id<Value>) -> Id<Value>,
```
aakoshh, line 34: Can you add some docs about what the `instruction` does? `acir_result` sounds like it could be execution result, but I guess it's the ID of the returned witness. Maybe call it `acir_return` instead of `_result`, so it doesn't give `Result<_, _>` vibes.

You could create a type alias for `fn(&mut FuzzerBuilder, Id<Value>, Id<Value>) -> Id<Value>` to capture the explanation in one place.

tooling/ssa_fuzzer/src/config.rs
```diff
@@ -0,0 +1,9 @@
+#![allow(dead_code)]
```
aakoshh, line 1: Still needed?

tooling/ssa_fuzzer/fuzzer/fuzz_targets/field.rs
```diff
@@ -0,0 +1,473 @@
...
+    /// then it means that no instructions were added, so we just return the last variable set
+    fn get_return_witnesses(&mut self) -> (Witness, Witness) {
+        let acir_result_index = *self.acir_ids.last().unwrap();
+        let brillig_result_index = *self.brillig_ids.last().unwrap();
+        let mut acir_result_witness = Witness(acir_result_index);
+        let mut brillig_result_witness = Witness(brillig_result_index);
+
+        if self.acir_ids.len() as u32 != config::NUMBER_OF_VARIABLES_INITIAL {
+            acir_result_witness = Witness(NUMBER_OF_VARIABLES_INITIAL);
+            brillig_result_witness = Witness(NUMBER_OF_VARIABLES_INITIAL);
```
aakoshh, line 358: > If acir_witnesses_indeces or brillig_witnesses_indeces are not equal to NUMBER_OF_VARIABLES_INITIAL,
> then ... we just return the last variable set

That's not what it seems to be doing 🤔

tooling/ssa_fuzzer/fuzzer/fuzz_targets/field.rs
```diff
@@ -0,0 +1,473 @@
...
+    }
+}
+
+/// Represents the data for the fuzzer
+/// `methods` - sequence of instructions to be added to the program
+/// `initial_witness` - initial witness values for the program as String
+#[derive(Arbitrary, Debug, Clone, Hash)]
+struct FuzzerData {
+    methods: Vec<Instructions>,
+    initial_witness: [String; config::NUMBER_OF_VARIABLES_INITIAL as usize],
```
aakoshh, line 377: Can this be any string whatsoever? Can't you create some new type and implement `Arbitrary` for it so that it generates valid fields?

tooling/ssa_fuzzer/fuzzer/fuzz_targets/field.rs
```diff
@@ -0,0 +1,473 @@
...
+        (self.acir_builder.compile(), self.brillig_builder.compile())
+    }
+}
+
+/// Represents the data for the fuzzer
+/// `methods` - sequence of instructions to be added to the program
+/// `initial_witness` - initial witness values for the program as String
+#[derive(Arbitrary, Debug, Clone, Hash)]
+struct FuzzerData {
+    methods: Vec<Instructions>,
```
aakoshh, line 376: How do the arbitrary `lhs` and `rhs` values produce valid instructions, running the full range of `u32`? Shouldn't; it be generated in a range that is limited to the inputs of the circuit or the variables created?

tooling/ssa_fuzzer/fuzzer/fuzz_targets/field.rs
```diff
@@ -0,0 +1,473 @@
...
+    let (acir_program, brillig_program) = match (acir_program, brillig_program) {
+        (Ok(acir), Ok(brillig)) => (acir, brillig),
+        (Err(_), Err(_)) => {
+            return;
+        }
+        (Ok(acir), Err(e)) => {
+            let acir_result = execute_single(&acir.program, initial_witness, acir_result_witness);
+            match acir_result {
+                Ok(result) => {
+                    println!(
```
aakoshh, line 420: Why not use `log::debug!` or other suitable level like above?

aakoshh (commented):
Comments on the targets.

tooling/ssa_fuzzer/src/builder.rs
```diff
@@ -0,0 +1,309 @@
...
+    pub fn compile(self) -> Result<CompiledProgram, CompileError> {
+        compile(self.builder, &CompileOptions::default())
+    }
+
+    /// Inserts initial variables of the given type into the function
+    pub fn insert_variables(&mut self, variable_type: Type) {
+        for _ in 0..NUMBER_OF_VARIABLES_INITIAL {
+            self.builder.add_parameter(variable_type.clone());
+        }
+        self.type_ = variable_type.clone();
```
aakoshh, line 60: Could this be done in the constructor? 

It looks like it would be nicer, because as it is one can call this method multiple times, always increasing the number of variables, and changing `type_` and `numeric_type` to the last one.

tooling/ssa_fuzzer/src/builder.rs
```diff
@@ -0,0 +1,309 @@
...
+        match variable_type {
+            Type::Numeric(numeric_type) => {
+                self.numeric_type = numeric_type;
+            }
+            _ => {
+                panic!("Unsupported variable type");
```
aakoshh, line 66: Maybe the check could be done first, no?

aakoshh (commented):
Comments on the FuzzerBuilder.

tooling/ssa_fuzzer/src/builder.rs
jewelofchaos9: No, because I need to test constant folding. Added it in latest commits

tooling/ssa_fuzzer/fuzzer/fuzz_targets/field.rs
jewelofchaos9: Its really hard to handle valid lhs and rhs, so in insert_instruction every argument taken modulo length of array of all possible args

tooling/ssa_fuzzer/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+
+        fn finalize_function(&mut self, return_value: Id<Value>) {
+            self.acir_builder.finalize_function(return_value);
+            self.brillig_builder.finalize_function(return_value);
+        }
+    }
+
+    /// Generates a random array with config::NUMBER_OF_VARIABLES_INITIAL elements
+    fn generate_values() -> Vec<u64> {
+        let mut rng = rand::thread_rng();
```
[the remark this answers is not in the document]
jewelofchaos9: Just a unit test

tooling/ssa_fuzzer/fuzzer/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "fuzz"
```
aakoshh, line 2: We might end up with having other fuzzers in the project, at which point this name will have to be a bit more specific I think.

tooling/ssa_fuzzer/fuzzer/src/field.rs
```diff
@@ -0,0 +1,86 @@
...
+    let _ = env_logger::try_init();
+    let type_ = Type::field();
+    let mut witness_map = WitnessMap::new();
+    let mut values = vec![];
+    for i in 0..config::NUMBER_OF_VARIABLES_INITIAL {
+        let witness = Witness(i);
+        let value = FieldElement::from(data.initial_witness.get(i as usize).unwrap());
+        witness_map.insert(witness, value);
+        values.push(value);
+    }
```
aakoshh, line 76: nit: It's a pity that `WitnessMap` doesn't have a constructor that takes an iterator, but you could avoid `.get(i as usize).unwrap()` by doing for example:

```rust
for (i, f) in data.initial_witness.iter().map(FieldElement::from).enumerate()) {
  witness_map.insert(Witness(i), f);
  values.push(f);
}
```

tooling/ssa_fuzzer/fuzzer/src/base_context.rs
```diff
@@ -0,0 +1,374 @@
...
+        // by default private variables ids are indexed from 0 to NUMBER_OF_VARIABLES_INITIAL - 1
+        for i in 0..config::NUMBER_OF_VARIABLES_INITIAL {
+            acir_ids.push(i);
+            brillig_ids.push(i);
+        }
+        Self {
+            acir_builder,
+            brillig_builder,
+            acir_ids,
+            brillig_ids,
```
aakoshh, line 103:
```suggestion
        // by default private variables ids are indexed from 0 to NUMBER_OF_VARIABLES_INITIAL - 1
        let ids = (0..config::NUMBER_OF_VARIABLES_INITIAL).collect::<Vec<_>>();
        Self {
            acir_builder,
            brillig_builder,
            acir_ids: ids.clone(),
            brillig_ids: ids,
```

tooling/ssa_fuzzer/src/builder.rs
aakoshh: Okay I see these two constructors:

```
    /// Creates a new fuzzer context with the given type
    pub(crate) fn new(type_: Type) -> Self {
        let mut acir_builder = FuzzerBuilder::new_acir();
        let mut brillig_builder = FuzzerBuilder::new_brillig();
        acir_builder.insert_variables(type_.clone());
        brillig_builder.insert_variables(type_.clone());
        ...
    }
    /// Creates a new fuzzer context with the given values for a constant folding checking
    pub(crate) fn new_constant(values: Vec<impl Into<FieldElement>>, type_: Type) -> Self {
        let mut acir_builder = FuzzerBuilder::new_acir();
        let mut brillig_builder = FuzzerBuilder::new_brillig();
        acir_builder.set_type(type_.clone());
        brillig_builder.set_type(type_.clone());
        ...
    }
```

So I understand why you don't want to add `NUMBER_OF_VARIABLES_INITIAL` parameters in the constructor, but it looks to me like `set_type` could be part of the `new_acir` and `new_brillig` methods, which would remove the possibility of changing it later. Maybe that's a use case as well, but it looks like it happens in both cases.

tooling/ssa_fuzzer/fuzzer/src/base_context.rs
```diff
@@ -0,0 +1,374 @@
...
+    pub(crate) fn insert_array_get(&mut self, array_idx: u32, index: u32) {
+        if self.acir_arrays.is_empty() {
+            // no arrays created
+            return;
+        }
+        // choose array by index
+        let acir_arrays_len = self.acir_arrays.len() as u32;
+        let brillig_arrays_len = self.brillig_arrays.len() as u32;
+        let acir_array = self.acir_arrays[(array_idx % acir_arrays_len) as usize];
+        let brillig_array = self.brillig_arrays[(array_idx % brillig_arrays_len) as usize];
```
aakoshh, line 163: Can `acir_array` and `brillig_array` be of different length? When do they diverge?

As I understand this is where you project arbitrary `lhs` and `rhs` numbers to something you can actually use. So, say you have 3 arrays in your program, and the arbitrary `lhs` is 1234, then here we will access array 1. 

What I'm not sure about is what happens if for example `acir_array` has 3 items, and `brillig_array` has 4 items; in that case we will access `acir_arrays[1]` and `brillig_arrays[2]`. I may be misunderstanding how they are supposed to work, but wondering if it's possible that we built up `acir_arrays` and `brillig_arrays` together, so e.g. `acir_arrays[1]` corresponds to `brillig_arrays[1]`, and to achieve consistent results we should be getting the same arrays, but for some reason later we insert `brillig_arrays[3]`, and that causes `insert_array_get(1, _)` to refer to two unrelated arrays now, because of the modulo. 

Similar thing in the next section where modulo is used for the index itself: if the two arrays can be of different lengths, then how can we reason about getting a random element from both.

tooling/ssa_fuzzer/fuzzer/src/base_context.rs
```diff
@@ -0,0 +1,374 @@
...
+            let brillig_len = self.brillig_ids.len();
+            acir_values_ids.push(elem % acir_len as u32);
+            brillig_values_ids.push(elem % brillig_len as u32);
+        }
+        let acir_len = acir_values_ids.len();
+        let brillig_len = brillig_values_ids.len();
+        let acir_array = self.acir_builder.insert_make_array(acir_values_ids);
+        let brillig_array = self.brillig_builder.insert_make_array(brillig_values_ids);
+        self.acir_arrays.push(Array { id: acir_array, length: acir_len as u32 });
+        self.brillig_arrays.push(Array { id: brillig_array, length: brillig_len as u32 });
```
aakoshh, line 150:
```suggestion
    pub(crate) fn insert_array(&mut self, elements: &[u32]) {
        let mut acir_values_ids = vec![];
        let mut brillig_values_ids = vec![];
        let acir_len = self.acir_ids.len();
        let brillig_len = self.brillig_ids.len();
        for elem in elements {
            acir_values_ids.push(elem % acir_len as u32);
            brillig_values_ids.push(elem % brillig_len as u32);
        }
        let acir_array = self.acir_builder.insert_make_array(acir_values_ids);
        let brillig_array = self.brillig_builder.insert_make_array(brillig_values_ids);
        self.acir_arrays.push(Array { id: acir_array, length: elements.len() as u32 });
        self.brillig_arrays.push(Array { id: brillig_array, length: elements.len() as u32 });
```
Most of the time we pass `&[_]` instead of `Vec<_>` unless you really want to pass ownership. You don't have to, just a bit more general as it works with arrays too.

tooling/ssa_fuzzer/fuzzer/src/base_context.rs
```diff
@@ -0,0 +1,374 @@
...
+        self.acir_ids.push(id_to_int(acir_return_id));
+        self.brillig_ids.push(id_to_int(brillig_return_id));
+    }
+
+    /// Sets an element in an array at the given index
+    pub(crate) fn insert_array_set(&mut self, array_idx: u32, index: u32, value: u32) {
+        if self.acir_arrays.is_empty() {
+            // no arrays created
+            return;
+        }
```
aakoshh, line 180: Perhaps this ties into my earlier question: how can you be sure that `self.brillig_arrays` is also empty?

tooling/ssa_fuzzer/fuzzer/src/base_context.rs
```diff
@@ -0,0 +1,374 @@
...
+        let brillig_arrays_len = self.brillig_arrays.len() as u32;
+        let acir_array = self.acir_arrays[(array_idx % acir_arrays_len) as usize];
+        let brillig_array = self.brillig_arrays[(array_idx % brillig_arrays_len) as usize];
+        let acir_array_id = acir_array.id;
+        let brillig_array_id = brillig_array.id;
+
+        let acir_id = self.acir_ids[(index % acir_array.length) as usize];
+        let brillig_id = self.brillig_ids[(index % brillig_array.length) as usize];
+        let acir_return_id = self.acir_builder.insert_array_get(acir_array_id, acir_id);
+        let brillig_return_id = self.brillig_builder.insert_array_get(brillig_array_id, brillig_id);
```
aakoshh, line 170: I'm trying to understand what the modulo operation do here. 

So we called `context.insert_array_get(array_idx, index)` with arbitrary `u32` values, like 12345 and 67890. We take do `array_idx % self.acir_arrays.len()` to pick an array that actually exists, say, `acir_arrays[2]`. Then, we look at the length of that array, and use `index % acir_array.length` to access `acir_ids`. Finally we use whatever is in that `acir_id` slot to actually issue a get against the array. 

I don't understand why we take a modulo of `acir_array.length` to access `acir_ids`. What do they have to do with each other? Say `acir_arrays[2]` has a length of 3; does it mean that `acir_ids[0..3]` are valid indexes for this particular array, but others are not?

When I look at `insert_array`, I see this:
```rust
  pub(crate) fn insert_array(&mut self, elements: Vec<u32>) {
        let mut acir_values_ids = vec![];
        for elem in elements {
            let acir_len = self.acir_ids.len();
            acir_values_ids.push(elem % acir_len as u32);
        }
        let acir_len = acir_values_ids.len();
        let acir_array = self.acir_builder.insert_make_array(acir_values_ids);
        self.acir_arrays.push(Array { id: acir_array, length: acir_len as u32 });
```
I interpret that as `elements` be a vector of random `u32` numbers, which we project down to the length of `acir_ids`, and then we create an array that consists of random variable IDs that already existed. 

Say we have 5 IDs in `acir_ids` and an `elements` consists of 100 `u32`, then by the end we should have 5 (or 6 with the new array?) values in `acir_ids` and 100 values on the `0..5` range in the array. Now, when we do `let acir_id = self.acir_ids[(index % acir_array.length) as usize];` we would be doing `acir_ids[67890 % 100]`, which is `acir_ids[90]`, which should cause a panic because the array only has 5 items.

tooling/ssa_fuzzer/fuzzer/src/base_context.rs
```diff
@@ -0,0 +1,374 @@
...
+    /// If we inserted some instructions, WitnessStack contains return value, so we return the last one
+    /// If we are in constant execution, we in witness stack only return value, so we return the Witness(0)
+    pub(crate) fn get_return_witnesses(&self) -> (Witness, Witness) {
+        if self.is_constant {
+            (Witness(0), Witness(0))
+        } else if self.acir_ids.len() as u32 != NUMBER_OF_VARIABLES_INITIAL {
+            (Witness(NUMBER_OF_VARIABLES_INITIAL), Witness(NUMBER_OF_VARIABLES_INITIAL))
+        } else {
+            (Witness(NUMBER_OF_VARIABLES_INITIAL - 1), Witness(NUMBER_OF_VARIABLES_INITIAL - 1))
+        }
```
aakoshh, line 364: Trying to understand this:
* When in constant mode, instead of adding `NUMBER_OF_VARIABLES_INITIAL` parameter witnesses, we insert them as constants into the circuit, which should allow constant folding to eliminate a lot of the code, and the witness #0 will be the return value. 
* When we have more (I would write `>= NUMBER_OF_VARIABLES_INITIAL`) variables than the initials, then we assume the one following the parameters is the return value. 
* If we have exactly `NUMBER_OF_VARIABLES_INITIAL` variables then it's just the inputs, with not even a return value, and we return the last input witness.
* It's not spelled out, but we won't have less than `NUMBER_OF_VARIABLES_INITIAL` variables in non-constant mode. 

What I don't fully understand is if `constant` mode will have a return value at 0, then why doesn't the non-constant case also have a return value at `NUMBER_OF_VARIABLES_INITIAL`. As I understand we always create `NUMBER_OF_VARIABLES_INITIAL` parameters but we might use less than that. Treating `NUMBER_OF_VARIABLES_INITIAL-1` as return value seems strange, as it's a) not a return value and b) might not even be used and c) trivially the same in both circuits, coming from the outside.

aakoshh (commented):
Thanks for following up on all my comments! 

Comparing with constant folding is a great idea 👏 
I sent some more questions about how things work.

tooling/ssa_fuzzer/fuzzer/src/base_context.rs
jewelofchaos9: It's my mistake, the index should be a number

tooling/ssa_fuzzer/fuzzer/src/base_context.rs
jewelofchaos9: You are right, the length of the arrays should always match

tooling/ssa_fuzzer/fuzzer/src/base_context.rs
jewelofchaos9: The only case where we return `NUMBER_OF_VARIABLES_INITIAL - 1` is when the fuzzer has not provided any valid instructions. 

Let's assume we have 3 variables in non-constant mode.

If no instructions were inserted, then we terminate the program with the last element of `ids`, which is equal to `Witness(2)`. After execution, the resulting witness will be `(Witness(0), Witness(1), Witness(2))`, we return `Witness(2)`.

If any number of instructions were inserted, then only one (return value) will be added to the final witness, which is (Witness(0), Witness(1), Witness(2), Witness(3)), we return Witness(3).

In constant mode, instead of variables, we insert constants to preserve the logic of interaction with the ids, so the resulting witness will have only one value (return value).

tooling/ssa_fuzzer/README.md
```diff
@@ -0,0 +1,33 @@
...
+This fuzzer generates random sequences of arithmetic and logical operations to craft SSA from scratch and than verify that both ACIR and Brillig implementations produce identical results. It helps catch potential bugs and inconsistencies between the two implementations.
+
+
+## Usage
+
+1. Setup environment:
+You need to delete `rust-toolchain.toml` from root of the project, because cargo-fuzz requires nightly compiler for address sanitizer.
+```
+rustup install nightly
+rustup default nightly
```
aakoshh, line 16: I found that this command works as well, so I don't have to default to `nightly`:

```
cargo +nightly fuzz ...
```

tooling/ssa_fuzzer/fuzzer/src/field.rs
```diff
@@ -0,0 +1,85 @@
...
+    for (i, witness_value) in data.initial_witness.iter().map(FieldElement::from).enumerate() {
+        witness_map.insert(Witness(i as u32), witness_value);
+        values.push(witness_value);
+    }
+    let initial_witness = witness_map;
+    log::debug!("instructions: {:?}", data.methods.clone());
+    log::debug!("initial_witness: {:?}", initial_witness);
+
+    let mut fuzzer = Fuzzer::new(type_.clone(), values);
+    for method in data.methods.clone() {
```
aakoshh, line 81:
```suggestion
    let initial_witness = witness_map;
    log::debug!("instructions: {:?}", data.methods);
    log::debug!("initial_witness: {:?}", initial_witness);

    let mut fuzzer = Fuzzer::new(type_, values);
    for method in data.methods {
```

```suggestion
    let mut fuzzer = Fuzzer::new(type_, values);
    for method in data.methods {
```
Printing shouldn't need cloning, and then both seem like the last time we use these variables, so we can move them.

tooling/ssa_fuzzer/src/builder.rs
```diff
@@ -0,0 +1,365 @@
...
+use noirc_evaluator::ssa::ir::types::{NumericType, Type};
+use noirc_evaluator::ssa::ir::value::Value;
+use noirc_frontend::monomorphization::ast::InlineType as FrontendInlineType;
+use std::panic::AssertUnwindSafe;
+use std::sync::Arc;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum FuzzerBuilderError {
+    #[error("Compilation panicked")]
```
aakoshh, line 21:
```suggestion
    #[error("Compilation panicked: {0}")]
```
At least I think we have to include the field to actually see the value.

tooling/ssa_fuzzer/src/builder.rs
aakoshh: It looks like you could have called `set_type`, although I still think it would be worth considering setting the type in the constructor, and verifying that the variables being inserted here are compatible, rather than allow changing the type.

tooling/ssa_fuzzer/src/compiler.rs
```diff
@@ -0,0 +1,227 @@
...
+    .finish();
+
+    // to triage ssa after optimizations, when crash found
+    let formatted_ssa = format!("{}", ssa);
+    log::debug!("formatted_ssa: {:?}", formatted_ssa);
+    match ssa.into_acir(&brillig, &BrilligOptions::default(), options.expression_width) {
+        Ok(artifacts) => Ok(ArtifactsAndWarnings(artifacts, vec![])),
+        Err(e) => Err(e),
+    }
+}
```
aakoshh, line 53: This looks like it could easily diverge from the `noirc_evaluator::ssa::optimize_into_acir` version. 

I see that in the beginning the difference is how the `SsaBuilder` is constructed, so maybe instead of this:

```rust
mod ssa {
  pub(crate) fn optimize_into_acir(
    program: Program,
    options: &SsaEvaluatorOptions,
  ) -> Result<ArtifactsAndWarnings, RuntimeError> {
    let builder = SsaBuilder::new(
        program,
        options.ssa_logging.clone(),
        options.print_codegen_timings,
        &options.emit_ssa,
    )?;
    let mut ssa = optimize_all(builder, options)?;
    ... // more steps on the SSA
  }
}

mod fuzzer {
  fn optimize_into_acir(
    builder: FunctionBuilder,
    options: SsaEvaluatorOptions,
  ) -> Result<ArtifactsAndWarnings, RuntimeError> {
    let builder = SsaBuilder {
        ssa: builder.finish(),
        ssa_logging: SsaLogging::None,
        print_codegen_timings: false,
    };
    let ssa = optimize_all(builder, &options)?;
    ... // same-ish steps on the SSA as above
  }
}
```

We could have a version of `optimize_into_acir` that takes an `SsaBuilder` and call that from both places:

```rust
mod ssa {

  pub fn optimize_ssa_builder_into_acir(
    builder: SsaBuilder,
    options: &SsaEvaluatorOptions,
  ) -> Result<ArtifactsAndWarnings, RuntimeError> {
    let mut ssa = optimize_all(builder, options)?;
    ... // more steps on the SSA
  }
  
  pub(crate) fn optimize_into_acir(
    program: Program,
    options: &SsaEvaluatorOptions,
  ) -> Result<ArtifactsAndWarnings, RuntimeError> {
    let builder = SsaBuilder::new(
        program,
        options.ssa_logging.clone(),
        options.print_codegen_timings,
        &options.emit_ssa,
    )?;
    optimize_ssa_builder_into_acir(builder, options)
}

mod fuzzer {
  fn optimize_function_builder_into_acir(
    builder: FunctionBuilder,
    options: SsaEvaluatorOptions,
  ) -> Result<ArtifactsAndWarnings, RuntimeError> {
    let builder = SsaBuilder {
        ssa: builder.finish(),
        ssa_logging: SsaLogging::None,
        print_codegen_timings: false,
    };
    ssa::optimize_ssa_builder_into_acir(builder)
  }
}
```

jewelofchaos9:
1) Added support for types, now the instruction returns a specific type along with the ID.  Store different types in separated vectors. 
2) Changed fuzzer logic. Now it produces different type of witnesses, instruction now has index + variable type (to choose Id from correct vector)
3) Combined all fuzz targets into big one. 
4) Deleted all arrays from fuzzing (todo)

Rumata888 (approved):
Skimmed it, should be ok to merge

tooling/ssa_fuzzer/fuzzer/src/base_context.rs
```diff
@@ -0,0 +1,353 @@
...
+
+    /// Finalizes the function by setting the return value
+    pub(crate) fn finalize_function(&mut self) {
+        match (self.last_value_acir.clone(), self.last_value_brillig.clone()) {
+            (Some(acir_result), Some(brillig_result)) => {
+                self.acir_builder.finalize_function(acir_result);
+                self.brillig_builder.finalize_function(brillig_result);
+            }
+            _ => {
+                // If no last value was set, use the first value from the first type in each map
```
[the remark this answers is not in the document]
jewelofchaos9: This case is only possible if no instructions were inserted, so the return value will always be one of inputs

PR #7660, opened 2025-03-11
feat(ssa): Basic control dependent LICM

### LINKED ISSUE #7569: Hoist instructions from loops in non control dependent blocks
### Problem

Certain instructions are safe to hoist out of a loop if they are dominated by the same predicate (e.g. such checked arithmetic operations). 
For example this code:
```noir
fn main(x: u32, y: u32) {
    for _ in 0..4 {
        let mut z = x * y;
        z = z * x;
        assert_eq(z, 12);
    }
}
```
Will result in the following SSA:
```
  b0(v0: u32, v1: u32):
    jmp b1(u32 0)
  b1(v2: u32):
    v5 = lt v2, u32 4
    jmpif v5 then: b3, else: b2
  b2():
    return
  b3():
    v6 = mul v0, v1
    v7 = mul v6, v0
    constrain v7 == u32 12
    v10 = unchecked_add v2, u32 1
    jmp b1(v10)
}
```
However, we should be able to hoist `v6 = mul v0, v1` out of the loop. This is restricted as the current loop invariant code does not account for whether the loop invariant is under the same predicate as the pre-header block. This PR https://github.com/noir-lang/noir/pull/7396 restricted hoisting these kinds of operations. 

### Happy Case

I should be able to hoist operations that only rely upon a predicate out of a loop when the loop's pre-header has the same predicate.

We currently the predicate for constant folding with ACIR: https://github.com/noir-lang/noir/blob/c2022f4fbab4ed5d9cd62b3ff5b9bba9bd36627e/compiler/noirc_evaluator/src/ssa/ir/instruction.rs#L595

These instructions have the same effects in Brillig and we should be able to track their predicates for hoisting out of loops. 
For example even though we want to hoist `let mut z = x * y;` out of the loop above. If it were under a predicate, it should not be hoisted:
```noir
fn main(x: u32, y: u32) {
    for _ in 0..4 {
        if x == 5 {
            let mut z = x * y;
            assert_eq(z, 12);
        }
    }
}
```
However, if `x == 5` captured the entire loop, we should hoist it. 

### Workaround

Yes

### Workaround Description

Hoisting manually. However, for operations generated by the compiler this unclear for developers and very unfriendly. 

### Additional Context

_No response_

### Project Impact

Blocker

### Blocker Context

I labelled this a blocker, simply due to the critical nature of optimizing our unconstrained execution.

Yes

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #7569 

Builds upon https://github.com/noir-lang/noir/pull/7595 (post-dominator tree) and https://github.com/noir-lang/noir/pull/7692 (dominance frontiers)

This PR resolves the optimizations described in the issue. However, the optimization can be expanded by further accounting for predicates and possibly even hoisting instructions under predicates as part of loop peeling/splitting. I leave these tasks to be captured in separate issues.

## Summary\*

We have certain instructions that we mark as unsafe for hoisting when they are only dependent upon a predicate and do not have any other side effects (e.g. a checked arithmetic operation). You can look into the linked issue for an example of logic that will not be hoisted. 

If a predicated instruction shares the same predicate as the entire loop we can now hoist this instruction. The new unit tests under `loop_invariant.rs` provide examples of when a predicated instruction can now be hoisted. 

We check for whether an instruction shares a predicate with the entire loop through control dependence analysis. If the block containing an instruction is control dependent on any blocks between the pre-header and the instruction's block we know that there has been some branching. We then mark the current block as being control dependent. 

If a block is not control dependent, we mark instructions that are reliant upon predicates (but do not have other side effects) as safe for hoisting. 

We determine whether a block is control dependent on a parent block by following Definition 3 in [Ferrante et al., 1987, The program dependence graph and its use in optimization](https://dl.acm.org/doi/pdf/10.1145/24039.24041).
```text
Definition 3. Let G be a control flow graph. Let X and Y be nodes in G. Y is
control dependent on X iff
(1) there exists a directed path P from X to Y with any 2 in P (excluding X
and Y) post-dominated by Y and
(2) X is not post-dominated by Y. 
```

Builds upon https://github.com/noir-lang/noir/pull/7595 to utilize the post-dominator tree. We can further optimize our control dependence check by looking at the post-dominator frontiers https://github.com/noir-lang/noir/pull/7692.

Verifying the above conditions for every loop block would be quite inefficient.
For example, let's say we just want to check whether a given loop block is control dependent at all
after the loop preheader. We would have to to verify the conditions above for every block between the loop preheader
and the given loop block. This is n^2 complexity in the worst case.

To optimize the control dependence checks, we can use post-dominance frontiers (PDF).
From Cooper, Keith D. et al. “A Simple, Fast Dominance Algorithm.” (1999).
```text
A dominance frontier is the set of all CFG nodes, y, such that
b dominates a predecessor of y but does not strictly dominate y.
```
Reversing this for post-dominance we can see that the conditions for control dependence
are the same as those for post-dominance frontiers.
Thus, we rewrite our control dependence condition as Y is control dependent on X iff Y is in PDF(Y).
We then can store the PDFs for every block as part of the context of this pass, and use it for checking control dependence.
Using PDFs gets us from a worst case n^2 complexity to a worst case n.

#### Hoisting vs. Deduplicating

This PR also adds a new method `can_be_hoisted` on `Instruction`. This is essentially the same as `can_be_deduplicated` except it accounts for that certain instructions are not safe for hoisting as lone instructions. The main difference is that we do not hoist `Constrain`, `ConstrainNotEqual`, `RangeCheck`, and functions with `Purity::PureWithPredicate` with predicates.  

The reason for this is that we may have dynamic loop bounds or a equal lower and upper loop bounds (e.g. a loop that will never execute). If we were to hoist the instructions above they could potentially cause a program to fail when it should not. When deduplicating this is allowed as we have a matching instruction, but for hoisting this can break the program logic.

## Additional Context

As mentioned in the PR description this work builds upon https://github.com/noir-lang/noir/pull/7595 and https://github.com/noir-lang/noir/pull/7692. You may want to review those PRs first as to have the full context. This is also outlined in the new doc comments on the loop invariant pass. 

We can probably further differentiate `Purity::PureWithPredicate` to differentiate between instructions pure with predicates for hoisting and pure with predicates for deduplication. I leave this for follow-up work. 

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -125,6 +181,17 @@ impl<'f> LoopInvariantContext<'f> {
         self.set_values_defined_in_loop(loop_);
 
         for block in loop_.blocks.iter() {
+            // The CFG will ultimately be entirely flattened in ACIR.
```
vezenovm, line 184:
```suggestion
            self.current_block_control_dependent = false;
            // The CFG will ultimately be entirely flattened in ACIR.
```

compiler/noirc_evaluator/src/ssa/ir/cfg.rs
```diff
@@ -89,10 +89,6 @@ impl ControlFlowGraph {
     /// Add a directed edge making `from` a predecessor of `to`.
     fn add_edge(&mut self, from: BasicBlockId, to: BasicBlockId) {
         let predecessor_node = self.data.entry(from).or_default();
-        assert!(
```
vezenovm, line 92: Now that we have reversed CFGs if a node has many predecessors, you can have multiple successors upon reversing the CFG. If we want to keep this assert I could potentially add a reversed status flag to the CFG.

vezenovm:
> `AztecProtocol_aztec-packages_noir-projects_noir-protocol-circuits_crates_rollup-lib`	`223` s	`158` s	`1.41`

This seems flakey as the following PR activating this optimization on ACIR functions shows [no regression](https://github.com/noir-lang/noir/pull/7700#pullrequestreview-2683121606).

compiler/noirc_evaluator/src/ssa/ir/cfg.rs
michaeljklein: Yes, I would prefer to keep this assert somehow

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -176,6 +270,10 @@ impl<'f> LoopInvariantContext<'f> {
         // set the new current induction variable.
         self.current_induction_variables.clear();
         self.set_induction_var_bounds(loop_, true);
+        // The previous loop may have set that the current block is control dependent.
+        // If we fail to reset for the next loop, a block may be misadvertently labelled
```
michaeljklein, line 274: nit, but avoids cspell for `misadvertently`:

```suggestion
        // If we fail to reset for the next loop, a block may be inadvertently labelled
```

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -900,32 +1009,44 @@ mod test {
 
     #[test]
     fn do_not_hoist_unsafe_div() {
-        // This test is similar to `nested_loop_invariant_code_motion`, the operation
-        // in question we are trying to hoist is `v9 = div i32 10, v0`.
-        // Check that the lower bound of the outer loop it checked and that we not
+        // This test is similar to `nested_loop_invariant_code_motion`, except that
+        // the loop logic is under a dynamic predicate.
+        // Divisions are only reliant upon predicates and do not have other side effects.
```
michaeljklein, line 1014:
```suggestion
        // Divisions are only reliant upon predicates and do not have other side effects.
        //
```

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -900,32 +1009,44 @@ mod test {
...
+        // If an unsafe division occurs in a loop block that is not control dependent,
+        // we can still safely hoist that division as that instruction is always going to be hit.
+        // Thus, we place the unsafe division under a predicate to ensure that we are testing
+        // division hoisting based upon loop bounds and nothing else.
+        //
+        // The operation in question we are trying to hoist is `v12 = div u32 10, v1`.
+        // Check that the lower bound of the outer loop is checked and that we do not
```
michaeljklein, line 1021: `Check that the lower bound of the outer loop is checked` check for which check on the lower bound of the outer loop?

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -969,29 +1095,202 @@ mod test {
...
+          b3():
+            return
+        }
+        ";
+
+        assert_normalized_ssa_equals(ssa, expected);
+    }
+
+    #[test]
+    fn non_control_dependent_loop_follows_control_depenent_loop() {
```
michaeljklein, line 1208: typo:
```suggestion
    fn non_control_dependent_loop_follows_control_dependent_loop() {
```

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -969,29 +1095,202 @@ mod test {
...
+        let ssa = Ssa::from_str(src).unwrap();
+
+        let ssa = ssa.loop_invariant_code_motion();
+        assert_normalized_ssa_equals(ssa, src);
+    }
+
+    #[test]
+    fn hoist_safe_mul_that_is_non_control_dependent() {
+        let src = "
+        brillig(inline) fn main f0 {
```
michaeljklein, line 1167: FYI it should be possible to use "regular" identifiers for these tests after https://github.com/noir-lang/noir/pull/7697

michaeljklein (changes requested):

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: I switched to regular identifiers

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: I later state "that we do not hoist an operation that can potentially error with a division by zero". I see how this comment is roundabout though, I have changed to 
```
        // Check whether the lower bound of the outer loop is zero and that we do not
        // hoist an operation that can potentially error with a division by zero.
```

compiler/noirc_evaluator/src/ssa/ir/cfg.rs
vezenovm: I added a `reversed` flag to keep this assert

vezenovm:
The tests in https://github.com/noir-lang/noir/pull/7705 actually flagged a bug with how we were specifying instructions safe for hoisting. I have added a new method `can_be_hoisted` on `Instruction` which I go over in the `Hoisting vs. Deduplicating` section of the PR description as well as additional logic and tests to the loop invariant pass.

In summary, we do not want to hoist certain instructions (which may be safe for deduplication) when a loop may not execute.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -969,26 +1110,350 @@ mod test {
...
+          exit():
+            return
+        }
+        ";
+
+        assert_normalized_ssa_equals(ssa, expected);
+    }
+
+    #[test]
+    fn do_not_hoist_constrain_in_loop_with_equal_non_zero_loop_bounds() {
```
vezenovm, line 1363: Not sure if we want to keep this test and `do_not_hoist_constrain_in_loop_with_zero_upper_bound`. With this test I wanted to guarantee we were not just checking that the upper bound is zero but that (upper - lower) is not zero. However, it makes the above test feel a bit unnecessary. Let me know if we would prefer both tests or just this one here.

vezenovm:
The execution trace reports look to be bugged out. `loop_invariant_regression` with an inliner aggressiveness of 0 is giving me 379 opcodes as expected, but the report is saying that there is no difference.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
michaeljklein: I think it's worth keeping both

PR #7684, opened 2025-03-12
chore(docs): Minor fixes on local documentation development workflows

Savio-Sou:
## Summary\*

- Supplement missing step of generating `versions.json` in `yarn dev`
- Fix missing cache cleaning in `yarn clean`
- Trim unnecessary commands and supplement missing commands in README
- Improve readability of steps in README

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "preprocess": "yarn workspace @noir-lang/acvm_js build && ./scripts/codegen_nargo_reference.sh && yarn node ./scripts/preprocess/index.js",
-    "dev": "yarn preprocess && ENV=dev docusaurus start",
+    "dev": "yarn preprocess && yarn version::stables && ENV=dev docusaurus start",
     "build": "yarn preprocess && docusaurus build",
     "clean": "rm -rf ./processed-docs ./processed-docs ./build",
     "version::stables": "ts-node ./scripts/setStable.ts",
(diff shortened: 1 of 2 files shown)


docs/package.json
```diff
@@ -4,9 +4,9 @@
   "private": true,
   "scripts": {
     "preprocess": "yarn workspace @noir-lang/acvm_js build && ./scripts/codegen_nargo_reference.sh && yarn node ./scripts/preprocess/index.js",
-    "dev": "yarn preprocess && ENV=dev docusaurus start",
+    "dev": "yarn preprocess && yarn version::stables && ENV=dev docusaurus start",
```
TomAFrench, line 7: I'd like to avoid running `yarn version::stables` as part of `yarn dev` as it's a fairly slow step which doesn't benefit from incremental compilation. We should update the docs to mention that this should be run before `yarn dev`.

Savio-Sou:
@TomAFrench do we use the https://github.com/noir-lang/noir/blob/6eaea18225305903570ec1139b99196b28217663/docs/scripts/cut_version.sh anywhere?

Running `yarn workspace docs version` as originally documented doesn't trigger the `version` command, but spits out the following error instead:

```
Unknown Syntax Error: Not enough positional arguments.

$ yarn version [-d,--deferred] [-i,--immediate] <strategy>
```

Thinking if we should simply remove the `version` command from _package.json_.

TomAFrench:
We use it as part of releases so it shouldn't be removed.

Savio-Sou:
Cool, no additional changes from me in that case; thanks!

TomAFrench (approved):

PR #7694, opened 2025-03-13
fix: wrong printing of line comment in quoted

### LINKED ISSUE #7693: Wrong printing of line comment in `quote { ... }`
### Aim

This code:

```noir
fn main() {
    comptime {
        let q = quote {
            // Hello world
            1
        };
        println(q);
    }
}
```

### Expected Behavior

Should print this at compile time:

```
quote {
    // Hello world
    1
}
```

### Bug

It prints this instead:

```
quote {
    // Hello world1
}
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #7693

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/display.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/display.rs
@@ -285,7 +285,6 @@ impl<'interner> TokenPrettyPrinter<'interner> {
             | Token::RawStr(..)
             | Token::FmtStr(..)
             | Token::Whitespace(_)
-            | Token::LineComment(..)
             | Token::BlockComment(..)
             | Token::AttributeStart { .. }
             | Token::Invalid(_) => {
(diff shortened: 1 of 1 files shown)


TomAFrench (approved):
Can you add a quick regression test for this?

asterite:
Right now there are no tests for this printing but I'll try to add some tomorrow.

asterite:
Ah, now I remember... the quoted values are printed with `println!` or `eprintln!`, that's the only way we can see those values. Or, well, we could produce a `Context`, then type-check some code, somehow find the tokens and print them.

For now I'll merge this. I've been thinking that one thing we could do is to set the output expectation for programs in `test_programs/execution_success`, not just for this but in general to prove that `println` works well. I'll try to do it as a follow-up PR.

PR #7696, opened 2025-03-13
chore(artifact_cli): Print circuit output to stdout

vezenovm:
# Description

## Problem\*

Resolves an issue raised by aztec team wanting to extract the output from stdout

## Summary\*

In https://github.com/noir-lang/noir/pull/7406 we removed printing the return value of a program to stdout. 

This simply brings it back.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/artifact_cli/src/execution.rs
+++ b/tooling/artifact_cli/src/execution.rs
@@ -89,7 +89,7 @@ pub fn save_and_check_witness(
     if let Some(witness_dir) = witness_dir {
         save_witness(&results.witness_stack, circuit_name, witness_dir, witness_name)?;
     }
-    check_witness(circuit, results.return_values)
+    check_witness(circuit, results.return_values, circuit_name)
 }
 
 /// Save the witness stack to a file.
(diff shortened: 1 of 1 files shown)


tooling/artifact_cli/src/execution.rs
```diff
@@ -127,6 +128,7 @@ pub fn check_witness(
                 if actual != expected {
                     return Err(CliError::UnexpectedReturn { expected, actual: Some(actual) });
                 }
+                println!("[{}] Circuit output: {actual:?}", circuit_name);
```
aakoshh, line 131: It looks like this will only be printed if the `return` is set in the input `Prover.toml` file, ie. if `expected_return` is [not empty](https://github.com/noir-lang/noir/blob/9116b271b4569c575b5d8d90dacfee0ca9a4c097/tooling/artifact_cli/src/execution.rs#L122).

I'd do it at the same level as all the other `if`s in this method, or potentially outside of it in `save_and_check_witness`.

vezenovm: ah yes, I was going to quickly. I will move it

aakoshh (approved):
Thanks, sorry I missed this when I moved the code 🙏

vezenovm:
> Thanks, sorry I missed this when I moved the code 🙏

No worries 😃 !

PR #7699, opened 2025-03-13
chore(frontend): Regression test for creating a mutable reference to an array element

### LINKED ISSUE #7698: Footgun when dealing with mutable references to arrays
This is an issue reported by Roman on the Taceo team. Consider the program below:

```noir
fn foo(x: &mut u32) {
    *x += 1;
}
fn main() {
    let mut state: [u32; 4] = [1, 2, 3, 4];
    foo(&mut state[0]);
    // state[0] += 1; // Works
    assert_eq(state[0], 2); // expect:2 got:1
}
```

It was expected that passing `&mut state[0]` into `foo` will  cause a mutation to the `state` array, however what seems to happen is that we perform a read of the value at `state[0]` and then pass a mutable reference to this new value. We then will just discard this value and the assertion acts on the initial `state` array.

vezenovm:
# Description

## Problem\*

Resolves #7698 

## Summary\*

This issue was resolved by https://github.com/noir-lang/noir/pull/7488. I have included an additional regression test for the snippet within the issue.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/.github/workflows/rustdoc.yml
+++ b/.github/workflows/rustdoc.yml
@@ -1,0 +1,78 @@
+name: Publish rustdoc
+
+on:
+  push:
+    branches:
+      - master
+
+# This will cancel previous runs when a branch or PR is updated
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Build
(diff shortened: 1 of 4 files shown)


vezenovm:
Looks like our docs do not note this. Will update the docs.

michaeljklein (approved):

PR #7708, opened 2025-03-14
feat: use relative paths in "Call stack" and "Witness saved to"

asterite:
# Description

## Problem

No issue, just something I wanted to do for some time now.

## Summary

For example, an output before this PR looked like this:

```
error: Failed constraint
  ┌─ /Users/asterite/Sandbox/one/src/main.nr:9:16
  │
9 │         assert(array[index - 1].counter < 3);
  │                ----------------------------
  │
  = Call stack:
    1. /Users/asterite/Sandbox/one/src/main.nr:9:16
 ```

Now it's:

```
error: Failed constraint
  ┌─ src/main.nr:9:16
  │
9 │         assert(array[index - 1].counter < 3);
  │                ----------------------------
  │
  = Call stack:
    1. src/main.nr:9:16
```

Similarly, if the program executes fine it would show:

```
[one] Circuit witness successfully solved
[one] Witness saved to /Users/asterite/Sandbox/one/target/one.gz
```

Now it shows:

```
[one] Circuit witness successfully solved
[one] Witness saved to target/one.gz
```

## Additional Context

There's a bit of code duplication here but it's just a couple of lines. We could extract these to a helper function but I don't know where we'd put them. There's also a Go proverb I like: ["A little copying is better than a little dependency"](https://go-proverbs.github.io/)

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/nargo_cli/src/cli/debug_cmd.rs
+++ b/tooling/nargo_cli/src/cli/debug_cmd.rs
@@ -214,8 +214,15 @@ fn run_async(
             }
 
             if let Some(witness_name) = witness_name {
-                let witness_path =
+                let mut witness_path =
                     save_witness_to_dir(&solved_witness_stack, witness_name, target_dir)?;
+
+                // See if we can make the file name a bit shorter/easier to read if it starts with the current directory
+                if let Ok(current_dir) = std::env::current_dir() {
+                    if let Ok(name_without_prefix) = witness_path.strip_prefix(current_dir) {
+                        witness_path = name_without_prefix.to_path_buf();
+                    }
+                }
 
                 println!("[{}] Witness saved to {}", package.name, witness_path.display());
             }
--- a/tooling/nargo_cli/src/cli/debug_cmd.rs
+++ b/tooling/nargo_cli/src/cli/debug_cmd.rs
@@ -217,7 +217,7 @@ fn run_async(
                 let mut witness_path =
                     save_witness_to_dir(&solved_witness_stack, witness_name, target_dir)?;
 
-                // See if we can make the file name a bit shorter/easier to read if it starts with the current directory
+                // See if we can make the file path a bit shorter/easier to read if it starts with the current directory
                 if let Ok(current_dir) = std::env::current_dir() {
                     if let Ok(name_without_prefix) = witness_path.strip_prefix(current_dir) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: tooling/nargo_cli/src/cli/debug_cmd.rs [hunk 1]
```diff
@@ -214,9 +214,16 @@ fn run_async(
             }
 
             if let Some(witness_name) = witness_name {
-                let witness_path =
+                let mut witness_path =
                     save_witness_to_dir(&solved_witness_stack, witness_name, target_dir)?;
 
+                // See if we can make the file name a bit shorter/easier to read if it starts with the current directory
```
michaeljklein, line 220:
```suggestion
                // See if we can make the file path a bit shorter/easier to read if it starts with the current directory
```

tooling/noirc_artifacts/src/debug.rs
```diff
@@ -177,7 +177,18 @@ impl<'a> Files<'a> for DebugArtifact {
     type Source = &'a str;
 
     fn name(&self, file_id: Self::FileId) -> Result<Self::Name, Error> {
-        self.file_map.get(&file_id).ok_or(Error::FileMissing).map(|file| file.path.clone().into())
+        let name = self.file_map.get(&file_id).ok_or(Error::FileMissing);
+        let name: Self::Name = name.map(|file| file.path.clone().into())?;
+
+        // See if we can make the file name a bit shorter/easier to read if it starts with the current directory
```
michaeljklein, line 183:
```suggestion
        // See if we can make the file path a bit shorter/easier to read if it starts with the current directory
```

michaeljklein (approved):
LGTM w/ a non-blocking nit that "file name" often refers to just the "base name" + file extension part of a path

aakoshh:
I think the full path in the `Witness saved to /Users/asterite/Sandbox/one/target/one.gz` example was actually helpful, that it tells you where the target is: is it a workspace level target directory or one under the package?

asterite:
> is it a workspace level target directory or one under the package?

I guess, if you are inside a package and it's save at the workspace level, it would still show the full path. It'll only show the path as relative if it's inside the current directory. That said, I started doing this PR for "Call stack" but also noticed it there, but we could leave the "Witness saved to" to a follow-up PR to discuss it a bit more.

PR #7713, opened 2025-03-14
chore: check test program execution success output

asterite:
# Description

## Problem

Addresses https://github.com/noir-lang/noir/pull/7694#pullrequestreview-2683016696

## Summary

Now when running "execution_success" tests we check that the output they produce matches an expected output. The rule is:
- if there is an expected output, it goes into "stdout.txt" inside the directory
- if there's no expected output there shouldn't be an "stdout.txt" (or there should be an empty one, it's the same)

This allows us to test comptime printing, but also would have prevented the regression where the output value wasn't printed anymore.

## Additional Context

When checking for non-determinism (which we do when running these tests) the program was compiled again. If it had comptime println the output would show up duplicated so I changed it to disable comptime output in that second compilation, while also adding a hidden compiler option to disable comptime output (this will be useful for `nargo expand` too).

## Documentation

Check one:

# PR Checklist


aakoshh:
Asking here in case it's useful for somebody else: I created a new test, and when I run it it compiles that it expects no output, but we do have one because the circuit returns a value:

```
---- tests::execution_success::test_regression_7964::forcebrillig_false_inliner_0_expects stdout ----
stdout does not match expected output. Expected:


Actual:
[regression_7964] Circuit output: Field(1)

thread 'tests::execution_success::test_regression_7964::forcebrillig_false_inliner_0_expects' panicked at tooling/nargo_cli/tests/execute.rs:135:21:
assertion `left == right` failed
  left: "[regression_7964] Circuit output: Field(1)"
 right: ""

```

Is there some automated way for the `stdout.txt` to be created if it doesn't exist, or should I create it by copying that string into a file?

TomAFrench:
Run `cargo test` with the env var `OVERWRITE_TEST_OUTPUT=1`

aakoshh:
Thanks @TomAFrench 🙏 

Maybe the output could mention this variable in the output, as @guipublic just ran across this as well, and none of us knew there was this env var. My first thought was that if `stdout.txt` is missing it could have just created it, but then again if it's not the final version, then the next failure will be equally, if not more, confusing.

asterite:
> My first thought was that if stdout.txt is missing it could have just created it

This is how it works for `stderr.txt` in `compile_failure`: because we always expect a failure, if the file isn't there it's automatically generated. For `stdout.txt` I didn't do this because that file is optional (not all programs produce an output). That said, it probably won't matter once we move to insta.

PR #7723, opened 2025-03-14
feat: add `std::warn::warn`

### LINKED ISSUE #7714: Allow producing warnings at compile-time
### Problem

Right now one can use `assert`, `assert_eq` and `panic` (which uses `assert` under the hood) to produce errors at compile-time. However, there's no equivalent to produce warnings.

### Happy Case

There's a way to produce warnings at compile-time.

There could be this function in the stdlib:

```noir
#[builtin(warn)]
pub comptime fn warn<T, let N: u32>(message: fmtstr<N, T>) {}
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #7714

## Summary

There's some special code here to produce an error if `warn` is used at runtime. Eventually we could control this with another attribute, maybe, though I don't know if we'll have many functions that can only be called at compile-time...

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/patterns.rs
+++ b/compiler/noirc_frontend/src/elaborator/patterns.rs
@@ -20,6 +20,7 @@ use crate::{
     node_interner::{
         DefinitionId, DefinitionInfo, DefinitionKind, ExprId, FuncId, GlobalId, TraitImplKind,
     },
+    token::FunctionAttribute,
 };
 
 use super::{Elaborator, ResolverMeta, path_resolution::PathResolutionItem};
@@ -572,6 +573,8 @@ impl Elaborator<'_> {
             }
         }
 
+        self.check_call_to_warn_builtin_at_runtime(&definition_kind, location);
+
         let id = self.interner.push_expr(HirExpression::Ident(expr.clone(), generics.clone()));
 
         self.interner.push_expr_location(id, location);
@@ -594,6 +597,28 @@ impl Elaborator<'_> {
             }
         } else {
             (id, typ)
+        }
+    }
+
+    // It's an error to call to the built-in function "warn" if we are not in a comptime context
+    fn check_call_to_warn_builtin_at_runtime(
+        &mut self,
+        definition_kind: &Option<DefinitionKind>,
+        location: Location,
+    ) {
+        let Some(DefinitionKind::Function(func_id)) = definition_kind else {
+            return;
+        };
+
+        if self.in_comptime_context() {
+            return;
+        }
+
+        let attribute = self.interner.function_attributes(func_id).function();
+        if let Some(FunctionAttribute::Builtin(builtin)) = attribute {
+            if builtin == "warn" {
+                self.push_err(ResolverError::WarnUsedAtRuntime { location });
+            }
         }
     }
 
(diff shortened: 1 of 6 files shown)


compiler/noirc_frontend/src/elaborator/patterns.rs
```diff
@@ -597,6 +600,28 @@ impl Elaborator<'_> {
         }
     }
 
+    // It's an error to call to the built-in function "warn" if we are not in a comptime context
+    fn check_call_to_warn_builtin_at_runtime(
```
TomAFrench, line 604: Is this not covered by how we disallow calling comptime functions from a runtime context?

asterite: Ah, you are right. However, that check happens during monomorphization (not sure why!) and before that it panics when it can't find the built-in in SSA generation. That said, I guess I'll introduce this built-in in SSA that does nothing to let monomorphization produce this error.

asterite: Hmm... but monomorphization happens before SSA, I don't know why it goes to SSA with errors 🤔

asterite: Aah... it's because built-in functions don't go through the monomorphizer's queue.

Would it be okay to move this error from monomorphization to where I put it? (generalize it to all comptime functions, not just `warn`).

TomAFrench: I'm stopping for the day so we can discuss next week. You can workaround this with a non-builtin wrapper function which is comptime - this will ensure that the error is thrown appropriately.

asterite: Good idea! And that explains why we have many wrappers like that.

But... it didn't work. The reason is that the warning is then produced in the std crate, and we don't show warnings if they aren't from the crate being compiled.

TomAFrench: Tbh that means that this feature isn't going to work for Nico's usage anyway

asterite: Good point! I'll close this for now.

TomAFrench: @asterite we link macro errors back to the location which invokes the macro now. Seems like this means that these warnings would be visible now?

asterite:
I'll try this in a new PR.

compiler/noirc_frontend/src/elaborator/patterns.rs
asterite: I just tried this PR again.

If we have the definition like this:

```noir
pub comptime fn warn<T, let N: u32>(message: fmtstr<N, T>) {
    _warn(message)
}

#[builtin(warn)]
comptime fn _warn<T, let N: u32>(message: fmtstr<N, T>) {}
```

then we get the warning location in the `_warn` call, which is in the stdlib. Then the warning doesn't show up.

However, I just remembered that the interpreter keeps the call stack, so if the warn built-in assumes it's going to be wrapped then using the parent location works fine. Should I open a PR for this?

asterite: Ah, but that won't work either. If a library has a macro that calls `warn` then those warnings will never be visible to callers of that library.

PR #7729, opened 2025-03-17
chore: removed redundant `.clone()` in `insert_expr` function

0xstratus:
# Description  
In the `insert_expr` function, I removed the unnecessary `.clone()` call on `coefficients` since `coefficients` is already an iterator. This change simplifies the code and avoids redundant cloning.

## Additional Context  
This change optimizes performance slightly by removing an unnecessary operation in the code. No logic was altered.

## Documentation  
Check one:

# PR Checklist

- [ ] I have formatted the changes with [[Prettier](https://prettier.io/)](https://prettier.io/) and/or `cargo fmt` on default settings.


jfecher (changes requested):
`coefficients` is used afterward so either the `.clone()` or a `.iter().cloned()` is necessary here

Issue #7686, opened 2025-03-13
Noir wasm compiler does not resolve deps correctly


ioedeveloper:
I tested "@noir-lang/noir_wasm": "1.0.0-beta.3" with recent changes to the stealth drop example in this PR https://github.com/noir-lang/noir-examples/pull/42, and had the error below.

<img width="601" alt="Image" src="https://github.com/user-attachments/assets/d56413cd-df49-45fb-9612-42324d7cce53" />

TomAFrench:
This is likely due to @signorecello not using fixed dependencies so https://github.com/signorecello/zk-nullifier-sig is pulling in a different version of bignum than expected.

signorecello:
Yeah I should pin that version. In any case this works with nargo compile, just not with noir_wasm according to @ioedeveloper

TomAFrench:
@signorecello fyi if you're creating a set of libraries which don't pin their dependencies to particular commits then we're going to deprioritise any debugging of issues which involve them as they're built on sand.

ioedeveloper:
I tested using the steathdrop example from `zkpedro/plume` branch, using the same config for Nargo.toml
```
[package]
name = "stealthdrop"
type = "bin"
authors = [""]

[dependencies]
trees = { git = "https://github.com/privacy-scaling-explorations/zk-kit.noir", tag = "main", directory = "packages/merkle-trees" }
ecrecover = { tag = "main", git = "https://github.com/colinnielsen/ecrecover-noir" }
plume = { tag = "main", git = "https://github.com/signorecello/zk-nullifier-sig", directory = "circuits/noir" }
noir_bigcurve = { tag = "v0.6.0", git = "https://github.com/noir-lang/noir_bigcurve" }

```

signorecello:
got it @TomAFrench in this case I told @ioedeveloper to open this issue because he told me it works with nargo and not noir_wasm so it doesn't seem like a versioning issue

signorecello:
just to be sure, pinned both versions to v0.6.0 on [zk-kit](https://github.com/privacy-scaling-explorations/zk-kit.noir) and [plume](https://github.com/signorecello/zk-nullifier-sig)

TomAFrench:
@ioedeveloper do you have an example of how you've set up `noir_wasm`? Ideally this would a repository that I can clone and run to experience the issue you're running into.

TomAFrench:
[...]

PR #7736, opened 2025-03-18
chore: push users towards nargo in tutorial

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

For some reason we're pushing people towards `noir_wasm` over `nargo` which just makes it more likely they're going to run into issues. e.g. #7686

We can avoid these issues by having people just use the CLI which is best practice anyway.

## Documentation\*

Check one:

# PR Checklist\*


docs/docs/tutorials/noirjs_app.md
```diff
@@ -20,17 +20,23 @@ Before we start, we want to make sure we have Node installed. For convenience (a
 curl -fsSL https://bun.sh/install | bash
 ```
 
+We'll also need version 1.0.0-beta.2 nargo installed:
```
vezenovm, line 23: nit: could just link to the quick start installation here

vezenovm (approved):

saleel:
I saw someone in Discord having a problem with https://github.com/noir-lang/tiny-noirjs-app which also uses the wasm to compile. I will do a PR to change this.
We can probably do a separate page for browser compilation is that is useful (like for folks building Remix plugin (I am just guessing they are compiling in the browser)), but this is a bit too complex for newcomers IMO.

Also, we know at-least two teams running into weird wasm problems with `bun` ([this](https://github.com/AztecProtocol/aztec-packages/issues/11113) and AnonCast), so I don't think we should use `bun` anywhere until its more mature.

CC: @signorecello

TomAFrench:
I agree on stripping out usage of bun in docs. Whole it may be nice on the happy path we should push people towards well tested flows and allow people to experiment with bun themselves.

signorecello:
Agree with removing bun if it causes problems but noir_wasm was meant to allow for an full e2e browser solution (ex. for Remix which needs to compile in the browser and can't use nargo compile).

Do we have an idea of what could be causing problems with noir_wasm?

signorecello:
@TomAFrench can we have a look at this? I understand noir_wasm sounds like "two ways of doing the same thing" but we will lose the Remix IDE plugin if we don't fix it. Same for the hardhat plugin for which Oleh has a grant for people to move to wasm.

Seems like it is broken on beta3, but removing will only make it worse because of all the projects that depend on it

TomAFrench:
> can we have a look at this?

Sure but I need reproduction code in order to do this. e.g. for https://github.com/noir-lang/noir/issues/7686 it would be great to have a minimal repository like tiny-noirjs-app which is trying to compile the circuit which is exhibiting issues and then we can look at it.

> Same for the hardhat plugin for which Oleh has a grant for people to move to wasm.

Why are we making a push for people on the commandline to not use our commandline tools?

Savio-Sou:
> Seems like it is broken on beta3

v1.0.0-beta.3 shouldn't have broken any existing usage with `noir_wasm`, we have simply discovered https://github.com/noir-lang/noir/issues/7823 recently.

That said, if we come across any additional bugs with `noir_wasm`, do create Issues for them.

PR #7753, opened 2025-03-19
Convert frontend tests to compilation tests

michaeljklein:
# Description

## Problem\*

Frontend tests are run in an environment that could become out of sync with `nargo compile`.

## Summary\*

During `get_program_with_options`, we can:
- Generate a crate name by hashing the source
- Use the source as the contents of `main.nr`
- Generate a simple `Nargo.toml`

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_driver/src/abi_gen.rs
+++ b/compiler/noirc_driver/src/abi_gen.rs
@@ -9,7 +9,7 @@ use noirc_abi::{
 use noirc_errors::Location;
 use noirc_evaluator::ErrorType;
 use noirc_frontend::TypeBinding;
-use noirc_frontend::ast::{Signedness, Visibility};
+use noirc_frontend::shared::{Signedness, Visibility};
 use noirc_frontend::{
     hir::Context,
     hir_def::{
--- a/compiler/noirc_evaluator/src/ssa.rs
+++ b/compiler/noirc_evaluator/src/ssa.rs
@@ -31,7 +31,7 @@ use acvm::{
 use ir::instruction::ErrorType;
 use noirc_errors::debug_info::{DebugFunctions, DebugInfo, DebugTypes, DebugVariables};
 
-use noirc_frontend::ast::Visibility;
+use noirc_frontend::shared::Visibility;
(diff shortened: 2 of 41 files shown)


jfecher:
How many of the errors caught by this PR would also be caught by adding monomorphizing to the `get_program` test function? It seems like a lot of machinery to add macros for updating all these generated files. I think if we can't get the existing test structure working with changes like now monomorphizing them, then I'd prefer us to just commit to moving these tests to dedicated test_programs instead of housing them in both locations & having CI check the source matches the generated version.

jfecher:
Discussed in standup: we're going to keep the current approach since it has things we need now like more testing for the formatter and future support for the fuzzer. We may revisit it in the future but for now we'll keep the CI check as well to ensure the generated tests are up to date.

jfecher:
(waiting for tests to pass until I review)

jfecher (approved):
👍

TomAFrench:
Looping back to this. @michaeljklein Do you have examples of where you were getting different behaviour between the two testing methods?

michaeljklein:
The main examples are the skipped tests [here](https://github.com/noir-lang/noir/pull/7753/files#diff-71ef4ebd1447a72f32396063dc17cc15fcbafdc1933dc9b07f57fcc30484b36eR159):
- https://github.com/noir-lang/noir/issues/7763
- https://github.com/noir-lang/noir/issues/7783
- https://github.com/noir-lang/noir/issues/7766

TomAFrench:
> #7763

This strikes me as expected (minus the panic that we should turn into an error). Here we're testing whether the frontend will catch this test or not. The fact that cases which we call out that we don't catch early fail elsewhere in the compiler doesn't invalidate the test in my mind.

michaeljklein:
@TomAFrench 

> The fact that cases which we call out that we don't catch early fail elsewhere in the compiler doesn't invalidate the test in my mind.

I'm not sure that it would "invalidate" the test, but I think that it might be good to at least add a comment r.e. the expected behavior of the test, if that makes sense?

Issue #5922, opened 2024-09-04
Debugger enters both the "then" and the "else" statement of a single "if"


nickysn:
### Aim

Executing a noir program one statement at a time inside the debugger.

### Expected Behavior

When stepping through an "if" conditional, it is expected that after evaluating the boolean condition, the debugger should step either inside the "then" statement, or inside the "else" statement, but not go through both.

### Bug

In certain cases, the debugger steps inside the "then" statement first, then on the next step it goes inside the "else" statement. In reality, only the "else" statement should be executed (because the boolean condition of the "if" is false) and the debugger should not enter the "then" statement at all.

### To Reproduce

1. Extract the Noir project that demonstrates the bug:
[if_then_else_reduced.tar.gz](https://github.com/user-attachments/files/16873108/if_then_else_reduced.tar.gz)
2. start "nargo debug" inside the project directory
3. type "next" 9 times until the "if" condition is highlighted:
![Screenshot from 2024-09-04 18-23-58](https://github.com/user-attachments/assets/d2c36b8a-273d-4fe9-ae55-3c5879f51cbc)
4. type "next" one more time. The debugger enters the "then" statement (which is incorrect):
![Screenshot from 2024-09-04 18-24-50](https://github.com/user-attachments/assets/944e9dea-2579-4760-9bbd-afdc8cff074d)
5. type "next" one more time. The debugger jumps inside the "else" statement:
![Screenshot from 2024-09-04 18-25-36](https://github.com/user-attachments/assets/f9a59123-bfd9-4095-9d1b-b40d142e80ed)

### Workaround

None

### Project Impact

Nice-to-have

### Would you like to submit a PR for this Issue?

None

Issue #7496, opened 2025-02-24
Debugger does not see changes, made to var values, done via mutable references


nickysn:
### Aim

Debugging a program and being able to examine local variables via the `vars` command.

### Expected Behavior

The debugger should never print wrong variable values.

### Bug

When a variable value is changed indirectly, via using a mutable reference type `&mut T`, the debugger should still see the new value.

### To Reproduce

1. `nargo new mutref_debugging`
2. Edit `main.rs`:
```
fn set_to_zero(x: &mut Field) {
    *x = 0;
}

fn main() {
    let mut y = 42;
    set_to_zero(&mut y);
    assert(y == 0);
}
```
3. `nargo debug`
4. Type `next`, until the debugger reaches the last line of the main function (line 9, this is the line after the assert). With the current git version of Noir, this took stepping 6 times with `next`.
5. Type `vars` to examine the local variables. You get:
```
main()
  y:Field = 0x2a
```
However, the correct value of `y` should be 0, not 0x2a (42).

### Workaround

None

### Project Impact

Blocker

### Would you like to submit a PR for this Issue?

None

PR #7758, opened 2025-03-20
feat(debugger): CodeTracer time-travelling debugger support

nickysn:
# Description
This adds support for the CodeTracer time-travelling debugger:

https://github.com/metacraft-labs/codetracer

It also includes several debugger fixes.

## Problem\*

1) Adds support for CodeTracer (`nargo trace` command).

2) Resolves:
https://github.com/noir-lang/noir/issues/5922

3) Resolves other bugs in the debugger (`nargo debug`)

## Summary\*

This adds a new command `nargo trace`, which produces an execution trace, compatible with CodeTracer.

It also includes several debugger fixes.

## Additional Context

This is extracted from the Blocksense Noir branch at:

https://github.com/blocksense-network/noir

By cherry picking only the commits, related to `nargo trace` and the debugger.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -322,11 +396,15 @@ pub fn create_program(
...
-    let error_types = error_types
+    let mut error_types: BTreeMap<_, _> = error_types
         .into_iter()
         .map(|(selector, hir_type)| (selector, ErrorType::Dynamic(hir_type)))
         .collect();
 
+    for acir in &mut generated_acirs {
+        error_types.append(&mut acir.error_types);
+    }
+
```
TomAFrench, line 407: Interesting, looks like this just a bugfix being bundled in? We should be able to address this down in lower level more cleanly.

Do you remember the situation where you ran into issues with this? It would be good if we can create a regression test for this.

testing/format_trace/README.md
TomAFrench, line 1: Can you expand on how you expect this to be used? It's not expected to be user-facing correct?

I'm considering whether this could be replaced with usage of `jq`.

tooling/nargo/src/foreign_calls/print.rs
```diff
@@ -8,12 +8,24 @@ use noirc_printable_type::{PrintableType, PrintableValueDisplay};
...
 use super::{ForeignCall, ForeignCallError, ForeignCallExecutor};
 
-#[derive(Debug, Default)]
+#[derive(Default)]
 pub enum PrintOutput<'a> {
     #[default]
     None,
     Stdout,
     String(&'a mut String),
+    PrintCallback(Box<dyn Fn(String) + 'a>),
```
TomAFrench, line 17: I've been warning to remove this enum and just pass in a writer for any output to be sent to for some time. If we're adding a callback here I think we've reached the point of where we should address this.

compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -53,10 +53,19 @@ pub enum SsaLogging {
     Contains(String),
 }
 
+#[derive(Debug, Clone)]
+pub enum OptimizationLevel {
+    Debug, // optimizations that don't change the control-flow
+    All,   // all available optimizations
+}
```
TomAFrench, line 60: Nice, we've been wanting to add debug levels for some time. #2128

compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -109,7 +118,10 @@ pub(crate) fn optimize_into_acir(
         &options.emit_ssa,
     )?;
 
-    let mut ssa = optimize_all(builder, options)?;
+    let mut ssa = match options.optimization_level {
+        OptimizationLevel::Debug => optimize_for_debug(builder, options),
+        OptimizationLevel::All => optimize_all(builder, options),
+    }?;
```
TomAFrench, line 124: I would like to avoid having two completely separate compilation pipelines here however as it's a recipe for inconsistencies popping up down the line.

tooling/nargo_cli/src/cli/trace_cmd.rs
```diff
@@ -0,0 +1,124 @@
...
+        args.compile_options.clone(),
+    )?;
+
+    trace_program_and_decode(
+        compiled_program,
+        package,
+        &args.prover_name,
+        &args.trace_dir,
+        args.compile_options.pedantic_solving,
+    )
```
TomAFrench, line 76: Have you considered having a separate tool which accepts a nargo build artifact as an input? Currently every usage of `nargo trace` would require a recompile, etc.

Separating this out would make it easier for you to compile a program once and then debug with multiple inputs within codetracer without the overhead of recompilation.

tooling/tracer/src/tracer_glue.rs
```diff
@@ -0,0 +1,202 @@
...
+                    runtime_tracing::TypeKind::Seq,
+                    &format!("{type_name_base}<{length}, ..>"),
+                ); // TODO: more precise?
+                ValueRecord::Sequence { elements: element_values, type_id }
+            } else {
+                panic!("type-value mismatch: value: {:?} does not match type Bool", value)
+            }
+        }
+        _ => {
+            // TODO(stanm): cover all types and remove `warn!`.
```
TomAFrench, line 148: Unresolved TODO

tooling/tracer/src/tracer_glue.rs
```diff
@@ -0,0 +1,202 @@
...
+    tracer.register_call(file_id, args);
+}
+
+/// Extracts the relevant information from the given `frame` to construct a vector of `ArgRecord`
+/// that the `Tracer` interface expects when registering function calls.
+fn convert_params_to_args_vec(tracer: &mut Tracer, frame: &StackFrame) -> Vec<FullValueRecord> {
+    let mut result = Vec::new();
+    for param_index in &frame.function_param_indexes {
+        let variable = &frame.variables[*param_index];
+        // TODO(stanm): maybe don't duplicate values?
```
TomAFrench, line 174: Unresolved TODO

TomAFrench (commented):
Hey, thank you for opening this PR and the fixes to the debugger!

One bit of feedback that may be controversial is that I think we should explore `noir-tracer` as a separate tool rather than having a `nargo trace` command. This would result in a more focused tool and avoid unnecessary recompilations when switching inputs.

Cargo.toml
```diff
@@ -93,11 +95,15 @@ nargo_fmt = { path = "tooling/nargo_fmt" }
 nargo_toml = { path = "tooling/nargo_toml" }
 noir_lsp = { path = "tooling/lsp" }
 noir_debugger = { path = "tooling/debugger" }
+noir_tracer = { path = "tooling/tracer" }
 noirc_abi = { path = "tooling/noirc_abi" }
 noirc_artifacts = { path = "tooling/noirc_artifacts" }
 noirc_artifacts_info = { path = "tooling/noirc_artifacts_info" }
 noir_artifact_cli = { path = "tooling/artifact_cli" }
 noir_protobuf = { path = "utils/protobuf" }
+bb_abstraction_leaks = { path = "tooling/bb_abstraction_leaks" }
```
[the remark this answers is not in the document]
nickysn: Removed: https://github.com/noir-lang/noir/pull/7758/commits/25cd930a72592bd3d4e8aa21927805be857bfb9d

Cargo.toml
```diff
@@ -93,11 +95,15 @@ nargo_fmt = { path = "tooling/nargo_fmt" }
...
+acvm_cli = { path = "tooling/acvm_cli" }
```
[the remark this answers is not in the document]
nickysn: Removed: https://github.com/noir-lang/noir/pull/7758/commits/c740c4263818d8c77373297e3b6b8d5ab153ec45

compiler/noirc_evaluator/src/ssa.rs
TomAFrench: This is being resolved in https://github.com/noir-lang/noir/pull/7911

tooling/tracer/src/tracer_glue.rs
nickysn: Added handlers for all types and removed this TODO. Some types (Enum) throw a todo!(), because they're still unstable/experimental and don't yet work in the debugger.

testing/format_trace/README.md
nickysn: It is useful for making tests for `nargo trace` and for debugging the `nargo trace` feature. Therefore it's useful to Noir compiler developers, and not so much for regular users. It used to be a separate tool, but we decided to integrate it into nargo. But now that you say this, I think it doesn't make much sense for this to be a separate command. It's probably best if it's kept as a separate tool, or it can be integrated into `nargo trace`. So, I'm thinking of two options:

1) We can easily revert this change and make it a separate tool again.
2) We can instead make it a suboption to `nargo trace`, e.g. `nargo trace --pretty-print-json`.

What do you think? Which one is better?

compiler/noirc_evaluator/src/ssa.rs
nickysn: We have a regression test for this (`trace/assert_message`), but it uses the `nargo trace` feature, so the test can't be separated, even though the bug fix can. I can separate the bug fix into a separate pull request, but I don't know how to make a test for it, without `nargo trace`. Any ideas how to do it?

nickysn:
> One bit of feedback that may be controversial is that I think we should explore `noir-tracer` as a separate tool rather than having a `nargo trace` command. This would result in a more focused tool and avoid unnecessary recompilations when switching inputs.

We can do this in the future, but I think this requires a refactoring of the debugger, because right now the `nargo trace` feature internally uses the same debugger that `nargo debug` uses, and this is currently integrated inside `nargo` and requires a recompilation. Moreover, the debugger injects some code in the front end at the AST level, in order to track variable values. There are many flaws to this approach, including bugs like: https://github.com/noir-lang/noir/issues/7496

Overall, I'm in favor of changing the way the debugger tracks variables, but that would be a very huge change. It requires rewriting the entire debugger, so it doesn't modify the compiled code, but instead adds debug information as a separate entity that describes the addresses of the local variables in memory. So, I think it's beyond the scope of this pull request and can be done later in a separate pull request. At least, that's how I see it, unless I'm missing something. Is there a different approach?

tooling/nargo_cli/src/cli/trace_cmd.rs
nickysn: See my comment in the general thread below. I imagine this would require a complete rewrite of the debugger, but I could be missing something.

tooling/tracer/src/tracer_glue.rs
nickysn: I don't believe there's anything needed to be done for this TODO, so I removed it in: https://github.com/noir-lang/noir/pull/7758/commits/39d31a1a9c91ca76c51508e67bccdbacbe08a331

tooling/tracer/src/tracer_glue.rs
nickysn: However, most types (all that were stable) were implemented, and I also added tests for them.

compiler/noirc_evaluator/src/ssa.rs
nickysn: I unified them in: https://github.com/noir-lang/noir/pull/7758/commits/43817f8705136a8c6aee49c984dd36e8856e250c

testing/format_trace/README.md
Aristotelis2002: We removed the `nargo format-trace` feature from `nargo` all together. Now it's a part of a separate project.

PR #7762, opened 2025-03-20
fix(ci): Fail the CI job on a Brillig report failure

vezenovm:
# Description

## Problem\*

We have had a couple situations where the Brillig reports fail silently. This is because we have accounted for some special case in our `nargo_cli` tests such as experimental features such as enums or different inliner settings, but our Brillig reports do not account for that special case. For example here is a PR trying to mitigate that (https://github.com/noir-lang/noir/pull/7761).

## Summary\*

I have added logic to fail our CI job for Brillig reports if the script fails rather than failing silently and just making it look as though there is no diff in the reports. 

This required adding `set -o pipefail` to the gate reports scripts themselves and `set -e` in the CI job,.

I expect this PR to fail until https://github.com/noir-lang/noir/pull/7761 is merged in master.

## Additional Context

We should add this for all reports including circuit sizes as well.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/.github/workflows/reports.yml
+++ b/.github/workflows/reports.yml
@@ -117,19 +117,19 @@ jobs:
       - name: Generate Brillig bytecode size report
         working-directory: ./test_programs
         run: |
-          # Exit immediately on error
-          set -e
-
           mkdir ./reports
 
-          ./gates_report_brillig.sh 9223372036854775807
-          jq '.programs |= map(.package_name |= (. + "_inliner_max"))' gates_report_brillig.json > ./reports/gates_report_brillig_inliner_max.json
-
-          ./gates_report_brillig.sh 0
-          jq '.programs |= map(.package_name |= (. + "_inliner_zero"))' gates_report_brillig.json > ./reports/gates_report_brillig_inliner_zero.json
-
-          ./gates_report_brillig.sh -9223372036854775808
-          jq '.programs |= map(.package_name |= (. + "_inliner_min"))' gates_report_brillig.json > ./reports/gates_report_brillig_inliner_min.json
+          # Map the inliner aggressiveness to its test name
+          declare -A inliner_settings=(
+            [9223372036854775807]="inliner_max"
+            [0]="inliner_zero"
+            [-9223372036854775808]="inliner_min"
+          )
+          for inliner_agg in "${!inliner_settings[@]}"; do
+            test_name="${inliner_settings[$inliner_agg]}"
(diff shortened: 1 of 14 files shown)


vezenovm:
This job now fails as I expect https://github.com/noir-lang/noir/actions/runs/13973066908/job/39119915348?pr=7762

.github/workflows/reports.yml
```diff
@@ -119,14 +119,18 @@ jobs:
...
+          # Map the inliner aggressiveness to its test name
+          declare -A inliner_settings=(
+            [9223372036854775807]="inliner_max"
+            [0]="inliner_zero"
+            [-9223372036854775808]="inliner_min"
+          )
+
+          for inliner_agg in "${!inliner_settings[@]}"; do
+            test_name="${inliner_settings[$inliner_agg]}"
+            ./gates_report_brillig.sh "$inliner_agg" || echo "Script exited early on $test_name" && exit 1
```
vezenovm, line 131:
```suggestion
            ./gates_report_brillig.sh "$inliner_agg" || echo "Job exited early on $test_name" && exit 1
```

.github/workflows/reports.yml
```diff
@@ -169,14 +173,18 @@ jobs:
...
+          # Map the inliner aggressiveness to its test name
+          declare -A inliner_settings=(
+            [9223372036854775807]="inliner_max"
+            [0]="inliner_zero"
+            [-9223372036854775808]="inliner_min"
+          )
+
+          for inliner_agg in "${!inliner_settings[@]}"; do
+            test_name="${inliner_settings[$inliner_agg]}"
+            ./gates_report_brillig_execution.sh "$inliner_agg" || echo "Script exited early on $test_name" && exit 1
```
vezenovm, line 185:
```suggestion
            ./gates_report_brillig_execution.sh "$inliner_agg" || echo "Job exited early on $test_name" && exit 1
```

test_programs/gates_report_brillig_execution.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -e
+set -o pipefail
```
TomAFrench, line 3:
```suggestion
set -eo pipefail
```

test_programs/gates_report_brillig.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -e
+set -o pipefail
```
TomAFrench, line 3:
```suggestion
set -eo pipefail
```

TomAFrench (approved):
why say many words when few words do trick.

PR #7769, opened 2025-03-20
fix(debugger): improve debugger_expected_call_stack test

anaPerezGhiglia:
# Description
The test was not reflecting the actual behavior of the debugger, and thus it was not reliable for future changes in it.

## Summary\*
The test was sending several debug commands, and then trying to read the produced lines all together. 
This was faulty since it seems that the internal read-buffer of the `dbg_session` was not big enough to retain all the produced lines.
Actually, the test was sending 16 commands but only asserting 13 of them..

### Changes
Change the test to assert each command output one by one.
Also add the missing outputs of the first three commands

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
Thank you, looks great!

PR #7775, opened 2025-03-21
chore(docs): updating branding

signorecello:
Changing the branding to the new one


Savio-Sou:
Looks like the "Ask Noir AI" button is still on the old logo:
image

Can we update it in one-go?

signorecello:
Ready for review @Savio-Sou I was forced to have some fun with JS but it worked. Also made some other changes to the consent modal Tiffany asked for

signorecello:
TBH this is not at all complete, I'm just patching CSS spaguetti but we should change the entire palette, I'm talking with Tiffany to get that done at once

Savio-Sou (approved):
Looking great!

PR #7781, opened 2025-03-24
chore: Make embedded curve opcodes return embedded curve points

sirasistant:
# Description

## Problem\*

Embedded curve opcodes returned 3 fields on noir land, but on barretenberg side is_infinite is a boolean. This mismatch is problematic for the AVM.

## Summary\*

I've changed ecadd and msm to return an array of 1 EmbeddedCurvePoint. I can't return an EmbeddedCurvePoint directly since the AVM doesn't handle proving opcodes with more than one return register. By making it an array, we just have one pointer in the opcode, which also makes it smaller. This quirk is not exposed since the opcode is wrapped in the stdlib. No change should happen for users.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/acir/mod.rs
+++ b/compiler/noirc_evaluator/src/acir/mod.rs
@@ -2188,7 +2188,7 @@ impl<'a> Context<'a> {
                 let inputs = vecmap(&arguments_no_slice_len, |arg| self.convert_value(*arg, dfg));
 
                 let output_count = result_ids.iter().fold(0usize, |sum, result_id| {
-                    sum + dfg.try_get_array_length(*result_id).unwrap_or(1) as usize
+                    sum + dfg.type_of_value(*result_id).flattened_size() as usize
                 });
 
                 let vars = self.acir_context.black_box_function(black_box, inputs, output_count)?;
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs
@@ -289,6 +289,10 @@ fn embedded_curve_add(
     let (p1x, p1y, p1inf) = get_embedded_curve_point(point1)?;
     let (p2x, p2y, p2inf) = get_embedded_curve_point(point2)?;
 
+    let (x, y, inf) = Bn254BlackBoxSolver(pedantic_solving)
+        .ec_add(&p1x, &p1y, &p1inf.into(), &p2x, &p2y, &p2inf.into())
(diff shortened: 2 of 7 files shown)


compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -2188,7 +2188,7 @@ impl<'a> Context<'a> {
                 let inputs = vecmap(&arguments_no_slice_len, |arg| self.convert_value(*arg, dfg));
 
                 let output_count = result_ids.iter().fold(0usize, |sum, result_id| {
-                    sum + dfg.try_get_array_length(*result_id).unwrap_or(1) as usize
+                    sum + dfg.type_of_value(*result_id).flattened_size() as usize
```
sirasistant, line 2191: We weren't considering arrays of tuples here. It was thinking that the output_count for [EmbeddedCurvePoint;1] was 1, when it should be 3, since it maps the output_count to witnesses

compiler/noirc_frontend/src/hir/comptime/interpreter/foreign.rs
```diff
@@ -283,22 +289,40 @@ fn embedded_curve_add(
     let (p1x, p1y, p1inf) = get_embedded_curve_point(point1)?;
     let (p2x, p2y, p2inf) = get_embedded_curve_point(point2)?;
 
+    let embedded_curve_point_typ = match &return_type {
+        Type::Array(_, item_type) => item_type.as_ref().clone(),
+        _ => {
+            return Err(InterpreterError::TypeMismatch {
```
sirasistant, line 295: It's so verbose when I need to fetch a type from the return type. Maybe there is a better way?

TomAFrench: I'd be inclined to pull the `EmbeddedCurvePoint` type out of one of the arguments and then use that to construct the return type.

TomAFrench (commented):
Looks sensible to me. Do we have a PR up in aztec-packages showing the benefits in the AVM.

:face_with_diagonal_mouth: at brillig regressions.

sirasistant:
> Do we have a PR up in aztec-packages showing the benefits in the AVM
I have a PR up to see that it doesn't break https://github.com/AztecProtocol/aztec-packages/pull/12935
The AVM won't benefit much in a sense of bytecode size of opcodes executed, it's more than the AVM will type is_infinite with type u1 and the previous signature expected a type field, so it can potentially crash in runtime based on the usage of the output

Will address now the comments!

TomAFrench (approved):

PR #7786, opened 2025-03-24
fix: mark array index as used in DIE pass when replacing read with explicit bounds check

### LINKED ISSUE #7785: Panic thrown in zk-regex test `capture::composite::from_addr::test_invalid_1`
This test in the https://github.com/Mach-34/noir-zk-regex repo is causing a panic which I can reproduce on master.

https://github.com/Mach-34/noir-zk-regex/blob/20de93bd8b30bb51271690c7d13fad86399527e1/packages/noir/src/capture/composite/from_addr.nr#L180-L186

Running this test causes a panic with the error

```
$ nargo compile
Message:  internal error: entered unreachable code: ICE: Should have been in cache v12741 Instruction { instruction: Id(13474), position: 0, typ: Numeric(NativeField) }
Location: compiler/noirc_evaluator/src/acir/mod.rs:1955
```

TomAFrench:
# Description

## Problem\*

Resolves #7785 

## Summary\*

Explanation of the bug is in the regression test. tl;dr is that if we have a chain of array_get instructions which each one uses the output of the last as its index, and then the final array_get is unused, then we end up marking all of the array_gets as unused.

## Documentation\*

Check one:

# PR Checklist\*


jfecher (commented):
Looks like some unrelated failures in CI

PR #7799, opened 2025-03-25
fix: improve misleading error message when object type is unknown

### LINKED ISSUE #7798: Misleading error message when object type is unknown in method call
### Aim

The error message given for this code is misleading.

```noir
fn foo<let N: u32>() -> [Field; N] {
  let mut bar = std::mem::zeroed();
  for i in 0..bar.len() { .... } 
  bar
}
```

### Expected Behavior

Currently the error message is:

```
error: No method named 'len' found for type '[Field; N]'
  ┌─ src/main.nr:3:15
  │
3 │   for i in 0..bar.len() { 
  │     
```

which is misleading because `[Field; N]` does have a `len` method.

### Bug

Instead, it should say that `bar`'s type is unknown at that point, like in Rust.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

Originally reported by @nventuro 

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #7798

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/tests.rs
+++ b/compiler/noirc_frontend/src/tests.rs
@@ -4188,9 +4188,8 @@ fn mutable_reference_to_array_element_as_func_arg() {
 fn object_type_must_be_known_in_method_call() {
     let src = r#"
     pub fn foo<let N: u32>() -> [Field; N] {
-        let mut bar = _;
-                      ^ in expressions, `_` can only be used on the left-hand side of an assignment
-                      ~ `_` not allowed here
+        let array = [];
+        let mut bar = array[0];
         let _ = bar.len();
                 ^^^ Object type is unknown in method call
                 ~~~ Type must be known by this point to know which method to call
--- a/compiler/noirc_frontend/src/tests.rs
+++ b/compiler/noirc_frontend/src/tests.rs
@@ -4183,3 +4183,21 @@ fn mutable_reference_to_array_element_as_func_arg() {
     "#;
     check_errors(src);
 }
+
+#[test]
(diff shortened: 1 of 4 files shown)


michaeljklein (approved):

compiler/noirc_frontend/src/tests.rs
```diff
@@ -4183,3 +4183,21 @@ fn mutable_reference_to_array_element_as_func_arg() {
...
     check_errors(src);
 }
+
+#[test]
+fn object_type_must_be_known_in_method_call() {
+    let src = r#"
+    pub fn foo<let N: u32>() -> [Field; N] {
+        let mut bar = _;
+                      ^ in expressions, `_` can only be used on the left-hand side of an assignment
+                      ~ `_` not allowed here
```
jfecher, line 4193: Does this really reproduce the original error? I'm not sure if bar's type would be an unbound typevar here or the error type. To guarantee it is unbound you could do something a bit roundabout:
```rs
let array = [];
let bar = array[0]; // this would panic at runtime but we don't run the program
```

jfecher (approved):

compiler/noirc_frontend/src/tests.rs
asterite: Nice, thanks! I was looking of a way to get an untyped value but couldn't figure out how. I checked that with `_` the error message is the same, but it just adds noise to the test.

PR #7801, opened 2025-03-25
fix: Reserve memory slots for debugging at the start of the global space

vezenovm:
# Description

## Problem\*

Bug from https://github.com/noir-lang/noir/pull/7789

rollup-block-merge: Total arrays copied: 42812 (seems to be a bug here, the count starts at 41911 for this test for some reason meaning the real count would be 901)

## Summary\*

Keeping track of memory slots can become muddled 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
+++ b/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
@@ -106,6 +106,17 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
         used_globals: &HashSet<ValueId>,
         hoisted_global_constants: &BTreeSet<(FieldElement, NumericType)>,
     ) -> HashMap<(FieldElement, NumericType), BrilligVariable> {
+        // Using the end of the global memory space adds more complexity as we
+        // have to account for possible register de-allocations as part of regular global compilation.
+        // Thus, we want to allocate any reserved global slots first.
+        //
+        // If this flag is set, compile the array copy counter as a global
+        if self.brillig_context.count_array_copies() {
+            let new_variable = allocate_value_with_type(self.brillig_context, Type::unsigned(32));
+            self.brillig_context
+                .const_instruction(new_variable.extract_single_addr(), FieldElement::zero());
+        }
+
         for (id, value) in globals.values_iter() {
             if !used_globals.contains(&id) {
                 continue;
@@ -134,13 +145,6 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
             }
         }
 
-        // If this flag is set, also compile the array copy counter as a global
-        if self.brillig_context.count_array_copies() {
-            let new_variable = allocate_value_with_type(self.brillig_context, Type::unsigned(32));
-            self.brillig_context
-                .const_instruction(new_variable.extract_single_addr(), FieldElement::zero());
-        }
-
         new_hoisted_constants
     }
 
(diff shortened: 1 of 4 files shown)


compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_globals.rs
```diff
@@ -235,6 +235,8 @@ pub(crate) type BrilligGlobalsArtifact = (
     HashMap<(FieldElement, NumericType), BrilligVariable>,
 );
 
+// pub(crate) type GlobalsCompi
+
```
vezenovm, line 239:
```suggestion
```

compiler/noirc_evaluator/src/brillig/brillig_ir/entry_point.rs
```diff
@@ -23,16 +23,11 @@ impl<F: AcirField + DebugToString> BrilligContext<F, Stack> {
...
         globals_init: bool,
-        mut globals_memory_size: usize,
+        globals_memory_size: usize,
         options: &BrilligOptions,
     ) -> BrilligArtifact<F> {
         let mut context = BrilligContext::new(options);
 
-        if options.enable_array_copy_counter {
-            // Reserve space for the array clone counter
-            globals_memory_size += 1;
```
vezenovm, line 33: We never needed this as we are allocating during `compile_globals`. As this is in `new_entry_point_artifact` (which uses a fresh context) as well, this logic would simply offset where we start the calldata offset by an additional slot that was never allocated. 

Something like this would have been needed if we never allocated for the debug slot, but we are and I think it is cleaner that way.

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -106,6 +106,17 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
         used_globals: &HashSet<ValueId>,
         hoisted_global_constants: &BTreeSet<(FieldElement, NumericType)>,
     ) -> HashMap<(FieldElement, NumericType), BrilligVariable> {
+        // Using the end of the global memory space adds more complexity as we
+        // have to account for possible register de-allocations as part of regular global compilation.
+        // Thus, we want to allocate any reserved global slots first.
```
jfecher, line 111: Does this mean we may deallocate globals? If not, what kinds of registers would we be deallocating?

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -164,7 +168,6 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
 
         if self.brillig_context.count_array_copies()
             && matches!(terminator_instruction, TerminatorInstruction::Return { .. })
-            && self.brillig_context.get_globals_memory_size().is_some()
```
jfecher, line 167: This flag should still be useful - it makes it so that we only emit this printout at the end of the current artifact, instead of at the end of every function called.

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
vezenovm: Certain intermediate registers are deallocated during codegen (specifically for making an array) and can be re-used. Here is an example of where we [deallocate](https://github.com/noir-lang/noir/blob/d18a1dd814b1b71032b429a56d1366e154efb898/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs#L1005). This mattered more when we used to have a hardcoded limit for the global space. The global space follows the other `RegisterAllocator` implementations for deallocate register. Another solution could have been to change the `deallocate_register` method for `impl RegisterAllocator for GlobalSpace` do nothing. I still find moving this to the front simpler as we now can use `GlobalSpace::start()` for reserved debug registers and do not have to thread the global memory size to procedures.

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
vezenovm: Ah yes I see

PR #7804, opened 2025-03-25
fix!: require indexing arrays and slices with u32

### LINKED ISSUE #7795: Automatically casting numeric types for array indices is a footgun.
I recently found out that we were automatically casting any numeric type into a `u32` while looking into #7785 which seemed really counterintuitive and footgunny to me. 

@jewelofchaos9 found this independently in https://github.com/noir-lang/noir/issues/7776#issuecomment-2749243766 along with a way to exploit this to get a difference in ACIR and brillig due to it.

```noir
fn main(x: u64, y: pub u64, index: u64) -> pub u64 {
  let t: [u64; 2] = [x, y];
  t[index]
}
```

This compiles down to the SSA

```
  b0(v0: u64, v1: u64, v2: u64):
    v3 = make_array [v0, v1] : [u64; 2]
    v4 = cast v2 as u32
    v5 = array_get v3, index v4 -> u64
    return v5
}
```

This results in a difference in behaviour between ACIR and brillig as casting in brillig implies a truncation whereas this is not the case in ACIR. You can then read from the array with an index which points way past the end of the array as long as `v4 % u32::MAX <= 1`

asterite:
# Description

## Problem

Resolves #7795

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/expressions.rs
+++ b/compiler/noirc_frontend/src/elaborator/expressions.rs
@@ -8,10 +8,10 @@ use crate::{
     ast::{
         ArrayLiteral, AsTraitPath, BinaryOpKind, BlockExpression, CallExpression, CastExpression,
         ConstrainExpression, ConstrainKind, ConstructorExpression, Expression, ExpressionKind,
-        Ident, IfExpression, IndexExpression, InfixExpression, ItemVisibility, Lambda, Literal,
-        MatchExpression, MemberAccessExpression, MethodCallExpression, Path, PathSegment,
-        PrefixExpression, StatementKind, TraitBound, UnaryOp, UnresolvedTraitConstraint,
-        UnresolvedTypeData, UnresolvedTypeExpression, UnsafeExpression,
+        Ident, IfExpression, IndexExpression, InfixExpression, IntegerBitSize, ItemVisibility,
+        Lambda, Literal, MatchExpression, MemberAccessExpression, MethodCallExpression, Path,
+        PathSegment, PrefixExpression, StatementKind, TraitBound, UnaryOp,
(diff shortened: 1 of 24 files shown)


asterite:
@noir-lang/core Seeing these regressions, should I proceed forward? I don't understand why circuit sizes increases so much

(I'll try to run the gates report on my machine but first I need a newer version of `bb` and it seems there are no binaries for mac so I'll need to compile them from source, but `./boostrap.sh` doesn't work for mac so I'll work on that first)

vezenovm:
> @noir-lang/core Seeing these regressions, should I proceed forward? I don't understand why circuit sizes increase so much

These tests most likely starting to using range gates that they weren't previously. Ultra uses lookup tables to prove cheap range constraints, but these tables have a base-line cost and people more useful in larger circuits. 

> **brillig_nested_arrays**	+2 ❌	**+50.00%**	+2,746 ❌	**+16152.94%**
> **brillig_uninitialized_arrays**	+2 ❌	**+40.00%**	+2,746 ❌	**+10170.37%**
> **nested_array_dynamic_simple**	+2 ❌	**+12.50%**	+2,746 ❌	**+5604.08%**

vezenovm:
> (I'll try to run the gates report on my machine but first I need a newer version of `bb` and it seems there are no binaries for mac so I'll need to compile them from source, but `./boostrap.sh` doesn't work for mac so I'll work on that first)

Also I think 0.82.2 has mac binaries

TomAFrench:
Looks like we're going to need a deprecation message :face_exhaling:

asterite:
Closing as turning this into a deprecation warning will require different code.

TomAFrench:
Reopening this as a draft so we can track status of repositories being updated to use correct array index types.

asterite:
Closed in favor of #7908

PR #7806, opened 2025-03-25
fix: better logic for determining whether a struct is constructed or not

### LINKED ISSUE #6172: Struct is not considered as "used" despite being passed as a generic when constructing another type.
Noticed this in `noir-bignum` but will add a smaller repro later.

The `Test2048Params` struct is never constructed within the library and should not be exposed as public:
https://github.com/noir-lang/noir-bignum/blob/910dcbb8c2f07efb45416d267e75fdb3361b5b01/src/bignum_test.nr#L12

However, I would consider the below line as a usage of this struct:
https://github.com/noir-lang/noir-bignum/blob/910dcbb8c2f07efb45416d267e75fdb3361b5b01/src/bignum_test.nr#L366

We should also check to see whether this type is passed as a generic parameter when constructing another type. I imagine that passing it as a generic to a function should also satisfy this check.

asterite:
# Description

## Problem

Resolves #6172 and a few other cases mentioned over Slack

## Summary

I initially thought about "solving" this by marking any struct as used if it's mentioned. However, that would have made any struct with an `impl` immediately used:

```noir
struct Foo {}

impl Foo { // Here Foo is referenced so it would be considered as used... but it's not!
}
```

So I ended up determining that if a type appears in a position where it's "constructed", then it's constructed. These are:
1. If a type is mentioned in a let statement, like `let x: Foo = ...` then it's constructed (the right-hand side could be `std::mem::zeroed()`)
2. If a type is used in an expression, like `Foo::bar` or `Foo::bar()`, then it's constructed
3. If a type is used in generic arguments in an expression, like `call::<Foo>(...)`, then it's constructed
4. If a type is used in a return type position, like `fn foo() -> Foo`, then it's constructed (the body could be `std::mem::zeroed()`)
6. Otherwise, type mentions don't mean they are constructed. For example: a function parameter, an impl position, a trait constraint, etc.

I think this logic makes sense, though I'm not 100% sure.

## Additional Context

One downside of this approach is that now, whenever we lookup a path or a type, we need to decide whether that lookup marks types as used or not. This ends up adding a new boolean to many methods, which has to be passed around in some cases. It makes the code a bit longer, but maybe it's fine?

## Documentation

Check one:

# PR Checklist


asterite:
Done! I ended up creating an enum, `PathResolutionMode`, that for now just controls whether datatypes are marked as referenced or used. In the future there could be another mode, `Expression`, which is used when solving paths inside expressions (that implies "mark as used" and also it should find values at the leaf and not types... I think! I didn't think too much about this yet).

compiler/noirc_frontend/src/elaborator/path_resolution.rs
```diff
@@ -102,12 +102,35 @@ enum MethodLookupResult {
     FoundMultipleTraitMethods(Vec<(TraitId, Ident)>),
 }
 
+/// Determines whether datatypes found along a path are to be marked as referenced
+/// or used (see UsageTracker::mark_as_referenced and UsageTracker::marke_as_used)
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub(super) enum PathResolutionMode {
+    MarkAsReferenced,
+    MarkAsUsed,
+}
```
TomAFrench, line 111: Can you turn these references into links for the docs?

asterite: Done!

compiler/noirc_frontend/src/elaborator/path_resolution.rs
```diff
@@ -110,12 +110,36 @@ enum MethodLookupResult {
     FoundMultipleTraitMethods(Vec<(TraitId, Ident)>),
 }
 
+/// Determines whether datatypes found along a path are to be marked as referenced
```
michaeljklein, line 113: Please note the cases where a type is constructed v.s. referenced in the docs

michaeljklein (changes requested):
LGTM besides a bit of docs. @jfecher do you want to re-review this?

compiler/noirc_frontend/src/elaborator/path_resolution.rs
asterite: Makes sense. Done!

michaeljklein (approved):

PR #7813, opened 2025-03-26
fix: implicit associated types come before explicit types

### LINKED ISSUE #7807: Compiler crashes on turbofish generic with associated type
### Aim

This code:

```noir
trait Trait {
    let N: u32;
}

pub struct Foo {}

impl Trait for Foo {
    let N: u32 = 1;
}

fn main() {
    foo::<Foo>(); // This call causes the crash
}

fn foo<T>()
where
    T: Trait,
{}
```

### Expected Behavior

Should compile.

### Bug

It crashes:

```
Message:  while substituting (unbound): expected kind of unbound TypeVariable (Numeric(u32)) to match the kind of its binding (Normal)
Location: compiler/noirc_frontend/src/hir_def/types.rs:2402
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #7807

## Summary

This one was tough to debug and, to be honest, I'm not sure this is the right fix. Or, put another way, it's hard to easily know where implicit generics go in a list of generics.

On one hand there's `Type::instantiate_with` that receives a count of implicit generics that are filled in. These are put **before** explicit generics:

https://github.com/noir-lang/noir/blob/fa7f175c38d0e9b08739020a2f5b195871df4437/compiler/noirc_frontend/src/hir_def/types.rs#L2308-L2336

Then, when building a function's generics, associated types are added to the list of generics:

https://github.com/noir-lang/noir/blob/fa7f175c38d0e9b08739020a2f5b195871df4437/compiler/noirc_frontend/src/elaborator/mod.rs#L938-L941

However, notice that they are added **after** the other generics. This causes a mismatch.

So, I changed the order. This made all existing tests pass, which means that either this fix is wrong or that it wasn't tested (which would hopefully explain why it didn't work in the first place).

I don't know if implicit generics are added in other places that would cause this change to break.

I do wonder why associated types exist in a single flattened list of generics. If they are associated to types, shouldn't they be "inside" those types somehow and not on a flattened list? I tried some snippets with multiple traits being passed and they seem to work fine though.

## Documentation

Check one:

# PR Checklist


jfecher (approved):
Thanks for looking into this. Honestly... I'm not 100% sure this is correct either (and that's a problem). This would be a good item to revisit & document more thoroughly pre 1.0. For now, as long as this passes all our tests & external checks we can merge it.

The reason I'm not 100% sure its correct is that I believe the prepend implicit generics comes from the fact we can have generics on an outer `impl<T> Foo { ... }` which applies to all methods inside the impl but are still excluded from being specified via turbofish on the method. Self is prepended for a similar reason. We've been a bit unstructured with this assumption in the past though (maybe we can abstract it out with a wrapper or something so only one place needs to handle it?). This makes enough sense to me to merge though.

asterite:
> The reason I'm not 100% sure its correct is that I believe the prepend implicit generics comes from the fact we can have generics on an outer impl<T> Foo { ... } which applies to all methods inside the impl but are still excluded from being specified via turbofish on the method.

Indeed! This is why I wasn't sure either and why I included a test for it (the second one), which somehow passes (and I verified that it works correctly with an executable program, though I didn't include that full program in this PR).

PR #7825, opened 2025-03-27
feat: add basic comptime correctness tests

rkarabut:
# Description

Adds initial basic implementation for comparison between comptime interpreter and compiled Noir results, testing arithmetic binary operations for now (a building block for future fuzzer).

## Additional Context

Bitwise operations tests are added but currently ignored, as they fail on compilation with the results of u32 <op> u32 literals not getting cast to Field as they probably should (maybe warrants its own issue?)

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/nargo_cli/tests/comptime_correctness.rs
+++ b/tooling/nargo_cli/tests/comptime_correctness.rs
@@ -1,0 +1,139 @@
+mod common;
+
+use std::sync::LazyLock;
+use std::{cell::RefCell, collections::BTreeMap};
+
+use acvm::{FieldElement, acir::native_types::WitnessStack};
+use nargo::{foreign_calls::DefaultForeignCallBuilder, ops::execute_program};
+use noirc_abi::input_parser::InputValue;
+use proptest::prelude::*;
+
+static NUM_CASES: LazyLock<u32> = LazyLock::new(|| {
+    std::env::var("COMPTIME_CORRECTNESS_TEST_NUM_CASES").unwrap_or("1".into()).parse().unwrap_or(1)
+});
+
+pub(crate) fn run_snippet(source: String, force_brillig: bool) -> InputValue {
+    let program = match common::prepare_and_compile_snippet(source.clone(), force_brillig) {
+        Ok((program, _)) => program,
+        Err(e) => panic!("failed to compile program; brillig = {force_brillig}:\n{source}\n{e:?}"),
+    };
+
+    let pedantic_solving = true;
+    let blackbox_solver = bn254_blackbox_solver::Bn254BlackBoxSolver(pedantic_solving);
+    let foreign_call_executor = RefCell::new(DefaultForeignCallBuilder::default().build());
+
+    let initial_witness = program.abi.encode(&BTreeMap::new(), None).expect("failed to encode");
+    let mut foreign_call_executor = foreign_call_executor.borrow_mut();
+
+    let witness_stack: WitnessStack<FieldElement> = execute_program(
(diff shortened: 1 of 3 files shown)


tooling/nargo_cli/tests/comptime_correctness.rs
```diff
@@ -0,0 +1,139 @@
...
+use std::sync::LazyLock;
+use std::{cell::RefCell, collections::BTreeMap};
+
+use acvm::{FieldElement, acir::native_types::WitnessStack};
+use nargo::{foreign_calls::DefaultForeignCallBuilder, ops::execute_program};
+use noirc_abi::input_parser::InputValue;
+use proptest::prelude::*;
+
+static NUM_CASES: LazyLock<u32> = LazyLock::new(|| {
+    std::env::var("COMPTIME_CORRECTNESS_TEST_NUM_CASES").unwrap_or("1".into()).parse().unwrap_or(1)
```
aakoshh, line 12:
```suggestion
    std::env::var("NOIR_COMPTIME_CORRECTNESS_TEST_NUM_CASES").unwrap_or("1".into()).parse().unwrap_or(1)
```

Just so every env var is easy to recognise where it belongs to, or list all NOIR_ env vars.

tooling/nargo_cli/tests/comptime_correctness.rs
```diff
@@ -0,0 +1,139 @@
...
+    .expect("failed to execute");
+
+    let main_witness = witness_stack.peek().expect("should have return value on witness stack");
+    let main_witness = &main_witness.witness;
+
+    let (_, return_value) = program.abi.decode(main_witness).expect("failed to decode");
+    return_value.expect("should decode a return value")
+}
+
+#[cfg(test)]
```
aakoshh, line 43: Do we need this `cfg`? I thought it's all tests in the `tests` directory.

rkarabut: Yeah I think we don't

tooling/nargo_cli/tests/comptime_correctness.rs
```diff
@@ -0,0 +1,139 @@
...
+fn comptime_check_field_expression(strategy: BoxedStrategy<String>, num_cases: u32) {
+    proptest!(ProptestConfig::with_cases(num_cases), |(expr in strategy)| {
+        let program = format!("
+        comptime fn comptime_code() -> Field {{
+            {expr}
+        }}
+        
+        fn runtime_code() -> Field {{
```
aakoshh, line 53: One aspect to consider here is whether to splice the `expr` verbatim into the ACIR code, or try to use input witnesses. When you give it like `1+2`, that expression is going to be evaluated at compile time by the SSA optimisation passes. That _is_ worth testing on its own, but be aware that there will be a single opcode, and potentially even the `assert_eq` at the bottom will say "trivially true". 

By contrast you could generate e.g. `fn runtime_code(a: u32, b: u32) -> Field { a + b }` and then call `main` with `a` and `b` passed as inputs that the compiler isn't aware of, in which case the assertion will be like:

```
fn main(a: u32, b: u32) {
  assert_eq(3, runtime_code(a, b));
}
```

Yet another aspect is to use `unconstrained fn runtime_code` to use Brillig. Can also be achieved using the `force_brillig` compile option.

rkarabut: For now replaced with the code preventing optimization pass evaluation. Will probably be best to test both later, yes

PR #7827, opened 2025-03-27
fix: checks for index out of bounds also for arrays

### LINKED ISSUE #7759: Unchecked mul/add/sub problems
### Aim

I'm trying to find differences between Brillig and ACIR execution results using fuzzing

### Expected Behavior

### Bug

Brillig works in context of the variable, but acir not. For example for u32
unchecked_mul(2^31, 2) = 0 in brillig
unchecked_mul(2^31, 2) = 2^32 in acir
unchecked_sub(0, 1) = 2^32 - 1 in brillig
unchecked_sub(0, 1) = -1 in acir

So there is bug found in Noir arrays with unchecked operations

### main.nr
```nr
pub fn main(v0: u64,v1: u64,v2: u64,v3: u64,v4: u64,v5: u64,v6: u64, index: u32) -> pub (u64, u64) {
  let v7: [(u64, u64); 2] = [(v1, v5), (v1, v5)];
  v7[index]
}
```

### Prover.toml
```
"v0" = "1"
"v1" = "2"
"v2" = "3"
"v3" = "4"
"v4" = "5"
"v5" = "6"
"v6" = "7"
"index" = "2147483648" # 2 **31
```

### SSA 
```
  b0(v0: u64, v1: u64, v2: u64, v3: u64, v4: u64, v5: u64, v6: u64, v7: u32):
    v8 = make_array [v1, v5, v1, v5] : [(u64, u64); 2]
    v10 = unchecked_mul v7, u32 2
    v11 = array_get v8, index v10 -> u64
    v13 = unchecked_add v10, u32 1
    v14 = array_get v8, index v13 -> u64
    return v11, v14
}
```

### nargo execute
`Failed to solve program: 'Index out of bounds, array has size 4, but index was 4294967296'`

### nargo execute --force-brillig
```
[heck_noir] Circuit witness successfully solved
[heck_noir] Circuit output: Vec([Field(2), Field(6)])
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #7759

## Summary\*
The purpose of unchecked operations in Noir is to avoid overflow checks, and therefore they are supposed to be used only if overflow is guaranteed to not happen.
In the issue, we have an unchecked operation that overflows, so I have simply added a check to avoid this.

## Additional Context

ACIR and Brillig behaviour are inconsistent, and there is no (easy) way to fully fix this:
Brillig arithmetic is wrapping, while Noir is mainly checked, with some unchecked variants. In the latter, overflows will really overflow the integer because ACIR base type is a field, while Brillig integer type cannot store more than the specified bit size.
However, Noir checked arithmetic is emulated for unconstrained code by adding overflow checks when generating the brillig bytecode, so the only inconsistency is with the Noir unchecked operations which should never overflow.

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
These regressions look to be happening from a few places. 

Taking the largest regression as an example:

> **global_var_regression_entry_points_inliner_max**	+133,797 ❌	**+284674.47%**

A lot of this regression looks to be from this loop:
```
    let mut acc: Field = 0;
    for i in 0..2 {
        for j in 0..2 {
            acc += EXPONENTIATE[i][j];
        }
    }
```
We should be able to avoid index out of bounds checks in this situation as we know the maximum bound of `i` and `j`. We utilize the loop induction variable for optimizations in the loop invariant code motion pass already. The pass does not have support for hoisting or removing less than instructions based off of the induction variable but this logic could be added.

We also already do [index checks during Brillig gen](https://github.com/noir-lang/noir/blob/6f0b1192be8fbd4936e9119c492a7fcf1b6b92f0/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs#L794). So this means we are doing double array checks. We already skip this check for slices (which have the check from SSA gen). We can just remove the Brillig gen for arrays fully now as well.

compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs
```diff
@@ -445,8 +445,22 @@ impl FunctionContext<'_> {
         let type_size = Self::convert_type(element_type).size_of_type();
         let type_size =
             self.builder.numeric_constant(type_size as u128, NumericType::length_type());
-        // This shouldn't overflow as we are reaching for an initial array offset
-        // (otherwise it would have overflowed when creating the array)
+
+        let array_type = &self.builder.type_of_value(array);
+
+        // Checks for index Out-of-bounds
```
vezenovm, line 451: Nice, I guess we are generating a bunch of repeat access checks for slices at the moment

compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs
```diff
@@ -480,11 +483,11 @@ impl FunctionContext<'_> {
     /// Prepare a slice access.
     /// Check that the index being used to access a slice element
     /// is less than the dynamic slice length.
-    fn codegen_slice_access_check(&mut self, index: ValueId, length: Option<ValueId>) {
+    fn codegen_access_check(&mut self, index: ValueId, length: Option<ValueId>) {
```
vezenovm, line 486:
```suggestion
    fn codegen_access_check(&mut self, index: ValueId, length: ValueId) {
```
nit: This should not need to accept an option anymore

vezenovm (approved):
Looks good! Just one nit with regards to the inputs to `codegen_access_check`. 

Looks like after #7893 this PR now brings some large Brillig improvements with the removal of the array access check in Brillig gen. I see we still have some small regressions for certain circuits. This is not surprising as we are adding extra checks onto array accesses and the regressions look manageable.

compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs
guipublic: Done

vezenovm (approved):

compiler/noirc_frontend/src/tests/unused_items.rs
```diff
@@ -332,12 +332,13 @@ fn considers_struct_as_constructed_if_trait_method_is_called() {
 #[test]
 fn considers_struct_as_constructed_if_mentioned_in_let_type() {
     let src = "
-    struct Bar {}
+    pub struct Bar {}
 
-    pub fn main() {
-        let array = [];
+    pub fn foo(array: [Bar; 1]) {
```
vezenovm, line 337: @asterite will adding `Bar` in the parameter here override that we are testing whether the struct is constructed in the let type?

asterite: Good question! No, `Bar` as an input argument won't be considered as constructed. It's only the case in "return" positions (a return type, a let type, etc.). That said, it's true that if that were the case then this test wouldn't be testing this thing properly. I'll add another test 👍 , just a sec...

asterite: Ah, actually I added `pub` here so that already breaks this...

Issue #7321, opened 2025-02-07
Experiment with lightweight immutable references


michaeljklein:
### Problem

- We're currently copying too much in Brillig, but there's no way for users to explicitly control copying
- `inc_rc`/`dec_rc` are tricky to get right: We are conservatively adding `inc_rc`'s to ensure the reference count is sufficiently large, but Akosh found a bunch of cases where they go negative regardless: https://github.com/noir-lang/noir/pull/7297#issuecomment-2639900761

This has led to suggestions such as [Experiment replacing inc_rc & dec_rc with normal operations on integer references](https://github.com/noir-lang/noir/issues/7287)

### Happy Case

Add immutable references with the following minimal features:
- An `[no-&/no-&mut]T` means that `T` is always copied when used in more than one source location
- An `&T` means that `T` is _never_ copied 
- `&(&mut T)` is an error
- `&&T` is equivalent to `&T`

Then, benchmark a few test programs, e.g. [`ram_blowup_regression`](https://github.com/noir-lang/noir/blob/master/test_programs/execution_success/ram_blowup_regression/src/main.nr), between the following three cases:
- Immutable references with the existing `inc_rc`/`dec_rc`'s
- Immutable references with emitting `inc_rc`/`dec_rc`'s disabled
- `master` branch

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

TomAFrench:
Bumping this to post 1.0 as we've made the change to how we do reference counting and we can add in immutable references down the line in a non-breaking change.

PR #7837, opened 2025-03-28
chore(debug): Add ownership analysis pass

jfecher:
# Description

## Problem\*

Working towards https://github.com/noir-lang/noir/issues/7321
Resolves https://github.com/noir-lang/noir/issues/7842

## Summary\*

Adds an explicit ownership pass instead of relying on ssa-gen to insert `inc_rc`s.

This pass traverses the monomorphized AST and currently only inserts a new `Clone(expr)` node exactly where SSA-gen used to insert inc_rcs. I've also documented these locations for future reference. I consider this pass to simplify things since it separates concerns a bit and lets us more easily see where clones are being inserted since `.clone()` is now printed out in each relevant location when running with `--show-monomorphized`.

In a future PR we can change the location of clones based on the `-Zownership` feature but this one aims to make as few changes as possible to start out with.

## Additional Context

The main place we can't quite replicate what SSA was doing is when inserting `dec_rc`s at the end of the function. SSA checked if we were returning the same ValueId as a parameter and did not issue it in that case. This pass does not have ValueIds and so can't do that. I think the old ssa check was unnecessary though (it can be circumvented using references as well) and have excluded it from this PR (so we currently never dec_rc). Currently waiting on CI to see performance implications of this. I'll likely re-add them but would have to always dec_rc even for returned parameters.

## Documentation\*

Check one:
  - No user-facing documentation at least

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
+++ b/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
@@ -901,11 +901,16 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
                 // and become usize::MAX, and then return to 1, then it will indicate
                 // an array as mutable when it probably shouldn't be.
                 if self.brillig_context.enable_debug_assertions() {
-                    let min_addr = ReservedRegisters::usize_one();
+                    let zero =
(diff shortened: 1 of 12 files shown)


jfecher:
This also fixes https://github.com/noir-lang/noir/issues/7842 which I found while working on this PR

PR #7839, opened 2025-03-28
feat: better error message when generic argument cannot be determined

asterite:
# Description

## Problem

Not existing issue, just something I noticed.

## Summary

If you have code like this:

```noir
fn main() {
    foo();
}

fn foo<T>() {}
```

the error you get before this PR is:

```
error: Type annotation needed
  ┌─ src/main.nr:2:5
  │
2 │     foo();
  │     --- Could not determine type of generic argument
```

I think by now we probably know that it refers to a generic argument on the function `foo`, though it's not clear which generic argument it is (there could be many) and the error Rust gives is a bit clearer, so users without too much Noir experience might be confused.

With this PR, the error is now this:

```
error: Type annotation needed
  ┌─ src/main.nr:2:5
  │
2 │     foo();
  │     --- Could not determine the type of the generic argument `T` declared on the function `foo`
```

The same thing is done when a generic argument for a struct or enum cannot be determined.

## Additional Context

I think the logic for searching these generic names is okay though I'm not entirely sure, or maybe there's a more correct way to do this.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/monomorphization/mod.rs
+++ b/compiler/noirc_frontend/src/monomorphization/mod.rs
@@ -9,12 +9,13 @@
 //! The entry point to this pass is the `monomorphize` function which, starting from a given
 //! function, will monomorphize the entire reachable program.
 use crate::ast::{FunctionKind, IntegerBitSize, UnaryOp};
-use crate::hir::comptime::InterpreterError;
+use crate::hir::comptime::{InterpreterError, Value};
 use crate::hir::type_check::{NoMatchingImplFoundError, TypeCheckError};
 use crate::node_interner::{ExprId, GlobalValue, ImplSearchErrorKind};
 use crate::shared::{Signedness, Visibility};
 use crate::signed_field::SignedField;
 use crate::token::FmtStrFragment;
+use crate::{DataType, Shared, TypeVariableId};
 use crate::{
     Kind, Type, TypeBinding, TypeBindings,
     debug::DebugInstrumenter,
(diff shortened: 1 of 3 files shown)


compiler/noirc_frontend/src/monomorphization/mod.rs
```diff
@@ -2297,8 +2371,8 @@ fn unwrap_struct_type(
     match typ.follow_bindings() {
         HirType::DataType(def, args) => {
             // Some of args might not be mentioned in fields, so we need to check that they aren't unbound.
-            for arg in &args {
-                Monomorphizer::check_type(arg, location)?;
+            for (index, arg) in args.iter().enumerate() {
+                check_struct_generic_type(arg, location, &def, index)?;
```
jfecher, line 2375: Looks like we call this for structs or enums but the error always says "struct"

jfecher (approved):
Great work

asterite:
@jfecher Regarding your comment about structs/enums, I was writing this:

> Ah, yes. I initially had this call (with a different name) for the enum case, thinking it would trigger when using an enum variant... but it didn't! It required the other piece of code. I'll try to see if I can trigger it here too.

asterite:
Looking at the code it seems it would trigger on enum constructors that are "functions", so I tried this:

```noir
enum Foo<T, U> {
    Bar(T),
}

fn main() {
    let _ = Foo::Bar(1);
}
```

and I get:

```
error: Type annotation needed
  ┌─ src/main.nr:6:18
  │
6 │     let _ = Foo::Bar(1);
  │                  --- Could not determine the type of the generic argument `U` declared on the function `Bar`
```

so it might be that the code for `unwrap_enum_type` happens after this. So, I'll merge this as-is for now.

PR #7847, opened 2025-03-28
fix: do not use enable_side_effect in acir-gen if instruction does not req…

### LINKED ISSUE #7836: Wrong modulo computation
### Aim

I ran into a bug using beta3.  
A modulo operation (i didnt try others) in a function returns the wrong result if the function is called previously in 
an if condition.  
Sorry description is not clear.

```rust
fn main(x: u32) {
    if () {
        bad_function(x);
    }
    bad_function(x);
}

fn bad_function(input: u32) {
    let mod_100 = input % 100;
}
```

### Expected Behavior

In the example above, `mod_100` returns 0 even if the input isn't divisible by 100

### Bug

I made a small repo with more explanations to reproduce the bug: https://github.com/teddav/noir-beta3-compiler-bug.  
I talked to @saleel who confirmed that there's an issue.

I'd love to fix the issue myself, it would be a great way for me to look into the Noir compiler. I'm not familiar with it at all though, so I'll definitely need some help

### Nargo Version

nargo version = 1.0.0-beta.3 noirc version = 1.0.0-beta.3+ceaa1986628197bd1170147f6a07f0f98d21030a

Yes

### Support Needs

where to start looking? 😊

guipublic:
…uire it

# Description

## Problem\*

Resolves #7836

## Summary\*
Instruction that does not requires predicate should not use it during acir-gen

## Documentation\*

Check one:

# PR Checklist\*


guipublic:
Interesting failure of regression_6834 with my fix. What happens is that we compute -1 under predicate and then divides it by 4, which fails because the quotient and remainder are bounded.
This means that a division can fail even if dividing by a constant.
I cancelled my fix by saying that they do require acir-gen predicate. I might as well completely remove my first fix now...

asterite:
Should all mentions of `current_side_effects_enabled_var` be replaced with `predicate` in `convert_ssa_instruction`? I see one more on line 789.

guipublic:
> Should all mentions of `current_side_effects_enabled_var` be replaced with `predicate` in `convert_ssa_instruction`? I see one more on line 789.

So I did my fix only for binary instructions, purposely not to impact the rest.
The predicate on line 789 will not impact the values so it is OK to leave it.
Furthermore, removing it will not work.
So strictly speaking that means that constraint-not-equal does use the predicate in acir-gen, but I'd like to review the impacts before changing this.

PR #7845, opened 2025-03-28
fix: address warnings in stdlib

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR resolves some warnings in the stdlib added by #7820 

## Documentation\*

Check one:
- [x] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #7848, opened 2025-03-28
feat: Remove range constraints on witnesses when use as array index is more restrictive

TomAFrench:
# Description

## Problem\*

Resolves comment  https://github.com/noir-lang/noir/pull/7845#issuecomment-2762017251

## Summary\*

I've updated the `redundant_range` optimizer to take into account the usage of witnesses in `MemOp` opcodes as indices as an implicit range constraint which allows us to remove the explicit range constraints on that witness.

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
The code looks good! That said, I don't understand how the ACVM works yet so someone else could also review this.

acvm-repo/acvm/src/compiler/optimizers/redundant_range.rs
```diff
@@ -157,20 +175,38 @@ impl<F: AcirField> RangeOptimizer<F> {
     }
 }
 
+fn memory_block_implied_max_bits(init: &[Witness]) -> u32 {
+    8 * size_of::<usize>() as u32 - init.len().leading_zeros()
+}
```
vezenovm, line 180:
```suggestion
fn memory_block_implied_max_bits(init: &[Witness]) -> u32 {
    32 - (init.len() as u32).leading_zeros()
}
```
Why bother with `size_of::<usize>()`? This will potentially change the calculation with a different target too. I find an explicit constant more clear

TomAFrench: I'm using this as I expect `init.len().leading_zeros()` to change to match any change in usize.

TomAFrench:
fyi @zac-williamson as I know you've been doing crimes against the type system to avoid these range constraints.

acvm-repo/acvm/src/compiler/optimizers/redundant_range.rs
vezenovm: I see how `init.len().leading_zeros()` will change to match a change in usize but couldn't this still then return something above 32 bits which is our usize in Noir. I don't foresee how this would be unsafe so I won't block on this, but just feels weird to rely on Rust's usize when it changes with the target while Noir's usize does not.

vezenovm (approved):

acvm-repo/acvm/src/compiler/optimizers/redundant_range.rs
```diff
@@ -49,6 +49,7 @@ impl<F: AcirField> RangeOptimizer<F> {
     /// be 16 bits.
     fn collect_ranges(circuit: &Circuit<F>) -> BTreeMap<Witness, u32> {
         let mut witness_to_bit_sizes: BTreeMap<Witness, u32> = BTreeMap::new();
+        let mut memory_block_lengths_bit_size: HashMap<BlockId, u32> = HashMap::new();
```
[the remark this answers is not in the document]
TomAFrench: I've added this to the module-level documentation (along with constant assignments)

TomAFrench:
I'm going to merge as I feel that I've addressed the documentation point, let me know if you think it needs extra attention.

PR #7860, opened 2025-03-31
chore(debug): Add ownership analysis pass

jfecher:
# Description

## Problem\*

Working towards https://github.com/noir-lang/noir/issues/7321
Resolves https://github.com/noir-lang/noir/issues/7842

## Summary\*

Adds an explicit ownership pass instead of relying on ssa-gen to insert `inc_rc`s.

This pass traverses the monomorphized AST and currently only inserts a new `Clone(expr)` node exactly where SSA-gen used to insert inc_rcs. I've also documented these locations for future reference. I consider this pass to simplify things since it separates concerns a bit and lets us more easily see where clones are being inserted since `.clone()` is now printed out in each relevant location when running with `--show-monomorphized`.

In a future PR we can change the location of clones based on the `-Zownership` feature but this one aims to make as few changes as possible to start out with.

## Additional Context

The main place we can't quite replicate what SSA was doing is when inserting `dec_rc`s at the end of the function. SSA checked if we were returning the same ValueId as a parameter and did not issue it in that case. This pass does not have ValueIds and so can't do that. I think the old ssa check was unnecessary though (it can be circumvented using references as well) and have excluded it from this PR (so we currently never dec_rc). Currently waiting on CI to see performance implications of this. I'll likely re-add them but would have to always dec_rc even for returned parameters.

This PR is a remake to update the CI comments.

## Documentation\*

Check one:
- [x] No documentation needed.
  - No user-facing documentation at least
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #7861, opened 2025-03-31
feat(experimental): Implement separate `-Zownership` analysis for ownership pass

jfecher:
# Description

## Problem\*

Working towards https://github.com/noir-lang/noir/issues/7321

## Summary\*

Extends the ownership pass (added in https://github.com/noir-lang/noir/pull/7860) to implement some experimental ownership semantics that change where we `clone` things, enabled by the `-Zownership` flag.

Terminology note: `clone` refers to where we insert `inc_rc`s, brillig's reference counting & copy on write semantics remain the same, although actual reference counts may differ.

With this flag set:
- Clones occur moreso where one coming from Rust may expect. Normally these happen when a variable is used multiple times or when a dereference occurs.
- A last use analysis is done so that we move the last instance of a variable instead of clone it unnecessarily.
- We still clone implicitly. Since https://github.com/noir-lang/noir/pull/7860 we can at least `--show-monomorphized` to see where clones are inserted.

Edit: Also see https://github.com/noir-lang/noir/pull/7951 for brillig performance differences of `-Zownership` versus master

## Additional Context

The main changes in this PR for reviewing are:
- The new `last_uses` pass which is done on each function before the ownership pass only if `-Zownership` is set. It's purpose is to find where we can move variables instead of cloning.
- The changes to the ownership pass to clone in different places depending on whether `experimental_ownership_feature` is set.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
Initial experiment results of this PR ([3ff2189](https://github.com/noir-lang/noir/pull/7861/commits/3ff218951eabd10abc3460bac4c9f9ae96471f4d)) 
`-Zownership`:

    private-kernel-init: Total arrays copied: 657
    private-kernel-tail: Total arrays copied: 3
    rollup-base-private: Total arrays copied: 5817
    rollup-base-public: Total arrays copied: 7371
    rollup-block-merge: Total arrays copied: 903

Master / no `-Zownership`:

    private-kernel-init: Total arrays copied: 136
    private-kernel-tail: Total arrays copied: 3
    rollup-base-private: Total arrays copied: 895
    rollup-base-public: Total arrays copied: 644
    rollup-block-merge: Total arrays copied: 902

Note: array copy count without the `-Zownership` flag has remained unchanged for these tests after the merge of the ownership pass PR

Interpretation
- Base `-Zownership` with no other changes is significantly less efficient as expected.
- Currently with that flag we clone on every variable usage and dereference, although the vast majority of clones seem to come from dereferences. Running without this is unsound but does result in ~200 copies in rollup-base-private versus 5817. Again, it is unsound though so it is not an actual option - just an indication of how expensive different places to inc_rc are and how often a variable is repeated versus the dereference operator being used.
- rollup-block-merge interestingly had a large but almost identical copy count in both cases. One reason why the `-Zownership` flag may not have substantially increased the copy count for this test is if almost every write in that test is already a clone...

jfecher:
Changes:
- Last use of a variable is now identified as a move and is no longer cloned
- In a field access `a.b.c` we now `a.b.c.clone()` instead of `a.clone().b.c`
- In an array access `array[index]` we now `array[index].clone()` instead of `array.clone()[index]`
- No longer clone `rhs` in `*rhs` since we clone the result already anyway
```
private-kernel-init: Total arrays copied: 657
private-kernel-tail: Total arrays copied: 3
rollup-base-private: Total arrays copied: 5600 (-217)
rollup-base-public: Total arrays copied: 7360 (-11)
rollup-block-merge: Total arrays copied: 903
```
Overall much fewer differences than I was expecting. Going to start looking into the tests specifically now starting with `private-kernel-tail`. This may be the limit of compiler changes without explicitly changing the stdlib & repositories.

jfecher:
Ok, I've been doing a lot of experimenting mostly in the [jf/ownership-2nd-class-refs](https://github.com/noir-lang/noir/tree/jf/ownership-2nd-class-refs) branch and I think I've narrowed down the main reason why we have so many clones.

In experimenting on rollup-base-public I've noticed:
- Manually adding `&` gets you nowhere at all. I've spent hours going through the repository adding these to not have the clone counter drop even by 1 (although there were fewer reference count increases, all we need is for it to be >1 to clone on a write).
- Overriding the compiler to (unsoundly):
  - Give no rc changes when passing anything into a function (treat all parameters as `&` automatically)
  - Give no rc changes when accessing a struct/tuple member (treat all `foo.bar.baz` as `&foo.bar.baz`)
  - Give no rc changes when extracting an array from a nested array
All of these resulted in not a single fewer array clone at runtime.

What did change things was giving no rc changes when dereferencing values like `*self`. This changed the rollup-base-public copy count from 7360 down to 32. For context, even when we take out all the 3 other "give no rc" items above, those additional changes only bring us down to 29 after. What is happening here is that it is very common to have a large `Context` object containing many arrays, passed around as `&mut self`. This gets dereferenced automatically whenever a field is accessed so we have many instances of `(*self).foo.bar` internally which with the ownership changes clones (inc_rcs) everything in `self`. I'm going to change this now to try to delay the dereference to `*(self.foo.bar)` instead and see how that affects the numbers.

jfecher:
Ok, made another small change in [01fb965](https://github.com/noir-lang/noir/pull/7861/commits/01fb96514dc956acadfe737b41fe757945ee33a7) to exclude `foo` in `*foo` from clones. This was already meant to be the case but if it was used in an array access it went through another method instead (`handle_reference_expression`) and ended up being cloned. I've added the case to that method and the results look much more like when I removed cloning from `*` entirely (although we still are currently! We're just doing the delaying optimization from `(*foo).bar` to `*(foo.bar)`). Anyways, here are the current results:
```
private-kernel-init: Total arrays copied: 50 (-371)
private-kernel-tail: Total arrays copied: 3
rollup-base-private: Total arrays copied: 587 (-2252)
rollup-base-public: Total arrays copied: 32 (-2469)
rollup-block-merge: Total arrays copied: 903 (still no change)
```

jfecher:
This is ready for review now. For reviewing, the main changes are to the ownership pass and the new last uses pass. No changes are expected if not passing the `-Zownership` flag (although this flag was removed when running our execution tests)

jfecher:
Looks like there is a test failure when running `nargo e -Zownership --force-brillig` on the `slice_regexes` test:
```
~/.../execution_success/slice_regex> nargo e --force --force-brillig -Zownership
warning: struct `Plus` is never constructed
    ┌─ src/main.nr:156:8
    │
156 │ struct Plus<T, let N: u32, let N_PRED: u32> {
    │        ---- struct is never constructed
    │

Match { succeeded: true, match_ends: 4, leftover: &[] }
error: Assertion failed: 'Bit size for lhs 32 does not match op bit size 8'
   ┌─ src/main.nr:44:34
   │
44 │                 matches_input &= first_elem == c;
   │                                  ---------------
   │
   = Call stack:
     1. src/main.nr:181:18
     2. src/main.nr:66:30
     3. src/main.nr:64:26
     4. src/main.nr:122:13
     5. src/main.nr:44:34
```

vezenovm:
> error: Assertion failed: 'Bit size for lhs 32 does not match op bit size 8'

This failure makes me think we are accessing the slice length rather than the first element somewhere (although this is speculation I have not looked at the code in this PR).

compiler/noirc_evaluator/src/brillig/brillig_ir/procedures/vector_pop_front.rs
```diff
@@ -91,11 +91,6 @@ pub(super) fn compile_vector_pop_front_procedure<F: AcirField + DebugToString>(
                 source_capacity.address,
                 BrilligBinaryOp::Sub,
             );
-            brillig_context.codegen_initialize_vector_metadata(
-                target_vector,
-                target_size,
-                Some(source_capacity),
-            );
```
jfecher, line 98: This seems to fix the odd bit size errors while not failing any tests. @vezenovm sanity check for this fix. It is odd to just remove code, at the same time though the rc is one so we are directly mutating the same slice.

jfecher:
Looks like the bug is present in master and was just triggered by the differing RC values in this PR. I've created an issue here https://github.com/noir-lang/noir/issues/7976 so that we can tackle it later. I've commented out the relevant code in slice pop for now (although the issue seems to be with slices more broadly) so that we can merge this PR separately.

compiler/noirc_frontend/src/ownership/last_uses.rs
```diff
@@ -0,0 +1,410 @@
...
+            BranchesPath::IfOrMatch { branch_index: _, rest } => *path = *rest,
+        }
+    }
+
+    fn loop_index(&self) -> usize {
+        self.current_loop_and_branch.len() - 1
+    }
+
+    fn declare_variable(&mut self, id: LocalId) {
+        let loop_index = self.current_loop_and_branch.len() - 1;
```
asterite, line 178: Maybe use `self.loop_index()` here?

compiler/noirc_frontend/src/ownership/last_uses.rs
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    fn track_variables_in_binary(&mut self, binary: &ast::Binary) {
+        self.track_variables_in_expression(&binary.lhs);
+        self.track_variables_in_expression(&binary.rhs);
+    }
+
+    /// - Extracting an array from another array (`let inner: [_; _] = array[0];`):
+    ///   - Extracting a nested array from its outer array will always increment the reference count
+    ///     of the nested array.
```
asterite, line 301: Is this comment relevant?

compiler/noirc_frontend/src/ownership/last_uses.rs
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    fn track_variables_in_assign(&mut self, assign: &ast::Assign) {
+        self.track_variables_in_lvalue(&assign.lvalue);
+        self.track_variables_in_expression(&assign.expression);
+    }
+
+    fn track_variables_in_lvalue(&mut self, lvalue: &ast::LValue) {
+        match lvalue {
+            ast::LValue::Ident(_) => (),
```
asterite, line 397: Do we need to handle this ident? If not, why can it be omitted?

compiler/noirc_frontend/src/ownership/last_uses.rs
```diff
@@ -0,0 +1,410 @@
...
+        let path =
+            self.current_loop_and_branch.last().expect("We should always have at least 1 scope");
+        let loop_index = self.loop_index();
+
+        if let Some((variable_loop_index, uses)) = self.last_uses.get_mut(&id) {
+            if *variable_loop_index == loop_index {
+                Self::remember_use_of_variable_rec(uses, path, variable);
+            } else {
+                *uses = Branches::None;
+            }
```
asterite, line 192: Is there a way to relate these two if branches to codes that would trigger them? I think I understand that if the variable is in a loop index and we are currently in that index, we update the use. But if it's in a different index (an upper one?) then why this is set to None?

compiler/noirc_frontend/src/ownership/last_uses.rs
jfecher: No - removed. It was originally pasted from the ownership pass

compiler/noirc_frontend/src/ownership/last_uses.rs
jfecher: Nope - I'll add a comment explaining why not. It's because identifiers are never moved into an lvalue position. That position always expects a mutable reference since if the value is moved there'd be no point assigning to it since it could not be used any more.

compiler/noirc_frontend/src/ownership/last_uses.rs
jfecher: There's code examples in the doc comments in the fields of the context struct for this case but I'll add a comment to mention it here.

This is set to None if we're in an upper loop index since that means we are in a loop that the variable wasn't defined in, and a variable can never be moved into a loop (it needs to be cloned). This is a situation like:
```
let foo = bar;  // loop index 0
loop {
    println(foo);  // loop index 1, needs to be copied
}
```
If it was defined in the same loop, it can be moved:
```
loop {
    let foo = bar; // loop index 1
    println(foo);  // loop index 1, can be moved
}
```

compiler/noirc_frontend/src/tests.rs
```diff
@@ -4354,23 +4360,6 @@ fn deny_attaching_mut_ref_to_immutable_object() {
     check_errors!(src);
 }
 
-#[named]
-#[test]
-fn immutable_references_with_ownership_feature() {
```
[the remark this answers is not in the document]
TomAFrench: Can't we just switch out `get_program_using_features!` for the non-macroified `get_program_using_features`? Also remove the `#[named]` attribute.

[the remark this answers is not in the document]
jfecher: The non-macroified version still emits the test_programs test I believe. The macro is just to automatically pass the `function_name!()` argument for the resulting test file name. Similarly the named attribute just provides that function_name macro.

compiler/noirc_evaluator/src/brillig/brillig_ir/procedures/vector_pop_front.rs
```diff
@@ -76,49 +76,50 @@ pub(super) fn compile_vector_pop_front_procedure<F: AcirField + DebugToString>(
...
-            brillig_context.codegen_mem_copy(
-                source_copy_pointer,
-                target_vector_items_pointer,
-                target_size,
-            );
-
-            brillig_context.deallocate_register(source_copy_pointer);
-            brillig_context.deallocate_register(target_vector_items_pointer);
-        }
+    brillig_context.codegen_branch(is_rc_one, |brillig_context, _is_rc_one| {
```
vezenovm, line 79: Do we even need the branch if we are not checking `is_rc_one`

compiler/noirc_frontend/src/ownership/mod.rs
```diff
@@ -260,180 +344,321 @@ fn recur_on_parameter<'typ>(
...
+        let unary = match expr {
+            Expression::Unary(unary) => unary,
+            other => panic!("handle_unary given non-unary expression: {other}"),
+        };
 
-fn handle_while(while_expr: &mut crate::monomorphization::ast::While) {
-    handle_expression(&mut while_expr.condition);
-    handle_expression(&mut while_expr.body);
-}
+        // Don't clone `rhs` if this is a reference or dereference exprssion.
```
vezenovm, line 519:
```suggestion
        // Don't clone `rhs` if this is a reference or dereference expression.
```

compiler/noirc_frontend/src/ownership/mod.rs
```diff
@@ -260,180 +344,321 @@ fn recur_on_parameter<'typ>(
...
+            clone_result = true;
+        } else {
+            self.handle_expression(&mut index.collection);
+        }
+
+        self.handle_expression(&mut index.index);
+
+        if clone_result && contains_array_or_str_type(&index.element_type) {
+            clone_expr(index_expr);
+        }
```
vezenovm, line 565:
```suggestion
        if self.experimental_ownership_feature {
            // Don't clone the collection, cloning only the resulting element is cheaper
            self.handle_reference_expression(&mut index.collection);
        } else {
            self.handle_expression(&mut index.collection);
        }

        self.handle_expression(&mut index.index);

        if contains_array_or_str_type(&index.element_type) {
            clone_expr(index_expr);
        }
```
clone_result will always be true if experimental_ownership_feature is false. Then if experimental_ownership_feature is true we end up setting clone_result to true anyway. Looks like we can get rid of this `clone_result` flag.

compiler/noirc_frontend/src/ownership/mod.rs
```diff
@@ -260,180 +344,321 @@ fn recur_on_parameter<'typ>(
...
     }
-}
 
-/// - Assignments (`x = <expression which returns an array>;`):
-///   - Assigning an array to an existing variable will also increment the reference
-///     count of the array unless it is an array literal.
-fn handle_assign(assign: &mut crate::monomorphization::ast::Assign) {
-    handle_lvalue(&mut assign.lvalue);
-    handle_expression(&mut assign.expression);
+    fn handle_tuple(&mut self, elems: &mut [Expression]) {
```
vezenovm, line 601:
```suggestion
    fn handle_tuple(&mut self, elements: &mut [Expression]) {
```
nit: or add `elems` to cspell

compiler/noirc_frontend/src/ownership/mod.rs
```diff
@@ -260,180 +344,321 @@ fn recur_on_parameter<'typ>(
...
+        for elem in elems {
```
vezenovm, line 602:
```suggestion
        for elem in elements {
```
same as above

compiler/noirc_frontend/src/tests.rs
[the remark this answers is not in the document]
vezenovm: I had old commits in https://github.com/noir-lang/noir/pull/7868 where I allowed passing an optional `test_name` to avoid emitting test programs. It seems like it would be useful for cases like the following. I could push something to could bring it back @jfecher @TomAFrench

vezenovm (approved):
LGTM just some nits

compiler/noirc_evaluator/src/brillig/brillig_ir/procedures/vector_pop_front.rs
jfecher: Removed

compiler/noirc_frontend/src/ownership/last_uses.rs
```diff
@@ -0,0 +1,430 @@
...
+    ) -> HashMap<LocalId, Vec<IdentId>> {
+        let mut context =
+            LastUseContext { current_loop_and_branch: Vec::new(), last_uses: HashMap::default() };
+        context.push_loop_scope();
+        for (parameter, ..) in &function.parameters {
+            context.declare_variable(*parameter);
+        }
+        context.track_variables_in_expression(&function.body);
+        context.get_variables_to_move()
+    }
```
TomAFrench, line 143:
```suggestion
    /// Traverse the given function and return the last use(s) of each local variable.
    /// A variable may have multiple last uses if it was last used within a conditional expression.
    pub(super) fn find_last_uses_of_variables(
        function: &Function,
    ) -> HashMap<LocalId, Vec<IdentId>> {
        Self::find_last_uses_of_variables_internal(function).get_variables_to_move()
    }
    
    fn find_last_uses_of_variables_internal(
        function: &Function,
    ) -> HashMap<LocalId, (usize, Branches)> {
        let mut context =
            LastUseContext { current_loop_and_branch: Vec::new(), last_uses: HashMap::default() };
        context.push_loop_scope();
        for (parameter, ..) in &function.parameters {
            context.declare_variable(*parameter);
        }
        context.track_variables_in_expression(&function.body);
        context.last_uses
    }
```

compiler/noirc_frontend/src/ownership/last_uses.rs
```diff
@@ -0,0 +1,430 @@
...
+            }
+            ast::LValue::MemberAccess { object, field_index: _ } => {
+                self.track_variables_in_lvalue(object);
+            }
+            ast::LValue::Dereference { reference, element_type: _ } => {
+                self.track_variables_in_lvalue(reference);
+            }
+        }
+    }
+}
```
TomAFrench, line 430:
```suggestion
}

#[cfg(test)]
mod test {
    use crate::{
        monomorphization::ast::{IdentId, LocalId},
        ownership::last_uses::Branches,
        test_utils::get_monomorphized,
    };

    #[test]
    fn smoke_test() {
        let src = "
        fn main(d: [Field; 2]) {
            if len(d) == 2 {              // use 1 of d
                if len(d) == 2 {          // use 2 of d
                    assert(eq(d, [5, 6]));  // use 3 of d
                }
            } else {
                assert(eq(d, [5, 6]));      // use 4 of d
            }
        }

        fn eq(lhs: [Field; 2], rhs: [Field; 2]) -> bool {
            (lhs[0] == rhs[0]) & (lhs[1] == rhs[1])
        }
        
        fn len(arr: [Field; 2]) -> u32 {
            2
        }
        ";

        let program = get_monomorphized(src, "foobar", crate::test_utils::Expect::Success).unwrap();

        let function = program.main();
        let last_uses = super::Context::find_last_uses_of_variables_internal(function);

        let d_local_id = LocalId(0);
        let (_loop_index, d_last_uses) = &last_uses[&d_local_id];
        assert_eq!(
            *d_last_uses,
            Branches::IfOrMatch(vec![Branches::Direct(IdentId(7)), Branches::Direct(IdentId(10))])
        );
    }
}
```

It's probably worth having a sanity check or two here so check that we get the correct paths to the last uses.

Annoyingly we run into the "every frontend test will spit out something into `test_programs`" problem here however.

TomAFrench (approved):
Nit plus we could probably benefit from some quick testing of the `last_uses` pass. Testing exact values for the `IdentId` is probably not the most useful but at least we can track structure.

Otherwise LGTM

compiler/noirc_frontend/src/tests.rs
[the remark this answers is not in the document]
TomAFrench: If it would be quick then please do. It's a little unfortunate that the program writing got embedded right at the very core of the testing logic for some reason.

compiler/noirc_frontend/src/ownership/last_uses.rs
jfecher: Why split this up into two functions?

compiler/noirc_frontend/src/tests.rs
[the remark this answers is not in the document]
vezenovm: For now I just did the optional field https://github.com/noir-lang/noir/pull/8062, but I agree ideally we should move the program writing out.

compiler/noirc_frontend/src/ownership/last_uses.rs
jfecher: This ended up being a much larger item to try to write tests where it is easy to see where the IdentIds correspond to. I'm going to leave this as a future issue for now and hopefully we should have a better testing solution before we enable this by default.

Savio-Sou:
This should be a "[For Experimental Features] Documentation to be submitted in a separate PR", rather than a "No documentation needed" if I understand it correctly? (That is, useful for Noir users to know but not now.)

TomAFrench:
Nope, the changes around ownership are completely hidden to the end user.

PR #7868, opened 2025-04-01
feat: SSA gen unit tests

### LINKED ISSUE #7784: SSA gen unit tests
### Problem

We currently do not have any unit tests for our initial SSA gen (that `ssa_gen` module under `noirc_evaluator/src/ssa`). 

When critical SSA gen changes are made unit tests allow us to easily visualize the changes and also understand why things are being done a certain way. This is useful for both contributors and external auditors. 

Unit tests can also help expose places where we may be not generating code in an ideal manner. 

### Happy Case

We should have a test that is written in Noir (this is a reduced version of our `execution_success/loop` integration test):
```noir
fn main(six_as_u32: u32) {
    assert(plain_loop() == six_as_u32);
}
fn plain_loop() -> u32 {
    let mut sum = 0;
    for i in 0..4 {
        sum = sum + i;
    }
    sum
}
```
And I should be able to check it against its initial SSA:
```
acir(inline) fn main f0 {
  b0(v0: u32):
    v2 = call f1() -> u32
    v3 = eq v2, v0
    constrain v2 == v0
    return
}
acir(inline) fn plain_loop f1 {
  b0():
    v1 = allocate -> &mut u32
    store u32 0 at v1
    jmp b1(u32 0)
  b1(v0: u32):
    v4 = lt v0, u32 4
    jmpif v4 then: b2, else: b3
  b2():
    v6 = load v1 -> u32
    v7 = load v1 -> u32
    v8 = add v7, v0
    store v8 at v1
    v10 = unchecked_add v0, u32 1
    jmp b1(v10)
  b3():
    v5 = load v1 -> u32
    return v5
}
```
With the addition of the SSA parser these tests should be easier to implement. 

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

These do not necessarily block work, but would be good to drill down why code is generated in a certain way. Especially as we begin to have more differences in SSA gen from ACIR/Brillig runtimes. 

Yes

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #7784 

## Summary\*

This PR provides the initial structure to check the generated initial SSA based upon a given Noir program. You can now write a Noir program as we do in the frontend tests and check its generated SSA against an expected SSA string.
For example:
```rust
    let assert_src = "
    fn main(input: u32) {
        assert(input == 5);
    }
    ";
    let assert_ssa = get_initial_ssa(assert_src).unwrap();

    let expected = "
    acir(inline) fn main f0 {
      b0(v0: u32):
        v2 = eq v0, u32 5
        constrain v0 == u32 5
        return
    }
    ";
    assert_normalized_ssa_equals(assert_ssa, expected);
```

This required breaking out some of the frontend tests utility methods. The reasoning for which can be found in the module doc comments.  

I only included two basic unit tests in this PR as I felt those could come in follow-ups and the meat of this PR is the initial testing structure and breaking out of the utility methods into their own module.

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
Looks great!

vezenovm:
I had some conflicts post https://github.com/noir-lang/noir/pull/7753. The test_utils module looks mostly the same as it requires all methods which are used internally by `get_monomorphized`. However, I decided to skip the `emit_compile_test` call if a test path is not provided. Otherwise, I will have to add a similar `function_path` macro or randomization of function names for these SSA gen unit tests. If it is preferred for me to not skip the compilation test I can revert the optional test path. cc @michaeljklein

test_programs/format.sh
```diff
@@ -16,7 +16,7 @@ function collect_dirs {
   for dir in $test_dirs; do
     # TODO(https://github.com/noir-lang/noir/issues/7835): example blocking issue
     # skip generated tests
-    if [[ "${dir}" =~ ^noirc_frontend_* ]]; then
+    if [[ "${dir}" =~ ^(noirc_frontend_ | noirc_evaluator_).*  ]]; then
```
vezenovm, line 19:
```suggestion
    if [[ "${dir}" =~ ^noirc_frontend_* | ^noirc_evaluator_* ]]; then
```

vezenovm:
> I think these generated tests skip the formatter check: [ab64d80#diff-bd2b9a68506a950d69db2a2825ec9bf955b461dcbf581c767db43468efa71b4a](https://github.com/noir-lang/noir/commit/ab64d8045e0f5ac7a83a7c1a19207a25be47dac6#diff-bd2b9a68506a950d69db2a2825ec9bf955b461dcbf581c767db43468efa71b4a)

Thanks for the pointer. I exclude `noirc_evaluator_*` tests now as well for now. We can include them again in a follow-up. I also had to separate the `assert` and `assert_eq` test due to the usage of the `function_path!` macro that has to be unique per test. If you want to take one more quick skim, I'll merge then.

michaeljklein (approved):
LGTM

Issue #7744, opened 2025-03-18
Unsafe doubling of invalid `EmbeddedCurvePoint` missing assignment for witness


michaeljklein:
### Aim

Attempted to test a variation of the regression test for https://github.com/noir-lang/noir/issues/5045, i.e. an inactive doubling of an `EmbeddedCurvePoint` off of the curve:

```noir
use std::embedded_curve_ops::{EmbeddedCurvePoint, embedded_curve_add_unsafe};

// is_active = false
fn main(is_active: bool) -> pub EmbeddedCurvePoint {
    let bad = EmbeddedCurvePoint { x: 0, y: 5, is_infinite: false };
    if is_active {
        embedded_curve_add_unsafe(bad, bad)
    } else {
        bad
    }
}
```

### Expected Behavior

Expected the invalid `embedded_curve_add_unsafe` call to be disabled by `is_active = false`:

```bash
❯ cat Prover.toml
is_active = false
```

### Bug

Execution fails with a low-level ACVM error:

```bash
❯ ~/.nargo/bin/nargo execute --force
warning: Return variable contains a constant value
   ┌─ src/main.nr:6:5
   │  
 6 │ ╭     if is_active {
 7 │ │         embedded_curve_add_unsafe(bad, bad)
 8 │ │     } else {
 9 │ │         bad
10 │ │     }
   │ ╰─────' This variable contains a value which is constrained to be a constant. Consider removing this value as additional return values increase proving/verification time
   │  
   = Call stack:
     1. src/main.nr:6:5

Failed to solve program: 'Cannot solve opcode: missing assignment for witness index 4'
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

Nice-to-have

### Nargo Version

nargo version = 1.0.0-beta.3 noirc version = 1.0.0-beta.3+45ad637273cef317eba42feaf2be0e59d34065ed (git version hash: 45ad637273cef317eba42feaf2be0e59d34065ed, is dirty: false)

### Would you like to submit a PR for this Issue?

None

Issue #5045, opened 2024-05-17
Point Addition in Unentered else Block Causes Circuit Failure


LHerskind:
### Aim

A point addition inside an else that is **NOT** "entered" will cause the circuit to fail entirely. The simulator seems to apply the checks when generating the witness and if using 0 as the base value just fails. 

I encountered this when having a computation that is only to happen when we have non-zero values, but anyway fails even if unused. A minimal test to reproduce is here. 

```
#[test]
fn weird_add() {
    let a = EmbeddedCurvePoint {
        x: 0x1d8eb4378a3bde41e0b6a9a8dcbd21b7ff9c51bdd6ca13ce989abbbf90df3666,
        y: 0x06075b63354f2504f9cddba0b94ed0cef35fc88615e69ec1f853b51eb79a24a0
    };

    let aa = a.double();

    if aa.x == a.add(a).x {
        println(f"We are not trying to do a double of zero");
    } else {
        println("We are trying to do a double of zero");
        let d = EmbeddedCurvePoint { x: 0, y: 0 }.double();
        println(f"d*2 {d}");
    }
}
```

Similar failure is possible with the `multi_scalar_mul` and `fixed_base_scalar_mul` if providing fields with values > 128 bits inside the if. 

### Expected Behavior

I expected to see 
```bash
We are not trying to do a double of zero
```
And have the test pass. 

### Bug

Instead the test fails with the following log.

```bash
We are not trying to do a double of zero
[noir_aztec] Testing encrypted_logs::payload::weird_add... FAIL
Failed to solve program: 'Failed to solve blackbox function: embedded_curve_add, reason: Point (0000000000000000000000000000000000000000000000000000000000000000, 0000000000000000000000000000000000000000000000000000000000000000) is not on curve'
```

### To Reproduce

1. Run the test
2. Profit???
3.
4.

### Project Impact

Blocker

### Impact Context

We need to use the computation to constrain log production in Aztec.

### Workaround

None

### Installation Method

None

### Would you like to submit a PR for this Issue?

None

TomAFrench:
This is addressed by https://github.com/AztecProtocol/aztec-packages/pull/6384

Dupe of https://github.com/noir-lang/noir/issues/4978

TomAFrench:
Reopening as this is actually different.

TomAFrench:
We'll add some behaviour in the flattening pass to apply predicates to these inputs in if statements. This will be blocked pending being able to represent the point at infinity for now though.

PR #7874, opened 2025-04-01
chore: experiment with null point for disabled elliptic curve points

michaeljklein:
# Description

From a discussion with @TomAFrench, setting `is_infinity=true` may be insufficient to ensure that points are on the curve, so this PR sets `x, y = 0` along with `is_infinity = true` to ensure all disabled elliptic curve points are on the curve.

However, when following up with @vezenovm, `is_infinity=true` may be sufficient so this PR experiments with the alternative.

## Documentation\*

Check one:

# PR Checklist\*


guipublic:
Are we relying on x, y to be zero for the infinite point somewhere in the code?

michaeljklein:
@guipublic I thought we might be relying on `x, y` to be zero when I got `Infinite input: embedded_curve_add(infinity, infinity)` errors on the regression tests for these issues:
- https://github.com/noir-lang/noir/issues/5045
- https://github.com/noir-lang/noir/issues/7744

But I'm getting the same errors with these changes:
- https://github.com/noir-lang/noir/actions/runs/14388910660/job/40351767749?pr=7874#step:6:534
- https://github.com/noir-lang/noir/actions/runs/14388910660/job/40351766939?pr=7874#step:6:540

Does `is_infinity=true` ensure that a point is completely valid?

Issue #6805, opened 2024-12-13
Generalise `(x | 1)` optimization to larger bitsize types


TomAFrench:
This optimization is just a special case of  `x | (2^(N) -1) as uN` for a bitsize `N`. While the boolean case is more common to occur we should also handle the others.

This is considered relatively low-priority as it's much less likely to trigger in usual scenarios.

TomAFrench:
Optimization in question being https://github.com/noir-lang/noir/pull/6795

PR #7880, opened 2025-04-02
feat: add simplified OR when `rhs` or `lhs` is all 1s

kashbrti:
# Description
Added simplified or logic for `x|(2^N -1)` for type `uN` 

## Problem\*

Resolves 
#6805 
## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/ir/types.rs
```diff
@@ -85,6 +85,16 @@ impl NumericType {
...
+
+    pub(crate) fn max_value(&self) -> Result<FieldElement, String> {
+        match self {
+            NumericType::Unsigned { bit_size } => Ok(FieldElement::from(2u128.pow(*bit_size) - 1)),
+            NumericType::Signed { bit_size } => Ok(FieldElement::from(2u128.pow(*bit_size) - 1)),
+            NumericType::NativeField => {
+                Err("Cannot get max value for native field type".to_string())
+            }
+        }
+    }
```
TomAFrench: Can you add a `proptest` which checks that this always returns `None` from `value_is_outside_limits`?

TomAFrench: I can help provide some guidance on this.

compiler/noirc_evaluator/src/ssa/ir/types.rs
```diff
@@ -85,6 +85,16 @@ impl NumericType {
     pub(crate) fn is_unsigned(&self) -> bool {
         matches!(self, NumericType::Unsigned { .. })
     }
+
+    pub(crate) fn max_value(&self) -> Result<FieldElement, String> {
+        match self {
+            NumericType::Unsigned { bit_size } => Ok(FieldElement::from(2u128.pow(*bit_size) - 1)),
+            NumericType::Signed { bit_size } => Ok(FieldElement::from(2u128.pow(*bit_size) - 1)),
```
TomAFrench: There's an issue with overflow here if `bit_size == 128`

compiler/noirc_evaluator/src/ssa/ir/dfg/simplify/binary.rs
```diff
@@ -62,6 +62,8 @@ pub(super) fn simplify_binary(binary: &Binary, dfg: &mut DataFlowGraph) -> Simpl
 
     let lhs_is_one = lhs_value.is_some_and(|lhs| lhs.is_one());
     let rhs_is_one = rhs_value.is_some_and(|rhs| rhs.is_one());
+    let lhs_is_max = lhs_value.is_some_and(|lhs| lhs == lhs_type.max_value().unwrap());
+    let rhs_is_max = rhs_value.is_some_and(|rhs| rhs == rhs_type.max_value().unwrap());
```
TomAFrench, line 66: This would panic in the case where `lhs_value` is a constant `Field`, right?

TomAFrench (commented):
I think this runs into issues with `u128` and `Field` types.

compiler/noirc_evaluator/src/ssa/ir/dfg/simplify/binary.rs
kashbrti: have added the check locally, haven't pushed it yet

compiler/noirc_evaluator/src/ssa/ir/types.rs
kashbrti: thanks that would be helpful

compiler/noirc_evaluator/src/ssa/ir/types.rs
kashbrti: I've already fixed this in [72e82d5](https://github.com/noir-lang/noir/pull/7880/commits/72e82d524431cecde5df0b1f720c42110ae4c9e9)

compiler/noirc_evaluator/src/ssa/ir/types.rs
```diff
@@ -85,6 +85,22 @@ impl NumericType {
...
+    pub(crate) fn max_value(&self) -> Result<FieldElement, String> {
+        match self {
+            NumericType::Unsigned { bit_size } => {
+                if *bit_size == 128 {
+                    Ok(FieldElement::from(u128::MAX))
+                } else {
+                    Ok(FieldElement::from(2u128.pow(*bit_size) - 1))
+                }
+            }
+            NumericType::Signed { bit_size } => Ok(FieldElement::from(2u128.pow(*bit_size) - 1)),
```
kashbrti, line 98: I'm only doing this for `unsigned` `NumericType`s now. should we also have the same for signed variants?

compiler/noirc_evaluator/src/ssa/ir/types.rs
```diff
@@ -85,6 +85,20 @@ impl NumericType {
...
         matches!(self, NumericType::Unsigned { .. })
     }
+
+    pub(crate) fn max_value(&self) -> Result<FieldElement, String> {
+        match self {
+            NumericType::Unsigned { bit_size } => match bit_size {
+                bit_size if *bit_size > 128 => {
+                    Err("Cannot get max value for unsigned type: bit size is greater than 128"
+                        .to_string())
+                }
```
TomAFrench, line 95: This seems like a kludge to get around the fact that `Arbitrary` is generating invalid bitsizes. Can you create an issue for us to restrict the allowable values for `NumericType` to be u8, u16, etc.?

TomAFrench (approved):
Test failure is a flake.

Kinda sad how we're handling invalid bitsizes being generated inside of `max_value`, but we can handle that in a followup.

compiler/noirc_evaluator/src/ssa/ir/types.rs
kashbrti: Yep, will do that. I was also surprised that we're creating a `u129` `NumericType`.

compiler/noirc_evaluator/src/ssa/ir/dfg/simplify.rs
```diff
@@ -465,4 +465,124 @@ mod tests {
         ";
         assert_normalized_ssa_equals(ssa, expected);
     }
+
+    #[test]
+    fn simplifies_or_when_one_side_is_all_1s_u128() {
```
kashbrti, line 470: @TomAFrench I added these manually for every type (u128, u64, u32, u16, u8). Is there a cleaner way I could do this?

TomAFrench: You can template the SSA and insert the differences for the different types

```rust
let test_cases = [("u8", "255"), ...];

for (type, constant) in test_cases {
    let src = format!(src_template, type, constant);
    let expected = format!(expected_template, type, constant);
    let ssa: Ssa = Ssa::from_str_simplifying(src).unwrap(); 
}
```

with the templates being something like:

```
acir(inline) pure fn main f0 {
   b0(v1: {type}):
        v2 = or {type} {constant}, v1
        return v2
 }
```

```
acir(inline) pure fn main f0 {
    b0(v1: {type}):
        return {type} {constant}
}
```

kashbrti: gottcha, thanks.

kashbrti: loving how annoying Rust can be with these :D 
The compiler was complaining that it needs a `string literal` as an input of `format!`. 
But then I couldn't convert the `&str` to a literal as well because it was complaining it doesn't know the 
size at comptime. had to use the `.replace` method to work around it.

TomAFrench:
Looks good!

PR #7890, opened 2025-04-02
feat(fuzz): AST function body

### LINKED ISSUE #7879: Arbitrary AST for function body
Generate function bodies with arbitrary `Expressions`:
* Access variables in `Context::globals` 
* Call other functions based on `Context::function_declarations`
* Maintain local variables scoped to blocks
* Generate random expressions of a certain type by looking for compatible local variables, including by using accessors or casting, falling back to generating literals if nothing else works
* Maintain a reverse index of which variable can produce what kind of type, so that we can jump to random variables in a search tree, before we have to exhaustively generate all potential ways to derive a given type from a certain variable
* Generate random binary expressions by: 1) pick a binary operation 2) pick a compatible type we can produce 3) pick a corresponding pair
* Maintain a configurable budget for the size of the function which decreases according to the size of the randomly created blocks/expressions

aakoshh:
# Description

## Problem\*

Resolves #7879

## Summary\*

Implements random function generation. 

- [x] Keep track of which type can be generated from which variable
- [x] Choose a random variable to produce a given type

Expressions to use:
- [x] `Ident`
- [x] `Index`
- [x] `Cast `
- [x] `Tuple`
- [x] `ExtractTupleField`
- [x] `Block`
- [x] `Semi`
- [x] `Unary`
- [x] `Binary`
- [x] `For`
- [ ] `Loop` with `Break` and `Continue`
- [ ] `While`
- [x] `If`
- [ ] `Match`
- [ ] `Call`
- [x] `Let`
- [ ] `Constrain`
- [x] `Assign`
- [x] `Clone`
- [x] `Drop`

Deferred for future PRs are:
* `Loop` and `While` because they can be tricky to make sure they actually exit
* `Call`, which will require looking for matching function signatures
* `Match` 
* `Constrain`, which will cause assertion failures to compare in the executor

### Testing

I'm running the following command to try to find issues:

```bash
cargo +nightly fuzz run acir_vs_brillig -- -max_total_time=60 -max_len=1048576
```

Bugs discovered so far:
* https://github.com/noir-lang/noir/issues/7952
* https://github.com/noir-lang/noir/issues/7961
* https://github.com/noir-lang/noir/issues/7964
* https://github.com/noir-lang/noir/issues/7965
* https://github.com/noir-lang/noir/issues/7973
* https://github.com/noir-lang/noir/issues/7975

## Additional Context

Example of a generated AST:

```console
❯ cargo run -q -p noir_ast_fuzzer --example sample

global $g0 = 5470048429820336031;
global $g1 = (-3670721263457880079, [-5749924319054596519], [false, true, false, true]);
global $g2 = [-4560164315696530988];
fn main$f0(a$l0: i64) -> [bool; 4] {
    if (G_B$g1.1[0] != (a$l0 + 4974860642070860469)) {
        if G_B$g1.2[(if G_B$g1.2[3] {
            3532461583
        } else {
            2588378952
        } % 4)] {
            let mut b$1 = false;
            for idx_c$2 in -6873237555381424684 .. -6873237555381424682 {
                for idx_d$3 in 1373333312 .. 1373333321 {
                    b$l1 = {
                        if (if true {
                            if b$l1 {
                                ((a$l0 as i128) + (idx_c$l2 as i128))
                            } else {
                                ((idx_c$l2 as i128) + (G_A$g0 as i128))
                            }
                        } else {
                            G_A$g0
                        } > (-1190579296 << 199)) {
                            {
                                idx_d$l3.drop();
                                let e$4 = [];
                            }
                        } else {
                            ()
                        };
                        ((a$l0 as i128) > (a$l0 as i128))
                    };
                    b$l1 = (!false);
                };
                let mut f$5 = b$l1;
            };
            b$l1 = b$l1;
        } else {
            ()
        };
        G_B$g1.2
    } else {
        [false, true, false, false]
    }
}



```

## Documentation\*

Check one:

# PR Checklist\*


tooling/ast_fuzzer/src/program/mod.rs
```diff
@@ -252,14 +266,36 @@ impl Context {
...
+    let mut name = Vec::new();
+    let start = if is_global { 65 } else { 97 };
+    loop {
+        let i = id % 26;
+        name.push(char::from(start + i as u8));
+        id -= i;
+        if id == 0 {
+            break;
+        }
+        id -= 26;
```
rkarabut, line 282: Should be `id /= 26` I think?

aakoshh: Ah yes, thanks, I thought that made more sense as well 🙏

rkarabut (approved):

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -121,37 +154,562 @@ impl<'a> FunctionContext<'a> {
...
+    ///
+    /// This will always succeed, because we can always return a literal expression.
+    fn gen_expr(
+        &mut self,
+        u: &mut Unstructured,
+        typ: &Type,
+        max_depth: usize,
+        flags: Flags,
+    ) -> arbitrary::Result<Expression> {
+        let mut freq = Freq::new(u, &self.ctx.config.expr_freqs)?;
```
jewelofchaos9, line 200: I'm not sure the frequency approach is correct. It might prevent the fuzzer from mutating correctly, Even if it finds a case with new coverage, that case may be skipped due to frequency. By the way, according to @Rumata888 , we want to use every test generated by the fuzzer to the maximum

aakoshh: Do you mean that it's not correct to use frequencies to choose what kind of expression or statement to generate?

jewelofchaos9: Yes

aakoshh: Or the way I implemented it is not correct (which I commented about, I first wanted it to be faster than correct).

Can you explain what you mean by "Even if it finds a case with new coverage, that case may be skipped due to frequency."? First I thought you meant minimisation. 

The frequencies seemed inevitable, e.g. they similarly appear in [noir smith](https://github.com/FuzzingLabs/aztec_fuzzing/blob/0216a5864fd0b00068e213f9e080922c1b899ea2/noir_fuzzers/noir_smith/src/functions/function.rs#L342). 

> we want to use every test generated by the fuzzer to the maximum

How do you define a "generated test" in this case? Unlike noir smith I tried not to enumerate [all possible ways](https://github.com/FuzzingLabs/aztec_fuzzing/blob/0216a5864fd0b00068e213f9e080922c1b899ea2/noir_fuzzers/noir_smith/src/variables/bloc_data.rs#L88) to generate a random value for example, only to discard all but one of them. I could think of an approach to enumerate all potential ASTs by visiting every option, like a BFS traversal, instead of starting from scratch, but I'm not sure it's feasible as the `Unstructured` data is limited.

jewelofchaos9: Im sorry, I may have misunderstood.

let's assume that there is a bug that allows dividing by a constant 0 in Field terms in brillig.

If we assume that the fuzzer found such `Unstructed` that there is a constant folding (for example binary `let b = a - a`), The fuzzer remembers the flow of this execution, saves it, and tries to mutate this input so as to preserve the flow. When it mutates this `Unstructed` it seems to me that the following can happen:

1) `x` in frequency will change and break `Freq::passed`;
2) `lhs_type`, `rhs_type`,  in `FunctionContext::gen_binary` will change.

In this case, the fuzzer may not get to the `let b = a - a; let c = a/b`.

If this really works the way I understand it, then the effect on the resulting program, depending on the input of the fuzzer, will be almost random, which is bad because the fuzzer must collect the coverage and work with it.

As far as I understand, `noir_smith` does not collect coverage and runs really random programs, which is much less efficient.

```
we want to use every test generated by the fuzzer to the maximum
```
I'm sorry about that, I didn't understand your code well enough when I wrote this.

jewelofchaos9: It seems that the same `Unstructed` is used too often, which seems to lead to an avalanche effect even when changing one bit in `Ustructed`

aakoshh: Right, I understand now. Indeed it is very difficult to predict how the AST will change if the random input data is mutated by the fuzzer. Like, I'm not making use of `Unstructured::size_hint` or `Unstructured::arbitrary_len` to scale the size based on what the fuzzer might be trying to minimise. 

I suppose it also depends on how it mutates the input. In your example if it generated `let b = a - a; let c = a / b;`, then it must have gone into `gen_block`, where it chose to generate at least two statements, and then chose `gen_let` statement for both, and then chose `gen_binary` for twice, and then it chose `gen_expr_from_vars`, etc.  Then we'll also have had to generate an `InputMap` that resulted in this flow being hit. If it wanted to minimise this flow, like, preserve these two statements, but get rid of unrelated loops and functions, it would be highly unlikely to succeed. If it just changed some bits, it might turn `let c = a / b` into `let c = a / 2`. 

I think we would need a completely different strategy for generating ASTs in a structure preserving way. I read about some generator that took working pieces for code and recombined them in arbitrary ways. In my top-down approach I don't see that happening, but perhaps an alternative would be to build expressions bottom-up and finding ways to plug them together.

aakoshh: Perhaps another way of achieving more stable swathes of code is to generate more functions that will be rolled into each other in the SSA. Maybe reserve ranges in the random input for each function, so the effects of mutation are localised. Say we have 1MB randomness and we generate 10 functions, then we can divide the 1MB into 10x 100KB, and if the fuzzer is mutating in the range of 200-300KB, which is assigned to function 2, and it changes something that would generate a 1 line function instead of a 25 line function, then it wouldn't have an effect on functions 3..10.

jewelofchaos9: Is it possible to describe AST with some linear set of instructions? For example [insert_var, insert_global_var, start_if_block...] and just use arbitrary vector of them?

aakoshh: I don't know, I haven't thought about that, sounds like an interesting idea. It would resemble more of the state machine testing that I'm used to, picking one random operation at a time, accumulating the state, resulting in a transcript by the end.

I'm sure that the instructions to generate a particular AST can be linearised; for example if I capture the logs of what my generator is doing, it can be replayed in that order to generate that AST by folding over the instructions, and maintaining a cursor into the data structure being built. I'm not sure though how it would work when the instruction list it's arbitrarily mutated, minimised, etc.

rkarabut: I think we should go with this for now just for random code generation, but definitely figure out mutating a linearised representation of the generated AST later.

aakoshh: I thought that maybe it's easier to achieve the sort of controlled changes in the AST if we generate an AST first, and then do random mutations on it, which is what one of the fuzz targets is supposed to do. We can do a few less or more mutations easier than generate similar trees from scratch. But I don't know how we can convince the fuzzer to keep the base AST unchanged; we could 1) write to a file and then read it for mutations, or 2) use a seed in the beginning to grow the tree, then the rest of the Unstructured to do mutations. 

We'll can also generate multiple random inputs for the same AST before generating a different one, but that's a separate issue from randomisation of the AST itself. 

On a different note, I the building up of the AST from linear instructions reminded me of [zippers](https://github.com/polytypic/fastener), but I didn't get a coherent train of thought out of it.

aakoshh:
@jewelofchaos9 FYI:

```console 
$ cargo +nightly fuzz run init_vs_final -- -max_total_time=60 -max_len=1048576
...
Done 42633 runs in 61 second(s)
```

tooling/ast_fuzzer/src/program/func.rs
aakoshh: https://github.com/noir-lang/noir/issues/7979

PR #7909, opened 2025-04-03
fix: always use u32 to index arrays in stdlib

asterite:
# Description

## Problem

Part of #7795

## Documentation

Check one:

# PR Checklist


TomAFrench:
The regressions in `regression_5252` in ACIR seems to be due to the fact that the additional range opcodes are preventing some witnesses from being removed due to them also being used in a range constraint.

TomAFrench:
Closing as poseidon is no longer in stdlib.

PR #7911, opened 2025-04-03
chore: replace `PrintOutput` with usage of `std::io::Write`

TomAFrench:
# Description

## Problem\*

Resolves comment in https://github.com/noir-lang/noir/pull/7758#discussion_r2013922009

## Summary\*

This PR removes the `PrintOutput` enum with a generic `std::io::Write` implementation. This then unifies ignoring the output, piping it into stdout and writing it to a `String` to be used by `nargo`.

At some layers of the foreign call stack I've used a trait object as we need to be able to handle passing different writers conditionally.

## Documentation\*

Check one:

# PR Checklist\*


tooling/nargo/src/ops/test.rs
```diff
@@ -68,17 +69,27 @@ where
...
                         .is_ok_and(|var| &var == "true");
 
-                let mut foreign_call_log = ForeignCallLog::from_env("NARGO_TEST_FOREIGN_CALL_LOG");
+                let writer: Box<dyn std::io::Write> =
+                    match std::env::var("NARGO_TEST_FOREIGN_CALL_LOG") {
+                        Err(_) => Box::new(std::io::empty()),
+                        Ok(s) if s == "stdout" => Box::new(std::io::stdout()),
+                        Ok(s) => Box::new(
+                            OpenOptions::new()
+                                .create_new(true)
```
aakoshh, line 78: I think this means `examples/oracle_transcript/log_and_exec_transcript.sh` should be modified to delete any existing file, otherwise it will fail to create it. 

The previous version was probably equivalent to `.create(true).truncate(true)`, so if there were multiple tests matched then the last one would win, but now the second test will fail, which is potentially safer in a way that it forces the user to run 1 test at a time.

aakoshh (approved):
Nice 👍

PR #7843, opened 2025-03-28
fix!: better error message when cannot infer generic numeric type

asterite:
~Blocked on noir-bignum getting a new release, then updating dependencies to use this new release.~

Blocked on noir-bignum dependencies (and Aztec-Packages, I think) updating to bignum 0.7.0.

# Description

## Problem

Resolves #7617

## Summary

Now this code:

```noir
pub fn foo<let N: u32>() -> BoundedVec<Field, N> {
    BoundedVec::new()
}

fn main() {
    let _ = foo();
}
```

Gives this error:

```
error: Type annotation needed
  ┌─ src/main.nr:6:13
  │
6 │     let _ = foo();
  │             --- Could not determine the value of the generic argument `N` declared on the function `foo`
```

I think it's fine that the error shows up there because, for example, one can specify the type of the let variable to fix this (the issue suggested showing the error on `BoundedVec::new()` instead).

This PR improves a bit how unbound numeric generics are shown in inlay hints.

Before:

image

After:

image

## Additional Context

I know there's #7889 but this also solves the issue of `default_type` being wrong for `Numeric`, and also the fix for #7889 will be a bit different.

Also marked as a breaking change because previously unused numeric generics didn't trigger the error if they were completely unused.

## Documentation

Check one:
- [x] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #7914, opened 2025-04-03
feat: show which generic argument cannot be inferred when it's on the…

asterite:
… associated type

# Description

## Problem

Follow-up to https://github.com/noir-lang/noir/pull/7839

## Summary

I noticed that the error message can still not be clear if the generic is on the associate type. For example:

```noir
pub struct Foo<T> {}

impl<T> Foo<T> {
    fn one() {}
}

fn main() {
    Foo::one();
}
```

Before:

```
error: Type annotation needed
  ┌─ src/main.nr:8:10
  │
8 │     Foo::one();
  │          --- Could not determine type of generic argument
```

Now:

```
error: Type annotation needed
  ┌─ src/main.nr:8:10
  │
8 │     Foo::one();
  │          --- Could not determine the type of the generic argument `T` declared on the struct `Foo`
```

## Additional Context

Eventually all of these errors will be moved to the frontend but fixing those here and adding tests now will help move all of these cases in one go to the frontend, plus it helps for debugging https://github.com/noir-lang/noir/pull/7843

## Documentation

Check one:

# PR Checklist


asterite:
I need to regenerate stderr but this is ready for review.

TomAFrench (approved):

Issue #7915, opened 2025-04-03
Variable type in `Monomorphizer::unpack_tuple_pattern` is incorrect


jfecher:
### Aim

When we destructure patterns in a let binding:
```noir
let (a, b) = (1, "");
```
We lower this code into (roughly):
```noir
{
    let _ = (1, "");
    let 0 = _.0; // yes, we use a numeral as the name of this variable
    let 1 = _.1;
    (0, 1) // referring to the variables here
}
```

### Expected Behavior

This lowering should have correct & consistent types internally (and clearer names too, even if it is just changed to `_0`, `_1`).

### Bug

https://github.com/noir-lang/noir/blob/master/compiler/noirc_frontend/src/monomorphization/mod.rs#L902 in that snippet we set the type of the collection variable to a different value each loop: it is set to the result type of the extract, not the collection type.

So essentially internally right now we have:
```noir
{
    let _ = (1, "");
    let 0 = (_ : Field).0;
    let 1 = (_ : str<0>).1;
    (0, 1) // referring to the variables here
}
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #7916, opened 2025-04-03
fix: Fix type of internal variable created while destructuring tuples and structs during monomorphization

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/7915

## Summary\*

Found while working on the `-Zownership` work. This doesn't cause any known bugs on master but was panicking in that PR when I was trying to unwrap a tuple type and discovered the tuple didn't have a tuple type.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/monomorphization/mod.rs
+++ b/compiler/noirc_frontend/src/monomorphization/mod.rs
@@ -856,7 +856,7 @@ impl<'interner> Monomorphizer<'interner> {
             HirPattern::Mutable(pattern, _) => self.unpack_pattern(*pattern, value, typ),
             HirPattern::Tuple(patterns, _) => {
                 let fields = unwrap_tuple_type(typ);
-                self.unpack_tuple_pattern(value, patterns.into_iter().zip(fields))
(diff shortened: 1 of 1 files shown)


asterite:
Is there a test that can be written for this?

compiler/noirc_frontend/src/monomorphization/mod.rs
```diff
@@ -895,8 +896,8 @@ impl<'interner> Monomorphizer<'interner> {
             let mutable = false;
             let definition = Definition::Local(fresh_id);
             let name = i.to_string();
-            let typ = Self::convert_type(&field_type, location)?;
 
+            let typ = Self::convert_type(tuple_type, location)?;
```
asterite, line 900: Could this be moved outside of the loop?

jfecher: Yes and no - we'd need another `Location` to pass since this one is on the tuple element

jfecher:
> Is there a test that can be written for this?

Not that I know of at least. It's an internal variable that doesn't exist beforehand and variable types aren't displayed in the hir printer.

PR #7922, opened 2025-04-04
chore: create new crate for running `test_programs`

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR moves the `execute.rs` integration test out of `tooling/nargo_cli` and into `test_programs`. The main benefit of this is that `nargo_cli`'s incremental build cache does not need to be aware of `test_programs` at all. This addresses [this TODO](https://github.com/noir-lang/noir/blob/19285a74ec30cef13fb1ee6eda40af7c35f170bb/tooling/nargo_cli/build.rs#L30-L31) and should result in faster compilation times in general.

I've also added usage of insta.rs rather than using the `stdout.txt` and `stderr.txt` files that were written into the test directories.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Parking this for a bit, happy for someone else to take a look to see if they can handle this cleanly.

The issue is that we need to find a way to ensure that `nargo` is built when running the tests on `test_programs`.  This needs to:

- handle being packed into a nextest archive so CI is happy.
- gracefully handle the case where we're running the tests  but `nargo_cli` hasn't been built yet.

PR #7941, opened 2025-04-07
fix: wrapping mul support for u128

### LINKED ISSUE #7528: `std::wrapping_mul`  return incorrect values for u128 arguments
Noticed this while removing the `U128` struct and porting this test over:

https://github.com/noir-lang/noir/blob/f0c1c7ba3feaaa28905bb0813d14acc6673ca797/noir_stdlib/src/uint128.nr#L534-L566

```noir
    #[test]
    fn test_wrapping_mul() {
        let zero: u128 = 0;
        let one: u128 = 1;
        let two_pow_64: u128 = 0x10000000000000000;
        let u128_max: u128 = 0xffffffffffffffffffffffffffffffff;

        // 1*0==0
        assert_eq(zero, wrapping_mul(zero, one));

        // 0*1==0
        assert_eq(zero, wrapping_mul(one, zero));

        // 1*1==1
        assert_eq(one, wrapping_mul(one, one));

        // 0 * ( 1 << 64 ) ==  0
        assert_eq(zero, wrapping_mul(zero, two_pow_64));

        // ( 1 << 64 ) * 0 == 0
        assert_eq(zero, wrapping_mul(two_pow_64, zero));

        // 1 * ( 1 << 64 ) == 1 << 64
        assert_eq(two_pow_64, wrapping_mul(two_pow_64, one));

        // ( 1 << 64 ) * 1 == 1 << 64
        assert_eq(two_pow_64, wrapping_mul(one, two_pow_64));

        // ( 1 << 64 ) * ( 1 << 64 ) == 1 << 64
        assert_eq(zero, wrapping_mul(two_pow_64, two_pow_64));
        // -1 * -1 == 1
        assert_eq(one, wrapping_mul(u128_max, u128_max));
    }
```

This test now fails which is pretty clear from the implementation of `wrapping_mul` as it discounts the potential for the multiplication to overflow the field modulus

```
pub fn wrapping_mul<T>(x: T, y: T) -> T
where
    T: AsPrimitive<Field>,
    Field: AsPrimitive<T>,
{
    AsPrimitive::as_(x.as_() * y.as_())
}
```

`wrapping_add` and `wrapping_sub` seem to be ok however. There's potential for the magic constant in `wrapping_sub` being an issue but it seems fine to me.

```noir
pub fn wrapping_sub<T>(x: T, y: T) -> T
where
    T: AsPrimitive<Field>,
    Field: AsPrimitive<T>,
{
    //340282366920938463463374607431768211456 is 2^128, it is used to avoid underflow
    AsPrimitive::as_(x.as_() + 340282366920938463463374607431768211456 - y.as_())
}

```

guipublic:
# Description

## Problem\*

Resolves #7528 

## Summary\*
Wapping_mul now supports u128 by doing multiplication using u64 limbs, ignoring the  terms above 2^{128} and casting the result back to the original type.

## Documentation\*

Check one:
- [] No documentation needed.

# PR Checklist\*


noir_stdlib/src/lib.nr
```diff
@@ -110,7 +110,43 @@ where
...
+    let (y_lo, y_hi) = split_into_64_bit_limbs(y);
+    // Multiplication using the limbs:(x_lo + 2**64*x_hi)*(y_lo + 2**64*y_hi)=x_lo*y_lo+...
+    // and skipping the terms over 2**128
+    // Working with u64 limbs ensures that we cannot overflow the field modulus.
+    let low = x_lo as Field * y_lo as Field;
+    let lo = low as u64 as Field;
+    let carry = (low - lo) / two_pow_64;
+    let high = x_lo as Field * y_hi as Field + x_hi as Field * y_lo as Field + carry;
+    let hi = high as u64 as Field;
+    (lo + two_pow_64 * hi)
```
kashbrti, line 149: Couldn't there be carries from `high` that affect the low? If the high is larger than `2^64`, there would be a carry that would wrap unto the final result. but it seems you're just truncating the "overflow" part of `high` by casting to u64.

guipublic:
I am wondering why is there a regression on circuit size tests. The u128 cast should be compile-time resolved and not induce any overhead for u64 or below. I will look into this.

noir_stdlib/src/lib.nr
```diff
@@ -110,7 +110,43 @@ where
...
+/// It will fail if the number is more than 128 bits
+fn split_into_64_bit_limbs(x: Field) -> (u64, u64) {
+    // Safety: the limbs are constrained below
+    let (x_lo, x_hi) = unsafe { split64(x) };
+    assert(x as Field == x_lo as Field + x_hi as Field * two_pow_64);
+    (x_lo, x_hi)
+}
+
+#[field(bn254)]
+pub fn wrapping_mul128(x: Field, y: Field) -> Field {
```
TomAFrench, line 138: This shouldn't need to be `pub` it looks like.

TomAFrench (commented):
This is looking like it should be implemented as a `WrappingMul` trait tbh as we're trying to push all of the native types down the same codepath which doesn't make fit properly for most of them.

TomAFrench:
I expect the constrained-unconstrained boundary is preventing type knowledge from being used here.

noir_stdlib/src/lib.nr
guipublic: Done

noir_stdlib/src/lib.nr
```diff
@@ -107,14 +107,9 @@ where
 
 pub fn wrapping_mul<T>(x: T, y: T) -> T
```
TomAFrench: Let's deprecate this function

noir_stdlib/src/lib.nr
```diff
@@ -152,6 +147,42 @@ pub fn wrapping_mul128(x: Field, y: Field) -> Field {
 #[builtin(as_witness)]
 pub fn as_witness(x: Field) {}
 
+trait Wapping {
```
TomAFrench, line 150:
```suggestion
trait WrappingMul {
```

TomAFrench, line 150: Let's move these away from the crate root, they'd fit better somewhere in the `ops` module.

guipublic:
I added a trait to perform the wrapping_mul128 only for u128.

noir_stdlib/src/lib.nr
guipublic: Should we have a wrapping trait and have wrapping_add,...wrapping_mul functions in the trait?
It feel a bit strange to handle the multiplication differently (in term of function / trait)

noir_stdlib/src/lib.nr
guipublic: Done

noir_stdlib/src/lib.nr
TomAFrench: I would have separate `WrappingAdd` `WrappingMul`, etc. traits to mirror the arithmetic traits. We can't guarantee that all types which implement one operation will implement the rest.

noir_stdlib/src/lib.nr
```diff
@@ -105,17 +105,48 @@ where
...
+    // Multiplication using the limbs:(x_lo + 2**64*x_hi)*(y_lo + 2**64*y_hi)=x_lo*y_lo+...
+    // and skipping the terms over 2**128
+    // Working with u64 limbs ensures that we cannot overflow the field modulus.
+    let low = x_lo as Field * y_lo as Field;
+    let lo = low as u64 as Field;
+    let carry = (low - lo) / two_pow_64;
+    let high = x_lo as Field * y_hi as Field + x_hi as Field * y_lo as Field + carry;
+    let hi = high as u64 as Field;
+    (lo + two_pow_64 * hi)
+}
```
TomAFrench, line 148: These shouldn't be at the crate root either.

guipublic: Done

docs/docs/noir/concepts/data_types/integers.md
```diff
@@ -112,19 +112,19 @@ fn main() {
...
-Although integer overflow is expected to error, some use-cases rely on wrapping. For these use-cases, the standard library provides `wrapping` variants of certain common operations:
+Although integer overflow is expected to error, some use-cases rely on wrapping. For these use-cases, the standard library provides `wrapping` variants of certain common operations via Wrapping traits in std::ops
 
 ```rust
-fn wrapping_add<T>(x: T, y: T) -> T;
-fn wrapping_sub<T>(x: T, y: T) -> T;
-fn wrapping_mul<T>(x: T, y: T) -> T;
+ fn wrapping_add(self, y: Self) -> Self;
+ fn wrapping_sub(self, y: Self) -> Self;
+ fn wrapping_mul(self, y: Self) -> Self;
```
TomAFrench, line 120:
```suggestion
fn wrapping_add(self, y: Self) -> Self;
fn wrapping_sub(self, y: Self) -> Self;
fn wrapping_mul(self, y: Self) -> Self;
```

docs/docs/noir/concepts/data_types/integers.md
```diff
@@ -112,19 +112,19 @@ fn main() {
 
 ### Wrapping methods
 
-Although integer overflow is expected to error, some use-cases rely on wrapping. For these use-cases, the standard library provides `wrapping` variants of certain common operations:
+Although integer overflow is expected to error, some use-cases rely on wrapping. For these use-cases, the standard library provides `wrapping` variants of certain common operations via Wrapping traits in std::ops
```
TomAFrench, line 115:
```suggestion
Although integer overflow is expected to error, some use-cases rely on wrapping. For these use-cases, the standard library provides `wrapping` variants of certain common operations via Wrapping traits in `std::ops`
```

noir_stdlib/src/ops/arith.nr
```diff
@@ -346,3 +348,227 @@ impl Neg for i64 {
...
+    Field: AsPrimitive<T>,
+{
+    //340282366920938463463374607431768211456 is 2^128, it is used to avoid underflow
+    AsPrimitive::as_(x.as_() + 340282366920938463463374607431768211456 - y.as_())
+}
+
+fn wrapping_mul_hlp<T>(x: T, y: T) -> T
+where
+    T: AsPrimitive<Field>,
+    Field: AsPrimitive<T>,
```
TomAFrench, line 565: oof, that's a doozy. We shouldn't allow this in the typesystem and we should have a `FromPrimitive` trait imo

TomAFrench: Ah, I must have never run into a situation where I needed this before.

PR #7943, opened 2025-04-07
chore: separate benchmarking from github actions more

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

I'm wanting to rework the benchmarking to make it easier to extend without an explosion of jobs.

The new workflow is that we have a single job which contains two scripts:
- `generate_benchmark_data` which downloads the project and runs it to provide raw data and save to a directory
- `process_benchmark_data` which consumes this folder to generate a JSON payload which can be easily merged across different projects to feed the benchmark action.

This makes it easier for us to run benchmarks locally, add new benchmarks, debug benchmarks, etc.

## Additional Context

I've removed the internal compilation and execution benchmarks as they haven't been published in some time and if we reintegrate them in future we can reuse the generic benchmarking mechanism.

## Documentation\*

Check one:

# PR Checklist\*


jfecher (approved):
LGTM

PR #7946, opened 2025-04-07
Get Random Blackbox Implementation

ozgurarmanc:
# Description
This PR will be about Get Random blackbox function implementation. More details can be found [here](https://github.com/gnosisguild/enclave/issues/276).

## Problem\*
The idea behind a blackbox get_random() function is that should handle the underlying complexity of hashing the transcript thus far and appropriately constraining the subsequent phase(s) such that the Noir developer doesn't need to understand what happens under the hood / where the (pseudo)randomness comes from and can treat the randomness much as they would in any other high-level language.

Resolves 
https://github.com/gnosisguild/enclave/issues/276 but to fully resolve this issue we will have to implement this blackbox function to backend as well.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
btw, I'd recommend using an oracle function to inject randomness rather than a blackbox function. Oracles are the standard way to inject non-determinism whereas blackbox functions are expected to be deterministic.

ozgurarmanc:
> btw, I'd recommend using an oracle function to inject randomness rather than a blackbox function. Oracles are the standard way to inject non-determinism whereas blackbox functions are expected to be deterministic.

Hey Tom, thanks for helping out. I misclicked it and opened the PR here instead of our repository. Sorry for that and we are using Blackbox exactly for that reason. I know the name is a bit confusing for the context but we want randomness for a challenge value which should be deterministic

PR #7970, opened 2025-04-08
fix: replace one more path pattern when producing stderr.txt

asterite:
# Description

## Problem

Should unblock https://github.com/AztecProtocol/aztec-packages/pull/12995

## Summary

It seems tests in Aztec-Packages run from the root directory via something like `target/debug/deps/execute-16a6287606ef3464 test_name` so the currently directory isn't inside `nargo_cli`, which is what I assumed would happen. So this PR takes that into account.

## Additional Context

I was able to reproduce this locally and checked that this change fixes the issue according to how tests run in Aztec-Packages CI.

Tests could theoretically run from other directories and maybe it will fail too. I'm not sure what the logic would be to clear up these paths, maybe we'd need to use a regex... but if tests in Aztec-Packages are always going to run from the root directory then this should be good enough.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/nargo_cli/tests/execute.rs
+++ b/tooling/nargo_cli/tests/execute.rs
@@ -63,7 +63,15 @@ mod tests {
     }
 
     fn delete_test_program_dir_occurrences(string: String, test_program_dir: &Path) -> String {
+        // Assuming `test_program_dir` is "/projects/noir/test_programs/compile_failure/some_program"...
+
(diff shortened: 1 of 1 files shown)


TomAFrench (approved):
:shrug: on implementation as this isn't too much extra code compared to regex and it works.

Thanks!

PR #7974, opened 2025-04-08
fix: don't solve `Self` in trait impl method via trait method lookup

### LINKED ISSUE #7968: Default implementations of trait methods apply `Self` from one impl of the trait onto a different impl
Consider the program below (which is a massively simplified instance of the bignum library:

```noir
pub trait MagicNumber {
    fn magic_value() -> Field;
    fn from_magic_value() -> Self {
        Self::from_value(Self::magic_value())
    }

    fn from_value(value: Field) -> Self;
}

mod foo {
    use super::MagicNumber;

    pub struct Foo {
        inner: Field,
    }

    impl MagicNumber for Foo {

        fn magic_value() -> Field {
            42
        }

        fn from_value(inner: Field) -> Self {
            Self { inner }
        }
    }

}

mod bar {
    use super::MagicNumber;

    pub struct Bar {
        inner: Field,
    }

    impl MagicNumber for Bar {

        fn magic_value() -> Field {
            421
        }

        fn from_value(inner: Field) -> Self {
            Self { inner }
        }
    }
}

```

I have a trait `MagicNumber` which types can implement. To make it easier to create the magic value of various types `MagicNumber` implements a default implementation of `from_magic_value` to construct `Self` from this.

I would expect this to pass type checking as `Self` in the context of `from_magic_value` should be able to be able to resolved to `Foo` and `Bar` at the correct times. Instead I get the error below:

```
$ nargo check
error: expected type Bar, found type Foo
  ┌─ src/main.nr:3:30
  │
3 │     fn from_magic_value() -> Self {
  │                              ---- expected Bar because of return type
4 │         Self::from_value(Self::magic_value())
  │         ------------------------------------- Foo returned here
  │

```

asterite:
# Description

## Problem

Resolves #7968

## Summary

There were two issues here:
1. We have special logic for solving `Self::method_name` to find `methd_name` via the trait `Self` refers to. However, there's no need to do this if we are inside a trait impl as we know what's `Self` (and in fact going via the trait is what caused the issue... though to be honest I feel that maybe should have worked in the first place, so there might be another bug somewhere).
2. The above point fixes the original issue, but I noticed that it doesn't work if we are inside a trait impl for a primitive type. In that case the regular path lookup won't work, which is why we represent `u32::method_name` with a different AST node (a `TypePath`). So, if we need to solve `Self::method_name` and `Self` is a primitive type, we solve it as a `TypePath`.

## Documentation

Check one:

# PR Checklist


TomAFrench (approved):
LGTM

jfecher (approved):
:+1:

Issue #7949, opened 2025-04-07
Add snapshot tests for the build artifacts as produced by `test_programs`


TomAFrench:
We should be committing snapshots of the build artifacts as a source of truth so that different architectures can assert that they reproduce that snapshot.

PR #7981, opened 2025-04-09
feat: add `--program-output-file` to `nargo compile`

asterite:
# Description

## Problem

Towards #7949

## Summary

When test programs run for a given options matrix they all generate artifacts with the same final destination. It's all good when `nargo compile` runs because there's a lock, but... after `nargo compile` finishes the lock is gone. What I need to do for snapshots is to read that artifact, but another `nargo compile` run in the same directory might start running, so we'd need an extra lock to prevent this.

Alternatively, we could allow specifying the output file and have each option in the matrix use a different file. Then there's no conflict. I think this is simpler. Plus, in any case, I wanted to store these separate artifacts in a directory that's not `target` because that's in `.gitignore` (I plan to use `test_target` for these artifacts).

## Additional Context

I made this a hidden option because I don't know if it's generally useful.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_driver/src/lib.rs
+++ b/compiler/noirc_driver/src/lib.rs
@@ -25,7 +25,7 @@ use noirc_frontend::monomorphization::{
 use noirc_frontend::node_interner::{FuncId, GlobalId, TypeId};
 use noirc_frontend::token::SecondaryAttribute;
 use std::collections::HashMap;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use tracing::info;
 
 mod abi_gen;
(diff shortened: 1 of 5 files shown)


TomAFrench:
We already allow setting the target directory, is this necessary compared to having a separate target directory for each matrix configuration?

asterite:
I'm trying the snapshot thing with `--target-dir` but it's not as straight-forward as I thought. The main issue is that I don't know what the final name of the JSON file is going to be. In most cases it matches the name of the directory, but it actually comes from the name inside `Nargo.toml`, and right now there are some mismatches. But even then, if we are compiling a workspace the target name comes from the default member, which means we have to parse the `Nargo.toml` file to extract this information. If we provide the filename to write to we then don't need to do all of this extra work.

(but I'll try work on parsing the `Nargo.toml` file for now until we decide otherwise)

asterite:
Alternative I can try to find the single JSON file in that directory so I don't have to figure out the name.

TomAFrench:
Yeah, in a few cases like this I've just used the `./target/*.json` glob which has worked nicely.

PR #7983, opened 2025-04-09
chore: Support printing the AST without variable IDs

### LINKED ISSUE #7978: Add option not to print IDs in AST
https://github.com/noir-lang/noir/pull/7890 added some changes to the AST to print the name and type of globals, but there could be some further changes to make it easier to take a printed AST and turn it into an integration test:
* Print the visibility of the return type from `main`, which is an information only available on the `Program`, not the `Function` that is being printed. Applies to parameters as well, but when I copied the code into an integration test, it always complained about non-public return from `main`. 
* Add an option to the `AstPrinter` to omit the `$f1` and `$l2` style IDs from the output; these have to be removed for the code to be valid Noir code, which is simple to do, but a chore.

aakoshh:
# Description

## Problem\*

Resolves #7978 

## Summary\*

Adds some customisation options to the `AstPrinter` to try to keep the printed code closer to Noir by not printing e.g. `a$l1` for a local variable but simply `a`. This saves us the trouble of having to get rid of all those IDs when we're trying to turn an AST printed when the fuzzer failed into a Noir integration test.

The printer now also shows the _main return visibility_ of the program, because Noir always complains if we try to return a value without `pub`. The generator will no longer generate private non-unit return types for `main`.

The PR also changes the display of input values when a program fails to TOML, so it's easier to paste into `Prover.toml`. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3616,6 +3616,7 @@ dependencies = [
  "function_name",
  "fxhash",
  "im",
+ "insta",
  "iter-extended",
  "noirc_arena",
  "noirc_errors",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3302,6 +3302,7 @@ dependencies = [
  "color-eyre",
  "libfuzzer-sys",
  "noir_ast_fuzzer",
+ "noirc_abi",
  "noirc_evaluator",
  "noirc_frontend",
 ]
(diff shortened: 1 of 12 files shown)


rkarabut (approved):

TomAFrench:
@aakoshh I've switched over the relevant test to use insta.rs which makes it easier to see what the differences are and also easier to update.

The change is in https://github.com/noir-lang/noir/pull/7984 which has this PR as a base. I've also committed the updated snapshot in a separate commit for easier review.

PR #7984, opened 2025-04-09
chore: switch to using insta for snapshot testing

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR switches us over to using insta.rs for snapshot testing the rewritten noir generated from ast.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/monomorphization/tests.rs
+++ b/compiler/noirc_frontend/src/monomorphization/tests.rs
@@ -116,17 +116,17 @@ fn simple_closure_with_no_captured_variables() {
 
     let program = get_monomorphized!(src, Expect::Success).unwrap();
     insta::assert_snapshot!(program, @r"
-    fn main$f0() -> Field {
-        let x$0 = 1;
-        let closure$3 = {
-            let closure_variable$2 = {
-                let env$1 = (x$l0);
+    fn main$f0() -> pub Field {
+        let x$l0 = 1;
+        let closure$l3 = {
+            let closure_variable$l2 = {
+                let env$l1 = (x$l0);
                 (env$l1, lambda$f1)
             };
             closure_variable$l2
         };
         {
-            let tmp$4 = closure$l3;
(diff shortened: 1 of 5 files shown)


aakoshh (approved):
Thanks!

compiler/noirc_frontend/src/monomorphization/tests.rs
```diff
@@ -121,23 +114,24 @@ fn simple_closure_with_no_captured_variables() {
...
     "#;
 
-    let expected_rewrite = r#"fn main$f0() -> Field {
-    let x$0 = 1;
-    let closure$3 = {
-        let closure_variable$2 = {
-            let env$1 = (x$l0);
-            (env$l1, lambda$f1)
+    let program = get_monomorphized!(src, Expect::Success).unwrap();
+    insta::assert_snapshot!(program, @r"
```
aakoshh, line 118: I've never used `insta`; a quick glance at the docs suggested that it would store some kind `.snap` file in a `snapshots` directory, and later compare against that. But here we're comparing against some hardcoded string, so I wonder if `similar_asserts::assert_eq!` we use to compare SSAs would also be an option? It provides diff-like presentation as well. Is `insta::assert_snapshot!` better in some way?

TomAFrench: If I just did `insta::assert_snapshot(program)` then we'd write out the value of `format!("{display}")` into a .snap file and compare against that as you said.

Here I'm using the inline snapshot feature (showm by preceding the string by an `@`), this means that whenever we update the snapshot's expected result it's automatically written into the following string literal. 

The main benefit here over just using `similar_asserts::assert_eq!` is when we intentionally modify the code in a way which would change the snapshots. Rather than needing to run each test and copy the new result from the terminal into the codebase one by one, we instead just run the tests and can review and update the snapshots by just running `cargo insta review`.

PR #7985, opened 2025-04-09
feat: add `loop` generator to AST fuzzer 

### LINKED ISSUE #7927: Generate arbitrary Loop statements
Follow up https://github.com/noir-lang/noir/pull/7890 with arbitrary `Loop` statements. 

The difficulty with this is that we should also make sure that the loop exits with `Break` at some point during its execution, so there has to be some condition in it which will be true at some point.

rkarabut:
# Description

## Problem\*

Resolves #7927 

## Documentation\*

Check one:

# PR Checklist\*


tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -729,4 +733,90 @@ impl<'a> FunctionContext<'a> {
...
+        })]);
+
+        let inner_block = self.gen_block(u, &Type::Unit)?;
+        let inner_block = expr::prepend_block(
+            inner_block,
+            vec![Expression::Semi(Box::new(Expression::If(If {
+                condition: Box::new(expr::binary(
+                    Expression::Ident(idx_ident.clone()),
+                    BinaryOp::Equal,
+                    expr::u32_literal(self.ctx.config.max_loop_size as u32),
```
aakoshh, line 768: You could choose a random limit, up to the maximum size.

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -729,4 +733,90 @@ impl<'a> FunctionContext<'a> {
...
+            id: idx_id,
+            mutable: true,
+            name: idx_name,
+            expression: Box::new(expr::u32_literal(0)),
+        })]);
+
+        let inner_block = self.gen_block(u, &Type::Unit)?;
+        let inner_block = expr::prepend_block(
+            inner_block,
+            vec![Expression::Semi(Box::new(Expression::If(If {
```
aakoshh, line 764: I think `Semi` is not necessary, it's just used when an expression is a last one in the block, to not treat it as a return value.

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -729,4 +733,90 @@ impl<'a> FunctionContext<'a> {
...
+        let inner_block = expr::extend_block(
+            inner_block,
+            vec![Expression::Assign(Assign {
+                lvalue: LValue::Ident(idx_ident.clone()),
+                expression: Box::new(Expression::Clone(Box::new(expr::binary(
+                    Expression::Ident(idx_ident.clone()),
+                    BinaryOp::Add,
+                    expr::u32_literal(1),
+                )))),
+            })],
```
aakoshh, line 784: Perhaps this and the `if` above could be made a bit less verbose by some helper in the `expr` module?

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -729,4 +733,90 @@ impl<'a> FunctionContext<'a> {
...
+                consequence: Box::new(Expression::Block(vec![Expression::Break])),
+                alternative: None,
+                typ: Type::Unit,
+            })))],
+        );
+        let inner_block = expr::extend_block(
+            inner_block,
+            vec![Expression::Assign(Assign {
+                lvalue: LValue::Ident(idx_ident.clone()),
+                expression: Box::new(Expression::Clone(Box::new(expr::binary(
```
aakoshh, line 779: Do we have to clone here?

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -729,4 +733,90 @@ impl<'a> FunctionContext<'a> {
...
+                    Expression::Ident(idx_ident.clone()),
+                    BinaryOp::Add,
+                    expr::u32_literal(1),
+                )))),
+            })],
+        );
+
+        let outer_block =
+            expr::extend_block(outer_block, vec![Expression::Loop(Box::new(inner_block))]);
```
aakoshh, line 788: It looks like you could just create a block out of the _let_ and the _loop_ expressions, there is no real reason to wrap, unwrap, extend and wrap again 👀

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -729,4 +733,90 @@ impl<'a> FunctionContext<'a> {
...
+    fn gen_loop(&mut self, u: &mut Unstructured) -> arbitrary::Result<Expression> {
+        // Declare break index variable visible in the loop body.
+        let idx_type = types::U32;
+        let idx_id = self.next_local_id();
+        let idx_name = format!("idx_{}", make_name(idx_id.0 as usize, false));
+        let idx_ident =
+            expr::ident_inner(VariableId::Local(idx_id), true, idx_name.clone(), idx_type);
+
+        // Add a scope which will hold the index variable.
+        //self.locals.add(idx_id, false, idx_name.clone(), idx_type.clone());
```
aakoshh, line 747: Now that we discussed that this guard variable will not be added to the scope, could you update the comment to explain the reason for that?

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -729,4 +733,90 @@ impl<'a> FunctionContext<'a> {
...
+
+        // Initialize index to 0
+        let outer_block = Expression::Block(vec![Expression::Let(Let {
+            id: idx_id,
+            mutable: true,
+            name: idx_name,
+            expression: Box::new(expr::u32_literal(0)),
+        })]);
+
+        let inner_block = self.gen_block(u, &Type::Unit)?;
```
aakoshh, line 761: If we use `gen_block` to generate the body of the loop, then I think the `self.locals.enter()` and `self.locals.exit()` isn't even needed, since we're not adding anything to the scope here and `gen_block` does it on its own.

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -729,4 +733,90 @@ impl<'a> FunctionContext<'a> {
...
+                    expr::u32_literal(self.ctx.config.max_loop_size as u32),
+                )),
+                consequence: Box::new(Expression::Block(vec![Expression::Break])),
+                alternative: None,
+                typ: Type::Unit,
+            })))],
+        );
+        let inner_block = expr::extend_block(
+            inner_block,
+            vec![Expression::Assign(Assign {
```
aakoshh, line 777: This could be in the `else` of the check above, so the two related statements are close together in the AST.

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -543,6 +543,10 @@ impl<'a> FunctionContext<'a> {
             return self.gen_for(u);
         }
 
+        if freq.enabled_when("loop", self.budget > 1) {
```
aakoshh, line 546:
```suggestion
        if freq.enabled_when("loop", self.budget > 1 && self.unconstrained()) {
```
At least I think we can't unroll `loop` in ACIR.

aakoshh (commented):
Cool! 

One thing that's missing is that now that you can generate `Loop`, you should also be able to generate `Break` and `Continue`.

It could be controlled for example by adding two `bool` flags: 
* `FunctionContext::in_loop` could enable those statements, set to `true` in `gen_loop`. 
* a `let was_in_loop = self.in_loop;` at the top of `gen_loop` and a `self.in_loop = was_in_loop;` at the bottom to restore it, so the break/continue can appear anywhere, and doesn't get reset in nested loops until all outer loops are exited.

tooling/ast_fuzzer/src/program/func.rs
rkarabut: Yes, true

tooling/ast_fuzzer/src/program/func.rs
rkarabut: I don't really know, an assign always looked like that when I looked through the AST samples I generated from actual code, and I didn't investigate further

aakoshh: Interesting. I assumed `Clone` was related to ownership, moving, and other experimental features, like `Drop`. I wonder if SSA optimisations remove any overhead. In any case we shouldn't have to clone the result of an `Add` 🤔

tooling/ast_fuzzer/src/program/func.rs
aakoshh: Ah, sorry, I thought you meant adding the guard counter to `for`. 

Well, I think we do need to add the index variable of the `for` loop to the scope, because we do want to be able to use that variable in the loop body, to access arrays or whatever. But you're right we should probably not be able to drop it. Hasn't caused an issue so far, so maybe it's a no-op for an integer.

rkarabut: But wouldn't it be possible for the inner generated code to make `for` infinite by messing with the index anyway?

rkarabut: Like setting it to a fixed value inside the range

aakoshh: Whereas this counter is a behind-the-scenes one that we don't necessarily want to influence the random program. Nothing bad would happen if it did, apart from the dropping, but in theory a `loop` doesn't need one, so it's better if the AST generator ignores it.

aakoshh: > But wouldn't it be possible for the inner generated code to make for infinite by messing with the index anyway?

Ah, good point. I declared the index variable as immutable, because it's governed by the for loop itself. Not sure what the compiler would say if we declared it as mutable and then tried to assign to it in the body. Apparently rust [does allow](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=736cc5a343a961bd46f6ba54d52fd2c8) assigning to the variable if it's declared to be mutable, but the effect doesn't persist beyond the body of the loop, it gets reassigned. 

Noir doesn't seem to allow the `for mut i in ...` syntax though.

aakoshh:
If you add `Break` and `Continue` then I think they should also be enabled in the `for` loop, if the code is `unconstrained`. I didn't realise that I could have done that (probably).

tooling/ast_fuzzer/src/program/func.rs
rkarabut: Done

tooling/ast_fuzzer/src/program/func.rs
rkarabut: Clone removed

tooling/ast_fuzzer/src/program/func.rs
rkarabut: Removed

tooling/ast_fuzzer/src/program/func.rs
rkarabut: Done

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -729,4 +733,83 @@ impl<'a> FunctionContext<'a> {
...
+        ));
+
+        // Put everything into if/else
+        let inner_block = Expression::Block(vec![expr::if_else(
+            expr::binary(
+                Expression::Ident(idx_ident.clone()),
+                BinaryOp::Equal,
+                expr::u32_literal(self.ctx.config.max_loop_size as u32),
+            ),
+            Expression::Block(vec![Expression::Break]),
```
aakoshh, line 778: Why surround with a block?

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -729,4 +733,83 @@ impl<'a> FunctionContext<'a> {
...
+            id: idx_id,
+            mutable: true,
+            name: idx_name,
+            expression: Box::new(expr::u32_literal(0)),
+        })];
+
+        // Get the randomized loop body
+        let inner_block = self.gen_block(u, &Type::Unit)?;
+        let Expression::Block(mut inner_stmts) = inner_block else {
+            unreachable!("generated a non-block expression: {}", inner_block)
```
aakoshh, line 762: Technically `gen_block` is currently allowed to return a non-block expression if it's called when there is no budget, which could be the case after decreasing it on line #749.

aakoshh: I would suggest not unwrapping here, but instead just adding it to a sub block as it is:
```
Block(vec![
  Let(idx, 0), 
  Loop(
    Block(vec![      
      If(
        Binary(idx, Eq, max_size),
        Break,
        Block(vec![
          Assign(idx, Binary(idx, Add, 1)),
          inner_block
        ])
      )
    ])
  )
])
```

tooling/ast_fuzzer/src/program/func.rs
rkarabut: Yeah I guess unnecessary

rkarabut:
> Cool!
> 
> One thing that's missing is that now that you can generate `Loop`, you should also be able to generate `Break` and `Continue`.
> 
> It could be controlled for example by adding two `bool` flags:
> 
>     * `FunctionContext::in_loop` could enable those statements, set to `true` in `gen_loop`.
> 
>     * a `let was_in_loop = self.in_loop;` at the top of `gen_loop` and a `self.in_loop = was_in_loop;` at the bottom to restore it, so the break/continue can appear anywhere, and doesn't get reset in nested loops until all outer loops are exited.

Let's do this in the `while` PR then, for every kind of a loop

tooling/ast_fuzzer/src/program/func.rs
rkarabut: I don't think it's good here semantically, as the guard index is explicitly applied over the generator scope and only serves to break out if nothing else does

aakoshh: Okay, fair enough 👍

rkarabut:
Hmm agreed

aakoshh (approved):
Smashing!

PR #7986, opened 2025-04-09
chore: add snapshot tests for the build artifacts as produced by test_programs

### LINKED ISSUE #7949: Add snapshot tests for the build artifacts as produced by `test_programs`

asterite:
# Description

## Problem

Resolves #7949

## Summary

Only adds snapshots for `compile_success_empty` and `execution_success`, but let me know if we should do this for other test programs too.

## Documentation

Check one:

# PR Checklist


TomAFrench:
Cool, I have some thoughts but will review this tomorrow.

asterite:
Asking for a review just to signal it's ready for review (CI is finally green)

asterite:
Strange! There is a `let _ = fs::remove_dir_all(target_dir);` line I added right before generating the snapshot. Maybe you stopped the `cargo test` run and it left some files because it was in the middle of running a test? Or did it leave all of these directories?

TomAFrench:
Yeah, I think this is from `cargo test` failing fast and that results in leftover directories not being cleaned up.

asterite:
Let me try using `Drop` to see if it happens even if you stop a test in the middle with ctrl+c...

asterite:
I ended up adding these dirs to .gitignore.

TomAFrench (approved):

TomAFrench:
I'm going to add some redaction for the hash emitted as I'm concerned about churn on that. We can always unredact it in future.

asterite:
Just in case: I see a bunch of `.span.new` that shouldn't have been committed (in case CI passes)

Issue #7965, opened 2025-04-08
ACIR "Index out of Bounds" and unsolvable; Brillig works


aakoshh:
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/7890

Similar to https://github.com/noir-lang/noir/issues/7952 and https://github.com/noir-lang/noir/issues/7961

The following program fails in ACIR, passes in Brillig:

```
global G_A: [[bool; 0]; 2] = [[], []];

fn main(a: bool, b: [(u1, bool, u1, str<2>); 4]) -> pub [bool; 0] {
    if (a) {
        G_A[((((b[0].0 as u8) + (b[0].0 as u8)) as u32) % 2)]
    } else {
        G_A[((((b[0].0 as u8) + (b[0].0 as u8)) as u32) % 2)]
    }
}
```

with these inputs:
```
a = 1
b = [[0, 1, 0, "\\\t"], [0, 1, 1, "mp"], [0, 0, 0, "\t/"], [0, 0, 0, "ab"]]
```

### Expected Behavior

The program should return `[]`.

### Bug

In `Brillig` it correctly returns `[]`. In ACIR it fails with `Index out of bound`, but unlike the other bugs in the summary, it fails with both `a=0` and `a=1`. 
```console
bash-5.2$ cargo run -q -p nargo_cli -- execute --force-brillig
[regression_fuzz_testing] Circuit witness successfully solved
[regression_fuzz_testing] Witness saved to target/regression_fuzz_testing.gz
[regression_fuzz_testing] Circuit output: Vec([])
bash-5.2$ cargo run -q -p nargo_cli -- execute 
bug: Assertion is always false
  ┌─ src/main.nr:7:9
  │
7 │         G_A[((((b[0].0 as u8) + (b[0].0 as u8)) as u32) % 2)]
  │         ----------------------------------------------------- As a result, the compiled circuit is ensured to fail. Other assertions may also fail during execution
  │
  = Call stack:
    1. src/main.nr:7:9

error: Assertion failed: Index out of bounds
  ┌─ src/main.nr:5:9
  │
5 │         G_A[((((b[0].0 as u8) + (b[0].0 as u8)) as u32) % 2)]
  │         -----------------------------------------------------
  │
  = Call stack:
    1. src/main.nr:5:9


```

This is the final SSA:

```
After Dead Instruction Elimination (3rd):
g0 = make_array [] : [u1; 0]
g1 = make_array [] : [u1; 0]
[...]

Issue #7952, opened 2025-04-07
ACIR "Index out of bounds" depending on condition; Brillig works


aakoshh:
### Aim

Trying to get https://github.com/noir-lang/noir/pull/7890 to pass by running `cargo +nightly fuzz run acir_vs_brillig -- -max_total_time=60 -max_len=1048576` seems to have found an issue where ACIR behaves different from Brillig. 

Based on what it printed I replicated it as an integration test and managed to reproduce the behaviour with the following program:

```rust
fn main(a: [[u32; 0]; 1], b: bool) -> pub [u32; 0] {
    if (b) {
        a[0]
    } else {
        a[0]
    }
}
```

and the following `Prover.toml`:
```toml
a = [[]]
b = 1
```

### Expected Behavior

ACIR and Brillig should behave the same way and return `[]`. 

### Bug

ACIR returns a runtime assertion error, while Brillig successfully returns `[]`:

```console
❯ nargo execute 
error: Assertion failed: Index out of bounds
  ┌─ /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/regression_zero_size_array_index/src/main.nr:3:9
  │
3 │         a[0]
  │         ----
  │
  = Call stack:
    1. /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/regression_zero_size_array_index/src/main.nr:3:9

Failed assertion

❯ nargo execute --force-brillig
[regression_ast_fuzz] Circuit witness successfully solved
[regression_ast_fuzz] Circuit output: Vec([])
[regression_ast_fuzz] Witness saved to /Users/aakoshh/Work/aztec/noir/test_programs/execution_success/regression_zero_size_array_index/target/regression_ast_fuzz.gz
```

If I change the inputs to contain `b=0` then ACIR passes as well. It also passes if the size of `[u32; 0]` array isn't zero. Since it passes with `b = 0`, I reckon ACIR should handle zero-sized arrays.

Executing with the `--show-ssa` options shows the following:
```
❯ nargo execute --show-ssa
After Initial SSA:
acir(inline) fn main f0 {
  b0(v0: [[u32; 0]; 1], v1: u1):
    jmpif v1 then: b2, else: b1
  b1():
    v4 = array_get v0, index u32 0 -> [u32; 0]
    jmp b3(v4)
  b2():
    v5 = array_get v0, index u32 0 -> [u32; 0]
    jmp b3(v5)
  b3(v2: [u32; 0]):
    return v2
}

[...]

PR #7995, opened 2025-04-09
fix(ssa): Remove OOB checks inserted during DIE

### LINKED ISSUE #7975: Brillig "Index out of bounds" and unsolvable; ACIR works
### Aim

Discovered in https://github.com/noir-lang/noir/pull/7890

Probably same as https://github.com/noir-lang/noir/issues/7952 https://github.com/noir-lang/noir/issues/7961 and https://github.com/noir-lang/noir/issues/7965 but I thought I'd capture it as this time it's Brillig that fails.

Program:
```rust
fn main(b: [[i1; 1]; 1]) -> pub [[i1; 1]; 1] {
    if (b[0][0] == b[0][0]) {
        b
    } else {
        b
    }
}
```

Input:
```
b = [[0]]
```

### Expected Behavior

Behaviour should be identical, returning `[[0]]`. 

### Bug

```console
error: Assertion failed: Index out of bounds
  ┌─ src/main.nr:2:9
  │
2 │     if (b[0][0] == b[0][0]) {
  │         ----
  │
  = Call stack:
    1. src/main.nr:2:9

Failed assertion
[regression_fuzz_testing] Circuit output: Vec([Vec([Field(0)])])
```

Final ACIR SSA:

```
acir(inline) pure fn main f0 {
  b0(v0: [[i1; 1]; 1]):
    return v0
}
```

Final Brillig SSA:
```
  b0(v0: [[i1; 1]; 1]):
    return v0
}
```

Unlike in the other bugs, the constraint does not depend on any input, it asserts that `0 == 1`, so it will always fail.

### To Reproduce

See above.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+cdd4da85b7713e64aa8ff5e712c312978fb4d02e (git version hash: 2bfb47fba0c541e2f2d53a1e40ad04d39a9763c8, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #7975, #7952, #7965

The snippet from issue #7975 has this SSA:
```
After Brillig Array Get Optimizations:
  b0(v0: [[i1; 1]; 1]):
    v2 = array_get v0, index u32 1 -> [i1; 1]
    return v0
}

After Dead Instruction Elimination (2nd):
  b0(v0: [[i1; 1]; 1]):
    return v0
}
```
We are inserting an OOB check as DIE is seeing an unsafe array access post the Brillig array gets optimization. OOB checks are now being inserted in #7827. We would expect the OOB on this array to be simplified upon its initial insertion and for an OOB check to not be inserted. 

Each of the other failures in the linked issues, had an SSA in the same theme as above.

## Summary\*

We added https://github.com/noir-lang/noir/pull/5691 several months ago in place of https://github.com/noir-lang/noir/pull/5610 as to avoid performance regressions. However, we now are adding OOB checks for arrays during SSA gen directly in https://github.com/noir-lang/noir/pull/7827 as it is needed for deeper inconsistencies with how ACIR/Brillig handles unchecked ops. Also, at the time of #5691 we did not even have Brillig bytecode reports, but now that we have included loop optimizations and cross-block constant folding so adding these OOB checks on all arrays should not instill too much of a cost. 

https://github.com/noir-lang/noir/pull/7827 also was missing access checks on certain array assignments, so I added those here.

## Additional Context

Brillig bytecode regressions:
https://github.com/noir-lang/noir/pull/8017 has been opened to outline the regressions on this PR and a possible solution. I don't think the regressions are as bad as they look if we chose not to go with #8017. The sample program is very small and is only regressing due to no longer triggering a loop unroll. A larger loop already would not have had triggered an unroll, and perhaps with more aggressive loop unrolling in the future it will not be an issue.

Circuit size regressions:
https://github.com/noir-lang/noir/pull/7995#issuecomment-2794958967

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
> **brillig_rc_regression_6123_inliner_max**	+121 ❌	**+605.00%**
> **brillig_rc_regression_6123_inliner_zero**	+121 ❌	**+605.00%**

Investigating these regressions. It looks like I am adding a repeat bounds check that is not being simplified out and is preventing an unrolling of the loop that occurs on master.

vezenovm:
> Program	ACIR opcodes (+/-)	%	Circuit size (+/-)	%
> **regression_7062**	+4 ❌	**+44.44%**	+2,746 ❌	**+6102.22%**
> **array_dynamic_main_output**	+4 ❌	**+12.50%**	+2,746 ❌	**+2096.18%**

These circuit size regressions are expected. For example looking at `regression_7062`:
```noir
fn main(value: Field, index: u32) {
    let mut args = &[0, 1];
    args[index] = value;
    // Safety: n/a
    unsafe { store(args) };
    // Dummy test to remove the 'underconstraint bug'
    assert(args[0] + args[1] != 0);
}

pub unconstrained fn store(_: [Field]) {}
```
We are going to add an index check for `args[index] = value;` in this PR that was not being generated previously.

guipublic (approved):

PR #8002, opened 2025-04-10
chore: clippy fixes

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

Clippy woke up today and decided to be a pedant.

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
Interesting that clippy is looking at doc comments. I also didn't know you could use a type without `{ ... }` in a pattern match as a short syntax to that.

TomAFrench:
This is only because `Allocate` is a unit type so there's no fields to ignore.

PR #8003, opened 2025-04-10
chore: add benchmark for ACVM arithmetic solver

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR adds an example benchmark for testing the speed of the ACVM when solving pure arithmetic circuits. I've chosen an example circuit which does not allow for any batch inversion as a baseline case.

We should also add some benchmarks which will show improvement once batch inversion is added but I'll leave that as an exercise to the reader.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
I've added another benchmark with a perfectly batchable program (no opcodes depend on each others' results). I've also set this up so we monitor this execution speed in CI.

noirwhal (approved):

PR #8006, opened 2025-04-10
feat: attribute locations

asterite:
# Description

## Problem

Part of #7540

## Summary

Some attribute errors were shown on the function name because we didn't track locations for all attributes. Now we do so the error is a bit more clearer.

## Additional Context

There's also a small LSP improvement that could be done here but I'll leave it as a follow-up PR for much later.

## Documentation

Check one:

# PR Checklist


jfecher (approved):
:+1:

michaeljklein (approved):
LGTM

PR #2498, opened 2023-08-30
chore: Remove `set_value_from_id` and `resolve` in the `DataFlowGraph`

jfecher:
# Description

## Problem\*

Resolves (none)

## Summary\*

This is another experimental change in the SSA IR. This one is aimed at reducing the surface for bugs by removing `DataFlowGraph::set_value_from_id` and subsequently the need to call `DataFlowGraph::resolve` for each value as well. This has historically been a source of bugs when one forgets to call resolve. Now that the mem2reg no longer relies on `set_value_from_id`, the only passes that still use it are cfg simplification which just needed a small change, and defunctionalization which still needs a larger change.

## Documentation

- [ ] This PR requires documentation updates when merged.

  

  - [ ] I will submit a noir-lang/docs PR.

  

  - [ ] I will request for and support Dev Rel's help in documenting this PR.

  
  

## Additional Context

# PR Checklist\*

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #8019, opened 2025-04-10
fix(ssa): Map terminator instructions after constant folding 

### LINKED ISSUE #7964: Compiler crashes with --force-brillig, passes with ACIR
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/7890

The following program crashes during SSA generation when it's compiled to Brillig:

```rust
global G_A: bool = false;
fn main(a: Field) -> pub Field {
    if (
        (
            if (!G_A) {
                if (!G_A) {
                    (a as u128)
                } else {
                    (a as u128)
                }
            } else {
                (a as u128)
            }
                <= (a as u128)
        ) as bool
    ) {
        (-(-a))
    } else {
        (-(-a))
    }
}
```

### Expected Behavior

It should compile.

### Bug

```console
bash-5.2$ cargo run -q -p nargo_cli -- compile --force-brillig
Message:  index out of bounds: the len is 1 but the index is 7
Location: compiler/noirc_evaluator/src/ssa/opt/simplify_cfg.rs:224

bash-5.2$ 

```

### To Reproduce

See above.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+cdd4da85b7713e64aa8ff5e712c312978fb4d02e (git version hash: 54d7cf108f636068b93285e5f5319c7457e30b26, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
…ant folding

# Description

## Problem\*

Resolves #7964 

I noticed the supplied program in the issue would not fail when I had `--show-ssa` activated. Looking at the SSA we can see the jmpif condition is simplified from instructions to a constant value during constant folding:
```
After Purity Analysis (2nd):
g0 = u1 0

  b0(v1: Field):
    v3 = truncate v1 to 128 bits, max_bit_size: 254
    v4 = cast v3 as u128
    v5 = truncate v1 to 128 bits, max_bit_size: 254
    v6 = cast v5 as u128
    v7 = lt v6, v4
    v8 = not v7
    jmpif v7 then: b1, else: b2
  b1():
    v12 = sub Field 0, v1
    v13 = sub Field 0, v12
    jmp b3(v13)
  b2():
    v10 = sub Field 0, v1
    v11 = sub Field 0, v10
    jmp b3(v11)
  b3(v2: Field):
    return v2
}

After Constant Folding:
g0 = u1 0

  b0(v1: Field):
    v3 = truncate v1 to 128 bits, max_bit_size: 254
    v4 = cast v3 as u128
    jmpif u1 0 then: b1, else: b2
  b1():
    v8 = sub Field 0, v1
    v9 = sub Field 0, v8
    jmp b3(v9)
  b2():
    v6 = sub Field 0, v1
    v7 = sub Field 0, v6
    jmp b3(v7)
  b3(v2: Field):
    return v2
}
```
The fact that the instruction ID is missing made me think to check whether we were resolving the terminators of blocks during constant folding.

## Summary\*

I simply added logic similar to `map_terminator_in_place` that exists on the `FunctionInserter`. I had tried to make some sort of the shared method on `BasicBlock` or `Function` but `resolve_cache` needs to borrow the dfg which runs into issues with the borrow checker. Thus, just setting and mapping the terminator during the pass felt simplest.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
@@ -234,9 +234,9 @@ impl<'brillig> Context<'brillig> {
         &mut self,
         function: &mut Function,
         dom: &mut DominatorTree,
-        block: BasicBlockId,
+        block_id: BasicBlockId,
     ) {
-        let instructions = function.dfg[block].take_instructions();
+        let instructions = function.dfg[block_id].take_instructions();
 
         // Default side effect condition variable with an enabled state.
         let mut side_effects_enabled_var =
@@ -246,12 +246,22 @@ impl<'brillig> Context<'brillig> {
             self.fold_constants_into_instruction(
                 function,
                 dom,
-                block,
+                block_id,
                 instruction_id,
                 &mut side_effects_enabled_var,
             );
         }
-        self.block_queue.extend(function.dfg[block].successors());
+
+        let cache = self.get_constraint_map(side_effects_enabled_var);
+        // Map a terminator in place, replacing any ValueId in the terminator with the
+        // resolved version of that value id from the simplification cache's internal value mapping.
+        let mut terminator = function.dfg[block_id].take_terminator();
+        terminator.map_values_mut(|value| {
+            Self::resolve_cache(block_id, &mut function.dfg, dom, self.get_constraint_map(side_effects_enabled_var), value)
+        });
+        function.dfg[block_id].set_terminator(terminator);
+
+        self.block_queue.extend(function.dfg[block_id].successors());
     }
 
     fn fold_constants_into_instruction(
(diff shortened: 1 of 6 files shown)


aakoshh (approved):
I don't fully understand the reasons, but if it fixes the crash, let's get it in!

asterite:
> Wish there was a good, clean solution to the issue of needing to resolve ValueIds. Wish we could remove it entirely.

When a ValueID needs to be replaced with another one, would replacing it in all subsequent instructions solve this? Is the issue that doing this is cumbersome and also probably not very efficient?

jfecher:
> When a ValueID needs to be replaced with another one, would replacing it in all subsequent instructions solve this? Is the issue that doing this is cumbersome and also probably not very efficient?

Correct. We have a couple passes that'd need to be rewritten which rely on this - notably defunctionalization. There's been several attempts to remove this feature in the past but they tended to lead to slowdowns and/or a large increase in code across several SSA passes. E.g. https://github.com/noir-lang/noir/pull/2498

asterite:
> notably defunctionalization

I just checked and I can't find calls to `set_value_from_id` there so maybe that's no longer the case?

In any case, I was always curious about this, and a lot of time passed since the last try, so I'll try it again (but in small PRs, trying to remove it from each pass in turn)

jfecher:
> I just checked and I can't find calls to set_value_from_id there so maybe that's no longer the case?

Ah, looks like I removed them in https://github.com/noir-lang/noir/pull/7222 along with the problematic mutating of existing value's types. I did think I touched this code since Dec 2023.

> In any case, I was always curious about this, and a lot of time passed since the last try, so I'll try it again (but in small PRs, trying to remove it from each pass in turn)

Good luck!

PR #8021, opened 2025-04-11
feat: add `while` generator to AST fuzzer

### LINKED ISSUE #7928: Generate arbitrary While statement
Follow up https://github.com/noir-lang/noir/pull/7890 with arbitrary `While` statement. 

Similar to #7927 we have to make sure that the condition at some point becomes false, and we don't end up in an infinite, or even unreasonably long running loop where the Brillig VM gets stuck.

rkarabut:
# Description

## Problem\*

Resolves #7928, also introduces break/continue statements in all types of loops

## Documentation\*

Check one:

# PR Checklist\*


tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -637,6 +660,16 @@ impl<'a> FunctionContext<'a> {
         Ok(Some(Expression::Assign(Assign { lvalue, expression: Box::new(expr) })))
     }
 
+    /// Break out of a loop.
+    fn gen_break(&mut self) -> Expression {
+        Expression::Break
+    }
```
aakoshh, line 666: I reckon we can just return `Expression::Break` in `gen_stmt` directly, and then these to `gen_` methods don't have to be the exceptions that don't take an `Unstructured`.

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -715,7 +748,9 @@ impl<'a> FunctionContext<'a> {
         // Decrease budget so we don't nest for loops endlessly.
         self.decrease_budget(1);
 
+        self.in_loop = true;
         let block = self.gen_block(u, &Type::Unit)?;
+        self.in_loop = false;
```
aakoshh, line 753: This isn't entirely correct: setting it back to `false` would destroy any previous information about what happened _before_ `gen_for` was invoked, so in a case like this:

```rust
loop {
   ...
   for i in 0 .. 10 {
      if ... { 
        continue;
      }
   }
   // it thinks we're no longer in a loop
   if ... {
     break; // this will not be generated 
   }
}
```

The interpreter has an [elegant way](https://github.com/noir-lang/noir/blob/3d1f8b2639d0b1faf31c26610d5a0d3b1e386b21/compiler/noirc_frontend/src/hir/comptime/interpreter.rs#L1234-L1259) using `std::mem:swap` of doing this.

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -820,3 +909,38 @@ fn test_loop() {
...
+        )
+    );
+
+    assert!(
+        while_code.ends_with(
+            &r#"idx_a$l0 = (idx_a$l0 + 1)
+        }
+    }
+}"#
+            .replace(" ", "")
```
aakoshh, line 943: FYI I changed this in https://github.com/noir-lang/noir/pull/7987 to be right after the `else`. I thought this:

```rust 
if idx == 10 {
  break;
} else {
  idx = idx + 1;
  // stmt 1
  // stmt 2
}
```

looks better than this when we print it:
```
if idx == 10 {
  break;
} else {
  {
    // stmt 1
    // stmt 2
  }
  idx = idx + 1;.
}
```

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -785,6 +818,62 @@ impl<'a> FunctionContext<'a> {
...
+                expr::u32_literal(self.ctx.config.max_loop_size as u32),
+            ),
+            Expression::Break,
+            Expression::Block(inner_stmts),
+            Type::Unit,
+        )]);
+
+        // Generate the `while` condition with depth 1, ensuring there is a variable in scope to use
+        stmts.push(self.gen_let(u, Some(true))?);
+        let condition = self.gen_expr(u, &Type::Bool, 1, Flags::CONDITION)?;
```
aakoshh, line 868: Ah, it took me a while to understand how this variable you declare but don't have an ID for ends up being used in the condition. 

I agree that giving it a chance to use something mutable is nice, but unless there is an actual guarantee that it will use it, I don't think it's much better than just generating a random bool expression. Even with max_depth=1 it might pick up a read-only variable from somewhere else. 

It's also possible that the `gen_let` created a complex type, and the max_depth=1 might prevent us from accessing its field.

I realise that if we don't do anything then the most likely scenario is that we will hit the max count. I think that's fine, though, as with a `loop` we also only have the count, and the chance of an early `break` to get us out.

tooling/ast_fuzzer/src/lib.rs
```diff
@@ -46,11 +46,14 @@ impl Default for Config {
             ("literal", 5),
         ]);
         let stmt_freqs = Freqs::new(&[
-            ("drop", 5),
+            ("drop", 3),
+            ("break", 1),
```
aakoshh, line 50: I would give `break` more chance simply to not spend more time in busy loops than we have to.

tooling/ast_fuzzer/src/lib.rs
```diff
@@ -46,11 +46,14 @@ impl Default for Config {
...
+            ("continue", 1),
             ("assign", 30),
             ("if", 10),
-            ("for", 10),
-            ("loop", 10),
+            ("for", 4),
+            ("loop", 3),
+            ("while", 3),
```
aakoshh, line 56: In ACIR the `loop` and `while` are disabled, so I wouldn't bring down `for` relative to `if`, as unrolling can be tricky, we should have coverage for it.

tooling/ast_fuzzer/src/program/func.rs
rkarabut: Hmm I think it's gonna be fine yeah

tooling/ast_fuzzer/src/lib.rs
rkarabut: Does drop 2 break 2 sound okay?

tooling/ast_fuzzer/src/program/func.rs
rkarabut: Oh yeah right. Fixing

tooling/ast_fuzzer/src/program/func.rs
rkarabut: Just a random bool expression then?

tooling/ast_fuzzer/src/lib.rs
rkarabut: Picking frequencies is getting tricky at this point hm, maybe move loop/for/while into its own set?

tooling/ast_fuzzer/src/program/func.rs
rkarabut: Or maybe introducing guaranteed usage _would_ be better hm

aakoshh: It is tempting to give it a chance to mutate something in the body that has some bearing on the condition, but there is no guarantee that it will. Even if we assign to the variable, it might assign something that never changes. I wouldn't worry about it too much.

tooling/ast_fuzzer/src/lib.rs
aakoshh: I don't know, I just run the `--example sample` thing and look for loops, and breaks. Drop seemed overpowered at some point, but `break` is only enabled in loops, and and it's nice as it gets us out of there so we can test other stuff, so I would give it a generous weight.

tooling/ast_fuzzer/src/lib.rs
aakoshh: I wouldn't complicate it with more, just give them all the same weight so they appear and we're not just full of `if`s. 

But this reminds me of something I saw in QuickCheck or ScalaCheck where it was able to print how many times the various generators were exercised, so we know we actually covered a property. Perhaps we could do something like that, to collect metrics that we can look at that, after generating 100 programs, what is the actual distribution of what we hit.

rkarabut: Created https://github.com/noir-lang/noir/issues/8026 for this, for now setting everything to 10.

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -598,28 +601,59 @@ impl<'a> FunctionContext<'a> {
...
+            return self.gen_while(u);
+        }
+
+        if freq.enabled_when("break", self.in_loop) {
+            return Ok(Expression::Break);
+        }
+
+        if freq.enabled_when("continue", self.in_loop) {
+            return Ok(Expression::Continue);
+        }
```
aakoshh, line 618:
```suggestion
        if freq.enabled_when("break", self.in_loop && self.unconstrained()) {
            return Ok(Expression::Break);
        }

        if freq.enabled_when("continue", self.in_loop && self.unconstrained()) {
            return Ok(Expression::Continue);
        }
```

We can't break out of ACIR loops.

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -598,28 +601,59 @@ impl<'a> FunctionContext<'a> {
...
+
         if freq.enabled_when("call", self.budget > 0) {
             if let Some(e) = self.gen_call(u, &Type::Unit, self.max_depth())? {
                 return Ok(e);
             }
         }
 
+        if freq.enabled("drop") {
```
aakoshh, line 626: This is a duplicate, I moved it up to the top earlier so it doesn't gain weight if other stuff is disabled, because then it starts dropping everything.

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -610,16 +629,25 @@ impl<'a> FunctionContext<'a> {
...
+        self.gen_let(u, None)
     }
 
-    /// Generate a `Let` statement.
-    fn gen_let(&mut self, u: &mut Unstructured) -> arbitrary::Result<Expression> {
+    /// Generate a `Let` statement, optionally requesting mutability.
+    fn gen_let(
+        &mut self,
+        u: &mut Unstructured,
+        mutable: Option<bool>,
```
aakoshh, line 639: Perhaps in the future we can extend this to be able to ask for a specific type.

aakoshh (approved):

PR #8025, opened 2025-04-11
chore: snapshot stdout/stderr with insta

### LINKED ISSUE #8022: Replace usage of `stdout.txt` and `stderr.txt` with insta
@asterite added snapshot testing for stdout/stderr test programs in https://github.com/noir-lang/noir/pull/7713. We've since moved to using insta.rs more generally for snapshot testing so for consistency we should migrate this over to use insta.rs as well.

TomAFrench:
# Description

## Problem\*

Resolves #8022 

## Summary\*

This PR moves the snapshot testing for stdout and stderr for `execution_success` and `compile_failure` tests to use insta. We now snapshot both stdout and stderr by default.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Yeah, it's a bit sad that differing compiler settings result in different output here. I was hoping we could get away with a single output snapshot per test

asterite:
I was thinking that maybe stdout should be the same for all configurations, while stderr could vary. In fact, it could probably be a good idea to check that stdout is the same for all configurations (which we did before this PR). That means that stdout and stderr might need to be split...

asterite:
> while stderr could vary

Maybe we could omit the check that produces those "bugs" here? I think it's the "check underconstrained values". Not sure if these are important for these tests.

TomAFrench:
> I was thinking that maybe stdout should be the same for all configurations, while stderr could vary. In fact, it could probably be a good idea to check that stdout is the same for all configurations (which we did before this PR). That means that stdout and stderr might need to be split...

Yeah, I think that's a good idea. Warnings may vary but the stdout behaviour should be consistent.

TomAFrench:
Looks like there's some non-determinism in the warnings output order (which I thought was fixed earlier?)

asterite:
It seems to. I fixed the order for unused struct, unused function, etc. The code for unused variables is somewhere else... but I just checked and it should be deterministic 🤔

michaeljklein (approved):
LGTM

PR #8028, opened 2025-04-11
feat: use batch inversion when writing to the witness map (WIP)

kashbrti:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*


acvm-repo/acvm/src/pwg/arithmetic.rs
```diff
@@ -1,14 +1,64 @@
...
+        Ok(())
+    }
+}
+
+pub(crate) fn batch_invert<F>(inp_list: &Vec<F>) -> Vec<F>
+where
+    F: AcirField,
+{
+    vec![F::one(); inp_list.len()]
+}
```
TomAFrench, line 53: https://github.com/arkworks-rs/algebra/blob/bf1c9b22b30325ef4df4f701dedcb6dea904c587/ff/src/fields/mod.rs#L353-L356

acvm-repo/acvm/src/pwg/arithmetic.rs
```diff
@@ -23,6 +73,144 @@ pub(crate) enum MulTerm<F> {
...
+                            })
+                        } else {
+                            Ok(())
+                        }
+                    } else {
+                        // normally we would do
+                        // let assignment = -total_sum / (q + b);
+                        // insert_value(&w1, assignment, initial_witness)
+                        // but we want to add this to pending_arithmetic_opcodes
+                        pending_arithmetic_opcodes.add_pending_op(total_sum, q + b, w1);
```
TomAFrench, line 136: Special case which is worth handling, if `q + b == FieldElement::one()` then we can skip the inversion entirely and just write directly into the witness map which will increase chances for successful batching down the line.

kashbrti:
@TomAFrench this pr is still way too far from being ready to review. Would be better to wait a bit before checking it

TomAFrench:
Yep, was just following up providing links to things I had spoken about previously.

acvm-repo/acvm/src/pwg/arithmetic.rs
kashbrti: I noticed we only have Arkworks as a dev dependency. Should I move it to dependencies for the acvm repo?

acvm-repo/acvm/src/pwg/arithmetic.rs
kashbrti: good point. I can try to do this for both 1 and -1

acvm-repo/acvm/src/pwg/arithmetic.rs
TomAFrench: We don't want to have arkworks as a dependency outside of `acir_field`.

Thinking more about the trait constraints involved, a generic `AcirField` implementation can't be passed to this cleanly so we may have to just vendor that code for now.

TomAFrench:
I'm going to close this PR as it looks like we weren't generating constraints which would benefit from it.

PR #8031, opened 2025-04-11
chore: remove unnecessary double compilation

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

After #7986, we're testing for non-determinism by checking against the build artifact snapshot. We can then avoid having to compile every program twice.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_driver/src/lib.rs
+++ b/compiler/noirc_driver/src/lib.rs
@@ -174,10 +174,6 @@ pub struct CompileOptions {
     #[arg(long, default_value = "false")]
     pub pedantic_solving: bool,
 
-    /// Used internally to test for non-determinism in the compiler.
-    #[clap(long, hide = true)]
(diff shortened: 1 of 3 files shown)


TomAFrench:
Asking for a review from @vezenovm as he added this feature originally.

vezenovm (approved):
> we're testing for non-determinism by checking against the build artifact snapshot

Are you just referring to the fact that we are checking snapshots on different architectures already?

TomAFrench:
I mean that even if we're only testing on x86 then compiling once and comparing with the snapshot is equivalent to compiling twice and comparing both results (except that we avoid the work of compiling a second time)

vezenovm:
> compiling once and comparing with the snapshot is equivalent

Ah yes don't even need to be checking against different archs

PR #8037, opened 2025-04-11
chore: don't use `set_value_from_id` in `remove_truncate_after_range_checks`

asterite:
# Description

## Problem

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/dfg.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/dfg.rs
@@ -383,12 +383,12 @@ impl DataFlowGraph {
         self.instructions[id] = instruction;
     }
 
-    /// Replaces all values in the given block with the values in the given map.
+    /// Replaces all values in the given blocks with the values in the given map.
     ///
     /// This method should be preferred over `set_value_from_id` which might eventually be removed.
-    pub(crate) fn replace_values_in_block(
+    pub(crate) fn replace_values_in_blocks(
         &mut self,
-        block: BasicBlockId,
+        blocks: impl Iterator<Item = BasicBlockId>,
         values_to_replace: &HashMap<ValueId, ValueId>,
     ) {
         if values_to_replace.is_empty() {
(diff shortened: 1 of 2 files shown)


TomAFrench:
I haven't thought about this in depth but does this make assumptions that there won't be a reference one of these value ids in a different block? This is possible (blocks can refer to valueids defined in dominators)

asterite:
Aaah... I didn't think value IDs could appear beyond blocks. So we'd need to traverse all instructions in all blocks. I'll change this, thanks!

asterite:
I changed an exiting test to make sure that values are replaced across blocks. This uncovered the issue that the current `remove_truncate_after_range_checks` optimization was only being applied block per block, so if a value had a range check in one block but it was truncated in another block, the optimization wasn't applied.

jfecher:
> I changed an exiting test to make sure that values are replaced across blocks. This uncovered the issue that the current remove_truncate_after_range_checks optimization was only being applied block per block, so if a value had a range check in one block but it was truncated in another block, the optimization wasn't applied.

I haven't reviewed the code yet but since we're dealing with side-effects like this, have you checked that we only replace the instruction in blocks which are dominated by the block the range check was in? E.g. if we have `v0` as a parameter to main but the range check is only within an `if` somewhere we don't want to modify any uses of v0 before that (or after that after the if ends)

asterite:
> have you checked that we only replace the instruction in blocks which are dominated by the block the range check was in?

I thought about it! I think we could mainly do it for performance reasons, to not lookup values in block where we know they won't show up.

However, at least in this pass the values to replace can appear in multiple blocks, so we'd end up potentially visiting a same block multiple times.

An alternative would be to at least keep track of the first block we encounter a value to replace, and then replace from that point onward. But, we could try to improve the performance if we see performance regressions (I also tried to avoid allocating memory here, like using `self.blocks[block_id]` instead of `self[block_id]` which makes Rust later complain with "cannot borrow self as mutable...").

asterite:
I'm already facing some issues with this approach in #8039 but I'll merge this anyway as it's mostly equivalent to the old code.

PR #8038, opened 2025-04-11
chore: migrate recursive proof test to ultrahonk

TomAFrench:
Looking to get a working e2e example of a recursive ultrahonk program being verified. It doesn't need to be done onchain yet but we should have an example on how to do this in JS ([people are coming in to make their own at this point](https://github.com/teddav/noir-recursive)).

Issues I ran into here:
- How big is a UH proof? It's not documented anywhere.
    - I've pulled the new values from `verify_honk_proof` test in aztec-packages.
    - `SIZE_OF_PROOF_IF_LOGN_IS_28` implies that this depends on circuit size (which I remember being a thing at one point). Is there a way I can get bb.js to calculate this for me rather than having to do trial and error?
- How big is a UH verification key in fields? It's not documented anywhere.
- [How do I get the verification key in JS?](https://github.com/AztecProtocol/barretenberg/issues/1337)
- [How do I get the verification key hash in JS?](https://github.com/AztecProtocol/barretenberg/issues/1337)

I threw this together based on the barretenberg readme and pulling code out of `bb.js` when necessary components aren't exposed. However I'm now getting unreachable errors inside `acirWriteVkUltraHonk` for some reason I haven't debugged as I'm blocked on the verification key hash anyway.

@ledwards2225 Are there resources which I'm missing on how to get these values out of barretenberg in JS? Perhaps @saleel can help do a pass on UX of bb.js for recursion considering Noirhack is coming up next week?


saleel:
I think `UltraHonkBackend::getVerificationKey()` + `deflattenFields(vk)` should solve the problem. `0x0` can be passed as the value for vk_hash for now. I can continue on this PR and push to the same branch if its ok.

We normally point to [noir-bb-bench](https://github.com/AztecProtocol/aztec-packages/tree/v0.84.0/yarn-project/noir-bb-bench) as an example for recursion.
After 0.85.0 the proof size of recursive and non-recursive proof will be the same, so this will be much cleaner. I think the methods in bb.js can be improved as well. I will take a look this week.

TomAFrench:
Hey, help on getting this PR over the line would be greatly appreciated. :pray: 

> I think UltraHonkBackend::getVerificationKey() + deflattenFields(vk) should solve the problem. 

I think that depends on whether all elements in the verification key are fields though? I remember with UP there was tight packing of some smaller types so I wasn't sure that we could just chunk it into 32 bytes segments.

In any case, having a dedicated "convert this binary blob of a vk into fields" function in bb.js  (even if it's just a `deflattenFields(vk)` under the hood) would be amazing and take all guesswork out of the user code.

> After 0.85.0 the proof size of recursive and non-recursive proof will be the same, so this will be much cleaner. I think the methods in bb.js can be improved as well. I will take a look this week.

Awesome, thank you so much. :)

On a related note, we should probably remove the `generateRecursiveProofArtifacts` method as it's causing more confusion that good now.

saleel:
> On a related note, we should probably remove the generateRecursiveProofArtifacts method as it's causing more confusion that good now.

Yes, this one should be removed. This was never implemented for Honk, but another method `generateProofForRecursiveAggregation()` which does the same thing was added on. We can use this method to fix this PR, but looks like the method got removed with 0.85.0. So I think we need to fix in bb.js first.
I will look in to this.

saleel:
I've pushed changes to make it work with bb 0.84.0 (current bb version in master)

saleel:
Updated to verify with Solidity.
+ This test was compiling the inner circuit with wasm (and toml parser to prepare inputs). Changing this to use an existing compiled circuit.

saleel (approved):

PR #8039, opened 2025-04-11
chore: don't use `set_value_from_id` in `as_slice_length`

asterite:
# Description

## Problem

## Summary

Three things:
1. I realized there's never a need to replace values in instruction results: these are always the output of something so they should never be a value previously seen
2. This PR also removes `replace_results` as it seems it's no longer needed
3. I realized that instead of gathering all values to replace, then replacing all of them by traversing all blocks and instructions at the end, we could instead replace values in blocks and instructions the first time we traverse them. This requires using `take_instructions()` but otherwise it ends up in almost the same code, but one that should be more efficient as we only traverse things once, and we only replace things if we ever find things to replace.

## Additional Context

I wonder if it would make sense to have two intrinsics:
- `as_slice` -> (u32, [...])
- `as_slice_without_length` -> [...]

That is, once this optimization runs we also replace the intrinsic with the second one that just returns the slice. Then:
- we don't have to worry about not replacing the result with a constant
- maybe brillig is a bit faster as it doesn't need to compute the length and put it in a register

## Documentation

Check one:

# PR Checklist


asterite:
After some refactors this PR is ready for review again.

jfecher (approved):

TomAFrench:
Looks like this absolutely nuked the bignum testsuite:
image

PR #8049, opened 2025-04-14
feat: skip inverting field elements with values +-1 in ACVM

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

I held off on this previously as supposedly this didn't affect benchmarks but it causes a 90% drop in the ACVM benchmarks I added recently.

We now skip the inversion when solving an `AssertZero` opcode in cases where the denominator is +-1 as we can convert them into a multiplication.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/acvm-repo/acvm/src/pwg/arithmetic.rs
+++ b/acvm-repo/acvm/src/pwg/arithmetic.rs
@@ -59,7 +59,7 @@ impl ExpressionSolver {
                             Ok(())
                         }
                     } else {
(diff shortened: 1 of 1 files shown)


kashbrti (approved):

Rumata888:
I don't get it. We literally did the same thing and noticed no change whatsoever. Maybe it was being recompiled and all the benefits from faster witness generation were being overshadowed

PR #8050, opened 2025-04-14
chore(docs): update bb commands to match 0.84.0

saleel:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*


docs/docs/getting_started/quick_start.md
```diff
@@ -110,7 +111,7 @@ The proof is now generated in the `target` folder. To verify it we first need to
 bb write_vk -b ./target/hello_world.json -o ./target
 
 # Verify the proof
-bb verify -k ./target/vk -p ./target/proof
+bb verify -k ./target/vk -p ./target/proof -i ./target/public_inputs
```
critesjosh, line 114: This worked for me on beta.4 and bb 0.84 without passing the `-i ./target/public_inputs`. Is this just the recommended way to do it now? Or will it be required in the future?

docs/docs/how_to/how-to-solidity-verifier.mdx
```diff
@@ -131,27 +131,22 @@ This will generate a `Prover.toml` you can fill with the values you want to prov
...
+:::
 
-PROOF_WITHOUT_PUBLIC_INPUTS="${PROOF_HEX:$((NUM_PUBLIC_INPUTS * 32 * 2))}"
+To verify with Remix, you can pass the proof bytes as a hex string and the public inputs as a `bytes32[]` to the `verify` function.
 
-echo 0x$PROOF_WITHOUT_PUBLIC_INPUTS
-echo [$SPLIT_HEX_PUBLIC_INPUTS]
+```bash
+# Print the proof bytes as a hex string
+cat ./target/proof | od -An -v -t x1 | tr -d $' \n'
```
critesjosh, line 145:
```suggestion
echo -n "0x"; cat ./target/proof | od -An -v -t x1 | tr -d $' \n'
```
Should we prefix with "0x" since remix expects it?

critesjosh (approved):

docs/docs/getting_started/quick_start.md
saleel: This value is the default path so it will work even if you don't pass it explicitly.

PR #8077, opened 2025-04-15
chore: Print `unsafe { ... }` around calls to Brillig from ACIR in AST

aakoshh:
# Description

## Problem\*

Resolves an issue I found when trying to copy-paste printed AST as Noir: it's easy to get rid of `$f1` IDs, but it's not easy to identify all calls made from ACIR functions to `unconstrained` ones and add `unsafe { }` around them to satisfy the compiler. 

## Summary\*

Adds an `in_unconstrained` flag to the `AstPrinter`, set in `print_function`, and used in `print_call` to decide whether `unsafe` should be surrounding the call.

_We could make this optional like with the `show_ids` field, so it only kicks in with the `DisplayAstAsNoir(&program)` construct. At the moment I'm still printing the AST with IDs if the compiler crashes, just in case that extra detail helps (it did once, but usually doesn't), and because IDs are easy to get rid of, whereas `unsafe` isn't easy to add, I thought I'd include it by default. If you think it's too syntax oriented let me know; for me it would be okay to always print using `DisplayAstAsNoir`._

## Additional Context

Here's an example of a random AST printed with the additional keyword:

```rust
fn func_3(mut ctx_depth: u32) -> Field {
    if (ctx_depth == 0) {
        -52159555816448338737321817283027243824
    } else {
        ctx_depth = (ctx_depth - 1);
        func_5(if unsafe { func_4("JOIQ", ([G_A, G_A], "JUAU", (false, 3099534544668430827120783512681897970, 99734623872525348178457395891102316144, "BXKT", true)), ctx_depth) }[0].0 {
            "ZZLW"
        } else {
            unsafe { func_4("AGTH", ([-110772968255620736211771354824630909717, G_A], "XAVG", (false, 167985346747896750986631363550989799897, 67461072483853528276210850475181508438, "PFDM", true)), ctx_depth) }[2].3
        }, if false {
            ([G_A, (-130943816147133127254214549314831761348)], [])
        } else {
            (unsafe { func_1([-83858775359775106149427711427413560182, 229195978581537808051305761843578807607], G_A, [G_A, -17581793525685428048035436850144654957]) }, [])
        })
    }
}
```

The compiler will still issue warnings like this:
```
warning: Unsafe block must have a safety comment above it
   ┌─ src/main.nr:44:13
   │
44 │             unsafe { func_4("AGTH", ([-110772968255620736211771354824630909717, G_A], "XAVG", (false, 167985346747896750986631363550989799897, 67461072483853528276210850475181508438, "PFDM", true)), ctx_depth) }[2].3
   │             ------ The comment must start with the "Safety: " word
```

But it will not error like it would without the `unsafe`:

```
error: Call to unconstrained function is unsafe and must be in an unconstrained function or unsafe block
   ┌─ src/main.nr:44:14
   │
44 │              func_4("AGTH", ([-110772968255620736211771354824630909717, G_A], "XAVG", (false, 167985346747896750986631363550989799897, 67461072483853528276210850475181508438, "PFDM", true)), ctx_depth)[2].3
   │              --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
   │
```

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/monomorphization/printer.rs
+++ b/compiler/noirc_frontend/src/monomorphization/printer.rs
@@ -1,6 +1,6 @@
 //! This module implements printing of the monomorphized AST, for debugging purposes.
 
-use crate::{ast::UnaryOp, shared::Visibility};
+use crate::{ast::UnaryOp, monomorphization::ast::Ident, shared::Visibility};
 
 use super::ast::{
     Definition, Expression, FuncId, Function, GlobalId, LValue, LocalId, Program, Type, While,
@@ -11,12 +11,13 @@ use std::fmt::{Display, Formatter};
 #[derive(Debug)]
 pub struct AstPrinter {
     indent_level: u32,
+    in_unconstrained: bool,
     pub show_id: bool,
 }
 
 impl Default for AstPrinter {
     fn default() -> Self {
-        Self { indent_level: 0, show_id: true }
+        Self { indent_level: 0, in_unconstrained: false, show_id: true }
     }
 }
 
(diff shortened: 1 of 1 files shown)


TomAFrench (approved):
LGTM

PR #8079, opened 2025-04-15
fix: add proper handling for `u128` in comptime interpreter 

### LINKED ISSUE #8075: Noir fails to compile it's standard library file bn254.nr
### Aim

So I have the following code that looks something like:

```constant.nr
use dep::std::{self, embedded_curve_ops::{EmbeddedCurveScalar, EmbeddedCurvePoint, multi_scalar_mul}};

// BasePoint G for public-key derivations
pub fn base_point() -> EmbeddedCurvePoint {
    let random: Field = 71; // ASCII for 'G'
    
    // Create an array of the ASCII values
    let ghost_pad = [random];
    
    // Hash the ASCII values using Pedersen hash
    let hash = std::hash::pedersen_hash(ghost_pad);
    
    // Get the generator point and perform scalar multiplication with the hash
    let generator = EmbeddedCurvePoint::generator();
    let hash_scalar = EmbeddedCurveScalar::from_field(hash);
    let g_point = multi_scalar_mul([generator], [hash_scalar]);
    g_point
}
```

```uril.nr
use dep::std::{self, embedded_curve_ops::{EmbeddedCurvePoint, EmbeddedCurveScalar, multi_scalar_mul}};
use crate::types::{Note, MerklePath};
use crate::constants::BASE_POINT;



// k_ephemeral = H(spendKey || ephemeralKeyGenRandomness)
pub fn derive_ephemeral_private_key(spend_key: Field, randomness: Field) -> Field {
    // Use standard library's Pedersen hash for secure key derivation
    let inputs = [spend_key, randomness];
    std::hash::pedersen_hash(inputs)
}


// Derive ephemeral public key from private key
// K_ephemeral = k_ephemeral * G
pub fn derive_ephemeral_public_key(k_ephemeral: Field) -> EmbeddedCurvePoint {
    // Use the BASE_POINT constant G for scalar multiplication
    let k_ephemeral_scalar = EmbeddedCurveScalar::from_field(k_ephemeral);
    let k_ephemeral_point = multi_scalar_mul([BASE_POINT], [k_ephemeral_scalar]);
    k_ephemeral_point
}
```

Then I am getting the following error:

```
error: Only numeric types may be casted
   ┌─ std/field/bn254.nr:14:15
   │
14 │     let low = (x as u128) as Field;
   │               -------------------- `u128` is non-numeric
   │
```

I used `noirup --version nightly ` to use the latest compiler. It seems that my standard library is too old. Do you know how to fix it?

### Expected Behavior

There should be no compiler issues when compiling the standard library, `bn254.nr`

### Bug

I just wanted to use the EmbbededCurve library and it fails to compile it's dependencies

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

noirc version = 1.0.0-beta.4+f2cfe3e90872e84957473e9640d361ed94afeb8d

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

TomAFrench:
# Description

## Problem\*

Resolves #8075 

## Summary\*

There's a number of places in the comptime interpreter where the `u128` was falling through a catch-all branch so it was silently not handled without raising an error. It was then not being treated as a numeric type in a few places

This PR adds the necessary `u128` branches to these match statements. I've also moved the casting logic into a separate module along with a smoke test.

I've added a little bit of hackiness to `evaluate_for` as it was previously using an `i128` to handle all for loop ranges as this covered all previous types. This doesn't handle u128 however so I've had to split this function into two branches based on whether the start value is a signed or unsigned type.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/hir/comptime/interpreter.rs
```diff
@@ -1210,49 +1211,96 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
...
+
+                let must_break = match self.evaluate(for_.block) {
+                    Ok(_) => false,
+                    Err(InterpreterError::Break) => true,
+                    Err(InterpreterError::Continue) => false,
+                    Err(error) => {
+                        result = Err(error);
+                        true
+                    }
+                };
```
asterite, line 1261: It's not a lot of code duplication but we could have a helper function like this:

```rust
fn evaluate_loop_body(&mut self, body: ExprId, result: &mut IResult<Value>) -> bool {
    match self.evaluate(body) {
        Ok(_) => false,
        Err(InterpreterError::Break) => true,
        Err(InterpreterError::Continue) => false,
        Err(error) => {
            *result = Err(error);
            true
        }
    }
}
```

then call it like this:

```rust
let must_break = self.evaluate_loop_body(for_.block, &mut result);
```

And we could also use this for `loop` and `while` handling.

asterite (approved):
LGTM

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
```diff
@@ -1210,49 +1211,96 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
...
-                }
+        let start_value = self.evaluate(for_.start_range)?;
+        let signed = match start_value {
+            Value::I8(_) | Value::I16(_) | Value::I32(_) | Value::I64(_) => true,
+            Value::U1(_)
+            | Value::U8(_)
+            | Value::U16(_)
+            | Value::U32(_)
+            | Value::U64(_)
+            | Value::U128(_) => false,
```
jfecher, line 1222: Can you move out this explicit match for signedness? Either to a helper function or we can do:
```rs
let index_type = start_value.get_type();
if index_type.is_signed() {

} else if index_type.is_unsigned() {

} else {
    Err(...)
}

File: compiler/noirc_frontend/src/hir/comptime/interpreter.rs [hunk 3]
```diff
@@ -1210,49 +1211,96 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
...
-            self.push_scope();
-            self.current_scope_mut().insert(for_.identifier.id, make_value(i));
+        if signed {
+            // i128 can store all values from i8 - u64
+            let get_index = |this: &mut Self, expr| -> IResult<(_, fn(_) -> _)> {
+                match this.evaluate(expr)? {
+                    Value::I8(value) => Ok((value as i128, |i| Value::I8(i as i8))),
+                    Value::I16(value) => Ok((value as i128, |i| Value::I16(i as i16))),
+                    Value::I32(value) => Ok((value as i128, |i| Value::I32(i as i32))),
+                    Value::I64(value) => Ok((value as i128, |i| Value::I64(i as i64))),
```
jfecher, line 1241: Can we can replace this with an as_i128 helper?

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
```diff
@@ -1210,49 +1211,96 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
...
+            }
+        } else {
+            // u128 can store all values from u8 - u128
+            let get_index = |this: &mut Self, expr| -> IResult<(_, fn(_) -> _)> {
+                match this.evaluate(expr)? {
+                    Value::U8(value) => Ok((value as u128, |i| Value::U8(i as u8))),
+                    Value::U16(value) => Ok((value as u128, |i| Value::U16(i as u16))),
+                    Value::U32(value) => Ok((value as u128, |i| Value::U32(i as u32))),
+                    Value::U64(value) => Ok((value as u128, |i| Value::U64(i as u64))),
+                    Value::U128(value) => Ok((value, |i| Value::U128(i))),
```
jfecher, line 1277: Same here but for as_u128

jfecher (commented):
Looks good, but we can be more aggressive about limiting the size of functions by pulling out large match expressions into helpers.

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
TomAFrench: Are you wedded to the idea of this being a globally accessible helper function? I think it's nice having this tightly scoped.

I've restructured this code to not be a single massive closure.

jfecher: It doesn't need to be globally accessible but I think private to the file would be fine - I mainly want to simplify the logic of this function so that if someone is reading over it they need to read over fewer unnecessary helper functions to understand the main logic.

Issue #7902, opened 2025-04-03
Update ACIR display output to display all necessary information to reconstruct the circuit.


TomAFrench:
We've currently got a very good SSA display format such that we can serialize SSA to it as a readable format and then deserialize it later for testing. We don't have the same for ACIR however, for instance 

- [ ] All range opcodes show up as `BLACKBOX::RANGE [(<witness_number>)] [ ]` so there's no sign of how many bits we're constraining it to.
- [ ] We're inconsistent with how we display witnesses, in the case above it's `(<witness_number>)`, in `AssertZero` opcodes it's `_<witness_number>`, and potentially differently in other places. We should harmonise that.
- [ ] We don't output the witnesses which we're initialising a memory block with.
- [ ] When dealing with blackbox functions with many witness inputs we don't output all of these.
- [ ] Brillig functions are displayed with their `Debug` trait implementation. This is plain unreadable.

PR #8084, opened 2025-04-15
chore: add acir-gen unit tests per ssa instruction

guipublic:
# Description

## Problem\*

Increase test coverage for acir-gen

## Summary\*
Add a framework for acir-gen unit tests.
It generates SSA for a single instruction, converts it using acir-gen and execute it.
Then we compare the result with a SSA evaluation using constant folding.
All 'simple' instructions are covered, in particular all the binary instructions.
More tests with different types or missing instructions such as array operations and calls could be added in future.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -3691,4 +3691,203 @@ mod test {
 
         assert!(matches!(acvm.solve(), ACVMStatus::Failure::<FieldElement>(_)));
     }
+
+    // Convert the SSA input into ACIR and use ACVM to execute it
+    // Returns the ACVM execution status and the value of the 'output' witness value,
+    // unless the provided output is None or ACVM fail execution.
```
michaeljklein, line 3697:
```suggestion
    // unless the provided output is None or the ACVM fails during execution.
```

compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -3691,4 +3691,203 @@ mod test {
...
+            (status, output.map(|o| acvm.witness_map()[o]))
+        } else {
+            (status, None)
+        }
+    }
+
+    fn get_constant_main_src(v0: FieldElement, v1: FieldElement, typ: &str) -> String {
+        format!(
+            "acir(inline) fn main f0 {{
+            b0(v0: [{typ}; 2]):
```
michaeljklein, line 3727: Can normal variable names be used here? E.g. `v0 -> input_array`, `v1 -> lhs`, etc.

TomAFrench:
@vezenovm If we're testing SSA -> ACIR then we don't need to fully complete #7902 in order to do this (obviously the test becomes more binding in that case) but we can probably get enough out with the current output.

vezenovm:
> If we're testing SSA -> ACIR then we don't need to fully complete #7902 in order to do this (obviously the test becomes more binding in that case) but we can probably get enough out with the current output.

Sounds good, agreed. In addition to your note, we already do test in this way for the existing ACIR gen unit tests. Either way I still think the ACIR gen unit tests should be more in the vein mentioned in my original comment. e.g. `does_not_generate_memory_blocks_without_dynamic_accesses` (post SSA parser) and `brillig_stdlib_calls_with_multiple_acir_calls` (pre SSA parser) are a good example of the kind of tests I would expect.

I mentioned completing #7902 and an ACIR parser as it may be worth it to still complete that task before adding more unit tests as it would not only make the unit tests more binding but quicker to write. If we have tests we can add now though without completing #7902 though I won't block those.

compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -3691,4 +3691,203 @@ mod test {
...
+            .collect::<BTreeMap<_, _>>();
+        let len = inputs.len() as u32;
+        let initial_witness = WitnessMap::from(inputs);
+
+        for op in operators {
+            let (src, with_output) = generate_test_instruction_from_operator(op);
+            let output = if with_output { Some(Witness(len)) } else { None };
+            let ssa = Ssa::from_str(&(main.to_owned() + &src)).unwrap();
+            let result = execute_ssa(ssa, initial_witness.clone(), output.as_ref());
+            let const_ssa = Ssa::from_str(&(const_main.to_owned() + &src)).unwrap();
```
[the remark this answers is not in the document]
guipublic: It would be possible by injecting the constant values directly in the dfg, that would be easy to do but would require to add some functions in the dfg, used only for testing purposes. I can try this.

compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -3691,4 +3691,203 @@ mod test {
...
+            let const_ssa = const_ssa.fold_constants();
+            let const_result = execute_ssa(const_ssa, initial_witness.clone(), output.as_ref());
+            match (result, const_result) {
+                ((ACVMStatus::Failure(_), _), (ACVMStatus::Failure(_), _)) => {
```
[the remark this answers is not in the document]
guipublic: I am not sure, the constant folding version can be quite different so I don't expect this. It's not easy to get a definitive answer.

compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -3691,4 +3691,203 @@ mod test {
...
+        test_operators(
+            &operators,
+            "Field",
+            &[FieldElement::from(1_usize), FieldElement::from(2_usize)],
+        );
+
+        // u32, equal inputs
+        test_operators(
+            &operators,
+            "u32",
```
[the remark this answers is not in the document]
guipublic: Since it test a single operation, it's very fast. I'd like to add more tests in a subsequent PR but I wanted to validate the approach first.
I agree that we should parse only once if we are using multiple inputs for the same test.

compiler/noirc_evaluator/src/acir/mod.rs
guipublic: Done

compiler/noirc_evaluator/src/acir/mod.rs
guipublic: Done

compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -3691,4 +3691,203 @@ mod test {
...
+        initial_witness: WitnessMap<FieldElement>,
+        output: Option<&Witness>,
+    ) -> (ACVMStatus<FieldElement>, Option<FieldElement>) {
+        let brillig = ssa.to_brillig(&BrilligOptions::default());
+        let (acir_functions, brillig_functions, _, _) = ssa
+            .into_acir(&brillig, &BrilligOptions::default(), ExpressionWidth::default())
+            .expect("Should compile manually written SSA into ACIR");
+        assert_eq!(acir_functions.len(), 1);
+        let main = &acir_functions[0];
+        let mut acvm = ACVM::new(
```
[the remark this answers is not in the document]
guipublic: For me it is a unit test because it test a single operator.

compiler/noirc_evaluator/src/acir/mod.rs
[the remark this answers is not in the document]
guipublic: Done, I now use the ssa interpreter instead of doing constant folding.
This highlighted some differences between ACVM execution and SSA interpreter, which I commented out.

compiler/noirc_evaluator/src/acir/tests/mod.rs
```diff
@@ -926,3 +927,208 @@ fn do_not_overflow_with_constant_constrain_neq() {
     assert_eq!(acir_functions.len(), 1);
     assert!(acir_functions[0].opcodes().is_empty());
 }
+
+// Convert the SSA input into ACIR and use ACVM to execute it
+// Returns the ACVM execution status and the value of the 'output' witness value,
+// unless the provided output is None or the ACVM fails during execution.
```
vezenovm, line 933:
```suggestion
/// Convert the SSA input into ACIR and use ACVM to execute it
/// Returns the ACVM execution status and the value of the 'output' witness value,
/// unless the provided output is None or the ACVM fails during execution.
```
nit: might as well make these doc comments if they are on the function

compiler/noirc_evaluator/src/acir/tests/mod.rs
```diff
@@ -926,3 +927,208 @@ fn do_not_overflow_with_constant_constrain_neq() {
...
+    format!(
+        "acir(inline) fn main f0 {{
+            b0(inputs: [{typ}; 2]):
+              lhs = array_get inputs, index u32 0 -> {typ}
+              rhs = array_get inputs, index u32 1 -> {typ}
+              "
+    )
+}
+
+// Create a ssa instruction corresponding to the operator, using v1 and v2 as operands.
```
vezenovm, line 970: same nit about changing to doc comments. Can also do ssa -> SSA

compiler/noirc_evaluator/src/acir/tests/mod.rs
```diff
@@ -926,3 +927,208 @@ fn do_not_overflow_with_constant_constrain_neq() {
...
+            ),
+            false,
+        )
+    }
+}
+
+// Execute a simple operation for each operators
+// The operation is executed from SSA IR using ACVM after acir-gen
+// and also directly on the SSA IR via constant folding, by hardcoding the values
+// via get_constant_main_src() prefix.
```
vezenovm, line 1019: same doc comment nit

compiler/noirc_evaluator/src/acir/tests/mod.rs
```diff
@@ -926,3 +927,208 @@ fn do_not_overflow_with_constant_constrain_neq() {
...
+        }}"
+            ),
+            false,
+        )
+    }
+}
+
+// Execute a simple operation for each operators
+// The operation is executed from SSA IR using ACVM after acir-gen
+// and also directly on the SSA IR via constant folding, by hardcoding the values
```
vezenovm, line 1018: Looks like this comment needs to be update that it is testing against the SSA interpreter

compiler/noirc_evaluator/src/acir/tests/mod.rs
```diff
@@ -926,3 +927,208 @@ fn do_not_overflow_with_constant_constrain_neq() {
...
+    let initial_witness = WitnessMap::from(inputs);
+
+    for op in operators {
+        let (src, with_output) = generate_test_instruction_from_operator(op);
+        let output = if with_output { Some(Witness(len)) } else { None };
+        let ssa = Ssa::from_str(&(main.to_owned() + &src)).unwrap();
+        // ssa execution
+        let result = ssa.interpret(vec![inputs_int.clone()]);
+        // acir execution
+        let result2 = execute_ssa(ssa, initial_witness.clone(), output.as_ref());
```
vezenovm, line 1048:
```suggestion
        let acir_result = execute_ssa(ssa, initial_witness.clone(), output.as_ref());
```
just so it is clear one is from executing ssa and one is from executing acir. Can even rename `result` above to `ssa_result`

compiler/noirc_evaluator/src/acir/tests/mod.rs
```diff
@@ -926,3 +927,208 @@ fn do_not_overflow_with_constant_constrain_neq() {
...
+        "unchecked_mul",
+        "mod",
+        "lt",
+        "or",
+        "not",
+        "range_check 8",
+        "truncate 8 32",
+    ];
+    let inputs = [FieldElement::from(2_usize), FieldElement::from(1_usize)];
+    test_operators(&operators, "u32", &inputs);
```
vezenovm, line 1112: I guess in follow-ups we can test the remaining numeric types then.

compiler/noirc_evaluator/src/acir/tests/mod.rs
```diff
@@ -926,3 +927,208 @@ fn do_not_overflow_with_constant_constrain_neq() {
...
+    let operators = [
+        "add",
+        "sub",
+        "mul",
+        "div",
+        "eq",
+        // Bitwise operations are not allowed on field elements
+        // SSA interpreter will emit an error but not ACVM
+        // "and",
+        // "xor",
```
vezenovm, line 1079: We can follow-up with expected error tests that would have corresponding issues.

vezenovm (approved):
Looks good! I am more comfortable with these tests being under `acir` now that they use the SSA interpreter vs. constant folding. I think ideally these would be moved into their own `interpreter` module or something, but as this PR has been up for a while I am good with the PR as is. The tests can reorganized in follow-ups. Just some last nits with regards to comments.

compiler/noirc_evaluator/src/acir/tests/mod.rs
guipublic: Done

compiler/noirc_evaluator/src/acir/tests/mod.rs
guipublic: Done

compiler/noirc_evaluator/src/acir/tests/mod.rs
guipublic: Done

compiler/noirc_evaluator/src/acir/tests/mod.rs
guipublic: Done

compiler/noirc_evaluator/src/acir/tests/mod.rs
guipublic: Done

guipublic:
Closing as duplicate.
The PR was duplicated in #9185

PR #8091, opened 2025-04-15
chore: don't use `set_value_from_id` in `constant_folding`

asterite:
# Description

## Problem

## Summary

This is the PR that removes the final occurrences of `set_value_from_id` so, in addition to that:
1. `set_value_from_id` is removed
2. `resolve` is removed, together with `replaced_value_ids`
3. a `ValueMapping` is introduced to wrap the `HashMap` so looking up values is more straight-forward without having to rely on another function to do the recursive lookup

## Additional Context

I originally wanted to do points 2 and 3 above in a separate PR. However, CI for noir-wasm was timing out... I don't know if it's because the new way is slower, plus having all of the `resolve` calls (which would now hit an empty HashMap so it shouldn't make things slower 🤷). After I pushed 2 here the CI issue was solved, so maybe those `resolve` calls were making things slower, or maybe it was a fluke. In any case we would have removed `resolve` in a future PR, so... 

## Documentation

Check one:

# PR Checklist


jfecher (approved):
No more resolve! :tada:

Issue #8011, opened 2025-04-10
Negative range literals are not iterated


aakoshh:
### Aim

Related to https://github.com/noir-lang/noir/issues/8009

I wanted to confirm whether negative ranges are supposed to work. The compiler doesn't allow me to use `for i in -5..10` because it says `-5` is outside the range of `u32`, but it compiles like this:

```rust
unconstrained fn main() {
    let s: i64 = -5;
    let e: i64 = 10;
    for i in s..e {
        println(i);
    }
}
```

### Expected Behavior

I would expect to see the numbers -5, -4, ..., 0, 1, ..., 9 to be printed, like in Rust.

### Bug

Nothing is printed. The SSA shows the entire loop is eliminated:

```console
$ cargo run -q -p nargo_cli -- execute --show-ssa-pass "Initial"
After Initial SSA:
brillig(inline) fn main f0 {
  b0():
    return
}
```

### To Reproduce

See above.

### Workaround

Yes

### Workaround Description

Convert to unsigned values, but it might be unexpected if somebody has negative values and expects them to work like Rust.

Actually if we run the program like this, then it prints the expected values:

```rust
unconstrained fn main(s: i64, e: i64) {
    for i in s..e {
        println(i);
    }
}
```

```toml
# Prover.toml
s = -5
e = 10
```

```console
bash-5.2$ cargo run -q -p nargo_cli -- execute --show-ssa-pass "Initial"
After Initial SSA:
brillig(inline) fn main f0 {
  b0(v0: i64, v1: i64):
    jmp b1(v0)
  b1(v2: i64):
    v3 = lt v2, v1
    jmpif v3 then: b2, else: b3
  b2():
    call f1(v2)
    v6 = unchecked_add v2, i64 1
    jmp b1(v6)
  b3():
    return
}
brillig(inline) fn println f1 {
  b0(v0: i64):
    call f2(u1 1, v0)
    return
}
brillig(inline) fn print_unconstrained f2 {
  b0(v0: u1, v1: i64):
    v20 = make_array b"{\"kind\":\"signedinteger\",\"width\":64}"
    call print(v0, v1, v20, u1 0)
    return
}

-5
-4
-3
-2
-1
0
1
2
3
4
5
6
7
8
9
[regression_fuzz_testing] Circuit witness successfully solved

```

### Project Impact

None

### Nargo Version

[...]

PR #8094, opened 2025-04-15
fix: Use `IntegerConstant` for loop boundaries in `unrolling`

### LINKED ISSUE #8009: Compiler crashes unrolling for loop with negative boundaries
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/7987

```bash
cargo +nightly fuzz run init_vs_final -- -max_total_time=60 -max_len=1048576
```

The compiler crashed during unrolling:
```
thread '<unnamed>' panicked at compiler/noirc_evaluator/src/ssa/opt/unrolling.rs:752:9:
attempt to multiply with overflow
```

Based on the printed AST, here's a different example of a program that doesn't behave as expected:
```rust
unconstrained fn main() -> pub u32 {
    let mut sum = 0;
    let start: i64 = -51675949543456665;
    let end: i64 = -1;
    let start = (start % 5) as u32;
    let end = (end % 5) as u32;
    for i in start..end {
        sum = sum + i;
    }
    sum
}
```

#### Negative ranges

I got the notion that loops should be able to handle negative ranges such as `for i in -10 .. -5` looking at `evaluate_for` in `interpreter.rs`, but it looks like Noir actually doesn't allow anything other than `u32`, so maybe this is outside the spec.

However, https://github.com/noir-lang/noir/issues/8011 confirmed in the workaround that negative values _can_ be iterated when they are coming from the inputs. It's only the SSA gen that can't handle them, when they are known literals.

### Expected Behavior

The program should compile and run a loop which is equivalent to `0 .. 4`:

```console
bash-5.2$ python3
>>> -51675949543456665 % 5
0
>>> -1 % 5
4
```

So it should return 0+1+2+3=6 in this case. 

EDIT: I see Rust returns `-1` for `-1 % 5`, and we're supposed to add 5 if we want it to be positive. The problem in this case is with the `as u32` of a negative number. See below for the original problem.

### Bug

```console
bash-5.2$ cargo run -q -p nargo_cli -- compile --show-ssa-pass Initial --force-brillig
After Initial SSA:
brillig(inline) fn main f0 {
  b0():
    v1 = allocate -> &mut u32
    store u32 0 at v1
    jmp b1(u32 0)
  b1(v0: u32):
    v4 = lt v0, u32 4294967295
    jmpif v4 then: b2, else: b3
  b2():
    v6 = load v1 -> u32
    v7 = add v6, v0
    store v7 at v1
    v9 = unchecked_add v0, u32 1
    jmp b1(v9)
  b3():
    v5 = load v1 -> u32
    return v5
}
```

This shows that the upper bound is determined to be `4294967295` instead of 4. 
The explanation for this turned out to be that `-1i32` is represented as `u32::MAX`, and casting to `u32` results in this.

In the original error the AST looked like this:
```
for idx_a$l0 in (-51675949543456665 % 5) .. (-1 % 5)
```
and the bounds in the SSA were different:
```
brillig(inline) fn main f0 {
  b0():
    jmp b1(i64 0)
  b1(v0: i64):
    v3 = lt v0, i64 18446744073709551615
    jmpif v3 then: b2, else: b3
  b2():
    v5 = make_array [] : [[[u8; 0]; 0]; 0]
    v6 = allocate -> &mut [[[u8; 0]; 0]; 0]
    store v5 at v6
    v8 = unchecked_add v0, i64 1
    jmp b1(v8)
  b3():
    return u1 0
}
```

That results in the `iterations * useful_instructions` overflowing the type in `unrolling`, because it treats `18446744073709551615` as it is, rather than as -1.

### To Reproduce

See above.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+a6e6ecd33bed019989bd3f9e84543826dd78817a (git version hash: 7bcc3a96fdbbda1ab8e6e7cecc06412fe0dabdc7, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #8009 

## Summary\*

Fixes the `get_const_lower_bound` and `get_const_upper_bound` methods in the `unrolling` module to use a new `IntegerConstant` type, which uses the `binary::try_convert_field_element_to_signed_integer` function for `Signed` types to convert them to `i128`, which can correctly represent negative values for our use case. `Unsigned` types are converted to `u128` instead.

## Additional Context

The program in the bug ticket was misleading: `-1 % 5` is 4 according to Python, but -1 according to Rust; it was turned into `4294967295` by the `as u32` operation, not the modulo. I added the cast because the frontend didn't let me use negative literals in the loop.

The negative result for modulo would not necessarily be a problem for the AST story: if we loop from 3..-1, it's not going to do anything, but that's okay, other times it will. Generating a block to declare a value, and then conditionally add the modulo to it to make it positive seems like extreme overkill (it would be unreadable in AST prints). 

The original problem where it turned into 18446744073709551615 did not involve any casting, only a literal in the AST. This literal cannot be used in the for loop in Noir, nor does it allow negatives in the loop, so I added a test directly creating SSA from a hand crafted AST to replicate this. It turned out this is expected behaviour; this number if `u64::MAX`, which is how negative values of `i64` are represented as a `Field` in the SSA. Special care needs to be taken to convert the value back to signed integers. (EDIT: It turns out it can be used if the bounds are declared separately as variables.)

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
fyi, we currently allow using u128s in loop bounds which may cause issues.

aakoshh:
I may have crossed my eyes, but I thought I restricted the index to be 64 bits because Noir wouldn’t let me use 128 bits:

https://github.com/noir-lang/noir/blob/b5918474ccf88fba83d29aa909a8d3048e2d0e71/tooling/ast_fuzzer/src/program/func.rs#L788

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -282,6 +286,17 @@ impl Loop {
         Self { header, back_edge_start, blocks }
     }
 
+    /// Handle negative values represented as fields.
+    fn get_const_as_i128(function: &Function, id: ValueId) -> Option<i128> {
+        let (value, typ) = function.dfg.get_numeric_constant_with_type(id)?;
+        if matches!(typ, NumericType::NativeField) {
+            // Shouldn't happen, as fields don't have meaningful ordering.
```
vezenovm, line 293: This comment in isolation I think adds confusion as the title and function logic alone indicates we are just attempting to fetch an i128. If we are assuming this function will only ever be called on the jump value of the loop let's comment that on the function itself and replace this with an assertion.

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -680,16 +691,14 @@ impl Loop {
...
-        let upper = upper.try_to_u64()?;
         let refs = self.find_pre_header_reference_values(function, cfg)?;
 
         let (loads, stores) = self.count_loads_and_stores(function, &refs);
         let increments = self.count_induction_increments(function);
         let all_instructions = self.count_all_instructions(function);
 
         Some(BoilerplateStats {
-            iterations: (upper - lower) as usize,
+            iterations: (upper - lower).max(0) as usize,
```
vezenovm, line 701: nit: maybe just a brief comment that we do not allow reverse ranges and this will need to be updated in that case

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -1199,6 +1207,27 @@ mod tests {
...
+        assert_eq!(stats.iterations, 0);
+        assert_eq!(stats.unrolled_instructions(), 0);
+    }
+
+    #[test]
+    fn test_boilerplate_stats_i64_non_empty() {
+        // Looping -4..-1, which should be 3 iterations.
+        let ssa = brillig_unroll_test_case_6470_with_params(
+            "i64",
+            "18446744073709551612",
```
vezenovm, line 1224: I see it is `-4` from the comment above, but perhaps we could do this through code? Or at least just an additional comment that this is how Noir represents negative numbers.

vezenovm (commented):
LGTM just some minor nits

vezenovm:
> Right. Well currently these two are just optimisations, so hopefully they would just be skipped as None

It does look like as this just used during optimizations we should just return None for u128 range values over i128. So this shouldn't cause any issues, but let's make an issue to follow-up on this (although it will be low priority). Could even have a small unit test that shows an optimization that works will a u32 or u64 bounds but not a certain u128 bound (e.g. like `hoist_array_gets_using_induction_variable_with_const_bound` or a different test that is using the induction variable bounds).

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
aakoshh: Removed this in favour of an `IntegerConstant` type, which has a constructor returning `None` if it was called with a `NativeField`, so there is no confusion.

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
aakoshh: Done 👍

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
aakoshh: Changed it to `u64::MAX - 3` and also added a comment.

aakoshh:
> It does look like as this just used during optimizations we should just return None for u128 range values over i128. 

@asterite pointed out in #8011 that using `u128` range bounds will crash the compiler, which it did. I got confused earlier because of this:

```
error: The value `170141183460469231731687303715884105715` cannot fit into `u32` which has range `0..=4294967295`
   ┌─ src/main.nr:22:14
   │
22 │     for _ in 170141183460469231731687303715884105715..170141183460469231731687303715884105730 {
   │              ---------------------------------------
   │
   = Call stack:
     1. src/main.nr:22:14

```
I forgot that you can still assign to a `u128` and it will accept it. 

I introduced a new `IntegerConstant` type which can be either `Signed` or `Unsigned` to handle all permissible cases, and updated the unrolling and hoisting to use this type instead of `FieldElement` for handling bounds.

compiler/noirc_evaluator/src/ssa/ir/integer.rs
```diff
@@ -0,0 +1,122 @@
...
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        match (self, other) {
+            (Self::Signed { value: a, .. }, Self::Signed { value: b, .. }) => a.partial_cmp(b),
+            (Self::Signed { value: a, .. }, Self::Unsigned { value: b, .. }) => {
+                if a.is_negative() || *b > i128::MAX as u128 {
+                    Some(Ordering::Less)
+                } else {
+                    let a: u128 = (*a).try_into().ok()?;
+                    a.partial_cmp(b)
+                }
```
vezenovm, line 109:
```suggestion
                if a.is_negative() {
                    Some(Ordering::Less)
                } else {
                    (*a).try_into().ok().and_then(|a: u128| a.partial_cmp(b))                
                }
```
nit: we should be able to simplify this if statement

compiler/noirc_evaluator/src/ssa/ir/integer.rs
```diff
@@ -0,0 +1,122 @@
...
+            }
+            (Self::Unsigned { value: a, .. }, Self::Signed { value: b, .. }) => {
+                if b.is_negative() || *a > i128::MAX as u128 {
```
vezenovm, line 112:
```suggestion
                if b.is_negative() {
```
nit: same thing here

compiler/noirc_evaluator/src/ssa/ir/integer.rs
```diff
@@ -0,0 +1,122 @@
...
+                    Some(Ordering::Less)
```
vezenovm, line 113:
```suggestion
                    Some(Ordering::Greater)
```
`Less` is incorrect no? If `b` is negative the unsigned `a` is always greater.

compiler/noirc_evaluator/src/ssa/ir/integer.rs
```diff
@@ -0,0 +1,122 @@
...
+                } else {
+                    let b: u128 = (*b).try_into().ok()?;
+                    a.partial_cmp(&b)
```
vezenovm, line 116:
```suggestion
                    (*b).try_into().ok().and_then(|b: u128| a.partial_cmp(&b))
```
nit: same thing here

vezenovm (commented):
Looks great! I like the new type. I just have some minor nits and one correction regarding comparisons in the new type

compiler/noirc_evaluator/src/ssa/ir/integer.rs
aakoshh: Ah yes, thanks, copy paste error 🤦

vezenovm (approved):

vezenovm:
> `rollup-block-root`	`152` s	`119` s	`1.28`

Pretty surprised to see this large of a regression. We do have large loops that need to be unrolled here. Perhaps we are doing a lot of unnecessary conversions?

PR #8097, opened 2025-04-15
chore(optimization): Enable experimental ownership clone scheme by default

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/7287

## Summary\*

Enables the ownership scheme implemented in https://github.com/noir-lang/noir/pull/7861 by default for fewer array copies - see the copy counts in that PR's comments.

## Additional Context

Do not be deceived by line count on this PR! Most are test snapshot changes and ~400 net lines removed from ownership/mod.rs.

Expecting to see CI results similar to https://github.com/noir-lang/noir/pull/7951.

Note that the `-Zownership` flag is still in the compiler, it just now only toggles whether immutable references are allowed.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/ownership/tests.rs
```diff
@@ -0,0 +1,118 @@
...
+unconstrained fn len$f1(arr$l1: [Field; 2]) -> u32 {
+    2
+}
+unconstrained fn eq$f2(lhs$l2: [Field; 2], rhs$l3: [Field; 2]) -> bool {
+    ((lhs$l2[0] == rhs$l3[0]) & (lhs$l2[1] == rhs$l3[1]))
+}
+";
+
+    let program = get_monomorphized_no_emit_test(src).unwrap();
+    similar_asserts::assert_eq!(expected, program.to_string());
```
TomAFrench, line 49: At the risk of being a broken record... `insta::assert_snapshot!` can be used in much the same way but makes it easier to update these in future if necessary.

jfecher: I don't really like how it makes it so that you can't see what `expected` code you're testing for. It makes it unclear from the source code what cases are actually covered. If you'd still prefer it I'll add it though.

jfecher: Ok, thank you. I was getting odd errors with `insta::assert_snapshot!(expected, program.to_string());` expecting a file name and must have gotten confused.

TomAFrench (approved):
LGTM, if you're unsure of this change then it might be a good idea to run aztec-packages CI with the experimental ownership flag turned on to make sure that this doesn't break some of the tests we don't run.

TomAFrench:
huh, looks like we have a failing test here as well.

jfecher:
> huh, looks like we have a failing test here as well.

Those were the inliner tests since the expected counts have now changed. They had some non-superficial changes so I was taking a little while to go through them and update them all. The main thing to note is that since we often clone into each function call, the actual call to `array_refcount` now clones the reference count which needs to be accounted for. Hopefully we can remove this when immutable references are later stabilized.

PR #8102, opened 2025-04-15
chore(experimental): Function::simple_optimization for SSA optimizations

asterite:
# Description

## Problem

No issue, but many SSA optimizations share a common pattern so it might be nice to abstract that pattern for readability purposes, and also to make it easier to add new passes that fit into this pattern.

## Summary

I was able to extract a common pattern for simple optimization into a function that takes a callback. So far I was able to apply this to 9 optimizations so I think it might be worth keeping it, specially if we later keep introducing "simple" optimizations.

These optimizations traverse all reachable blocks and all instructions, receiving a callback to "process" an instruction at a time. You can then choose to keep that instruction (the default behavior), remove it, or insert it before the callback ends in case you need to add more instructions afterwards. You can also replace a value with another one with a single function call and `simple_optimization` will take care of replacing all subsequent values.

## Additional Context

I think performance of running these optimizations is more or less the same (CI didn't time out). I also compiled `rollup-block-root` and it takes around the same time and memory.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs
@@ -41,6 +41,10 @@ impl Function {
             return;
         }
 
+        // Make sure this optimization runs when there's only one block
+        let block = self.entry_block();
+        assert_eq!(self.dfg[block].successors().count(), 0);
+
         let one = self.dfg.make_constant(FieldElement::one(), NumericType::bool());
         let mut active_condition = one;
         let mut last_side_effects_enabled_instruction = None;
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs
@@ -9,13 +9,11 @@
 //!     - An [Instruction] with side-effects is encountered, if so then insert the currently saved [Instruction::EnableSideEffectsIf]
 //!       before the [Instruction]. Continue inserting instructions until the next [Instruction::EnableSideEffectsIf] is encountered.
 //!
-use std::collections::HashSet;
 
 use acvm::{FieldElement, acir::AcirField};
 
 use crate::ssa::{
     ir::{
-        basic_block::BasicBlockId,
         dfg::DataFlowGraph,
         function::{Function, RuntimeType},
         instruction::{BinaryOp, Hint, Instruction, Intrinsic},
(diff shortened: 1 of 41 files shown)


compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs
```diff
@@ -43,244 +41,205 @@ impl Function {
             return;
         }
 
-        let mut context = Context::default();
-        context.block_queue.push(self.entry_block());
-
-        while let Some(block) = context.block_queue.pop() {
```
vezenovm, line 49: Almost all of these optimizations operate over `self.reachable_blocks` as they do not do anything that depends upon the ordering of blocks. However, using `reachable_blocks` in place of the post order has led to errors in the past (e.g. https://github.com/noir-lang/noir/pull/7140). In this pass it looks like we are setting up a block queue that essentially maps the post order vs. using `reachable_blocks` which is a single ordered set of all blocks. 

Based upon the module doc comments on this pass it would be best to still use the post order vs. reachable blocks in this case even though we don't have any failures on our test suite. This looks to be because this pass is ran after our CFG is totally flattened. Let's add an assert here that we have a flat program. 

We could always use the post order for `simple_optimization` but recomputing the post-order when we don't need to do so seems unnecessarily cumbersome so maybe it should just be another option for `simple_optimization`. We can differentiate post order vs reachable blocks in a follow-up if it is needed. I think for now just an assertion we have a single block should sufifce.

vezenovm (changes requested):
Looks good! Just one comment regarding adding a sanity assertion

compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs
asterite: > Let's add an assert here that we have a flat program.

Sounds good! I guess this also means we could simplify things here a bit and just operate on the single entry block instead of keeping `visited_blocks` and `block_queue`, right? (I assume yes so I'll start working on this, but let me know otherwise!)

compiler/noirc_evaluator/src/ssa/opt/simple_optimization.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// are then traversed in turn. For each one, `f` will be called with a context.
+    ///
+    /// The current instruction will be inserted at the end of the callback given to `mutate` unless
+    /// `remove_current_instruction` or `insert_current_instruction` are called.
+    ///
+    /// `insert_current_instruction` is useful if you need to insert new instructions after the current
+    /// one, so this can be done before the callback ends.
+    ///
+    /// `replace_value` can be used to replace a value with another one. This substitution will be
+    /// performed in all subsequent instructions.
```
vezenovm, line 22: Let's also add a safety comment that this function uses `reachable_blocks` internally. Replacing instructions which depend upon the ordering of blocks with this method can potentially lead to correctness errors.

vezenovm (approved):

compiler/noirc_evaluator/src/ssa/opt/simple_optimization.rs
asterite: We could also rename `simple_optimization` to `simple_reachable_blocks_optimization` to make this clearer. It's a bit longer but it's just one call per SSA pass that uses it. Then eventually we could add one for post order if needed. What do you think?

asterite: I ended up renaming the method plus adding a link to `reachable_blocks` in the doc comments.

vezenovm: Thanks! I like the rename as well.

Issue #8083, opened 2025-04-15
Possibility to take `Field` mod `Field` in Brillig


jewelofchaos9:
### Aim

.

### Expected Behavior

Execution failed or compilation panicked

### Bug

```
brillig(inline) predicate_pure fn main f0 {
  b0(v0: Field, v1: Field):
    v2 = mod v0, v1
    return v2
}
```
is valid SSA

### To Reproduce
In `tooling/ssa_fuzzer/lib.rs`
```rs
mod tests {
    ........
    use noirc_evaluator::ssa::ir::types::NumericType;
    use noirc_evaluator::ssa::ir::instruction::BinaryOp;
    use crate::runner::execute_single;
    #[test]
    fn regression_fields_mod_op_brillig() {
        let mut brillig_builder = FuzzerBuilder::new_brillig();
        let field_var_brillig_id_1 = brillig_builder.insert_variable(ValueType::Field.to_ssa_type()).value_id;
        let field_var_brillig_id_2 = brillig_builder.insert_variable(ValueType::Field.to_ssa_type()).value_id;

        let mod_brillig = brillig_builder.builder.insert_binary(field_var_brillig_id_1, BinaryOp::Mod, field_var_brillig_id_2);

        brillig_builder.builder.terminate_with_return(vec![mod_brillig]);
        let brillig_program = brillig_builder.compile().unwrap();

        let mut initial_witness = WitnessMap::new();
        initial_witness.insert(Witness(0), FieldElement::from(13371337_u32));
        initial_witness.insert(Witness(1), FieldElement::from(12341234_u32));

        let result_witness = Witness(2);
        let execution_result = execute_single(&brillig_program.program, initial_witness, result_witness);
        println!("{:?}", execution_result);
        assert!(execution_result.is_err());
    }
}
```

and run `cargo test`

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

TomAFrench:
We disallow this operation in the typechecker along with other operations which are invalid on `Field`s such as bitwise operations.

https://github.com/noir-lang/noir/blob/9158338da31218d8f0a28d3b13491028dfa2d040/compiler/noirc_frontend/src/elaborator/types.rs#L1196-L1207

jewelofchaos9:
For security reasons, we want full compliance between ACIR and Brillig execution. I think security should be ensured at every level. And if a bug can't be reproduced now with `nargo`, it may be critical in the future.

[...]

PR #8105, opened 2025-04-16
fix(ssa): fix possibility to `Field % Field` operaions in Brillig from SSA

jewelofchaos9:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/8083

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/parser/tests.rs
+++ b/compiler/noirc_evaluator/src/ssa/parser/tests.rs
@@ -647,9 +647,8 @@ fn regression_modulo_fields_brillig() {
         }
         ";
     let ssa = Ssa::from_str(src).unwrap();
-    match std::panic::catch_unwind(AssertUnwindSafe(|| ssa.to_brillig(&BrilligOptions::default())))
-    {
-        Ok(_) => panic!("Should fail on compilation"),
-        Err(_) => (),
-    }
+    assert!(
+        std::panic::catch_unwind(AssertUnwindSafe(|| ssa.to_brillig(&BrilligOptions::default())))
+            .is_err()
+    );
 }
--- a/compiler/noirc_evaluator/src/ssa/parser/tests.rs
+++ b/compiler/noirc_evaluator/src/ssa/parser/tests.rs
@@ -638,7 +638,6 @@ fn test_parses_keyword_in_function_name() {
 #[should_panic = "Attempt to modulo fields"]
 fn regression_modulo_fields_brillig() {
     use crate::brillig::BrilligOptions;
-    use std::panic::AssertUnwindSafe;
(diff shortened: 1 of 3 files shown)


tooling/ssa_fuzzer/src/lib.rs
```diff
@@ -225,4 +227,39 @@ mod tests {
...
+
+        brillig_builder.builder.terminate_with_return(vec![mod_brillig]);
+        let brillig_program = brillig_builder.compile();
+        let brillig_program = match brillig_program {
+            Ok(program) => program,
+            Err(_e) => {
+                // compilation failed, its nice
+                return;
+            }
+        };
```
TomAFrench, line 253: Can you replace this with usage of the SSA parser?

TomAFrench: I'm not clear on why this is succeeding in that case as this should panic before we reach execution.

jewelofchaos9: It is not succeeding, fails on `::compile()`. I cat delete the rest of the test to avoid confusion

jfecher: That is what I expected as well while reading this.

compiler/noirc_evaluator/src/ssa/parser/tests.rs
```diff
@@ -633,3 +633,22 @@ fn test_parses_keyword_in_function_name() {
...
+            v2 = mod v0, v1
+            return v2
+        }
+        ";
+    let ssa = Ssa::from_str(src).unwrap();
+    assert!(
+        std::panic::catch_unwind(AssertUnwindSafe(|| ssa.to_brillig(&BrilligOptions::default())))
+            .is_err()
+    );
+}
```
TomAFrench, line 654:
```suggestion
#[test]
#[should_panic = "Attempt to modulo fields"]
fn regression_modulo_fields_brillig() {
    use crate::brillig::BrilligOptions;
    use std::panic::AssertUnwindSafe;

    let src = "
        brillig(inline) predicate_pure fn main f0 {
          b0(v0: Field, v1: Field):
            v2 = mod v0, v1
            return v2
        }
        ";
    let ssa = Ssa::from_str(src).unwrap();
    ssa.to_brillig(&BrilligOptions::default());
}
```

I would write this test like this. If compiling from ssa to brillig succeeds (i.e. we're doing a bad modulo) then the test will fail. If the test panics with a different message then the test will fail. The test will only succeed if we hit the panic with the message "Attempt to modulo fields".

TomAFrench (approved):
LGTM

PR #8106, opened 2025-04-16
chore: simpler `make_mutable` in `array_set` optimization

asterite:
# Description

## Problem

None, just something I noticed.

## Summary

While looking at existing SSA optimizations I noticed that in `array_set` we change some instructions. These are modified directly through the dfg but the code was traversing all blocks, taking instructions, etc., and there's no need to do that.

It likely won't improve performance in a noticeable way but at least it makes the code shorter and clearer.

## Documentation

Check one:

# PR Checklist


jfecher (approved):
:+1:

PR #8111, opened 2025-04-16
feat: optimize get_value_max_num_bits to take into account multiplications

asterite:
# Description

## Problem

No issue, just a common operation I noticed in the bignum library that we could maybe optimize.

## Summary

If we have code like this:

```noir
fn main(x: u1, z: u32) -> pub u32 {
    let m = x as u32;
    z * m
}
```

then we don't check for overflow in that multiplication because we know `m` was casted from `u1` which can only be 0 or 1.

If we have code like this:

```noir
fn main(x: u1, y: u1, z: u32) -> pub u32 {
    let x = x as u32;
    let y = y as u32;
    let m = x * y;
    z * m
}
```

then we shouldn't need to check for overflow either as we know `x * y` can be 1 or 0. However, there was no such check in our code so that's what this PR does.

## Additional Context

I also made the cast case recurse so that if you cast a multiplication to `u32`, we'll check the max bitsize of that multiplication according to these rules.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/dfg.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/dfg.rs
@@ -540,7 +540,7 @@ impl DataFlowGraph {
                 let value_bit_size = self.type_of_value(value).bit_size();
                 match self[instruction] {
                     Instruction::Cast(original_value, _) => {
-                        let original_bit_size = self.get_value_max_num_bits(original_value);
+                        let original_bit_size = self.type_of_value(original_value).bit_size();
                         // We might have cast e.g. `u1` to `u8` to be able to do arithmetic,
                         // in which case we want to recover the original smaller bit size;
                         // OTOH if we cast down, then we don't need the higher original size.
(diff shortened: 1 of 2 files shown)


asterite:
Hmm... it seems this destroys performance and programs in test_programs didn't change... maybe it's not a good optimization after all 😅

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -537,14 +538,22 @@ impl DataFlowGraph {
...
-                    let original_bit_size = self.type_of_value(original_value).bit_size();
-                    // We might have cast e.g. `u1` to `u8` to be able to do arithmetic,
-                    // in which case we want to recover the original smaller bit size;
-                    // OTOH if we cast down, then we don't need the higher original size.
-                    value_bit_size.min(original_bit_size)
-                } else {
-                    value_bit_size
+                match self[instruction] {
+                    Instruction::Cast(original_value, _) => {
+                        let original_bit_size = self.get_value_max_num_bits(original_value);
```
jfecher, line 543: Could it be this recursive call that is affecting compile speeds?

asterite: Good point! I'll try...

asterite:
Closing in favor of #8113 to get clean reports.

PR #8113, opened 2025-04-16
feat: optimize get_value_max_num_bits to take into account multiplications

asterite:
Opened another PR to get clean reports. It seems the performance is okay now, though we don't recursively optimize.

# Description

## Problem

No issue, just a common operation I noticed in the bignum library that we could maybe optimize.

## Summary

If we have code like this:

```noir
fn main(x: u1, z: u32) -> pub u32 {
    let m = x as u32;
    z * m
}
```

then we don't check for overflow in that multiplication because we know `m` was casted from `u1` which can only be 0 or 1.

If we have code like this:

```noir
fn main(x: u1, y: u1, z: u32) -> pub u32 {
    let x = x as u32;
    let y = y as u32;
    let m = x * y;
    z * m
}
```

then we shouldn't need to check for overflow either as we know `x * y` can be 1 or 0. However, there was no such check in our code so that's what this PR does.

## Additional Context

I also made the cast case recurse so that if you cast a multiplication to `u32`, we'll check the max bitsize of that multiplication according to these rules.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/dfg.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/dfg.rs
@@ -540,7 +540,7 @@ impl DataFlowGraph {
                 let value_bit_size = self.type_of_value(value).bit_size();
                 match self[instruction] {
                     Instruction::Cast(original_value, _) => {
-                        let original_bit_size = self.get_value_max_num_bits(original_value);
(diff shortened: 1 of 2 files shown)


asterite:
Actually, nevermind... without recursion then it would only optimize multiplying two `u1` which isn't common...

Issue #6826, opened 2024-12-16
Add SSA interpreter for debugging


jfecher:
### Problem

SSA can be difficult to test:
- The output of SSA programs cannot be automatically tested without converting to brillig or acir and bringing in their runtimes.
- SSA programs can often be quite large making it difficult to go through code manually.
- Per the last point, there are many SSA passes and it can be difficult when output is large to pinpoint exactly which one SSA output changes (if there is one)
  - Often when there is an SSA bug which changes the output of a program, it is a bug that originates from one pass. To find which pass though, the developer must comb through the SSA dump of every pass in order until they find the one that changes the output. This can be difficult if the program is large or even just a medium size.

### Happy Case

We add a debug interpreter for SSA programs which can be run on the result of every SSA pass and flag any for which the result changes on.

We should add an SSA interpreter so that we can debug SSA code or automatically test it more easily. This should also aide in auditing since auditors will not need to touch or be concerned with code outside of SSA to test the SSA passes alone. This interpreter should serve as the source of truth of what executed SSA should do.

Additionally, the interpreter could be run after each pass to ensure the expected result does not change after each pass, or catch when it does change.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

aakoshh:
> Additionally, the interpreter could be run after each pass to ensure the expected result does not change after each pass, or catch when it does change.

One aspect of SSA passes that could work against realising this is that some instructions in that cause constraint failures in pass N can be eliminated in pass N+1. In those cases they could mask problems with the return value, because by the time it's possible to run the SSA to the end, the return value could have been borked.

PR #8115, opened 2025-04-16
feat(testing): Add SSA interpreter for testing SSA

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/6826

## Summary\*

Implements an SSA interpreter purely for debugging purposes.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
The interpreter is currently in a very incomplete yet semi-working state so I thought I'd put this up to see if anyone had any comments or suggestions.

Going forward I think most of the remaining work will be:
- Implementing each builtin
- Threading errors throughout the interpreter instead of panics
- Ways to invoke the interpreter from the CLI instead of just via unit tests.
  - Run each test_program against the ssa interpreter automatically and ensure values match?
  - CLI option to run the interpreter after each SSA pass to catch if a pass ever changes the result of the program.
  - Debugger-like options for the interpreter? E.g. "give me the value of v203." This one at least is out of scope for this PR.

jfecher (commented):
This PR should be ready for review now. Some things to note:
- ~2.3k lines of this PR are just tests in `ssa/interpreter/tests/mod.rs` and `ssa/interpreter/tests/instructions.rs`.
- The interpreter currently still panics instead of returning error variants on error. We should update this later but I'm considering this ok for an MVP for testing.
- A few intrinsics are still not implemented but again I'd rather leave them out of the MVP. These are:
  - `Intrinsic::ApplyRangeConstraint` we could just apply a range constraint but there's no documentation on any of the intrinsics so I'm not sure argument ordering, etc.
  - `Intrinsic::BlackBox(_)` all black box functions are unimplemented
  - `Intrinsic::Hint(_)` unsure what to do here
  - `DerivePedersenGenerators` this one was just a lot of code
  - We should implement these in follow-ups. I'll make an issue.
- There are some necessary differences between this and SSA. Notably `assert_constant` is always true. I had it failing before on non-constant ValueIds but this meant we'd fail the same SSA on different optimization levels which is undesirable. `static_assert` is similar but just functions exactly like `assert` now.
- The interpreter can still be used for unit testing and can be easily fed inputs and outputs. Constrains in the code are still asserted, etc.
- I removed the CLI code from this PR, we can add this in a later PR so that it can be invoked via e.g. `nargo interpret-ssa`. Embedding it in other commands like `nargo execute` isn't ideal since we'd need to pass the program arguments down to `noirc_evaluator`.

compiler/noirc_evaluator/src/ssa/interpreter/tests/instructions.rs
```diff
@@ -0,0 +1,840 @@
...
+    ",
+    );
+    assert_eq!(value, Value::Numeric(NumericValue::I32(10000)));
+}
+
+/// TODO: Replace panic with error
+#[test]
+#[should_panic(expected = "called `Option::unwrap()` on a `None` value")]
+fn sub_underflow() {
+    executes_with_no_errors(
```
aakoshh, line 80: This is a bit confusing. Maybe use `expect_error` instead, so the `should_panic` and `executes_with_no_errors` don't contradict each other? Same for the other tests where `should_panic` is used below.

compiler/noirc_evaluator/src/ssa/interpreter/tests/instructions.rs
```diff
@@ -0,0 +1,840 @@
...
+    let value = expect_value(
+        "
+        acir(inline) fn main f0 {
+          b0():
+            v0 = shl u8 3, u32 9
+            return v0
+        }
+    ",
+    );
+    assert_eq!(value, Value::from_constant(12_u128.into(), NumericType::unsigned(8)));
```
aakoshh, line 320: Copy pasted from the test above it?

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
```diff
@@ -0,0 +1,922 @@
...
+    }
+}
+
+type IResult<T> = Result<T, RuntimeError>;
+type IResults = IResult<Vec<Value>>;
+
+#[allow(unused)]
+impl Ssa {
+    pub(crate) fn interpret(&self) -> IResults {
+        self.interpret_function(self.main_id, Vec::new())
```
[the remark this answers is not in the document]
jfecher: I've been using `interpret_function` in that case but can change it

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
```diff
@@ -0,0 +1,922 @@
...
+        results: &[ValueId],
+    ) -> Result<(), RuntimeError> {
+        match instruction {
+            Instruction::Binary(binary) => {
+                let result = self.interpret_binary(binary)?;
+                self.define(results[0], result);
+            }
+            // Cast in SSA changes the type without altering the value
+            Instruction::Cast(value, numeric_type) => {
+                let field = self.lookup(*value).as_numeric().unwrap().convert_to_field();
```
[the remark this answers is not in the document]
jfecher: I think this would be good to implement for https://github.com/noir-lang/noir/issues/8284

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
```diff
@@ -0,0 +1,922 @@
...
+                let rhs = self.lookup(*rhs);
+                if self.side_effects_enabled() && lhs != rhs {
+                    panic!("Constrain {lhs} == {rhs} failed!");
+                }
+            }
+            Instruction::ConstrainNotEqual(lhs, rhs, constrain_error) => {
+                let lhs = self.lookup(*lhs);
+                let rhs = self.lookup(*rhs);
+                if self.side_effects_enabled() && lhs == rhs {
+                    panic!("Constrain {lhs} != {rhs} failed!");
```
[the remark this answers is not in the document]
jfecher: Not too bad hopefully - see https://github.com/noir-lang/noir/issues/8284. This work has been going on long enough though that I wanted to separate out this change since the interpreter is at least usable now.

compiler/noirc_evaluator/src/ssa/interpreter/tests/instructions.rs
jfecher: `expect_error` won't work on its own currently since we panic instead of returning an error. I intended this to be temporary until https://github.com/noir-lang/noir/issues/8284 but should be able to change `expect_error` to use `catch_unwind` in the meantime. I'll update it.

jfecher: Unfortunately, we can't change this currently. I tried adding a catch_unwind to expect_error but `Ssa` contains a frontend Type in its error selector field which makes it unsafe to pass a catch unwind boundary so Rust errors.
For now we'll need to keep the `should_panic` versions and when error variants are added we can move to `expect_error`

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
```diff
@@ -0,0 +1,922 @@
...
+                    U128(value) => U128(value.checked_shl(rhs).expect(overflow_msg)),
+                    I8(value) => I8(value.checked_shl(rhs).expect(overflow_msg)),
+                    I16(value) => I16(value.checked_shl(rhs).expect(overflow_msg)),
+                    I32(value) => I32(value.checked_shl(rhs).expect(overflow_msg)),
+                    I64(value) => I64(value.checked_shl(rhs).expect(overflow_msg)),
+                }
+            }
+            BinaryOp::Shr => {
+                let rhs = rhs.as_u32().expect("Expected rhs of shr to be a u32");
+                let overflow_msg = "Overflow when evaluating `shr`, `rhs` is too large";
```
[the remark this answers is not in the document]
jfecher: Yeah, when writing this I didn't remember us overflowing here but the Rust method did so I figured we wanted that behavior

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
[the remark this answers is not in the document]
aakoshh: I think it would make sense that `interpret` has the `main` endpoint, and being a public function it caters for real-world SSA with args. Maybe `interpret_no_args` could be one to cater for tests?

compiler/noirc_evaluator/src/ssa/interpreter/tests/instructions.rs
aakoshh: Actually all I really meant was whether it would work like this:
```rust
[should_panic(expected = "called `Option::unwrap()` on a `None` value")]
fn sub_underflow() {
    expect_error(
```
that is, which is just a nit to say that "should panic" contradicts "expect no errors" in spirit, and I thought "expect error" could communicate better that in the future, when `should_panic` is removed, we still expect an error, but a runtime one, not a panic.

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
```diff
@@ -0,0 +1,922 @@
...
+        let else_value = self.lookup(else_value);
+
+        // Note that `then_condition = !else_condition` doesn't always hold!
+        // Notably if this is a nested if expression we could have something like:
+        //   then_condition = outer_condition & a
+        //   else_condition = outer_condition & !a
+        // If `outer_condition` is false, both will be false.
+        assert!(!then_condition || !else_condition);
+
+        let new_result = if then_condition { then_value } else { else_value };
```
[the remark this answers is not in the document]
jfecher: I do not know honestly. IfElse never makes it all the way to acir and I don't think there's any way to inspect a completely disabled value like that in normal code. I'll try the following in the SSA executor:
```
enable_side_effects u1 0
v0 = if ... then 100 else 200
enable_side_effects u1 1
call print(v0)
```

compiler/noirc_evaluator/src/ssa/interpreter/tests/instructions.rs
jfecher: Oh, I was thinking the opposite. That "should_panic" would contradict "expect error" because it'd imply you're expecting an error, didn't find one, and panicked.

aakoshh: Ah okay, never mind then 😄

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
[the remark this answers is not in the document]
jfecher: Following the linked discussion (https://github.com/noir-lang/noir/pull/7509) in your ticket I've updated this to return 0 on shr overflow now to match our current behavior. Still not sure the reasoning given there is the best but at least this matches for now.

compiler/noirc_evaluator/src/ssa/interpreter/tests/instructions.rs
jfecher: Ah - just updated this anyway :man_shrugging:

aakoshh (approved):
Awesome :) 

This is going to be be a great reference in itself for what the behaviour should be 👍

PR #8118, opened 2025-04-17
feat(greybox_fuzzer): Should_fail and should_fail_with

Rumata888:
- Updated error messages in the lexer to include new fuzz attribute forms.
- Modified parser to recognize and handle `should_fail` and `should_fail_with` attributes.
- Added new test programs to validate the behavior of these attributes.
- Enhanced the fuzzer to support expected failure scenarios based on the new attributes
- Added parsing of partial witness to detect triggered boolean conditions under "should_fail\*" scenarios

# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*


tooling/greybox_fuzzer/src/coverage.rs
```diff
@@ -532,6 +516,14 @@ impl AccumulatedFuzzerCoverage {
     ) {
         // Insert all ACIR states and replace testcase association
         for acir_bool_state in new_coverage.acir_bool_coverage.iter() {
+            // If the witness is non-boolean according the merged non-bool witness list, skip it
+            if self
+                .non_bool_witness_list
+                .witness
+                .contains(&Witness::new(acir_bool_state.witness_id))
```
asterite, line 523: Extracting this argument to a local variable reduces the number of lines a bit.

tooling/greybox_fuzzer/src/lib.rs
```diff
@@ -271,10 +274,13 @@ pub struct FuzzedExecutorExecutionConfiguration {
...
+    /// Fail on specific asserts (when you use `fuzz(only_fail_with = "...")`)
     pub fail_on_specific_asserts: bool,
 
-    /// Failure reason
+    /// We expect the fuzzed program to always fail (when you use `fuzz(should_fail)` or `fuzz(should_fail_with = "...")` )
+    pub should_always_fail: bool,
+
+    /// Failure reason (when you use `fuzz(should_fail_with = "...")` or `fuzz(only_fail_with = "...")`)
     pub failure_reason: Option<String>,
 }
```
asterite, line 285: I wonder if it wouldn't be better to have this configuration be an enum, similar to `FuzzingScope`, but maybe with four variants: `None`, `OnlyFailWith`, `ShouldFail` and `ShouldFailWith`. Then in `single_fuzz_...` we could pattern match each of these and handle them separately. Mainly because right now we get the failure reason and then assume one of the other booleans is true, adding `unreachable!` otherwise. I think with pattern-matching we wouldn't have an unreachable state and it might also lead to clearer code (right now we are checking these booleans many times).

tooling/greybox_fuzzer/src/lib.rs
```diff
@@ -1023,6 +995,14 @@ impl<
         // Parse results
         match (result_acir, result_brillig) {
             (Ok(witnesses), Ok((_map, brillig_coverage))) => {
+                // If we expect the program to always fail, we need to return a counter example
```
asterite, line 998: I just noticed that this block is duplicated in two more places. Well, not exactly duplicated but it's almost exactly like the other two cases. I wonder if a function couldn't be extracted where `coverage` can be passed as an `Option`, together with the durations (passing 0 when we don't have one, like you are doing now).

tooling/greybox_fuzzer/src/coverage.rs
Rumata888: Done

tooling/greybox_fuzzer/src/lib.rs
Rumata888: Changed

tooling/greybox_fuzzer/src/lib.rs
Rumata888: Moved it out

tooling/greybox_fuzzer/src/lib.rs
```diff
@@ -996,22 +992,28 @@ impl<
...
+                    // If we expect the program to always fail, we need to return a counter example, if it actually passes
+                    FuzzedExecutorFailureConfiguration::ShouldFail
+                    | FuzzedExecutorFailureConfiguration::ShouldFailWith(_) => {
+                        HarnessExecutionOutcome::CounterExample(CounterExampleOutcome {
+                            case_id: testcase.id(),
+                            exit_reason: "".to_string(),
+                            counterexample: testcase.value().clone(),
+                        })
+                    }
+                    _ => {
```
asterite, line 1005: Just in case, I would explicitly list the other two cases here in case we add more in the future.

asterite (approved):
Looks great! I left a final comment.

tooling/greybox_fuzzer/src/lib.rs
Rumata888: Yes, thanks for noticing. Changed everywhere

Issue #8081, opened 2025-04-15
Compiler crash inlining recursive lambda


aakoshh:
### Aim

Trying to replicate a crash from https://github.com/noir-lang/noir/pull/8048

Before that PR I generated ASTs where Brillig called ACIR functions, and one of the random programs hit the recursion limit of 1000 inlining calls. We do collect `InlineInfos` which should mark recursive functions as such, but these are only consulted when we're calling Brillig functions; if the callee is ACIR, it will be inlined. 

The monomorphizer normally doesn't produce such code; it will duplicate ACIR functions as Brillig, in which case the `InlineInfos` should prevent this from happening. However, I realised I knew one edge case where we can still make Brillig call ACIR, which is lambdas; see https://github.com/noir-lang/noir/issues/7289

The following code triggers this situation:

```rust
fn main() {
    unsafe {
        foo(bar);
    }
}

unconstrained fn foo<Env>(f: fn[Env]()) {
    f();
    f();
}

fn bar() {
    bar();
    bar();
}
```

The instructions are duplicated so we don't go into the same route as https://github.com/noir-lang/noir/issues/8078

### Expected Behavior

I expected it to crash actually. Not sure what we can do, I suppose we would have to turn the called function into Brillig on the fly, because then we would _not_ have to inline it. Otherwise we would need to fix the story about passing around lambdas as a tuple.

This is contrived also because the infinite recursion happens purely in the ACIR function, which would crash the compiler if it was called from ACIR anyway.

### Bug

```console
bash-5.2$ cargo run -q -p nargo_cli -- compile --skip-brillig-constraints-check --show-ssa
...

After Preprocessing Functions (1):
acir(inline) fn main f0 {
  b0():
    call f1(Field 2)
    return
}
brillig(inline) fn foo f1 {
  b0(v0: Field):
    call f2()
    call f2()
    return
}
acir(inline) fn bar f2 {
  b0():
    call f2()
    call f2()
    return
}

The application panicked (crashed).
Message:  Attempted to recur more than 1000 times during inlining function 'bar':
acir(inline) fn bar f2 {
  b0():
    call f2()
    call f2()
    return
}
Location: compiler/noirc_evaluator/src/ssa/opt/inlining.rs:283
```

I also printed the following `InlineInfos`:

```
f0: InlineInfo { is_brillig_entry_point: false, is_acir_entry_point: true, is_recursive: false, should_inline: false, weight: 0, cost: 0 }
[...]

PR #8127, opened 2025-04-17
fix(ssa): Do not inline simple recursive functions

### LINKED ISSUE #8078: Compiler crash inlining simple recursive function
### Aim

Trying to replicate something from https://github.com/noir-lang/noir/pull/8048

The original error was more involved, including going between ACIR to Brillig and back, but when trying to replicate I found another bug. (What I was trying to replicate is that when dealing with mutual recursion, only the first function that appears repeatedly in the call stack is marked as recursive, the one calling it gets a cost. I'm trying to find a way to exploit this). 

The following code crashes the compiler:

```rust
fn main() {
    unsafe {
        foo();
    }
}

unconstrained fn foo() {
    bar();
}

unconstrained fn bar() {
    bar();
}
```

### Expected Behavior

I thought the program should compile, because we prepare `InlineInfos` and there which should classify `bar` as recursive, and not try to inline it into `foo`. 

### Bug

```console
bash-5.2$ RUST_BACKTRACE=1 cargo run -q -p nargo_cli -- compile --skip-brillig-constraints-check
Message:  Attempted to recur more than 1000 times during inlining function 'bar':
brillig(inline) fn bar f2 {
  b0():
    call f2()
    return
}
Location: compiler/noirc_evaluator/src/ssa/opt/inlining.rs:280

                                ⋮ 13 frames hidden ⋮                              
  14: noirc_evaluator::ssa::opt::inlining::InlineContext::inline_function::h000ed29d636ddb49
      at /Users/aakoshh/Work/aztec/noir/compiler/noirc_evaluator/src/ssa/opt/inlining.rs:280
...
  4033: noirc_evaluator::ssa::opt::inlining::<impl noirc_evaluator::ssa::ssa_gen::program::Ssa>::inline_simple_functions::h3e69d7e38395438e
      at /Users/aakoshh/Work/aztec/noir/compiler/noirc_evaluator/src/ssa/opt/inlining.rs:118
...
```

The problem is that `inline_simple_functions` tries to inline functions regardless of what `InlineInfos` would say, which is only considered in `inline_functions` and `inline_functions_with_no_predicates`. It inlines anything that has 1 instruction; indeed changing the program as follows circumvents the bug:

```rust
unconstrained fn bar() {
    bar();
    bar();
}
```

### To Reproduce

See above.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+b9cadfe9a31fc848d94fa6b597a65b5e357d4d7f (git version hash: b8b7eb47edf9df7441c93be9147c6db4bcf5bace, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #8078

## Summary\*

We were not taking into account whether the single instruction a single function has may be a recursive call.

I chose to add a special check for this inside of `inline_simple_functions` rather than computing the `InlineInfos` as done in `inline_functions` and `inline_functions_with_no_predicates`. The simple functions check can avoid running through every function to compute the inline infos when we can just directly check the single instruction that we have. 

## Additional Context

I added a test similar to what is in #8081 but that test is failing due to different reasons than #8078. If we were appropriately marking `bar` as unconstrained input it would not be appropriately inlined and execute with a stack overflow as expected.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/inlining.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/inlining.rs
@@ -94,6 +94,9 @@ impl Ssa {
         self
     }
 
+    /// Inline "simple" functions.
+    /// A simple function being a function with one or less instructions.
+    /// Simple functions are still restricted to not inlined if they are recursive or marked with no predicates.
     pub(crate) fn inline_simple_functions(mut self: Ssa) -> Ssa {
         let should_inline_call = |callee: &Function| {
             if let RuntimeType::Acir(_) = callee.runtime() {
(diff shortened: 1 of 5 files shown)


asterite (approved):
LGTM

PR #8132, opened 2025-04-17
chore(ssa): Test terminator value constant folding and resolve cache for data bus

vezenovm:
# Description

## Problem\*

No issue just something small I noticed while working on other tasks.

~~After https://github.com/noir-lang/noir/pull/8091 the fix in https://github.com/noir-lang/noir/pull/8019 is now redundant.~~

It is in fact not redundant, but necessary for terminator optimizations from constraint folding: https://github.com/noir-lang/noir/pull/8132#issuecomment-2816231179. 

## Summary\*

~~I removed the logic introduced in #8019 to defer to the strategy of resolving IDs introduced in #8091.~~

I have simply added additional comments about why we are resolving the constraint simplification cache for the terminator and data bus when we are also replacing their values from `values_to_replace`.

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
> **tuples_inliner_max**	+4 ❌	**+12.50%**
> **tuples_inliner_min**	+4 ❌	**+12.50%**
> **tuples_inliner_zero**	+4 ❌	**+12.50%**

Interesting, this gives us some minor regressions and improvements. The original failure fixed in #8019 still passed for me locally with this PR. Will investigate these before marking ready for review. We may be missing values to replace somewhere.

vezenovm:
https://github.com/noir-lang/noir/pull/8091 ultimately looks to be missing some logic surrounding when we should replace result IDs for constraints. However, we should only be replacing these constraints under specific predicates and only per block so the fix is going to be a bit more involved.

vezenovm:
Turns out we do in fact miss out on some optimizations with just `replace_values_in_block_terminator`. The resolution of the constraint simplification cache across dominated blocks can let us simplify the CFG further. The value mapping in constant folding is only updated to map old instruction results to new instruction results. However, constraints do not have "results" like other instructions, thus are not included in `values_to_replace`. Adding the constraint simplification directly causes issues as well, as we have to consider the predicate and dom tree when simplifying from constraints.

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -242,10 +242,8 @@ impl<'brillig> Context<'brillig> {
             function.dfg.make_constant(FieldElement::one(), NumericType::bool());
 
         for instruction_id in instructions {
-            if !self.values_to_replace.is_empty() {
```
vezenovm, line 245: We check `!mapping.is_empty()` internally in `Instruction::replace_values`

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -256,11 +254,17 @@ impl<'brillig> Context<'brillig> {
...
 
+        // Map the block terminator, resolving any values in the terminator with the
+        // internal value mapping generated by this pass.
         function.dfg.replace_values_in_block_terminator(block_id, &self.values_to_replace);
         function.dfg.data_bus.replace_values(&self.values_to_replace);
 
         // Map a terminator in place, replacing any ValueId in the terminator with the
         // resolved version of that value id from the simplification cache's internal value mapping.
+        // We need this in addition to the value replacement above in order to take advantage
+        // of constraints that may have advised simplifications..
```
vezenovm, line 265:
```suggestion
        // of constraints that may have advised simplifications.
```

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -1946,4 +1959,62 @@ mod test {
         }
         ");
     }
+
+    #[test]
+    fn constant_fold_terminator_argument_from_constrain() {
+        // The only instructions advising simplifications for v0 are
+        // constrain instructions. We want to make sure that they simplifications
+        // are still used for any terminator arguments.
```
vezenovm, line 1967:
```suggestion
        // constrain instructions. We want to make sure that those simplifications
        // are still used for any terminator arguments.
```

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
asterite: Good point! I think I did it to avoid hitting `function.dfg[instruction_id]` but it seems that's just an array index, not a map lookup, so it shouldn't impact performance.

PR #8133, opened 2025-04-17
fix(acir): Check whether opcodes were laid down for non-equality check before fetching payload locations

### LINKED ISSUE #7372: Unconstrained function call only fails when input is constant
```noir
pub unconstrained fn foo(input: Field) -> Field {
    // moving the `!= 0` from `bar` to here fixes the panic:
    (input == 0) as Field
}

pub fn bar(input: Field) {
    // Safety: test
    let output = unsafe { foo(input) };

    // removing the empty assert message fixes the panic:
    assert(output != 0, "");
}

mod test {
    use crate::bar;

    // this panics:
    #[test]
    fn test_local() {
        bar(0);
    }

    global x_global: Field = 0;

    // this panics:
    #[test]
    fn test_global() {
        bar(x_global);
    }

}

// it works when calling or inlining `bar` in `main` (with `input = 0`)
fn main(input: Field) {
    bar(input)
}
```

### Expected Behavior

Expected the tests to pass, as they do with `nargo` from `noirup`:

```bash
❯ ~/.nargo/bin/nargo --version
nargo version = 1.0.0-beta.2
noirc version = 1.0.0-beta.2+97afa52f5212be2d05af26b9e8dde9c3ea7a1d2e
(git version hash: 97afa52f5212be2d05af26b9e8dde9c3ea7a1d2e, is dirty: false)

❯ ~/.nargo/bin/nargo test

[regression_noir_sort] Running 2 test functions
[regression_noir_sort] Testing test::test_local ... ok
[regression_noir_sort] Testing test::test_global ... ok
[regression_noir_sort] 2 tests passed
```

### Bug

The tests fail with `overflow` errors:

```bash
❯ cargo run test
   Compiling noir_debugger v1.0.0-beta.4 (/home/michael/coding/rust/noir/tooling/debugger)
   Compiling nargo_cli v1.0.0-beta.4 (/home/michael/coding/rust/noir/tooling/nargo_cli)
    Finished `dev` profile [optimized + debuginfo] target(s) in 6.15s
     Running `/home/michael/coding/rust/noir/target/debug/nargo test`
[regression_7372] Running 2 test functions
Message:  attempt to subtract with overflow
Location: compiler/noirc_evaluator/src/acir/acir_context/generated_acir/mod.rs:646

[regression_7372] Testing test::test_local ... FAIL
attempt to subtract with overflow

Message:  attempt to subtract with overflow
Location: compiler/noirc_evaluator/src/acir/acir_context/generated_acir/mod.rs:646

[regression_7372] Testing test::test_global ... FAIL
attempt to subtract with overflow


[regression_7372] Failures:
     test::test_local
     test::test_global

[regression_7372] 2 tests failed
```

Where the panicking line is in [`GeneratedAcir::last_acir_opcode_location`](https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/acir/acir_context/generated_acir/mod.rs#L645C1-L647C6):

```rust
OpcodeLocation::Acir(self.opcodes.len() - 1)
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
… for neq

# Description

## Problem\*

Resolves #7372 

This is the SSA in the leadup to ACIR gen where the overflow is triggered:
<p>

```
After Dead Instruction Elimination (2):
  b0():
    v2 = call f1(Field 0) -> Field
    constrain v2 != Field 0, ""
    return
}
  b0(v0: Field):
    v2 = eq v0, Field 0
    v3 = cast v2 as Field
    return v3
}

After Inlining Brillig Calls (1):
  b0():
    constrain Field 1 != Field 0, ""
    return
}
  b0(v0: Field):
    v2 = eq v0, Field 0
    v3 = cast v2 as Field
    return v3
}

After Removing Unreachable Functions (4):
  b0():
    constrain Field 1 != Field 0, ""
    return
}

After Dead Instruction Elimination (3):
  b0():
    constrain Field 1 != Field 0, ""
    return
}
```

</p>
</details> 

The issue boils down to that inversion (which is used by our neq check internally) can potentially be a no-op in ACIR. As we have a constant non equality check, the first SSA instruction is in fact a no-op. When we then go to write the assertion payload location we use [GeneratedAcir::last_acir_opcode_location](https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/acir/acir_context/generated_acir/mod.rs#L645C1-L647C6). This causes us to subtract with an overflow as we are doing `0 - 1`. 

## Summary\*

I simply now check before fetching a payload location whether any opcodes were actually laid down. If they were not, we skip adding an assertion payload as we know that the neq check passed.

`constrain Field 1 != Field 0, ""` makes it down to ACIR gen and is not simplified during constant folding as it only appears after inlining Brillig calls with constant inputs. This is most likely a rare case but important to handle nonetheless. 

The current logic also means we are adding assertion payloads for any neq var check even when we they are no-ops during ACIR gen. This shouldn't affect very many assertion payloads, but is a nice side benefit either way.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/acir/acir_context/mod.rs
+++ b/compiler/noirc_evaluator/src/acir/acir_context/mod.rs
@@ -499,8 +499,20 @@ impl<F: AcirField, B: BlackBoxFunctionSolver<F>> AcirContext<F, B> {
     ) -> Result<(), RuntimeError> {
         let diff_var = self.sub_var(lhs, rhs)?;
 
+        let old_opcodes_len = self.acir_ir.opcodes().len();
         let _ = self.inv_var(diff_var, predicate)?;
         if let Some(payload) = assert_message {
+            // Non-equality can potentially be a no-op if we have all constant
+            // inputs that we know satisfy the non-equality check.
+            // If a no-op non-equality check were to then add an assertion payload
+            // opcode location based upon `GeneratedAcir::last_acir_opcode_location`,
+            // it would be pointing at the previous opcode location.
+            // This at best leads to a mismatch in assertion payload opcode locations
+            // and at worst an attempt to subtract with overflow if the non-equality
+            // check is the first opcode.
+            if self.acir_ir.opcodes().len() - old_opcodes_len == 0 {
+                return Ok(());
+            }
             self.acir_ir
                 .assertion_payloads
                 .insert(self.acir_ir.last_acir_opcode_location(), payload);
--- a/compiler/noirc_evaluator/src/acir/mod.rs
+++ b/compiler/noirc_evaluator/src/acir/mod.rs
@@ -2856,4 +2856,29 @@ mod test {
 
         assert!(matches!(acvm.solve(), ACVMStatus::Failure::<FieldElement>(_)));
     }
+
+    #[test]
+    fn do_not_overflow_with_constant_constrain_neq() {
+        // Test that we appropriately fetch the assertion payload opcode location.
+        // We expect this constrain neq to be simplified and not lay down any opcodes. 
+        // As the constrain neq is the first opcode, if we do not fetch the last opcode
+        // location correctly we can potentially trigger an overflow. 
+        let src = r#"
+        acir(inline) predicate_pure fn main f0 {
+          b0():
+            constrain Field 1 != Field 0, ""
+            return
+        }
+        "#;
+
+        let ssa = Ssa::from_str(src).unwrap();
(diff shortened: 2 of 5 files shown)


guipublic (approved):
LGTM

PR #7834, opened 2025-03-28
chore: remove unused `predicate` from `Opcode::MemoryOp`

asterite:
# Description

## Problem

It seems the field is unused (always `None`).

## Documentation

Check one:
- [x] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #8134, opened 2025-04-18
chore!: several ACIR serialisation changes

### LINKED ISSUE #7920: Remove `num_bits` from `FunctionInput`
These are more a property of the blackbox function itself so should be stored on there instead. In most cases this is implicit in the definition of the blackbox function (e.g. keccakf1600 will always have 64 bit inputs and outputs) but it needs to be explicit in a few cases (i.e. XOR and AND).

This removes unnecessary degrees of freedom from the ACIR spec and will reduce artifact sizes slightly

guipublic:
# Description

## Problem\*

Related to #7934
Resolves #7920 

## Summary\*
This PR currently combines #7834, #7716 and the removal of num-bits in ACIR.
It is in a draft because may be more serialisation changes will be included, and the changes must also be taken into account on Barrentenberg side.

## Documentation\*

Check one:

# PR Checklist\*


acvm-repo/acir/src/proto/acir/circuit.proto
```diff
@@ -108,9 +108,9 @@ message BlackBoxFuncCall {
     Sha256Compression sha256_compression = 20;
   }
   message AES128Encrypt {
-    repeated FunctionInput inputs = 1;
-    repeated FunctionInput iv = 2;
-    repeated FunctionInput key = 3;
+    repeated ConstantOrWitnessEnum inputs = 1;
+    repeated ConstantOrWitnessEnum iv = 2;
+    repeated ConstantOrWitnessEnum key = 3;
```
aakoshh, line 113: Just to note that this would be a breaking change even in Protobuf. Since we're not using it that's fine, but it's another reason not to bother with deprecating the other field. In fact I wouldn't even bother with `reserved` if we do this.

aakoshh:
I wonder if it would be easier on all sides to keep `FunctionInput` and just remove/deprecate its `num_bits` field:
* That _would_ be a backwards compatible change in Protobuf (not that we're using it, so it doesn't matter that much).
* The model on the BB side wouldn't change by a lot (although removing `.input` from all paths doesn't seem like a big deal either).
* Maybe avoid churn in our own code by keeping a `FunctionInput` as a _newtype_?

TomAFrench:
Thoughts on further serialization changes:

I'm not sure it makes sense to have `ExpressionWidth` on the `Circuit`. There's nothing that guarantees that it's accurate so backends will need to handle the case where the expression width is too large anyway. It's mostly just something for the `CSATTransformer` so we should pass to that directly rather than having it sit on the `Circuit`.
https://github.com/noir-lang/noir/blob/6145aaab2ac7ff370bc439f10844f2ecfc559d7a/acvm-repo/acir/src/circuit/mod.rs#L65-L66


These names being completely dissociated from the underlying ACIR/brillig bytecode is a massive pain when it comes to attaching metrics to the circuits (e.g. gates/brillig reports). I think it's worth attaching these directly to the relevant functions:
https://github.com/noir-lang/noir/blob/6145aaab2ac7ff370bc439f10844f2ecfc559d7a/tooling/noirc_artifacts/src/program.rs#L41-L43

TomAFrench:
@guipublic see above comment

acvm-repo/acir/src/circuit/opcodes/black_box_function_call.rs
TomAFrench, line 1: Can you add a predicate onto the ECDSA blackbox ops as well? #8991

PR #8148, opened 2025-04-21
fix(brillig): SliceRefCount reads from the appropriate pointer

### LINKED ISSUE #7976: `pop_slice_front` can overwrite a slice's reference count with nonsense values
### Aim

A slice's reference count seems to get overwritten with seemingly random values when `pop_slice_front` is called. This remains the case even if the `is_rc_one` check is disabled in brillig gen for that method.

This can be seen when adding a println of the slice's refcount to the slice_regex test:
```noir
impl<let N: u32> Regex for str<N> {
    fn find_match(self, input: [u8]) -> Match {
        let mut leftover = input;
        let mut matches_input = true;
        let self_as_bytes = self.as_bytes();
        for c in self_as_bytes {
            if leftover.len() != 0 {
                let rc = std::mem::slice_refcount(leftover);
                println(f"rc of leftover = {rc}");

                let (first_elem, popped_slice) = leftover.pop_front();
                leftover = popped_slice;
                matches_input &= first_elem == c;
            } else {
                matches_input = false;
            }
        }
        if matches_input {
            Match { succeeded: true, match_ends: self_as_bytes.len(), leftover }
        } else {
            Match { succeeded: false, match_ends: 0, leftover: input }
        }
    }
}
```

### Expected Behavior

The reference count should hover between relatively low values

### Bug

The actual output is:
```
rc of leftover = 0
rc of leftover = 0
rc of leftover = 1
rc of leftover = 32930
Match { succeeded: true, match_ends: 4, leftover: &[] }
rc of leftover = 0
rc of leftover = 0
rc of leftover = 1
rc of leftover = 1
rc of leftover = 33175
Match { succeeded: true, match_ends: 4, leftover: &[] }
rc of leftover = 33178
rc of leftover = 1
rc of leftover = 0
rc of leftover = 1
rc of leftover = 33219
rc of leftover = 33229
Match { succeeded: true, match_ends: 5, leftover: &[] }
rc of leftover = 114
rc of leftover = 33178
rc of leftover = 2
rc of leftover = 0
rc of leftover = 1
rc of leftover = 33284
Match { succeeded: true, match_ends: 6, leftover: &[] }
```
It looks like the reference count of the slice is being overwritten with values near 33000 somewhat often. I'm not sure what the significance of this offset is. There is also a single outlier `114` value which may just be very inefficient reference counting but could also be another overwrite. Interestingly, `--enable-brillig-debug-assertions` does not catch this which (in addition to the random numbers rather than a value near 2^N) makes me say it is an overwrite rather than an overflow.

When narrowing down exactly where the bogus reference count originates, it originates from the pop slice front call:
```
impl<let N: u32> Regex for str<N> {
    fn find_match(self, input: [u8]) -> Match {
        let mut leftover = input;
        let mut matches_input = true;
        let self_as_bytes = self.as_bytes();
        for c in self_as_bytes {
            if leftover.len() != 0 {
                let rc = std::mem::slice_refcount(leftover);
                println(f"rc of leftover = {rc}");

                let (first_elem, popped_slice) = leftover.pop_front(); // popped_slice here is the first instance of the 33k RC
                leftover = popped_slice;
                matches_input &= first_elem == c;
            } else {
                matches_input = false;
            }
        }
        if matches_input {
            Match { succeeded: true, match_ends: self_as_bytes.len(), leftover }
        } else {
            Match { succeeded: false, match_ends: 0, leftover: input }
        }
    }
}
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #7976 

As I looked at the vector_pop_front procedure I was struggling to find anything incorrect. It turns out the issue was with `SliceRefCount` itself using the same code gen as `ArrayRefCount` which was not accounting for that slices are tuples and the slice contents are prefixed by the slice length.

## Summary\*

I have fixed the `SliceRefCount` code gen in Brillig.

I now get this output for the `slice_regex` test when we only do the first match in `main`:
```
rc of leftover = 4
rc of leftover = 3
rc of leftover = 4
rc of leftover = 4
[slice_regex] Circuit witness successfully solved
[slice_regex] Witness saved to target/slice_regex.gz
```

I have also added a `reference_counts_slices_inliner_0` test that would have caught this discrepancy in the ref count implementations.
Comments from the test:
```
This test is exactly the same as `reference_counts_inliner_0` which uses
arrays rather than slices.
This test exists to make sure that our reference counting debug methods match
between arrays and slices.
We could most likely combine the code for these tests (e.g. using generics),
but it is simpler to debug isolated tests.
It should only be necessary to have a test at one inliner setting, as we
are just checking for discrepancies between the array and slice debugging builtin functions.
The actual functionality of reference counting is tested with the `reference_counts_*` tests.
We went with testing at an inliner aggressiveness of zero, as this is generally
the most useful inliner setting for unconstrained functions.
```

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
LGTM

PR #8099, opened 2025-04-15
fix(ssa): Recursive shared Brillig entry points 

vezenovm:
# Description

## Problem\*

Resolves #8069 

For more context surrounding what the Brillig entry point analysis pass is doing and why you can reference https://github.com/noir-lang/noir/pull/7277 and the doc comments on the pass itself. 

The current SSA for the new `brillig_entry_points_regression_8069` test:
<details><summary>Details</summary>
<p>

```
After Array Set Optimizations (1):
acir(inline) impure fn main f0 {
  b0():
    v4 = call f1(u1 1, i8 1, u32 5) -> u1
    v5 = not v4
    enable_side_effects v5
    v8 = call f2(Field 1, u1 1, u32 5) -> [i8; 4]
    enable_side_effects u1 1
    return u1 0
}
brillig(inline) impure fn func_1 f1 {
  b0(v0: u1, v1: i8, v2: u32):
    v5 = eq v2, u32 0
    jmpif v5 then: b1, else: b2
  b1():
    jmp b3(u1 0)
  b2():
    v7 = sub v2, u32 1
    v8 = cast v0 as Field
    v11 = call f2(v8, u1 0, v7) -> [i8; 4]
    v13 = array_get v11, index u32 3 -> i8
    v15 = call f1(u1 0, v13, v7) -> u1
    jmp b3(v15)
  b3(v3: u1):
    return v3
}
brillig(inline) impure fn func_2 f2 {
  b0(v0: Field, v1: u1, v2: u32):
    v5 = eq v2, u32 0
    jmpif v5 then: b1, else: b2
  b1():
    v17 = make_array [i8 6, i8 101, i8 92, i8 30] : [i8; 4]
    jmp b3(v17)
  b2():
    v7 = sub v2, u32 1
    v10 = call f1(v1, i8 0, v7) -> u1
    v12 = call f2(v0, v10, v7) -> [i8; 4]
    jmp b3(v12)
  b3(v3: [i8; 4]):
    return v3
}

After Brillig Entry Point Analysis (1):
acir(inline) impure fn main f0 {
  b0():
    v4 = call f1(u1 1, i8 1, u32 5) -> u1
    v5 = not v4
    enable_side_effects v5
    v8 = call f2(Field 1, u1 1, u32 5) -> [i8; 4]
    enable_side_effects u1 1
    return u1 0
}
brillig(inline) impure fn func_1 f1 {
  b0(v0: u1, v1: i8, v2: u32):
    v5 = eq v2, u32 0
[...]

PR #8152, opened 2025-04-21
chore(ssa): Do not run Brillig entry point analysis when no globals have been declared

vezenovm:
# Description

## Problem\*

No issue just something I noticed while looking at https://github.com/noir-lang/noir/pull/8099.

We already skip Brillig entry point specialization if we have a single Brillig entry point (e.g. `main` is Brillig). We can also avoid specializing Brillig functions for programs that do not use globals at all.

## Summary\*

Check whether any globals have been declared. If not we can skip the Brillig entry point analysis.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
We've got a panic in `brillig_calls_array` due to "f4 has multiple entry points"

vezenovm:
> We've got a panic in `brillig_calls_array` due to "f4 has multiple entry points"

Yeah, we then need to handle that we did not specialize entry points during Brillig globals code gen. We also hoist constants during the globals code gen so that complicates things a bit further as well. This is looking to be a bit more trouble than it is worth at the moment, so I am going to close.

PR #8161, opened 2025-04-22
chore: remove uneeded references to FileId

asterite:
# Description

## Problem

While working on something else I noticed we were passing `FileId` around and creating `Location` instances from an `Ident`'s span. However, `Ident` now carries a location so in all cases we can take the `Location` from it, and it's actually a bit more accurate.

## Summary

Removes some `FileId` from some structs and methods.

## Documentation

Check one:

# PR Checklist


asterite:
Something broke but I don't understand what. I'll open another PR later on.

PR #8163, opened 2025-04-22
chore: move unsigned overflow check from acir/brillig to ssa

### LINKED ISSUE #8162: chore: move checking of unsigned overflow from acir-gen/brillig-gen to SSA
Right now when we generate ACIR or Brillig, for some checked math operations (add, sub, mul) we check if they could possibly overflow. If not, no range check is inserted after the operation. This works, but we could instead move this logic to an SSA pass that would transform these checked operations to unchecked. Then, when ACIR/Brillig find a checked add/sub/mul they can be sure a range check is needed.

This better separates concerns, plus we can add SSA tests for this (it's kind of tricky to add these tests in ACIR/Brillig, and they would be duplicated).

asterite:
# Description

## Problem

Resolves #8162

## Additional Context

When inserting binary operations into the SSA we currently transform them a bit. For example:
1. if we have `unchecked_mul` between fields we turn that to `mul` because checked/unchecked doesn't make sense for fields
2. if we have `unchecked_mul` between bools we turn that into `mul` because it can never overflow

We did this to reduce the noise in the SSA, so that we don't get to see `unchecked_mul` between fields and wonder why it's unchecked if it doesn't make sense for it to be unchecked.

But... like that, and especially for point 2, it means that boolean multiplication will be checked unless we don't check it specifically for bools.

So... I wonder if instead `unchecked_mul` should be used whenever two values are being multiplied and they can never overflow. That means that `mul` between fields doesn't make sense anymore, and `unchecked_mul` is what makes sense. And the same applies for bools.

But... for now in this PR I just made those checked operations unchecked. 

---

A next step, which I'll try as a follow-up PR, is to try https://github.com/noir-lang/noir/pull/8111 again but this time memoizing the maximum bit size of values as we traverse the instructions (to avoid recursing multiple times, maybe that was the cause of the slowdown).

---

Yet another thought: `unchecked: false` is kind of hard to understand because it's a double negation. I wonder if renaming `unchecked` to `checked` and flipping all bools would make things a bit clearer.

## Documentation

Check one:

# PR Checklist


asterite:
Cool, it went from a regression to an optimization! (not sure why, though, but maybe unchecked operations have more consequences along the way...)

compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -2692,13 +2697,11 @@ mod test {
     }
 
     #[test]
-    fn multiply_with_bool_should_not_emit_range_check() {
+    fn unchecked_mul_should_not_have_range_check() {
```
asterite, line 2700: I changed this test because part of this logic was moved to SSA. However, we can still check that unchecked multiplications don't produce a range check, and there's an SSA test that makes sure this particular case will transform that `mul` to `unchecked_mul`.

compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -1071,11 +1071,16 @@ impl<'a> Context<'a> {
...
         binary: &Binary,
-        dfg: &DataFlowGraph,
         predicate: AcirVar,
     ) -> Result<(), RuntimeError> {
-        let Some(msg) = binary.check_unsigned_overflow_msg(dfg, bit_size) else {
-            return Ok(());
+        let msg = match binary.operator {
+            BinaryOp::Add { unchecked: false } => "attempt to add with overflow",
+            BinaryOp::Sub { unchecked: false } => "attempt to subtract with overflow",
+            BinaryOp::Mul { unchecked: false } if bit_size != 1 => {
```
TomAFrench, line 1079: Why are we not handling this case within SSA?

asterite: I explained it a bit in "Additional Context" though I might have wrote too much 😅 

I pushed a commit for this. I was mainly trying to avoid producing a larger diff, but changing this makes sense.

compiler/noirc_evaluator/src/ssa/opt/checked_to_unchecked.rs
```diff
@@ -0,0 +1,226 @@
...
+          b0(v0: u16):
+            v1 = cast v0 as u32
+            v3 = unchecked_sub u32 65536, v1
+            return v3
+        }
+        ");
+    }
+
+    #[test]
+    fn checked_to_unchecked_when_multiplying_bools() {
```
vezenovm, line 140: nits: let's add this to cspell

compiler/noirc_evaluator/src/ssa/opt/checked_to_unchecked.rs
```diff
@@ -0,0 +1,226 @@
...
+        acir(inline) fn main f0 {
+          b0(v0: u1, v1: u1):
+            v2 = unchecked_mul v0, v1
+            return v2
+        }
+        ");
+    }
+
+    #[test]
+    fn checked_to_unchecked_when_multiplying_upcasted_bool_with_u32() {
```
vezenovm, line 160: nit: cspell

vezenovm (approved):
LGTM, I have wanted to move more of this shared logic into the SSA so happy to see it!

compiler/noirc_evaluator/src/ssa/opt/checked_to_unchecked.rs
asterite: Done!

compiler/noirc_evaluator/src/ssa/opt/checked_to_unchecked.rs
asterite: Done!

PR #8167, opened 2025-04-22
fix(acir): Div/mod for i1

### LINKED ISSUE #7973: `i1 mod i1` behaves different in ACIR than Brillig
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/7890

This program fails in ACIR and passes in Brillig:

```rust
fn main(a: i1) -> pub i1 {
    a % a
}
```

using the following input:

```
a = -1
```

### Expected Behavior

ACIR and Brillig should both succeed, or both fail.

### Bug

```
error: Failed constraint
  ┌─ src/main.nr:2:5
  │
2 │     a % a
  │     -----
  │
  = Call stack:
    1. src/main.nr:2:5

bash-5.2$ 
```

The final SSA is identical in both cases:

```
  b0(v0: i1):
    v1 = mod v0, v0
    return v1
}

```

### To Reproduce

See above.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+cdd4da85b7713e64aa8ff5e712c312978fb4d02e (git version hash: 54d7cf108f636068b93285e5f5319c7457e30b26, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #7973 

## Summary\*

From comments in the PR:
```
When dividing the twos complement lhs and rhs, we subtract one from the bit size
to accurately represent the unsigned values we are dividing.
We then internally constrain `q < 2^{max_q_bits}` during euclidean division.
If we do not special case for binary types of one bit, we will range constrain the resulting
quotient to be zero bits, which will cause inadvertent failures.
We could theoretically still use this algorithm as is by incrementing the bit size by one,
but the special case is very simple and then we don't mix the degenerate type
into our general signed division algorithm.
```
i1 division:
```
We know that a signed integer of bit size one can only be -1 or 0
Thus, we know the only valid signed division can be 0 / -1.
0 / -1, valid as the quotient is 0
-1 / -1, invalid as the quotient is 1, which cannot be represented in an i1
All divisions by zero are invalid
```
i1 modulo:
```
We know that a signed integer of bit size one can only be -1 or 0
Thus, we know the only valid signed modulo can be 0 % -1 or -1 % -1
0 / -1, valid as the remainder is 0
-1 / -1, valid as the remainder is 0
All divisions by zero are invalid
```

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/acir/acir_context/mod.rs
```diff
@@ -1029,6 +1029,30 @@ impl<F: AcirField, B: BlackBoxFunctionSolver<F>> AcirContext<F, B> {
...
+        // Thus, we know the only valid signed division can be 0 / -1.
+        // 0 / -1, valid as the quotient is 0
+        // -1 / -1, invalid as the quotient is 1, which cannot be represented in an i1
+        // All divisions by zero are invalid
+        if bit_size == 1 {
+            let lhs_is_zero = self.eq_var(lhs, zero)?;
+            // The value is represented internally as one with its sign stored separately
+            // As we are within signed division we can just check whether we have one here.
+            let rhs_is_one = self.eq_var(rhs, one)?;
+            let both_true = self.and_var(lhs_is_zero, rhs_is_one, AcirType::unsigned(1))?;
```
aakoshh, line 1051: Shouldn't this be `AcirType::signed`?

vezenovm: I am comparing `lhs_is_zero` and `rhs_is_one` which should be two booleans. We could also just do a multiplication here, and in fact that is what `and_var` will do internally for bit size one, I just thought this read more clearly with `and`.

aakoshh: Ah, of course! 😵‍💫

aakoshh (approved):
Great, the detailed proof/explanation is super helpful 👏

TomAFrench (changes requested):
We should be blocking the i1 type.

vezenovm:
> We should be blocking the i1 type.

Wasn't sure if we just wanted to block i1. Even after blocking i1 in the frontend, would we still not want this logic as to prevent any differences between ACIR and Brillig?

TomAFrench:
I can have a think about this tomorrow but I'm pretty sure that SSA can create arbitrary bitsize types which gives opportunities for all kinds of funky behaviour. We could also just disallow creating `i1`s in SSA too.

vezenovm:
> We could also just disallow creating `i1`s in SSA too

This would be simplest. Otherwise I do think we probably need to handle i1 as done here.

PR #8184, opened 2025-04-23
fix: don't take implicitly added named generics when checking where clauses

### LINKED ISSUE #8177: Inclusion of traits with associated constants in trait constraints on trait methods results in ` impl has stricter requirements than trait` error
```noir
trait Bar {
    let N: Field;
}

impl Bar for Field {
    let N: Field = 42;
}

trait Foo {
    fn foo<B>(b: B) where B: Bar; 
}

impl Foo for Field{
    fn foo<B>(_: B) where B: Bar {} 
}
```

I expect that this would compile fine but I instead get the below error

```
% nargo check                                                                                                                                                                                                                                                         ~/noir/temp master tom-box
error: impl has stricter requirements than trait
   ┌─ src/main.nr:10:8
   │
10 │     fn foo<B>(b: B) where B: Bar; 
   │        --- definition of `foo` from trait
   ·
14 │     fn foo<B>(_: B) where B: Bar {} 
   │                              --- impl has extra requirement `B: Bar<N = <B as Bar>::N>`
   │

```

It seems like we're being too strict with the associated constants here but I'd need to check against Rust's behaviour.

asterite:
# Description

## Problem

Resolves #8177

## Summary

In a code like this:

```noir
trait Bar {
    let N: Field;
}

impl Bar for Field {
    let N: Field = 42;
}

trait Foo {
    fn foo<B>(b: B) where B: Bar; 
}

impl Foo for Field{
    fn foo<B>(_: B) where B: Bar {} 
}
```

in the past, the compiler would fail to compile the above just because `B: Bar` would error saying that it's missing `N` from `Bar`'s generics (that is, `Bar<N = ...>`). So in #7026 this was now allowed, and `N` was added implicitly as an unbound named generic.

When comparing where clauses, these are compared by equality (after binding substitution). The issue was that the implicitly added named generic in the first `foo` ended up with a type variable that's different than the one used for the second `foo`. Then, they aren't equal.

What's done in this PR, and I don't know if it's the correct way to solve this, was to mark these implicitly added named generics as such. Then, two unbound implicitly added named generics are considered equal, regardless of their type variable ID. For this I extracted a new `NamedGeneric` type instead of adding another argument to `NamedGeneric(..., /* implicit /* bool)` for clarity.

## Documentation

Check one:

# PR Checklist


TomAFrench (approved):

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -3170,7 +3190,16 @@ impl std::hash::Hash for Type {
                 alias.hash(state);
                 args.hash(state);
             }
-            Type::TypeVariable(var) | Type::NamedGeneric(var, ..) => var.hash(state),
+            Type::NamedGeneric(NamedGeneric { type_var, implicit: true, .. }) => {
+                // An implicitly added unbound named generic's hash must be the same as any other
+                // implicitly added unbound named generic's hash.
+                if !type_var.borrow().is_unbound() {
+                    type_var.hash(state);
+                }
```
jfecher, line 3198: Does it though?
We issue implicit NamedGenerics for associated types so this seems like it'd consider the generics from two separate associated types as equal.

asterite: Good point.

Originally I fixed this issue in a different way: when comparing a method where clause in a impl against its trait, I removed all implicit generics from the comparison (there's an equality comparison somewhere there). But then I thought I could also fix it by changing the equality... maybe I should have done the former?

PR #8185, opened 2025-04-23
fix: Use `get_local_or_global_instruction` when simplifying `IfElse`

### LINKED ISSUE #8174: Compiler crash in `simplify.rs`: `index out of bounds`
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8173

```rust
global G_A: (str<3>, [str<3>; 2]) = ("AWQ", ["LNV", "ZIH"]);
unconstrained fn main(a: [str<3>; 2], b: bool) -> pub [[str<3>; 2]; 1] {
    [if (!b) {
        a
    } else {
        G_A.1
    }]
}
```

### Expected Behavior

### Bug

```console
bash-5.2$ cargo run -q -p nargo_cli -- compile --force
Message:  index out of bounds: the len is 3 but the index is 3
Location: compiler/noirc_evaluator/src/ssa/ir/dfg/simplify.rs:233
```

### To Reproduce

See above.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+b9cadfe9a31fc848d94fa6b597a65b5e357d4d7f (git version hash: 8e8b07b5e55d2819112d193e4fe4dc62c6574fe8, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #8174 

## Summary\*

Changes the way `IfElse` is simplified so that the `then_value` and `else_value` are looked up with `DataFlowGraph::get_local_or_global_instruction`, directly returning an `Instruction` from either `dfg` or `dfg.globals`, rather than the two step process where we 1) get the `Value::Instruction` transparently from `dfg` or `dfg.global` and then 2) go to `dfg` to get the instruction.

Thanks @vezenovm for the tip about this method 🙏 

## Additional Context

See https://github.com/noir-lang/noir/issues/8174#issuecomment-2824549898 for an illustration of how the issue arises: 
1. We see a `ValueId` in the terminator of a block in the SSA, which in this case is a global ID passing an array to the next block. 
2. The `Index` by `ValueId` of `DataFlowGraph` checks if the ID is global. If it is, it's returned from the `GlobalsGraph` instead of `DataFlowGraph::values`. 
3. The value in this case was a `Value::Instruction`, which carries an `InstructionId` that we need to look up, however the ID doesn't say if it's a global or a local instruction. Because it came from a global `ValueId`, it should have been looked up in `dfg.globals`, not `dfg`. 

#### Any others?

I had a brief look at [other instances](https://github.com/search?q=repo%3Anoir-lang%2Fnoir%20%22dfg%5B*ins%22&type=code), looking for other potential places this method could be used. Perhaps `simplify::try_optimize_array_set_from_previous_get`? It doesn't seem to have the short circuit of global arrays considered immutable like `constant_folding::remove_possibly_mutated_cached_make_arrays` does.

## Documentation\*

Check one:

# PR Checklist\*


vezenovm (approved):
Looks good!

aakoshh:
> We shouldn't have any array set instructions acting on global arrays due to their immutable nature.

Yes, what I wasn't sure about is that in `constant_folding` we [check that the instruction is an ArraySet and then skip if it's global](https://github.com/noir-lang/noir/blob/e1ec3d941c3803ec8bbd29ba71f6ced6653d998c/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs#L713C70-L714) (which shouldn't even happen, but it checks it nevertheless), while in [try_optimize_array_set_from_previous_get](https://github.com/noir-lang/noir/blob/e1ec3d941c3803ec8bbd29ba71f6ced6653d998c/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify.rs#L380-L413) this possibility is not entertained, unlike in [try_optimize_array_get_from_previous_set](https://github.com/noir-lang/noir/blob/e1ec3d941c3803ec8bbd29ba71f6ced6653d998c/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify.rs#L313) where it is. As long as you're happy with that, so am I 🫡

vezenovm:
> `constant_folding` we [check that the instruction is an ArraySet and then skip if it's global](https://github.com/noir-lang/noir/blob/e1ec3d941c3803ec8bbd29ba71f6ced6653d998c/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs#L713C70-L714) (which shouldn't even happen, but it checks it nevertheless)

We also are checking whether the instruction is a Store. 

> while in [try_optimize_array_set_from_previous_get](https://github.com/noir-lang/noir/blob/e1ec3d941c3803ec8bbd29ba71f6ced6653d998c/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify.rs#L380-L413) this possibility is not entertained, unlike in [try_optimize_array_get_from_previous_set](https://github.com/noir-lang/noir/blob/e1ec3d941c3803ec8bbd29ba71f6ced6653d998c/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify.rs#L313) where it is

I have thought about switching all instances of `dfg[instruction_id]` with `get_local_or_global_instruction` to prevent the possibility of these kinds of panics. I ultimately went with just using `get_local_or_global_instruction` wherever we may have a global MakeArray instruction as to avoid too many pervasive changes. I wouldn't be against experimenting with making this switch but it should be a separate PR.

PR #8187, opened 2025-04-23
fix(debugger): send idle at loop end + fix test

asterite:
# Description

## Problem

Resolves https://github.com/AztecProtocol/aztec-packages/pull/13755#issuecomment-2823802201

## Summary

This PR has two fixes:
1. The debugger communicated over channels now since the last debugger PR. An "idle" status is expected after sending a command over one channel. The problem (I think) is that the "idle" status was reported **before** commands were processes instead of after.
2. A test was ignoring lines that matched "At {path}", but the debugger will try to use relative paths when possible so when the test ran from the top level, that line didn't match. Trying to use relative paths in the test fixes that (and this is the one that should unblock the merge on Aztec-Packages, but I fixed 1 first and then noticed the test was still failing, so...)

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/debugger/tests/debug.rs
+++ b/tooling/debugger/tests/debug.rs
@@ -144,6 +144,18 @@ mod tests {
             VecDeque::from(["foo(x);", "bar(y);", "fn bar(y: Field) {"]),
             VecDeque::from(["foo(x);", "bar(y);", "assert(y != 0);"]),
         ];
+
+        // Try to use relative paths if possible, otherwise the test breaks when run from the repository root
+        let at_filename = if let Ok(current_dir) = std::env::current_dir() {
+            if let Ok(stripped) = test_program_path.strip_prefix(current_dir) {
+                format!("At {}", stripped.display())
+            } else {
+                format!("At {}", test_program_dir)
+            }
+        } else {
+            format!("At {}", test_program_dir)
+        };
+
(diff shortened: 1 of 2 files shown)


tooling/debugger/tests/debug.rs
```diff
@@ -144,6 +144,18 @@ mod tests {
...
+        // Try to use relative paths if possible, otherwise the test breaks when run from the repository root
+        let at_filename = if let Ok(current_dir) = std::env::current_dir() {
+            if let Ok(stripped) = test_program_path.strip_prefix(current_dir) {
+                format!("At {}", stripped.display())
+            } else {
+                format!("At {}", test_program_dir)
+            }
+        } else {
+            format!("At {}", test_program_dir)
+        };
```
vezenovm, line 157:
```suggestion
        let at_filename = std::env::current_dir()
            .ok()
            .and_then(|dir| test_program_path.strip_prefix(&dir).ok())
            .map(|stripped| format!("At {}", stripped.display()))
            .unwrap_or_else(|| format!("At {}", test_program_dir));
```
nit: we can reduce nesting here with a clear single fallback

vezenovm (approved):

anaPerezGhiglia (approved):
✨ Thanks for the catch and fix 🙇‍♀️

PR #8193, opened 2025-04-23
fix: Detect ABI change of fuzzing harnesses

Rumata888:
# Description

## Problem\*

Changing a fuzzing harness with an existing corpus could lead to panics, if former elements couldn't be cast to appropriate types. 

## Summary\*

Before fuzzing with the starting corpus we now try to decode all testcases with the current ABI. If a testcase fails, then we collect parse it for dictionary elements.
For example, if you first fuzz 

```rust
#[fuzz(should_fail)]
fn fuzz_main(x: Field, y: u64, z: u64) {
    
     let prod=y*z;
     assert(y<0x100000000);
     assert(z<0x100000000);
     assert(prod-0xdeadbeefdeadbeef<0xdeadbeef);
     assert(x==0xdeadbeefdeadbeefdeadbeefdeadbeef);
 }
```
and find a failure (`nargo fuzz`), after changing to 
```rust
#[fuzz(should_fail)]
fn fuzz_main(x: u64, y: u64, z: u64) {
    
    let prod=y*z;
    assert(x<0x100000000);
    assert(y<0x100000000);
    assert(z<0x100000000);
    assert(prod-0xdeadbeefdeadbeef<0xdeadbeef);
}
```
The fuzzer will inform you that the ABI has changed and will almost immediately find a failing testcase, because it already has useful values that it needs to put into `y` and `z`
## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/greybox_fuzzer/src/lib.rs
+++ b/tooling/greybox_fuzzer/src/lib.rs
@@ -24,7 +24,7 @@ mod mutation;
 mod types;
 
 use corpus::{Corpus, DEFAULT_CORPUS_FOLDER, TestCase, TestCaseId};
-use mutation::InputMutator;
+use mutation::{InputMutator, add_elements_from_input_map_to_vector_without_abi};
 use rayon::iter::ParallelIterator;
 use termcolor::{ColorChoice, StandardStream};
 pub use types::FuzzTestResult;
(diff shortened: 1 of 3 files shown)


asterite (approved):
Looks great!

PR #8197, opened 2025-04-23
fix: Do not panic if RHS constant in division has more bits than the operand

### LINKED ISSUE #8195: Compiler crash in `acir_context`: `attempted to divide by constant larger than operand type`
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8048

```rust
global G_A: u16 = 41618;
fn main(a: u16) -> pub u16 {
    ((13834 - a) % (G_A * G_A))
}
```

### Expected Behavior

It shouldn't crash. It compiles with `--force-brillig`.

### Bug

```console
Message:  attempted to divide by constant larger than operand type
Location: compiler/noirc_evaluator/src/acir/acir_context/mod.rs:827
```

Works with `--force-brillig`:
```console
bash-5.2$ cargo run -q -p nargo_cli -- compile --force-brillig
bash-5.2$ 
```
Executing with the following `Prover.toml`:
```
a = 0
```
works and returns an assertion failure:
```console
  ┌─ src/main.nr:3:21
  │
3 │     ((13834 - a) % (G_A * G_A))
  │                     ---------
  │
  = Call stack:
    1. src/main.nr:3:21

Failed assertion
```

### To Reproduce

See above.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+ae008d31e289ddb086b4c69e3fd287c5167fc4dc (git version hash: e5484ec4a9c9e341996f9e4f057fc5a36c450eeb, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #8195 

## Summary\*

Changes `acir_context::euclidean_division_var` to not assert that the RHS, when it's a constant, must not have more bits than the operand type, and instead it lets the runtime handle the problem.

## Additional Context

In the failing program we have the following final SSA:

```
g0 = u16 41618

  b0(v1: u16):
    v3 = sub u16 13834, v1
    v4 = mul u16 41618, u16 41618
    v5 = mod v3, v4
    return v5
}
```
During ACIR generation `convert_ssa_binary` calls `AcirContext::mul_var` to handle `mul u16 41618, u16 41618`, which runs the following:
```rust
            (AcirVarData::Const(lhs_constant), AcirVarData::Const(rhs_constant)) => {
                self.add_constant(lhs_constant * rhs_constant)
            }
```
That adds an `AcirVar` with the value `1732057924`. Then, back in `convert_ssa_binary`, we insert a range constraint in `check_unsigned_overflow`, and move on. When we handle `mod v3, v4`, we retrieve `1732057924` for `v4`, and panic because it wouldn't fit into 16 bits. But at runtime we would have failed the multiplication overflow constraint already.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench (approved):

TomAFrench:
I'm a little concerned as we'll be telling the brillig call that the inputs are a smaller size than they are in reality but we should always error beforehand.

TomAFrench:
Perhaps we should inject a failing constraint at this point just to be sure?

aakoshh:
> I'm a little concerned as we'll be telling the brillig call that the inputs are a smaller size than they are in reality

Yes, that bothers me as well. I was wondering if it would be worth adjusting it, but then again this shouldn't actually be invoked, and I'm not sure it would even make sense or whether it could handle it.

PR #8203, opened 2025-04-24
chore: replace usage of `to_u128` with `try_into_u128`, etc.

TomAFrench:
# Description

## Problem\*

Followup to #8180 

## Summary\*

Having silently truncating conversion functions don't seem to be helping us. This PR removes these and replaces their use with the `try_into_u128` function. We can then unwrap this which will panic if we're using it in an incorrect way.

## Documentation\*

Check one:

# PR Checklist\*


aakoshh (commented):
Why not keep `to_u128` and panic in there?

That way we could provide a more meaningful error message than `attempted to unwrap None` or whatever it will say.

PR #8206, opened 2025-04-24
fix: let `nargo test --show-output` work for tests with arguments

### LINKED ISSUE #8194: `nargo test --show-output` doesn't show output for tests with arguments
### Aim

Given this code:

```noir
#[test]
fn test(byte: u8) {
    println(byte);
}
```

Running `nargo test --show-output` doesn't show any output.

This seems to only happen when the test has arguments.

### Expected Behavior

The output is shown, like in tests without arguments.

### Bug

The output isn't shown.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8194

## Summary

This was tricky because in order to make it work we need `output` to be `Clone`. `std::io::stdout()` isn't `Clone` so I have to think about ways to make it work.

To be honest I'm not sure this is the best solution so let me know if there's a better way to do this.

## Additional Context

- For `BytesWriter` I took some inspiration from https://andrewra.dev/2019/08/05/testing-in-rust-writing-to-stdout/
- Note that for a test with arguments, we'll now get to see the output of each run for each input value combination... maybe each run could be split somehow but I think for now this is fine.

## Documentation

Check one:

# PR Checklist


TomAFrench:
I can take a look at this tomorrow. I think that if we're printing out the results from every iteration then this will be extremely noisy, restricting it to only the last instance would be better if we want to move forwards with this.

asterite:
> I think that if we're printing out the results from every iteration then this will be extremely noisy, restricting it to only the last instance would be better if we want to move forwards with this.

Makes sense. I'll see if I can make it work like that.

asterite:
I got it working!

So now if you have a program like this:

```noir
#[test]
fn foo(x: u8) {
    println(x);
}
```

You get:

```
$ nargo test --show-output
[one] Running 1 test function
[one] Testing foo ... ok
--- foo stdout ---
253
------------------
[one] 1 test passed
```

But if you have this:

```noir
#[test]
fn foo(x: u8) {
    println(x);
    assert(x >= 2);
}
```

you get this:

```
$ nargo test --show-output
[one] Running 1 test function
[one] Testing foo ... FAIL

--- foo stdout ---
0
------------------

[one] Failures:
     foo

[one] 1 test failed
```

or this:

```
$ nargo test --show-output
[one] Running 1 test function
[one] Testing foo ... FAIL

--- foo stdout ---
1
------------------

[one] Failures:
     foo

[one] 1 test failed
```

which is really nice because from a user perspective it always looks like it just ran once with some correct value, or once with an incorrect value.

This implementation will eventually change to use the new fuzzer, but I think this PR is a good warm-up as we'll likely want a similar logic there too (right now if you do `nargo fuzz` you get all outputs).

TomAFrench:
Nice! I'll have a more in-depth look tomorrow but I think we can remove the `Clone` trait constraint now as each fuzz instance gets a buffer to write into and then we copy that into stdout only once.

tooling/fuzzer/src/lib.rs
```diff
@@ -51,36 +56,66 @@ where
...
     /// Fuzzes the provided program.
-    pub fn fuzz(&self) -> FuzzTestResult {
+    pub fn fuzz(&mut self) -> FuzzTestResult {
         let dictionary = build_dictionary_from_program(&self.program.bytecode);
         let strategy = strategies::arb_input_map(&self.program.abi, &dictionary);
 
+        // Each fuzz run produces output from `print` and `println` calls.
+        // We don't print the output of all runs. Instead, two things can happen:
+        //  1. There is a counter-example, in which case we'd like to show the output for that counter-example.
+        //  2. There is no counter-example, in which case we'd like to show just one of the successfull cases.
```
michaeljklein, line 66: typo:
```suggestion
        //  2. There is no counter-example, in which case we'd like to show just one of the successful cases.
```

tooling/nargo/src/ops/test.rs
```diff
@@ -163,9 +164,12 @@ where
                     }
                 };
 
-                let fuzzer = FuzzedExecutor::new(compiled_program.into(), executor, runner);
+                let mut fuzzer = FuzzedExecutor::new(compiled_program.into(), executor, runner);
 
                 let result = fuzzer.fuzz();
+
+                let _ = output.write(&result.output);
```
michaeljklein, line 171: Don't we want to handle this `Result`?

michaeljklein (changes requested):
What about making `std::io::stdout` clonable by wrapping in `Rc`/`RefCell`? I'd expect it to be faster since we'd avoid copying the buffer.

E.g.

```rust
use std::rc::Rc;
use std::cell::RefCell;

#[derive(Clone)]
struct Foo {
    inner: Rc<RefCell<std::io::Stdout>>,
}
```

asterite:
> What about making std::io::stdout clonable by wrapping in Rc/RefCell?

In the end a copyable std::io::Write wasn't needed, so `std::io::stdout()` is now passed directly.

michaeljklein:
@asterite then I guess my question becomes: can we avoid the use of `BytesWriter` (and cloning the buffer) by passing around `Stdout` or similar?

asterite:
I tried but I couldn't make it work. I tried passing `&mut Vec<u8>` to the callback but then I get an error that fuzz doesn't belong to that particular generic instance, and I don't understand why.

TomAFrench:
Perhaps the executor function should just return an owned Vec rather than needing to take in a buffer (with either interior mutability or needing to figure out the why the type system doesn't like mutable references)

asterite:
> Perhaps the executor function should just return an owned Vec rather than needing to take in a buffer

I'm trying it right now but I'm getting more errors.

Here's the diff:

```diff
diff --git a/tooling/fuzzer/src/lib.rs b/tooling/fuzzer/src/lib.rs
index 070c02d8e3..f84f72fc16 100644
--- a/tooling/fuzzer/src/lib.rs
+++ b/tooling/fuzzer/src/lib.rs
@@ -47,8 +47,7 @@ where
     E: Fn(
         &Program<FieldElement>,
         WitnessMap<FieldElement>,
-        BytesWriter,
-    ) -> Result<WitnessStack<FieldElement>, String>,
+    ) -> (Result<WitnessStack<FieldElement>, String>, Vec<u8>),
 {
     /// Instantiates a fuzzed executor given a [TestRunner].
     pub fn new(program: ProgramArtifact, executor: E, runner: TestRunner) -> Self {
@@ -124,9 +123,7 @@ where
     /// or a `CounterExampleOutcome`
     pub fn single_fuzz(&self, input_map: InputMap) -> Result<FuzzOutcome, TestCaseError> {
         let initial_witness = self.program.abi.encode(&input_map, None).unwrap();
-        let bytes_writer = BytesWriter::default();
-        let result = (self.executor)(&self.program.bytecode, initial_witness, bytes_writer.clone());
-        let output = bytes_writer.into_bytes();
+        let (result, output) = (self.executor)(&self.program.bytecode, initial_witness);
 
         // TODO: Add handling for `vm.assume` equivalent
 
diff --git a/tooling/nargo/src/ops/test.rs b/tooling/nargo/src/ops/test.rs
index 8486ed77b2..4f78421a5d 100644
--- a/tooling/nargo/src/ops/test.rs
+++ b/tooling/nargo/src/ops/test.rs
@@ -131,38 +131,42 @@ where
                 let abi = compiled_program.abi.clone();
                 let debug = compiled_program.debug.clone();
 
-                let executor = |program: &Program<FieldElement>,
-                                initial_witness: WitnessMap<FieldElement>,
-                                output|
-                 -> Result<WitnessStack<FieldElement>, String> {
-                    // Use a base layer that doesn't handle anything, which we handle in the `execute` below.
-                    let inner_executor =
-                        build_foreign_call_executor(Box::new(output), layers::Unhandled);
-
-                    let mut foreign_call_executor = TestForeignCallExecutor::new(inner_executor);
-
-                    let circuit_execution = execute_program(
-                        program,
-                        initial_witness,
-                        blackbox_solver,
-                        &mut foreign_call_executor,
-                    );
-
-                    // Check if a failure was actually expected.
-                    let status = test_status_program_compile_pass(
-                        test_function,
-                        &abi,
-                        &debug,
-                        &circuit_execution,
-                    );
-
-                    if let TestStatus::Fail { message, error_diagnostic: _ } = status {
-                        Err(message)
-                    } else {
-                        // The fuzzer doesn't care about the actual result.
-                        Ok(WitnessStack::default())
-                    }
-                };
+                let executor =
+                    |program: &Program<FieldElement>,
+                     initial_witness: WitnessMap<FieldElement>|
+                     -> (Result<WitnessStack<FieldElement>, String>, Vec<u8>) {
+                        // Use a base layer that doesn't handle anything, which we handle in the `execute` below.
+                        let mut output = Vec::<u8>::new();
+                        let inner_executor =
+                            build_foreign_call_executor(Box::new(&mut output), layers::Unhandled);
+
+                        let mut foreign_call_executor =
+                            TestForeignCallExecutor::new(inner_executor);
+
+                        let circuit_execution = execute_program(
+                            program,
+                            initial_witness,
+                            blackbox_solver,
+                            &mut foreign_call_executor,
+                        );
+
+                        // Check if a failure was actually expected.
+                        let status = test_status_program_compile_pass(
+                            test_function,
+                            &abi,
+                            &debug,
+                            &circuit_execution,
+                        );
+
+                        let result =
+                            if let TestStatus::Fail { message, error_diagnostic: _ } = status {
+                                Err(message)
+                            } else {
+                                // The fuzzer doesn't care about the actual result.
+                                Ok(WitnessStack::default())
+                            };
+                        (result, output)
+                    };
 
                 let mut fuzzer = FuzzedExecutor::new(compiled_program.into(), executor, runner);
 
```

Here are the errors I'm getting:

```
error[E0597]: `output` does not live long enough
   --> tooling/nargo/src/ops/test.rs:141:66
    |
40  | pub fn run_test<'a, W, B, F, E>(
    |                 -- lifetime `'a` defined here
...
139 |                         let mut output = Vec::<u8>::new();
    |                             ---------- binding `output` declared here
140 |                         let inner_executor =
141 |                             build_foreign_call_executor(Box::new(&mut output), layers::Unhandled);
    |                             -------------------------------------^^^^^^^^^^^---------------------
    |                             |                                    |
    |                             |                                    borrowed value does not live long enough
    |                             argument requires that `output` is borrowed for `'a`
...
169 |                     };
    |                     - `output` dropped here while still borrowed

error[E0505]: cannot move out of `output` because it is borrowed
   --> tooling/nargo/src/ops/test.rs:168:34
    |
40  | pub fn run_test<'a, W, B, F, E>(
    |                 -- lifetime `'a` defined here
...
139 |                         let mut output = Vec::<u8>::new();
    |                             ---------- binding `output` declared here
140 |                         let inner_executor =
141 |                             build_foreign_call_executor(Box::new(&mut output), layers::Unhandled);
    |                             ---------------------------------------------------------------------
    |                             |                                    |
    |                             |                                    borrow of `output` occurs here
    |                             argument requires that `output` is borrowed for `'a`
...
168 |                         (result, output)
    |                                  ^^^^^^ move out of `output` occurs here
    |
help: consider cloning the value if the performance cost is acceptable
    |
141 |                             build_foreign_call_executor(Box::new(&mut output.clone()), layers::Unhandled);
    |                                                                             ++++++++
```

but of course we can't clone output there.

I _think_ I understand what the error is saying: the lifetime of `output` should be that of the function, but in this case we are passing an `output` that has a shorter lifetime. So I don't know what lifetime we can specify here.

asterite:
With the PR I'm working on, this is not going to be needed. Maybe some parts will be useful, but most of this will be gone.

jfecher:
@asterite what is the status of this PR and https://github.com/noir-lang/noir/issues/8194?

asterite:
This PR has code changes over code that no longer exists, since we migrated to the greybox fuzzer. I'll reply on the issue a bit more about this.

asterite:
Oh, I thought the comment was on the issue so I'll answer here: I think it #8194 could still be done... though I don't know if showing output when fuzzing makes sense, so maybe it's fine to close that issue.

PR #8207, opened 2025-04-24
feat(fuzz): Add experimental comptime code generator

rkarabut:
# Description

Printing `noir_ast_fuzzer` as Noir, we should make it possible to fuzz comptime output against that of the equivalent compiled code. This adds an experimental version of valid comptime code generator, building a comptime function and wrapping its call with appropriate literal arguments in a `comptime` block within `main` .

Example of running `cargo run -p noir_ast_fuzzer --example sample_comptime`:

```
fn main() -> pub Field {
    comptime {
        func_1(18278151528840398125126511725758739340)
    }
}
comptime fn func_1(a: Field) -> Field {
    let b = if true {
        for idx_c in 1637873142 .. 1637873144 {
            let mut d = false;
            d = d;
            d = {
                d = d;
                d = (4163569491 < 2148799602);
                if (!false) {
                    for idx_e in -6940455744395345574 .. -6940455744395345566 {
                        if d {
                            ()
                        };
                        let f = (2544915962152465123 < -1838314094939617270);
                    };
                    ((idx_c as u128) <= 20225268882629477752056013539620857730)
                } else {
                    d
                }
            };
            d = d;
        };
        (a, a)
    } else {
        (-80171648342934597222692573068578438886, (-a))
    };
    a
}
```

## Documentation\*

Check one:

# PR Checklist\*


tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -924,6 +933,43 @@ impl<'a> FunctionContext<'a> {
...
+                typ: Type::Function(
+                    param_types,
+                    Box::new(callee.return_type.clone()),
+                    Box::new(Type::Unit),
+                    callee.unconstrained,
+                ),
+                id: self.next_ident_id(),
+            })),
+            arguments: args,
+            return_type: callee.return_type.clone(),
```
aakoshh, line 966: These `.clone()` calls should not be needed as `callee` has already been cloned and won't be used any more.

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -196,6 +196,15 @@ impl<'a> FunctionContext<'a> {
         Ok(body)
     }
 
+    /// Generate the function body, wrapping a function call with literal arguments.
+    pub fn gen_lit_call_wrapper_body(
```
aakoshh, line 200: Could you add a bit more explanation to the docstring about the rationale of this method? 

I'd also call it `gen_body_with_lit_call`, so we find all the different `gen_body*` variants that consume `self` with the same prefix.

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -196,6 +196,15 @@ impl<'a> FunctionContext<'a> {
...
+        mut self,
+        u: &mut Unstructured,
+        callee_id: FuncId,
+    ) -> arbitrary::Result<Expression> {
+        self.gen_lit_call(u, callee_id)
```
aakoshh, line 205: So here we have a `FunctionContext` that I could have been initialised with a number of parameters and potentially global values. Yet its body will exclusively use literals to call another function, not the globals and not the parameters either. Is that correct?

If so, I wonder if there should be another `FunctionContext` level constructor to completely bypass `new`, to generate a parameterless function. 

What about the return value of this callee, is it guaranteed to match up with the body of where this is supposed to be spliced in? The dedicated constructor could also take care of that.

tooling/ast_fuzzer/src/program/mod.rs
```diff
@@ -109,9 +146,17 @@ impl Context {
...
+    /// Generate random function names and signatures, optionally specifying their exact number.
+    fn gen_function_decls(
+        &mut self,
+        u: &mut Unstructured,
+        num: Option<usize>,
+    ) -> arbitrary::Result<()> {
+        let num_non_main_fns = match num {
+            Some(num) => num,
+            None => u.int_in_range(0..=self.config.max_functions)?,
+        };
```
aakoshh, line 158: I think it would be nicer to add a `self.config.min_functions`, and you can set both min and max to `arb_program_comptime`, so there is a single way to set this via the configuration.

tooling/ast_fuzzer/src/program/mod.rs
```diff
@@ -31,13 +31,50 @@ mod tests;
...
+        id: FuncId(0),
+        name: decl_main.name.clone(),
+        parameters: decl_main.params.clone(),
+        body,
+        return_type: decl_main.return_type.clone(),
+        unconstrained: decl_main.unconstrained,
+        inline_type: decl_main.inline_type,
+        func_sig: decl_main.signature(),
+    };
+    ctx.functions.insert(FuncId(0), func_main);
```
aakoshh, line 72: Maybe with a bit of refactoring we could change this so we can avoid generating 2 random functions and then replacing `main` completely, so it looks something like this:
```    
    ctx.gen_function(u, 1)?;
    let decl_main = FunctionDelcartion { name: "main", ... };
    ctx.set_function_decl(FuncId(0), decl_main);
    ctx.gen_function(u, 0)?;
    ctx.rewrite_functions(u)?;
```

When `gen_function(u, 0)` is called, because of your suggestion to make sure `main` calls the other function, it will have no chance but to generate a `Call`, and because it has no parameters, and there are no globals, it will have to use literals.

aakoshh: I'm also not sure that because you use `ctx.rewrite_functions(u)?;` and then replace the main body with the literal call, it won't be passing the `ctx_limit` to the callee any more, it won't even initialise it. Instead it will pass some random `u32`, which could be a big value. 

With my suggestion it will preserve that and pass it by value (to do that you might have to make sure that `func_1` is `unconstrained`, though).

tooling/ast_fuzzer/examples/sample_comptime.rs
```diff
@@ -0,0 +1,27 @@
...
+    let data = {
+        let mut rng = rand::thread_rng();
+        let mut data = [0u8; 1024 * 1024];
+        rng.fill_bytes(&mut data);
+        data
+    };
+    let mut u = Unstructured::new(&data);
+
+    let mut config = Config::default();
+    config.include_brillig = false;
```
aakoshh, line 22: Does this have to be disabled? If so, should it be set by `arb_program_comptime` instead?

tooling/ast_fuzzer/src/program/func.rs
rkarabut: Done

compiler/noirc_frontend/src/monomorphization/printer.rs
```diff
@@ -87,15 +88,47 @@ impl AstPrinter {
...
+        self.print_function(function, return_visibility, f)
+    }
+    
+    /// Debug method used in fuzzing. Prints a function wrapping a comptime function call
+    pub fn print_function_as_comptime_wrapper(
+        &mut self,
+        function: &Function,
+        return_visibility: Option<Visibility>,
+        f: &mut Formatter,
+    ) -> std::fmt::Result {
```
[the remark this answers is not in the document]
rkarabut: Done

tooling/ast_fuzzer/src/program/mod.rs
rkarabut: Added

tooling/ast_fuzzer/src/program/func.rs
rkarabut: Fixed (well the first one's needed not the second)

tooling/ast_fuzzer/src/program/mod.rs
rkarabut: I'd like to compose the wrapper call explicitly to prevent any possible confusion later on (and make it more readable). Though maybe it would be better if _just_ main is generated, and then repurposed as comptime and wrapped with the "real" main. Also yes, I don't think `rewrite_functions` is needed. On it

tooling/ast_fuzzer/src/program/func.rs
rkarabut: That's right. I guess yeah it could be made a bit cleaner with a dedicated `FunctionContext` constructor getting rid of locals/globals, but say return value isn't covered by that, it gets replaced in the declaration

tooling/ast_fuzzer/src/program/mod.rs
aakoshh: I do think `rewrite_functions` is needed, because even a single function can be infinitely recursive. 

With my suggestion it would be something like this:

```
fn main() -> u32 {
    comptime {
        let mut ctx_limit = 25;
        let a = 123;
        if func_1(a, &mut ctx_limit) + func_1(456, &mut ctx_limit) > 789 {
          func_1(1123, &mut ctx_limit)
        } else {
          1456
        }
    }
}

comptime fn func_1(a: u32, ctx_limit: &mut u32) -> u32 {
  if *ctx_limit == 0 {
    246
  } else {
    *ctx_limit = *ctx_limit - 1;
    func_1(a - 1, ctx_limit)
  }
}
```

Which I think is nice as you have more variety than calling a single function with random args, although main will only ever call a single function, but it can do so multiple times.

aakoshh: > Though maybe it would be better if just main is generated

I think with the way the generator works, the way you did it make sense: if you only tried to generate `main`, it would never recurse because nothing is allowed to call `main`, so you would never generate an entire class of programs.

With `func_1` and `main` you get a lot more variety.

aakoshh: > to do that you might have to make sure that func_1 is unconstrained, though

I just realised writing the above snippet that this isn't the case: as long as both `main` and `func_1` have the same runtime, and the entire body of `main` is wrapped in `comptime`, then passing by reference should work, if the interpreter supports it. 

I'd actually use brillig for both functions, so instead of `include_brillig` it would be `force_brillig`, so you get `loop` and `while`, which the interpreter handles AFAIK.

aakoshh: > I'd like to compose the wrapper call explicitly to prevent any possible confusion later on (and make it more readable).

Are you talking about the readability of the generator, or the generated AST? What is causing confusion?

rkarabut: Actually both, but mostly the generator. I would prefer main to stay a simple clean wrapper, as pretty much the entry point to comptime. We can expand the generator to bring in more functions and allow for their interaction afterwards, but for starters I think just calling a non-recursive function juggling a few statements would suffice

rkarabut: > Which I think is nice as you have more variety than calling a single function with random args. Although main will only ever call a single function, but it can do so multiple times.

Although I do like the idea hm

rkarabut: > Actually both, but mostly the generator. I would prefer main to stay a simple clean wrapper, as pretty much the entry point to comptime. We can expand the generator to bring in more functions and allow for their interaction afterwards, but for starters I think just calling a non-recursive function juggling a few statements would suffice

For readability in the generated code I mean that it's nice to have a clear top-level call to what's in fact the comptime's main, with pregenerated literal arguments. It seems it would make comparing it to the output of equivalent compiled code easier, with the compiler entry point being the actual wrapped main.

rkarabut: Supplying the right arguments would be more difficult if there's some code besides the main function wrapper

aakoshh: In that case I think what could possibly work is to do things in this order:
```rust
ctx.config.force_brillig = true; 
for i in 1..=2 { // If we want more
  ctx.set_function_decl(FuncId(i), ctx.gen_function_decl(u)?);
  ctx.gen_function(i);
}
ctx.set_function_decl(FuncId(0), FunctionDeclaration { ..., return_type = ... }); // parameterless main, return from func_q
ctx.set_function(FuncId(0), ctx.gen_function_just_calling(0, FuncId(1))?);
ctx.rewrite_functions(u)?; // Now do the recursion stuf
```

* By calling `rewrite_functions` last, it will initialise the `ctx_limit` in `main`, and add it to the call it makes
* By forcing all functions to be `unconstrained`, all functions to support loop, while, anything the interpreter can do, and because they are all the same runtime, there is no `unsafe` keyword, and `ctx_limit` is passed by ref

rkarabut: Done

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -184,7 +184,7 @@ impl<'a> FunctionContext<'a> {
     }
 
     /// Generate the function body.
-    pub fn gen_body(mut self, u: &mut Unstructured) -> arbitrary::Result<Expression> {
+    pub fn gen_body(&mut self, u: &mut Unstructured) -> arbitrary::Result<Expression> {
```
aakoshh, line 187: Does it need to be `&mut` instead of `mut`? The budget won't reset after this.

tooling/ast_fuzzer/src/program/mod.rs
```diff
@@ -189,23 +233,40 @@ impl Context {
...
+
+    /// Generate function with a specified body generator.
+    fn gen_function_with_body(
+        &mut self,
+        u: &mut Unstructured,
+        id: FuncId,
+        f: impl Fn(&mut Unstructured, &mut FunctionContext) -> arbitrary::Result<Expression>,
+    ) -> arbitrary::Result<()> {
+        let mut fctx = FunctionContext::new(self, id);
+        let body = f(u, &mut fctx)?;
```
aakoshh, line 254: Right, because of this. Maybe it could be `f: impl Fn(&mut Unstructured, FunctionContext)` instead?

tooling/ast_fuzzer/src/program/mod.rs
```diff
@@ -109,9 +151,11 @@ impl Context {
         Ok((name, typ, val))
     }
 
-    /// Generate random function names and signatures.
+    /// Generate random function names and signatures, optionally specifying their exact number.
```
aakoshh, line 154:
```suggestion
    /// Generate random function names and signatures.
```

aakoshh (approved):
Cool, happy fuzzing!

PR #8213, opened 2025-04-24
feat: add `--no-fuzz` and `--only-fuzz` to `nargo test`

asterite:
# Description

## Problem

Part of #8214

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/nargo_cli/src/cli/test_cmd.rs
+++ b/tooling/nargo_cli/src/cli/test_cmd.rs
@@ -23,7 +23,7 @@ use nargo::{
 };
 use nargo_toml::PackageSelection;
 use noirc_driver::{CompileOptions, check_crate};
-use noirc_frontend::hir::{FunctionNameMatch, ParsedFiles};
+use noirc_frontend::hir::{FunctionNameMatch, ParsedFiles, def_map::TestFunction};
 
 use crate::errors::CliError;
 
@@ -478,7 +478,7 @@ impl<'a> TestRunner<'a> {
 
         let tests: Vec<Test> = test_functions
             .into_iter()
-            .map(|test_name| {
+            .map(|(test_name, test_function)| {
                 let test_name_copy = test_name.clone();
                 let root_path = root_path.clone();
                 let package_name_clone = package_name.clone();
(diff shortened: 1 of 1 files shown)


compiler/noirc_frontend/src/hir/def_map/mod.rs
```diff
@@ -391,6 +393,10 @@ impl TestFunction {
         self.location.file
     }
 
+    pub fn has_arguments(&self) -> bool {
+        self.has_arguments
```
michaeljklein, line 397: Could there be cases where this is false at one point and true at another point? E.g. if a function is modified by metaprogramming? If so, then I think we should calculate it each time instead of caching

tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -497,11 +505,15 @@ impl<'a> TestRunner<'a> {
...
-            .get_all_test_functions_in_crate_matching(&crate_id, &self.pattern)
-            .into_iter()
-            .map(|(test_name, _)| test_name)
-            .collect())
+        let mut tests = context.get_all_test_functions_in_crate_matching(&crate_id, &self.pattern);
+        if self.args.no_fuzz {
+            tests.retain(|(_, test)| !test.has_arguments());
+        }
+        if self.args.only_fuzz {
+            tests.retain(|(_, test)| test.has_arguments());
```
michaeljklein, line 513: What do you think of marking these tests as skipped, e.g. `test_name -- SKIP`?

michaeljklein (changes requested):

compiler/noirc_frontend/src/hir/def_map/mod.rs
asterite: I don't think this is possible. This is computed after the program was compiled.

tooling/nargo_cli/src/cli/test_cmd.rs
asterite: Oh! I thought about doing it that way, then... I forgot 😅 I think that makes sense, I'll do the change.

asterite: Done!

michaeljklein (approved):

PR #8216, opened 2025-04-24
feat: AST fuzzer compare result between minimum and full SSA pipeline

### LINKED ISSUE #8034: Fuzz with random prefix of the SSA passes
After #7873 we can choose a random prefix of the SSA passes and check that their execution is equivalent to the initial SSA.

aakoshh:
# Description

## Problem\*

Resolves #8034

## Summary\*

Changes the `init_vs_final` fuzz target to do no, or minimum amount of SSA passes. (So far it compared min vs max inlining).

## Additional Context

```shell
cd tooling/ast_fuzzer
cargo +nightly fuzz run init_vs_final
```

For now I'm trying to figure out how minimal the SSA passes can be. A crash can be investigated with for example:

```shell
❯ NOIR_AST_FUZZER_SHOW_AST=1 cargo +nightly -q fuzz run init_vs_final fuzz/artifacts/init_vs_final/crash-adc83b19e793491b1c6ea0fd8b46cd9f32e592fc
...
---
global G_A: bool = false;
global G_B: bool = false;
    (!(!G_A))
}

---

thread '<unnamed>' panicked at compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs:1794:25:
ICE: Global value not found in cache v0
```

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/ast_fuzzer/fuzz/fuzz_targets/acir_vs_brillig.rs
+++ b/tooling/ast_fuzzer/fuzz/fuzz_targets/acir_vs_brillig.rs
@@ -1,9 +1,12 @@
 //! Compare the execution of random ASTs between the normal execution
 //! vs when everything is forced to be Brillig.
-
+//!
+//! ```text
+//! cargo +nightly fuzz run acir_vs_brillig
+//! ```
 #![no_main]
 
-use color_eyre::eyre::{self, Context};
+use color_eyre::eyre;
 use libfuzzer_sys::arbitrary::Unstructured;
 use libfuzzer_sys::fuzz_target;
 use noir_ast_fuzzer::Config;
@@ -19,9 +22,8 @@ fn fuzz(u: &mut Unstructured) -> eyre::Result<()> {
     let inputs = CompareMutants::arb(
         u,
         Config::default(),
-        |_u, program| {
+        |_u, mut program| {
             // Change every function to be unconstrained.
-            let mut program = program.clone();
             for f in program.functions.iter_mut() {
                 f.unconstrained = true;
             }
@@ -30,7 +32,7 @@ fn fuzz(u: &mut Unstructured) -> eyre::Result<()> {
         |program| create_ssa_or_die(program, &options, None),
     )?;
 
-    let result = inputs.exec().wrap_err("exec")?;
+    let result = inputs.exec()?;
 
     compare_results(&inputs, &result, |inputs| [&inputs.program.0, &inputs.program.1])
 }
--- a/tooling/ast_fuzzer/fuzz/fuzz_targets/init_vs_final.rs
+++ b/tooling/ast_fuzzer/fuzz/fuzz_targets/init_vs_final.rs
@@ -1,15 +1,20 @@
 //! Compare the execution of random ASTs between the initial SSA
 //! (or as close as we can stay to the initial state)
 //! and the fully optimized version.
+//!
+//! ```text
+//! cargo +nightly fuzz run init_vs_final
+//! ```
 #![no_main]
 
-use color_eyre::eyre::{self, Context};
+use color_eyre::eyre;
 use libfuzzer_sys::arbitrary::Unstructured;
 use libfuzzer_sys::fuzz_target;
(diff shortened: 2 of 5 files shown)


tooling/ast_fuzzer/src/program/mod.rs
```diff
@@ -332,10 +347,15 @@ impl Context {
...
                     // i1 is deprecated, and i128 does not exist yet
-                    let sign = *u.choose(&[Signedness::Signed, Signedness::Unsigned])?;
+                    let sign = if is_frontend_friendly {
+                        Signedness::Unsigned
+                    } else {
+                        *u.choose(&[Signedness::Signed, Signedness::Unsigned])?
+                    };
                     let sizes = IntegerBitSize::iter()
                         .filter(|bs| {
                             !(sign.is_signed() && (bs.bit_size() == 1 || bs.bit_size() == 128))
```
rkarabut, line 357: This condition gets somewhat confusing, could we rewrite it a bit clearer?

tooling/ast_fuzzer/src/program/mod.rs
```diff
@@ -301,11 +311,15 @@ impl Context {
     /// functions.
     ///
     /// With a `max_depth` of 0 only leaf types are created.
+    ///
+    /// With `is_frontend_friendly` we try to only consider types which are less likely to result
+    /// in literals that the frontend does not like when it has to infer their types.
```
rkarabut, line 316: Should clarify that this currently relates to non-u32 and empty arrays specifically I think

tooling/ast_fuzzer/src/program/rewrite.rs
```diff
@@ -286,3 +286,18 @@ fn next_local_and_ident_id(func: &Function) -> (u32, u32) {
     });
     (next_local_id, next_ident_id)
 }
+
+/// Turn all ACIR functions into Brillig functions.
+///
+/// This is more involved then flipping the `unconstrained` property because of the
```
rkarabut, line 292:
```suggestion
/// This is more involved than flipping the `unconstrained` property because of the
```

tooling/ast_fuzzer/src/program/types.rs
```diff
@@ -203,6 +203,13 @@ pub(crate) fn can_binary_op_return(op: &BinaryOp, typ: &Type) -> bool {
     }
 }
 
+/// Can the binary operation result in an over/underflow or division by zero.
+/// These are operations that commonly fail with random constants.
+pub(crate) fn can_binary_op_overflow(op: &BinaryOp) -> bool {
+    use BinaryOpKind::*;
+    matches!(op, Add | Subtract | Multiply | Divide | ShiftLeft | ShiftRight)
```
rkarabut, line 210:
```suggestion
    matches!(op, Add | Subtract | Multiply | Divide | Modulo | ShiftLeft | ShiftRight)
```

tooling/ast_fuzzer/src/program/mod.rs
rkarabut: Sorry, meaning lines 357-358 rather

aakoshh: Added comments to both lines and changed the first to look like the second: `(!matters || is_okay)`

tooling/ast_fuzzer/src/program/types.rs
aakoshh: I was thinking about adding modulo, but then I thought it should not cause an issue unless we actually choose the 0 literal. OTOH it did come up two bugs recently in a different format, e.g. `a % (a % a)`: https://github.com/noir-lang/noir/issues/8261 and https://github.com/noir-lang/noir/issues/8235

That's why I was reluctant to exclude it. I suppose the same argument could be said about `Divide` as well 🤔 

Maybe we should separate these two out into a "zero related" issues that are less likely to occur due to combining random literals, and more to do with expressions?

rkarabut: Yeah, seems like a good idea

PR #8224, opened 2025-04-25
fix(parser): do not use `Ident::default()`

asterite:
# Description

## Problem

Similar to #8178, `Ident::default()` ends up using a default/dummy location that would sometimes crash the LSP server.

## Summary

One case where it crashed was if you had code like this:

```noir
#[test]
fn () {
    let x = 1;
    let y = 2;
}
```

The name is missing. In my case that happens when I rename the function: I use VIM mode so the command first removes the entire name and lets me type a new one.

The error happened because in this case the parser ended up using `Ident::default()` for the name, and when LSP tried to use that location it crashed.

In addition to this I improved the error message shown in this case. Previously it would not continue parsing the function and error on each successive token. Now the parser will assume the function definition continues if a `(` or `<` comes next.

Finally, checking the error messages now use `assert_snapshot!` which will be handy if we later change/improve the produced error messages.

## Additional Context

I'm trying to fix these LSP bugs as soon as I find them because I usually left them for later, but then I couldn't remember how to trigger them.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/parser/parser.rs
+++ b/compiler/noirc_frontend/src/parser/parser.rs
@@ -530,11 +530,10 @@ impl<'a> Parser<'a> {
     }
 
     fn location_at_previous_token_end(&self) -> Location {
-        Location::new(self.span_at_previous_token_end(), self.previous_token_location.file)
-    }
-
-    fn span_at_previous_token_end(&self) -> Span {
-        Span::from(self.previous_token_location.span.end()..self.previous_token_location.span.end())
+        let span_at_previous_token_end = Span::from(
+            self.previous_token_location.span.end()..self.previous_token_location.span.end(),
+        );
+        Location::new(span_at_previous_token_end, self.previous_token_location.file)
     }
 
     fn expected_identifier(&mut self) {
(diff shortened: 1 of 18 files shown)


compiler/noirc_frontend/src/parser/parser.rs
```diff
@@ -530,11 +530,14 @@ impl<'a> Parser<'a> {
...
+        let span_at_previous_token_end = Span::from(
+            self.previous_token_location.span.end()..self.previous_token_location.span.end(),
+        );
+        Location::new(span_at_previous_token_end, self.previous_token_location.file)
     }
 
-    fn span_at_previous_token_end(&self) -> Span {
-        Span::from(self.previous_token_location.span.end()..self.previous_token_location.span.end())
+    fn empty_ident_at_previous_token_end(&self) -> Ident {
+        Ident::new(String::new(), self.location_at_previous_token_end())
```
michaeljklein, line 540: I wonder if having the `Ident`'s string be empty could cause similar issues. Perhaps the default string could be something like `"{unknown}"` for clearer error messages?

michaeljklein (approved):

compiler/noirc_frontend/src/parser/parser.rs
asterite: Done! I also pushed another commit to not produce a module AST if an identifier doesn't come after "mod" or "contract" because otherwise the error was a bit strange (it was already strange).

Issue #8155, opened 2025-04-21
Document SSA Pass - Inlining - `inline_simple_functions`


Savio-Sou:
Document the `inline_simple_functions` function in https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/inlining.rs.

PR #8227, opened 2025-04-25
chore: rename `inline_simple_functions` to `inline_functions_with_at_most_on…

asterite:
…e_instruction`

# Description

## Problem

Related to #8155

## Summary

In https://github.com/noir-lang/noir/pull/8169 I suggested maybe moving the `inline_simple_functions` to its own module. This is what this PR does, but it also renames the function so it's clearer.

The reasons:
- The function didn't use code from the rest of the file
- It being in a separate module means we can document that module (like we currently document all SSA optimization modules)
- It's clearer that the tests in the module belong to this particular optimization and not other inlining types
- The name `inline_functions_with_at_most_one_instruction` is a big longer, but with "simple" you had no idea what that means unless you read the docs. Now there less need to read the docs (there are caveats, like the "no predicates" case).

## Documentation

Check one:

# PR Checklist


michaeljklein (approved):
LGTM

PR #8239, opened 2025-04-25
feat(ssa): Prune dead block parameters

### LINKED ISSUE #8233: Only Brillig: `attempted to subtract with overflow`
### Aim

The following program behaves different between Brillig and ACIR:

```rust
global G_A: u128 = 34475917616803563277280217276819298828;
fn main(c: bool) -> pub u32 {
    let mut d: [Field; 1] = func_3(779302847);
    d = if c {
        d[0] = (G_A as Field);
        d
    } else { d };
    1948156509
}

fn func_3(b: u32) -> [Field; 1] {
    [
        if true {
            if (if (!false) { 1061432097 } else { (b >> 41) } < (506440519 - 3531841762)) {
                if true {
                    ((G_A as Field) - (297742546099112882246414176413370796930 + (b as Field)))
                } else {
                    (G_A as Field)
                }
            } else {
                (G_A as Field)
            }
        } else {
            (b as Field)
        },
    ]
}
```

### Expected Behavior

They should agree. 

This is similar to the last 3 bugs I created. It's defining a variable that won't ultimately be used in the return value, but we set one of its elements based on the input. 

### Bug

```console
   ┌─ src/main.nr:14:66
   │
14 │             if (if (!false) { 1061432097 } else { (b >> 41) } < (506440519 - 3531841762)) {
   │                                                                  ----------------------
   │
   = Call stack:
     1. src/main.nr:3:29
     2. src/main.nr:14:66

Failed assertion
[regression_fuzz_testing] Circuit output: Field(1948156509)
  ┌─ src/main.nr:8:5
  │
8 │     1948156509
  │     ---------- This variable contains a value which is constrained to be a constant. Consider removing this value as additional return values increase proving/verification time
  │
  = Call stack:
    1. src/main.nr:8:5

[regression_fuzz_testing] Circuit output: Field(1948156509)

```

### To Reproduce

See above.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+ae008d31e289ddb086b4c69e3fd287c5167fc4dc (git version hash: 9f595ea5ff5da20490c263de3283ee23d9f8bde8, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #8237: SSA: Remove dead block parameters and adjust terminators
### Problem

From https://github.com/noir-lang/noir/issues/8229#issuecomment-2831112993:

For this program:
```noir
global G_B: u32 = 2825334515;
fn main(b: [[bool; 4]; 4]) -> pub bool {
    let _a = if b[0][(G_B % 4)] {
        ((601072115 * G_B) as u64)
    } else {
        3513574538769362461
    };
    false
}
```
When compiling to ACIR we have the following SSA before DIE:
```
g0 = u32 2825334515

  b0(v1: [[u1; 4]; 4]):
    v3 = array_get v1, index u32 0 -> [u1; 4]
    v5 = array_get v3, index u32 3 -> u1
    enable_side_effects v5
    v7 = mul u32 601072115, u32 2825334515
    v8 = cast v7 as u64
    v9 = not v5
    enable_side_effects u1 1
    v11 = cast v5 as u64
    v12 = cast v9 as u64
    v13 = unchecked_mul v11, v8
    v15 = unchecked_mul v12, u64 3513574538769362461
    v16 = unchecked_add v13, v15
    return u1 0
}
```
After DIE we have the following:
```
g0 = u32 2825334515

  b0(v1: [[u1; 4]; 4]):
    v3 = array_get v1, index u32 0 -> [u1; 4]
    v5 = array_get v3, index u32 3 -> u1
    enable_side_effects v5
    enable_side_effects u1 1
    return u1 0
}
```
We can see that the overflowing instruction has been removed as it is never used.

While when force compiling to Brillig we have the following SSA before and after DIE:
<p>

```
After Check u128 mul overflow (1):
g0 = u32 2825334515

  b0(v1: [[u1; 4]; 4]):
    v4 = array_get v1, index u32 0 -> [u1; 4]
    inc_rc v4
    v6 = array_get v4, index u32 3 -> u1
    jmpif v6 then: b1, else: b2
  b1():
    v9 = mul u32 601072115, u32 2825334515
    v10 = cast v9 as u64
    jmp b3(v10)
  b2():
    jmp b3(u64 3513574538769362461)
  b3(v2: u64):
    return u1 0
}

After Dead Instruction Elimination (1):
g0 = u32 2825334515

  b0(v1: [[u1; 4]; 4]):
    v4 = array_get v1, index u32 0 -> [u1; 4]
    v6 = array_get v4, index u32 3 -> u1
    jmpif v6 then: b1, else: b2
  b1():
    v9 = mul u32 601072115, u32 2825334515
    v10 = cast v9 as u64
    jmp b3(v10)
  b2():
    jmp b3(u64 3513574538769362461)
  b3(v2: u64):
    return u1 0
}
```

</p>
</details> 

As the instruction is used in a block terminator it is never removed and we still trigger an overflow in our final program. So this issue looks to actually be a discprenacy in how ACIR and Brillig handle DIE. 

Also, in Rust this would fail at compile-time as a possible overflow. We should consider adding compile-time overflow checks in a separate issue. However, this issue would still arise if we made either the lhs or rhs of the overflowing operation a dynamic input. 

https://github.com/noir-lang/noir/issues/8236 and https://github.com/noir-lang/noir/issues/8233 look to be related.

### Happy Case

We should track parameter usage as part of DIE. If a parameter is not used in a block we should remove it and update all terminators that jump to the block with the dead parameters. Upon updating all terminators to account for dead block parameters, running DIE again will propagate its usage backwards in the blocks which called the block with the dead parameter.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

I have labeled this an enhancement, but it also will potentially fix multiple bugs (https://github.com/noir-lang/noir/issues/8229, https://github.com/noir-lang/noir/issues/8230, https://github.com/noir-lang/noir/issues/8231, https://github.com/noir-lang/noir/issues/8233, https://github.com/noir-lang/noir/issues/8236)

### Project Impact

Blocker

### Blocker Context

Not handling this logic leaves discrepancies between how Brillig and ACIR handle DIE

Yes

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #8237, #8229

## Summary\*

We now remove dead block parameters and update the respective predecessor terminators. This was done as follows:
1. Tracking unused parameters during DIE
2. Returning a map of (BasicBlockId -> unused parameters) for a given function
3. Return a map of (FunctionId -> map(BasicBlockId -> unused parameters)) for the given Ssa
4. Now go through the map of unused parameters returned from DIE
5. If a block has unused parameters remove those parameters (except for the entry block)
6. For each predecessor of the block for which we removed parameters, update the terminator instruction of the predecessor to remove the corresponding arguments

## Additional Context

Partially resolves #8233 except for the minimum inliner setting https://github.com/noir-lang/noir/issues/8233#issuecomment-2831420153. Perhaps we are ok with that though as ACIR cannot be run with a different inliner setting either way. So for the same inliner setting they do in fact still match. 

I could add #8233 as a regression test and just override the minimum inliner setting if we do view this as matching execution. 

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
Looks like the last failing test is in the debugger. The debugger is most likely keeping the bad overflowing instruction in use, thus we error out vs. normal execution. This PR should still be fine to review. I will add this as an exception to the debugger tests as we do in fact expect an error here with the debugger then.

TomAFrench:
Thoughts on just including this as part of the DIE pass?

TomAFrench:
This would also make it easier to trigger another round of DIE if necessary.

vezenovm:
> Thoughts on just including this as part of the DIE pass?

I wanted the pruning to be separate from the DIE logic, but you raise a good point that it is simpler to have them combined. I'll merge the passes.

TomAFrench:
I think we can still keep it relatively separate but if we're always calling this immediately after DIE then it probably makes sense to have a "DIE+block params" pass which calls the two sub-passes internally. This also makes it easier for us to pass the state without having to pass it up into the dfg and down again.

vezenovm:
> if we're always calling this immediately after DIE then it probably makes sense to have a "DIE+block params" pass which calls the two sub-passes internally. This also makes it easier for us to pass the state without having to pass it up into the dfg and down again.

Yeah I decided to go with turning DIE into a "DIE+block params" pass. However, for passing the state I still find it a bit cleaner to simply attach the unused parameters onto the `BasicBlock` itself. `prune_dead_parameters` doesn't need to accept any extra state as input then and `Function::dead_instruction_elimination` remains a bit simpler. DIE already returns some state for unused globals so perhaps we would prefer to have all state that DIE produces for follow-up passes to be returned and passed directly into the follow-up pass. Let me know what you prefer.

compiler/noirc_evaluator/src/ssa/opt/die/prune_dead_parameters.rs
```diff
@@ -0,0 +1,250 @@
...
+        let post_order = PostOrder::with_cfg(&cfg);
+
+        for block in post_order.as_slice() {
+            let block = *block;
+
+            let unused_params = self.dfg[block].unused_parameters().to_vec();
+            if unused_params.is_empty() {
+                // Nothing to do if the block has no unused params
+                continue;
+            }
```
jfecher, line 76: Hmm I was thinking this may cause problems if a block uses another block's parameter (in a way that that instruction would also be marked unused) but it doesn't seem to cause issues in a couple programs I've tested. It also looks like this pass will only remove 1 "layer" of unused block parameters at a time so if you have a program like:
```noir
unconstrained fn main(input: i16) {
    let result = if input > 3 {
        if input > 4 {
            1
        } else {
            2
        }
    } else {
        3
    };

    let result2 = if input > 5 {
        result
    } else {
        result + 1
    };
}
```
which has 2 layers of them it takes a couple die passes. Removing this continue doesn't make it take 1 pass so maybe DIE itself is saying the `result` block parameter is used the first time and only marking `result2` as unused. Then in a second DIE `result` is marked unused and removed.

compiler/noirc_evaluator/src/ssa/opt/die/prune_dead_parameters.rs
```diff
@@ -0,0 +1,250 @@
...
+        brillig(inline) fn test f0 {
+          b0():
+            jmp b1(Field 2)
+          b1(v0: Field):
+            return v0
+        }"#);
+    }
+
+    #[test]
+    fn prune_unused_block_params_multiple_predecessors() {
```
jfecher, line 165: Can you add a larger test which shows a few unused results from e.g. the merge of a few `if`s?

jfecher (changes requested):

TomAFrench:
> DIE already returns some state for unused globals so perhaps we would prefer to have all state that DIE produces for follow-up passes to be returned and passed directly into the follow-up pass.

I would prefer for us to pass this state around outside of the DFG as it prevents something which is now internal to this new SSA pass leaking into the global SSA state.

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -97,6 +108,14 @@ impl Function {
...
         for block in blocks.as_slice() {
             context.remove_unused_instructions_in_block(self, *block);
+
+            let unused_params = self.dfg[*block]
+                .parameters()
+                .iter()
+                .filter(|value| !context.used_values.contains(value))
+                .copied()
+                .collect::<Vec<_>>();
+            self.dfg[*block].set_unused_parameters(unused_params);
```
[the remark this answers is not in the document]
vezenovm: I'll make a separate mapping to store this state.

compiler/noirc_evaluator/src/ssa/opt/die/prune_dead_parameters.rs
```diff
@@ -0,0 +1,250 @@
...
+//! - Detects and removes unused block parameters, except for those on the entry block.
+//! - Clears the list of unused block parameters after removing them from the block.
+//! - **Entry block parameters** are never removed. These often correspond to function inputs and
+//!   must remain to preserve the function's external interface, even if they're unused internally.
+//! - Updates the corresponding argument lists in predecessor terminator instructions to keep
+//!   them aligned with the new parameter lists.
+//!
+//! ## Preconditions:
+//! - This pass should be run *after* [Dead Instruction Elimination (DIE)][super] so that parameter
+//!   liveness is up-to-date.
```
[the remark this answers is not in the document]
vezenovm: Switching to a map of the ValueIDs to remove as mentioned in the other comments. However, a precondition for block parameter liveness will still exist.

compiler/noirc_evaluator/src/ssa/opt/die/prune_dead_parameters.rs
```diff
@@ -39,43 +39,31 @@
...
+        for &block in post_order.as_slice() {
+            let empty_params = Vec::new();
+            let unused_params = unused_params.get(&block).unwrap_or(&empty_params);
             if unused_params.is_empty() {
                 // Nothing to do if the block has no unused params
                 continue;
             }
 
-            self.dfg[block].clear_unused_parameters();
+            // self.dfg[block].clear_unused_parameters();
```
jfecher, line 66:
```suggestion
```

compiler/noirc_evaluator/src/ssa/ir/basic_block.rs
```diff
@@ -25,6 +25,10 @@ pub struct BasicBlock {
     /// This will be a control flow instruction. This is only
     /// None if the block is still being constructed.
     terminator: Option<TerminatorInstruction>,
+
+    /// List of parameters that have been found to be unused after running
+    /// dead instruction elimination
+    unused_parameters: Vec<ValueId>,
```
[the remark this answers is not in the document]
jfecher: Unsure if [115d125](https://github.com/noir-lang/noir/pull/8239/commits/115d125235ed7aac4c6c38f2de1d40029724a9cf) was meant to be a full switch or not but this field  remains despite the methods being removed

compiler/noirc_evaluator/src/ssa/opt/die/prune_dead_parameters.rs
vezenovm: Still making some changes, so I removed this locally

compiler/noirc_evaluator/src/ssa/ir/basic_block.rs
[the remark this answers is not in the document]
vezenovm: still making updates

compiler/noirc_evaluator/src/ssa/opt/die/prune_dead_parameters.rs
```diff
@@ -0,0 +1,250 @@
...
+
+impl Ssa {
+    /// See [`prune_dead_parameters`][self] module for more information.
+    #[tracing::instrument(level = "trace", skip(self))]
+    pub(crate) fn prune_dead_parameters(mut self) -> Self {
+        for function in self.functions.values_mut() {
+            function.prune_dead_parameters();
+        }
+        self
+    }
```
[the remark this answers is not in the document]
vezenovm: I originally did `Function::prune_dead_parameters` as to simplify the inputs to the pruning pass, however, I ultimately decided to return the full unused parameter state from DIE. 

I felt having a pass of DIE that only removes dead instructions to be simpler and better for testing. Pruning parameters changes the signatures of blocks and rewrites terminators which changes the CFG. This feels like it would be better to have separated from instruction removal. Keeping pruning totally separate allows us to more easily unit test pruning without DIE. We will also have more flexibility in the future by keeping them totally separate.

compiler/noirc_evaluator/src/ssa/opt/die/prune_dead_parameters.rs
vezenovm: Yeah so it shouldn't cause problems, as we are checking unused params against the entire used values context in DIE. 

> maybe DIE itself is saying the `result` block parameter is used the first time and only marking `result2` as unused. Then in a second DIE `result` is marked unused and removed.

However, this is occurring. DIE is going to mark `result` as used the first time and only `result2` will be marked unused. 

We have this SSA before DIE:

<p>

```
        b0(v0: i16):
          v4 = lt i16 3, v0
          jmpif v4 then: b1, else: b2
        b1():
          v7 = lt i16 4, v0
          v8 = not v7
          v9 = cast v7 as Field
          v10 = cast v8 as Field
          v12 = mul v10, Field 2
          v13 = add v9, v12
          jmp b3(v13)
        b2():
          jmp b3(Field 3)
        b3(v1: Field):
          v15 = lt i16 5, v0
          jmpif v15 then: b4, else: b5
        b4():
          jmp b6(v1)
        b5():
          v17 = add v1, Field 1
          jmp b6(v17)
        b6(v2: Field):
          return
      }
```

</p>
</details> 

It will mark `v2` (e.g. `result2`) as unused and `v1` (e.g. `result`) as used. I am going to look at updating DIE to handle multiple layers of unused params. If it is too complex I'll add it in a follow-up PR, however, I don't think it should be too bad.

jfecher (approved):

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -19,20 +22,28 @@ use crate::ssa::{
...
     }
 
     /// Post the Brillig generation we do not need to run this pass on Brillig functions.
     #[tracing::instrument(level = "trace", skip(self))]
     pub(crate) fn dead_instruction_elimination_acir(self) -> Ssa {
-        self.dead_instruction_elimination_inner(true, true)
+        let ssa = self.dead_instruction_elimination_inner(true, true);
+        // The liveness analysis for block parameters is handled during DIE,
+        // thus pruning parameters should always follow DIE.
+        ssa.prune_dead_parameters()
```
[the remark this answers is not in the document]
vezenovm: I ended up making a new method `dead_instruction_elimination_with_pruning` that calls the pruning pass internally and both DIE entry points call the new method. This then lets `dead_instruction_elimination_inner` only remove instructions and gather liveness information.

compiler/noirc_evaluator/src/ssa/opt/die/prune_dead_parameters.rs
vezenovm: done

compiler/noirc_evaluator/src/ssa/opt/die/prune_dead_parameters.rs
vezenovm: Split out marking transitive unused params in DIE https://github.com/noir-lang/noir/pull/8254

PR #8243, opened 2025-04-25
fix(acir): Do not do the bound_constraint_with_offset optimization when the rhs is const and zero

### LINKED ISSUE #8235: Only ACIR: `Failed constraint` (unused variable, division by zero)
### Aim

The following program fails in ACIR but passes in Brillig:

```rust
fn main(a: bool) -> pub bool {
    if (a) {
        let b = func_4(1569582039);
        true
    } else {
        false
    }
}
fn func_4(a: u32) -> u32 {
    (a / (a % a))
}
```

Using the following input:
```toml
a = false
```

### Expected Behavior

They should agree. 

It fails because it's doing `a / 0`, but Brillig doesn't go into that branch because `a` is `false`. If `a` is true, it fails the same way. Is it okay for ACIR to fail in both cases? `b` isn't even used, shouldn't it be eliminated?

### Bug

```console
warning: unused variable b
  ┌─ src/main.nr:3:13
  │
3 │         let b = func_4(1569582039);
  │             - unused variable
  │

  ┌─ src/main.nr:2:5
  │  
2 │ ╭     if (a) {
3 │ │         let b = func_4(1569582039);
4 │ │         true
5 │ │     } else {
6 │ │         false
7 │ │     }
  │ ╰─────' This variable contains a value which is constrained to be a constant. Consider removing this value as additional return values increase proving/verification time
  │  
  = Call stack:
    1. src/main.nr:2:5

error: Failed constraint
   ┌─ src/main.nr:10:6
   │
10 │     (a / (a % a))
   │      -----------
   │
   = Call stack:
     1. src/main.nr:3:17
     2. src/main.nr:10:6

warning: unused variable b
  ┌─ src/main.nr:3:13
  │
3 │         let b = func_4(1569582039);
  │             - unused variable
  │

bash-5.2$ 


```

### To Reproduce

See above.

### Workaround

Yes

### Workaround Description

Don't divide by zero. Nevertheless it's curious because `b` isn't even used, shouldn't it be eliminated?

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+b0d9c680c3e9f6a818382b4f57ae2b5a93c657ab (git version hash: 1f595f71320be9464dc527946c88d9bdf264dd9f, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #8235 

## Summary\*

In `bound_constraint_with_offset` we have an optimization for when the rhs is constant and fits in a u128. However, if the rhs is zero this optimization will lead a range constraint of zero bits which will always fail. I simply check whether the constant rhs is zero, and if it is we do not run the optimization.

## Documentation\*

Check one:

# PR Checklist\*


test_programs/execution_success/regression_8235/src/main.nr
```diff
@@ -0,0 +1,11 @@
...
+    if (a) {
+        let b = func_4(1569582039);
+        true
+    } else {
+        false
+    }
+}
+fn func_4(a: u32) -> u32 {
+    (a / (a % a))
+}
```
vezenovm, line 11:
```suggestion
}

```

compiler/noirc_evaluator/src/acir/acir_context/mod.rs
```diff
@@ -981,7 +981,10 @@ impl<F: AcirField, B: BlackBoxFunctionSolver<F>> AcirContext<F, B> {
 
         // Optimization when rhs is const and fits within a u128
         let rhs_expr = self.var_to_expression(rhs)?;
-        if rhs_expr.is_const() && rhs_expr.q_c.num_bits() <= 128 {
+        // Do not attempt this optimization when the q_c is zero as otherwise
+        // we will compute an rhs offset of zero and ultimately lay down a range constrain of zero bits
+        // which will always fail.
```
guipublic, line 986: Instead of avoiding this when the range constraint has bit-size 0, I'd rather fix the range constraint so that it does not fail with 0: if the bit size is 0, we can simply AssertZero() the variable (to be 0) in the range constraint.

TomAFrench (approved):
Looks sensible to me

compiler/noirc_evaluator/src/acir/acir_context/mod.rs
TomAFrench: We've defined elsewhere in the codebase that a value of 0 requires 1 bit to store it so there's scope for introducing new issues due to this.

vezenovm: I'll capture this in an issue

vezenovm: https://github.com/noir-lang/noir/issues/8251

PR #8249, opened 2025-04-28
fix: Call the ownership analysis

### LINKED ISSUE #8240: AST fuzzer should insert clone when assigning arrays to mutable variables
### Aim

This was elusive, so I thought I'd capture it on its own as a bug ticket, even if the bug is in the AST generator.

The following command generates an AST and compares it between ACIR and Brillig:

```shell
NOIR_ARBTEST_SEED=0x67cd723f00100000 NOIR_AST_FUZZER_SHOW_SSA=1 cargo test -p noir_ast_fuzzer_fuzz acir_vs_brillig
```

They disagree on the return value. 

The program in question looks like this:
```rust
global G_A: ([Field; 2], Field) = ([128616173893067586065949749649520356159, -39992705687682713591668456281436754155], 89994797223517382485415940006161433966);
global G_B: Field = 57174925586013130257604257278735153383;
fn main(a: str<3>, b: ([Field; 2], Field), c: ([Field; 2], Field)) -> pub ([Field; 2], Field) {
    for idx_d in 183346773020840504720451580705483121440 .. 183346773020840504720451580705483121444 {
        let mut e = b.0;
        for idx_f in 54819 .. 54825 {
            let mut g = (-(idx_d as Field));
        };
        let i = {
            if true {
                e[(if true {
                    2976294725
                } else {
                    if true {
                        1414144453
                    } else {
                        2046950688
                    }
                } % 2)] = if false {
                    c.1
                } else {
                    G_A.1
                };
                let h = G_B;
            };
            (-(idx_d as Field))
        };
        let mut j = "";
    };
    b
}
```

And the inputs:
```
a = "^JW"
b = [["0x2c1372fda47f62e2bd91ac18c787cc41db48cb93d2e87b66add9fa1e6d5fe1c1", "0x297913438907994aea8174019625a39297153a6041e55277d60ffd8218275053"], "0x27517ab8abc9d6cfb994977b8d9bc8b8b7e18ac66094c7b69df7cb48f36a1bc2"]
c = [["0x1d9306f095717f46fe45378a95004b07530493fd43205c1187be6536c8175386", "0x0c5d0c416538604e2cc73908d4da6e735606b9e0c804c985a162cf0abfccf41c"], "0x18b9e8650245eb7201209bc636582a45941941198fe2fb0d5cfc7c04295fdeba"]
```

### Expected Behavior

I expect them to return the same value. 

### Bug

```console
$ NOIR_ARBTEST_SEED=0x67cd723f00100000 NOIR_AST_FUZZER_SHOW_SSA=1 cargo test -p noir_ast_fuzzer_fuzz acir_vs_brillig
...
called `Result::unwrap()` on an `Err` value: programs disagree on return value: Some(Vec([Vec([Field(-1952113802671104411634548797245179493731731032183837380843626758631556390464), Field(-3129494631159243151911699491131808756609541060152924275722698262321447874478)]), Field(-4104080173068908072510107902313196041529891370866016099070729286085726561343)])) != Some(Vec([Vec([Field(-1952113802671104411634548797245179493731731032183837380843626758631556390464), Field(89994797223517382485415940006161433966)]), Field(-4104080173068908072510107902313196041529891370866016099070729286085726561343)]))
```

If we look closely, what happened was that the input array `b` got one of its items overwritten with `89994797223517382485415940006161433966` coming from `G_A`. 

However, when I tried to replicate this test, it returned the same value as the ACIR in the fuzzer:
```console
$ cargo run -q -p nargo_cli -- execute --skip-underconstrained-check --skip-brillig-constraints-check --silence-warnings --show-monomorphized
...
[regression_fuzz_testing] Circuit output: Vec([Vec([Field(-1952113802671104411634548797245179493731731032183837380843626758631556390464), Field(-3129494631159243151911699491131808756609541060152924275722698262321447874478)]), Field(-4104080173068908072510107902313196041529891370866016099070729286085726561343)])
```

Upon closer inspection, the initial SSA between the fuzzer and when I tried to replicate this in isolation were different. 

The fuzzer:
```
...
brillig(inline) fn main f0 {
  b0(v5: [u8; 3], v6: [Field; 2], v7: Field, v8: [Field; 2], v9: Field):
    v13 = array_get v8, index u32 0 -> Field
    v15 = array_get v8, index u32 1 -> Field
    jmp b1(u128 183346773020840504720451580705483121440)
  b1(v10: u128):
    v18 = lt v10, u128 183346773020840504720451580705483121444
    jmpif v18 then: b2, else: b3
  b2():
    v19 = allocate -> &mut [Field; 2]
    store v6 at v19
    jmp b4(u16 54819)
...
```
the reproduction:
```
brillig(inline) fn main f0 {
  b0(v5: [u8; 3], v6: [Field; 2], v7: Field, v8: [Field; 2], v9: Field):
    jmp b1(u128 183346773020840504720451580705483121440)
  b1(v10: u128):
    v14 = lt v10, u128 183346773020840504720451580705483121444
    jmpif v14 then: b2, else: b3
  b2():
    inc_rc v6
    v15 = allocate -> &mut [Field; 2]
    store v6 at v15
    jmp b4(u16 54819)
```

The fuzzer misses the `inc_rc v6`, which is why it is allowed to overwrite the array. 

The monomorphized AST of the reproduction was different than what I generated in the fuzzer:
```
let mut e = b.0.clone();
```

So because the AST generator doesn't insert an explicit clone, the Brillig VM, which uses refcounts is allowed to modify the array, whereas ACIR, which doesn't use this mechanism, is correct. Compiling the AST printed and re-parsing it by the compiler implicitly fixes the issue. 

#### Action

I need to figure out where to insert clones:
* array assigned to mutable variable
* array assigned to non-mutable variable?
* array passed to a function as an argument? what if the parameter is mutable?

Another issue might be that to turn an originally ACIR function into Brillig, all I do is flip the `unconstrained` flag on the `Function` in the monomorphized `Program`. Could it be that the monomorphizer inserts `.clone()` only if it's unconstrained? Indeed, running `--show-monomorphized` vs `--show-monomorphized --force-brillig` above shows that the ACIR version doesn't have `.clone()`, only the latter has.

### To Reproduce

See above.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+b0d9c680c3e9f6a818382b4f57ae2b5a93c657ab (git version hash: 7fcfd508a094d4baabaccc86882981f79193ce06, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #8240

## Summary\*

Makes `Program::handle_ownership` public and calls it in `Context::finalize` to insert `Clone` into `unconstrained` functions where necessary. Added a `change_functions_into_unconstrained` to flip the `unconstrained` flag in every function of a `Program` and run the `handle_ownership` on them, which can now also be called on `Function`. 

### Where to do it

I went back and forth between calling `handle_ownership` in `Context::finalize`, versus calling it in `create_ssa_or_die`. I thought we might have _rewrites_ in the future which could result in this method having to run again on an unconstrained function (where it would panic). In the end I put it in `finalize`, because that way the `CompareSsa::program` field contains an AST which is as close as possible to what we will execute, which could avoid some confusion in debugging. 

We might not want to show `.clone()` and `.drop()` in prints, because they are are not valid Noir code, but at least we would see them if we wanted to compare the AST to what `nargo` is doing, without further hidden steps happening during SSA conversion.

### Drop
The `ownership` module contains this comment:

> Clones & Drops are only inserted by this pass so we can assume any code they

However, I don't see `Drop` being inserted. Maybe it's reserved for the future. For now I adjusted the frequency of `"drop"` to 0 but left `gen_drop` in, although I expect we can remove it in the future, unless an explicit `drop` function will be available.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/lib.rs
+++ b/compiler/noirc_frontend/src/lib.rs
@@ -26,7 +26,7 @@ pub mod lexer;
 pub mod locations;
 pub mod monomorphization;
 pub mod node_interner;
-pub(crate) mod ownership;
+pub mod ownership;
 pub mod parser;
 pub mod resolve_locations;
 pub mod shared;
--- a/compiler/noirc_frontend/src/monomorphization/printer.rs
+++ b/compiler/noirc_frontend/src/monomorphization/printer.rs
@@ -23,11 +23,12 @@ pub struct AstPrinter {
     indent_level: u32,
     in_unconstrained: bool,
     pub show_id: bool,
+    pub show_clone_and_drop: bool,
 }
 
 impl Default for AstPrinter {
     fn default() -> Self {
-        Self { indent_level: 0, in_unconstrained: false, show_id: true }
+        Self { indent_level: 0, in_unconstrained: false, show_id: true, show_clone_and_drop: true }
     }
 }
 
@@ -177,11 +178,17 @@ impl AstPrinter {
             Expression::Continue => write!(f, "continue"),
             Expression::Clone(expr) => {
                 self.print_expr(expr, f)?;
-                write!(f, ".clone()")
+                if self.show_clone_and_drop {
+                    write!(f, ".clone()")?;
+                }
+                Ok(())
             }
             Expression::Drop(expr) => {
                 self.print_expr(expr, f)?;
-                write!(f, ".drop()")
+                if self.show_clone_and_drop {
+                    write!(f, ".drop()")?;
+                }
+                Ok(())
             }
         }
     }
--- a/compiler/noirc_frontend/src/ownership/mod.rs
+++ b/compiler/noirc_frontend/src/ownership/mod.rs
@@ -50,7 +50,7 @@ mod tests;
 impl Program {
     /// Perform "ownership analysis".
     ///
-    /// See [noirc_frontend::ownership] for details.
+    /// See [ownership](crate::ownership) for details.
     ///
     /// This should only be called once, before converting to SSA.
     pub fn handle_ownership(mut self) -> Self {
@@ -64,7 +64,7 @@ impl Program {
 impl Function {
     /// Perform "ownership analysis".
     ///
-    /// See [noirc_frontend::ownership] for details.
+    /// See [ownership](crate::ownership) for details.
     ///
     /// This should only be called on a function once.
     pub fn handle_ownership(&mut self) {
--- a/compiler/noirc_frontend/src/ownership/mod.rs
+++ b/compiler/noirc_frontend/src/ownership/mod.rs
@@ -48,13 +48,28 @@ mod last_uses;
 mod tests;
 
 impl Program {
-    pub(crate) fn handle_ownership(mut self) -> Self {
-        let mut context = Context { variables_to_move: Default::default() };
-
+    /// Perform "ownership analysis".
+    ///
+    /// See [noirc_frontend::ownership] for details.
+    ///
+    /// This should only be called once, before converting to SSA.
+    pub fn handle_ownership(mut self) -> Self {
         for function in self.functions.iter_mut() {
-            context.handle_ownership_in_function(function);
+            function.handle_ownership();
         }
         self
+    }
+}
+
+impl Function {
+    /// Perform "ownership analysis".
+    ///
+    /// See [noirc_frontend::ownership] for details.
+    ///
+    /// This should only be called on a function once.
+    pub fn handle_ownership(&mut self) {
+        let mut context = Context { variables_to_move: Default::default() };
+        context.handle_ownership_in_function(self);
     }
 }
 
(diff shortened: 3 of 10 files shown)


rkarabut (approved):

jfecher:
> However, I don't see Drop being inserted. Maybe it's reserved for the future. For now I adjusted the frequency of "drop" to 0 but left gen_drop in, although I expect we can remove it in the future, unless an explicit drop function will be available.

They used to be inserted in the old ownership analysis. But it turned out to be faster in brillig to just exclude them. In the new ownership analysis it wasn't as straightforward to add them so they weren't included for the MVP of that different scheme. Yet it turned out that scheme already showed measurable reductions in array copies in larger programs so we went ahead to use that as the default. We may still add Drops later but it isn't a priority for 1.0.

Issue #7991, opened 2025-04-09
Revisit `STDIN`-to-`main.nr` fuzzing


michaeljklein:
### Problem

https://github.com/noir-lang/noir/pull/5591 allowed testing `nargo` by giving the contents of `main.nr` through STDIN and auto-generating a simple `Nargo.toml`:
```bash
cat src/main.nr | nargo compile --debug-compile-stdin
```

This was useful for testing `nargo` against batches of Rust files, e.g. finding a few parser panics, but:
- Compilation was around 10x too slow
- Compilation, especially SSA passes, often took GB's RAM
- It was too slow for AFL fuzzing to efficiently search the state space

Now, we're expected to be 10x+ faster with significant reductions in RAM usage so this approach can be revisited.

### Happy Case

1. Make a new version of the `--debug-compile-stdin` flag
2. Test performance of AFL with `--debug-compile-stdin`
3. Test performance on batches of Rust files

### Workaround

None

### Workaround Description

I don't think the AFL part can really be worked around because:
- Any wrapper to call `nargo` will be fuzzed as part of the target (so e.g. a `bash` script wrapper would fuzz `bash` along with `nargo`)
- I've found such wrappers to often be the slowest part when testing, e.g. because they may require re-launching the target binary, writing files, etc.

### Project Impact

Nice-to-have

### Would you like to submit a PR for this Issue?

None

PR #8253, opened 2025-04-28
feat: add `--debug-compile-stdin` to read `main.nr` from `STDIN` for testing

michaeljklein:
# Description

## Problem\*

Prerequisite of https://github.com/noir-lang/noir/issues/7991

## Summary\*

- Accept `main.nr` from STDIN when `--debug-compile-stdin` is enabled
- Use an implicit `Nargo.toml`

## Additional Context

Not testing here because I plan to add tests in a follow up PR, but I could add a smoke test

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_driver/src/lib.rs
```diff
@@ -174,6 +174,11 @@ pub struct CompileOptions {
     #[arg(long, default_value = "false")]
     pub pedantic_solving: bool,
 
+    /// Accept a single '.nr' file through STDIN instead of reading from the
+    /// root directory
+    #[arg(long, hide = true)]
+    pub debug_compile_stdin: bool,
```
vezenovm, line 180: Could you expand the comment here `debug_compile_stdin`. This is where I would look if I was unsure of why we call `with_maybe_dummy_workspace` for the compile command.

tooling/nargo_cli/src/cli/mod.rs
```diff
@@ -178,6 +181,42 @@ fn read_workspace(
...
 }
 
+/// "with_workspace", but use a dummy workspace when 'debug_compile_stdin' is enabled
+fn with_maybe_dummy_workspace<R>(
+    cmd: compile_cmd::CompileCommand,
+    config: NargoConfig,
+    run: R,
+) -> Result<(), CliError>
+where
+    R: FnOnce(compile_cmd::CompileCommand, Workspace) -> Result<(), CliError>,
```
vezenovm, line 191: We only call the compile command so we can just do it directly.

asterite:
I'm trying it like this:

```
echo "fn main() {}" | nargo compile --debug-compile-stdin
```

and I get this error:

```
Cannot find file src/main.nr which is defaulted due to specifying `type = "bin"` in Nargo.toml
```

I think it's erroring before reaching the new conditional.

michaeljklein:
@asterite where does `nargo` point to on your path?

With `$NARGO="target/debug/nargo"`, I'm getting:

```bash
❯ cat src/main.nr | $NARGO compile --debug-compile-stdin
❯ echo "fn main() {}" | $NARGO compile --debug-compile-stdin
❯ echo "fn main() {}" | $NARGO compile --debug-compile-stdin --print-acir
func 0
current witness index : _0
private parameters indices : []
return value indices : []
```

asterite:
Huh, now it's working. I did `git pull`, I don't know if the last merge did it, or maybe I was at an old point (though it was the first time I jumped into this branch).

I guess nevermind then, sorry!

asterite (approved):
Looks great!

tooling/nargo_cli/src/cli/mod.rs
```diff
@@ -181,6 +184,40 @@ fn read_workspace(
     Ok(workspace)
 }
 
+/// "with_workspace", but use a dummy workspace when 'debug_compile_stdin' is enabled
+fn compile_with_maybe_dummy_workspace(
+    cmd: compile_cmd::CompileCommand,
+    config: NargoConfig,
+) -> Result<(), CliError>
+where
```
vezenovm, line 192:
```suggestion
```
empty where statement?

vezenovm (approved):

PR #8273, opened 2025-04-29
fix: use correct meta attribute names in contract custom attributes

### LINKED ISSUE #8255: "(quoted)" function custom attribute
### Aim

I was trying to get rid of the full path of the utility function in the function custom attribute by doing the following in the Aztec macros:

Image

Unfortunately I got this in the artifact instead of "utility":

Image

### Expected Behavior

Get "utility" in the custom attributes.

### Bug

...

### To Reproduce

1. Checkout `04-28-chore_utility_func_attr_without_full_path_in_abi` branch in aztec-packages
2. bootstrap the repo
3. in `noir-projects/noir-contracts/target/auth_registry_contract-AuthRegistry.json` in search for `sync_notes` function and you should see "(quoted)" in the function attributes.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8255

## Summary

To do this:
- I wanted a test for this and one thing we can do is to start capturing snapshots of contract artifacts, which is what I did here (second commit)
- However, when I did that I noticed the output of functions was not deterministic. So the first commit makes it deterministic by sorting by function name.
- The third commit is the actual fix to the issue, and you can see in the diff how "(quoted)" changed to "utility" (and also how "super::add_utility" changed to "add_utility"), proving that the fix works 😊 

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_driver/src/lib.rs
+++ b/compiler/noirc_driver/src/lib.rs
@@ -594,7 +594,9 @@ fn compile_contract_inner(
             .iter()
             .filter_map(|attr| match &attr.kind {
                 SecondaryAttributeKind::Tag(contents) => Some(contents.clone()),
-                SecondaryAttributeKind::Meta(attribute) => Some(attribute.to_string()),
+                SecondaryAttributeKind::Meta(meta_attribute) => {
+                    context.def_interner.get_meta_attribute_name(meta_attribute)
+                }
                 _ => None,
             })
             .collect();
--- a/compiler/noirc_driver/src/lib.rs
+++ b/compiler/noirc_driver/src/lib.rs
@@ -497,16 +497,19 @@ pub fn compile_contract(
 fn read_contract(context: &Context, module_id: ModuleId, name: String) -> Contract {
     let module = context.module(module_id);
 
-    let functions = module
+    let mut functions: Vec<(ContractFunctionMeta, &str)> = module
         .value_definitions()
         .filter_map(|id| {
             id.as_function().map(|function_id| {
+                let name = context.def_interner.function_name(&function_id);
                 let attrs = context.def_interner.function_attributes(&function_id);
                 let is_entry_point = attrs.is_contract_entry_point();
-                ContractFunctionMeta { function_id, is_entry_point }
+                (ContractFunctionMeta { function_id, is_entry_point }, name)
             })
         })
         .collect();
+    functions.sort_by_key(|(_, name)| *name);
+    let functions = vecmap(functions, |(function, _)| function);
 
     let mut outputs = ContractOutputs { structs: HashMap::new(), globals: HashMap::new() };
 
--- a/compiler/noirc_driver/src/lib.rs
+++ b/compiler/noirc_driver/src/lib.rs
@@ -497,19 +497,16 @@ pub fn compile_contract(
 fn read_contract(context: &Context, module_id: ModuleId, name: String) -> Contract {
     let module = context.module(module_id);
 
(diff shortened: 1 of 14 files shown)


tooling/nargo_cli/tests/execute.rs
```diff
@@ -341,19 +360,50 @@ mod tests {
...
+        force_brillig: ForceBrillig,
+        inliner: Inliner,
+    ) {
+        let artifact_filename =
+            find_program_artifact_in_dir(target_dir).expect("Expected an artifact to exist");
+
+        let artifact_file = fs::File::open(&artifact_filename).unwrap();
+        let artifact: ContractArtifact = serde_json::from_reader(artifact_file).unwrap();
+
+        let _ = fs::remove_dir_all(target_dir);
```
michaeljklein, line 381: Is this silently ignoring failures?

michaeljklein (changes requested):

compiler/noirc_driver/src/lib.rs
```diff
@@ -497,16 +497,19 @@ pub fn compile_contract(
...
             id.as_function().map(|function_id| {
+                let name = context.def_interner.function_name(&function_id);
                 let attrs = context.def_interner.function_attributes(&function_id);
                 let is_entry_point = attrs.is_contract_entry_point();
-                ContractFunctionMeta { function_id, is_entry_point }
+                (ContractFunctionMeta { function_id, is_entry_point }, name)
             })
         })
         .collect();
+    functions.sort_by_key(|(_, name)| *name);
```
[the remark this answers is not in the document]
asterite: 1. I don't think so. You can't have two functions named the same inside a contract.
2. It isn't because the values come from a HashMap and that's not ordered.
3. No, because the functions here are just for one contract so it's impossible to end up with functions from different contracts here.

Then I thought: but wait! What if you have more than one contract in the same crate? But it seems that errors with "Packages are limited to a single contract" so I think there's no chance of collition.

tooling/nargo_cli/tests/execute.rs
asterite: Yes, but I'm not sure we should fail the test because of a missing cleanup. Or yes?

michaeljklein: Yes, I think so: missing cleanup could cause knock-on effects that we could have trouble debugging otherwise

compiler/noirc_driver/src/lib.rs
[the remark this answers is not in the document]
michaeljklein: > 2. It isn't because the values come from a HashMap and that's not ordered.

Ah, then would it make sense to use a `BTreeMap` or other more deterministic data structure for this so that we don't accidentally cause other forms of non-determinism by accessing it elsewhere?

tooling/nargo_cli/tests/execute.rs
asterite: Done!

compiler/noirc_driver/src/lib.rs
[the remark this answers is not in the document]
asterite: Done!

michaeljklein (approved):

jfecher (approved):
:+1:

PR #8276, opened 2025-04-29
fix: don't disallow writing to memory after passing it to brillig

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This fixes an issue that @MirandaWood was running into on her [batched blobs branch](https://github.com/AztecProtocol/aztec-packages/pull/13817).

@kashbrti and I found the root cause is that if a memory block is passed as an argument to a brillig call then the `CircuitSimulator` marks it as "used" at which point it will reject any future writes to that memory block. I don't see any reason for why the simulator should reject this so I've replaced this logic with a simple check to ensure that we don't perform any reads/writes from uninitialized memory blocks.

@guipublic I think you originally wrote this logic, can you explain the motivation behind the old behaviour in case I'm missing something?

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/acvm-repo/acvm/src/compiler/simulator.rs
+++ b/acvm-repo/acvm/src/compiler/simulator.rs
@@ -41,7 +41,9 @@ impl CircuitSimulator {
             Opcode::AssertZero(expr) => {
                 for (_, w1, w2) in &expr.mul_terms {
                     if !self.solvable_witness.contains(w1) {
-                        if !self.solvable_witness.contains(w2) {}
+                        if !self.solvable_witness.contains(w2) {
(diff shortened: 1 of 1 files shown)


TomAFrench:
> Should we note this behavior in the BrilligInputs docs?

I'm not clear what behaviour you mean, the ability to pass a memory block as input?

> Is the original example too complex for a Noir-level regression test?

We could construct one but I think a more focused test is better.

michaeljklein:
> I'm not clear what behaviour you mean, the ability to pass a memory block as input?

Mainly how the initialization works

PR #8279, opened 2025-04-29
fix: constants unify with implicitly added named generics

### LINKED ISSUE #8252: Using traits with associated constants in trait bounds in trait impls results in `No matching impl found`
Consider the program below

```noir

trait TraitWithAssociatedConstant{
    let N: u32;

    fn foo(_: Self) -> bool {
        true
    }
}

struct Foo{}

impl TraitWithAssociatedConstant for Foo {
    let N: u32 = 42;
}


struct Wrapper<T>{
    inner: T
}

impl<T> std::cmp::Eq for Wrapper<T>
where
    T: TraitWithAssociatedConstant,
{

    fn eq(self, other: Self) -> bool {
        self.inner.foo()
    }
}


fn main() {
    let wrapper = Wrapper{ inner: Foo {}};
    assert_eq(wrapper, wrapper);
}

```

This fails to compile with the error message:

```
% nargo compile                                                                                                                                                                                                                                                     ~/noir/temp master tom-box
error: No matching impl found for `Foo: TraitWithAssociatedConstant<N = <T as TraitWithAssociatedConstant>::N>`
   ┌─ src/main.nr:27:9
   │
27 │         self.inner.foo()
   │         -------------- No impl for `Foo: TraitWithAssociatedConstant<N = <T as TraitWithAssociatedConstant>::N>`
   │

```

I would expect that we would be able to detect this implementation as `Foo` has an implementation of `TraitWithAssociatedConstant`.


---


Note that if we replace the `std::cmp::Eq` trait implementation with a regular impl block then this will compile fine so it's purely an issue in trait impls.

```noir

trait TraitWithAssociatedConstant{
    let N: u32;

    fn foo(_: Self) -> bool {
        true
    }
}

struct Foo{}

impl TraitWithAssociatedConstant for Foo {
    let N: u32 = 42;
}


struct Wrapper<T>{
    inner: T
}

impl<T>  Wrapper<T>
where
    T: TraitWithAssociatedConstant,
{

    fn eq(self, other: Self) -> bool {
        self.inner.foo()
    }
}


fn main() {
    let wrapper = Wrapper{ inner: Foo {}};
    assert(wrapper.eq(wrapper));
}

```

asterite:
# Description

## Problem

Resolves #8252

## Additional Context

I think unbound named generics or unbound type variables should always unify with constants, but I just added the minimum necessary thing to solve the issue... mainly because I'm not entirely sure that's true.

## Documentation

Check one:

# PR Checklist


TomAFrench (approved):

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -1951,6 +1951,14 @@ impl Type {
...
                 }
             }
+            (
+                Constant(_, kind),
+                Type::NamedGeneric(types::NamedGeneric { type_var, implicit: true, .. }),
+            )
+            | (
+                Type::NamedGeneric(types::NamedGeneric { type_var, implicit: true, .. }),
+                Constant(_, kind),
+            ) if type_var.borrow().is_unbound() => kind.unify(&type_var.kind()),
```
jfecher, line 1961: So if the type variable is unbound we don't unify it - we only unify it's Kind?
So the original issue was that the constant kind and the type variable's Kind didn't match - maybe because the constant's kind was inferred?

I'm not sure this is the right solution. This'd let any implicitly added named generic unify with any constant value. E.g. you could unify it with both `5` and `8` as long as they all had the same kind. Usually we also don't bind NamedGenerics at all until monomorphization. I suppose binding the kind isn't bad but there's nothing preventing this from unifying against 5 and 8.

I haven't looked into it too much so I don't know what issue the original example stemmed from - if it was a constant with a polymorphic kind or maybe the kind in the NamedGeneric was polymorphic for some reason - but I don't think we should be unifying NamedGenerics with constants unless we know that NamedGeneric must be that constant. It'd be better to go all the way and actually unify `type_var` with the constant that way when you unify with `5` it'll later fail when you unify again with `8` - although I'm not 100% on this either and would need to investigate some more.

jfecher (changes requested):

compiler/noirc_frontend/src/hir_def/types.rs
asterite: Thanks! Do you mind taking a look at this, then? There's a small test for this in the frontend tests.

jfecher: Sure, I just got to a stopping point with the SSA interpreter. I'll take a look at this tomorrow.

asterite: Thank you! There's another PR that's related to this: https://github.com/noir-lang/noir/pull/8184 . There I changed the equality of these implicitly added named generics should it's maybe safer or more correct, though now I realize I didn't check their kind (but I won't fix that until understanding a bit more about all of this)

test_programs/compile_success_no_bug/noirc_frontend_tests_traits_implicit_generics_unification/src/main.nr
```diff
@@ -0,0 +1,38 @@
+
+    trait TraitWithAssociatedConstant {
+        let N: u32;
+
+        fn foo(_: Self) -> bool {
+            true
+        }
+    }
+
+    struct Foo {}
```
jfecher, line 10: Is this the right repro? This passes on master

jfecher: Testing against the original code in https://github.com/noir-lang/noir/issues/8252 I still get the same error with this PR

asterite: Gah, I copied the second snippet in the issue, which is very similar to the first one and works fine. I'll update the test.

asterite: Huh, it doesn't even reproduce if I don't use the `Eq` trait from the stdlib and `assert_eq`.

asterite: This is interesting. In the original snippet we have:

```noir
    assert_eq(wrapper, wrapper);
```

That fails to compile. If we change it to this:

```noir
    assert(wrapper == wrapper);
```

it still fails to compile.

But if we change it to this:

```noir
    assert(wrapper.eq(wrapper));
```

then it compiles, so I'm starting to suspect it has something to do with operator overloading.

Still need to update the test, I'll do it now.

asterite: Done!

compiler/noirc_frontend/src/hir_def/types.rs
asterite: > This'd let any implicitly added named generic unify with any constant value. E.g. you could unify it with both 5 and 8 as long as they all had the same kind.

What's a code example for this?

jfecher:
```
trait TraitWithAssociatedConstant{
    let N: u32;

    fn foo(_: Self) -> bool {
        true
    }

    fn make_array(self) -> [Field; Self::N];
}

struct Foo{}

impl TraitWithAssociatedConstant for Foo {
    let N: u32 = 5;

    fn make_array(self) -> [Field; Self::N] {
        std::mem::zeroed()
    }
}


struct Wrapper<T>{
    inner: T
}

impl<U> std::cmp::Eq for Wrapper<U>
where
    U: TraitWithAssociatedConstant,
{

    fn eq(self, _other: Self) -> bool {
        let _array1: [Field; 5] = self.inner.make_array();
        let _array2: [Field; 6] = self.inner.make_array();
        self.inner.foo()
    }
}


fn main() {
    let wrapper = Wrapper{ inner: Foo {}};
    assert_eq(wrapper, wrapper);
}
```
Type checks with this PR for example (although it fails during monomorphization with "Could not determine array length ..."

asterite: Thanks! I changed the code to still unify named generics with constants, but now they are tried to be unified with bindings, etc. (I copied the code from a bit above, I don't fully understand how type bindings work though I have an intuition). Now the old test still passes, and the program you put above fails to compile.

Would that be the correct solution?

jfecher: Unfortunately not, I tried that when I first looked at this PR but realized that'd permanently bind the generic when we don't want to bind named generics at all.

E.g. add another trait call
```
let wrapper = Wrapper{ inner: Bar {}};
assert_eq(wrapper, wrapper);
```
To another empty struct Bar with an identical impl but with `let N: u32 = 6;` and it now fails on Bar.

PR #8286, opened 2025-04-29
fix: allow names to collide in the values/types namespaces

asterite:
# Description

## Problem

Just something small I noticed that works in Rust but didn't in Noir.

## Summary

The compiler stores types and values in separate namespaces yet when a path is looked up and the same name exists in both namespaces it returned just one of them (the one in values). So in the end that namespace separation, I _think_, was never really used.

Now when looking up a path we need to specify if we want to search in the type namespace or in the value namespace. The other namespace is still looked up as a fallback option if nothing is found on the target namespace so we can give errors like "foo is not a type, it's a function".

## Documentation

Check one:

# PR Checklist


asterite:
I asked for a review earlier but the implementation was too complex. Now I found a way to implement this in a much simpler way.

jfecher (approved):
Neat!

PR #8289, opened 2025-04-30
fix(docs): bar -> foo correction

ciaranightingale:
# Description

I noticed that the docs referenced `bar::from_foo()` when this function is from the parent module `foo.nr` 

Let me know if I am misunderstanding!

## Problem\*

Comments reference function from child module rather than parent module.

## Documentation\*

https://noir-lang.org/docs/noir/modules_packages_crates/modules#referencing-a-parent-module

Check one:
- [ -] Documentation included in this PR.


aakoshh (approved):
LGTM, thanks!

michaeljklein (changes requested):
.

Issue #8271, opened 2025-04-29
AST fuzzer smoke tests fail in aztec-packages


aakoshh:
### Aim

https://github.com/AztecProtocol/aztec-packages/pull/13919

The smoke tests fail. The current `noir-repo-ref` is `nightly-2025-04-28`. 

### Expected Behavior

The smoke tests should not crash, unless there is still some bug in the compiler or the AST generator.

### Bug

http://ci.aztec-labs.com/list/ee2aaf4f842c6c24 shows the list of tests. Here are the failing examples:

* http://ci.aztec-labs.com/823d56115ad010f1 seed 0x3a81946100100000
* http://ci.aztec-labs.com/6629687952481e5d seed 0x5324728b00100000
* http://ci.aztec-labs.com/e87817fd7d4caefb seed 0xc7ace5fb00100000
* http://ci.aztec-labs.com/37fbf51c64241bff seed 0xca45d9d300100000

1-2 and 3-4 have the same error message. 

### To Reproduce

Run `NOIR_ARBTEST_SEED=0x... cargo test -p noir_ast_fuzzer --test smoke` with the above seeds.

### Workaround

None

### Project Impact

None

### Nargo Version

nightly-2025-04-28

### Would you like to submit a PR for this Issue?

None

aakoshh:
I reckon even after the fix in #8294 we can leave the smoke test marked as flaky in https://github.com/AztecProtocol/aztec-packages/pull/13919 for now, so we keep being pinged about any issues on Slack without failing the aztec-packages build.

PR #8294, opened 2025-04-30
fix: Return zero and insert an assertion if RHS bit size is over the limit in euclidian division

### LINKED ISSUE #8272: Only ACIR: `Range constraint of 254 bits is too large for the Field size` with `u32` modulo
### Aim

Minimised from the first failure in https://github.com/noir-lang/noir/issues/8271

The following program fails to compile as ACIR but does with Brillig:

```rust
fn main(b: [u32; 1]) {
    func_1(3032078109, b);
}
fn func_1(a: u32, mut b: [u32; 1]) {
    b[0] = (b[0] % (869374236 - a));
}
```

### Expected Behavior

They should agree. 

On a a general note, I'm not sure where we are on compile time overflow checks. I'm running the smoke tests under the impression that even something that we know will overflow (e.g. constants) should compile, and return a failure at runtime. The smoke test fails if it cannot compile a program, not just when it hits ICE. But in this case the fact that `--force-brillig` works indicates that we have another issue anyway.

### Bug

```console
$ cargo run -q -p nargo_cli -- compile --skip-underconstrained-check --skip-brillig-constraints-check --silence-warnings 
error: Range constraint of 254 bits is too large for the Field size
  ┌─ src/main.nr:6:13
  │
6 │     b[0] = (b[0] % (869374236 - a));
  │             ----------------------
  │
  = Call stack:
    1. src/main.nr:3:5
    2. src/main.nr:6:13

$ cargo run -q -p nargo_cli -- compile --skip-underconstrained-check --skip-brillig-constraints-check --silence-warnings  --force-brillig

```

### To Reproduce

See above. 

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+36db3d070a6614fc948566e98b39bd389fdfa3be (git version hash: e4835312386ecea728ba0e22580b3dac6cf0205f, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #8274: Compiler crash, Only ACIR: `attempt to shift left with overflow` (unused variable, modulo)
### Aim

This is a minimised version of the last failure in https://github.com/noir-lang/noir/issues/8271

The following program crashes the compiler, but passes with `--force-brillig`:

```rust
fn main(a: u8, b: u8) {
    let h = (((a as u64) >> b) % (11912905567223247326 * 14329851068374824036));
}
```

Maybe related to https://github.com/noir-lang/noir/pull/8243 (that fix is not included where this test failed, in `nightly-2028-04-28`). 

### Expected Behavior

Compiler should not crash. 

### Bug

```console
$ cargo run -q -p nargo_cli -- compile --skip-underconstrained-check --skip-brillig-constraints-check --silence-warnings --force --force-brillig
$ cargo run -q -p nargo_cli -- compile --skip-underconstrained-check --skip-brillig-constraints-check --silence-warnings --force 
Message:  attempt to shift left with overflow
Location: compiler/noirc_evaluator/src/acir/acir_context/mod.rs:996

```

### To Reproduce

See above. 

This is not exactly the failure on CI, where it was this instead:
```
18:46:46 thread 'arb_program_can_be_executed' panicked at compiler/noirc_evaluator/src/acir/acir_context/mod.rs:1000:13:
18:46:46 assertion failed: bits + r_bit_size < F::max_num_bits()
```

Both the 3rd and the 4th seed gave the error seen in this ticket instead.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+36db3d070a6614fc948566e98b39bd389fdfa3be (git version hash: e4835312386ecea728ba0e22580b3dac6cf0205f, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #8272 
Also fixes #8274

In https://github.com/noir-lang/noir/pull/8197 I changed the code to not panic if the RHS is larger than the expected bit size because of an overflow, but only for determining the quotient bit size, I still passed it to Brillig for the RHS and apparently also used it to set a range constraint on the result. 

This time the value underflowed, and the negative result was actually represented by a 254 bit Field, which triggered another compile time failure (not panic) for ACIR (but not Brillig).  

## Summary\*

~This PR changes the code to completely ignore the RHS bit sizes over the operand size, trusting that there simply a range constraint inserted where the overflow happens which will prevent us from seeing any adverse affects from doing so.~

This PR changes the code so that instead of ignoring the RHS bit size being over the operand size, it inserts an `assert_eq(0, 1, "attempted to divide by constant larger than operand type")` constraint, which will fail at runtime, unless of course we right assuming that there will have been an earlier overflow check which will catch the problem before we get here. 

In practice this is what we see when we try to compile and execute the program:
```console
❯ cargo run -q -p nargo_cli -- compile --force 
  ┌─ src/main.nr:5:13
  │
5 │     b[0] = (b[0] % (869374236 - a));
  │             ---------------------- As a result, the compiled circuit is ensured to fail. Other assertions may also fail during execution
  │
  = Call stack:
    1. src/main.nr:2:5
    2. src/main.nr:5:13

  ┌─ src/main.nr:5:21
  │
5 │     b[0] = (b[0] % (869374236 - a));
  │                     -------------
  │
  = Call stack:
    1. src/main.nr:2:5
    2. src/main.nr:5:21

Failed assertion
```

## Additional Context

This fixes both bugs that cause flakiness in aztec-packages captured in https://github.com/noir-lang/noir/issues/8271 when the `smoke.rs` test in `noir_ast_fuzzer` generates a random AST that doesn't compile.

I have not attempted to figure out how to replicate this behaviour for Brillig (the printing of a "bug"), since it does not fail to compile this program, and works identically at runtime.

## Documentation\*

Check one:

# PR Checklist\*


guipublic (changes requested):
lhs and rhs are supposed to be of size 'bit_size', if it is not the case, then this is an error.

I think at that point you should simply return 0 and add an assert(false, "invalid bit size in euclidian division, executed {bit_size}, got {max_rhs_bits}").

compiler/noirc_evaluator/src/acir/acir_context/mod.rs
```diff
@@ -840,6 +840,13 @@ impl<F: AcirField, B: BlackBoxFunctionSolver<F>> AcirContext<F, B> {
             } else {
                 bit_size - max_rhs_bits + 1
             };
+
+            // If the RHS is a result of an underflow, it could be a negative number which is represented by
+            // a very large positive Field, which could fail to compile to ACIR in `range_constrain_var` below,
+            // because it can use all 254 bits. To avoid that, ignore even what we know about the RHS and trust
+            // that a previous range constraint will fail at runtime before it comes to executing this instruction.
+            let max_rhs_bits = max_rhs_bits.min(bit_size);
```
michaeljklein, line 848: I think I understand the issue with underflowing, but I would expect a fix to be elsewhere and I would expect to bubble up this "will always fail at runtime" result as a warning (like we do with `assert(false)`.

How did this cause "flakiness in aztec-packages in http://ci.aztec-labs.com/list/ee2aaf4f842c6c24"? If this was e.g. panicking nondeterministically, I would expect it to be due to either 1) the `aztec-packages` test setup generating invalid inputs, or 2) non-determinism in Noir itself. I'm concerned that such potential bugs could be covered up by skipping this check.

Also, how can we tell at this point whether it's due to such an underflow (which we expect to check at runtime) or due to a different issue?

michaeljklein (changes requested):

compiler/noirc_evaluator/src/acir/acir_context/mod.rs
aakoshh: But, using bit_size is not the end of the world, as that will happen for any non-constant value anyway, which can also over/underflow at runtime

guipublic:
> @guipublic do you mean to insert an assertion instruction into ACIR?
Yes

> do you know how the Brillig track handles this by any chance?
I'd need to look, but the linked issue is about ACIR, saying it works for Brillig?

aakoshh:
@guipublic yes, as the ticket shows if we use —force-brillig it compiles

compiler/noirc_evaluator/src/acir/acir_context/mod.rs
aakoshh: @michaeljklein ti be clear Aztec-packages runs all Noir tests, including the AST fuzzer smoke tests. It just happens to have failed for them, not us.

regarding a fix elsewhere: I looked at it last time; where the binary instruction is handled it just does it on Fields, and the adds an overflow check. It does not care if it’s constant or not. Only here do we look at it. I didn’t want to rewrite how that works, as it seems to be fine for Brillig, only ACIR gen assumes stuff about constants.

aakoshh: Changed it to include an assertion as @guipublic suggested.

michaeljklein (approved):
LGTM

guipublic (approved):

PR #8298, opened 2025-04-30
fix(ssa): Remove unused calls to pure functions

### LINKED ISSUE #8233: Only Brillig: `attempted to subtract with overflow`

vezenovm:
# Description

## Problem\*

Resolves no issue, but a basic optimization that we were missing out on.

## Summary\*

We were marking all calls as being side effectual during DIE, causing us to not eliminate calls to pure functions. Pure w/ predicates functions can be handled in a separate PR that accesses whether the current predicate is false. In ACIR, as the pure function is inlined all of its internal instructions were ultimately removed. While in Brillig, the function call remained thus triggering a failure that we did not see in Brillig. 

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -390,7 +391,14 @@ fn can_be_eliminated_if_unused(
             Value::ForeignFunction(_) => false,
 
             // We must assume that functions contain a side effect as we cannot inspect more deeply.
```
TomAFrench, line 393:
```suggestion
```

Seems like this comment is superceded

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -390,7 +391,14 @@ fn can_be_eliminated_if_unused(
...
-            Value::Function(_) => false,
+            // We use purity to determine whether functions contain side effects.
+            // If we have an impure function, we cannot remove it even if it is unused.
+            Value::Function(function_id) => match function.dfg.purity_of(function_id) {
+                Some(Purity::Pure) => true,
+                Some(Purity::PureWithPredicate) => true,
```
jfecher, line 398: Is this right? Seems like this case should depend on the predicate value. E.g. a function with a `constrain` inside will be marked as this and we don't want to remove those unless we know the predicate is false.

vezenovm: Yeah you are correct. I just wanted to see the effects here. I hadn't looked deeply into the purity types but I see now that we mark functions with `constrain` as being `PureWithPredicate`. I am most likely going to add some more granular purity types.

jfecher: What would the more granular cases be for?

vezenovm: I was thinking along the lines of that handling `constrain` and binary ops are different depending on whether we want to deduplicate, hoist, or eliminate. However, after looking a bit deeper we probably just want to not eliminate checked binary ops at all as suggested by @TomAFrench. 

For now, I think I will just switch `PureWithPredicate` to `false`. We can add checking whether the predicate is false in a separate PR, as we will need to handle checking both control dependence for Brillig and the side effects var for ACIR.

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -389,8 +390,14 @@ fn can_be_eliminated_if_unused(
             // from the ACVM to the external world during execution.
             Value::ForeignFunction(_) => false,
 
-            // We must assume that functions contain a side effect as we cannot inspect more deeply.
-            Value::Function(_) => false,
+            // We use purity to determine whether functions contain side effects.
+            // If we have an impure function, we cannot remove it even if it is unused.
+            Value::Function(function_id) => match function.dfg.purity_of(function_id) {
+                Some(Purity::Pure) => true,
+                Some(Purity::PureWithPredicate) => true,
```
vezenovm, line 397:
```suggestion
                Some(Purity::PureWithPredicate) => false,
```

compiler/noirc_evaluator/src/ssa/opt/die.rs
vezenovm: Tracking here https://github.com/noir-lang/noir/issues/8302

vezenovm:
@jfecher @TomAFrench This doesn't really make an meaningful changes in our test programs but I added a unit test to display the functionality.

TomAFrench (approved):
LGTM

PR #8299, opened 2025-05-01
feat: Metamorphic AST fuzzing

### LINKED ISSUE #7875: Fuzz random equivalence AST mutations
Perform arbitrary equivalence transformations on parts of the AST, and test whether the side effects are correctly preserved when the two versions are compiled to SSA and executed. 

For example:
* `a + b` is equivalent to `b + a`
* `5` is equivalent to `2 + 3`
* etc. 

Maybe related to the concept of _Equivalence Modulo Inputs_. It's not the same as _Mutation Testing_ which focuses on introducing small mutations which result in a _non_-equivalent program, and checking whether the test suite catches the regression or not.

aakoshh:
# Description

## Problem\*

Part 1 of #7875 

## Summary\*

Apply random semantic preserving changes to the AST, then compare the execution results with the original.

- [x] `num_plus_minus_zero`
- [x] `bool_or_self`
- [x] `bool_xor_self`
- [x] `bool_xor_rand`
- [x] Limit the number of rewrites
- [x] Dynamic probability of rewrites
- [x] Skip morphing unreachable functions 

I'll follow up a second PR to implement Michael's ideas described in the ticket.

### Bugs discovered

*  https://github.com/noir-lang/noir/issues/8305: The AST generator actually uses literals for ACIR, so after discovering this bug I stopped mutating them in this fuzzer.
* https://github.com/noir-lang/noir/issues/8310
* https://github.com/noir-lang/noir/issues/8319 (technically it came from `init_vs_final`, but came down to inliner aggro)

## Documentation\*

Check one:

# PR Checklist\*


jfecher (approved):
:+1:

PR #8304, opened 2025-05-01
[docs] Output file command incorrect

ciaranightingale:
# Description

Need to specify an output file as ./target.vk not ./target


TomAFrench:
Thanks for opening this PR. There's been changes to the bb CLI recently and on current versions it's correct to just do `./target`

ciaranightingale:
Then the comment needs to change

I updated to the latest version and still had this issue. What version of bb requires ./target instead? So I can check my version.

PR #8312, opened 2025-05-02
chore: document flattening pass

### LINKED ISSUE #8121: Document SSA Pass - Flattening

guipublic:
# Description

## Problem\*

Resolves #8121 

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -698,12 +748,12 @@ impl<'f> Context<'f> {
                     Instruction::Constrain(lhs, rhs, message)
                 }
                 Instruction::Store { address, value } => {
-                    // If this instruction immediately follows an allocate, and stores to that
-                    // address there is no previous value to load and we don't need a merge anyway.
+                    // There is no side effect for 'Store' instructions on arrays allocated in the current branch, because
+                    // these arrays do not exist in other branches.
```
jfecher, line 752: I think the old comment may have been more accurate here. Since this check isn't for handling side-effects but rather for merging the previous value with the current value, which isn't possible as the comment notes if there was no previous value. Additionally it can be any value, not just arrays.

jfecher (commented):
Looks good - just one suggestion

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -326,6 +347,10 @@ impl<'f> Context<'f> {
     }
 
     /// Returns the current condition
+    /// The conditions are in a stack, they are added as conditional branches are encountered
```
TomAFrench, line 350:
```suggestion
    /// Returns the current condition
    ///
    /// The conditions are in a stack, they are added as conditional branches are encountered
```

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -782,6 +835,10 @@ impl<'f> Context<'f> {
         }
     }
 
+    /// 'Cast' the 'condition' to 'value' type
+    /// This needed because we need to multiply the condition with several values
+    /// in order to 'nullify' side-effects when the 'condition' is false (in 'handle_instruction_side_effects()' function).
+    /// Since the condition is a boolean, it can be safely casted to any other type.
```
TomAFrench, line 841:
```suggestion
    /// 'Cast' the 'condition' to 'value' type
    ///
    /// This needed because we need to multiply the condition with several values
    /// in order to 'nullify' side-effects when the 'condition' is false (in 'handle_instruction_side_effects()' function).
    /// Since the condition is a boolean, it can be safely casted to any other type.
```

TomAFrench (commented):
You need a empty line in order for rustdoc to separate out the short summary line.

jfecher (approved):
👍

PR #8322, opened 2025-05-02
feat: let static_assert accept any type for its message

### LINKED ISSUE #6884: Allow fmtstr to be passed to `static_assert` instead of str.
A minor feature request, if it's even possible. 

E.g. to print the values of numeric generics that are being compared with `static_assert`.

asterite:
# Description

## Problem

Resolves #6884

## Summary

At first I didn't know how we could turn any value into a string, but I noticed `println` can do that. So... I used the same approach here (printable types). If the static assertion is true, all the printable type information added to the SSA is eventually removed by DIE.

Here's a commit that shows the output of `static_assert` when given a fmtstr as a message: https://github.com/noir-lang/noir/pull/8322/commits/1c1de883c79c2e12a0c79ec2284af0928d7aa81e

## Additional Context

I made some refactors to be able to reuse some of the logic for decoding a PrintableValueDisplay from foreign call params from the noirc_evaluator crate.

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/ssa/opt/assert_constant.rs
```diff
@@ -108,27 +113,71 @@ fn evaluate_static_assert(
...
-            let message = function
-                .dfg
-                .get_string(arguments[1])
-                .expect("Expected second argument to be a string");
+        return Ok(false);
+    }
+
+    let call_stack = function.dfg.get_instruction_call_stack(instruction);
+    if !function.dfg.is_constant(arguments[0]) {
+        return Err(RuntimeError::StaticAssertDynamicPredicate { call_stack });
```
michaeljklein, line 134: Can this error be delayed until we've attempted to calculate the message?

compiler/noirc_evaluator/src/ssa/opt/assert_constant.rs
```diff
@@ -108,27 +113,71 @@ fn evaluate_static_assert(
...
+
+    match PrintableValueDisplay::<FieldElement>::try_from_params(&foreign_call_params) {
+        Ok(display_values) => {
+            let message = display_values.to_string();
             Err(RuntimeError::StaticAssertFailed { message, call_stack })
-        } else {
-            Err(RuntimeError::StaticAssertDynamicPredicate { call_stack })
         }
+        Err(err) => {
+            let message = match err {
```
michaeljklein, line 154: I think we should panic on ICE. Does this return a normal error or is it turned into a panic later?

compiler/noirc_printable_type/src/lib.rs
```diff
@@ -273,6 +274,178 @@ fn format_field_string<F: AcirField>(field: F) -> String {
...
+
+                struct_map.insert(field_key.to_owned(), field_value);
+            }
+
+            PrintableValue::Struct(struct_map)
+        }
+        PrintableType::Function { env, .. } => {
+            let field_element = field_iterator.next().unwrap();
+            let func_ref = PrintableValue::Field(field_element);
+            // we want to consume the fields from the environment, but for now they are not actually printed
```
michaeljklein, line 335: Does this need follow up?

tooling/nargo/src/foreign_calls/mod.rs
```diff
@@ -93,3 +94,14 @@ pub enum ForeignCallError {
     #[error("Failed to replay oracle transcript: {0}")]
     TranscriptError(String),
 }
+
+impl From<TryFromParamsError> for ForeignCallError {
+    fn from(err: TryFromParamsError) -> Self {
+        match err {
```
michaeljklein, line 100: Do both enums have the same variants? If so, could one be an alias of the other?

noir_stdlib/src/lib.nr
```diff
@@ -84,9 +84,9 @@ fn verify_proof_internal<let N: u32, let M: u32, let K: u32>(
 #[builtin(assert_constant)]
 pub fn assert_constant<T>(x: T) {}
 
-// Asserts that the given value is both true and known at compile-time
+// Asserts that the given value is both true and known at compile-time.
```
michaeljklein, line 87: Please update the docs with the new behavior

compiler/noirc_printable_type/src/lib.rs
```diff
@@ -273,6 +274,178 @@ fn format_field_string<F: AcirField>(field: F) -> String {
     "0x".to_owned() + &trimmed_field
 }
 
+/// Assumes that `field_iterator` contains enough field elements in order to decode the [PrintableType]
+pub fn decode_value<F: AcirField>(
```
michaeljklein, line 278: Please add some more complex tests for this function

michaeljklein (changes requested):

noir_stdlib/src/lib.nr
asterite: You mean, talking about message? What should I say about it?

tooling/nargo/src/foreign_calls/mod.rs
asterite: No, ForeingCallError has many more variants

compiler/noirc_printable_type/src/lib.rs
asterite: The comment was already there, it was just moved to a helper function

compiler/noirc_evaluator/src/ssa/opt/assert_constant.rs
asterite: Ah,I actually made that intentional do we don't have to calculate the message if it's not needed. Why you'd prefer it the other way around?

compiler/noirc_printable_type/src/lib.rs
asterite: I initially did this PR by copying the relevant methods I needed into the `assert_constant.rs` file. Then I tried to introduce some refactors to avoid that copying.

First, we need `try_from_params`. That was in `nargo` and we can't have `noirc_evaluator` depend on `nargo`. But... I didn't know where to move that. Well, `try_from_params` takes an array of `ForeignCallParam` and returns a `PrintableValueDisplay`. `PrintableValueDisplay` lives in the `noirc_printable_type` crate so I thought it made sense for it to be there (though it's a bit weird that `ForeignCallParam` is there, because that's from the acvm repo... but `noirc_printable_type` already depended on acvm 🤷)

Then, `try_from_params` depends on `decode_value`, which is in the noirc_abi. We can't have `noirc_printable_type` depend on the ABI so I moved it to `noirc_printable_type`... which maybe makes sense? The return value is a `PrintableValue`, a type that lives in `noirc_printable_type`.

But maybe I shouldn't have gone with all that refactor before checking with you all on what's the best way to approach this.

compiler/noirc_evaluator/src/ssa/opt/assert_constant.rs
asterite: Ah, I saw `RuntimeError::InternalError` already has some variants which produce "ICE: ..." as a message, that's why I copied that pattern. But it should be fine to use panics here. Should I?

noir_stdlib/src/lib.nr
asterite: We can say something like "The message can be a string, a format string, or any value, as long as it is known at compile-time". What do you think?

compiler/noirc_printable_type/src/lib.rs
vezenovm: > though it's a bit weird that `ForeignCallParam` is there, because that's from the acvm repo... but `noirc_printable_type` already depended on acvm 🤷

You mean that it is used there? Why do you find it weird?

> We can't have `noirc_printable_type` depend on the ABI so I moved it to `noirc_printable_type`... which maybe makes sense? The return value is a `PrintableValue`, a type that lives in `noirc_printable_type`

I think this is fine. `noirc_abi` already depends on `noirc_printable_type`. Let's just get rid of the one in `noirc_abi` and use this one. Thanks for the clarification that makes sense.

We are also implicitly testing this decoding through `stdout` in our integration tests. I think this is sufficient for now.

vezenovm (approved):

compiler/noirc_printable_type/src/lib.rs
asterite: > You mean that it is used there? Why do you find it weird?

It looks like the ABI is leaking into that crate, but... I guess that's fine, as the ABI is everywhere in the compiler and, well, it's the ABI after all :-)

> Let's just get rid of the one in noirc_abi and use this one.

Sounds good, I didn't know if it was desired to have a pub re-export from noirc_abi. I'll remove that re-export.

vezenovm: > Sounds good, I didn't know if it was desired to have a pub re-export from noirc_abi. I'll remove that re-export.

If we need it later on we can add it back.

asterite:
Ah, one more thing I'm remembering now. There were three copies of `decode_string_value`. Now there's one, in `noirc_printable_type`, but I don't know if that's correct because it has nothing to do with printable types. Should I keep the three copies? It's not that much code (around 10 lines). Otherwise, I don't know where turning a list of field element values into a string by assuming they are bytes could be done so it can be shared with other crates.

compiler/noirc_evaluator/src/ssa/opt/assert_constant.rs
michaeljklein: Yes: I think it would be easier to debug with the message

compiler/noirc_evaluator/src/ssa/opt/assert_constant.rs
michaeljklein: Yes, please use panics for ICE so that it's easier to detect them when testing, e.g. ICE will only be found by fuzzers when it panics.

noir_stdlib/src/lib.nr
michaeljklein: > You mean, talking about message?

Yes, and also [here](https://github.com/noir-lang/noir/blob/b8d04d6f1c5dd0bd304df970bf433f5fa60d826a/docs/docs/noir/concepts/assert.md?plain=1#L48), e.g. as mentioned for assert [here](https://github.com/noir-lang/noir/blob/b8d04d6f1c5dd0bd304df970bf433f5fa60d826a/docs/docs/noir/concepts/assert.md?plain=1#L31C1-L32C1).

> We can say something like "The message can be a string, a format string, or any value, as long as it is known at compile-time". What do you think?

Sounds great

compiler/noirc_evaluator/src/ssa/opt/assert_constant.rs
asterite: Done!

noir_stdlib/src/lib.nr
asterite: Done!

michaeljklein (approved):

Issue #8214, opened 2025-04-24
Use greybox fuzzer for `nargo test`


asterite:
The overall task is to start using the greybox fuzzer when doing `nargo test` instead of the old fuzzer.

Together with this we'd like to also account for these:
- [ ] There is an option to run only non-argument tests (the ones that don't require fuzzing) or only fuzzing test
- [ ] By default testing uses a temporary directory, but there is an option to specify one
- [ ] Fuzzing on each harness runs until it either reaches a certain number of iterations or a certain timeout (I guess 10k or 100k iterations by default and timeout of 10 seconds). Both are configurable.
- [ ] There is an option to trigger corpus minimization

I think at least the first point can be done even before switching to the new fuzzer.

PR #8326, opened 2025-05-02
feat: add `--fuzz-timeout` to `nargo test` options

asterite:
# Description

## Problem

Part of #8214

## Summary

I let the default timeout be 1 second. The original issue says 10 seconds but I think that's too long if tests are supposed to pass. But let me know otherwise! (additionally: 1 second was the default value before this PR too, though just for `nargo test`).

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/nargo/src/ops/test.rs
+++ b/tooling/nargo/src/ops/test.rs
@@ -49,6 +49,7 @@ impl TestStatus {
 
 pub struct FuzzConfig {
     pub folder_config: FuzzFolderConfig,
+    pub execution_config: FuzzExecutionConfig,
 }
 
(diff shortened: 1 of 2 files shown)


TomAFrench (approved):
If 1s was the default before I see no reason why we have to change it in this PR. We can modify in future if we want.

Rumata888:
> If 1s was the default before I see no reason why we have to change it in this PR. We can modify in future if we want.

Until we add the number of executions as a limiter, I agree with 1s. After we add the number of executions, we should switch to 10k or 100k execution or 10s. Whichever comes first

PR #8328, opened 2025-05-02
fix(ownership): Clone global arrays 

### LINKED ISSUE #8259: ACIR != Brillig: mutating global arrays with minimum inliner aggressiveness
### Aim

The following program was flagged up as different between the `--minimal-ssa` and the `--force-brillig` pass. When I tried to replicate, it turned out `--force-brillig` was correct and agreed with the ACIR version, but only with the default inliner setting.

```rust
global G_C: [bool; 3] = [true, false, true];
fn main(a: bool) -> pub [bool; 3] {
    let b = func_1(a, G_C);
    if a {
        G_C
    } else {
        b
    }
}
fn func_1(a: bool, mut b: [bool; 3]) -> [bool; 3] {
    b[1] = a;
    b
}
```

`Prover.toml`:

```toml
a = true
```

### Expected Behavior

ACIR and Brillig should agree. Brillig should not mutate the global array.

### Bug

```console
bash-5.2$ cargo run -q -p nargo_cli -- execute --skip-underconstrained-check --skip-brillig-constraints-check --silence-warnings --inliner-aggressiveness -9223372036854775808
[regression_fuzz_testing] Circuit output: Vec([Field(1), Field(0), Field(1)])
bash-5.2$ cargo run -q -p nargo_cli -- execute --skip-underconstrained-check --skip-brillig-constraints-check --silence-warnings --inliner-aggressiveness -9223372036854775808 --force-brillig
[regression_fuzz_testing] Circuit output: Vec([Field(1), Field(1), Field(1)])
```

I see that unlike with #8240 the monomorphized AST and the initial does not contain any cloning of the global array:
```
unconstrained fn main$f0(a$l0: bool) -> pub [bool; 3] {
    let b$l1 = func_1$f1(a$l0, G_C$g0);
    if a$l0 {
        G_C$g0
    } else {
        b$l1
    }
}
unconstrained fn func_1$f1(a$l2: bool, mut b$l3: [bool; 3]) -> [bool; 3] {
    b$l3[1] = a$l2;
    b$l3
}
```

Looking at the SSA with the default aggressiveness, the passing of the global array to `func_1` disappears during inlining:
```
bash-5.2$ cargo run -q -p nargo_cli -- execute --skip-underconstrained-check --skip-brillig-constraints-check --silence-warnings --inliner-aggressiveness 0 --force-brillig --show-ssa
...
After Preprocessing Functions (1):
g0 = u1 1
g1 = u1 0
g2 = make_array [u1 1, u1 0, u1 1] : [u1; 3]

brillig(inline) fn main f0 {
  b0(v3: u1):
    v6 = call f1(v3, g2) -> [u1; 3]
    jmpif v3 then: b1, else: b2
  b1():
    jmp b3(g2)
  b2():
    jmp b3(v6)
  b3(v4: [u1; 3]):
    return v4
}
  b0(v3: u1, v4: [u1; 3]):
    v6 = array_set v4, index u32 1, value v3
    return v6
}

After Inlining (1):
g0 = u1 1
g1 = u1 0
g2 = make_array [u1 1, u1 0, u1 1] : [u1; 3]

brillig(inline) fn main f0 {
  b0(v3: u1):
    v5 = make_array [u1 1, v3, u1 1] : [u1; 3]
    jmpif v3 then: b1, else: b2
  b1():
    jmp b3(g2)
  b2():
    jmp b3(v5)
  b3(v4: [u1; 3]):
    return v4
}
...
```

That's what gave me idea to try to see what happens if we don't inline what we don't have to.

### To Reproduce

See above.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #8259 

Taking the snippet from the issue:
```noir
global G_C: [bool; 3] = [true, false, true];
fn main(a: bool) -> pub [bool; 3] {
    let b = func_1(a, G_C);
    if a {
        G_C
    } else {
        b
    }
}
fn func_1(a: bool, mut b: [bool; 3]) -> [bool; 3] {
    b[1] = a;
    b
}
```
We have the following SSA:
```
g0 = u1 1
g1 = u1 0
g2 = make_array [u1 1, u1 0, u1 1] : [u1; 3]

  b0(v3: u1):
    v5 = call f1(v3, g2) -> [u1; 3]
    v6 = not v3
    v7 = if v3 then g2 else (if v6) v5
    return v7
}
brillig(inline) predicate_pure fn func_1 f1 {
  b0(v3: u1, v4: [u1; 3]):
    v6 = array_set v4, index u32 1, value v3
    return v6
}
```
The crux of the issue is a missing `inc_rc` on `g2`. This indicates a bug in the ownership pass and that we should clone the global. 

## Summary\*

We now clone any globals which contain arrays.

The snippet above produces the following SSA:
```
g0 = u1 1
g1 = u1 0
g2 = make_array [u1 1, u1 0, u1 1] : [u1; 3]

  b0(v3: u1):
    inc_rc g2
    v5 = call f1(v3, g2) -> [u1; 3]
    inc_rc g2
    v6 = not v3
    v7 = if v3 then g2 else (if v6) v5
    return v7
}
brillig(inline) predicate_pure fn func_1 f1 {
  b0(v3: u1, v4: [u1; 3]):
    v6 = array_set v4, index u32 1, value v3
    return v6
}
```

## Additional Context

This PR initially considering optimizing this process by initializing global arrays to have an RC > 1. This was chosen as unideal as it complicates array initialization in Brillig gen. It also essentially means our monomorphization output is false and being correct by Brillig gen. You can follow the discussion on this PR for why we moved away from this optimization to simplfy always cloning global arrays.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/ownership/mod.rs
```diff
@@ -209,6 +209,11 @@ impl Context {
                 return true;
             }
         }
+        // if let Definition::Global(_) = ident.definition {
+        //     if contains_array_or_str_type(&ident.typ) {
+        //         return true;
+        //     }
+        // }
```
vezenovm, line 216:
```suggestion
```

vezenovm:
I think it may be best to simply update the ownership pass. It would simplify Brillig gen and how we initialize arrays.

jfecher:
> I think it may be best to simply update the ownership pass. It would simplify Brillig gen and how we initialize arrays.

I think so too. It is an easy change in `should_clone_ident` to add the global case:
```rs
    fn should_clone_ident(&self, ident: &Ident) -> bool {
        match &ident.definition {
            Definition::Local(local_id) => {
                contains_array_or_str_type(&ident.typ) && !self.should_move(*local_id, ident.id)
            },
            // Globals are always cloned if they contain arrays
            Definition::Global(_) => contains_array_or_str_type(&ident.typ),
            _ => false,
        }
    }
```
This PRs version with starting RC of globals at 2 could be more efficient but I don't like that it means our monomorphized output is essentially incorrect and is relying on brillig to later correct it. For debugging in the future it can also complicate the mental model if we have to think about global arrays in SSA differently from normal arrays.

jfecher:
Aside: all the snap changes make this PR look much larger than it actually is

vezenovm:
> I would like to track globals in last uses as well

For now, I will make this a follow-up as to fix the issue first.

jfecher:
> I would like to track globals in last uses as well

We shouldn't track globals in last_uses since we should never move globals values. We don't know the ordering globals are used in in-between function calls which prevents us from determining where they're last used generally. If we limit the analysis to only entry point functions, these can be called recursively. I think it'd be a lot of effort to remove just 1 inc_rc from each global in the best case.

vezenovm:
> We shouldn't track globals in last_uses since we should never move globals values. We don't know the ordering globals are used in in-between function calls which prevents us from determining where they're last used generally. If we limit the analysis to only entry point functions, these can be called recursively. I think it'd be a lot of effort to remove just 1 inc_rc from each global in the best case.

Ah yes! Good point that would way too much effort. Makes things easier!

compiler/noirc_evaluator/src/brillig/brillig_ir/codegen_memory.rs
```diff
@@ -403,39 +403,20 @@ impl<F: AcirField + DebugToString, Registers: RegisterAllocator> BrilligContext<
...
-
-        // Write capacity
-        self.codegen_usize_op_in_place(write_pointer, BrilligBinaryOp::Add, 1);
-        self.store_instruction(write_pointer, capacity.unwrap_or(size).address);
-
-        self.deallocate_register(write_pointer);
     }
 
-    /// Initializes a vector, allocating memory to store its representation and initializing the reference counter, size and capacity
+    /// Initializes a vector, allocating memory to store its representation and initializing the reference counter
```
vezenovm, line 419:
```suggestion
    /// Initializes a vector, allocating memory to store its representation and initializing the reference counter, size and capacity
```

compiler/noirc_evaluator/src/brillig/brillig_ir/codegen_memory.rs
```diff
@@ -456,6 +437,28 @@ impl<F: AcirField + DebugToString, Registers: RegisterAllocator> BrilligContext<
         self.codegen_initialize_vector_metadata(vector, size, capacity);
     }
 
+    /// Writes vector metadata (reference count, size, and capacity) into the allocated memory
+    pub(super) fn codegen_initialize_vector_metadata(
```
vezenovm, line 441: I just moved this after `codegen_initialize_vector` and changed the vis to `super`

test_programs/execution_success/global_slice_rc_regression_8259/src/main.nr
```diff
@@ -0,0 +1,9 @@
+global G_C: [bool] = &[true, false, true];
+fn main(a: bool) {
+    println(func_1(a, G_C));
```
vezenovm, line 3: We cannot return slices from main so this was the simplest way to trigger the same bug but with slices.

vezenovm:
@jfecher this is ready for review again. As you said the actual changes are very small and almost all of the diff are snapshots.

jfecher (approved):
LGTM

PR #8338, opened 2025-05-05
fix: Ordering of items in callstacks

sirasistant:
# Description

## Problem\*

Resolves: Wrong order in callstack resolution

## Summary\*

We are exploring the locations tree child first. However, callstacks are parent=>child. That implies that using push was ordering in reverse the callstack items.

## Documentation\*

Check one:

# PR Checklist\*


sirasistant:
Found another issue in the ordering for acir => brillig calls

vezenovm (approved):

tooling/noir_js/src/debug.ts
```diff
@@ -110,7 +110,8 @@ function getCallStackFromLocationNode(
 
     callStack = location_tree[callStack.parent];
   }
-  return result;
+  // Reverse since we explored the child nodes first
+  return result.reverse();
```
vezenovm, line 114: I switched to just reverse at the end of the loop after pushing the stack rather than unshifting for every element. `unshift` has a complexity of O(n) where `n` is size of the array vs an O(1) push per element and just reversing a single time at the end of the loop. cc @sirasistant

Issue #8337, opened 2025-05-05
Comptime disagrees with Brilllig (handling conditional reassignment)


rkarabut:
### Aim

The following program returns [1429786010, 955969569]:

```
fn main() -> pub [i32; 2] {                                                                                                                                                                   
    comptime {                                                                                                                                                                                
        let mut ctx_limit = 25;                                                                                                                                                               
        unsafe { func_1_proxy(([1027925660, -1166341724], (-725726628, -262097387523026142812360820973506556332, -200882972683916250307467337508223428900, 418428712, true), false), ctx_limit) }                                                                                                                                                                                           
    }                                                                                                                                                                                         
}   

unconstrained fn func_1(mut a: ([i32; 2], (i32, Field, Field, i32, bool), bool), ctx_limit: &mut u32) -> [i32; 2] {  
    if ((*ctx_limit) == 0) {
        [1429786010, 955969569]
    } else {
        *ctx_limit = ((*ctx_limit) - 1);
        a.1 = if a.1.4 {                                                                                                                                                               
            a.0 = {                                                                                                                                                                           
                {                                                                                                                                                                             
                    a.1 = a.1;                                                                                                                                                                
                    a = (func_1(a, ctx_limit), if false {                                                                                                                                     
                        if a.2 {                                                                                                                                                                                          (if a.2 {                                                                                                                                                         
                                a.0[(if a.1.4 {                                                                                                                                               
                                    if (false != a.1.4) {                                                                                                                                     
                                        518870855                                                                                                                                             
                                    } else {                                                                                                                                                  
                                        2359665939                                                                                                                                            
                                    }                                                          
                                } else {                                                                                                                                                      
                                    if (-310728077952386577195164557033335247947 == -250209217142728108008477671960776659104) {                                                               
                                        if (a.2 <= a.1.4) {                                                                                                                                   
                                            2326121486                                                                                                                                        
                                        } else {                                                                                                                                              
                                            2041231551                                                                                                                                        
                                        }                                                                                                                                                     
                                    } else {                                                                                                                                                  
                                        3793351281                                                                                                                                            
                                    }                                                                                                                                                         
                                } % 2)]                                                                                                                                                       
                            } else {                                                                                                                                                          
                                a.0[0]                                                                                                                                                        
                            }, a.1.2, (a.1.4 as Field), a.0[0], (-58363765325703312851688177430427377548 != -261065546223799612669139593700581409424))                                        
                        } else {                                                                                                                                                              
[...]

Issue #8262, opened 2025-04-29
Init vs Final SSA: array mutation; ACIR and Brillig both wrong


aakoshh:
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8216

The following program returns a different result with `--minimal-ssa` compared to both ACIR and Brillig:

```rust
fn main() -> pub [Field; 3] {
    let mut a = [10, 20, 30];
    a[1] = {
        a = {
            a[2] = a[0];
            [a[0], 40, a[2]]
        };
        a[1]
    };
    a
}
```

### Expected Behavior

The expected return value is `[10, 40, 10]`, confirmed by [Rust](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=4b65f09cb7d373d223962d3c7fc84dc1)

### Bug

```console
$ cargo run -q -p nargo_cli -- execute --skip-underconstrained-check --skip-brillig-constraints-check --silence-warnings --force 
[regression_fuzz_testing] Circuit witness successfully solved
[regression_fuzz_testing] Witness saved to target/regression_fuzz_testing.gz
[regression_fuzz_testing] Circuit output: Vec([Field(10), Field(40), Field(30)])
$ cargo run -q -p nargo_cli -- execute --skip-underconstrained-check --skip-brillig-constraints-check --silence-warnings --force --force-brillig
[regression_fuzz_testing] Circuit witness successfully solved
[regression_fuzz_testing] Witness saved to target/regression_fuzz_testing.gz
[regression_fuzz_testing] Circuit output: Vec([Field(10), Field(40), Field(30)])
$ cargo run -q -p nargo_cli -- execute --skip-underconstrained-check --skip-brillig-constraints-check --silence-warnings --force --minimal-ssa
[regression_fuzz_testing] Circuit witness successfully solved
[regression_fuzz_testing] Witness saved to target/regression_fuzz_testing.gz
[regression_fuzz_testing] Circuit output: Vec([Field(10), Field(40), Field(10)])
$ 


```

### To Reproduce

See above.

### Workaround

None

### Project Impact

None

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+b0d9c680c3e9f6a818382b4f57ae2b5a93c657ab (git version hash: 0879bacabb4dcbf5f436575049b464e8fb35abac, is dirty: true)

### Would you like to submit a PR for this Issue?

None

aakoshh:
[...]

PR #8342, opened 2025-05-05
fix: Fix evaluation order of assignments with side-effects in their rhs

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/8262
Resolves https://github.com/noir-lang/noir/issues/8337

## Summary\*

Changed evaluation order of assign-subexpressions to match the comptime interpreter.

## Additional Context

@vezenovm mentioned here https://github.com/noir-lang/noir/issues/8262#issuecomment-2851205648 that we could also desugar the blocks to remove any internal statements beforehand. This would still be beneficial since we can still get confusing results by doing:
```noir
a.0[{ a = ([1,2], 3); 1  }] = 7;
```
ie. assigning within the lvalue will now assign _after_ the `[4, 5]` assignment instead of before. This is probably unexpected though it should be internally consistent now at least. That being said, changing this to remove internal statements beforehand to remove all sub-expressions requires an entire compiler pass which would be a larger pass so I thought I'd get this out first.

Edit: We're still not completely consistent here. There are differences between the comptime interpreter and SSA/brillig in compiling the above example but both are incorrect.

## Documentation\*

Check one:

# PR Checklist\*


vezenovm (approved):
Sounds like we still have a failure for the case you posted in the description?
```noir
a.0[{ a = ([1,2], 3); 1  }] = [4, 5];
```
Can we add a failing test for this before merging with a link to a new issue? Happy to translate my comments on https://github.com/noir-lang/noir/issues/8262 into an issue as well

jfecher:
@vezenovm added, bug is here: https://github.com/noir-lang/noir/issues/8344. It's actually worse than I thought since we're not consistent between comptime & SSA.

PR #8343, opened 2025-05-05
fix: let comptime Field value carry the field's sign

### LINKED ISSUE #8341: Comptime function won't return i32 from -1 literal
### Aim

This program fails:

```noir
fn main() -> pub i32 {
    comptime {
        -1
    }
}
```

The error is

```
error: The value `-1` cannot fit into `i32` which has range `-2147483648..=2147483647`
  ┌─ src/main.nr:2:5
  │  
2 │ ╭     comptime {
3 │ │         -1
4 │ │     }
  │ ╰─────'
  │  
  = Call stack:
    1. src/main.nr:2:5
```

The same seems to happen with other signed types.

### Expected Behavior

Function returns -1.

### Bug

```
nargo execute --force
error: The value `-1` cannot fit into `i32` which has range `-2147483648..=2147483647`
  ┌─ src/main.nr:2:5
  │  
2 │ ╭     comptime {
3 │ │         -1
4 │ │     }
  │ ╰─────'
  │  
  = Call stack:
    1. src/main.nr:2:5
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8341

## Summary

When the compiler saw `comptime { -1 }` it would evaluate that and store it as a `Value::Field`. The problem was that `Value::Field` carried just a `FieldElement`, so in this case the sign was lost and we ended up with the largest FieldElement value.

In this PR `Value::Field` now carries a `SignedField`.

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -1656,7 +1657,10 @@ fn expr_as_binary_op(
             let binary_op_value = infix_expr.operator.contents as u128;
 
             let mut fields = HashMap::default();
-            fields.insert(Rc::new("op".to_string()), Value::Field(binary_op_value.into()));
+            fields.insert(
+                Rc::new("op".to_string()),
+                Value::Field(SignedField::positive(binary_op_value)),
```
asterite, line 1662: Can the sign be ignored here?

compiler/noirc_frontend/src/signed_field.rs
```diff
@@ -82,6 +82,61 @@ impl SignedField {
...
+        } else if self.is_negative && !rhs.is_negative {
+            if self.field > rhs.field {
+                // For example "-4 + 3", so "-(4 - 3)"
+                Self::new(self.field - rhs.field, true)
+            } else {
+                // For example "-4 + 5", so "5 - 4"
+                Self::new(rhs.field - self.field, false)
+            }
+        } else {
+            // !self.is_negative && rhs.is_negative
```
vezenovm, line 106:
```suggestion
```

compiler/noirc_frontend/src/signed_field.rs
```diff
@@ -204,4 +259,67 @@ mod tests {
...
+    fn multiplication() {
+        let two = SignedField::positive(2_u32);
+        let three = SignedField::positive(3_u32);
+        let six = SignedField::positive(6_u32);
+        let minus_two = SignedField::negative(2_u32);
+        let minus_three = SignedField::negative(3_u32);
+        let minus_six = SignedField::negative(6_u32);
+
+        assert_eq!(two * three, six); // positive * positive
+        assert_eq!(two * minus_three, minus_six); // possitive * negative
```
vezenovm, line 306:
```suggestion
        assert_eq!(two * minus_three, minus_six); // positive * negative
```
nit

compiler/noirc_frontend/src/signed_field.rs
```diff
@@ -204,4 +259,67 @@ mod tests {
...
+    fn division() {
+        let two = SignedField::positive(2_u32);
+        let three = SignedField::positive(3_u32);
+        let six = SignedField::positive(6_u32);
+        let minus_two = SignedField::negative(2_u32);
+        let minus_three = SignedField::negative(3_u32);
+        let minus_six = SignedField::negative(6_u32);
+
+        assert_eq!(six / two, three); // positive * positive
+        assert_eq!(six / minus_three, minus_two); // possitive * negative
```
vezenovm, line 321:
```suggestion
        assert_eq!(six / minus_three, minus_two); // positive * negative
```

compiler/noirc_frontend/src/signed_field.rs
```diff
@@ -204,4 +259,67 @@ mod tests {
 
         assert!(two > neg_two);
     }
+
+    #[test]
+    fn addition() {
```
vezenovm, line 264: These tests look good! Could we add some cases with zero to these operators before merging though?

vezenovm (approved):

compiler/noirc_frontend/src/signed_field.rs
```diff
@@ -82,6 +82,61 @@ impl SignedField {
...
+            }
+        }
+    }
+}
+
+impl std::ops::Sub for SignedField {
+    type Output = Self;
+
+    fn sub(self, rhs: Self) -> Self::Output {
+        self + (-rhs)
```
jfecher, line 121:
```suggestion
        self + -rhs
```

compiler/noirc_frontend/src/signed_field.rs
jfecher: I think tests to show we can't construct a `-0` value would be good too. Maybe it should be an assertion when constructing a SignedField as well?

asterite: Done! Yes, when I added tests for zero what you mentioned started to happen, so I had to add a few `if`s to account for that.

PR #8350, opened 2025-05-05
fix: preserve type when turning comptime Value into an Expression

### LINKED ISSUE #8340: Returning i64 requires explicit cast normally but not in comptime
### Aim

Running this successfully returns 1:

```noir
fn main() -> pub i64 {
    comptime {
        unsafe { func_1(1) }
    }
}

unconstrained fn func_1(b: i32) -> i32 {
    b                                                                                                                                                             
}
```

But removing comptime leads to an error:
```
error: expected type i64, found type i32
  ┌─ src/main.nr:1:18
  │
1 │ fn main() -> pub i64 {
  │                  --- expected i64 because of return type
2 │     //comptime {
3 │         unsafe { func_1(1) }
  │         --------------------- i32 returned here
  │
```

### Expected Behavior

Running the code with and without `comptime` should agree on results, or fail in a similar way.

### Bug

```
nargo execute --force
[test1] Circuit output: Field(1)
```

After removing `comptime`:
```
error: expected type i64, found type i32
  ┌─ src/main.nr:1:18
  │
1 │ fn main() -> pub i64 {
  │                  --- expected i64 because of return type
2 │     //comptime {
3 │         unsafe { func_1(1) }
  │         -------------------- i32 returned here
  │

```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8340

## Summary

For example, given this code:

```noir
fn main() -> pub i64 {
    comptime {
        1 as i32
    }
}
```

The compiler would see `comptime { 1 as i32 }`. The type of `1 as i32` is `i32`. But the compiler also interprets `1 as i32` and gets a "1" as a `Value::I32`. Then it turns that comptime value into an Expression and get the literal "1", then it types that (we get a polymorphic integer) and that type was given to the `comptime` block. That polymorphic integer matches fine with `i64`, that's why we didn't previously get an error.

In this PR that `Value::I32` is turned into `1 as i32` in `into_expression` so the type is perserved.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
@@ -2,7 +2,6 @@ use std::collections::VecDeque;
 use std::{collections::hash_map::Entry, rc::Rc};
 
 use acvm::blackbox_solver::BigIntSolverWithId;
-use acvm::{FieldElement, acir::AcirField};
 use im::Vector;
 use iter_extended::try_vecmap;
 use noirc_errors::Location;
@@ -859,9 +858,9 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
 
         use BinaryOpKind::*;
         let less_or_greater = if matches!(operator, Less | GreaterEqual) {
-            FieldElement::zero() // Ordering::Less
+            SignedField::zero() // Ordering::Less
         } else {
-            2u128.into() // Ordering::Greater
+            SignedField::positive(2u128) // Ordering::Greater
         };
 
         if matches!(operator, Less | Greater) {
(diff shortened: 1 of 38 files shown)


jfecher (approved):
This PR looks good too

asterite:
I'll go with this one after merging conflicts and fixing some tests (mainly some `nargo expand` programs changed to include the new cast)

PR #8355, opened 2025-05-05
fix(ssa): Mislabeled instructions with side effects in EnableSideEffectsIf removal pass

### LINKED ISSUE #8236: ACIR != Brillig: passing unused parameters
### Aim

The following program returns different result in ACIR vs Brillig:

```rust
global G_C: u8 = 158;
fn main(a: (Field, str<4>, [bool; 4], bool, [str<3>; 1])) -> pub bool {
    let mut ctx_limit = 25;
    (
        195033874435849113062101315935534676719
            > (
                func_5(
                    [],
                    [
                        (a.3, -580975298, a.3, a.4[0], 2312041605788857155769242809839111790),
                        (a.3, -1047272313, a.3, a.4[0], (a.3 as Field)),
                        (false, -1615095934, a.3, "MGZ", (G_C as Field)),
                    ],
                    (&mut ctx_limit),
                ) as u128
            )
    )
}
unconstrained fn func_1(
    a: ((bool, i32, bool, str<3>, Field), [bool; 3], [u16; 4]),
) -> (str<2>, bool, u8, bool) {
    ("PY", true, 203, true)
}
unconstrained fn func_2(
    a: [(bool, i32, bool, str<3>, Field); 3],
    b: (str<4>, u16, u8, bool, [bool; 4]),
    ctx_limit: &mut u32,
) -> (Field, str<4>, [bool; 4], bool, [str<3>; 1]) {
    if ((*ctx_limit) == 0) {
        (
            -138077626322456242457068228915022246858, "UDJA", [false, true, true, false], false,
            ["LIA"],
        )
    } else {
        *ctx_limit = ((*ctx_limit) - 1);
        (
            (
                if (1683200272229996428 <= -8184679232303581313) {
                    (-(G_C as Field))
                } else {
                    (G_C as Field)
                }
                    / (G_C as Field)
            ),
            func_2(
                if func_1((a[2], [b.4[1], b.4[0], true], [60430, 36232, 4452, 29090])).1 {
                    [a[(2075443353 % 3)], a[0], a[2]]
                } else {
                    [
                        a[((G_C as u32) % 3)],
                        a[((G_C as u32) % 3)],
                        (false, a[0].1, (!true), func_2(a, b, ctx_limit).4[0], (G_C as Field)),
                    ]
                },
                b,
                ctx_limit,
            )
                .1, b.4, (b.2 == (b.2 / b.2)),
            if b.3 {
                [a[1].3]
            } else {
                [a[((b.2 as u32) % 3)].3]
            },
        )
    }
}
fn func_5(a: [Field; 0], b: [(bool, i32, bool, str<3>, Field); 3], ctx_limit: &mut u32) -> Field {
    if ((*ctx_limit) == 0) {
        -116691602079003650843727709629323708719
    } else {
        *ctx_limit = ((*ctx_limit) - 1);
        (
            -(
                unsafe {
                    func_1((
                        (false, 1033139569, b[0].2, b[1].3, (b[0].0 as Field)),
                        [b[2].0, false, true], [42093, 59962, 15862, 26716],
                    ))
                }
                    .1 as Field
            )
        )
    }
}
```

`Prover.toml`:
```toml
a = [
    "0x14c6e4c56b87071d0ee3c4b9a32352235577e833c97e1b31b3d48bd8b812c75c",
    "ABCD",
    [
        true,
        false,
        false,
        false,
    ],
    false,
    [
        "DEF",
    ],
]
```

### Expected Behavior

They should agree. 

### Bug

```console
bash-5.2$ cargo run -q -p nargo_cli -- execute --skip-underconstrained-check --skip-brillig-constraints-check --silence-warnings
bash-5.2$ cargo run -q -p nargo_cli -- execute --skip-underconstrained-check --skip-brillig-constraints-check --silence-warnings --force-brillig
[regression_fuzz_testing] Circuit output: Field(1)
```

If I delete the parameters of `func_1`, which are all unused, then ACIR also returns 1. 

### To Reproduce

See above.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+b0d9c680c3e9f6a818382b4f57ae2b5a93c657ab (git version hash: 1f595f71320be9464dc527946c88d9bdf264dd9f, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #8236 

For the given snippet in the issue we have the following SSA before constant folding:
<p>

```
  b0(v8: [u16; 3], v9: [u1; 1]):
    v10 = make_array [u16 59890, u16 17417, u16 14409] : [u16; 3]
    v13 = call f1(v10, u1 1) -> [u1; 1]
    v15 = array_get v8, index u32 0 -> u16
    v16 = cast v15 as u32
    v18 = mod v16, u32 14409
    v19 = array_get v13, index u32 0 -> u1
    enable_side_effects v19
    v20 = array_get v9, index u32 0 -> u1
    v21 = call f1(v8, v20) -> [u1; 1]
    v22 = array_get v9, index u32 0 -> u1
    v23 = not v19
    enable_side_effects v23
    v24 = array_get v8, index u32 0 -> u16
    v25 = cast v24 as u32
    v27 = mod v25, u32 3
    v28 = lt v27, u32 3
    v29 = unchecked_mul v28, v23
    constrain v29 == v23, "Index out of bounds"
    v30 = array_get v8, index v27 -> u16
    v31 = cast v30 as u32
    v33 = lt u32 24993, v31
    enable_side_effects u1 1
    v34 = unchecked_mul v19, v22
    v35 = unchecked_mul v23, v33
    v36 = unchecked_add v34, v35
    return v36
}
```

</p>
</details> 

After Constant Folding:
<p>

```
  b0(v8: [u16; 3], v9: [u1; 1]):
    v10 = make_array [u16 59890, u16 17417, u16 14409] : [u16; 3]
    v13 = call f1(v10, u1 1) -> [u1; 1]
    v15 = array_get v8, index u32 0 -> u16
    v16 = cast v15 as u32
    v18 = mod v16, u32 14409
    v19 = array_get v13, index u32 0 -> u1
    enable_side_effects v19
    v20 = array_get v9, index u32 0 -> u1
    v21 = call f1(v8, v20) -> [u1; 1]
    v22 = not v19
    enable_side_effects v22
    v24 = mod v16, u32 3
    v25 = lt v24, u32 3
    v26 = unchecked_mul v25, v22
    constrain v26 == v22, "Index out of bounds"
    v27 = array_get v8, index v24 -> u16
    v28 = cast v27 as u32
    v30 = lt u32 24993, v28
    enable_side_effects u1 1
    v31 = unchecked_mul v19, v20
    v32 = unchecked_mul v22, v30
    v33 = unchecked_add v31, v32
    return v33
}
```

</p>
</details> 

After EnableSideEffectsIf removal:
<p>

```
  b0(v8: [u16; 3], v9: [u1; 1]):
    v10 = make_array [u16 59890, u16 17417, u16 14409] : [u16; 3]
    v13 = call f1(v10, u1 1) -> [u1; 1]
    v15 = array_get v8, index u32 0 -> u16
    v16 = cast v15 as u32
    v18 = mod v16, u32 14409
    v19 = array_get v13, index u32 0 -> u1
    enable_side_effects v19
    v20 = array_get v9, index u32 0 -> u1
    v21 = call f1(v8, v20) -> [u1; 1]
    v22 = not v19
    v24 = mod v16, u32 3
    v25 = lt v24, u32 3
    enable_side_effects v22
    v26 = unchecked_mul v25, v22
    constrain v26 == v22, "Index out of bounds"
    v27 = array_get v8, index v24 -> u16
    v28 = cast v27 as u32
    v30 = lt u32 24993, v28
    enable_side_effects u1 1
    v31 = unchecked_mul v19, v20
    v32 = unchecked_mul v22, v30
    v33 = unchecked_add v31, v32
    return v33
}
```

</p>
</details> 

In the initial SSA, can see how `v24 = array_get v8, index u32 0 -> u16` was deduplicated to a matching instruction that occurred under a different predicate. The result `v24` is then used to compute the dynamic index of another array get. Those instructions calculating the index are being moved incorrectly causing an incorrect index to be computed.

Looking at the new SSA (After EnableSideEffectsIf removal) with the deduplicated array_get, we now use the actual result `v15` from `v15 = array_get v8, index u32 0 -> u16` when computing `v16 = cast v15 as u32` -> `v24 = mod v16, u32 3` -> `v27 = array_get v8, index v24 -> u16`. 

After the EnableSideEffectsIf removal pass the modulo `v24 = mod v16, u32 3` occurs under a different predicate from where it was previously. If we have a zero predicate our [euclidean division returns zero](https://github.com/noir-lang/noir/blob/76c9937918ca59fa2b36abc5e106307d362ea338/compiler/noirc_evaluator/src/acir/acir_context/mod.rs#L772) for both the quotient and the remainder. 

We have this final SSA with the enable side effects vars simplified:
```
  b0(v8: [u16; 3], v9: [u1; 1]):
    v11 = array_get v8, index u32 0 -> u16
    v12 = cast v11 as u32
    enable_side_effects u1 0
    v14 = array_get v9, index u32 0 -> u1
    v16 = call f1(v8, v14) -> [u1; 1]
    v18 = mod v12, u32 3
    v19 = lt v18, u32 3
    enable_side_effects u1 1
    constrain v19 == u1 1, "Index out of bounds"
    v21 = array_get v8, index v18 -> u16
    v22 = cast v21 as u32
    v24 = lt u32 24993, v22
    enable_side_effects u1 1
    return v24
}
```
We can see how the modulo is now under a false predicate when it would have been under a true predicate previously. Thus, we are getting a result of zero from the modulo operation instead of one as we would like, triggering an incorrect result.

## Summary\*

I simply switched to using `Instruction::requires_acir_gen_predicate` instead of `responds_to_side_effects_var` where `Binary::requires_acir_gen_predicate` is called internaly and `BinaryOp::Div`/`BinaryOp::Mod` are correctly labeled as always requiring ACIR gen predicates.  

I could have went with just updating the `responds_to_side_effects_var` function inside of `remove_enable_side_effects.rs`, however, I felt it would be better to lean into `Instruction::requires_acir_gen_predicate` as they are attempting to do the same thing.
Comments above the call to `responds_to_side_effects_var`:
```
// If we hit an instruction which is affected by the side effects var then we must insert the
// `Instruction::EnableSideEffectsIf` before we insert this new instruction.
```
Comments above `Instruction::requires_acir_gen_predicate`:
```
/// If true the instruction will depend on `enable_side_effects` context during acir-gen.
```
We also had other mismatches between the methods on what instructions returned true for being affected by the side effects var. I felt it is better to be unified here and have alignment on which instructions depend on an ACIR predicate. 

[Disabling of the side effects var](https://github.com/noir-lang/noir/blob/ea71a0e769fc691053d430eb854f445fd5e03950/compiler/noirc_evaluator/src/acir/mod.rs#L499) in ACIR gen is also based upon `requires_acir_gen_predicate`. 

## Additional Context

The snapshosts make the PR seem much bigger than it is in reality. The majority of the changes are contained within `compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs`. 

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
> (i.e. why does the predicate attached to the other array_get not handle this?)

The predicate of the other `array_get` is going to be one.
Taking this snippet:
```
    v19 = array_get v13, index u32 0 -> u1
    enable_side_effects v19
    v20 = array_get v9, index u32 0 -> u1
    v21 = not v19
    enable_side_effects v21
    v22 = array_get v8, index u32 0 -> u16
    v23 = cast v22 as u32
    v25 = mod v23, u32 3
    v26 = lt v25, u32 3
    v27 = unchecked_mul v26, v21
    constrain v27 == v21, "Index out of bounds"
    v28 = array_get v8, index v25 -> u16
```
We can see that the second array get occurs under the `v21` side effects var. This is the conditional of the if statement in the snippet. We then have the result of the second array get `v22` used in `v23 = cast v22 as u32` to `v25 = mod v23, u32 3` to `v28 = array_get v8, index v25 -> u16`. Following the SSA further we will see that `v28` is then used as part of the computing the final result we return from flattening. Using the earlier array get in fact gives a different index for `v25`.

compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs
```diff
@@ -216,34 +141,81 @@ mod test {
...
         acir(inline) fn main f0 {
-          b0(v0: Field, v1: Field):
-            enable_side_effects v0
-            v3 = allocate -> &mut Field
-            enable_side_effects v0
-            v4 = allocate -> &mut Field
+          b0(v0: [u16; 3], v1: u32, v2: u32):
             enable_side_effects v1
-            v5 = allocate -> &mut Field
+            v3 = array_get v0, index v1 -> u16
```
vezenovm, line 146: Switching to an actual acir predicate side effectual instruction

compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs
```diff
@@ -102,79 +100,6 @@ impl Function {
...
-            BinaryOp::Add { .. } | BinaryOp::Sub { .. } | BinaryOp::Mul { .. } => {
-                dfg.type_of_value(binary.lhs).is_unsigned()
-            }
-            BinaryOp::Div | BinaryOp::Mod => {
-                if let Some(rhs) = dfg.get_numeric_constant(binary.rhs) {
-                    rhs == FieldElement::zero()
-                } else {
-                    true
-                }
-            }
```
TomAFrench, line 118: Looks like we're losing this optimization if we rely on `requires_acir_gen_predicate`, can you move this across to that method?

compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs
```diff
@@ -102,79 +100,6 @@ impl Function {
...
-
-        // Some `Intrinsic`s have side effects so we must check what kind of `Call` this is.
-        Call { func, .. } => match dfg[*func] {
-            Value::Intrinsic(intrinsic) => match intrinsic {
-                Intrinsic::SlicePushBack
-                | Intrinsic::SlicePushFront
-                | Intrinsic::SlicePopBack
-                | Intrinsic::SlicePopFront
-                | Intrinsic::SliceInsert
-                | Intrinsic::SliceRemove => true,
```
TomAFrench, line 149: This is changing whether a lot of intrinsics are treated by this pass. Let's have tests for these.

TomAFrench (commented):
I found https://github.com/noir-lang/noir/issues/8387 while reviewing this.

I'm supportive of consolidating onto `requires_acir_gen_predicate` if we find it fits our needs here although I'm quite leery of the number of instructions this change is affecting.

Ideally we should have tests which show that the new behaviour is desirable for each of these instructions.

compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs
vezenovm: This is the optimization I would have simply removed if I had kept the `responds_to_side_effects_var` method. 

As per the PR description:
> After the EnableSideEffectsIf removal pass the modulo `v24 = mod v16, u32 3` occurs under a different predicate from where it was previously. If we have a zero predicate our [euclidean division returns zero](https://github.com/noir-lang/noir/blob/76c9937918ca59fa2b36abc5e106307d362ea338/compiler/noirc_evaluator/src/acir/acir_context/mod.rs#L772) for both the quotient and the remainder.

Even if the rhs is not zero, the modulo operation depends upon the predicate. Moving this over would cause the new `keep_enable_side_effects_for_safe_modulo` included in this PR to fail.

TomAFrench: I'd say that this points to us not correctly performing ACIRgen for constant rhs modulos properly rather than a constant modulo not being safe to perform without a predicate though?

TomAFrench: Yes, the predicate applied is 1 if `requires_acir_gen_predicate` returns false so if we fix this in that method then ACIRgen will properly perform a constant division without a predicate.

vezenovm: This is the final SSA with the side effects vars simplified:
```
  b0(v8: [u16; 3], v9: [u1; 1]):
    v11 = array_get v8, index u32 0 -> u16
    v12 = cast v11 as u32
    enable_side_effects u1 0
    v14 = array_get v9, index u32 0 -> u1
    v16 = call f1(v8, v14) -> [u1; 1]
    v18 = mod v12, u32 3
    v19 = lt v18, u32 3
    enable_side_effects u1 1
    constrain v19 == u1 1, "Index out of bounds"
    v21 = array_get v8, index v18 -> u16
    v22 = cast v21 as u32
    v24 = lt u32 24993, v22
    enable_side_effects u1 1
    return v24
}
```
If `requires_acir_gen_predicate` returns false, `v18 = mod v12, u32 3` uses a predicate of zero. Where do you see the predicate applied being `1` if `requires_acir_gen_predicate` returns false?

vezenovm: I'm seeing behavior slightly different than I had before... need a moment to reproduce. I may be running something incorrectly. I'm going to try bringing back the optimization.

vezenovm: Marking Div/Mod as not requiring a predicate when `rhs.is_zero()` causes failures in `regression_6834`. You can see https://github.com/noir-lang/noir/pull/7847#issuecomment-2762154020 for more information where `requires_acir_gen_predicate` for Div/Mod was set to always return true. 

> Yes, the predicate applied is 1 if `requires_acir_gen_predicate` returns false

I see what you were referencing here now. [Disabling of the side effects var](https://github.com/noir-lang/noir/blob/ea71a0e769fc691053d430eb854f445fd5e03950/compiler/noirc_evaluator/src/acir/mod.rs#L499) is based upon `requires_acir_gen_predicate`.  This allowed the new `regression_8236` in this PR to pass with the Div/Mod rhs optimization put back in place as the predicate was being overwritten to a constant of one.

TomAFrench: Ah yeah, sorry I wasn't explicit. Looks like you've made some changes locally so I'll wait until those are pushed to talk more on this.

vezenovm: > I can also trigger a failure under a predicate for modulo in `regression_6834` by replacing `(year - 2) / 4` with `(year - 2) % 5` (using `4` here would simplify the modulo to truncate).

Captured this difference in functionality here https://github.com/noir-lang/noir/issues/8402

compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs
TomAFrench: Ah, I kinda meant that we should have tests to check that if we place these under an if-statement then the ACIRgen is correct. e.g. no popping

vezenovm: Currently blocked by https://github.com/noir-lang/noir/issues/8408. I attempted to workaround #8408 with the SSA gen test utilities I added under `ssa_gen::tests`, however, these units tests do not have access to the stdlib so this causes separate panics for unresolved method calls when working with slice intrinsics. I will look at updating the SSA parser tomorrow (or if @asterite has a moment) to unblock these tests. I basically was trying to assert that given an SSA like such:
```
  b0(v0: u32, v1: u1):
    v4 = make_array [Field 2, Field 3] : [Field; 2]
    v5 = make_array [Field 1, v4] : [Field, [Field; 2]]
    v7 = array_set v5, index v0, value Field 4
    enable_side_effects v1
    v10, v11 = call slice_insert(u32 3, v7, u32 1, Field 1, v4) -> (u32, [Field, [Field; 2]])
    return v10
}
```
If we were to remove `enable_side_effects v1` we would in fact be changing the program and have a different ACIR. While for the unaffected slice intrinsics we would assert they are unchanged. 

I can still add tests asserting that the ACIR gen for all the slice intrinsics using non-composite slices are as expected, however, testing whether certain slice intrinsics (insert and remove) depend upon a predicate will remain untested as part of ACIR gen. These tests could also be added in a follow-up as they are more testing our marking of instructions during `requires_acir_gen_predicate`. I could also add some more complex integration tests for insert and remove to `execution_success`, but the predicated nature of `get_flattened_index` has already been tested implicitly by our various composite dynamic array tests.

vezenovm: I have captured these tests in a separate PR https://github.com/noir-lang/noir/pull/8421 @TomAFrench so they can be more easily reviewed and we can agree on the approach there separate from this PR.

TomAFrench (approved):

jfecher (approved):
LGTM. Fewer functions that do the same thing is good

PR #8357, opened 2025-05-05
fix: unify type from elaborator and from comptime value

asterite:
# Description

## Problem

Resolves https://github.com/noir-lang/noir/issues/8340

## Summary

Yet another approach. The problem with the two previous approaches were:
1. When only taking the type of the elaborator we were discarding the type of the comptime value. Then it seems somewhere along the road there's a type mismatch (I didn't debug this, to be honest)
2. Turning `Value::I32` into "1 as i32" works, but for "Value::Field" it's a bit trickier. If we have `[1, 2, 3]`, each of those will end up being `Value::Field`, but if the target type is `[i32; 3]` then we can't turn `[1, 2, 3]` into `[1 as Field, 2 as Field, 3 as Field]` as that doesn't compile.

In this PR we just unify the type we got from the Elaborator with the type we get from the comptime value... which, if it holds integers, should be a polymorphic integer (unless casts are used, etc.). Then the "untyped" literals will work fine.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/expressions.rs
+++ b/compiler/noirc_frontend/src/elaborator/expressions.rs
@@ -1241,13 +1241,19 @@ impl Elaborator<'_> {
         location: Location,
         target_type: Option<&Type>,
     ) -> (ExprId, Type) {
-        let (block, _typ) = self.elaborate_in_comptime_context(|this| {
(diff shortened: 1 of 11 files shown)


asterite:
Hmm... Maybe this isn't the right solution either. I'll leave this issue for later.

Issue #8281, opened 2025-04-29
nargo expand follow-up: make ignored tests work


asterite:
In `tooling/nargo_cli/build.rs` there are constants with names of test_programs that are ignored by the `nargo expand` tests. Most of these are ignored because of `nargo expand` bugs. The task here is to make `nargo expand` work well in those cases.

PR #8374, opened 2025-05-06
fix(expand): use re-exports for non-visibile items

asterite:
# Description

## Problem

Part of #8281

## Summary

Though it didn't fix any existing test, `nargo expand` did produce invalid code by using private items. These just produce a warning now but eventually they'll be an error.

You can see what changed in this PR by looking at how the snapshots changed.

## Documentation

Check one:

# PR Checklist


TomAFrench:
Not sure if there's an issue for this but I noticed in the snapshots a few imports like `use std::prelude::Vec;`. We should be able to just omit any imports from the `prelude` directory.

asterite:
Good point.

There are programs in "test_programs" that have imports like this:

```noir
use std::collections::vec::Vec;
```

`nargo expand` tries to preserve those `use` to show the expanded program as close to the original program. In this case that import is preserved, except that it's not directly visible so the visible re-export is used. But it's true that `nargo expand` could check that this `use` item refers to an item that exists in the prelude, and directly omit it.

In real programs I guess nobody would import these types... but I'll capture an issue anyway.

tooling/nargo_cli/src/cli/expand_cmd/printer.rs
```diff
@@ -25,22 +26,24 @@ use super::items::{Impl, Import, Item, Module, TraitImpl};
 mod hir;
 mod types;
 
-pub(super) struct ItemPrinter<'interner, 'def_map, 'string> {
+pub(super) struct ItemPrinter<'interner, 'def_map, 'deps, 'string> {
     crate_id: CrateId,
     interner: &'interner NodeInterner,
     def_maps: &'def_map DefMaps,
+    dependencies: &'deps Vec<Dependency>,
     string: &'string mut String,
```
TomAFrench, line 34: I feel like we're creating way more lifetimes than necessary here.

compiler/noirc_frontend/src/modules.rs
```diff
@@ -209,3 +214,54 @@ pub fn module_def_id_relative_path(
 
     Some(path)
 }
+
+/// Returns true if the given ModuleDefId is visible from the current module, given its visibility.
+/// This will in turn check if the ModuleDefId parent modules are visible from the current module.
```
TomAFrench, line 219:
```suggestion
///
/// This will in turn check if the ModuleDefId parent modules are visible from the current module.
```

TomAFrench (approved):

PR #8376, opened 2025-05-06
chore: move acirgen tests into separate file

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

I'd like to slim down `acir/mod.rs` as it's a few thousand lines at this point. To help with this I've moved a bunch of the test code into a new `tests` directory similar to how we do it for `noirc_frontend`.

I've also added more usage of the SSA parser in these tests.

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
Looks good! I'm curious about why changing it to use the SSA parser didn't work (I tried it locally too and it failed), there must be a difference between how the SSA is constructed manually and what the SSA parser does. I'll investigate it after this merges.

TomAFrench:
The issue is because we're injecting callstacks when using the builder but we can't do the same when using the SSA parser.  We could perhaps make the SSA parser inject similar callstack information to get around this but I think it's not worth it atm.

Issue #8352, opened 2025-05-05
Parse test function name without splitting for `debugger` tests


michaeljklein:
### Problem

Currently, the `test_name`'s of test functions, collected by the `debugger`'s `build.rs` uses `.split(..)` and `.unwrap()` to parse the name of the function:

```rust
            // TODO: get test name by regex perhaps?
            let test_name = test_name_line
                .split("fn ")
                .collect::<Vec<&str>>()
                .get(1)
                .unwrap()
                .split("<")
                .next()
                .unwrap()
                .split("(")
                .next()
                .unwrap();
```

### Happy Case

Use regex or another more flexible form of parsing

### Workaround

None

### Project Impact

Nice-to-have

### Would you like to submit a PR for this Issue?

None

PR #8381, opened 2025-05-06
fix: Use regex for test function name parsing in debugger build script

GarmashAlex:
# Description

This PR replaces the multiple .split() and .unwrap() calls used for parsing test function names in the debugger build script with a regex-based approach. The new implementation:

1. Adds a regex dependency to tooling/debugger/Cargo.toml
2. Uses a regex pattern to extract function names that handles generics and arguments properly
3. Maintains a fallback to the old method for backward compatibility

This makes the test function name parsing more robust and maintainable, while avoiding potential panic situations when parsing complex function signatures.

## Problem\*

Fix  issue #8352 

## Additional Context

The previous implementation was using a series of string splits and unwraps which could fail when dealing with more complex function signatures. The regex approach provides a cleaner and more reliable solution.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/debugger/build.rs
+++ b/tooling/debugger/build.rs
@@ -1,3 +1,4 @@
+use regex::Regex;
 use std::collections::HashSet;
 use std::fs::File;
 use std::io::{BufRead, BufReader, Write};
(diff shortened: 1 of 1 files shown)


TomAFrench:
We have a policy of not accepting PRs which consist entirely of LLM output due to the fact that it ends up functionally being a DOS on maintainers because we then need to do all the checking which would otherwise be done by the author (as shown by this PR keeping the old method as a fallback).

I'm sorry but I'm going to close this PR.

Issue #8344, opened 2025-05-05
Comptime & SSA differ in evaluating assignment with side-effects and array in lvalue


jfecher:
### Aim

```noir
fn main() {
    bug();
    comptime { bug() };
}

fn bug() {
    let mut a = ([1, 2], 3);
    a.0[{
        a = ([4,5], 6);
        1
    }] = 7;

    // acir+brillig yield ([1, 7], 3)
    // comptime yields ([1, 7], 6)
    // both should yield:
    assert_eq(a, ([4, 7], 6));
}
```

### Expected Behavior

both should yield: `([4, 7], 6))`

### Bug

acir+brillig yield `([1, 7], 3)`
comptime yields `([1, 7], 6)`

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

vezenovm:
Cross-posting a proposed solution for visibility https://github.com/noir-lang/noir/issues/8262#issuecomment-2851205648

PR #8384, opened 2025-05-06
fix: Fix sequencing of side-effects in lvalue

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/8344

## Summary\*

I removed statements from the index part of an lvalue to sequence the side-effects consistently. This is done while converting the Ast to the Hir so the change is seen by both the comptime interpreter and SSA.

## Additional Context

This changes output of `nargo expand` so I added a quick check to skip it if the expression trivially contains no side-effects (is already a variable or literal) to minimize changes but there are still several.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/elaborator/statements.rs
```diff
@@ -354,7 +370,13 @@ impl Elaborator<'_> {
         }
     }
 
-    fn elaborate_lvalue(&mut self, lvalue: LValue) -> (HirLValue, Type, bool) {
+    /// Elaborates an lvalue returning:
+    /// - The HirLValue equivalent of the given `lvalue`
+    /// - The type being assigned to
+    /// - Whether the underlying variable is mutable
+    /// - A vector of new statements which need to prefix the resulting assign statement.
+    ///   This hoists out any sub-expressions to simpilfy sequencing of side-effects.
```
vezenovm, line 378:
```suggestion
    ///   This hoists out any sub-expressions to simplify sequencing of side-effects.
```

vezenovm (approved):

jfecher:
Looks like
```
tests::nargo_expand_execution_success::test_aes128_encrypt
tests::nargo_expand_execution_success::test_encrypted_log_regression
tests::nargo_expand_execution_success::test_nested_array_dynamic
```
are failing - I wasn't aware we tested expanded code to ensure it runs but it's changing now because I'm creating a new variable `i` for each index de-sugared in this way. Internally there's no confusion since they have different `DefinitionId`s but nargo expand drops the `$id` suffix so it results in code that uses `i`, `i`, `i` where each i is meant to be different as the same i, leading to an assertion failure.

asterite:
We could ignore those tests for now. In the long run, I'm not sure how `nargo expand` can work with these introduced variables.

asterite:
Maybe the compiler could have a counter for temporary variables, then use "__var{index}" for these. During compilation it wouldn't change anything, but at least for `nargo expand` it would mean there's less chance of collisions.

jfecher:
> Maybe the compiler could have a counter for temporary variables, then use "__var{index}" for these. During compilation it wouldn't change anything, but at least for nargo expand it would mean there's less chance of collisions.

If we just want to reduce the chance of collisions we could add `_{definition_id}` as a suffix since that is already effectively a counter. I'll add that for this PR but there is still a chance of collision which is unsatisfactory.

PR #8395, opened 2025-05-07
fix: error on unused generic in trait impl

### LINKED ISSUE #8394: Unused generics do not error on trait impls
### Aim

Given a code like this:

```noir
trait Trait {}

impl<T> Trait for i32 {}
```

### Expected Behavior

The compiler should error on `T`, saying it's not mentioned in either the trait or the object type, similar to what was done in https://github.com/noir-lang/noir/issues/5855 for regular impls.

### Bug

The compiler doesn't error.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8394

## Summary

The only problem now is that `HashMap` and `UHashMap` fail to compile because they have unused generics, but... I fail to see how this can be fixed. I feel it's a bit strange that `HashMap` has a `B` generic which is `BuildHasher<H>`, but that `H` isn't mentioned anywhere... it's like that `H` appears from thin air. So maybe `HashMap` should have an `H` generic, but I'm not sure how... maybe `BuildHasher<H>` shouldn't exist and instead `_build_hasher` should be `_build_hasher: fn() -> H` or something like that, not sure... 🤔 

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
> Asking for a review because I'm not sure how to fix HashMap and UHashMap in the standard library so this error doesn't trigger... but maybe one of you know!

Rust's BuildHasher uses an associated type to avoid this: https://doc.rust-lang.org/std/hash/trait.BuildHasher.html. So we should do the same since we don't want to allow multiple impls of BuildHasher for the same input type either.

asterite:
Cool, thank you! I'll try to do that.

asterite:
Hmm, I think it still won't work. In Rust it's something like this:

```rust
pub trait BuildHasher {
    type H: Hasher;
}
```

so we know the type `H` is a `Hasher`. Right now in Noir that doesn't work but we can do this:

```noir
pub trait BuildHasher {
    type H;

    fn build_hasher(self) -> H
    where
        H: Hasher;
}
```

but it's not exactly the same. Whenever we have something of type `B` we have to constrain that to `B: BuildHasher<H>, H: Hasher` over and over again, but sometimes we can't introduce that type `H`, for example in a trait implementation where we can't add new where clauses.

Is this something that should be implemented for Noir 1.0? Otherwise once we do this it's going to be a large breaking change.

TomAFrench:
I updated the bigcurve dep on bignum

asterite:
Ah! Mystery solved 😄 Thanks!

asterite:
This is ready for review now.

TomAFrench (approved):

PR #8412, opened 2025-05-08
fix: print slice composite types surrounded by parentheses

### LINKED ISSUE #8408: SSA parser cannot handle composite slice types
### Aim

I want to pass a string like the following to the SSA parser:
```
      b0(v0: u32, v1: u1):
        v4 = make_array [Field 2, Field 3] : [Field; 2]
        v5 = make_array [Field 1, v4] : [Field, [Field; 2]]
        v7 = array_set v5, index v0, value Field 4
        enable_side_effects v1
        v10, v11 = call slice_insert(u32 3, v7, u32 1, Field 1, v4) -> (u32, [Field, [Field; 2]])
        return v10
    }
``` 

### Expected Behavior

I expect to be able to generate an SSA object for the above snippet using the parser.

### Bug

I get this panic:
```
---- acir::tests::slice_insert_affected_by_predicate stdout ----

thread 'acir::tests::slice_insert_affected_by_predicate' panicked at compiler/noirc_evaluator/src/acir/tests/mod.rs:987:58:
called `Result::unwrap()` on an `Err` value: 
      b0(v0: u32, v1: u1):
        v4 = make_array [Field 2, Field 3] : [Field; 2]

        v5 = make_array [Field 1, v4] : [Field, [Field; 2]]
                                              ^
                                              Expected ']', found ','

        v7 = array_set v5, index v0, value Field 4
        enable_side_effects v1
        v10, v11 = call slice_insert(u32 3, v7, u32 1, Field 1, v4) -> (u32, [Field, [Field; 2]])
        return v10
    }
```

### To Reproduce

1. Make an SSA unit test and create a string literal using the snippet above.
2. Pass the SSA string to `Ssa::from_str(src_with_pred).unwrap()`
3. Run

### Workaround

Yes

### Workaround Description

1. Manually build the SSA using the function builder as was done pre SSA parser
2. I am could use the Noir code directly to get my SSA rather than the SSA parser

### Additional Context

_No response_

### Project Impact

Blocker

### Blocker Context

Blocks the ability to add clean ACIR gen unit tests for whether certain slice intrinsics are dependent upon a predicate in ACIR gen in https://github.com/noir-lang/noir/pull/8355. Predicates are only used when we have slices of composite elements. We otherwise have to build the SSA manually. 

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+ea71a0e769fc691053d430eb854f445fd5e03950 (git version hash: ea71a0e769fc691053d430eb854f445fd5e03950, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8408

## Summary

Let me know if composite types shouldn't be surrounded by parentheses. It's something I already did for arrays a long time ago (when writing the SSA parser, or maybe even before that).

## Additional Context

I can't remember why I added those parentheses, it was maybe for clarity or to avoid an ambiguity.

Then I just tried to generate SSA for this code:

```noir
fn main() {
    let _ = [(); 4];
}
```

The result is:

```noir
acir(inline) fn main f0 {
  b0():
    v0 = make_array [] : [(); 4]
    return
}
```

I guess without those `()` it would look like `[; 4]` so we'd have to assume, when parsing, that if no type comes it's an empty list... but maybe that makes it harder to distinguish between that and a mistake.

I think back when I implemented the SSA parser I wasn't sure what all types were, so when parsing parameters `b0: Field, b1: Field`, if we allowed composite types in the parameters it would be harder to parse because we see `Field, ...` and we don't know if another type comes. Well, composite types aren't allowed in parameters, but in the end using parentheses is maybe clearer and more consistent.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/types.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/types.rs
@@ -315,7 +315,11 @@ impl std::fmt::Display for Type {
             }
             Type::Slice(element) => {
                 let elements = vecmap(element.iter(), |element| element.to_string());
-                write!(f, "[{}]", elements.join(", "))
+                if elements.len() == 1 {
+                    write!(f, "[{}]", elements.join(", "))
+                } else {
+                    write!(f, "[({})]", elements.join(", "))
+                }
             }
             Type::Function => write!(f, "function"),
         }
--- a/compiler/noirc_evaluator/src/ssa/parser/tests.rs
+++ b/compiler/noirc_evaluator/src/ssa/parser/tests.rs
@@ -90,6 +90,19 @@ fn test_make_composite_array() {
 }
 
 #[test]
+fn test_make_composite_slice() {
+    let src = "
+        acir(inline) predicate_pure fn main f0 {
+          b0():
+            v2 = make_array [Field 2, Field 3] : [Field; 2]
+            v4 = make_array [Field 1, v2] : [(Field, [Field; 2])]
+            return v4
+        }
+        ";
+    assert_ssa_roundtrip(src);
+}
+
+#[test]
 fn test_make_byte_array_with_string_literal() {
     let src = "
         acir(inline) fn main f0 {


vezenovm:
> if we allowed composite types in the parameters it would be harder to parse because we see `Field, ...` and we don't know if another type comes. Well, composite types aren't allowed in parameters, but in the end using parentheses is maybe clearer and more consistent.

We do allow composite types in parameters they just will be flattened to separate variables as you have outlined.

vezenovm (approved):
I find the parentheses clearer and it aligns with what we do for arrays. Thanks!

PR #8416, opened 2025-05-08
feat: Skip SSA passes that contain any of the given messages

aakoshh:
# Description

## Problem\*

Something I suggested in https://github.com/noir-lang/noir/pull/8407/files#r2079545908 instead of adding individual filters for passes.

## Summary\*

Adds a `--skip-ssa-pass` compiler option which is passed to the `SsaBuilder` to tell it to skip SSA passes we want to exclude from the pipeline, for testing purposes. 

Multiple passes can be skipped by repeating the options, e.g. `--skip-ssa-pass Inlining --skip-ssa-pass "Removing Unreachable Functions (2)"`

Note that the skipped passes are not printed at all, so the kind of numbering done by https://github.com/noir-lang/noir/pull/7853 would be disrupted by it. I thought an alternative to that would be to print the unchanged SSA with a "SKIPPED" label, but it might be confusing, and if I just print a message, it could disrupt the splitting in that PR if it expects that every step is parseable. The count in the message is maintained though, so we can filter e.g. the 2nd occurrence and still display the 3rd. 

## Additional Context

Example:
```console
❯ bash
bash-5.2$ PS1="$ "
$ cargo run -q -p nargo_cli -- compile --show-ssa-pass "at most one" 
After Inlining functions with at most one instruction (1):
acir(inline) fn main f0 {
  b0():
    call f1()
    return
}
brillig(inline) fn foo f1 {
  b0():
    call f2()
    return
}
brillig(inline) fn bar f2 {
  b0():
    call f2()
    return
}

warning: function `bar` cannot return without recursing
   ┌─ src/main.nr:12:18
   │
12 │ unconstrained fn bar() {
   │                  --- function cannot return without recursing
   │

$ cargo run -q -p nargo_cli -- compile --show-ssa-pass "at most one" --skip-ssa-pass "at most one" 
warning: function `bar` cannot return without recursing
   ┌─ src/main.nr:12:18
   │
12 │ unconstrained fn bar() {
   │                  --- function cannot return without recursing
   │


```

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
Looks good!

It wasn't introduced in this PR but I initially didn't understand what the `passed` field of `SsaBuilder` was about. Now I understand it's for the counter, but a small comment on top of it might be useful.

aakoshh:
Thanks @asterite I'll add an explanation 👍

PR #8417, opened 2025-05-08
fix: general solution for accessing associated constants

### LINKED ISSUE #7812: Can't access numeric associated value in method
### Aim

This code:

```noir
trait Trait {
    let N: u32;

    fn foo() -> u32 {
        Self::N
    }
}

impl Trait for i32 {
    let N: u32 = 10;
}

fn main() {
    println(i32::foo());
}
```

and this code:

```noir
trait Trait {
    let N: u32;
}

impl Trait for i32 {
    let N: u32 = 10;
}

fn foo<T>() -> u32
where
    T: Trait,
{
    T::N
}

fn main() {
    println(foo::<i32>())
}
```

### Expected Behavior

Should compile and print 10.

### Bug

It errors with this:

```
error: Could not resolve 'Self' in path
  ┌─ src/main.nr:5:9
  │
5 │         Self::N
  │         ----
```

and this:

```noir
error: Could not resolve 'T' in path
   ┌─ src/main.nr:17:5
   │
17 │     T::N
   │     -
```

### To Reproduce

1.
2.
3.
4.

### Workaround

There's a workaround for the first snippet:

```noir
trait Trait {
    let N: u32;

    fn foo() -> u32 {
        [0; Self::N].len() // <-- here
    }
}

impl Trait for i32 {
    let N: u32 = 10;
}

fn main() {
    println(i32::foo());
}
```

### Workaround Description

Because associated constants seem to resolve fine in type positions, we can create an array of that length and ask for its length, which is equivalent to directly getting the length.

### Additional Context

It also doesn't work with associated types. All of these do work at the type level (for example in function signatures).

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Follow-up to #8403
Fixes #7812
Fixes #7812

## Summary

1. For each associated constant we create a definition (`DefinitionId`) with a new kind, `DefinitionKind::AssociatedConstant`, that holds a trait impl ID and the associated constant name
2. When looking up an associated constant, we know it's numeric type. We can also get a `DefinitionId` for it, so we elaborate it with those things.
3. Eventually during monomorphization and in the comptime interpreter, when we come across this new `DefinitionKind` we can look up the associated constant type and try to evaluate it to a number.

## Documentation

Check one:

# PR Checklist


TomAFrench:
Looks like this requirement is going to be a requirement for #8545.

asterite:
Force-pushed because there was a bad merge along the way.

This still doesn't work as when I wrote it I was just trying to see if an approach like this would work.

compiler/noirc_frontend/src/monomorphization/mod.rs
```diff
@@ -1078,6 +1078,24 @@ impl<'interner> Monomorphizer<'interner> {
                 let value = SignedField::positive(value);
                 ast::Expression::Literal(ast::Literal::Integer(value, typ, location))
             }
+            DefinitionKind::AssociatedConstant(trait_impl_id, name) => {
```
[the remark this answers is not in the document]
asterite: I see that `take_unresolved_associated_types` returns `Vec<(Ident, UnresolvedType)>`, so these are unresolved types, so no type variables. Then `GenericTypeArgs` has a `named_args` field, but it's also `Vec<(Ident, UnresolvedType)>`. Then associated types seem to always be a list of `NamedType` which holds a `Type` but, as in the example above, it's not always a `TypeVariable`.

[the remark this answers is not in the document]
jfecher: Ah, my bad. Looks like they're resolved in `resolve_trait_args_from_trait_impl` which doesn't wrap them in type variables.
We don't do that for regular numeric generics either though.

You could try taking the generic from the parent trait. That should be what is bound to the named generic argument later. https://github.com/noir-lang/noir/blob/master/compiler/noirc_frontend/src/elaborator/traits.rs#L50-L53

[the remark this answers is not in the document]
asterite: I tried wrapping the type inside `NamedType` inside a type variable and that almost worked but no. 

> You could try taking the generic from the parent trait.

I'm not sure that can work. We need the type associated to the trait impl associated constant, which is the one that will eventually evaluate to a constant, and it's the constant that we need (not the generic type).

I went back to using `DefinitionKind::AssociatedConstant`, together with creating a `DefinitionId` for each associated constant, and it seems to work. But I'm sure there's a better way to do it.

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
```diff
@@ -573,10 +573,12 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
     }
 
     pub(super) fn evaluate_ident(&mut self, ident: HirIdent, id: ExprId) -> IResult<Value> {
-        let definition = self.elaborator.interner.try_definition(ident.id).ok_or_else(|| {
-            let location = self.elaborator.interner.expr_location(&id);
-            InterpreterError::VariableNotInScope { location }
-        })?;
+        let location = self.elaborator.interner.expr_location(&id);
```
michaeljklein, line 576: It looks like this `location` only needs to be looked-up in the `ok_or_else` case below?

compiler/noirc_frontend/src/monomorphization/mod.rs
```diff
@@ -1079,6 +1079,34 @@ impl<'interner> Monomorphizer<'interner> {
...
+                let Kind::Numeric(numeric_type) = associated_type.typ.kind() else {
+                    unreachable!("Expected associated type to be numeric");
+                };
+                match associated_type
+                    .typ
+                    .evaluate_to_field_element(&associated_type.typ.kind(), location)
+                {
+                    Ok(value) => {
+                        let typ = Self::convert_type(&numeric_type, location)?;
+                        let value = SignedField::positive(value);
```
michaeljklein, line 1098: Will this always be a `SignedField::positive`? I would expect this to work with signed and negative associated constants

tooling/nargo_cli/build.rs
```diff
@@ -133,6 +133,8 @@ const IGNORED_NARGO_EXPAND_EXECUTION_TESTS: [&str; 8] = [
     "regression_5045",
     // bug
     "regression_7744",
+    // bug
+    "trait_associated_constant",
```
michaeljklein, line 137: Does this need a follow-up?

michaeljklein (changes requested):

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
asterite: It's also used on lines 672 and another one. But I'll push a commit to only fetch it when needed.

compiler/noirc_frontend/src/monomorphization/mod.rs
asterite: Right, it's something I noticed too. Associated constants can't hold negative values right now. I thought about capturing an issue about it right now, but I don't know if it's intentional. This code:

```noir
trait Foo {
    let N: i32;
}

impl Foo for i32 {
    let N: i32 = -1;
}
```

gives this error:

```
error: Invalid type expression: '-1'. Only unsigned integer constants up to `u32`, globals, generics, +, -, *, /, and % may be used in this context.
  ┌─ src/main.nr:6:18
  │
6 │     let N: i32 = -1;
  │                  --
```

On the other hand this code gives a different error:

```noir
trait Foo {
    let N: i32;

    fn n() -> i32 {
        Self::N
    }
}

impl Foo for i32 {
    let N: i32 = 1 - 10;
}

fn main() {
    println(i32::n());
}
```

It says:

```
error: Could not determine the value of associated constant `N`, encountered error: `The value `-9` cannot fit into `numeric i32` which has a maximum size of `2147483647``
  ┌─ src/main.nr:5:9
  │
5 │         Self::N
  │         -------
```

which is a bit misleading. It **could** determine the value, and it actually fits the range of `i32`.

I captured https://github.com/noir-lang/noir/issues/8649 for this. I'm not sure I should leave a TODO for this in the code because it would affect many places in the code.

tooling/nargo_cli/build.rs
asterite: It's this one: https://github.com/noir-lang/noir/issues/8281

I add to the ignore list instead of fixing new bugs right away because I think `nargo expand` has less priority than everything else.

jfecher (approved):

Issue #8218, opened 2025-04-24
Canary audit selected parts of the SSA internally


Savio-Sou:
Audit selected parts of the SSA ourselves to practice auditing and improve security of our code.

### Scope

A subset of the SSA passes:

1. `remove_unreachable_functions`
https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/remove_unreachable.rs
2. `defunctionalize`
https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
3. `inline_simple_functions`
https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/inline_simple_functions.rs

As part of the above, the "general machinery" of SSA, e.g. the DFG https://github.com/noir-lang/noir/tree/master/compiler/noirc_evaluator/src/ssa/ir is also in scope.

### Out of scope

- Any SSA codegen from AST
  - Arbitrary test SSA can be constructed as part of this audit (subject to any documented restrictions on the inputs on the above SSA passes)
- All other SSA passes
- ACIR/Brillig gen (The SSA interpreter may be used in order to test execution directly)

### Timeline

- Audit begins: 6 May 2025
- Audit ends: 16 May 2025

### Resources

- Noir Documentation: https://noir-lang.org/docs/
- SSA Documentation: https://noir-lang.github.io/noir/docs/noirc_evaluator/ssa/index.html

PR #8393, opened 2025-05-07
feat: SSA pass impact report

aakoshh:
# Description

## Problem\*

Part of https://github.com/noir-lang/noir/issues/8218

Looking at any given SSA pass, they have a few unit tests, but it's not obvious which integration tests (ie. Noir projects under `test_programs`) are relevant to them.

## Summary\*

Added a tool to list the integration tests where a certain SSA pass has the most impact on the IR. 

It can be executed as follows:
```shell
cargo run -q -p nargo_cli --example ssa_pass_impact -- --ssa-pass <ssa-pass-name>
```

Currently it only looks at `execution_success`.

### Impact

The _impact_ is based on the _similarity_ between the SSA of the pass we're interested in, and the one before it. A pass can appear multiple times in the pipeline, and these are all considered separately.

The _similarity_ is calculated based on the number of lines where the string representation of the two SSA equal each other, which is computed using the `similar` crate. Before comparing them, the numeric values from the variable, block and function IDs are stripped away, so a re-numbering resulting from removing or adding instructions is ignored, and only the general structure and instruction names count.

The formulas are:
```
similarity(a, b) = 2 * lines_equal(a, b) / (lines(a) + lines(b))
impact(a, b) = 1 - similarity(a, b)
```

### Build pipeline

I ended up duplicating a bunch of boilerplate from the `compile_cli` and `nargo` crates, and bypassing the `SsaBuilder` to run passes one by one. This has some pros and cons:

Pros:
* I can stop the compilation as soon as we have passed the last occurrence of the pass we're interested in
* There is no need to compile into Brillig and ACIR

Cons:
* The code does not do everything the main compiler does, and can get out of sync
* I don't execute the secondary pipeline with the Brillig VM passes
* Every execution of the tool compiles every program

These could be alleviated if instead of compiling in the tool we relied on what Tom started in https://github.com/noir-lang/noir/pull/7853 and looked for the printed SSA saved into files, just doing the string comparisons.

### Audited passes

The following is the impact on tests of the passes selected for audit:

#### Remove unreachable functions

```console
❯ cargo run -q -p nargo_cli --example ssa_pass_impact -- --ssa-pass "Removing Unreachable Functions"
Packages: 267
Passes total: 1068
Passes with no impact: 1063
Passes most impacted by 'Removing Unreachable Functions' (top 50):
	0.156 impact: step 35 following 'Brillig Entry Point Analysis' in regression_fuzz_testing
[...]

PR #8418, opened 2025-05-08
chore(test): Add more unit tests for `inline_functions_with_at_most_one_instruction`

aakoshh:
# Description

## Problem\*

Part of https://github.com/noir-lang/noir/issues/8218

## Summary\*

Added some unit tests to cover the conditions in `should_inline_call` by first commenting out one of the `if` and then check if there is a unit test that fails.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/.inline_functions_with_at_most_one_instruction.rs.pending-snap
```diff
@@ -0,0 +1,9 @@
+{"run_id":"1746712579-849205000","line":140,"new":{"module_name":"noirc_evaluator__ssa__opt__inline_functions_with_at_most_one_instruction__test","snapshot_name":"does_not_inline_functions_that_require_multiple_passes","metadata":{"source":"compiler/noirc_evaluator/src/ssa/opt/inline_functions_with_at_most_one_instruction.rs","assertion_line":140,"expression":"mut_ssa"},"snapshot":"acir(inline) fn main f0 {\n  b0(v0: Field):\n    v2 = call f2(v0) -> Field\n    return v2\n}\nacir(inline) fn foo f1 {\n  b0(v0: Field):\n    return v0\n}\nacir(inline) fn bar f2 {\n  b0(v0: Field):\n    v1 = add v0, v0\n    return v1\n}"},"old":{"module_name":"noirc_evaluator__ssa__opt__inline_functions_with_at_most_one_instruction__test","metadata":{},"snapshot":"acir(inline) fn main f0 {\nb0(v0: Field):\n    v2 = call f2(v0) -> Field\n    return v2\n}\nacir(inline) fn foo f1 {\nb0(v0: Field):\n    return v0\n}\nacir(inline) fn bar f2 {\nb0(v0: Field):\n    v1 = add v0, v0\n    return v1\n}"}}
```
asterite, line 1: I think this file should be removed

compiler/noirc_evaluator/src/ssa/opt/inline_functions_with_at_most_one_instruction.rs
```diff
@@ -25,18 +25,19 @@ impl Ssa {
...
                 return false;
             }
 
             // Only inline functions with 0 or 1 instructions
-            if entry_block.instructions().len() > 1 {
+            if instructions.len() > 1 {
                 return false;
             }
 
-            let instructions = callee.dfg[entry_block_id].instructions();
```
michaeljklein, line 39: Is this a fix, separate from the tests?

michaeljklein (changes requested):

compiler/noirc_evaluator/src/ssa/opt/inline_functions_with_at_most_one_instruction.rs
aakoshh: It's not fixing anything, just a small refactoring because these instructions have been accessed in a couple of ways.

compiler/noirc_evaluator/src/ssa/ir/map.rs
```diff
@@ -295,3 +295,12 @@ impl<T> Default for AtomicCounter<T> {
...
 }
+
+impl<T> Clone for AtomicCounter<T> {
+    fn clone(&self) -> Self {
+        Self {
+            next: AtomicU32::new(self.next.load(Ordering::Relaxed)),
+            _marker: std::marker::PhantomData,
+        }
+    }
+}
```
vezenovm, line 306:
```suggestion
```

compiler/noirc_evaluator/src/ssa/ssa_gen/program.rs
```diff
@@ -16,7 +16,7 @@ use super::ValueId;
 
 /// Contains the entire SSA representation of the program.
 #[serde_as]
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Clone)]
```
vezenovm, line 19:
```suggestion
#[derive(Serialize, Deserialize)]
```

compiler/noirc_evaluator/src/ssa/opt/inline_functions_with_at_most_one_instruction.rs
```diff
@@ -104,6 +111,69 @@ mod test {
...
+            v2 = call f1(v0) -> Field
+            v3 = add v1, v2
+            return v3
+        }
+        ";
+        let ssa = Ssa::from_str(src).unwrap();
+
+        // In the first pass it won't recognize that `main` could be simplified.
+        let ssa = ssa.inline_functions_with_at_most_one_instruction();
+        assert_ssa_snapshot!(ssa.clone(), @r"
```
vezenovm, line 140:
```suggestion
        // In the first pass it won't recognize that `main` could be simplified.
        let ssa = ssa.inline_functions_with_at_most_one_instruction();
        assert_ssa_snapshot!(ssa, @r"
```
I know we have to clone here with the way the test is written right now, but we should be able to avoid cloning the entire SSA.

compiler/noirc_evaluator/src/ssa/opt/inline_functions_with_at_most_one_instruction.rs
```diff
@@ -104,6 +111,69 @@ mod test {
...
+        }
+        acir(inline) fn bar f2 {
+          b0(v0: Field):
+            v1 = add v0, v0
+            return v1
+        }
+        ");
+
+        // After `bar` has been simplified, it does `main` as well.
+        let ssa = ssa.inline_functions_with_at_most_one_instruction();
```
vezenovm, line 158:
```suggestion
        let ssa = Ssa::from_str(src).unwrap();
        // After `bar` has been simplified, it does `main` as well.
        let ssa = ssa
            .inline_functions_with_at_most_one_instruction()
            .inline_functions_with_at_most_one_instruction();
```
let's just regenerate the SSA rather than cloning it

aakoshh: Okay. Curious though, what problem do you see in having a `Ssa::clone`? After all the SSA can be serialised and deserialised, it can be printed and parsed, all of which provide roundabout ways to clone it. Originally I wanted to capture the subsequent states of it in https://github.com/noir-lang/noir/pull/8393, but ended up just using the string representation. I'll remove this commit from that PR as well if it's considered harmful.

vezenovm: I'm not very against it being used in tests like how you did, but if we do not need to clone I think it is good to discourage cloning the entire Ssa structure. Thus, not deriving clone on `Ssa`. 
- Cloning a large Ssa can potentially be very expensive if it works its way into our normal compilation pipeline
- For our central Ssa data structure it is good to enforce ownership tracking. Cloning could let us avoid that friction more easily, but we don't really want that

> SSA can be serialised and deserialised, it can be printed and parsed, all of which provide roundabout ways to (partially) clone it

Yes, but as you said they are roundabout ways to perform a clone as it is not the actual desired functionality in each of those cases.

aakoshh: Okay, thanks 👍 
Found a way to avoid the duplication using `&mut` with the macro, hopefully best of both worlds.

vezenovm (approved):
Only `&mut` worked not `&`? Approving if that is the case, otherwise lets switch that before merging.

aakoshh:
> Only &mut worked not &? 

Unfortunately not, because the macro needs mutable reference or take ownership to be able to normalize the IDs:
```rust
let mut mut_ssa = $ssa;
mut_ssa.normalize_ids();
```

michaeljklein (approved):

PR #8420, opened 2025-05-08
fix!: accept and return visibilities for StructDef

### LINKED ISSUE #6229: `StructDef::fields` and `StructDef::set_fields` don't take visibility into account
### Aim

In #6221 we'll introduce struct field visibility. The existing macro methods for dealing with struct fields don't know about visibility, but they probably should.

### Expected Behavior

There should be a way to know a struct field's visibility, and to specify it when using `set_fields`.

### Bug

There's no way to deal with struct field visibilities in macros.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #6229

## Summary

This is a breaking change, but it should be easy to fix on the Aztec-Packages side.

Now that this is producing a compile error in Aztec-Packages, that error was produced during comptime code. There was an issue where in many cases we were pushing a scope, evaluating something with `?` and then popping the scope. The problem was that `?` returns earlier and we would end up with extra scopes... I think in the case of enter/exit function it lead to an empty array of scopes. I fixed that and put an assertion so that we never end up with an empty scope list.

## Additional Context

I thought about introducing a `Visibility` struct, similar to how we have `UnaryOp`, `BinaryOp`, etc., but given that in Rust you can have `pub(super)` and many other visibilities (I think it can target any module) I made the visibility be Quoted.

## Documentation

Check one:

# PR Checklist


asterite:
The diff to fix Aztec-Packages is relatively small:

```diff
diff --git a/noir-projects/aztec-nr/aztec/src/macros/notes.nr b/noir-projects/aztec-nr/aztec/src/macros/notes.nr
index 8d1e97e0a3..873e9ba80a 100644
--- a/noir-projects/aztec-nr/aztec/src/macros/notes.nr
+++ b/noir-projects/aztec-nr/aztec/src/macros/notes.nr
@@ -156,7 +156,7 @@ comptime fn generate_note_properties(s: TypeDefinition) -> Quoted {
     let note_fields = s.fields_as_written();

     let properties_types = note_fields
-        .map(|(name, _): (Quoted, Type)| quote { pub $name: $property_selector_type })
+        .map(|(name, _, _): (Quoted, Type, Quoted)| quote { pub $name: $property_selector_type })
         .join(quote {,});

     // TODO #8694: Properly handle non-field types https://github.com/AztecProtocol/aztec-packages/issues/8694
@@ -276,7 +276,7 @@ comptime fn index_note_fields(
     let mut indexed_nullable_fields = &[];
     let mut counter: u32 = 0;
     for field in s.fields_as_written() {
-        let (name, typ) = field;
+        let (name, typ, _visibility) = field;
         if nullable_fields.all(|field| field != name) {
             indexed_fixed_fields = indexed_fixed_fields.push_back((name, typ, counter));
         } else {
@@ -393,7 +393,7 @@ comptime fn assert_has_owner(note: TypeDefinition) {
     let fields = note.fields_as_written();
     let mut has_owner = false;
     for i in 0..fields.len() {
-        let (field_name, _) = fields[i];
+        let (field_name, _, _) = fields[i];
         if field_name == quote { owner } {
             has_owner = true;
             break;
diff --git a/noir-projects/aztec-nr/aztec/src/macros/storage.nr b/noir-projects/aztec-nr/aztec/src/macros/storage.nr
index a621e9c52e..99c07bc770 100644
--- a/noir-projects/aztec-nr/aztec/src/macros/storage.nr
+++ b/noir-projects/aztec-nr/aztec/src/macros/storage.nr
@@ -45,7 +45,7 @@ pub comptime fn storage(s: TypeDefinition) -> Quoted {
     //let context_generic = s.add_generic("Context");
     for field in s.fields_as_written() {
         // FIXME: This doesn't handle field types with generics
-        let (name, typ) = field;
+        let (name, typ, _visibility) = field;
         let (storage_field_constructor, storage_size) =
             generate_storage_field_constructor(typ, quote { $slot });
         storage_vars_constructors =
diff --git a/noir-projects/aztec-nr/aztec/src/macros/utils.nr b/noir-projects/aztec-nr/aztec/src/macros/utils.nr
index 9ea763d34a..540be8f856 100644
--- a/noir-projects/aztec-nr/aztec/src/macros/utils.nr
+++ b/noir-projects/aztec-nr/aztec/src/macros/utils.nr
@@ -147,7 +147,7 @@ comptime fn signature_of_type(typ: Type) -> Quoted {
         let (s, generics) = typ.as_data_type().unwrap();
         let field_signatures = s
             .fields(generics)
-            .map(|(_, typ): (Quoted, Type)| signature_of_type(typ))
+            .map(|(_, typ, _): (Quoted, Type, Quoted)| signature_of_type(typ))
             .join(quote {,});
         f"({field_signatures})".quoted_contents()
     } else if typ.as_tuple().is_some() {
@@ -220,7 +220,7 @@ pub(crate) comptime fn compute_event_selector(s: TypeDefinition) -> Field {
     let event_name = s.name();
     let args_signatures = s
         .fields_as_written()
-        .map(|(_, typ): (Quoted, Type)| {
+        .map(|(_, typ, _): (Quoted, Type, Quoted)| {
             // signature_of_type can handle structs, so this supports nested structs
             // FIXME: Field generics are not handled here!
             signature_of_type(typ)
diff --git a/noir-projects/noir-protocol-circuits/crates/types/src/meta/mod.nr b/noir-projects/noir-protocol-circuits/crates/types/src/meta/mod.nr
index 2b8b636471..5c02151585 100644
--- a/noir-projects/noir-protocol-circuits/crates/types/src/meta/mod.nr
+++ b/noir-projects/noir-protocol-circuits/crates/types/src/meta/mod.nr
@@ -124,7 +124,7 @@ pub comptime fn generate_deserialize_from_fields(

         // Iterate over each field in the struct
         for field in nested_def.fields(generics) {
-            let (field_name, field_type) = field;
+            let (field_name, field_type, _visibility) = field;
             // Recursively call `generate_deserialize_from_fields` for each field in the struct
             let (deserialized_field, num_consumed_in_recursion) = generate_deserialize_from_fields(
                 field_name,
@@ -344,7 +344,7 @@ pub comptime fn generate_serialize_to_fields(
         // For struct we pref
         let nested_struct = typ.as_data_type().unwrap();
         let params = nested_struct.0.fields(nested_struct.1);
-        let struct_flattened = params.map(|(param_name, param_type): (Quoted, Type)| {
+        let struct_flattened = params.map(|(param_name, param_type, _visility): (Quoted, Type, Quoted)| {
             let maybe_prefixed_name = if name == quote {} {
                 // Triggered when the param name is of a value available in the current scope (e.g. a function
                 // argument) --> then we don't prefix the name with anything.
```

asterite:
@TomAFrench This is a breaking change but there's no way to fix Aztec-Packages before merging this PR. It seems external check failures block merging PRs... is that expected?

TomAFrench:
Yep, that's expected. In situations like this you can mark the failures as expected and then merge the PR. Once aztec-packages is fixed and we bump the commit then we can mark them as expected to work again.

asterite:
Aaaah... I forgot about that! I implemented that 🤦😅

asterite:
@TomAFrench It seems "Compile `noir-contracts`" is currently a hard requirement for CI, so this PR can't be merged.

We could remove that, or... there's something else we could do so that Aztec-Packages compiles even with this change. We could introduce tuples with a "rest" pattern, like `(x, y, ..)`. It should be easy to implement (allowing that syntax, then the pattern succeeds if it has at least 2 elements (in this case)). Then on Aztec-Packages we could use `, ..` to ignore the new visibility, even before it's introduced. What do you think?

TomAFrench:
We can just modify the workflow to assert that the compilation fails and then switch back once updated.

TomAFrench:
Can you modify the workflow so that it directly asserts that the compilation fails? If we just make it optional then it'll never get turned back on plus master will show as failing.

asterite:
I searched but I couldn't find how to do it. I don't know much about bash, if that's what github actions run. I have zsh locally... I tried something like `RESULT=nargo compile ...` but that's not valid syntax in zsh (so later I can check RESULT). 

I'll continue searching...

TomAFrench:
I can solve this on monday.

asterite:
@TomAFrench I think CI is "green", that is, all required checks pass even though reports and some external repositories fail to compile.

.github/workflows/reports.yml
```diff
@@ -754,9 +754,11 @@ jobs:
         - name: Report overall success
           run: |
             if [[ $FAIL == true ]]; then
-                exit 1
-            else
+                echo "Expected reports-end to fail but it didn't. Please update the workflow to assert that the command now does not fail."
```
TomAFrench, line 757:
```suggestion
```

.github/workflows/reports.yml
```diff
@@ -754,9 +754,11 @@ jobs:
...
                 exit 0
+            else
+                echo "This step is currently expected to fail."
+                exit 1
```
TomAFrench, line 761:
```suggestion
                exit 1
            else
                exit 0
```

TomAFrench (approved):

PR #8424, opened 2025-05-08
fix!: error on tuple mismatch

### LINKED ISSUE #8422: Exact number of tuple elements isn't required in lambdas
### Aim

Given this code:

```node
fn main() {
    let x = [(1, 2, 3)];
    let _ = x.map(|(a)| a);
}
```

### Expected Behavior

The code shouldn't compile, as `(a)` doesn't match the expected tuple length. It should be something like `(a, b, c)`.

### Bug

The compiler crashes:

```
Message:  ICE: at least one variant should exist for a dynamic call Signature { params: [Numeric(NativeField), Numeric(NativeField), Numeric(NativeField)], returns: [Numeric(NativeField)] }
Location: compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs:374
```

It also crashes if we use `(a, b)`, but once we use `(a, b, c)` in at least one `map` call then it compiles fine.

It also even compiles fine if we give more tuple elements than needed:

```noir
fn main() {
    let x = [(1, 2, 3)];
    let _ = x.map(|(a, b, c, d)| a);
}
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8422

## Documentation

Check one:

# PR Checklist


jfecher (approved):
A bit embarrassing we were missing this for so long!

asterite:
There are some errors on Aztec-Packages that we could fix by ignoring the extra elements, but I'd leave that to others because maybe they shouldn't be ignored.

jfecher:
@asterite agreed, this should be a hard error. Aztec packages will need to be fixed - they should not be writing code that relies on this to begin with.

PR #8432, opened 2025-05-09
feat(greybox_fuzzer): Parallel fuzz tests

Rumata888:
# Description

## Problem\*

Resolves 

## Summary\*

Fuzzing tests are now run after the main tests and each fuzzing test uses parallelisation internally instead of just being assigned 1 thread. The limits have also been changed to 100k executions and 1 second.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/docs/docs/tooling/testing.md
+++ b/docs/docs/tooling/testing.md
@@ -88,7 +88,7 @@ fn test_basic(a: Field, b: Field) {
     assert(a + b == b + a);
 }
 ```
-The test above is not expected to fail. By default, the fuzzer will run for 10 seconds and use 100000 executions (whichever comes first). All available threads will be used for each fuzz test.
+The test above is not expected to fail. By default, the fuzzer will run for 1 seconds and use 100000 executions (whichever comes first). All available threads will be used for each fuzz test.
 The fuzz tests also work with `#[test(should_fail)]` and `#[test(should_fail_with = "<the reason for failure>")]`. For example:
 
 ```rust
--- a/docs/docs/tooling/testing.md
+++ b/docs/docs/tooling/testing.md
@@ -134,9 +134,9 @@ There are some fuzzing-specific options that can be used with `nargo test`:
           If given, store the failing input in the given folder
 
       --fuzz-timeout <FUZZ_TIMEOUT>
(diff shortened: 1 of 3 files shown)


tooling/nargo_cli/src/cli/test_cmd.rs
```diff
@@ -93,9 +97,17 @@ pub(crate) struct TestCommand {
     #[arg(long)]
     fuzzing_failure_dir: Option<String>,
 
-    /// Maximum time in seconds to spend fuzzing (default: 1 second)
+    /// Maximum time in seconds to spend fuzzing (default: 1 seconds)
     #[arg(long, default_value_t = 1)]
     fuzz_timeout: u64,
```
asterite, line 102: Should this be changed to 10 seconds?

asterite (approved):
Looks great!

tooling/nargo_cli/src/cli/test_cmd.rs
Rumata888: No, I left it as 1, because it blew up the timing on aztec.nr.

docs/docs/tooling/testing.md
```diff
@@ -76,4 +76,76 @@ fn test_king_arthur() {
...
+
+You can write fuzzing harnesses that will run on `nargo test` by using the decorator `#[test]` with a function that has arguments. For example:
+
+```rust
+#[test]
+fn test_basic(a: Field, b: Field) {
+    assert(a + b == b + a);
+}
+```
+The test above is not expected to fail. By default, the fuzzer will run for 10 seconds and use 100000 executions (whichever comes first). All available threads will be used for each fuzz test.
```
asterite, line 91: But then this should say it's 1 second, right?

asterite (changes requested):
I think there's a mismatch between the 1 second by default and the documented 10 seconds.

docs/docs/tooling/testing.md
Rumata888: You're right

Rumata888:
> I think there's a mismatch between the 1 second by default and the documented 10 seconds.

Yes, thanks for catching that

asterite (approved):

PR #8436, opened 2025-05-09
feat(fuzz): AST fuzzing with SSA interpreter

### LINKED ISSUE #8413: Run the SSA pipeline up to a given prefix and execute with the interpreter
Create a fuzz target that does the following:
1. Generate an arbitrary AST
2. Generate an arbitrary input based on the ABI
3. Generate the initial SSA from the AST
4. Choose an arbitrary prefix length N of the SSA pipeline
5. Apply the first N-1 passes on the SSA
6. Apply the N-th pass on the SSA
7. Execute the interpreter on the SSAs of the previous two steps with the random inputs and compare that the result is the same.

aakoshh:
# Description

## Problem\*

Resolves #8413 
Part of https://github.com/noir-lang/noir/issues/8218

## Summary\*

Adds a new `pass_vs_prev` target, which uses the SSA interpreter to execute two consecutive passes and check that the results between them don't change. 

Currently limited to the first 3 passes that participate in the canary audit.

## Additional Context

This includes a few fixes for the SSA interpreter:
* Allow `u1` shifting
* Handle negative integers

Regarding https://github.com/noir-lang/noir/issues/8451 that came up testing this, we're discussing whether we should ban 0 array inputs (at least for `main`). If not, the mapping of ABI to SSA values should work differently when the parameter is `call_data` and an empty array, because the call_data expects empty arrays to be skipped, while normal data access expects them to be present.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/interpreter/value.rs
```diff
@@ -39,7 +39,7 @@ pub(crate) enum NumericValue {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub(crate) struct ReferenceValue {
+pub struct ReferenceValue {
```
jfecher, line 42: If you're going to automatically compare values across iterations, you should ignore the `original_id` field here since that may change from optimization passes

aakoshh: Thanks! I only plan to compare the final return value, so I don't expect it to contain references, but good to have that insight 🙏

compiler/noirc_evaluator/src/ssa/interpreter/value.rs
```diff
@@ -227,38 +228,28 @@ impl NumericValue {
...
-            other => panic!("Unsupported numeric type: {other}"),
+            typ => match IntegerConstant::from_numeric_constant(constant, typ) {
+                Some(IntegerConstant::Unsigned { value, bit_size }) => match bit_size {
+                    8 => Self::U8(value.try_into().expect("not u8")),
+                    16 => Self::U16(value.try_into().expect("not u16")),
+                    32 => Self::U32(value.try_into().expect("not u32")),
+                    64 => Self::U64(value.try_into().expect("not u64")),
+                    128 => Self::U128(value),
+                    bs => panic!("Unsupported unsigned type: u{bs}"),
+                },
```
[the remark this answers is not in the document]
aakoshh: Great, thank you! In the meantime I discovered that `u1` also needs to support arithmetic operations. I'm working on adding them, but I expect it's not going to be pretty. I'll push it, and then if you don't like it then please ignore that commit.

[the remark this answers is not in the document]
jfecher: I'll include those in the PR too then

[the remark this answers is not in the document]
aakoshh: I pushed my changes but haven't fixed the unit tests for the interpreter. It passed a few times on the first 3 passes. I created tickets for failures that appear later in the pipeline.

tooling/ast_fuzzer/src/input/dictionary.rs
```diff
@@ -0,0 +1,9 @@
+use std::collections::BTreeSet;
+
+use acir::FieldElement;
+use noirc_evaluator::ssa::ssa_gen::Ssa;
+
+pub(crate) fn build_dictionary_from_ssa(_ssa: &Ssa) -> BTreeSet<FieldElement> {
+    // TODO: Traverse the SSA to collect fields.
```
aakoshh, line 7: Follow up issue: https://github.com/noir-lang/noir/issues/8467

tooling/ast_fuzzer/src/compare/interpreted.rs
```diff
@@ -0,0 +1,303 @@
...
+        // println!(
+        //     "input values:\n{}\n",
+        //     self.input_values
+        //         .iter()
+        //         .enumerate()
+        //         .map(|(i, v)| format!("{i}: {v}"))
+        //         .collect::<Vec<_>>()
+        //         .join("\n")
+        // );
+        // println!("SSA after step {} ({}):\n{}\n", self.ssa1.step, self.ssa1.msg, self.ssa1.ssa);
```
rkarabut, line 90:
```suggestion
```

rkarabut, line 90:
```suggestion
```

rkarabut: Oh, maybe add `tracing` along with subscriber into the crate then? It would actually be pretty useful to leave `debug!` here and there

PR #8437, opened 2025-05-09
fix: always type-check turbofish, and error when it's not allowed

### LINKED ISSUE #8434: Turbofish is ignored in many cases
### Aim

There are many situations where turbofish is allowed syntactically but is ignored or leads to a strange error when used in places that don't expect a turobfish.

Some example:

### Turbofish in a global

```noir
global x: Field = 1;

fn main() {
    let _ = x::<hello, world>;
}
```

No error at all. There should be three errors: "hello" and "world" can't be found, then that "x" doesn't accept generic arguments.

### Turbofish in a local variable

```noir
fn main() {
    let x = 1;
    let _ = x::<hello, world>;
}
```

This errors with:

```
error: Could not resolve 'x' in path
  ┌─ src/main.nr:3:13
  │
3 │     let _ = x::<hello, world>;
  │             -
```

which is misleading. And, again, "hello" and "world" don't produce errors.

### Turbofish after "Self"

```noir
struct Foo {}

impl Foo {
    fn new() {
        Self::<i32> {}
    }
}
```

Gives this error:

```
error: struct Foo expects 0 generics but 1 was given
  ┌─ src/main.nr:5:13
  │
5 │         Self::<i32> {}
  │             -------
  │

error: expected type (), found type Foo
  ┌─ src/main.nr:4:13
  │
4 │     fn new() {
  │             - expected () because of return type
5 │         Self::<i32> {}
  │         -------------- Foo returned here
  │
  = help: try adding a return type: `-> Foo`
```

However, Rust disallows turobfish for "Self" (semantically) so we should probably do the same here.

### Turbofish on a module

```noir
mod moo {
    fn foo() {}
}

fn main() {
    moo::<hello>::foo();
}
```

The error is:

```
error: turbofish (`::<_>`) not allowed on module `moo`
  ┌─ src/main.nr:6:8
  │
6 │     moo::<hello>::foo();
  │        ---------
```

which is good, but it doesn't also say that "hello" cannot be resolved.

### Turbofish on an associated constant

```noir
trait Foo {
    let N: u32;

    fn foo() -> [Field; Self::N::<i32>] {
        [0; Self::N::<i32>]
    }
}
```

No error at all.

### Expected Behavior

The compiler should type turbofish in all cases, and error when these are not supposed to be there in some cases.

Right now when elaborating a variable we always do that on a `Path` that has `UnresolvedType` for the turbofish segments. It would be better to first type-check all those `UnresolvedType`s into another type, maybe `TypedPath`, then use that type to fully resolve the path. That way we also wouldn't need to type turbofish in many places, just in one.

### Bug

What's in the "aim" section.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8434

## Summary

Now when we get to elaborate a path we first type-check all of its turbofish. That already makes sure all turbofish is type-checked: previously it was only type-checked in some scenarios. For this I introduced two new types, `TypedPath` and `TypedPathSegment`... those names don't convince me much because the only thing typed there is the turbofish, but I couldn't find a better name.

Then, looking up paths always takes a `TypedPath`, which makes sure turbofish is resolved in all cases.

Finally, in some existing scenarios I made the compiler produce an error if there's turbofish when it's not allowed. At least all the cases in #8434 are covered.

## Additional Context

Before this PR we were solving turobfish with target Kinds in mind. For example if a generic kind was `let N: u32` then we'd solve constants with those kinds. Now that we solve turbofish before knowing what they'll be matched against, these are solved with `Kind::Any`. Then, once we know the target kind, any `Kind::Any` is replaced with the target kind.

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -2843,6 +2843,46 @@ impl Type {
             _ => None,
         }
     }
+
+    /// Subtitute any [`Kind::Any`] in this type, for types that hold kinds (like [`Type::Constant`])
```
TomAFrench, line 2847:
```suggestion
    /// Substitute any [`Kind::Any`] in this type, for types that hold kinds (like [`Type::Constant`])
```

TomAFrench (approved):

PR #8440, opened 2025-05-09
fix: Fix nested trait dispatch with associated types

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/8252

## Summary\*

This ended up being a case of doing something almost but not quite right. We were throwing away instantiation bindings during impl search which ends up being fine most of the time when you already have an impl (by using method syntax which looks inside the struct type to find the exact impl method beforehand). If you don't however, you'll only have the instantiation bindings from instantiating your current trait - not the trait impl! So when doing `Eq::eq(a, b)` we'd instantiate the object types since Eq requires us instantiate Self, but we don't get the additional type variables on the specific Eq impl which requires the implicit generics from associated types from other traits (in the case of the bugged program at least).

I adjusted impl search slightly to return the instantiation bindings of the selected impl as well, and store them if `select_impl` is true. If we store them without `select_impl` being true we can accidentally override other type bindings - I made a note of this in the code.

## Additional Context

Dobby is a free elf now!

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
@asterite do you know where the blacklist is for the nargo expand test?
Looks like it produces invalid syntax for the regression example preventing the expanded regression from compiling.

asterite:
It's in builders inside nargo_cli. There are many constants for all the tests we ignore.

asterite (approved):
Looks good!

What do you think if we also add the tests from https://github.com/noir-lang/noir/pull/8279 ?

TomAFrench:
I've ignored the `nargo expand` test as it's unrelated to the issue which this PR is fixing. I've opened an issue here: https://github.com/noir-lang/noir/issues/8449

TomAFrench:
I've tested that this fixes the failures in https://github.com/noir-lang/noir_bigcurve/pull/46. Miranda is going to be blocked by this very soon so merging so this ends up in the next nightly.

TomAFrench:
:huh:

jfecher:
Disabled auto-merge, I'm adding the extra failing test from https://github.com/noir-lang/noir/pull/8279. The others should be covered by the existing test.

Issue #7228, opened 2025-01-29
Consolidate multiple implementations of function call graphs across codebase


TomAFrench:
We've gradually accumulated multiple implementations of a graph to track which functions are calling other functions during SSA.

Examples are:
https://github.com/noir-lang/noir/blob/0c4687025ad8ff2faa0512223147977b1bc28b46/compiler/noirc_evaluator/src/ssa/opt/inlining.rs#L309-L344

The purity callgraph added in https://github.com/noir-lang/noir/pull/7197.

The callgraph to mark reachable globals in https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_globals.rs added in #7188 

We should see if we can standardize these to avoid code duplication and potential inconsistencies.

PR #8447, opened 2025-05-12
fix(ssa): Mark mutually recursive simple functions

### LINKED ISSUE #8414: Compiler crash inlining simple mutually recursive functions
### Aim

Follow up for https://github.com/noir-lang/noir/issues/8078 

https://github.com/noir-lang/noir/pull/8127 solved the above issue by not inlining simple self recursion, but it doesn't handle the case of mutual recursion. 

```rust
fn main() {
    // Safety: testing context
    unsafe {
        starter();
    }
}

unconstrained fn starter() {
    ping();
}

unconstrained fn ping() {
    pong()
}

unconstrained fn pong() {
    ping()
}
```

### Expected Behavior

This is a pathological example because the the crash happens in `inline_functions_with_at_most_one_instruction`, and to do anything useful in `ping` or `pong`, e.g. to decrement a parameter, requires at least another instruction, and this program _is_  infinitely recursive at runtime,  however, it should still compile. 

### Bug

```console
❯ cargo run -q -p nargo_cli -- compile --show-ssa-pass "at most one"
Message:  Attempted to recur more than 1000 times during inlining function 'ping':
brillig(inline) fn ping f2 {
  b0():
    call f3()
    return
}
Location: compiler/noirc_evaluator/src/ssa/opt/inlining.rs:252
```

### To Reproduce

See above. 

Note that we need the `starter` function: if we call `ping` from `main` the crash does not happen. That's because once we inline `pong` into `ping`, and then recursively try to inline `ping` into itself, `PerFunctionContext::should_inline_call` knows that it's trying to inline the callee into the entry, which won't work. But if the entry is another wrapper function, it will try. `main` does not count as a wrapper in this case because it's ACIR, and we would not try to inline a Brillig function into it. 

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+d8e4de40d05efe79b970055f42bd72f061f6aae6 (git version hash: d8e4de40d05efe79b970055f42bd72f061f6aae6, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #8414 

We were not correctly checking for mutually recursive functions in `inline_functions_with_at_most_one_instruction`.

## Summary\*

I chose to lean into computing a full graph here as to push for re-use of the same call graph. As noted by https://github.com/noir-lang/noir/issues/7228 we have multiple implementations of call graphs. Even though `inline_functions_with_at_most_one_instruction` should ideally only process functions with a single instruction, I felt leaning into a single call graph was better and simpler than expanding the `inline_functions_with_at_most_one_instruction` recursion check to account for mutual recursion cycles.

I went with using the call graph from `opt::pure`. I went with this graph as it uses [petgraph](https://github.com/petgraph/petgraph). Using this crate simplifies our graph generation and provides nice utility methods out of the box such as fetching strongly connected components (SCC). 

After generating the `CallGraph` we determine the recursive functions by checking the graph's SCCs. 

## Additional Context

I experimented with using the `InlineInfo` computation. This revealed that the call graph there is not appropriately marking recursive functions. I have captured an issue here https://github.com/noir-lang/noir/issues/8448 and I have added a test with `#[should_panic]`. I will resolve this issue in a follow-up where the `InlineInfo` computation uses the same call graph as we use in this PR. 

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/inlining/inline_info.rs
```diff
@@ -365,3 +366,54 @@ fn mark_brillig_functions_to_retain(
...
+          b0():
+            call f2()
+            return
+        }
+        ";
+
+        let ssa = Ssa::from_str(src).unwrap();
+        let inline_infos = compute_inline_infos(&ssa, false, i64::MAX);
+
+        let func_0 = inline_infos.get(&Id::test_new(0)).expect("Should have made ");
```
aakoshh, line 407: "Should have made " looks like it's missing something.

compiler/noirc_evaluator/src/ssa/opt/inlining/inline_info.rs
```diff
@@ -365,3 +366,54 @@ fn mark_brillig_functions_to_retain(
...
+#[cfg(test)]
+mod tests {
+    use crate::ssa::{ir::map::Id, ssa_gen::Ssa};
+
+    use super::compute_inline_infos;
+
+    // This panic is a regression that will need to be updated by fixing the inline infos computation.
+    // https://github.com/noir-lang/noir/issues/8448
+    #[test]
+    #[should_panic]
```
aakoshh, line 379: Can you perhaps indicate the expected panic message, to give a hint as to what is wrong with the inline infos?

aakoshh (approved):
It's great to see general graph algorithms coming into action 👍

compiler/noirc_evaluator/src/ssa/opt/inlining/inline_info.rs
vezenovm: Good idea, done

PR #8452, opened 2025-05-12
chore: add some simple post-checks to certain SSA passes

TomAFrench:
# Description

## Problem\*

From discussion with @michaeljklein 

## Summary\*

This PR adds some sanity checks to SSA passes so that, for instance,  the SSA pass to remove bitshifts does not leave behind any bitshifts in ACIR functions.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/pure.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/pure.rs
@@ -49,8 +49,21 @@ impl Ssa {
             function.dfg.set_function_purities(purities.clone());
         }
 
+        purity_analysis_post_check(&self);
+
         self
     }
+}
+
+/// Post-check condition for [Function::remove_bit_shifts].
+///
+/// Returns `true` if:
+///   - all functions have a purity status attached to it.
+///
+/// Otherwise returns `false`.
+fn purity_analysis_post_check(ssa: &Ssa) -> bool {
+    // TODO: We have N different copies of the purity hashmap (one per function), ideally this would be deduplicated.
+    ssa.functions.iter().all(|(id, function)| function.dfg.purity_of(*id).is_some())
 }
 
 pub(crate) type FunctionPurities = HashMap<FunctionId, Purity>;
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
@@ -79,6 +79,8 @@ impl Function {
 
             context.replace_value(old_result, new_result);
         });
+
+        debug_assert!(remove_bit_shifts_post_check(self));
     }
 }
 
@@ -341,6 +343,35 @@ impl Context<'_, '_, '_> {
             self.call_stack,
         )
     }
+}
+
+/// Post-check condition for [Function::remove_bit_shifts].
+///
+/// Returns `true` if:
+///   - `func` is not an ACIR function, OR
+///   - `func` does not contain any bitshift instructions.
+///
+/// Otherwise returns `false`.
+fn remove_bit_shifts_post_check(func: &Function) -> bool {
+    // Non-ACIR functions should be unaffected.
+    if !func.runtime().is_acir() {
+        return true;
+    }
+
+    // Otherwise there should be no shift-left or shift-right instructions in any reachable block.
+    for block_id in func.reachable_blocks() {
+        let instruction_ids = func.dfg[block_id].instructions();
+        for instruction_id in instruction_ids {
+            if matches!(
+                func.dfg[*instruction_id],
+                Instruction::Binary(Binary { operator: BinaryOp::Shl | BinaryOp::Shr, .. })
+            ) {
+                return false;
+            }
+        }
+    }
+
+    return true;
 }
 
 #[cfg(test)]
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_if_else.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_if_else.rs
@@ -43,6 +43,7 @@ impl Function {
         } else {
             Context::default().remove_if_else(self);
         }
+        debug_assert!(remove_if_else_post_check(self));
     }
 }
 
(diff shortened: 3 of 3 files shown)


compiler/noirc_evaluator/src/ssa/opt/pure.rs
```diff
@@ -49,10 +49,23 @@ impl Ssa {
             function.dfg.set_function_purities(purities.clone());
         }
 
+        purity_analysis_post_check(&self);
```
asterite, line 52:
```suggestion
        debug_assert!(purity_analysis_post_check(&self));
```

compiler/noirc_evaluator/src/ssa/opt/pure.rs
```diff
@@ -49,10 +49,23 @@ impl Ssa {
...
 }
 
+/// Post-check condition for [Function::remove_bit_shifts].
+///
+/// Returns `true` if:
+///   - all functions have a purity status attached to it.
+///
+/// Otherwise returns `false`.
+fn purity_analysis_post_check(ssa: &Ssa) -> bool {
+    // TODO: We have N different copies of the purity hashmap (one per function), ideally this would be deduplicated.
```
asterite, line 65: Should an issue be captured for this?

vezenovm: `FunctionPurities` are contained within an Arc which provides shared ownership.

compiler/noirc_evaluator/src/ssa/opt/pure.rs
```diff
@@ -49,10 +49,23 @@ impl Ssa {
             function.dfg.set_function_purities(purities.clone());
         }
 
+        debug_assert!(purity_analysis_post_check(&self));
```
vezenovm, line 52: We are cloning each purity above. We should be able to just do this debug assertion a single time before storing the purities in each function DFG

compiler/noirc_evaluator/src/ssa/opt/remove_if_else.rs
```diff
@@ -224,6 +225,32 @@ fn slice_capacity_change(
...
+///   - `func` does not contain any if-else instructions.
+///
+/// Otherwise returns `false`.
+fn remove_if_else_post_check(func: &Function) -> bool {
+    // Brillig functions should be unaffected.
+    if func.runtime().is_brillig() {
+        return true;
+    }
+
+    // Otherwise there should be no shift-left or shift-right instructions in any reachable block.
```
vezenovm, line 241:
```suggestion
    // Otherwise there should be no if-else instructions in any reachable block.
```

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -79,6 +79,8 @@ impl Function {
 
             context.replace_value(old_result, new_result);
         });
+
+        debug_assert!(remove_bit_shifts_post_check(self));
```
aakoshh, line 83: Perhaps instead of returning a bool, we could create a macro that calls the function only if `#[cfg(debug_assertions)]` as suggested [here](https://users.rust-lang.org/t/show-value-only-in-debug-mode/43686/3), and then we can include a bit more information in the panic.

aakoshh: I did that in https://github.com/noir-lang/noir/pull/8481

TomAFrench:
Huh, seems like we have a legit issue with purity analysis?

compiler/noirc_evaluator/src/ssa/opt/pure.rs
```diff
@@ -100,14 +118,24 @@ impl std::fmt::Display for Purity {
 
 impl Function {
     fn is_pure(&self) -> (Purity, BTreeSet<FunctionId>) {
+        // Note, this function must be allowed to complete despite the fact that once the function is marked as impure
+        // then its final purity is known. This is because we need to collect all of the dependencies of the function
+        // to ensure that they are processed.
+        //
+        // This can be relaxed if we calculate the callgraph separately.
```
TomAFrench, line 125: cc @vezenovm as you're centralising how we generate callgraphs.

TomAFrench:
This is now ready for review

compiler/noirc_evaluator/src/ssa/opt/pure.rs
vezenovm: This is referring more to how we collect the function dependencies and when `is_pure` should be called. We ultimately use our `CallGraph` in this pass still.

vezenovm (approved):

compiler/noirc_evaluator/src/ssa/opt/pure.rs
```diff
@@ -49,10 +49,28 @@ impl Ssa {
...
         }
 
+        #[cfg(debug_assertions)]
+        purity_analysis_post_check(&self);
+
         self
     }
 }
 
+/// Post-check condition for [Function::remove_bit_shifts].
```
vezenovm, line 59:
```suggestion
/// Post-check condition for [Ssa:: purity_analysis].
```

TomAFrench:
@aakoshh Can you take all the fuzz stuff out of the merge queue? It seems like every other PR gets blocked on this.

compiler/noirc_evaluator/src/ssa/opt/pure.rs
```diff
@@ -116,13 +144,9 @@ impl Function {
...
-        let mut result = if self.runtime().is_acir() {
-            Purity::Pure
-        } else {
-            // Because we return bogus values when a brillig function is called from acir
-            // in a disabled predicate, brillig functions can never be truly pure unfortunately.
-            Purity::PureWithPredicate
-        };
+        if self.parameters().iter().any(&contains_reference) {
+            result = Purity::Impure;
+        }
```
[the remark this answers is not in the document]
TomAFrench: Yeah, I reverted the changes to which avoided an early return in this PR in #8667. This implementation should still be correct however as we use `result = new_result.unify(result)` which always takes the most impure status for the function. We only directly assign to `result` when we're setting it as impure.

PR #8470, opened 2025-05-12
feat: primitive types are no longer keywords

### LINKED ISSUE #8023: Confusing error message about using `i128`
### Aim

I was trying to [find out ](https://github.com/noir-lang/noir/pull/7987#discussion_r2039245363)if a user can use `i128` in their programs, because older versions of `nargo` underlined it as an error, and some arithmetic operations on them cause compiler crashes. 

I tried now with this program:
```rust
fn main(a: i128, b: i128) -> pub bool {
    a < b
}
```

### Expected Behavior

I wanted to run this program to see:
* if the compiler accepts `i128` as user input
* if it accepts them, then can it compare the values without crashing

### Bug

The compiler tells me that my use of `i128` is invalid, but doesn't mention the _sign_ of the type; instead it lists 128 bits as one of the possible values I can use:

```console
error: Use of invalid bit size 128
  ┌─ src/main.nr:2:12
  │
2 │ fn main(a: i128, b: i128) -> pub bool {
  │            ---- Allowed bit sizes for integers are 1, 8, 16, 32, 64, 128
  │

error: Use of invalid bit size 128
  ┌─ src/main.nr:2:21
  │
2 │ fn main(a: i128, b: i128) -> pub bool {
  │                     ---- Allowed bit sizes for integers are 1, 8, 16, 32, 64, 128
  │

```

### To Reproduce

See above.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+0f903df32beefdcae0f1465abb28bcc98eca49d1 (git version hash: 10bce87eac63bf4313fb707848720b8a7d0b0939, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

No issue, just noticed that Noir doesn't work like Rust in this regards.

Also trivially fixes #8023

This also uncovered an inconsistency in the language: you write `let _ = Vec::<i32>::new();` but for strings you currently have to write it like `let _ = str<N>::something();` (no `::` between `str` and `<`). We could accept that inconsistency if we keep parsing primitive types in a special form, or we could improve things now to avoid breaking changes after 1.0. For this, we could maybe first create a PR to allow writing `str::<N>::something();` (would parse to a `TypePath`), give a warning if someone writes `str<N>` in an expression context, then use the correct syntax in external crates, then this PR wouldn't produce errors anymore.

## Summary

In Rust, primitive types like `i32`, `bool`, etc., are not keywords. They just resolve to types at the very last moment, after searching through existing data types. In Noir, before this PR, they were keywords. Not anymore.

This isn't super important, but it gives more flexibility to the language. For example:
-  a few times in the past I wrote `let bool = true;` for some quick prototype code, only to find out that that doesn't parse (but it does in Rust)
- if you want to have a factory of values, you couldn't have a function `i32` that returned `i32`. Now you can.
- if you wanted to define a type in your crate called `Expr` for some reason, you couldn't as it was a keyword
- if you wanted to invoke a method on `str`, for example `str::as_bytes(...)`, you couldn't, as we'd only parse `str` followed by generics
- now `str<1, 2>` is parsed correctly, except that it gives a generics count mismatch error (again, not super important, but it's nice for LSP recovery as that's not a syntax error, just a semantic error)

This also allows us to introduce new primitive types in the future without breaking any code, as users can use any name for their types, for example `char`, and if we introduce a `char` type, the `char` in their code would refer to their `char`, not the primitive type.

I think this also simplifies the compiler a bit as we no longer need a keyword, an unresolved type, and logic to resolve that unresolved type, for every primitive type. (also: the diff shows that this ends up using less code)

## Additional Context

1. To solve a path like `i32::default()`, which was previously parsed as a `TypePath`, I needed the same logic to finding direct functions or trait functions in scope that exists somewhere else. There's a bit of code duplication but it's not that much. I couldn't find a way to improve this, but we could capture an issue to try to do it.
2. In Rust you can define `struct i32 { ... }` which won't conflict with the primitive type `i32`. But what if you wanted to refer to the primitive type `i32` in that same file? You can, using `std::primitive::i32`. We could do the same by using type aliases (I think!) but that case should be rare so I decided not to do that in this PR.
3. While making this change code like this stared failing: `assert!(3 as u8 < 4)`. It actually fails in Rust too! The reason is that `u8 < ` actually denotes a generic type `u8<...>`. In Rust you have to write this as `assert!((3 as u8) < 4)`. So this PR would introduce a breaking change. To avoid that, I made the parser not parse generics on the type given to `as`. However, I think the "correct" thing would be to error like in Rust, especially if in the future we find some use case for casting between other types that aren't primitive types and that might be generics. In that case it'll break code so maybe we should introduce that breaking change before 1.0.

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/ast/expression.rs
```diff
@@ -116,17 +117,24 @@ impl UnresolvedGeneric {
...
-
-        match typ.typ {
-            FieldElement => Ok(Type::FieldElement),
-            Integer(sign, bits) => Ok(Type::Integer(sign, bits)),
-            // Only fields and integers are supported for numeric kinds
-            _ => Err(UnsupportedNumericGenericType {
-                ident: self.ident().clone(),
-                typ: typ.typ.clone(),
-            }),
+        // TODO: this should be done with resolved types
```
TomAFrench, line 120: Unresolved TODO

compiler/noirc_frontend/src/ast/expression.rs
```diff
@@ -116,17 +117,24 @@ impl UnresolvedGeneric {
...
+        use crate::ast::UnresolvedTypeData::Named;
+
+        if let Named(path, _generics, _) = &typ.typ {
+            if path.segments.len() == 1 {
+                if let Some(primitive_type) =
+                    PrimitiveType::lookup_by_name(path.segments[0].ident.as_str())
+                {
+                    // TODO: check generics
```
TomAFrench, line 128: Unresolved TODO

compiler/noirc_frontend/src/elaborator/path_resolution.rs
```diff
@@ -400,16 +409,27 @@ impl Elaborator<'_> {
...
+                importing_module,
+                intermediate_item,
+                target,
+                mode,
+            ),
+            Err(PathResolutionError::Unresolved(err)) => {
+                if let Some(resultion) =
+                    self.resolve_primitive_type_or_function(path, importing_module)
+                {
+                    return resultion;
```
TomAFrench, line 427:
```suggestion
                if let Some(result) =
                    self.resolve_primitive_type_or_function(path, importing_module)
                {
                    return result;
```

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -1365,7 +1365,21 @@ fn unresolved_type_is_bool(
 ) -> IResult<Value> {
     let self_argument = check_one_argument(arguments, location)?;
     let typ = get_unresolved_type(interner, self_argument)?;
-    Ok(Value::Bool(matches!(typ, UnresolvedTypeData::Bool)))
+
+    // TODO: we should resolve the type here instead of just checking the name
```
TomAFrench, line 1369: Unresolved TODO

compiler/noirc_frontend/src/hir/comptime/interpreter/builtin.rs
```diff
@@ -1376,7 +1390,21 @@ fn unresolved_type_is_field(
 ) -> IResult<Value> {
     let self_argument = check_one_argument(arguments, location)?;
     let typ = get_unresolved_type(interner, self_argument)?;
-    Ok(Value::Bool(matches!(typ, UnresolvedTypeData::FieldElement)))
+
+    // TODO: we should resolve the type here instead of just checking the name
```
TomAFrench, line 1394: Unresolved TODO

compiler/noirc_frontend/src/hir/def_collector/dc_mod.rs
```diff
@@ -892,16 +893,23 @@ impl ModCollector<'_> {
...
     ) -> Type {
-        match &typ.typ {
-            UnresolvedTypeData::FieldElement => Type::FieldElement,
-            UnresolvedTypeData::Integer(sign, bits) => Type::Integer(*sign, *bits),
-            _ => {
-                let error =
-                    ResolverError::AssociatedConstantsMustBeNumeric { location: typ.location };
-                errors.push(error.into());
-                Type::Error
+        // TODO: delay this to the Elaborator
```
TomAFrench, line 896: Unresolved TODO

TomAFrench (commented):
Quick first pass, there's a few TODOs that have been added to the codebase. Can you resolve these or create issues (and link to them from the codebase)?

asterite:
Done! All those TODOs are for these two issues:
- https://github.com/noir-lang/noir/issues/8504
- https://github.com/noir-lang/noir/issues/8505

I removed the "TODO: check generics" comments as they are also related to the first issue above (once that's solved, generics will be checked).

TomAFrench (approved):

PR #8473, opened 2025-05-13
chore(test): Replicate comptime stack overflow in a test

aakoshh:
# Description

## Problem\*

Resolves an issue frequently mentioned by @rkarabut, that the comptime vs brillig fuzzer gets stack overflow.

## Summary\*

I ran the following test after adding a few prints to check how far it gets before it gets a stack overflow, and printed the source that caused it. 
```shell
cargo test -p noir_ast_fuzzer_fuzz comptime_vs_brillig -- --nocapture
```

We get this:
```
thread 'targets::comptime_vs_brillig::tests::fuzz_with_arbtest' has overflowed its stack
```

It turned out it was during the call to `prepare_and_compile_snippet`. I added a unit test with one of the offending examples. 

I cannot replicate this stack overflow with `nargo`.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/ast_fuzzer/src/compare/comptime.rs
+++ b/tooling/ast_fuzzer/src/compare/comptime.rs
@@ -129,3 +129,73 @@ impl HasPrograms for CompareComptime {
         vec![&self.program]
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::prepare_and_compile_snippet;
+
+    #[test]
+    fn test_prepare_and_compile_snippet() {
+        let src = r#"
+fn main() -> pub ((str<2>, str<2>, bool, str<2>), bool, [str<2>; 3]) {
+    comptime {
+        let mut ctx_limit = 25;
+        unsafe { func_1_proxy(ctx_limit) }
+    }
+}
+unconstrained fn func_1(ctx_limit: &mut u32) -> ((str<2>, str<2>, bool, str<2>), bool, [str<2>; 3]) {
+    if ((*ctx_limit) == 0) {
+        (("BD", "GT", false, "EV"), false, ["LJ", "BB", "CE"])
+    } else {
+        *ctx_limit = ((*ctx_limit) - 1);
+        let g = if true {
+            let f = {
+                {
+                    let mut idx_a = 0;
+                    loop {
+                        if (idx_a == 4) {
+                            break
+                        } else {
+                            idx_a = (idx_a + 1);
+                            let mut e = if true {
+                                if func_1(ctx_limit).0.2 {
+                                    {
+                                        let b = 38;
+                                        {
+                                            let mut c = false;
+                                            let d = if c {
+                                                c = false;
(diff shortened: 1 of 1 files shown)


aakoshh:
> Ary Borenszweig
It's interesting because it doesn't overflow if you paste that code into a file and compile it
But... if you change ctx_limit to 50 then it stack overflows
My guess is that because it runs in a test, the stack has already some depth so it reaches the overflow faster

> That said... the situation can be improved a bit
If you apply this patch, it doesn't stack overflow anymore, at least not with 25:

```
diff --git a/compiler/noirc_frontend/src/hir/comptime/interpreter.rs b/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
index 605c46115a..9b0d6416f6 100644
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
@@ -508,6 +508,7 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {

     /// Evaluate an expression and return the result.
     /// This will automatically dereference a mutable variable if used.
+    #[inline(always)]
     pub fn evaluate(&mut self, id: ExprId) -> IResult<Value> {
         match self.evaluate_no_dereference(id)? {
             Value::Pointer(elem, true, _) => Ok(elem.borrow().clone()),
@@ -518,6 +519,7 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
     /// Evaluating a mutable variable will dereference it automatically.
     /// This function should be used when that is not desired - e.g. when
     /// compiling a `&mut var` expression to grab the original reference.
+    #[inline(always)]
     fn evaluate_no_dereference(&mut self, id: ExprId) -> IResult<Value> {
         match self.elaborator.interner.expression(&id) {
             HirExpression::Ident(ident, _) => self.evaluate_ident(ident, id),
@@ -710,6 +712,7 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
         evaluate_integer(typ, value, location)
     }

+    #[inline(always)]
     pub fn evaluate_block(&mut self, mut block: HirBlockExpression) -> IResult<Value> {
         let last_statement = block.statements.pop();
         self.push_scope();
@@ -1053,6 +1056,7 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
         Ok(Value::Quoted(Rc::new(tokens)))
     }

+    #[inline(always)]
     pub fn evaluate_statement(&mut self, statement: StmtId) -> IResult<Value> {
         match self.elaborator.interner.statement(&statement) {
             HirStatement::Let(let_) => self.evaluate_let(let_),
```

> I used that trick to allow for more deeply nested structures in the parser too: https://github.com/noir-lang/noir/pull/8347
> When I kept adding #[inline] it would cause compile times (to compile the compiler) to balloon
> Like, it never finished compiling

aakoshh:
Thanks for investigating @asterite

rkarabut (approved):

Issue #8474, opened 2025-05-13
No visibility error when calling primitive function


asterite:
### Aim

NOTE: there are no longer private primitive methods in the standard library so this might not be needed after all, unless we introduce new private methods in the future (but there should never be a need to do so).

The following code doesn't produce any error or warning:

```noir
fn main() {
    Field::__assert_max_bit_size(1, 2);
}
```

### Expected Behavior

It should say that `__assert_max_bit_size` is private, as a warning for now.

### Bug

It doesn't produce an error or warning.

This only happens with "type paths".

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Additional Context

Found while working on #8470 . It's not fixed by #8470

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

jfecher:
> NOTE: there are no longer private primitive methods in the standard library so this might not be needed after all, unless we introduce new private methods in the future (but there should never be a need to do so).

Going to close this 🙂

PR #8496, opened 2025-05-13
fix: remove private builtins from `Field` impl

TomAFrench:
# Description

## Problem\*

A little bit of a weaselly fix for #8474, we'll have other private methods in future I'm sure so we should do a proper fix as well.

## Summary\*

You can't call private method on a primitive type if the type doesn't have any private methods. This also gives us a nicer distinction between the builtin methods and the actual logic of a `Field`.

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
Looks good! Now all these built-ins are grouped together.

PR #8501, opened 2025-05-14
fix: Change `can_be_main` to be recursive

aakoshh:
# Description

## Problem\*

Fixes #8465 

## Summary\*

Fixes an oversight: `can_be_main` has to be recursive otherwise it can pick one of the existing `global` types which contains 0 length sub-types.

## Additional Context

Need this to stop the flaking smoke test on aztec-packages.

## Documentation\*

Check one:

# PR Checklist\*


tooling/ast_fuzzer/src/program/mod.rs
```diff
@@ -425,6 +425,9 @@ impl Context {
                 break;
             }
         }
+        if is_main {
```
michaeljklein, line 428: Leftover debugging print?

tooling/ast_fuzzer/src/program/mod.rs
```diff
@@ -425,6 +425,9 @@ impl Context {
...
+            println!("TYPE FOR MAIN: {typ}");
+        }
```
TomAFrench, line 430:
```suggestion
```

michaeljklein (approved):
LGTM besides possibly-leftover debug `println`

TomAFrench (approved):
Approved once debug code removed. Would also be good to have a frontend test for this.

aakoshh:
> Would also be good to have a frontend test for this.

I think the understanding was that we will actually disallow this in the frontend. I'll crate a ticket.

Issue #8498, opened 2025-05-14
-0 != 0 in comptime


rkarabut:
### Aim

The program

```
fn main() -> pub bool {
  comptime {
    unsafe { -0 == 0}
  }
}
```

returns 0 with the comptime block modifier, and 1 without.

(minimized from a find by @aakoshh running the comptime_vs_brillig target)

### Expected Behavior

Should probably return 1.

### Bug

```
nargo execute -- force
[test1] Circuit witness successfully solved
[test1] Witness saved to target/test1.gz
[test1] Circuit output: Field(0)
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

rkarabut:
(happened originally in form of `-(a as Field) == b as Field` with boolean `a` and `b`)

PR #8511, opened 2025-05-14
fix: Prevent negative zero

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/8498

## Summary\*

Prevents negative zero from being constructed in a SignedField wherever SignedField is used. This was only a known issue for the comptime interpreter but I thought it'd be safer to make these fields private so we could have the constructor check it everywhere.

## Documentation\*

Check one:

# PR Checklist\*


test_programs/compile_success_empty/comptime_negative_zero/src/main.nr
```diff
@@ -0,0 +1,9 @@
+fn main() {
+    comptime {
+        assert_eq(-0, 0);
+        assert_eq(-1 + 1, 0);
+        assert_eq(-1 - -1, 0);
+        assert_eq(0 * -1, 0);
+        assert_eq(0 / -1, 0);
```
jfecher, line 7: These weren't failing before (the issue was in `impl Neg for SignedField`) but I figured they were good for completeness.

vezenovm (approved):

rkarabut:
@jfecher looks like SignedField needs its fields to be public?

jfecher:
@rkarabut those just need to be updated to use the accessor methods now instead.

Didn't know this PR didn't merge already, I'll fix it today.

PR #8512, opened 2025-05-14
feat: `..` in tuple pattern

asterite:
# Description

## Problem

No issue, but:
- with this we could move forward with #8420 without ever introducing a breaking change (we can add `, ..` in the existing code where the new tuple element will show up before merging that PR)
- I guess it's something we were going to eventually implement, so...

## Summary

I decided to implement this to unblock #8420 because I thought it was going to be easy: "just" parse the new syntax, then ignore the extra elements. But in the end it was a bit more complex:
- I initially thought that ".." could only show up at the end of a tuple pattern, but in reality it can show up in any part of it
- The ".." logic must also be applied during monomorphization in a couple of places, and for the interpreter

That said, besides that it was relatively straight-forward.

I included a big test that proves this is working well in many cases.

## Additional Context

- There was a `pattern_to_string` function that replicated the logic of turning a pattern into a string, so I removed it
- I took the opportunity to extract some code into helper functions to reduce the size of some big functions
- I noticed some methods were unused so I removed them

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/ast/statement.rs
```diff
@@ -568,6 +568,7 @@ pub enum Pattern {
     Tuple(Vec<Pattern>, Location),
     Struct(Path, Vec<(Ident, Pattern)>, Location),
     Parenthesized(Box<Pattern>, Location),
+    DoubleDot(Location),
```
jfecher, line 571: What do you think of including this as a field on Tuple and Struct variants instead of having it be its own pattern?

Notably:
- We should not have more than 1 DoubleDot in a single tuple/struct pattern
- Whether a tuple/struct pattern has a double dot can be a `bool` since its location in the pattern doesn't matter.
  - For formatting it can always be placed at the end of the pattern
- I haven't gotten to the rest of the PR yet but it seems like a DoubleDot would be something that'd be handled within the tuple/struct pattern logic anyway rather than the recursive pattern case. E.g. when checking if all struct fields are present you'd do `!missing_fields.is_empty() || has_double_dot` (pseudocode).

asterite: I actually started implementing it this way!

> We should not have more than 1 DoubleDot in a single tuple/struct pattern

This is true, though at least in Rust this is syntactically allowed but gives a semantic error (for example the formatter will format the file even if you have multiple ".." in a tuple pattern).

> Whether a tuple/struct pattern has a double dot can be a bool since its location in the pattern doesn't matter

That's what I initially thought, because I thought ".." could only be placed at the end of a tuple pattern. But it turns out you can place it in the beginning or in the middle too. For example `let (x, .., z) = (1, 2, 3, 4, 5);` will have `x == 1` and `z == 5`.

For struct patterns, yes, it's just a boolean as only one ".." can show up, and only at the end.

> I haven't gotten to the rest of the PR yet but it seems like a DoubleDot would be something that'd be handled within the tuple/struct pattern logic anyway

This is correct! I actually initially had `DoubleDot` on the frontend AST nodes, and just an `Option<Location>` with the location of the double dot for `HirPattern`. However, printing patterns became a bit cumbersome because the ".." can show up in any index, and the current printing logic just iterates the tuple patterns and puts "," between each pattern. Well, to be honest, we could at least add `Option<Location>` for `HirPattern::Tuple` instead of having `HirPattern::DoubleDot`, and handle the printing logic differently... but I didn't feel that it would be much better. For example, adding `Option<Location>` to `HirPattern::Tuple` will make all `HirPattern::Tuple` bigger in size (though it doesn't matter because it doesn't make `HirPattern` bigger), and then the ".." felt like its own pattern. But I'll try to push a commit that at least changes this for `HirPattern`.

jfecher: My gut is that `..` being on frontend nodes as well could help clean up elaborator code (unverified assumption though of course) and that we could still issue having multiple of them just as a parser error since we control those errors as well.

For the problem of `..` needing an index for tuples, what are your thoughts on splitting tuple patterns into having `fields_before_range: Vec<Pattern>, field_after_range: Vec<Pattern>` where if there is no range the later would be empty. I could also see this being unwieldy though and potentially worse so just wanted to flag it as an option.

asterite: I'll do the change to the frotnend tuple pattern type. I'm not sure about splitting the fields in two, I might try it later

asterite: Done too for the AST node!

> For the problem of .. needing an index for tuples, what are your thoughts on splitting tuple patterns into having fields_before_range: Vec<Pattern>, fields_after_range: Vec<Pattern> where if there is no range the later would be empty.

It's a good idea! But I don't know how we would distinguish between `(x,)` and `(x, ..)`. In both cases the "before" would have one element, the "after" would have zero.

Right now the only "duplicate" code is splitting the patterns at the double dot index, then getting the corresponding values or types according to that. But it's not that much code, and it couldn't be encoded in the tuple pattern itself I think.

asterite: Or maybe it could be a separate variant, TupleWithDoubleDots or something like that that would have the two lists of patterns. I might try it tomorrow.

asterite: Done! It ended up being much, much better than before. A tuple with ".." has more complex logic than a tuple without ".." so now those two patterns are separate and represented differently. The one with ".." has `before` and `after`, like you suggested. To pack of all of that I introduced a new struct, `TupleWithDoubleDot`, that has several helper methods, for example to split a slice of elements into three (before, at dot, after), to iterate all patterns together with "..", etc. It all ends up being much easier to understand, I think. Thank you!

Because `TupleWithDoubleDot` has useful methods that I needed for AST patterns and HIR patterns, I made it generic over the pattern type. That means that a `HirPattern::TupleWithDoubleDot` will reference this type that exists in the "AST namespace". If that's not good we could always duplicate the code, but maybe it's fine.

jfecher: > If that's not good we could always duplicate the code, but maybe it's fine.

A generic type shared between multiple passes is fine. It's a technique used by a lot of compilers. Sometimes even the entire Ast is generic over additional types. E.g. your after-parsing Ast would be `Ast<()>` then you fill in types after type inference and it becomes `Ast<Type>`, etc.

TomAFrench (changes requested):
As discussed in 1:1 we're not going to add this feature pre-1.0. We can work around this issue by just breaking CI for a short period and fixing aztec packages.

PR #8513, opened 2025-05-14
chore(ssa): Add weighted edges to call graph, move callers and callees methods to call graph

vezenovm:
# Description

## Problem\*

No issue but part of the general efforts to consolidate into a centralized call graph structure. 

## Summary\*

The following was refactored in this PR:
- `compute_callees` is now a private method in the `call_graph` module and is only used for computing weighted call dependencies
- We have a new public `from_ssa_weighted` constructor that uses `compute_callees` internally to call `from_deps_weighted` which now adds edges with weights.
- The find the callers, callees, and times_called maps we parse the call graph directly.
- Moved the relevant tests to the call graph module

## Additional Context

Due to certain assertions in our topological sort `compute_bottom_up_order` I add some extra insertions to ensure that all nodes are in the maps returned from `callers`, `callees`, and `times_called`.

A next step could be to experiment with replacing `compute_bottom_up_order` with `petgraph::algo::toposort`. However, this is not entirely straightforward as the topological sort has some custom logic for exiting cycles and accumulating weights. It is still possible though that moving to `petgraph::algo::toposort` would be easier to maintain and simpler. However, as this sort should not be the cause of any correctness issues I am ok with it remaining.

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
> `rollup-base-private`	`20.26` s	`16.3` s	`1.24`

Need to check what is going on here, as this PR aims to improve compilation times. Probably can further reduce re-computation of the callers and callees.

vezenovm:
> `rollup-base-private`	`16.44` s	`15.78` s	`1.04`

Perhaps the runner was just busy as the time looks to be a bit flakey. We do run this circuit 5 times though, but looks like the times are fine.

jfecher (approved):

vezenovm:
Taking this off the queue temporarily so https://github.com/noir-lang/noir/pull/8568 can get in first.

PR #8526, opened 2025-05-15
fix: don't overflow when casting signed value to u128

### LINKED ISSUE #8494: panic "attempt to multiply with overflow" during ssa_gen
### Aim

Found while testing casting in the SSA interpreter:
```noir
fn main(x: i8) {
    let y: i8 = -1;

    println(x as u128);
    println(y as u128);
}
```

### Expected Behavior

Expected it to output what the expected result of `-1_i8 as u128` should be (255 or u128::MAX?)

Edit: Expected value is `u128::MAX` for `x as u128` but if we do `x as Field as u128` it is `255`. Seems confusing but the `u128::MAX` is Rust's behavior as well at least.

### Bug

```
Message:  attempt to multiply with overflow
Location: ~/.rustup/toolchains/1.85.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/num/mod.rs:328

                                ⋮ 8 frames hidden ⋮                               
   9: core::panicking::panic_const::panic_const_mul_overflow::hf1c8947532f8c86b
      at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:178
  10: core::num::<impl i128>::pow::h2e7e2ac4ac139d27
      at ~/.rustup/toolchains/1.85.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/num/int_macros.rs:<unknown line>
  11: noirc_evaluator::ssa::ssa_gen::context::FunctionContext::insert_safe_cast::h4521f55a674f9fe2
      at ~/Code/Noir/noir/compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs:<unknown line>
  12: noirc_evaluator::ssa::ssa_gen::context::FunctionContext::insert_safe_cast::h4521f55a674f9fe2
      at ~/Code/Noir/noir/compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs:<unknown line>
  13: noirc_evaluator::ssa::ssa_gen::<impl noirc_evaluator::ssa::ssa_gen::context::FunctionContext>::codegen_cast::hb9d04da57adf7035
      at ~/Code/Noir/noir/compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs:508
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8494

## Summary

A few issues here:
1. `2_u128.pow(...)` would overflow for 128
2. There's no way to have `2_u128.pow(128)` work (it overflows the max of u128) so I changed it to be one value less than that for both incoming and target sizes. That works because then the difference of those values is computed and ultimately used, so subtracting one from both sides would give the same result.
3. SSA ends up generating and using values of type `i128`, even if that type doesn't exist in the frontend. For that I had to relax the check in `validate_numeric_type`.

## Documentation

Check one:

# PR Checklist


jfecher (approved):
Nice

Issue #8457, opened 2025-05-12
Investigate brillig regression caused by #8454


TomAFrench:
https://github.com/noir-lang/noir/pull/8454

As part of fixing the visibility issues in the test suite, we replaced direct manipulation of `BoundedVec` storage with usage of `set_unchecked`. I'd expect that the compiler should be smart enough to optimize these down to be equivalent to directly writing into the storage array but it doesn't seem like it is.

vezenovm:
> As in these method calls aren't being inlined?

That is the suspicion, just need to confirm it.

asterite:
I started investigating this but noticed that the regressions happen with "inliner_zero" and "inliner_min", so maybe it's fine that those calls aren't inlined in those cases?

vezenovm:
> I started investigating this but noticed that the regressions happen with "inliner_zero" and "inliner_min", so maybe it's fine that those calls aren't inlined in those cases?

For `inliner_min` I expect it. But for `inliner_zero` I am surprised to not see the function inlined as all `set_unchecked` does is write to the `storage` array.

asterite:
- In the case where `set_unchecked` is called twice, the "cost" given to its `InlineInfo` is `-3`. Because it's less than zero, it's inlined
- When it's called twice, the "cost" given to it is `0`. Then it's not inlined.
- Each time the function is called, its cost increases by 3

Maybe then it makes sense, because if we have to inline it three times the final could would increase more than doing a call... though I guess the optimizer doesn't realize that maybe by inlining it will be able to further optimize things.

vezenovm:
> Maybe then it makes sense, because if we have to inline it three times the final could would increase more than doing a call...

It would potentially increase the byte code size more and that is what our calculation is deciding to avoid. In my opinion our inlining calculation is somewhat off as we are only increasing the cost for each time called. This is relevant for reducing bytecode size but not execution trace. Generally we want to inline all small and hot functions, and `set_unchecked` in this case is hot relative to the size of main. We will want to update our inlining to more clearly weight the size vs. performance trade off. Thanks for investigating, it does make sense with how the inline info computation is currently done.

vezenovm:
I'm going to close this issue for now as the regression was resolved by #8533. We can open a separate issue about improving the inliner cost analysis to better account for byte code size vs. execution trace.

Issue #8321, opened 2025-05-02
Consider propagating constant function call arguments


vezenovm:
### Problem

We have the following SSA on master for `to_bytes_consistent` with `--inliner-aggressiveness 0`:

<details><summary>Details</summary>
<p>

```
After Removing Paired rc_inc & rc_decs (1):
brillig(inline) fn main f0 {
  b0(v0: Field):
    v3 = call f1(v0) -> [u8; 31]
    v5 = call f1(Field 2040124) -> [u8; 31]
    inc_rc v5
    inc_rc v3
    inc_rc v5
    jmp b1(u32 0)
  b1(v1: u32):
    v8 = lt v1, u32 31
    jmpif v8 then: b2, else: b3
  b2():
    v9 = lt v1, u32 31
    constrain v9 == u1 1, "Index out of bounds"
    v11 = array_get v5, index v1 -> u8
    v12 = lt v1, u32 31
    constrain v12 == u1 1, "Index out of bounds"
    v13 = array_get v3, index v1 -> u8
    v14 = eq v11, v13
    constrain v11 == v13
    v16 = unchecked_add v1, u32 1
    jmp b1(v16)
  b3():
    return
}
brillig(inline) fn to_be_bytes f1 {
  b0(v0: Field):
    v30 = make_array [u8 1, u8 0, u8 0, u8 240, u8 147, u8 245, u8 225, u8 67, u8 145, u8 112, u8 185, u8 121, u8 72, u8 232, u8 51, u8 40, u8 93, u8 88, u8 129, u8 129, u8 182, u8 69, u8 80, u8 184, u8 41, u8 160, u8 49, u8 225, u8 114, u8 78, u8 100, u8 48] : [u8]
    v46 = make_array b"N must be less than or equal to modulus_le_bytes().len()"
    v49 = call to_be_radix(v0, u32 256) -> [u8; 31]
    return v49
}
brillig(inline) fn to_be_radix f2 {
  b0(v0: Field, v1: u32):
    v3 = call to_be_radix(v0, v1) -> [u8; 31]
    return v3
}

After Preprocessing Functions (1):
brillig(inline) fn main f0 {
  b0(v0: Field):
    v3 = call f1(v0) -> [u8; 31]
    v5 = call f1(Field 2040124) -> [u8; 31]
    inc_rc v5
    inc_rc v3
    inc_rc v5
    jmp b1(u32 0)
  b1(v1: u32):
    v8 = lt v1, u32 31
[...]

PR #8533, opened 2025-05-15
chore(inlining): Mark functions with <= 10 instructions and no control flow as inline always 

vezenovm:
# Description

## Problem\*

Experimenting with resolving #8457 

## Summary\*

We have a pass to inline functions with a single instruction (or none). In general, small functions without any control flow should be inlined as they should almost always lead to further optimizations. 

This PR changes `inline_functions_with_at_most_one_instruction` to `inline_simple_functions`.

As per the doc comments in this PR a simple function is defined as the following:
- Contains no more than 10 instructions
- The function only has a single block (e.g. no control flow or conditional branches)
- It is not marked with the no predicates inline type

## Additional Context

See https://github.com/noir-lang/noir/pull/8533#issuecomment-2884547837 to understand the regressions for `to_bytes_consistent_inliner_min`.

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
So we have this size regression:
> **to_bytes_consistent_inliner_min**	+53 ❌	**+76.81%**
And this execution trace regression:
> **to_bytes_consistent_inliner_min**	510 (+33)	**+6.92%**

We have this final SSA on master:
```
  b0(v0: Field):
    v3 = call f1(v0) -> [u8; 31]
    v5 = call f1(Field 2040124) -> [u8; 31]
    jmp b1(u32 0)
  b1(v1: u32):
    v8 = lt v1, u32 31
    jmpif v8 then: b2, else: b3
  b2():
    v9 = array_get v5, index v1 -> u8
    v10 = array_get v3, index v1 -> u8
    constrain v9 == v10
    v12 = unchecked_add v1, u32 1
    jmp b1(v12)
  b3():
    return
}
brillig(inline) predicate_pure fn to_be_bytes f1 {
  b0(v0: Field):
    v3 = call to_be_radix(v0, u32 256) -> [u8; 31]
    return v3
}
```
And this SSA on this PR:
```
  b0(v0: Field):
    v4 = call to_be_radix(v0, u32 256) -> [u8; 31]
    v9 = make_array [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 31, u8 33, u8 60] : [u8; 31]
    jmp b1(u32 0)
  b1(v1: u32):
    v12 = lt v1, u32 31
    jmpif v12 then: b2, else: b3
  b2():
    v13 = array_get v9, index v1 -> u8
    v14 = array_get v4, index v1 -> u8
    constrain v13 == v14
    v16 = unchecked_add v1, u32 1
    jmp b1(v16)
  b3():
    return
}
```

The second call to f1 is with a constant input. This gets simplified to a single make array. As `to_be_radix` is a single Brillig opcode the simplified SSA actually leads to a larger bytecode size as well as execution trace. This was captured in an issue already from a similar case here https://github.com/noir-lang/noir/issues/8321.

Due to the other benefits in this PR I think it is ok to eat this edge case. Especially as it is only occurring for `_inliner_min` which is also an unlikely aggressiveness developers would use in practice (an aggressiveness at zero is the more useful setting).

compiler/noirc_evaluator/src/ssa/opt/inline_functions_with_at_most_one_instruction.rs
```diff
@@ -1,6 +1,11 @@
-//! This modules defines an SSA pass that inlines calls to functions with at most one instruction.
+//! This modules defines an SSA pass that inlines calls to simple functions.
```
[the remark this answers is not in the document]
vezenovm: done

asterite (approved):
Looks good!

jfecher (approved):
We should wait to merge this until late Friday since it touches audit code

TomAFrench (approved):
LGTM, will defer to @michaeljklein or @aakoshh when it comes to hitting merge.

compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -141,7 +141,7 @@ pub fn primary_passes(options: &SsaEvaluatorOptions) -> Vec<SsaPass> {
         SsaPass::new(Ssa::remove_unreachable_functions, "Removing Unreachable Functions"),
         SsaPass::new(Ssa::defunctionalize, "Defunctionalization"),
         SsaPass::new(
-            Ssa::inline_functions_with_at_most_one_instruction,
+            Ssa::inline_simple_functions,
             "Inlining functions with at most one instruction",
```
aakoshh, line 145: Should the message be updated?

compiler/noirc_evaluator/src/ssa/opt/inline_simple_functions.rs
```diff
@@ -1,6 +1,11 @@
-//! This modules defines an SSA pass that inlines calls to functions with at most one instruction.
+//! This modules defines an SSA pass that inlines calls to simple functions.
 //! That is, the contents of the called function is put directly into the caller's body.
 //! Functions are still restricted to not be inlined if they are recursive or marked with no predicates.
+//!
+//! A simple function is defined as the following:
+//! - Contains no more than 10 instructions
+//! - The function only has a single block (e.g. no control flow or conditional branches)
+//! - It is not marked with the no predicates [inline type][noirc_frontend::monomorphization::ast::InlineType]
```
aakoshh, line 8: It would be great to include an explainer about why "no predicates" are treated separately, or a link to any existing documentation about it. There is some description on `InlineType::NoPredicate`, but it only states that it's inlined later, without explaining why.

compiler/noirc_evaluator/src/ssa/opt/inline_simple_functions.rs
```diff
@@ -34,8 +39,8 @@ impl Ssa {
                 return false;
             }
 
-            // Only inline functions with 0 or 1 instructions
-            if instructions.len() > 1 {
+            // Only inline functions with 10 or less instructions
+            if instructions.len() > 10 {
```
aakoshh, line 43: Should we make this configurable in the `SsaEvaluatorOptions`, seeing that the inliner aggressiveness is configurable as well? It would avoid a magic number in the code.
Also leaving some hints about how this sweet spot has been determined (even if it's just an expert guess) would help auditors and future devs.

compiler/noirc_evaluator/src/ssa/opt/inline_simple_functions.rs
```diff
@@ -124,14 +129,22 @@ mod test {
...
-            v3 = add v1, v2
-            return v3
+            v3 = call f1(v0) -> Field
+            v4 = call f1(v0) -> Field
+            v5 = call f1(v0) -> Field
+            v6 = call f1(v0) -> Field
+            v7 = call f1(v0) -> Field
+            v8 = call f1(v0) -> Field
+            v9 = call f1(v0) -> Field
+            v10 = call f1(v0) -> Field
```
aakoshh, line 139: Can you leave a comment about why these unused calls have been added? Previously it was 1 per argument of `add`, which was more intuitive. For example it looks like they will all be removed from `bar` anyway, so, does matter how many were added?

aakoshh (commented):
I have no problem with merging stuff into the audited code, I'm just writing a post-audit report at this point, but I think some extra comments and magic-number outsourcing could be helpful to explain in the code why we went from 1 to 10 in particular.

compiler/noirc_evaluator/src/ssa/opt/inline_simple_functions.rs
vezenovm: I will make some more comments about this number and move it into a constant global. We could make this configurable, although I'm a little hesitant as this pass is a bit of a workaround of our inliner cost analysis and aggressiveness setting not already marking these functions to be inlined. Improving the inliner cost analysis is a larger task in its own so I think have this kind of inlining override is ok for now though.

compiler/noirc_evaluator/src/ssa/opt/inline_simple_functions.rs
aakoshh: Right, I just saw they were eliminated from `f2` completely. I guess they could be any instruction whatsoever, it was just slightly suspect that it's a call to `f1` itself, like, not clear if we're trying to express that "if you try to inline f1 more than 10 times then X happens", rather than "if f2 has more than 10 instructions it's not inlined into f0"

compiler/noirc_evaluator/src/ssa.rs
vezenovm: done

compiler/noirc_evaluator/src/ssa/opt/inline_simple_functions.rs
vezenovm: done

compiler/noirc_evaluator/src/ssa/opt/inline_simple_functions.rs
aakoshh: Awesome, thanks!

aakoshh (approved):
👍

Issue #8165, opened 2025-04-22
Generic impl not found when searching for impl on a concrete type


nventuro:
### Aim

I'm trying to write a derive macro for a trait which recursively invokes the trait on each struct field (not unlike `Eq` requires for all fields to also be `Eq` and then calls `eq`). Some base cases must be defined by hand (e.g. the impl for `Field`). One of these impls is generic for an array where all elements implement the trait:

```noir
#[derive_via(derive_sum_all_values)]
pub trait SumAllValues {
    fn sum_all_values(self) -> Field;
}

impl SumAllValues for Field {
    fn sum_all_values(self) -> Field {
        self
    }
}

impl<T, let N: u32> SumAllValues for [T; N]
where
    T: SumAllValues,
{
    fn sum_all_values(self) -> Field {
        let mut sum = 0;
        
        for i in 0..N {
            sum += self[i].sum_all_values();
        }

        sum
    }
}
```

Now I want to write the `derive` macro so that the impl is autogenerated for any struct where all fields impl the trait. In order to avoid forcing the user to import the trait themselves to resolve ambiguity, I find the trait method in the attribute itself, resolving it on the macro and not in the returned quote:

```noir
comptime fn get_trait_impl_method(
    typ: Type,
    target_trait: Quoted,
    target_method: Quoted,
) -> TypedExpr {
    let trait_constraint = target_trait.as_trait_constraint();
    typ
        .get_trait_impl(trait_constraint)
        .expect(f"Could not find impl for {target_trait} for type {typ}")
        .methods()
        .filter(|m| m.name() == target_method)[0]
        .as_typed_expr()
}
```

This works fine for structs with `Field`, but fails if my struct has a `Field` array. The reason seems to be that the finds a generic impl over `T` with length `N`, but instead gets my concrete array with some length and type `Field`.

### Bug

Below is the full code:

```noir
#[derive_via(derive_sum_all_values)]
pub trait SumAllValues {
    fn sum_all_values(self) -> Field;
}

impl SumAllValues for Field {
    fn sum_all_values(self) -> Field {
        self
    }
}

impl<T, let N: u32> SumAllValues for [T; N]
where
    T: SumAllValues,
{
    fn sum_all_values(self) -> Field {
        let mut sum = 0;
        
        for i in 0..N {
            sum += self[i].sum_all_values();
        }

        sum
    }
}

comptime fn get_trait_impl_method(
    typ: Type,
    target_trait: Quoted,
    target_method: Quoted,
[...]

PR #8537, opened 2025-05-15
feat: Implement $crate

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/8165

## Summary\*

Implements `$crate` which when used within a `quote` will always resolve to the crate the `quote` was originally made in.

## Additional Context

This is done by having a new token kind to hold the CrateId and allowing this as a new PathKind as well. The actual code is fairly small but it touches the stdlib so there are a lot of .snap changes.

## Documentation\*

Check one:

# PR Checklist\*


tooling/nargo_fmt/src/formatter/use_tree_merge.rs
```diff
@@ -268,6 +268,11 @@ fn merge_imports_in_tree(imports: Vec<UseTree>, mut tree: &mut ImportTree) {
             PathKind::Super => tree.insert(Segment::Super),
             PathKind::Dep => tree.insert(Segment::Dep),
             PathKind::Plain => &mut tree,
+            PathKind::Resolved(_) => {
+                // TODO: Don't know whether to insert $crate or <crate name> here, or what the crate name is
+                //       Probably better to guess instead of panicking in the LS though.
```
asterite, line 273: This case is probably never hit as the formatter runs on a parsed AST but without doing semantic analysis. Then I think `Resolved` will never be created.

tooling/nargo_fmt/src/formatter/path.rs
```diff
@@ -23,6 +23,10 @@ impl Formatter<'_> {
                 self.write_keyword(Keyword::Super);
                 self.write_token(Token::DoubleColon);
             }
+            PathKind::Resolved(_) => {
+                self.write_token(Token::DollarSign);
+                self.write_keyword(Keyword::Crate);
+            }
```
asterite, line 29: Like the other comment, we could panic here.

asterite (approved):
Looks good!

I wonder if there's a solution to referencing types in the current crate's dependencies. Otherwise I think the workaround is doing `use dependency;`, then `$crate::dependency`. But with `$dep::dependency` it could be simpler (not sure if there's something like that in Rust either, I couldn't find anything)

noir_stdlib/src/default.nr
```diff
@@ -8,7 +8,7 @@ pub trait Default {
 // docs:end:default-trait
 
 comptime fn derive_default(s: TypeDefinition) -> Quoted {
-    let name = quote { Default };
+    let name = quote { $crate::default::Default };
     let signature = quote { fn default() -> Self };
     let for_each_field = |name| quote { $name: Default::default() };
```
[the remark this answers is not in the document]
jfecher: No, this is the trait name not the method name.
I suppose if we're fully resolving this we could do the same for the method name on the `for_each_field` line (or is that what you meant?)

tooling/nargo_fmt/src/formatter/use_tree_merge.rs
jfecher: Sounds good, I'll edit the comment. Do you think we should still avoid a panic or would it be better to have a panic to verify these assumptions?

jfecher:
> I wonder if there's a solution to referencing types in the current crate's dependencies. Otherwise I think the workaround is doing use dependency;, then $crate::dependency. But with $dep::dependency it could be simpler (not sure if there's something like that in Rust either, I couldn't find anything)

Yeah, there is the workaround you mentioned. Not sure what rust does here either. Another option is to re-export only the function(s) you care about `pub use dependency::foo` then you can inject `$crate::my_path::foo`.

noir_stdlib/src/default.nr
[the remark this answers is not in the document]
asterite: Right, I meant the one in line 13.

tooling/nargo_fmt/src/formatter/use_tree_merge.rs
asterite: I think a panic is good. For example when a formatter encounters an interned pattern, it panics too.

PR #8546, opened 2025-05-16
fix(licm): Check whether the loop is executed when hoisting with a predicate

### LINKED ISSUE #8319: Brillig: Assertion failed with minimal inliner aggressiveness only
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8299

The following program passes with the default inliner aggressiveness, but fails with the minimum:

```rust
unconstrained fn main() {
    func_4(([982714762, 1608230664, 671108204]))
}

unconstrained fn func_4(a: [u32; 3]) {
    {
        let mut idx_b = 0;
        loop {
            if (idx_b == 6) {
                break
            } else {
                idx_b = (idx_b + 1);
                {
                    let mut idx_c = 0;
                    loop {
                        if (idx_c == 7) {
                            break
                        } else {
                            idx_c = (idx_c + 1);
                            for idx_d in ((a[2] / 1193346642) % 4)..(a[1] % 4) {
                                {
                                    let mut idx_g = 0;
                                    loop {
                                        if (idx_g == 3) {
                                            break
                                        } else {
                                            idx_g = (idx_g + 1);
                                            let mut h = (a[0] % idx_d);
                                        }
                                    }
                                };
                                break;
                            }
                        }
                    }
                };
            }
        }
    }
}
```

### Expected Behavior

I expected that the behaviour would be the same. 

I thought maybe https://github.com/noir-lang/noir/pull/8303 causes this, because the failure is happening during the calculation of an unused variable, but it's strange that the loop in which the failure happens should never actually be executed. That is something the compiler wouldn't know unless it inlines `func_4` into `main`. 

### Bug

```console
$ RUST_BACKTRACE=1 cargo run -q -p nargo_cli -- execute --skip-underconstrained-check --skip-brillig-constraints-check --silence-warnings --force --inliner-aggressiveness -9223372036854775808
$ RUST_BACKTRACE=1 cargo run -q -p nargo_cli -- execute --skip-underconstrained-check --skip-brillig-constraints-check --silence-warnings --force --inliner-aggressiveness -9223372036854775808
error: Assertion failed: 'Attempted to divide by zero'
   ┌─ src/main.nr:28:58
   │
28 │                                             let mut h = (a[0] % idx_d);
   │                                                          ------------
   │
   = Call stack:
     1. src/main.nr:2:5
     2. src/main.nr:28:58

Failed assertion
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+904693b1ae2eb71621d99ce1608a9d220fa5c73d (git version hash: 96dc632eeac458657c6a6d95e6e5f2f3891802b9, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #8521: Brillig divides by zero in a loop that never runs
### Aim

Discovered running `cargo test -p noir_ast_fuzzer_fuzz min_vs_full`. 

This could be the same issue as https://github.com/noir-lang/noir/issues/8319 but it's simpler. 

The following program behaves different when executed with `--minimal-ssa`:

```rust
unconstrained fn main() -> pub str<3> {
    let mut idx_b = 0;
    while true {
        if (idx_b == 0) {
            break
        } else {
            idx_b = (idx_b + 1);
            let mut c = (209991686817458660262701076546592445127 / (false as Field));
        }
    }
    "SQF"
}
```

### Expected Behavior

It should return "SQF" in all cases, because we immediately break out of the loop.

### Bug

```console
$ cargo run -q -p nargo_cli -- execute --force --silence-warnings --minimal-ssa; cargo run -q -p nargo_cli -- execute --force --silence-warnings
[regression_fuzz_testing] Circuit output: String("SQF")
error: Assertion failed: 'Attempted to divide by zero'
  ┌─ src/main.nr:8:26
  │
8 │             let mut c = (209991686817458660262701076546592445127 / (false as Field));
  │                          ----------------------------------------------------------
  │
  = Call stack:
    1. src/main.nr:8:26

Failed assertion
```

### To Reproduce

See above. 

My suspicion is that `c` is hoisted out of the loop as an invariant, but it should not be executed at all because the loop never runs.

Indeed:
```
After Purity Analysis (1):
  b0():
    v0 = allocate -> &mut Field
    store Field 0 at v0
    jmp b1()
  b1():
    v2 = load v0 -> Field
    v3 = eq v2, Field 0
    jmpif v3 then: b2, else: b3
  b2():
    v13 = make_array b"SQF"
    return v13
  b3():
    v4 = load v0 -> Field
    v6 = add v4, Field 1
    store v6 at v0
    v8 = div Field 209991686817458660262701076546592445127, Field 0
    v9 = allocate -> &mut Field
    jmp b1()
}

After Loop Invariant Code Motion (1):
  b0():
    v0 = allocate -> &mut Field
    store Field 0 at v0
    v3 = div Field 209991686817458660262701076546592445127, Field 0
    jmp b1()
  b1():
    v4 = load v0 -> Field
    v5 = eq v4, Field 0
    jmpif v5 then: b2, else: b3
  b2():
    v13 = make_array b"SQF"
    return v13
  b3():
    v6 = load v0 -> Field
    v8 = add v6, Field 1
    store v8 at v0
    v9 = allocate -> &mut Field
    jmp b1()
}

```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+e796dfd67726cbc28eb9991782533b211025928d (git version hash: bb532cc1c0e314c13a7f8a79610d3592d7ba3d4d, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #8319 
Resolves #8521 

## Summary\*

We were not checking whether the loop body is guaranteed to execute when hoisting with a predicate. This is logic we already were doing when simplifying loop bounds for instructions such as constrain and pure with predicate calls.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -2222,4 +2238,47 @@ mod control_dependence {
...
+            v3 = load v1 -> Field
+            v4 = eq v3, Field 0
+            jmpif v4 then: b2, else: b3
+          b2():
+            return
+          b3():
+            v6 = div u32 5, v0
+            jmp b1()
+        }
+        ");
```
aakoshh, line 2282:
```suggestion
        // We expect the SSA to be unchanged
        assert_normalized_ssa_equals(ssa, src);
```
Just so it's obviously the same.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -403,24 +413,29 @@ impl<'f> LoopInvariantContext<'f> {
             }
             Binary(binary) => self.can_evaluate_binary_op(binary),
```
aakoshh, line 414: So it wasn't the `div` being treated as a binary that can be hoisted?

vezenovm: This was the issue but the error was occurring earlier inside of `can_hoist_invariant`. `can_be_hoisted` were `hoist_with_predicate == true` needed the extra check that the loop body does not execute, we were only checking that the instruction was not in a control dependent block. `can_evaluate_binary_op` would return `false` for the `div` from the regression as we cannot ensure that the divisor cannot be zero.

PR #8550, opened 2025-05-16
feat(performance): brillig array set optimization

asterite:
# Description

## Problem

Follow up to #8536

## Summary

Similar logic to the one of ArraySet.

While doing this I introduced a few refactors, for example just having one `insert_array_get` and `insert_array_set` on a `FunctionBuilder`.

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -838,12 +832,21 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
                     dfg,
                 );
 
+                let offseted = if dfg.is_constant(*index) {
```
vezenovm, line 835:
```suggestion
                let has_offset = if dfg.is_constant(*index) {
```
Let's change this from `offseted` which triggers cspell

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -838,12 +832,21 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
...
+                    // For constant indices it must be the case that they have been offseted during SSA
```
vezenovm, line 836:
```suggestion
                    // For constant indices it must be the case that they have been offset during SSA
```

compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -210,7 +210,7 @@ pub fn primary_passes(options: &SsaEvaluatorOptions) -> Vec<SsaPass> {
         // We can safely place the pass before DIE as that pass only removes instructions.
         // We also need DIE's tracking of used globals in case the array get transformations
         // end up using an existing constant from the globals space.
-        SsaPass::new(Ssa::brillig_array_gets, "Brillig Array Get Optimizations"),
+        SsaPass::new(Ssa::brillig_array_get_and_set, "Brillig Array Get Optimizations"),
```
vezenovm, line 213:
```suggestion
        SsaPass::new(Ssa::brillig_array_get_and_set, "Brillig Array Get and Set Optimizations"),
```
nit

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -1110,6 +1113,32 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
...
+            (items_pointer, true)
+        };
+
+        self.brillig_context.codegen_load_with_offset(
+            items_pointer,
+            index_variable,
+            destination_variable.extract_register(),
+        );
+
+        if deallocate {
```
vezenovm, line 1137: looks like we could just re-use the `offseted` flag here

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
```diff
@@ -1164,7 +1201,9 @@ impl<'block, Registers: RegisterAllocator> BrilligBlock<'block, Registers> {
             );
         }
 
-        self.brillig_context.deallocate_register(items_pointer);
+        if deallocate {
```
vezenovm, line 1204: same thing about reusing the offset flag

vezenovm (commented):
Looks good just some nits

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
asterite: Done!

compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs
asterite: Done!

compiler/noirc_evaluator/src/ssa.rs
asterite: Good catch!

asterite:
I looked at the regression in `fold_call_witness_condition`, which isn't a big deal (just one more instruction). It seems there's a new constant needed now, with value `2`, where previously there was no such constant. Maybe because of the offset it appears now. But I guess in most cases this PR will make things better.

vezenovm:
> I looked at the regression in `fold_call_witness_condition`, which isn't a big deal (just one more instruction). It seems there's a new constant needed now, with value `2`, where previously there was no such constant. Maybe because of the offset it appears now. But I guess in most cases this PR will make things better.

Yeah I noticed the same thing. For most cases this optimization should make things better and this regression is tiny so it is fine.

compiler/noirc_evaluator/src/ssa/ir/instruction.rs
```diff
@@ -652,7 +652,7 @@ impl Instruction {
     }
 }
 
-/// Determines whether an ArrayGet or ArraySet index has been offseted by a given value.
+/// Determines whether an ArrayGet or ArraySet index has been shifted by a given value.
 /// Offsets are set during `crate::ssa::opt::brillig_array_gets` for brillig arrays
 /// and vectors with constant indicces.
```
vezenovm, line 657: nit: indicces -> indices

vezenovm (approved):

Issue #8545, opened 2025-05-16
Comptime gen interpreter stuck in an infinite loop


benesjan:
### Aim

I was trying to use derive_serialize in more places in aztec-packages but the code suddenly failed to compile. Managed to isolate it and the problematic code can be cloned from https://github.com/benesjan/noir_slow

I know it's an issue with code generation because I put a printout into the derive_serialize function and I know it does not get executed for all the funcs.

### Expected Behavior

It compiles and everyone's happy

### Bug

It doesn't compile and everyone's sad

### To Reproduce

1. git clone https://github.com/benesjan/noir_slow.git
2. cd noir_slow
3. nargo compile

You should get the thing to hang:

<img width="565" alt="Image" src="https://github.com/user-attachments/assets/4869aa35-5299-4104-8b23-beb1c1d2cd33" />

### Workaround

None

### Project Impact

None

### Nargo Version

nargo version = 1.0.0-beta.5 noirc version = 1.0.0-beta.5+0000000000000000000000000000000000000000

### Would you like to submit a PR for this Issue?

None

TomAFrench:
```noir
use std::meta::derive;

pub(crate) comptime fn derive_serialize(s: TypeDefinition) -> Quoted {
    let typ = s.as_type();
    std::println("derive_serialize");
    std::println(typ);
    let nested_struct = typ.as_data_type().unwrap();
    let params = nested_struct.0.fields(nested_struct.1);

    // The following will give us [<type_of_struct_member_1 as Serialize>::N, <type_of_struct_member_2 as Serialize>::N, ...]
    let components_of_definition_of_n = params.map(|(_, param_type): (Quoted, Type)| {
        quote {
            <$param_type as Serialize>::N
        }
    });

    // The following will give us <type_of_struct_member_1 as Serialize>::N + <type_of_struct_member_2 as Serialize>::N + ...
    let right_hand_side_of_definition_of_n = components_of_definition_of_n.join(quote {+});

    let array_of_quotes_serializing_each_struct_member = params
        .map(|(param_name, param_type): (Quoted, Type)| {
            quote {
            let serialized_member = self.$param_name.serialize();
            let serialized_member_len = <$param_type as Serialize>::N;
            for i in 0..serialized_member_len {
                result[i + offset] = serialized_member[i];
            }
            offset += serialized_member_len;
        }
        });

    let serialization_of_struct_members =
        array_of_quotes_serializing_each_struct_member.join(quote {});

    quote {
[...]

PR #8575, opened 2025-05-19
chore(opt): Memoize `Type::canonicalize`

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/8545

## Summary\*

Optimizes `Type::canonicalize` by remembering the returned result which prevents this function from being quadratic in input size since we're no longer re-doing work. This also required a small fix to keep the `arithmetic_generics` test working: we can't do `if let Type::InfixExpr(..) = ...` since that doesn't match on `Type::CheckedCast` variants which are meant to be transparent. With this fix that was previously causing canonicalize to leave some terms unsimplified.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
+++ b/compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
@@ -1,3 +1,4 @@
+use fxhash::FxHashMap as HashMap;
 use std::collections::BTreeMap;
 
 use acvm::{AcirField, FieldElement};
--- a/compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
+++ b/compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
@@ -1,3 +1,4 @@
+use fxhash::FxHashMap as HashMap;
 use std::collections::BTreeMap;
 
 use acvm::{AcirField, FieldElement};
@@ -19,7 +20,7 @@ impl Type {
         match self.follow_bindings() {
             Type::CheckedCast { from, to } => Type::CheckedCast {
                 from: Box::new(from.canonicalize_checked()),
-                to: Box::new(to.canonicalize_unchecked()),
+                to: Box::new(to.canonicalize_unchecked(&mut HashMap::default())),
             },
 
             other => {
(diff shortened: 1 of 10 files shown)


asterite (approved):
Nice!

I thought the solution could be memoizing but I didn't know if it was possible because types don't have IDs, though they do have equality and has so I guess it works.

It might be nice to add `bench-show` to see if there's any noticeable improvements in existing projects (just out of curiosity).

asterite:
Um, I approved but I see there are some regressions, so maybe this broke something. In any case the changes look good.

jfecher:
> Um, I approved but I see there are some regressions, so maybe this broke something. In any case the changes look good.

Yeah, quite large ones :sweat_smile:. I'll look into those before merging. Not sure what they are currently. The main difference I can think of is that some `Type::CheckedCast` may be different but those should be checked at monomorphization time rather than runtime IIRC.

jfecher:
> Review the following changes in direct dependencies. Learn more about [Socket for GitHub](https://socket.dev/?utm_medium=gh).

This makes me think it is just because I didn't merge master originally.

compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
```diff
@@ -269,17 +323,16 @@ impl Type {
...
+    ) -> Option<(Type, BinaryTypeOperator, FieldElement, FieldElement)> {
         let kind = lhs.infix_kind(rhs);
         let dummy_location = Location::dummy();
         let rhs = rhs.evaluate_to_field_element(&kind, dummy_location).ok()?;
 
-        let Type::InfixExpr(l_type, l_op, l_rhs, _) = lhs.follow_bindings() else {
-            return None;
-        };
+        let lhs = lhs.follow_bindings();
+        let (l_type, l_op, l_rhs, _) = lhs.as_infix_expr()?;
```
michaeljklein, line 332: Is it possible that recursing to find an `InfixExpr` here and/or in `try_unify_by_moving_constant_terms` ends up removing `CheckedCast`'s?

michaeljklein (changes requested):

compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
jfecher: Yes - although without doing so we were failing in the `arithmetic_generics` test from failing to catch this case for results which were now cached to include CheckedCast where they previously dropped it

jfecher:
Closing this - https://github.com/noir-lang/noir/pull/8826 should be a valid substitute.
The original issue is compiling faster but has errors during compilation (already captured in https://github.com/noir-lang/noir/issues/8797). We can revisit/close it once the unrelated errors are fixed.

PR #8578, opened 2025-05-19
chore(refactor): Switch unreachable function removal to use centralized call graph

vezenovm:
# Description

## Problem\*

Part of https://github.com/noir-lang/noir/issues/7228

## Summary\*

I just noticed that `remove_unreachable` was also building its own call graph to determine transitively reachable functions. We can reduce complexity here by re-using our `CallGraph` and off the shelf DFS. 

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/remove_unreachable.rs
```diff
@@ -33,66 +34,35 @@ use crate::ssa::{
...
+            is_entry_point.then_some(id)
+        });
+
+        // Build call graph dependencies using this passes definition of reachability.
+        let dependencies = self
+            .functions
+            .iter()
+            .map(|(&id, func)| {
+                let used = used_functions(func);
+                (id, used)
```
asterite, line 52: nit: if it's written like `(id, used_functions(func))` then this declaration is reduced to two lines.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable.rs
```diff
@@ -33,66 +34,35 @@ use crate::ssa::{
...
+            })
+            .collect();
```
asterite, line 54: I think this is the result:

```suggestion
        let dependencies =
            self.functions.iter().map(|(&id, func)| (id, used_functions(func))).collect();
```

PR #8580, opened 2025-05-19
feat(experimental): keep inlining small functions

asterite:
# Description

## Problem

In a previous PR @aakoshh noted that `inline_simple_functions` just runs once and it's not meant to keep inlining functions as long as it's possible (for example, inlining a function could turn a parent function into a "simple" function).

## Summary

Here just to try this out I put another "inline_simple_functions" pass just to see if it improves things. If so, we could consider doing this properly.

## Documentation\*

Check one:

# PR Checklist\*


asterite:
Probably not worth it.

vezenovm:
> Probably not worth it.

I don't see much much harm in making a feedback loop to inline small functions until there are no more. But as you said the benefits looks to be not worth it at the moment.

PR #8591, opened 2025-05-20
chore(fuzz): Build the dictionary from the SSA

### LINKED ISSUE #8467: Build fuzzing dictionary from SSA
Similar to how the dictionary for ABI input generation is created from the bytecode, look for interesting values in the SSA instructions.

aakoshh:
# Description

## Problem\*

Resolves #8467 

## Summary\*

Pull the constants from the SSA to build the fuzzing input dictionary.

## Documentation\*

Check one:

# PR Checklist\*


michaeljklein (approved):

compiler/noirc_evaluator/src/ssa/ir/function.rs
```diff
@@ -217,6 +218,17 @@ impl Function {
             })
             .sum()
     }
+
+    /// Iterate over the numeric constants in the function.
+    pub fn constants(&self) -> impl Iterator<Item = (&FieldElement, &NumericType)> {
```
vezenovm, line 223: We store a `constants: HashMap<(FieldElement, NumericType), ValueId>` field in the dfg that is used for optimizing the creation of constants. We could probably just expose fetching this field or at least its keys rather than iterating through all the values.

compiler/noirc_evaluator/src/ssa/ir/function.rs
```diff
@@ -217,6 +218,17 @@ impl Function {
...
+        self.dfg.values_iter().filter_map(|(_, value)| {
+            if let Value::NumericConstant { constant, typ } = value {
```
vezenovm, line 225: This looks like it would be missing global constants?

compiler/noirc_evaluator/src/ssa/ir/function.rs
aakoshh: Yeah, I wasn't sure what would the team prefer. Everything in the `DataFlowGraph` is `pub(crate)`, as well as the `dfg` field of `Function`. Instead of starting to make this field or that field public, I thought I'd add a helper function to do exactly what I needed for now, to avoid all arguments: can't expose less than this. 

In general I remember `foo.bar.qux.baz` being an anti pattern as well, so it would at least need to be a read-only `.dfg()` method, etc. 

Open to suggestions if there is consensus over how much to expose.

compiler/noirc_evaluator/src/ssa/ir/function.rs
aakoshh: Good point! Chained them in.

PR #8593, opened 2025-05-20
fix(licm): Account for nested loops being control dependent when analyzing outer loops

### LINKED ISSUE #8586: Loop invariant motion: failing constraint hoisted out of `for`/`if`/`for`
### Aim

Discovered while testing https://github.com/noir-lang/noir/pull/8584

The following program executes with the `--minimal-ssa` flag but fails with the defaults:

```rust
fn main(a: pub bool) {
    for _ in 0..1 {
        if a {
            for _ in 0..1 {
                let _ = (1 / (a as Field));
            }
        };
    }
}
```

Input:
```
a = false
```

### Expected Behavior

The execution should succeed, because when `a` is `false` it should not run `1 / (false as Field)`. 

### Bug

```console
$ cargo run -q -p nargo_cli -- execute --silence-warnings 
error: Failed constraint
  ┌─ src/main.nr:5:26
  │
5 │                 let _ = (1 / (a as Field));
  │                          ----------------
  │
  = Call stack:
    1. src/main.nr:5:26
$ cargo run -q -p nargo_cli -- execute --silence-warnings  --minimal-ssa
```

### To Reproduce

See above. 

The SSA shows that the division is hoisted out of the loop:

```
After Purity Analysis (1):
  b0(v0: u1):
    jmp b1(u32 0)
  b1(v1: u32):
    v4 = eq v1, u32 0
    jmpif v4 then: b2, else: b3
  b2():
    jmpif v0 then: b4, else: b5
  b3():
    return
  b4():
    jmp b6(u32 0)
  b5():
    v7 = unchecked_add v1, u32 1
    jmp b1(v7)
  b6(v2: u32):
    v5 = eq v2, u32 0
    jmpif v5 then: b7, else: b8
  b7():
    v8 = cast v0 as Field
    v10 = div Field 1, v8
    v11 = unchecked_add v2, u32 1
    jmp b6(v11)
  b8():
    jmp b5()
}

After Loop Invariant Code Motion (1):
  b0(v0: u1):
    v3 = cast v0 as Field
    v5 = div Field 1, v3
    jmp b1(u32 0)
  b1(v1: u32):
    v7 = eq v1, u32 0
    jmpif v7 then: b2, else: b3
  b2():
    jmpif v0 then: b4, else: b5
  b3():
    return
  b4():
    jmp b6(u32 0)
  b5():
    v10 = unchecked_add v1, u32 1
    jmp b1(v10)
  b6(v2: u32):
    v8 = eq v2, u32 0
    jmpif v8 then: b7, else: b8
  b7():
    v11 = unchecked_add v2, u32 1
    jmp b6(v11)
  b8():
    jmp b5()
}
```

A recent fix to this pass was in https://github.com/noir-lang/noir/pull/8546 

The failure disappears if either of the `for` are commented out, that is, both `for/if` and `if/for` work as expected.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+1be0da333770ee5a026a9952783b16c39b9ab7b4 (git version hash: 20c5dac61923d7b65cbd7f52c3917a4fbe0153fb, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
…ks control dependence

# Description

## Problem\*

Resolves #8586 

The bug was that we were not appropriately accounting for a nested loop being control dependent. Taking the issue snippet:
```noir
fn main(a: pub bool) {
    for _ in 0..1 {
        if a {
            for _ in 0..1 {
                let _ = (1 / (a as Field));
            }
        };
    }
}
```
For completeness this has the following SSA before LICM:
<p>

```
          b0(v0: u1):
            jmp b1(u32 0)
          b1(v1: u32):
            v4 = eq v1, u32 0
            jmpif v4 then: b2, else: b3
          b2():
            jmpif v0 then: b4, else: b5
          b3():
            return
          b4():
            jmp b6(u32 0)
          b5():
            v7 = unchecked_add v1, u32 1
            jmp b1(v7)
          b6(v2: u32):
            v5 = eq v2, u32 0
            jmpif v5 then: b7, else: b8
          b7():
            v8 = cast v0 as Field
            v10 = div Field 1, v8
            v11 = unchecked_add v2, u32 1
            jmp b6(v11)
          b8():
            jmp b5()
        }
```

</p>
</details> 

When checking whether inner loop block (b7) is control dependent, we will get `false`. This is because we are not accounting for transitive control dependence. When initially adding control dependence to LICM I felt doing so (which would most likely require a full control dependence graph (CDG)) was a bit overkill as an initial control dependent LICM could simply check for control dependence after the loop header. 

## Summary\*

A full CDG I think would still be overkill at the moment and I simply expanded our control dependence to check for whether the block we are processing is within a nested loop that is control dependent.

Changes:
1. We now pass all the remaining loops we have yet to process
2. When checking whether a block is control dependent we now iterate through all loops, skipping potential nested loops where the block being checked for control dependence is not in the nested loop (this simply means we do not have a relevant nested loop).
3. If we find a nested loop we run the same check where we look at whether the current block is control dependent on any of the predecessors between itself and the outer loop header. Except now, we are checking the control dependence on the nested loop header rather than the current block.
4. If we have a control dependent nested loop (like in the issue snippet), that block will be marked as control dependent for the outer loop LICM. Once we get to the inner loop LICM the control dependence checks will reset, and if appropriate we can still hoist out of the inner loop.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -40,7 +40,7 @@
 //! ```
 //! Reversing this for post-dominance we can see that the conditions for control dependence
 //! are the same as those for post-dominance frontiers.
-//! Thus, we rewrite our control dependence condition as Y is control dependent on X iff Y is in PDF(Y).
+//! Thus, we rewrite our control dependence condition as Y is control dependent on X iff X is in PDF(Y).
```
vezenovm, line 43: This comment was incorrect. However, the code is correct. Just noting this change for the reviewer.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -257,30 +265,91 @@ impl<'f> LoopInvariantContext<'f> {
...
+            return;
+        }
+
+        self.is_nested_loop_control_dependent(loop_, block, all_loops, all_predecessors);
+    }
+
+    /// Determines if the `block` is in a nested loop that is control dependent
+    /// on a block in the outer loop.
+    /// If this is the case, we block hoisting as control is not guaranteed.
+    /// If the block is not control dependent on the inner loop itself, it will be marked appropriately
```
vezenovm, line 307:
```suggestion
    /// If the block is not control dependent on the inner loop's header, it will be marked appropriately
```

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -257,30 +265,91 @@ impl<'f> LoopInvariantContext<'f> {
...
+    /// when the inner loop is being processed later.
```
vezenovm, line 308:
```suggestion
    /// If the block is not control dependent on the inner loop itself, it will be marked appropriately
    /// when the inner loop is processed later.
```

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -257,30 +265,91 @@ impl<'f> LoopInvariantContext<'f> {
...
+    /// It is expected that `all_predecessors` contains at least all of these blocks.
+    fn is_nested_loop_control_dependent(
+        &mut self,
+        loop_: &Loop,
+        block: BasicBlockId,
+        all_loops: &[Loop],
+        all_predecessors: Vec<BasicBlockId>,
+    ) {
+        // Now check for nested loops within the current loop
+        for nested in all_loops.iter() {
```
vezenovm, line 321: I decided to start from the front of the list of loops as I was thinking nested loops are most likely smaller. Although I whether checking nested loop from the front or back is faster feels largely dependent on the program itself. Ideally we would have a tree structure of some kind to model nested loop explicitly. As that is a larger refactor I went with this approach of searching all loops for now.

jfecher:
> rollup-block-root	9.85 s 11.6 s 0.85

Looks like it updated with numbers more similar to the previous values now unfortunately

jfecher (approved):
:+1:

vezenovm:
> Looks like it updated with numbers more similar to the previous values now unfortunately

Ah dang. Yeah looks like we execute this circuit a single time (https://github.com/noir-lang/noir/blob/0e62fed6423c1c9b25105337f64d7947c9ed2e05/.github/benchmark_projects.yml#L74). This was set quite some time ago (`rollup-block-root` used to take >40s to execute). Perhaps it is bumping this to 5 runs (or maybe 3 runs if we think 11s is still too long for 5 runs).

PR #8594, opened 2025-05-20
feat: trait associated type bounds

asterite:
# Description

## Problem

A requirement for #8395

## Summary

In order to implement this several things needed to be done:
1. First, of course, introduce the new syntax
2. Then, we need to resolve and store the bounds on trait associated types. This is stored in the `associated_type_bounds` field of `Trait`.
3. We also need to verify that associated types on trait impls match the above bounds.
4. Then, if we have a where clause like `T: Foo` where `Foo` has an associated type `E` with a bound on it, we need to introduce those bounds on the associated type. However, I faced a challenge here because traits were collected after defining function metas, and it's when collecting traits that we find out about these bounds. For this, I split `collect_traits` into two: `collect_traits` and `collect_trait_methods`. `collect_traits` runs before defining function metas and it's here that we collect these bounds. This worked!
5. Another challenge was that after these bounds were turned into trait constraints, in trait impls they would give an error saying that these constraints didn't match the parent trait. For example in this code:
    ```noir
    trait Foo {
        type E: Bar;
        fn bar(self) -> Self::E;
    }

    trait Bar {
        fn bar_method(self);
    }

    impl<T> Eq for Baz<T> where T: Foo {
        fn eq(self, other: Self) -> bool {}
    }
    ````
    We need the constraint on `Foo::E` to be `Bar` inside `eq`, but for that to work it needed to be put in the `trait_constraints` field of `FuncMeta`. But if we do that then those constraints don't match the one (none) on the `eq` definition of the `Eq` trait.
    For this, I decided to split the `trait_constraints` field of `FuncMeta` into two:
    - `trait_constraints`: these are just the ones from the where clause on the function
    - `extra_trait_constraints`: these are the ones from the parent trait impl, and also ones for implicitly added named generics (like `Foo:E` above)
    Then when checking that the trait constraints match, we just check in `trait_constraints`. For knowing all the trait constraints needed for the function to work, there's a new `all_trait_constraints()` method.
6. Finally, to really prove that this works as expected, I changed `BuildHasher` from being `BuildHasher<H>` to having `H` as an associated type with a bound. It worked!

The one thing missing (there's a TODO about this) is being able to put a where clause on a trait, with that where clause referring to a trait that has an associated type with bounds. This example in particular:

```noir
trait Foo {
    type E: Bar;
    fn bar(self) -> Self::E;
}

trait Bar {
    fn bar_method(self);
}

trait Baz<T>
where
    T: Foo,
{
    fn baz(t: T) {
        let bar = t.bar();
        bar.bar_method(); // This method here is not found
    }
}
```

Given that we currently don't have code for this, and it's not needed for `BuildHasher`, I'd prefer to implement it as a follow-up PR. A `Trait` object has a `where_clause`, but we'd likely need to split that into two too (not sure!).

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1001,13 +1004,35 @@ impl<'context> Elaborator<'context> {
         let func_generics = vecmap(&self.generics, |generic| generic.type_var.clone());
 
         let associated_generics = self.desugar_trait_constraints(&mut func.def.where_clause);
-        let mut generics = vecmap(associated_generics, |generic| generic.type_var);
+
+        let mut generics = Vec::with_capacity(associated_generics.len());
+        let mut associated_generics_trait_contraints = Vec::new();
```
TomAFrench, line 1009:
```suggestion
        let mut associated_generics_trait_constraints = Vec::new();
```

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1001,13 +1004,35 @@ impl<'context> Elaborator<'context> {
...
+
+        for (associated_generic, bounds) in associated_generics {
+            for bound in bounds {
+                let typ = Type::TypeVariable(associated_generic.type_var.clone());
+                let location = associated_generic.location;
+                self.add_trait_bound_to_scope(location, &typ, &bound, bound.trait_id);
+                associated_generics_trait_contraints
```
TomAFrench, line 1016:
```suggestion
                associated_generics_trait_constraints
```

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -2164,7 +2237,15 @@ impl<'context> Elaborator<'context> {
             trait_impl.resolved_generics = self.generics.clone();
 
             let new_generics = self.desugar_trait_constraints(&mut trait_impl.where_clause);
-            for new_generic in new_generics {
+            let mut new_generics_trait_contraints = Vec::new();
```
TomAFrench, line 2240:
```suggestion
            let mut new_generics_trait_constraints = Vec::new();
```

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -2164,7 +2237,15 @@ impl<'context> Elaborator<'context> {
...
+            for (new_generic, bounds) in new_generics {
+                for bound in bounds {
+                    let typ = Type::TypeVariable(new_generic.type_var.clone());
+                    let location = new_generic.location;
+                    self.add_trait_bound_to_scope(location, &typ, &bound, bound.trait_id);
+                    new_generics_trait_contraints
```
TomAFrench, line 2246:
```suggestion
                    new_generics_trait_constraints
```

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1001,13 +1004,35 @@ impl<'context> Elaborator<'context> {
...
+        }
 
         // We put associated generics first, as they are implicit and implicit generics
         // come before explicit generics (see `Type::instantiate_with`).
         generics.extend(func_generics);
 
         let mut trait_constraints = self.resolve_trait_constraints(&func.def.where_clause);
+        let mut extra_trait_constraints =
+            vecmap(extra_trait_constraints, |(constraint, _)| constraint);
+        extra_trait_constraints.extend(associated_generics_trait_contraints);
```
TomAFrench, line 1035:
```suggestion
        extra_trait_constraints.extend(associated_generics_trait_constraints);
```

TomAFrench (approved):
LGTM

jfecher (approved):
LGTM

PR #8604, opened 2025-05-21
feat(fuzz): Generate println statements in unconstrained code

### LINKED ISSUE #7931: Generate arbitrary print statements
Follow up https://github.com/noir-lang/noir/pull/7890 with the generation of print statements, which are captured and compared by the execution. 

The difficulty here is that we need to call a builtin function with a format description for the type we pass.

aakoshh:
# Description

## Problem\*

Resolves #7931

## Summary\*

Adds a `gen_print` method to print an arbitrary local variable in the AST fuzzer, for now only in unconstrained functions. Adds the ability to the `AstPrinter` to display calls to the `#[oracle(print)]` as if they were the `std` version `print` or `println`, which we use with `DisplayAstAsNoir`. 

## Additional Context

In this PR prints are only enabled in unconstrained functions, because in constrained code the compiler demands that the call to an oracle is wrapped in a proxy function. I'll do that in #8605 

Prints are disabled in comptime code because the `Interpreter` currently prints straight to stdout, which isn't captured. I'll try to change that in #8606 

The printed values are captured by the executor and compared in the tests (which is how I know that the interpreter output isn't part of that, because they are failed).

## Documentation\*

Check one:

# PR Checklist\*


tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -715,6 +718,14 @@ impl<'a> FunctionContext<'a> {
         }
 
         if self.unconstrained() {
+            // For now only try prints in unconstrained code, were we don't need to create a proxy.
```
michaeljklein, line 721: R.e. `For now`: does this need a follow up?

compiler/noirc_frontend/src/monomorphization/printer.rs
```diff
@@ -455,6 +481,25 @@ impl AstPrinter {
...
+        let Expression::Literal(Literal::Bool(with_newline)) = args[0] else {
+            unreachable!("the first arg of print is a bool");
+        };
+        if with_newline {
+            write!(f, "println")?;
+        } else {
+            write!(f, "print")?;
+        }
+        write!(f, "(")?;
+        self.print_expr(&args[1], f)?;
```
michaeljklein, line 498: Would it be useful to include the other two arguments when printing this?

```
        // The print oracle takes 2 parameters: the newline marker and the value,
        // but it takes 2 more arguments: the type descriptor and the format string marker.
```

michaeljklein (approved):

tooling/ast_fuzzer/src/program/func.rs
aakoshh: Yes, thanks for reminding, created https://github.com/noir-lang/noir/issues/8605

compiler/noirc_frontend/src/monomorphization/printer.rs
aakoshh: I'll extend the comment to explain that these parameters are automatically added by the monomorphizer, but they are not parameters of `println` itself.

PR #8607, opened 2025-05-21
fix(ssa): Remove the array cache from the function inserter

### LINKED ISSUE #8333: internal error: entered unreachable code: All Value::Instructions should already be known during inlining after creating the original inlined instruction
### Aim

I want to parse packed data structures, where I can control the number of data structures per field with a global variable (so, in case I later want to change it, I can simply change the global variables value)

### Expected Behavior

When running the parsing loop 

    `for i in 0..(MAX_CHARACTERS as u8) {
        let (curr_valid, curr_c) = Character::parse(i, data, actions_data, wait_action);
        is_valid &= curr_valid;
        result[i] = curr_c;
    }`

the expected behavior is to parse MAX_CHARACTERS Character data structures.

### Bug

Instead the following error is thrown:
Message:  internal error: entered unreachable code: All Value::Instructions should already be known during inlining after creating the original inlined instruction. Unknown value v5845 = Instruction { instruction: Id(7439), position: 0, typ: Array([Numeric(Unsigned { bit_size: 8 })], 32) }
Location: compiler/noirc_evaluator/src/ssa/opt/inlining.rs:352

[skplg] Testing test_parse_character_round0 ... FAIL
An unexpected error happened

### To Reproduce

```noir
global WIDTH_MASK: u8 = 31;
global CLASS_MASK: u8 = 7;
global WIDTH_BITS: u8 = 5;
global MAX_CHARACTERS: u32 = 5;

struct Character {
    pub x: u8, // 5 bit
    pub class: u8, // 3 bit
    pub actions: [[u8; 32]; 7],
}
impl Character {
    fn is_class_valid(class: u8) -> bool {
        class as u32 <= 5
    }

    pub fn is_valid(self) -> bool {
        let mut valid = true;
        let x = self.x;
        let class = self.class;
        println(x);
        valid &= Character::is_class_valid(class);
        valid
    }

    fn parse(
        id: u8,
        data: Field,
        actions_data: [Field; 30],
        wait_action: [u8; 32],
    ) -> (bool, Character) {
        let mut is_valid = true;

        let split: [u8; 32] = data.to_be_bytes::<32>();
        assert(split.len() == 32);
        let x_packed = split[2 + id * 6]; // 5 bit x, 3 bit class

        let x = x_packed & WIDTH_MASK;
        let mut class = (x_packed >> WIDTH_BITS) & CLASS_MASK;
        if !Character::is_class_valid(class) {
            class = 0;
            is_valid = false;
        }

        let c = Character {
            x: x,
            class: class,
            actions: [
                wait_action,
                actions_data[id * 6 as u8].to_be_bytes::<32>(),
                actions_data[id * 6 as u8 + 1].to_be_bytes::<32>(),
                actions_data[id * 6 as u8 + 2].to_be_bytes::<32>(),
                actions_data[id * 6 as u8 + 3].to_be_bytes::<32>(),
                actions_data[id * 6 as u8 + 4].to_be_bytes::<32>(),
                actions_data[id * 6 as u8 + 5].to_be_bytes::<32>(),
            ],
        };
        (is_valid, c)
    }
}

fn create_character(x: u8, class: u8) -> Character {
    let c = Character { x: x, class: class, actions: [[0 as u8; 32]; 7] };
    assert(c.is_valid());
    c
}
fn parse_characters(data: Field, actions_data: [Field; 30]) -> (bool, [Character; MAX_CHARACTERS]) {
    let mut is_valid = true;

    let wait_action = [
        0 as u8, 0 as u8, 240 as u8, 0 as u8, 63 as u8, 0 as u8, 0 as u8, 1 as u8, 0 as u8,
        32 as u8, 0 as u8, 0 as u8, 0 as u8, 0 as u8, 32 as u8, 0 as u8, 0 as u8, 0 as u8, 0 as u8,
        32 as u8, 0 as u8, 0 as u8, 0 as u8, 0 as u8, 32 as u8, 0 as u8, 0 as u8, 0 as u8, 0 as u8,
        0 as u8, 0 as u8, 0 as u8,
    ];

    // no need to optimize, compiler automatically rolls this out
    let mut result: [Character; MAX_CHARACTERS] = [
        create_character(0, 0),
        create_character(0, 0),
        create_character(0, 0),
        create_character(0, 0),
        create_character(0, 0),
    ];
    for i in 0..(MAX_CHARACTERS as u8) {
        let (curr_valid, curr_c) = Character::parse(i, data, actions_data, wait_action);
        is_valid &= curr_valid;
        result[i] = curr_c;
    }

    // let (curr_valid, curr_c) = Character::parse(0, data, actions_data, wait_action, events, enemy_advance);
    // is_valid &= curr_valid;
    // result[0] = curr_c;
    // let (curr_valid, curr_c) = Character::parse(1, data, actions_data, wait_action, events, enemy_advance);
    // is_valid &= curr_valid;
    // result[1] = curr_c;
    // let (curr_valid, curr_c) = Character::parse(2, data, actions_data, wait_action, events, enemy_advance);
    // is_valid &= curr_valid;
    // result[2] = curr_c;
    // let (curr_valid, curr_c) = Character::parse(3, data, actions_data, wait_action, events, enemy_advance);
    // is_valid &= curr_valid;
    // result[3] = curr_c;
    // let (curr_valid, curr_c) = Character::parse(4, data, actions_data, wait_action, events, enemy_advance);
    // is_valid &= curr_valid;
    // result[4] = curr_c;
    (is_valid, result)
}

#[test]
fn test_parse_character_round0() {
    let data = 0x2912640000004b03190000006c04142000008a0464000000aa47640b340a;

    let parsed = parse_characters(
        data,
        [
            0x03f00001300314002000000000100300000020000000002000000000000000,
            0x02f0000110071020100a000020100a00000020000000002000000000000000,
            0x02f0000a3b1901002000000000000801000020000000002000000000000000,
            0x023100081b161118120a010018120a01000020000000002000000000000000,
            0xf0003f000001002000000000200000000020000000002000000000000000,
            0xf0003f000001002000000000200000000020000000002000000000000000,
            0x03f00001300414002000000000100300000020000000002000000000000000,
            0x02f0000110071008100a000008100a00000020000000002000000000000000,
            0xf0003f1b1c11002000000000000702000020000000002000000000000000,
            0xff003f1b1c1100200e013f00000e013f0020000000002000000000000000,
            0xf0003f000001002000000000200000000020000000002000000000000000,
            0xf0003f000001002000000000200000000020000000002000000000000000,
            0x03f00001300314002000000000100300000020000000002000000000000000,
            0x02f0000110071018100a000018100a00000020000000002000000000000000,
            0x09f0000130060c002000000000200000000020000000002000000000000000,
            0xf00001100610001005000100100500010020000000002000000000000000,
            0xf00001301c0100100d000000100d00000020000000002000000000000000,
            0x0bf00001100c00002000000000200000000010040a00002000000000000000,
            0x03f00001300414002000000000100300000020000000002000000000000000,
            0x02f0000110061028100a000028100a00000020000000002000000000000000,
            0xf0003f000001002000000000200000000020000000002000000000000000,
            0xf0003f000001002000000000200000000020000000002000000000000000,
            0xf0003f000001002000000000200000000020000000002000000000000000,
            0xf0003f000001002000000000200000000020000000002000000000000000,
            0x03f00001300414002000000000100300000020000000002000000000000000,
            0x02f0000110071010100a000010100a00000020000000002000000000000000,
            0xf000071b1611002000000000000705000020000000002000000000000000,
            0x6300061b161164150a020164150a02010020000000002000000000000000,
            0xf0003f000001002000000000200000000020000000002000000000000000,
            0xf0003f000001002000000000200000000020000000002000000000000000,
        ],
    );
    assert(parsed.0);
}
```

### Workaround

Yes

### Workaround Description

There are 3 ways to make above code work (which I compressed to the parts that trigger the bug):

1.) Roll out the loop (which also works with the full version of the posted code):

Remove the loop and replace it with the rolled out variant that I commented out

2.) Remove the assert in create_character:

Interestingly, for this reduced version, removing that assert (that actually is only there to avoid accidentally configuring invalid default values), fixes the bug.

3.) Remove the println in is_valid:

Also only relevant for this reduced version: Removing that println fixes the bug (worthless for the full code, as this applies to anything that where you reference any other property of the character).

4.) Remove the actions property (everywhere):

This was the most interesting part: Removing that big array also fixes the bug, which makes no sense to me, but it does ...

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.3 noirc version = 1.0.0-beta.3+ceaa1986628197bd1170147f6a07f0f98d21030a (git version hash: ceaa1986628197bd1170147f6a07f0f98d21030a, is dirty: false)

### NoirJS Version

1.0.0-beta.3

Barretenberg - bb CLI 0.82.2 - 0.86.0 (tested in all of those)

None

### Support Needs

_No response_

vezenovm:
…ay repitition in unrolling to hoisting

# Description

## Problem\*

Resolves #8333 

Take this reduction of the original program:
```noir
fn main() -> pub [[u8; 32]; 2] {
    let mut result = [[0; 32]; 2];
    let actions_data: [Field; 30] = [0; 30];
    for i in 0..1 {
        result = [actions_data[i].to_be_bytes::<32>(), actions_data[i + 1].to_be_bytes::<32>()];
    }
    result
}
```
As part of https://github.com/noir-lang/noir/pull/6071 an array cache was added to the function inserter which is used during unrolling to prevent OOM panics. This array cache also hoists make array instructions out of the loop https://github.com/noir-lang/noir/blob/72ce94beeac6d15b67141307bc38f9e2b0020b20/compiler/noirc_evaluator/src/ssa/ir/function_inserter.rs#L146

In the above program we have a call whose result is used in a make array. That make array containing the call result is then hoisted out of the loop, but the call itself is not hoisted. This leads to a mismatch in the ordering of instructions, thus triggering a panic during inlining. The same panic can be triggerd by normalizing IDs after the unrolling pass (e.g. --show-ssa).

## Summary\*

https://github.com/noir-lang/noir/pull/6071 was built before LICM and Brillig globals existed at all (we used to inline arrays in plan wherever the global was used). The cache was a workaround to avoid compilation time/memory blow-ups for large constant arrays used in a loop. I chose to lean into LICM and remove the cache entirely. This allows us to simplify the function inserter and have better separation of concerns (hoisting vs. unrolling). 

## Additional Context

After this PR it may be worth considering to require LICM to precede unrolling. We currently always run LICM before unrolling, however, we may want to make this requirement more explicit. 

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/normalize_value_ids.rs
```diff
@@ -178,7 +178,7 @@ impl IdMaps {
             value @ Value::Instruction { instruction, .. } => {
                 *self.values.get(&old_value).unwrap_or_else(|| {
                     let instruction = &old_function.dfg[*instruction];
-                    unreachable!("Unmapped value with id {old_value}: {value:?}\n  from instruction: {instruction:?}, SSA: {old_function}")
+                    unreachable!("Unmapped value with id {old_value}: {value:?}\n  from instruction: {instruction:?}, from function: {}", old_function.id())
```
vezenovm, line 181: This clutters the output if the function is big. The engineer debugging can always manually print the entire function SSA if they desire.

vezenovm:
Huh looks like all the reports aside semaphore are gone.

vezenovm:
> Yeah, they got broken by a previous PR. 🤷

Assumed as such, I just didn't have the chance to investigate yet.

TomAFrench:
It needs a sync to be fixed.

TomAFrench (approved):

PR #8613, opened 2025-05-21
feat: Replace converging jmpif with empty block destinations with a single jmp

vezenovm:
# Description

## Problem\*

No issue but something I noticed originally when making the pruning dead parameters PR and was surprised to not see certain jumps simplified.

## Summary\*

Attempt to simplify a `jmpif` terminator if both branches converge.

We define convergence as when two branches of a `jmpif` ultimately lead to the same
destination block, after following chains of empty blocks. If they do, the conditional
jump is unnecessary and can be replaced with a simple `jmp`.

There is room for follow-up PRs that do not restrict jump chains to empty blocks. For example, if we had entirely pure instructions in a block it would also be safe to follow the jump chain as we do in this PR. For now, we restrict a jump chain to only be empty blocks with no parameters.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Be aware that this can break flattening (I think).

TomAFrench:
Ah, I misunderstood what you meant by converging

compiler/noirc_evaluator/src/ssa/opt/simplify_cfg.rs
```diff
@@ -254,6 +250,68 @@ fn check_for_negated_jmpif_condition(
...
+    };
+
+    let then_final = resolve_jmp_chain(function, *then_destination);
+    let else_final = resolve_jmp_chain(function, *else_destination);
+
+    // If both branches end at the same target, we can replace the jmpif with a jmp
+    if then_final == else_final {
+        let jmp = TerminatorInstruction::Jmp {
+            destination: then_final,
+            // The blocks in a jmp chain are assumed to have empty arguments
```
michaeljklein, line 277:
```suggestion
            // The blocks in a jmp chain are checked to have empty arguments by resolve_jmp_chain
```

michaeljklein (approved):
Small doc change but LGTM

PR #8635, opened 2025-05-22
Af/8606 capture comptime print

### LINKED ISSUE #8606: Capture prints at compile time
Follow up https://github.com/noir-lang/noir/pull/8604 with capturing `println` output from the comptime interpreter.

aakoshh:
# Description

## Problem\*

Resolves #8606

## Summary\*

Adds an `interpreter_output` to `Context` which is accessed by the comptime `Interpreter` through the `Elaborator` to do the printing. The field holds a trait object which implements `std::io::Write`; we use dynamic dispatch to access it for writing.

In the fuzz test tests the content goes in a byte vector and is used to compare the output from the interpreter with the real execution. DRAFT because the formats are different.

## Additional Context

I started off by adding a generic `W: std::io::Write`, and borrowing a mutable reference to it through the `Elaborator`, but the number of places I had to add it was huge relative to how little I expect comptime printing to be used. I thought the performance impact of dynamic dispatch should be negligible, as this is most likely used for debugging, and this way the code churn is kept to a minimum. 

I ended up using `Rc<RefCell<dyn std::io::Write>>` so that the we can share the ownership over something writeable between the test and the `Context` without losing type information, which is what would happen if we used `Box` and passed it over to `Context`, then took it back. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/comptime.rs
+++ b/compiler/noirc_frontend/src/elaborator/comptime.rs
@@ -89,6 +89,7 @@ impl<'context> Elaborator<'context> {
             self.def_maps,
             self.usage_tracker,
             self.crate_graph,
+            self.interpreter_output,
             self.crate_id,
             self.interpreter_call_stack.clone(),
             self.options,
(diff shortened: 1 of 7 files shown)


TomAFrench (approved):
LGTM

PR #8637, opened 2025-05-22
fix: use correct module for copied trait impl default method

### LINKED ISSUE #8632: Import errors on partial trait implementations
### Aim

I have a trait `PartialTrait` which has one implemented method and one unimplemented method. I want to implement the unimplemented method for a type. If the already implemented method calls a function that is not visible _to the implementing type_, then I get compiler errors, despite the function being visible to the implementation where this is used.

### Expected Behavior

The following code should compile:

```noir
mod my_trait {
    fn do_thing<T>(value: T, other: T) -> Field
    where
        T: Eq,
    {
        if value.eq(other) {
            5
        } else {
            6
        }
    }

    pub trait PartialTrait: Eq {
        fn unimplemented_method() -> Field;

        fn implemented_method(self, other: Self) -> Field {
            do_thing(self, other)
        }
    }
}

use my_trait::PartialTrait;

#[derive(Eq)]
struct Foo {
    foo: Field,
}

impl PartialTrait for Foo {
    fn unimplemented_method() -> Field {
        7
    }
}
```

### Bug

`nargo check` emits an error with the above: 
```
error: cannot find `do_thing` in this scope
   ┌─ src/main.nr:17:13
   │
17 │             do_thing(self, other)
   │             -------- not found in this scope
   │
```

If I add `use my_trait::do_thing`, to bring the function in scope, then `nargo check` no longer emits this error.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8632

## Documentation

Check one:

# PR Checklist


asterite:
This now breaks when a trait impl is specified in a different crate than the trait it's implementing (when the trait has a default method). It's tricky because we could replace both the local module ID and crate ID, but then it panics because we don't allow elaborating functions that are from a different crate.

I think the solution is to not elaborate these copied default functions. It should be enough to elaborate them once in the context of the trait. Then, yes, they should be monomorphized but only that.

PR #8638, opened 2025-05-22
fix(ssa interpreter): Default to zero when we have an overflowing shl

### LINKED ISSUE #8615: SSA Interpreter: LICM: Always failing constraint moved without the corresponding operation
### Aim

Found testing https://github.com/noir-lang/noir/pull/8610

I'm sure this is a bug actually, but I thought I'd rather report it to discuss. 

The failure can be reproduced with the following command:
```
NOIR_ARBTEST_SEED=0x3ffa561e00045623 cargo test -p noir_ast_fuzzer_fuzz pass_vs_prev
```

The code in question can be minimised to this:
```rust
fn main() {
    for _ in 0..1 {
        let h = 861;
        for _ in 0..1 {
            for _ in 0..1 {
                let mut k = (h << 69) as u128;
            }
        }
    }
}

```

### Expected Behavior

My initial expectation was that the different passes should fail in the same way, doing the same operation. I can relax this assumption, but wanted to confirm if there is anything else we have to do in this case. 

### Bug

Before Loop Invariant Code Motion we have somewhere in the SSA this:
```
  b0(v14: u1, v15: u128, v16: Field):
    ...
    jmp b1(u128 52359585508132965347953154358054934246)
  b16():
    jmp b18(u16 861)
  b17():
    jmp b18(u16 861)
  b18(v62: u16):
    v87 = shl v62, u8 69
    constrain u1 0 == u1 1, "attempt to bit-shift with overflow"
    ...
```
After the LICM it changes to this:
```
  b0(v14: u1, v15: u128, v16: Field):
    ...
    constrain u1 0 == u1 1, "attempt to bit-shift with overflow"
    ...
    jmp b1(u128 52359585508132965347953154358054934246)
  ...
  b16():
    jmp b18(u16 861)
  b17():
    jmp b18(u16 861)
  b18(v62: u16):
    v158 = shl v62, u8 69
     ...
```

So at compile time we know that we will perform `861 << 69` which will fail, and insert an always-failing assertion with the appropriate error message. 

The first SSA fails in the interpreter with this message:
```
Overflow { instruction: "`shl v62, u8 69` (u16 861 << 69)" }
```
While the second SSA fails with the constraint:
```
ConstrainEqFailed { lhs: "u1 false", lhs_id: Id(66), rhs: "u1 true", rhs_id: Id(20) }
```

The `shl` operation _is_ a loop invariant, but it's not recognised as such and stays in place, changing the way the interpreter fails. 

Is that okay? Should the interpreter have failed on the constraint instead of the `shl`, or should ACIR/Brillig fail in the `shl` and not need the constraint? 

If this is fine I'll update my test to consider `Overflow` and `ConstrainEqFailed` equivalents (I don't have location information for them in my tests, although it's something I could potentially use to reinforce this).

### To Reproduce

See above.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+b2bfdb7bb0fc661bfb27da365f108cf131771295 (git version hash: ef1c8dcac7d0542a5cdc21c9cc5e59b93f26be2a, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #8615 

## Summary\*

As overflowing shift left returns zero just like a shift right during execution:
https://github.com/noir-lang/noir/blob/8f660a3a6c8b1e48b9e988d181a65df44e0bc715/acvm-repo/brillig_vm/src/arithmetic.rs#L245

However, in both the comptime interpreter and and the SSA interpreter we simply errored out with an overflow.

## Additional Context

https://github.com/noir-lang/noir/pull/8638#issuecomment-2902061280

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
@@ -1069,12 +1069,6 @@ impl Interpreter<'_> {
                     }));
                 };
 
-                let overflow = || {
-                    let instruction =
-                        format!("`{}` ({lhs} << {rhs})", display_binary(binary, self.dfg()));
-                    InterpreterError::Overflow { instruction }
-                };
-
                 let rhs = rhs as u32;
                 use NumericValue::*;
                 match lhs {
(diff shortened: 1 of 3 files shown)


asterite (approved):

vezenovm:
It looks like overflowing shift left in ACIR is wrapping https://github.com/noir-lang/noir/blob/8f660a3a6c8b1e48b9e988d181a65df44e0bc715/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs#L96

However, as linked in the OP in Brillig we return zero on an overflowing shl. This is slightly different from this PR as it means we have a shl discrepancy between ACIR/Brillig. I will try to reproduce an execution discrepancy in a separate test. I was thinking we can keep this PR as it still prevents error'ing out on an overflowing shl in the interpreter which is the expected behavior and ultimately resolves #8615.

michaeljklein (approved):

Issue #8620, opened 2025-05-21
SSA Printer does not display the assertion message of range checks


aakoshh:
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8610

### Expected Behavior

The printed SSA should represent the instruction faithfully. 

### Bug

I get an error like this:
```
Range check of v164 = Field 781294165659571600 failed.
  Max bits allowed by range check = 32
  Actual bit count = 60
  attempt to multiply with overflow
```
or in Debug format:
```
RangeCheckFailedWithMessage { value: "Field 781294165659571600", value_id: Id(164), actual_bits: 60, max_bits: 32, message: "attempt to multiply with overflow" }
```

but in the SSA it only shows up like so:
```
range_check Field 781294165659571600 to 32 bits
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

```shell
NOIR_ARBTEST_SEED=0x1c9d6d5800100000 cargo test -p noir_ast_fuzzer_fuzz pass_vs_prev
```

### Project Impact

None

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+b2bfdb7bb0fc661bfb27da365f108cf131771295 (git version hash: ef1c8dc, is dirty: true)

### Would you like to submit a PR for this Issue?

None

asterite:
I think "Actual bit count = 60" comes from the actual Field value, in this case 781294165659571600. The SSA interpreter can show it because it knows the value. But the SSA printer (and the instruction) don't hold an "actual bit count".

Maybe the interpreter error message shouldn't show this? Though it's useful to better understand the error message. Otherwise I don't think there's anything that needs to be done here.

aakoshh:
Sorry for the confusing ticket. What I meant was that in the error there is the message that doesn't show up in the printed SSA itself, ie. this:

```
range_check Field 781294165659571600 to 32 bits
```

should look like this:
```
range_check Field 781294165659571600 to 32 bits, "attempt to multiply with overflow"
```

similarly to how `constrain u1 0 == u1 1, "attempt to bit-shift with overflow"` is printed. It's just that the instruction printer for `RangeCheck` ignores fields with `..` and forgets to show them. I suppose if we parsed the SSA we would not have the same error message, because it's lost in the roundtrip.

aakoshh:
[...]

PR #8647, opened 2025-05-22
chore(fuzz): Fuzz the SSA parser

aakoshh:
# Description

## Problem\*

Related to https://github.com/noir-lang/noir/issues/8620

Some parts of the SSA `Instruction` may not be displayed by the printer, which means the information would not survive an SSA print+parse roundtrip. 

## Summary\*

Adds a new test under the `ast_fuzzer` which generates an arbitrary `Program`, chooses an arbitrary SSA pass, compiles the program up to that point, prints the SSA into a string, parses the SSA, then compares that all values in the DFG are the same.

Run it like so:
```shell
cargo test -p noir_ast_fuzzer --test parser
```
or to replicate a failure:
```
NOIR_ARBTEST_SEED=0x9ecc135e00100000 cargo test -p noir_ast_fuzzer --test parser
```

## Additional Context

It fails currently with various seeds. One problem based on the 3rd example seems to be that the parser must be going in the order of appearance of blocks, which has been changed to alphabetical order, rather than following the CFG topological ordering, so it looks for IDs before they are defined.

##### `0x9ecc135e00100000`
```
v36 = make_array b"\u{c}"
                     ^^^^
                     '\u' is not a valid escape sequence. Use '\' for a literal backslash character.
```

##### `0xa683a43200100000`
```
    v1842, v1843, v1844 = call f1(v1838, v1827, v1840, v1841) -> (u1, u1, u128)
                                         ^^^^^
                                         Unknown variable 'v1827'
```

##### `0x8e62ca0c00100000`
```
    v370 = make_array [v367, v262, v369] : [[Field; 4]; 3]
                             ^^^^
                             Unknown variable 'v262'
```

## Documentation\*

Check one:

# PR Checklist\*


tooling/ast_fuzzer/tests/parser.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        // Print to str and parse back.
+        let ssa2 = Ssa::from_str(&ssa1.to_string()).unwrap_or_else(|e| {
+            let msg = passes.last().map(|p| p.msg()).unwrap_or("Initial SSA");
+            print_ast_and_panic(&format!(
+                "Could not parse SSA after step {last_pass} ({msg}): \n{e:?}"
+            ))
+        });
+
+        // Not everything is populated by the parser, and unfortunately serializing to JSON doesn't work either.
```
asterite, line 86: Is this comment still accurate?

asterite (approved):
Looks great!

tooling/ast_fuzzer/tests/parser.rs
aakoshh: Yes, but it's not because of any deficiency in the parser: for example `used_globals` are only populated by the DIE pass, so after parsing the SSA we don't have the full data structure we printed, even if the SSA implemented `PartialEq`. 

I thought no problem, these fields are marked as `#[serde(drop)]` anyway, so I'll just serialise to JSON and compare that. That doesn't work: attempting to serialise to JSON complains that "only string keys can be used". I spent a few minutes trying to figure out which field isn't marked as serialised to `Vec<_>`, or but gave up and went with the comparison of values instead.

PR #8646, opened 2025-05-22
chore(ci): Do not run ast_fuzzer orig vs. morph in ci

vezenovm:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/pull/8452#issuecomment-2901736666

This fuzzer test is failing on all several PRs non-deterministically.

## Summary\*

Just blocks the test in CI for now, we can bring it back later once the failures are resolved.

## Documentation\*

Check one:
- [X] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [X] I have tested the changes locally.
- [X] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #8664, opened 2025-05-23
chore(ci): avoid running fuzzer tests in the merge queue

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR disables running the fuzzer in the merge queue as it's just causing unnecessary friction as people need to push an otherwise good PR back in to the queue.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,10 @@
+[profile.ci]
+# Do not cancel the test run on the first failure.
+fail-fast = false
+
+[profile.merge-queue]
+# fail fast to kick from merge queue faster.
+fail-fast = true
+
+# Disable fuzzing to avoid flakiness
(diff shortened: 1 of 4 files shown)


aakoshh:
There is another way to disable these quick fuzzers with a comment about which bug is preventing them from being successful, such as https://github.com/noir-lang/noir/pull/8646

Maybe we should run them longer than 10s, or run them as advisory and not cause the entire pipeline to fail, but it would be good to run them to be aware of regressions.

aakoshh:
Ah, okay, I see it's disabled in the merge queue but not in the preceding CI. That's cool 👏

aakoshh (approved):
So this means when someone opens their PR the quick fuzzers get a chance to flag up regressions, but once they pass and they get in the merge queue they don't flake any more, which is great 👍

PR #8665, opened 2025-05-23
fix: Handle `&mut function` in defunctionalize

### LINKED ISSUE #8662: Compiler crash: defunctionalize post check failure
### Aim

The following command fails when run against the `next` branch of `aztec-packages`:

```shell
cargo run -p nargo_cli -- --program-dir /Users/aakoshh/Work/aztec/aztec-packages/noir-projects/noir-contracts/contracts/test/test_contract compile
```

### Expected Behavior

It should compile the contract.

### Bug

```console
❯ cargo run -q -p nargo_cli -- --program-dir /Users/aakoshh/Work/aztec/aztec-packages/noir-projects/noir-contracts/contracts/test/test_contract compile
Message:  Blocks are not expected to take function parameters any more.
Location: compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs:567
```

### To Reproduce

See above. 

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+973a76c3130708d11eaea7474e45e85dac831a8b (git version hash: 973a76c3130708d11eaea7474e45e85dac831a8b, is dirty: true)

### NoirJS Version

_No response_

_No response_

Yes

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #8662 

## Summary\*

The defunctionalize pass did not handle `&mut function` parameters, just `function`. OTOH the post-check looked for function references and failed. This didn't cause a problem on CI because the assertion only works in debug mode, and CI runs in release mode. Only by accident did we find that using `cargo run` to try to compile one of the contracts in aztec-packages fails due to the post check. 

## Additional Context

The AST fuzzer did not find this case because it doesn't generate reference parameters yet.

## Documentation\*

Check one:

# PR Checklist\*


test_programs/execution_success/regression_8662/src/main.nr
```diff
@@ -0,0 +1,23 @@
+
+struct Context {
+    f: fn() -> bool
+}
+fn main(c: bool) -> pub bool {
```
vezenovm, line 5: Missing a stdout?

vezenovm (approved):

test_programs/execution_success/regression_8662/src/main.nr
aakoshh: You mean the `stdout.txt` for the test? I thought they were replaced by insta in https://github.com/noir-lang/noir/pull/8576

vezenovm: oops yes

aakoshh:
My changes to the defunctionalization resulted in the expected SSA string, but later during inlining it turned out that the value of it was wrong, causing a crash during Brillig gen.

The fix was this: https://github.com/noir-lang/noir/pull/8665/commits/642e660b3c61b703ada9659e25a7d6b2a13bf00f 
But the unit test doesn't capture the problem.

I tested the crash with this:
```
cargo run -q -p nargo_cli -- --program-dir /Users/aakoshh/Work/aztec/aztec-packages/noir-projects/noir-contracts/contracts/app/easy_private_voting compile --silence-warnings --package easy_private_voting_contract --show-ssa
```

There is a part which before defunctionalization looks like this:
```
brillig(inline) fn add_to_tally_public f0 {
  b0(v2: Field):
    ...
    v48, v49, v50, v51 = call f10(v30, v31, v32, v33, v34, v2) -> (&mut u1, &mut Field, &mut function, Field)

brillig(inline) fn at f10 {
  b0(v2: &mut u1, v3: &mut Field, v4: &mut function, v5: Field, v6: function, v7: Field):
    v9 = call f16(v5, v7) -> Field
    v10, v11, v12, v13 = call v6(v2, v3, v4, v9) -> (&mut u1, &mut Field, &mut function, Field)
    return v10, v11, v12, v13
}
```
after defunctionalization:
```
brillig(inline) fn at f10 {
  b0(v2: &mut u1, v3: &mut Field, v4: &mut Field, v5: Field, v6: Field, v7: Field):
    v9 = call f16(v5, v7) -> Field
    v10, v11, v12, v13 = call v6(v2, v3, v4, v9) -> (&mut u1, &mut Field, Field, Field)
    return v10, v11, v12, v13
}
```
but after inlining:
```
brillig(inline) fn add_to_tally_public f0 {
  b0(v2: Field):
    ...
    v29, v30, v31, v32 = call Field 33(v3, v5, v7, v27) -> (&mut u1, &mut Field, Field, Field)
```
and then it crashes because it's trying to call Field 33 instead of f33 which it would if the defunctionalization worked correctly.

Just thought I would share that sometimes looking at the SSA in the same pass doesn't actually show the problem, which is hidden in the DFG values. I'll try to extend the AST fuzzer to use references to cover this.

aakoshh:
I'll try to replicate it as an integration test as well next week.

PR #8667, opened 2025-05-23
fix: relax connectedness requirement on purity analysis pass

### LINKED ISSUE #8666: Purity analysis pass fails on inputs with unreachable functions
The current purity analysis pass assumes that all functions are reachable from the main program entrypoint. Any programs which have unreachable functions don't have those functions analysed which go on to fail the post-check condition.

The following test then panics when it shouldn't
```rust
#[test]
    fn handles_unreachable_functions() {
        let src = r#"
        brillig(inline) fn main f0 {
          b0():
            return
        }
          b0():
            return
        }"#;

        let ssa = ssa.purity_analysis();

        let purities = &ssa.main().dfg.function_purities;
        assert_eq!(purities[&FunctionId::test_new(0)], Purity::PureWithPredicate);
        assert_eq!(purities[&FunctionId::test_new(1)], Purity::PureWithPredicate);
    }
```

TomAFrench:
# Description

## Problem\*

Resolves #8666 

## Summary\*

This PR addresses some fuzzing failures which were caused by functions being made unreachable between the `remove_unreachable_functions` and `purity_analysis` passes. The `purity_analysis` assumes that all functions are reachable from `main` so this resulted in some functions not being analysed.

We now run the callgraph analysis from all functions which have no callers to ensure that we hit every function in the callgraph.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/pure.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/pure.rs
@@ -1,4 +1,3 @@
-use std::collections::BTreeSet;
 use std::sync::Arc;
 
 use fxhash::FxHashMap as HashMap;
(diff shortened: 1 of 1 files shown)


vezenovm:
> functions being made unreachable between the `remove_unreachable_functions` and `purity_analysis` passes

Could we just run `remove_unreachable_functions` before the second purity_analysis pass as an alternative? We could then have additional debug assertions that the Ssa input to `purity_analysis` has no unreachable functions?

TomAFrench:
We could but I thought it best to remove the precondition on this pass rather than require us to always fulfill it.

vezenovm:
> We could but I thought it best to remove the precondition on this pass rather than require us to always fulfill it.

Fair enough. I also agree that is nice to get rid of dependency tracking in this pass as well.

vezenovm (approved):

compiler/noirc_evaluator/src/ssa/opt/pure.rs
```diff
@@ -28,20 +27,15 @@ impl Ssa {
...
-
         // First look through each function to get a baseline on its purity and collect
         // the functions it calls to build a call graph.
-        for function in self.functions.values() {
-            let (purity, dependencies) = function.is_pure();
-            purities.insert(function.id(), purity);
-            called_functions.insert(function.id(), dependencies);
-        }
+        let purities: HashMap<_, _> =
+            self.functions.values().map(|function| (function.id(), function.is_pure())).collect();
```
jfecher, line 33: Is the comment above this still accurate? Looks like the call dependencies are moved below

TomAFrench:
I can deal with the comment in a followup. This PR fixes flakiness which is affecting all PRs.

Issue #6467, opened 2024-11-06
Where clause on generic trait's generic argument fails to resolve


michaeljklein:
### Aim

Attempted to compile a program where a trait constrains one of its generic arguments:

```noir
trait Foo {
    fn foo(self) -> bool;
}

trait Bar<T> where T: Foo {}

impl<T, U> Bar<T> for U where T: Foo {}

pub fn bar<T, U>(x: T, _: U) -> bool where U: Bar<T> {
    x.foo()
}

fn main() {}
```

Note: this is blocking [trait aliases](https://github.com/noir-lang/noir/pull/6431) of the form:

```noir
trait Qux<T> = Foo + Bar<T> where T: Baz;
//                          ^ this where clause results in the same "unresolved method call" error
```

### Expected Behavior

Expected the `x.foo()` call to resolve since `x: T` and `U: Bar<T>` implies `T: Foo`

### Bug

The method call does not resolve:

```noir
❯ ~/.nargo/bin/nargo compile
error: No method named 'foo' found for type 'T'
   ┌─ src/main.nr:10:5
   │
10 │     x.foo()
   │     -------
   │

Aborting due to 1 previous error
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Nargo Version

nargo version = 0.37.0 noirc version = 0.37.0+fb1a8ca67c58d87991358078e6c532b49824fdb8 (git version hash: fb1a8ca67c58d87991358078e6c532b49824fdb8, is dirty: false)

### Would you like to submit a PR for this Issue?

None

asterite:
> Expected the x.foo() call to resolve since x: T and U: Bar<T> implies T: Foo

I'm not sure if this explicitness is expected from the language... Or, well, at least in Rust I get an error that says "the trait bound `T: Foo` is not satisfied, the trait `Foo` is not implemented for `T`" on `Bar<T>`.

PR #8686, opened 2025-05-26
fix: catch unbound type variables during frontend compilation

### LINKED ISSUE #7889: Catch unbound type variables earlier in compilation
### Aim

```noir
fn main() {
    let _slice = &[];
}
```

### Expected Behavior

`nargo check` should issue an error that the type variables for the slice type assigned to `_slice` are unknown and need to be specified.

### Bug

We only issue this error during monomorphization currently.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #8464: Misleading error message with unused generic in trait impls
### Aim

I'm trying to make a function return flexible types depending on some input with generics. I have a reproduction [here](https://github.com/sklppy88/noir-bug-repo).

### Expected Behavior

The LSP understands the type I want, and I would expect the code example to compile.

### Bug

Requires me to explicitly annotate the type, even though the LSP seems to be able to detect it automatically.

`Object type is unknown in method call
lib.nr(70, 27): Type must be known by this point to know which method to call
lib.nr(70, 27): Try adding a type annotation for the object type before this method call`

### To Reproduce

Repo [here](https://github.com/sklppy88/noir-bug-repo). I'm at latest master. `nargo compile` fails.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Blocker

### Blocker Context

Trying to clean up some interfaces around calls, although it is non-critical, it would be very nice to have relatively soon.

### Nargo Version

nargo version = 1.0.0-beta.4 noirc version = 1.0.0-beta.4+0000000000000000000000000000000000000000 (git version hash: 0000000000000000000000000000000000000000, is dirty: false)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #7889
Resolves #8464

## Summary

I started implementing this a few times until I landed with the current implementation. At first I thought we'd need to track all type variables, but for example when you have `fn foo<T>() { ... }`, that `T` in that function will remain unbound. It's only when it's called that it will become bound. This is why we can't track all type variables.

Then, I noticed that, _maybe_, the places where we need to track them are:
- array and slice literals (well, for the type of the array literal, and probably only when it's empty, but here it's done unconditionally)
- struct constructors, as the struct generics, if any, must be determined at the end of the function
- identifiers in expressions, corresponding to function calls and enum variants

But let me know if this isn't right.

Then, many errors started to show up. The reason is that we'd previously error during monomorphization. That means that if there was such error in a function that wasn't reached from `main`, that error never showed up. Now it does! And this is why tests will have a few more errors now. And, for example, there was a `#[test]` on top of a function were generics couldn't be deduced. That was previously only erroring when that test function was monomorphized. Now it always errors, which makes the test impossible to write, so I removed it.

Finally, when type-checking `FunctionDefinition::as_typed_expr` the type variables aren't checked because they are later checked when the value is interpolated somewhere else. In reality I'm not sure `as_typed_expr` should be using `type_check_variable_with_bindings`, so let me know if it should be done in a different way.

## Additional Context

I'm unsure about the name `RequiredTypeVariable`, happy to rename it!

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/tests/turbofish.rs
+++ b/compiler/noirc_frontend/src/tests/turbofish.rs
@@ -24,8 +24,11 @@ fn turbofish_numeric_generic_nested_function_call() {
     assert_no_errors!(src);
 }
 
+// TODO: this test should pass, but it doesn't because of this issue:
+// https://github.com/noir-lang/noir/issues/8900
 #[named]
 #[test]
+#[should_panic]
 fn turbofish_numeric_generic_nested_method_call() {
     // Check for turbofish numeric generics used with method calls
     let src = r#"
@@ -44,9 +47,7 @@ fn turbofish_numeric_generic_nested_method_call() {
     }
 
     fn bar<let N: u32>() -> [u8; N] {
-        let _ = Foo::static_method::<N>();
-                     ^^^^^^^^^^^^^ Type annotation needed
-                     ~~~~~~~~~~~~~ Could not determine the type of the generic argument `T` declared on the struct `Foo`
+        let _ = Foo::<u8>::static_method::<N>();
         let x: Foo<u8> = Foo { a: 0 };
         x.impl_method::<N>()
     }
@@ -57,7 +58,7 @@ fn turbofish_numeric_generic_nested_method_call() {
         let _ = bar::<M>();
     }
     "#;
-    check_errors!(src);
+    assert_no_errors!(src);
 }
 
 #[named]


asterite:
@jfecher I feel like the current implementation is correct so I'm a bit hesitant to try a different approach where I'm not sure it'll improve things. I'm just wondering if there's some code (some type variables) that aren't caught in this PR. If so, I can work on fixing those with this approach and, if that doesn't work, I can try the other approach.

asterite:
This is ready for review again, and all external libraries in our control compile file.

compiler/noirc_frontend/src/elaborator/function_context.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+impl Elaborator<'_> {
+    /// Push a type variable into the current FunctionContext to be defaulted if needed
+    /// at the end of the earlier of either the current function or the current comptime scope.
+    pub(super) fn push_defaultable_type_variable(&mut self, typ: Type) {
+        self.get_function_context().defaultable_type_variables.push(typ);
+    }
+
+    /// Push a type variable (it's ID and type) as a required type variable: it must be
```
TomAFrench:
```suggestion
    /// Push a type variable (its ID and type) as a required type variable: it must be
```

compiler/noirc_frontend/src/elaborator/function_context.rs
```diff
@@ -0,0 +1,253 @@
...
+        expr_id: ExprId,
+        select_impl: bool,
+    ) {
+        self.get_function_context().trait_constraints.push((constraint, expr_id, select_impl));
+    }
+
+    fn get_function_context(&mut self) -> &mut FunctionContext {
+        let context = self.function_context.last_mut();
+        context.expect("The function_context stack should always be non-empty")
+    }
```
TomAFrench:
```suggestion
    fn get_function_context_mut(&mut self) -> &mut FunctionContext {
        let context = self.function_context.last_mut();
        context.expect("The function_context stack should always be non-empty")
    }
```

It's more idiomatic to have a `_mut` suffix here.

TomAFrench: Ah, I can see this is a pre-existing method which just got moved. Not a blocker in that case.

compiler/noirc_frontend/src/tests/turbofish.rs
```diff
@@ -45,6 +45,8 @@ fn turbofish_numeric_generic_nested_method_call() {
 
     fn bar<let N: u32>() -> [u8; N] {
         let _ = Foo::static_method::<N>();
+                     ^^^^^^^^^^^^^ Type annotation needed
+                     ~~~~~~~~~~~~~ Could not determine the type of the generic argument `T` declared on the struct `Foo`
```
TomAFrench:
```suggestion
        let _ = Foo::<u8>::static_method::<N>();
```

I think we want to supply `T` rather than asserting an error here.

compiler/noirc_frontend/src/tests/traits.rs
```diff
@@ -450,14 +448,13 @@ fn trait_alias_polymorphic_where_clause() {
...
-                                          ^^^ No matching impl found for `T: Baz`
-                                          ~~~ No impl for `T: Baz`
         }
     "#;
 
     // TODO(https://github.com/noir-lang/noir/issues/6467)
     // assert_no_errors!(src);
-    check_errors!(src);
+    let errors = get_program_errors!(src);
+    assert!(!errors.is_empty());
```
TomAFrench: Why do we just assert that errors exist rather than their exact form now?

TomAFrench (commented):
Looks like there's some test failures now. Not sure how many are flakes but the noir-contracts issue is real.

compiler/noirc_frontend/src/tests/traits.rs
asterite: Ah, it's because the code now has new frontend errors that collide with existing ones, and the `check_errors` function doesn't handle multiple errors in the same place.

Looking again at this test, it's for this issue which I think is invalid: https://github.com/noir-lang/noir/issues/6467 (I added a comment there explaining why). Maybe this test should be removed?

compiler/noirc_frontend/src/tests/turbofish.rs
asterite: You are right. However, it still doesn't compile for some reason so I'll try to figure out why...

asterite: That said, the code doesn't compile in master either, so the bug is unrelated to this PR. Should I try to fix it in this PR?

asterite:
> the noir-contracts issue is real

Right, it seems a bunch of types were introduced with a generic `N` that is only used in a specific impl... I just tried to fix it but I'm not sure how to do it.

asterite:
I think I got it to compile... I'll create a PR.

compiler/noirc_frontend/src/tests/turbofish.rs
TomAFrench: In that case I think we can just ensure that we have an issue open for this bug and link to it here.

compiler/noirc_frontend/src/tests/traits.rs
TomAFrench: Hmm, that's an unfortunate limitation.

Re #6467, if rust doesn't support this then I think we can punt that issue from being a bug to a feature request. We can remove this test as a result.

jfecher: It is worth mentioning Rust also doesn't support trait aliases in stable yet, which we do. Hence why we wanted to allow this code originally

asterite:

jfecher (approved):

PR #8689, opened 2025-05-27
chore: ensure `blake3` input length is less than 1024 bytes

kashbrti:
# Description
BB does not behave correctly when the `blake3` input length is longer than 1024 bytes. 
This leads to programs that we can't generate the BB circuits for correctly. 
However, due to quirks in BB, this will only show up as a proof not being able to be verified. 
So I think it'd be a good idea to catch this in `stdlib`

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench (changes requested):
I've made #8690 which addresses this inside of the stdlib rather than the ACVM.

kashbrti:
ok great. Gave your PR a quick review. should we close this then?

PR #8469, opened 2025-05-12
chore: Add u1 ops to interpreter, convert Value panics to errors

jfecher:
# Description

## Summary\*

Breaks out ssa interpreter changes from this PR: https://github.com/noir-lang/noir/pull/8436

## Additional Context

We apparently allow `u1 + u1` in addition to other odd operations like `u1 << u1`

## Documentation\*

Check one:
- [x] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #8696, opened 2025-05-27
fix: Fix comptime casts of negative integer to field

jfecher:
# Description

## Problem\*

A test case I was using for https://github.com/noir-lang/noir/pull/8669 was incorrect. It stated we expect `-1_u8 as Field == -1` while the rest of our code base uses `-1_u8 as Field = 255`

## Summary\*

Not sure how I missed this originally but the problem is given above. We always zero-extend which casting to fields regardless of whether the source was signed or not.

There is something to be said about this being unintuitive. Even for existing languages casting a signed value to an unsigned one still sign-extends it.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter/cast.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter/cast.rs
@@ -249,10 +249,11 @@ mod tests {
             // Widen
             (Value::I8(127), unsigned(SixtyFour), Value::U64(127)),
             (Value::I8(127), signed(SixtyFour), Value::I64(127)),
-            // Widen negative: zero extend
+            // Widen signed->unsigned: sign extend
             (Value::I8(-1), unsigned(Sixteen), Value::U16(65535)),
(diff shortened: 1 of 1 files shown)


compiler/noirc_frontend/src/hir/comptime/interpreter/cast.rs
```diff
@@ -267,7 +252,8 @@ mod tests {
             // Widen negative: zero extend
             (Value::I8(-1), unsigned(Sixteen), Value::U16(65535)),
```
aakoshh, line 253: I'm just re-reading your explanation of extension [here](https://github.com/noir-lang/noir/pull/8469#issuecomment-2880842026) and wondering: when we go from `-1: i8` which is represented as `255: u8` to `u16` and end up with `65535`, which is `11111111 11111111`, aren't we _sign_ extending?

jfecher: Ah yes, the comment is wrong. We sign extend `signed -> unsigned` integer conversions, we only zero extend signed integers for `signed -> field`

PR #8672, opened 2025-05-23
fix(defunctionalize): Higher order functions dynamic dispatch 

vezenovm:
# Description

## Problem\*

Working towards https://github.com/noir-lang/noir/issues/5503 but not that issue.

The case this fixes has been reduce to the following:
```noir
fn main(x: u32) {
    println(lambdas_with_input_and_return_values(x));
}
fn lambdas_with_input_and_return_values(x: u32) -> u32 {
    let f1 = |value: u32| value;
    let f2 = |value: u32| value + 1;
    let f3 = |value: u32| value + 2;
    let f = if x == 0 {
        f1
    } else if x == 1 {
        f2
    } else {
        f3
    };
    f(x)
}
```

This is the following SSA before defunctionalization: 
<details><summary>Details</summary>
<p>

```
acir(inline) fn main f0 {
  b0(v0: u32):
    v2 = call f2(v0) -> u32
    call f1(v2)
    return
}
acir(inline) fn println f1 {
  b0(v0: u32):
    call f6(u1 1, v0)
    return
}
acir(inline) fn lambdas_with_input_and_return_values f2 {
  b0(v0: u32):
    v4 = eq v0, u32 0
    jmpif v4 then: b1, else: b2
  b1():
    jmp b3(f3)
  b2():
    v6 = eq v0, u32 1
    jmpif v6 then: b4, else: b5
  b3(v1: function):
    v10 = call v1(v0) -> u32
    return v10
  b4():
    jmp b6(f4)
  b5():
    jmp b6(f5)
  b6(v2: function):
    jmp b3(v2)
}
acir(inline) fn lambda f3 {
  b0(v0: u32):
    return v0
}
acir(inline) fn lambda f4 {
  b0(v0: u32):
    v2 = add v0, u32 1
    return v2
}
acir(inline) fn lambda f5 {
  b0(v0: u32):
    v2 = add v0, u32 2
    return v2
}
brillig(inline) fn print_unconstrained f6 {
  b0(v0: u1, v1: u32):
    v21 = make_array b"{\"kind\":\"unsignedinteger\",\"width\":32}"
    call print(v0, v1, v21, u1 0)
    return
}
```

</p>
</details> 

On master that gives the following apply function:
<details><summary>Details</summary>
<p>

```
acir(inline_always) fn apply f7 {
  b0(v0: Field, v1: u32):
    v4 = eq v0, Field 3
    jmpif v4 then: b3, else: b2
  b1(v2: u32):
    return v2
  b2():
[...]

PR #8697, opened 2025-05-27
fix(defunctionalize): Create a placeholder function for first-class function calls with no variants

vezenovm:
# Description

## Problem\*

Resolves last bug from https://github.com/noir-lang/noir/pull/8672 with empty arrays.

Alternative to https://github.com/noir-lang/noir/pull/8677

## Summary\*

For function calls with no variants we no longer skip updating the calls. We generate a dummy function so that there is still a concrete call to a real function rather than an SSA value ID. This was something I was debating when originally fixing the compiler panic for not having any function variants https://github.com/noir-lang/noir/pull/8573. At the time it did not cause issues to simply not update those variant-less calls. To be safe, we can now add a dummy function that allows us to continue compilation and not break the semantics of other passes. This includes generating the approach parameters and return values for the function (e.g., what if inlining attempts to inline the dummy function). This also makes the defunctionalization pass more accurate, as no matter what every function used as a value will be transformed into a concrete call, even those that come from a broken IR (e.g. OOB access on a zero-length array). 

The other solution aside #8677 would be to ban accessing zero-length arrays. However, this feels a bit more like a hack and this PR's solution helps prevent similar bugs in the future.

## Additional Context

https://github.com/noir-lang/noir/pull/8677 and its parent https://github.com/noir-lang/noir/pull/8672

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
Actually I'm going to try one more alternative, where we just insert an `Instruction::Noop` for these calls.

compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
```diff
@@ -945,8 +1008,12 @@ mod tests {
...
         brillig(inline) fn func_2 f2 {
           b0(v0: Field):
-            v2 = call v0(u128 1) -> u1
-            return v2
+            v3 = call f3(u128 1) -> u1
+            return v3
+        }
+        brillig(inline_always) pure fn apply_dummy f3 {
+          b0(v0: u128):
+            return
```
vezenovm, line 1016: At the moment the dummy function return values do not match the signature of the function being called. This IR can panic if the call still exists during inlining. Keeping the IR well formed was the motivation with adding the function parameters. I think it would be good to have IR truly well formed and remove assumptions between the passes. If we prefer this dummy function approach to https://github.com/noir-lang/noir/pull/8677 or accessing zero-length arrays, I will update this PR to have a well formed dummy function (e.g. approach arguments and return values).

vezenovm: I ended up generating dummy return data to keep a well formed SSA. I think this is the better approach and guarantees a correct SSA.

vezenovm:
This PR is ready for review again.

compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
```diff
@@ -616,6 +622,93 @@ fn create_apply_function(
...
+            }
+            function_builder.insert_make_array(array, typ.clone())
+        }
+        Type::Slice(element_types) => {
+            let mut array = im::Vector::new();
+            // The length of the slice does not matter for a dummy function, we simply
+            // desire to have a well formed SSA by returning the correct value for a type.
+            for typ in element_types.iter() {
+                array.push_back(make_dummy_return_data(function_builder, typ));
+            }
```
asterite: Would not inserting anything in this array still work?

asterite (approved):
Looks great!

compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
vezenovm:
```suggestion
            // The contents of a slice do not matter for a dummy function, we simply
            // desire to have a well formed SSA by returning the correct value for a type.
            // Thus, we return an empty slice here.
```

vezenovm: Yes you are correct. I have switched this to return an empty slice.

compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
```diff
@@ -616,6 +622,91 @@ fn create_apply_function(
...
+            let mut array = im::Vector::new();
+            for _ in 0..*len {
+                for typ in element_types.iter() {
+                    array.push_back(make_dummy_return_data(function_builder, typ));
+                }
+            }
+            function_builder.insert_make_array(array, typ.clone())
+        }
+        Type::Slice(element_types) => {
+            let mut array = im::Vector::new();
```
vezenovm:
```suggestion
        Type::Slice(_) => {
            let array = im::Vector::new();
```

Issue #8503, opened 2025-05-14
Make a binary for the SSA interpreter


michaeljklein:
### Problem

When testing with the SSA interpreter, it's useful to recreate the behavior "from scratch" to ensure that it matches the "default SSA interpreter behavior" when using a test harness.

### Happy Case

Add a new binary `ssa_interpreter` or similar:
- It may also be useful to have a command-line flag to run a full execution with `noirc_evaluator`, starting from input SSA
- Accept SSA through STDIN or a file

### Workaround

Yes

### Workaround Description

Make new unit tests

### Project Impact

Nice-to-have

### Would you like to submit a PR for this Issue?

None

PR #8700, opened 2025-05-28
feat(cli): Add `nargo interpret` command

aakoshh:
# Description

## Problem\*

@guipublic said he would like to use the SSA interpreter to evaluate the SSA at a certain point in the pipeline to debug some issue. 

This is related to https://github.com/noir-lang/noir/issues/8503 , but instead of working on a stand-alone SSA snippet where we would have to come up with some input format as well, it taps into the existing flow of `nargo`, dealing with Noir source code and TOML/JSON input, which I thought was a low hanging fruit and fits into our typical workflow.

## Summary\*

Adds a new `nargo interpret` command compiles the source code, looks for binary packages, parses their `Prover.toml` file, then uses the SSA Interpreter to evaluate the SSA after each pass, according to the `--ssa-pass` filter, which can accept multiple passes, similar to `--show-ssa-pass`. 

At the moment it only considers the `primary` passes, it doesn't do the `secondary` like the normal flow. 

## Additional Context

I added a reference from `nargo_cli` to:
* `noirc_evaluator`: so that I can access the interpreter
* `noir_ast_fuzzer`: so that I can turn ABI inputs into SSA interpreter values

We could migrate some of the stuff from the AST fuzzer into the libraries, but I thought it could be done as a follow up, to keep the code churn to a minimum. 

The new command also has quite a lot of boilerplate due to having to compile from source into monomorphized AST, which isn't exposed by the compiler library as a flow. In this it is similar to what I had to do in #8393 

Example:
```console
❯ cargo run -q -p nargo_cli -- interpret --silence-warnings --ssa-pass Flattening --ssa-pass "Removing bit shifts" 
--- Interpreter result after Flattening (step 17):
Ok([Numeric(U8(78))])
---
--- Interpreter result after Removing Bit Shifts (step 18):
Err(Internal(ConstantDoesNotFitInType { constant: -409655725810913393472212215421341542392503380204074161341709628484773418406, typ: Unsigned { bit_size: 64 } }))
---
```

The options include all the `CompileOptions`, such as `--force-brillig` and `--show-ssa`, plus some from `execute`:
```console
❯ cargo run -q -p nargo_cli -- interpret --help
Compile the program and interpret the SSA after each pass, printing the results to the console

Usage: nargo interpret [OPTIONS]

Options:
      --package <PACKAGE>
          The name of the package to run the command on. By default run on the first one found moving up along the ancestors of the current directory

      --workspace
          Run on all packages in the workspace

      ...

  -p, --prover-name <PROVER_NAME>
          The name of the TOML file which contains the ABI encoded inputs
          
          [default: Prover]

      --ssa-pass <SSA_PASS>
          The name of the SSA passes we want to interpret the results at.
          
          When nothing is specified, the SSA is interpreted after all passes.

  -h, --help
          Print help (see a summary with '-h')

```
Options which only affect ACIR or Brillig generation are ignored.

The command itself is hidden, like the SSA options in general. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/ast_fuzzer/src/compare/interpreted.rs
+++ b/tooling/ast_fuzzer/src/compare/interpreted.rs
@@ -160,7 +160,7 @@ impl Comparable for ssa::interpreter::value::Value {
 }
 
 /// Convert the ABI encoded inputs to what the SSA interpreter expects.
-fn input_values_to_ssa(abi: &Abi, input_map: &InputMap) -> Vec<ssa::interpreter::value::Value> {
+pub fn input_values_to_ssa(abi: &Abi, input_map: &InputMap) -> Vec<ssa::interpreter::value::Value> {
     let mut inputs = Vec::new();
     for param in &abi.parameters {
         let input = &input_map
--- a/tooling/nargo_cli/src/cli/interpret_cmd.rs
+++ b/tooling/nargo_cli/src/cli/interpret_cmd.rs
@@ -1,0 +1,224 @@
+//! Use the SSA Interpreter to execute a SSA after a certain pass.
+
+use acvm::acir::circuit::ExpressionWidth;
+use fm::{FileId, FileManager};
+use nargo::constants::PROVER_INPUT_FILE;
+use nargo::ops::report_errors;
+use nargo::package::Package;
+use nargo::workspace::Workspace;
+use nargo_toml::PackageSelection;
+use noirc_driver::{CompilationResult, CompileOptions};
+
+use clap::Args;
+use noirc_errors::CustomDiagnostic;
+use noirc_evaluator::brillig::BrilligOptions;
+use noirc_evaluator::ssa::ssa_gen::{Ssa, generate_ssa};
+use noirc_evaluator::ssa::{SsaEvaluatorOptions, SsaLogging, primary_passes};
+use noirc_frontend::debug::DebugInstrumenter;
+use noirc_frontend::hir::ParsedFiles;
+use noirc_frontend::monomorphization::ast::Program;
+use noirc_frontend::monomorphization::{monomorphize, monomorphize_debug};
+
+use crate::errors::CliError;
+
+use super::compile_cmd::parse_workspace;
+use super::{LockType, PackageOptions, WorkspaceCommand};
+
+/// Compile the program and interpret the SSA after each pass,
+/// printing the results to the console.
+#[derive(Debug, Clone, Args)]
+pub(super) struct InterpretCommand {
+    #[clap(flatten)]
+    pub(super) package_options: PackageOptions,
+
+    #[clap(flatten)]
+    pub(super) compile_options: CompileOptions,
+
+    /// The name of the TOML file which contains the ABI encoded inputs.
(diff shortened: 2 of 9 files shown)


tooling/ast_fuzzer/src/compare/interpreted.rs
```diff
@@ -160,7 +160,7 @@ impl Comparable for ssa::interpreter::value::Value {
 }
 
 /// Convert the ABI encoded inputs to what the SSA interpreter expects.
-fn input_values_to_ssa(abi: &Abi, input_map: &InputMap) -> Vec<ssa::interpreter::value::Value> {
+pub fn input_values_to_ssa(abi: &Abi, input_map: &InputMap) -> Vec<ssa::interpreter::value::Value> {
```
asterite, line 163: Should this function be somewhere near the interpreter? It's a bit strange that this is in the fuzzing tool and it's called from `nargo_cli`.

tooling/nargo_cli/src/cli/interpret_cmd.rs
```diff
@@ -0,0 +1,225 @@
...
+            noir_artifact_cli::fs::inputs::read_inputs_from_file(&prover_file, &abi)?;
+        let ssa_input = noir_ast_fuzzer::input_values_to_ssa(&abi, &prover_input);
+
+        // Generate the initial SSA.
+        let mut ssa = generate_ssa(program)
+            .map_err(|e| CliError::Generic(format!("failed to generate SSA: {e}")))?;
+
+        print_and_interpret_ssa(&ssa_options, &args.ssa_pass, &mut ssa, "Initial SSA", &ssa_input);
+
+        // Run all SSA passes that match the filter, printing the result after each step.
```
asterite, line 97: Shouldn't all passes run, but we'd only print the ones that were passed to the command line?

asterite (approved):
Looks good!

tooling/ast_fuzzer/src/compare/interpreted.rs
aakoshh: I agree, as I said in the description, I would follow up with a PR that moves some of the utility functions from the fuzzer to common libraries. Another example is the function that produces an ABI description from a `Program`. 

I just didn't want to make this PR bigger than it has to be.

tooling/nargo_cli/src/cli/interpret_cmd.rs
aakoshh: There is a number of these filters now:
* `--show-ssa-pass` filters which SSA passes are printed
* `--skip-ssa-pass` is a list of SSA passes which are not actually run, which is something Michael found useful in testing
* `--ssa-pass` of the `interpreter` command lists the passes where we execute the interpreter. 

This comment is a bit confusing, I'll rephrase it. I meant that run the ones which aren't skipped, but altogether it's probably better not to say anything about filtering.

tooling/ast_fuzzer/src/compare/interpreted.rs
aakoshh: Added https://github.com/noir-lang/noir/issues/8707 to follow up on this.

PR #8702, opened 2025-05-28
fix: better way to do LSP file overrides

### LINKED ISSUE #8670: LSP: Crash after switching branches where file doesn't exist
### Aim

The Noir Language Server in my VSCode crashed and I couldn't restart it without it crashing again, after switching git branches from where a file currently open in my editor was added to another branch where it did not exist, but it was still open in the editor. 

### Expected Behavior

The LSP should ignore the editor windows which show files that don't exist on the file system.

### Bug

This is the output of the server:

```
thread 'main' has overflowed its stack
[Error - 12:25:07] The Noir Language Server server crashed 5 times in the last 3 minutes. The server will not be restarted. See the output for more information.
[31mThe application panicked (crashed).[0m
Message:  [36mfiles are expected to be added to the FileManager before reaching the compiler file_path: "/Users/aakoshh/Work/aztec/noir/test_programs/execution_success/regression_8662/src/main.nr"[0m
Location: [35mcompiler/noirc_driver/src/lib.rs[0m:[35m318[0m

```

### To Reproduce

1. Check out new git branch.
2. Create a file on the new branch. Open it in VS Code for editing. 
3. Switch back to the old branch where the file doesn't exist. See that it has a strikethrough in its name in the VS Code tab.
4. See what happens with the LSP, try to restart it without crashing.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+611cbeb1ee64f6a707b7e2fc26518aa112201e9c (git version hash: 611cbeb1ee64f6a707b7e2fc26518aa112201e9c, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8670

## Summary

When you open a Noir file in LSP, and whenever you edit it, we cache that file's path and source code. Then when building a FileManager we pass these files as "overrides" so their contents are read from this cache instead of the filesystem. We do this so that even if you don't save a file, code that references that unsaved content works fine. This is how Rust Analyzer seem to work too (I'm not sure how they implemented it, though, but the exhibited behavior is that).

For overrides, what we used to do before this PR is traversing all files in the filesystem for the relevant packages and adding them to the FileManager, except that if there's an override for a path we'd use that override's source code. That works fine except when a file is deleted from the filesystem. In that case the override won't be added because we never find that path in the filesystem to be able to override it!

Initially I thought we could add the overrides at the end, at least those that we didn't find in the filesystem. The problem with that is that the override's FileID will end up being different than the one before the file was deleted, so things like "Go to definition", etc., which rely on FileIDs won't work well.

An alternative approach, which is what this PR does, is to first compute all the filenames that need to go into the FileManager. Those come from the filesystem, but also from the overrides. Then, to have a consistent FileID order, we can sort all of these filenames, and just then add them to the FileManager (considering overrides here too).

Here's how it works in Rust when you delete a file but continue editing it. Notice how we can still jump to that file from another file, and even if you add a new definition to the deleted file, it will be suggested in other files:

lsp-rust

And now it works the same way in Noir LSP:

lsp-noir

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/lsp/src/requests/mod.rs
+++ b/tooling/lsp/src/requests/mod.rs
@@ -446,7 +446,7 @@ where
 
     // `path` might be a relative path so we canonicalize it to get an absolute path
     let path_buf = PathBuf::from(path);
-    let path_buf = path_buf.canonicalize().unwrap_or(path_buf);
+    let path_buf = std::path::absolute(path_buf.clone()).unwrap_or(path_buf);
 
     let uri = Url::from_file_path(path_buf.to_str()?).ok()?;
 
--- a/tooling/lsp/src/requests/workspace_symbol.rs
+++ b/tooling/lsp/src/requests/workspace_symbol.rs
@@ -11,7 +11,7 @@ use async_lsp::lsp_types::{
 };
 use fm::{FileManager, FileMap};
 use fuzzy_matcher::{FuzzyMatcher, skim::SkimMatcherV2};
-use nargo::{insert_all_files_under_path_into_file_manager, parse_all};
+use nargo::{insert_all_files_under_path, parse_all};
 use noirc_errors::{Location, Span};
 use noirc_frontend::{
     ast::{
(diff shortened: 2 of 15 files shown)


asterite:
Hmm... the order of the files will change depending on the absolute filename they have.

jfecher (approved):

PR #8703, opened 2025-05-28
fix!: disallow casting numeric to bool

### LINKED ISSUE #8157: `Field as bool` acir vs brillig
### Aim

+

### Expected Behavior

## main.nr 
```
fn main(v6: Field) -> pub bool {
	let v10 = v6 - 1;
	let v12 = v10 as bool;
	v12
}
```

## Prover.toml
```
"v6" = 1337
```

## `nargo execute`
```
error: Range constraint of 254 bits is too large for the Field size
  ┌─ src/main.nr:3:12
  │
3 │     let v12 = v10 as bool;
  │               -----------
  │
  = Call stack:
    1. src/main.nr:3:12
```

## `nargo execute --force-brillig`
```
[heck] Circuit witness successfully solved
[heck] Witness saved to target/heck.gz
[heck] Circuit output: Field(1)
```

By the way the following program does not fail: 

```
fn main(v6: Field) -> pub bool {
	v6 as bool
}
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8157

## Summary

This seems to be a big breaking change but I think we should still disallow it. Casting using `as` for  numeric values means truncating or extending a value. When casting a `u32` to `u1` we only keep the least significant byte, so casting 1 to `u1` gives 0, but casting 2 to `u1` gives 0. But when casting `u32` to `bool` it's not clear what should happen: is the rule that 0 is false and 1 is true, or is the least significant byte what's used (similar to `u1`). Or maybe 0 is false, 1 is true and everything else is...? That's my guess of why Rust disallows this and asks to compare with zero, because like that there's no room for confusion.

The other way around, that is, casting from bool to a numeric value, is allowed in Rust and is still allowed in Noir. Here we get 0 for false and 1 for true, but I think that's expected (you wouldn't expect any other values here).

## Documentation

Check one:

# PR Checklist


test_programs/compile_failure/cast_numeric_to_bool/src/main.nr
```diff
@@ -0,0 +1,10 @@
+fn main() {
+    let x = 1;
+    let _ = x as bool;
```
vezenovm, line 3: We will also want to block this in SSA. The linked issue shows an SSA snippet.

vezenovm:
> It seems that before this PR casting a numeric value to bool actually truncates the number to one bit, then that bit is the one used for the bool. That means that casting 2 to bool gives false 😨 This is why [AztecProtocol/aztec-packages#14593](https://github.com/AztecProtocol/aztec-packages/pull/14593) is likely failing.

Ahhh! 😨

Your intuition about the failure makes sense. This just reiterates to me that we should disallow casting to bool.

test_programs/compile_failure/cast_numeric_to_bool/src/main.nr
asterite: I was about to implement this now but I noticed that the `Cast` SSA instruction takes a numeric type. So technically there's no way to cast to bool in SSA, only to `u1`. But now that it's disallowed in the frontend I think all is good.

vezenovm: > `Cast` SSA instruction takes a numeric type. So technically there's no way to cast to bool in SSA, only to `u1`. But now that it's disallowed in the frontend I think all is good.

Oh ok so for casting a numeric to a bool we are good then. Can you still trigger the ACIR vs. Brillig difference by directly writing the SSA but with `as u1`? If so this is a separate bug though and we can do it in a follow-up. In general I think we should lean into the SSA always being well formed even if the frontend prevents it (https://github.com/noir-lang/noir/issues/8705).

asterite: I just checked and with `u1` both programs give 0 as a result.

What's strange is that with `bool` and running in `master` I also get `0` for the brillig case, so I don't know why for the OP it gives 1.

asterite:
The "performance improvements" are just because I had to remove some code from test programs that checked that casting numbers to bools worked well.

TomAFrench (approved):

Issue #8629, opened 2025-05-22
Breaking out of a loop inside a `let` statement should be disallowed


rkarabut:
### Aim

A program like this:

```noir
unconstrained fn main() -> pub (Field, Field) {
        let mut a = (1, 1);
        loop {
            let mut b = {
                break;
                (0, 0)
            };
            a = (b.0, b.1);
        }
        a
}
```

would fail with

```
error: Object type is unknown in field access
  ┌─ src/main.nr:9:20
  │
9 │             a = (b.0, b.1);
  │                    - Type must be known by this point
  │
  = Try adding a type annotation for the object type before this expression

error: Object type is unknown in field access
  ┌─ src/main.nr:9:25
  │
9 │             a = (b.0, b.1);
  │                         - Type must be known by this point
  │
  = Try adding a type annotation for the object type before this expression
```

As breaking/continuing before the final assignment would always lead to this, it would probably be better to error out earlier on using `break` inside a let statement.

### Expected Behavior

An error on usage of `break` inside a `let` statement.

### Bug

```
$ nargo execute --force
warning: Unreachable statement
  ┌─ src/main.nr:6:17
  │
6 │                 break;
  │                 ------ Any code following this expression is unreachable
7 │                 (0, 0)
  │                 ------ Unreachable statement
  │

error: Object type is unknown in field access
  ┌─ src/main.nr:9:20
  │
9 │             a = (b.0, b.1);
  │                    - Type must be known by this point
  │
  = Try adding a type annotation for the object type before this expression

error: Object type is unknown in field access
  ┌─ src/main.nr:9:25
  │
9 │             a = (b.0, b.1);
  │                         - Type must be known by this point
  │
  = Try adding a type annotation for the object type before this expression
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

asterite:
https://github.com/noir-lang/noir/pull/8712 will fix this. The code will still compile and run fine, like in Rust, it's just that the loop will never execute because of the `break`, and the output of the program will be `(1, 1)`.

PR #8708, opened 2025-05-29
fix(fuzz): Prevent breaking/continuing out from let blocks in the AST fuzzer

rkarabut:
# Description

While generating loops, the fuzzer would often place `break` inside a let stament in a loop, leading to https://github.com/noir-lang/noir/issues/8629. This should be a quick fix for these appearances.

## Documentation\*

Check one:

# PR Checklist\*


aakoshh (approved):
👍

PR #8712, opened 2025-05-29
fix: stop inserting instructions after break and continue

### LINKED ISSUE #8629: Breaking out of a loop inside a `let` statement should be disallowed

### LINKED ISSUE #8710: In Brillig, conditional break doesn't seem to work
### Aim

The following program:
```
fn main() {
    unsafe {
        func_1();
    }
}

unconstrained fn func_1() {
    loop {
        if true {
            break;
        }
        println("bug");
    }
}
```

prints out "bug".

Removing the `if true` and making the break unconditional does prevent `println` from triggering. Surrounding the call with `comptime` also makes it work as expected.

### Expected Behavior

There should be no output.

### Bug

```
$ nargo execute --force
bug
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #8770: Compiler crash: `break` in assignment
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8728
Related to https://github.com/noir-lang/noir/issues/8629 and https://github.com/noir-lang/noir/pull/8712

```rust
    let mut a: [&mut bool; 2] = [(&mut true), (&mut false)];
    loop {
        a = {
            break;
            [a[1], a[0]]
        };
    }
    *a[0]
}
```

### Expected Behavior

It should compile and return `true`, ie. not modify the `a[0]`. 

### Bug

```console
Message:  internal error: entered unreachable code: assign: Expected lhs and rhs values to match but found Leaf(Mutable(Id(5), Array([Reference(Numeric(Unsigned { bit_size: 1 }))], 2))) and Branch([])
Location: compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs:1098
```

### To Reproduce

See above.

### Workaround

None

### Workaround Description

https://github.com/noir-lang/noir/pull/8708 already changed the fuzzer not to generate such `let` statements, but we forgot about assignments.

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+7e6d7cdb32c806aa0622c1c73e2255a79d93637d (git version hash: 49d8f8206799a3d346d8cf990d14029672ede096, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8710
Resolves #8629
Resolves #8770

## Summary

Now `FunctionBuilder` tracks whether the current block is closed. If so, no more instructions can be added to it. Then, on the codegen side, when we codegen `break` or `continue` we close the current block.

A few cases also need to check if the current block is closed to prevent trying to insert instructions via functions like `insert_allocate` and so on which check that a result is returned (but none is returned if the instruction is removed because the block is closed).

Finally, I realized that a `break` doesn't have a type (or, well, the type is a type variable) and that type was given to a block if it had an unconditional break. That's not correct. The correct thing to do would be to give the block the type "no return" or `!` in Rust, but we don't have that. Instead, I chose unit as a type. At least it prevents code like `1 + { break; }` to compile, and also makes #8629 give a better error.

To give a bit more detail on the last paragraph, the type of `break` should be `!` by itself but, because we don't have that I chose `()`. However, in Rust this code:

```noir
let a = { break; 1 };
```

has `a` be `i32`, not `!`. That is, the type of a block is always the type of the last statement, regardless of it having a break/continue or not. I chose to do the same thing in Noir. It maybe leads to simpler error messages and more intuitive types.

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/ssa/function_builder/mod.rs
```diff
@@ -40,6 +41,8 @@ use super::{
 pub struct FunctionBuilder {
     pub current_function: Function,
     current_block: BasicBlockId,
+    current_block_closed: bool,
+    closed_blocked: HashSet<BasicBlockId>,
```
jfecher, line 45: What happens if we consider the block closed when we insert a terminator in general? Have we tried that approach?

jfecher: Ok, let me know which approach you think would be better - I just wanted to flag that case in case it wasn't considered.
I'd be okay with merging the original approach if you think it is better.

compiler/noirc_evaluator/src/ssa/function_builder/mod.rs
```diff
@@ -3,6 +3,7 @@ pub mod data_bus;
 use std::{borrow::Cow, collections::BTreeMap, sync::Arc};
 
 use acvm::{FieldElement, acir::circuit::ErrorSelector};
+use im::HashSet;
```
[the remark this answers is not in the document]
asterite: Thanks! I probably imported the first suggestion without realizing it. I changed it to FxHashSet.

PR #8720, opened 2025-05-29
fix: cast signed to u1

### LINKED ISSUE #8704: Crash when casting `i32` to `u1`
### Aim

This code:

```noir
fn main() {
    let x: i32 = 1;
    let _ = x as u1;
}
```

Crashes the compiler:

```
Message:  Invalid bit size for signed numeric type: 1. Expected one of 8, 16, 32, 64 or 128.
Location: compiler/noirc_evaluator/src/ssa/function_builder/mod.rs:552
```

### Expected Behavior

It should compile.

### Bug

It doesn't.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8704

## Documentation

Check one:

# PR Checklist


test_programs/compile_success_empty/cast_signed_to_u1/src/main.nr
```diff
@@ -0,0 +1,4 @@
+fn main() {
+    let x: i32 = 1;
+    let _ = x as u1;
```
jfecher, line 3: Can you expand this a tad?

Might as well add tests asserting e.g. `1 as u1 == 1, 2 as u1 == 0, 3 as u1 == 1`

jfecher (approved):

test_programs/compile_success_empty/cast_signed_to_u1/src/main.nr
asterite: Good idea! I tried it but was lazy to write a test.

Now that I did, it was a bit strange. Casting 2 as i32 to u1 gave me zero, at least when I printed it. But when I compared it with zero it failed. I don't understand why. But now I changed the casting to how it was, except that we now allow again the `i1` type in SSA. That works... but I still don't know why directly casting to `u1` didn't work (I'll take a look later).

asterite: I'll leave it as it is now. It was working fine before we started validating numeric types, but this is the second time this validation was too restrictive. The previous case was for i128, which doesn't exist in Noir but is used in SSA just fine, and now it's for i1.

asterite:
Ugh, apparently the commit wasn't pushed. I'll open a follow-up PR...

PR #8724, opened 2025-05-29
fix(licm): Preserve semantic ordering of side-effectual instructions when hoisting

### LINKED ISSUE #8715: Should not hoist instruction dependent upon predicates when there are side effectual instructions before it
### Aim

Take this example:
```noir
// x = 2, y = 3
fn main(x: u32, y: u32) {
    simple_loop(4, x, y);
}
fn simple_loop(upper_bound: u32, x: u32, y: u32) {
    for _ in 0..upper_bound {
        let mut z = x * y;
        z = z * x;
        println("hello world");
        assert_eq(z, 2);
    }
}
```

However, 

### Expected Behavior

We should print "hello world" a single time and then error out with a failed constraint.

### Bug

Executing this code we will give the following error:
```
error: Failed constraint
   ┌─ src/main.nr:21:19
   │
21 │         assert_eq(z, 2);
   │                   ----
   │
   = Call stack:
     1. src/main.nr:6:5
     2. src/main.nr:21:19
```
But no print output

### To Reproduce

1. Run the snippet above with `nargo execute`

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Blocker

### Blocker Context

This is a critical bug. 

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+00d636c75d21a699e4561a7e15e4927ee89ce7ef (git version hash: 00d636c75d21a699e4561a7e15e4927ee89ce7ef, is dirty: false)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #8715 

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -975,6 +997,63 @@ fn can_be_hoisted(
     }
 }
 
+/// Indicates if the instruction has a side effect, ie. it can fail, or it interacts with memory.
+fn has_side_effects(instruction: &Instruction, dfg: &DataFlowGraph) -> bool {
```
asterite, line 1001: I guess this is the same as the one in constant_folding, right? If so, maybe it could be moved to `Instruction`. It was previously moved here because it was the only place it was used, but now it could maybe move back to where it was.

vezenovm: Yeah it is the same. I was unsure on whether we wanted to keep these methods separate still and just duplicate the code within the pass. The semantics of a side-effectual instructions shouldn't matter per pass like deduplicated vs. hoisting though, so I can de-duplicate this method.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -2421,4 +2448,119 @@ mod control_dependence {
...
+        let mut ssa = ssa.loop_invariant_code_motion();
+        ssa.normalize_ids();
+
+        let got = ssa
+            .interpret(vec![
+                from_constant(2_u128.into(), NumericType::Unsigned { bit_size: 32 }),
+                from_constant(3_u128.into(), NumericType::Unsigned { bit_size: 32 }),
+            ])
+            .expect_err("Should have error");
+        dbg!(&got);
```
```suggestion
```

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -254,6 +261,8 @@ impl<'f> LoopInvariantContext<'f> {
                             .insert_instruction_and_results(inc_rc, *block, None, call_stack);
                     }
                 } else {
+                    let dfg = &self.inserter.function.dfg;
+                    self.current_block_impure = dfg[instruction_id].has_side_effects(dfg);
```
asterite: Can it happen that `current_block_impure` was `true` before reaching this line, and this sets it to `false`?

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -285,6 +294,17 @@ impl<'f> LoopInvariantContext<'f> {
...
+        let dfg = &self.inserter.function.dfg;
+        // When hoisting a control dependent instruction, if a side effectual instruction comes in the predecessor block
+        // of that instruction we can no longer hoist the control dependent instruction.
+        // This is important for maintaining ordering semantic correctness of the code.
+        self.current_block_impure = all_predecessors.iter().any(|block| {
+            dfg[*block]
+                .instructions()
+                .iter()
+                .any(|instruction| dfg[*instruction].has_side_effects(dfg))
+        });
```
asterite, line 311: Can it happen that `current_block_impure` was `true` before reaching this line, and this sets it to `false`?

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: Good call. I will gate this to only check side effects if the block is currently labelled as pure.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: In this case we just reset `current_block_impure` to false so this should not be possible. However, I will add a sanity assertion.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: done

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: done

TomAFrench:
> This can lead much larger issues if we have oracles that we expect to be executed that are skipped by hoisting inappropriately.

Oracle calls are being hoisted?

Issue #8727, opened 2025-05-30
Execution crash: value is not typed as brillig usize


aakoshh:
### Aim

I found during working on references that the following program crashes Brillig:

```rust
global G_A: (str<2>, [str<2>; 2], [bool; 0]) = ("IU", ["TC", "JC"], []);
unconstrained fn main() -> pub str<2> {
    (*func_2(G_A)[0])
}
unconstrained fn func_2(mut b: (str<2>, [str<2>; 2], [bool; 0])) -> [&mut str<2>; 1] {
    let _c: [str<2>; 1] = [b.0];
    [(&mut b.0)]
}
```

### Expected Behavior

It should return `"IU"`. 

### Bug

```console
❯ cargo run -q -p nargo_cli -- execute
The application panicked (crashed).
Message:  value is not typed as brillig usize
Location: /Users/aakoshh/Work/aztec/noir/acvm-repo/brillig_vm/src/memory.rs:62
```

### To Reproduce

See above. 

NB `compile` works, only `execute` fails. It stops crashing if for example I remove the `, [bool; 0]` from `G_A`. 

### Workaround

None

### Project Impact

None

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+7e6d7cdb32c806aa0622c1c73e2255a79d93637d (git version hash: 7e6d7cdb32c806aa0622c1c73e2255a79d93637d, is dirty: true)

### Would you like to submit a PR for this Issue?

None

asterite:
Apparently this was fixed by #8801

@aakoshh @vezenovm Do you think we should add a regression test for this, or the one in #8801 also covers this case?

vezenovm:
The regression in #8801 is similar, but I guess for this one the array reference input comes from a parameter vs. a constant, so doesn't hurt to have another test.

asterite:
Cool, I'll send one soon.

Issue #8723, opened 2025-05-29
Compiler crash: condition value is not a boolean


aakoshh:
### Aim

The following program crashes the compiler:

```rust
fn main() -> pub Field {
    unsafe {
        baz(true)
    }
}
unconstrained fn foo(c: bool) -> [(Field, &mut bool); 1] {
    if c {
        [(0, &mut false)]
    } else {
        [(1, &mut true)]
    }
}
unconstrained fn baz(c: bool) -> Field {
    if (*foo(c)[0].1) {
        2
    } else {
        foo(c)[0].0
    }
}
```

### Expected Behavior

Shouldn't crash, it should return 0, or it should return an error saying it can't return `&mut` from `if` (if that's the resolution in #8722). 

### Bug

```console
❯ cargo run -q -p nargo_cli -- execute 
warning: Unsafe block must have a safety comment above it
  ┌─ src/main.nr:2:5
  │
2 │     unsafe {
  │     ------ The comment must start with the "Safety: " word
  │

The application panicked (crashed).
Message:  condition value is not a boolean: MismatchedBitSize { value_bit_size: 254, expected_bit_size: 1 }
Location: /Users/aakoshh/Work/aztec/noir/acvm-repo/brillig_vm/src/lib.rs:455
```

### To Reproduce

See above. 

The following changes result in a different behaviour:

1. If we change `baz` to not call `foo` again:
```rust
unconstrained fn baz(c: bool) -> Field {
    if (*foo(c)[0].1) {
        2
    } else {
        3
    }
}
```
then instead of crashing we get an assertion failure:
```
error: Assertion failed: 'Bit size for value 254 does not match the expected bit size 1'
   ┌─ src/main.nr:14:10
   │
14 │     if (*foo(c)[0].1) {
   │          ---------
   │
   = Call stack:
     1. src/main.nr:3:9
     2. src/main.nr:14:10

Failed assertion
```

2. If we change `foo` to not have an `if`:
```rust
unconstrained fn foo(c: bool) -> [(Field, &mut bool); 1] {
    [(0, &mut false)]
}
```
then it correctly returns 0 from the execution.

### Workaround

None

### Project Impact

None

### Nargo Version

9b7320908e74377bfd55a8760f16974ef1d07c05

### Would you like to submit a PR for this Issue?

None

vezenovm:
This was resolved by https://github.com/noir-lang/noir/pull/8731.

This code now errors with the following:
```
[...]

Issue #8722, opened 2025-05-29
Compiler crash: returning `&mut` from `if`


aakoshh:
### Aim

The following program crashes the compiler:

```rust
fn main(c: bool) -> pub Field {
    let mut x = 1;
    let mut y = 2;
    let z = if c { &mut x } else { &mut y };
    *z
}
```

### Expected Behavior

Return a compiler error without crashing.

### Bug

```console
❯ cargo run -q -p nargo_cli -- compile
The application panicked (crashed).
Message:  Cannot return references from an if expression
Location: compiler/noirc_evaluator/src/ssa/opt/flatten_cfg/value_merger.rs:68
```

### To Reproduce

See above.

### Workaround

Yes

### Workaround Description

Don't return mutable references from if-then-else.

### Project Impact

None

### Nargo Version

9b7320908e74377bfd55a8760f16974ef1d07c05

### Would you like to submit a PR for this Issue?

None

nventuro:
Reviving this as this is causing us issues (e.g. it makes https://github.com/AztecProtocol/aztec-packages/pull/17024 fail) that directly affect our most important goal (being able to provide nice APIs).

We mostly run into this problem through two scenarios. The first one is where both branches of the `if` hold a reference to the same object:

```noir
struct Foo {
  x: &mut Field,
  y: Field,
}

fn main(mut a: pub Field, b: pub Field, cond: pub bool) -> pub Field {
  let foo = if cond {
    Foo{ x: &mut a, y: b }
  } else {
    Foo{ x: &mut a, y: 5 }
  };

  *(foo.x) + foo.y
}
```

This sounds solvable by merging the references when they are the same. In our specific use case we just so happen to be able to address that on our own by removing the reference and attaching it again (see [this commit](https://github.com/AztecProtocol/aztec-packages/pull/17024/commits/f82194c944f7ff4cdaea3b119d567aef64bf9039)).

The second scenario is when we have an `Option` of such an object, e.g.:

```noir
struct Foo {
  x: &mut Field,
  y: Field,
}

fn main(mut a: pub Field, b: pub Field, cond: pub bool) -> pub Option<Field> {
  let maybe_foo = if cond {
    Option::some(Foo{ x: &mut a, y: b })
  } else {
    Option::none()
  };

  maybe_foo.map(|foo| *(foo.x) + foo.y)
}
```

[...]

PR #8728, opened 2025-05-30
feat(fuzz): Generate references in the AST fuzzer

### LINKED ISSUE #7939: Generate mutable reference parameters
Currently functions only get parameters which are ABI compatible, but the non-entry functions can use `&mut` parameters as well.

aakoshh:
# Description

## Problem\*

Resolves #7939 

## Summary\*

Generates parameter and return types with `&mut`. 

TODO:
* Need to change generation to avoid dynamic indexes to access arrays with references after https://github.com/noir-lang/noir/pull/8888

## Additional Context

Some bugs I discovered during testing this were:
* https://github.com/noir-lang/noir/issues/8727
* https://github.com/noir-lang/noir/issues/8723
* https://github.com/noir-lang/noir/issues/8722
* https://github.com/noir-lang/noir/issues/8729
* https://github.com/noir-lang/noir/issues/8739
* https://github.com/noir-lang/noir/issues/8741
* https://github.com/noir-lang/noir/issues/8748
* https://github.com/noir-lang/noir/issues/8750
* https://github.com/noir-lang/noir/issues/8761
* https://github.com/noir-lang/noir/issues/8755
* https://github.com/noir-lang/noir/issues/8770
* https://github.com/noir-lang/noir/issues/8786
* https://github.com/noir-lang/noir/issues/8803
* https://github.com/noir-lang/noir/issues/8909

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/ast_fuzzer/src/program/func.rs
+++ b/tooling/ast_fuzzer/src/program/func.rs
@@ -368,7 +368,7 @@ impl<'a> FunctionContext<'a> {
                 .choose_producer_filtered(u, typ.as_ref(), |id, (mutable, _, prod)| {
                     *mutable
                         && (typ.as_ref() == prod || !types::is_array_or_slice(prod))
-                        && (!self.in_no_dynamic || !self.is_dynamic(&id))
+                        && (!self.in_no_dynamic || !self.is_dynamic(id))
                 })
                 .map(|id| id.map(VariableId::Local));
         }


tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -817,7 +873,14 @@ impl<'a> FunctionContext<'a> {
             .locals
             .current()
             .variables()
-            .filter_map(|(id, (mutable, _, _))| mutable.then_some(id))
+            .filter(|(_, (mutable, _, typ))| {
+                // Trying to re-assign a mutable reference
```
aakoshh:
```suggestion
                // Trying to re-assign a mutable reference in an if leads to panics or errors
```

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -336,13 +384,27 @@ impl<'a> FunctionContext<'a> {
...
-                None => self.find_global_function_with_signature(u, typ),
-            };
-        }
+        if types::is_function(types::unref(typ)) {
+            if let Type::Reference(typ, _) = typ {
+                // We might have an `&mut &mut fn(...) -> ...`; we recursively peel
+                // off layers of references until we can generate an expression that
+                // returns an `fn`, and then take a reference over it to restore.
+                let expr = self.gen_expr(u, typ, max_depth, flags)?;
+                // If the expression is a read-only ident global ident, then assign a variable first.
```
rkarabut:
```suggestion
                // If the expression is a read-only global ident, then assign a variable first.
```

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -477,12 +542,10 @@ impl<'a> FunctionContext<'a> {
                 src_as_tgt()
             }
             (Type::Reference(typ, _), _) if typ.as_ref() == tgt_type => {
-                let e = if bool::arbitrary(u)? {
-                    Expression::Clone(Box::new(src_expr))
-                } else {
-                    expr::deref(src_expr, tgt_type.clone())
-                };
-                Ok(Some(e))
+                Ok(Some(expr::deref(src_expr, tgt_type.clone())))
```
rkarabut: Why are we only using deref here now?

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -758,14 +821,7 @@ impl<'a> FunctionContext<'a> {
...
         let comptime_friendly = self.is_comptime_friendly();
         let typ = self.ctx.gen_type(u, max_depth, false, false, true, comptime_friendly)?;
-
-        // Temporarily set in_loop to false to disallow breaking/continuing out
-        // of the let blocks (which would lead to frontend errors when reversing
-        // the AST into Noir)
-        let was_in_loop = std::mem::replace(&mut self.in_loop, false);
         let expr = self.gen_expr(u, &typ, max_depth, Flags::TOP)?;
-        self.in_loop = was_in_loop;
-
```
rkarabut: Possibly deleted the break disallow by mistake?

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -817,7 +873,14 @@ impl<'a> FunctionContext<'a> {
             .locals
             .current()
             .variables()
-            .filter_map(|(id, (mutable, _, _))| mutable.then_some(id))
+            .filter(|(_, (mutable, _, typ))| {
+                // Trying to re-assign a mutable reference in an if leads to panics or errors
```
rkarabut: Is that covered with an issue? I guess gotta mention it if so

tooling/ast_fuzzer/src/program/types.rs
```diff
@@ -188,13 +189,49 @@ pub(crate) fn is_reference(typ: &Type) -> bool {
     matches!(typ, Type::Reference(_, _))
 }
 
+/// Check if tye type is a function.
```
rkarabut:
```suggestion
/// Check if the type is a function.
```

tooling/ast_fuzzer/src/program/func.rs
aakoshh: I think this was a remnant from when I didn't understand that `Clone` is only something inserted by the monomorphizer, not a language construct. I thought if we have `let a: &mut [u8; 1]` and we need to call a function that takes `[u8; 1]` then we can do `*a` or `a.clone()`, but since `Clone` is something technical, perhaps related to ownership feature, I thought we should stick to `*a`.

tooling/ast_fuzzer/src/program/func.rs
aakoshh: Right, I think it works now after https://github.com/noir-lang/noir/pull/8712

I had a similar issue with https://github.com/noir-lang/noir/issues/8770 where `break` appeared in assignment. I wanted to use this fix in `gen_assignment` as well, but then Ary fixed the codegen. IIRC I tried it with `let` and it was fine, so I thought we don't need to avoid it any more.

tooling/ast_fuzzer/src/program/func.rs
aakoshh: Yes it's this one: https://github.com/noir-lang/noir/issues/8741
It's not clear that this will be solved, it looks like it will never be allowed, and potentially even have a blanket deny on reassigning anything with a mutable reference in it completely, as discussed in #8783 , so I thought it's more like a general language thing we have to avoid.

aakoshh: But you are right, since the situation is evolving, some link to the issues will be useful, especially if we have to widen the ban. I'll add a comment above the `avoid_*` flag.

PR #8172, opened 2025-04-22
fix!: disallow the `i1` type

TomAFrench:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/pull/8167#pullrequestreview-2785394880

## Summary\*

This PR adds a new parser error for if the `i1` type is constructed.

## Documentation\*

Check one:
- [x] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #8730, opened 2025-05-30
fix: cast signed to u1 follow-up

asterite:
# Description

## Problem

Follow-up to https://github.com/noir-lang/noir/pull/8720

## Summary

I didn't intent to merge that PR like that. I thought I pushed one more commit (the one in this PR) before merging, but apparently not.

See this comment: https://github.com/noir-lang/noir/pull/8720#discussion_r2115791385

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/ssa/function_builder/mod.rs
```diff
@@ -556,10 +556,10 @@ impl std::ops::Index<BasicBlockId> for FunctionBuilder {
 fn validate_numeric_type(typ: &NumericType) {
     match &typ {
         NumericType::Signed { bit_size } => match bit_size {
-            8 | 16 | 32 | 64 | 128 => (),
+            1 | 8 | 16 | 32 | 64 | 128 => (),
```
jfecher, line 559: Why are we allowing `i1` now?
The type would just be a single sign bit so I'm not sure it makes much sense. I saw your previous comment about the inconsistent result when printing zero versus comparing with zero but maybe we could add special logic to casting to avoid the `i1` intermediate?

vezenovm: We ban it in the frontend https://github.com/noir-lang/noir/pull/8172 so we should ban it in SSA as well. Not sure why it is being used, if it is we should move away from it.

asterite: In the end a truncate was missing.

Issue #8705, opened 2025-05-28
SSA Validation Module


vezenovm:
### Problem

We have had several issues come up where someone can cause differences between ACIR/Brillig or other issues by writing malformed SSA (e.g. https://github.com/noir-lang/noir/issues/8089). We may block this SSA being generated by the frontend, however, we do not want to rely on the what the frontend will produce for security reasons. Especially if we ever start to accept SSA from an untrusted source. Examples of untrusted sources could be if we wanted to enable using SSA IR artifacts as part of our compilation flow or custom optimization passes (LLVM allows these). 

In general having a validation pass on our SSA provides a greater security assurance that the SSA we are creating is well formed. Validation passes are common in modern compiler toolchains. e.g., cranelift calls this a [verifier](https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/codegen/src/verifier/mod.rs) and exposes it is as part of its testing https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/docs/testing.md#test-verifier. Validation could even assist developers working on a separate frontend for the Noir SSA. 

### Happy Case

We have a separate validation/verification module (such as cranelift's linked above) to check whether an SSA function is well formed.

We have validation logic scattered in a few places (e.g., when making an instruction, when creating a numeric constant, inserting a cast instruction, and post SSA function generation after the #8089 fix). It would be cleaner and more extensible to have all validation logic in a single pass that can be run on an SSA function. This also follows what cranelift does and I think follows fundamental compiler infrastructure better (e.g. separate passes for separate functionality). 

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

vezenovm:
I'm going to close this issue as the initial issues highlighting breakages in ACIR vs. Brillig from ill formed SSA have been resolved. There are a lot more validation checks we can add to guarantee well formed SSA, however, we can begin to capture these in separate issues.

PR #8731, opened 2025-05-30
fix!: Prevent returning references from if expressions

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/8722

## Summary\*

This was never supported in SSA (we would always panic eventually) but is technically a breaking change since it could previously be simplified away if the if condition was statically known where now we error before that point.

## Additional Context

I tried erroring in Ssa in jf/reference-if-error but this led to `Result` poisoning every method in the `FunctionBuilder` and was generally too large of a code change.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/interpreter/tests/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/interpreter/tests/mod.rs
@@ -245,43 +245,6 @@ fn without_defunctionalize() {
 }
 
 #[test]
-fn keep_repeat_loads_with_alias_store() {
-    let src = "
-    acir(inline) fn main f0 {
-      b0(v0: u1):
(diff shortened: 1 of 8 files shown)


vezenovm (approved):
Could we just make sure we cannot build this in SSA directly as well? You could check during `Function::assert_valid` for now. Eventually I plan to move this into its own validation pass module https://github.com/noir-lang/noir/issues/8705.

vezenovm:
> @vezenovm I ended up removing the SSA validation. We were hitting it many times while running `noir-contracts` although I'm not sure why exactly. When I tested locally I can't compile that crate at all.

Ok sounds good. We can follow-up with it.

PR #8738, opened 2025-05-30
fix: Add missing cases for finding function values in `find_functions_as_values`

jfecher:
# Description

## Problem\*

Resolves "Missing ArraySet case in Defunctionalize pass" [issue](https://audithub.veridise.com/app/organizations/85/projects/277/project-viewer?version=433&bpt=1&issue=15&thread=2873) from Veridise

## Summary\*

Instead of adding only ArraySet I figured it'd be more future-proof to match on the values from any instruction.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
@@ -340,9 +340,6 @@ fn find_variants(ssa: &Ssa) -> Variants {
 }
 
 /// Finds all literal functions used as values in the given function
-///
-/// It is assumed that function values will only ever be used in a call instruction
-/// or a store instruction.
 fn find_functions_as_values(func: &Function) -> BTreeSet<FunctionId> {
     let mut functions_as_values: BTreeSet<FunctionId> = BTreeSet::new();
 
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
```diff
@@ -1304,4 +1297,32 @@ mod tests {
         }
         ");
     }
+
+    #[test]
+    fn find_functions_as_values_finds_function_in_arrayset() {
```
vezenovm, line 1302:
```suggestion
    fn find_functions_as_values_finds_function_in_array_set() {
```
cspell nit

compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
```diff
@@ -1304,4 +1297,32 @@ mod tests {
...
+          b0(v0: u32):
+            return v0
+        }
+        "#;
+
+        let ssa = Ssa::from_str(src).unwrap();
+        let main = ssa.main();
+
+        let functions = super::find_functions_as_values(main);
+        assert_eq!(functions.len(), 2); // f0 and f1
```
vezenovm, line 1326: Could assert on the IDs to make this stronger but its non-blocking

jfecher:
@asterite I think the unit test is sufficient in this case

PR #8740, opened 2025-05-30
chore: repro for very strange bug

asterite:
# Description

## Problem

The compiler fails to compile a program if we add/remove unused variables.

## Summary

If we compile the `serialize_1` as it is in this PR, it fails to compile. But if we uncomment the `let a1 = quote {};` variable, it compiles.

Note that if it's `let a1 = 1;` then it **doesn't** compile. For some reason if we use `quote {}` it compiles.

Then, if we keep uncomment variables the program will compile or not compile. It seems a bit random.

This makes me think there's a bug somewhere in the comptime code. We already [observed](https://github.com/noir-lang/noir/issues/8545#issuecomment-2887599812) that if some code is generated via macros instead of written by hand then it might lead to slower compilation, when it shouldn't, so maybe all of these things are related. I wonder if something is leaking from the comptime interpreter that's modifying something outside of it, some global state (or bindings, type vars, not sure).

---

Another way: instead of uncommenting any of those variables, add this to `lib.nr` for the standard library:

```noir
fn foo() {
    comptime {
        let _ = quote {};
    }
}
```

then the code compiles.

## Documentation

Check one:

# PR Checklist\*

### DIFF
--- a/noir_stdlib/src/convert.nr
+++ b/noir_stdlib/src/convert.nr
@@ -175,23 +175,24 @@ comptime fn generate_as_primitive_impls(_: FunctionDefinition) -> Quoted {
     let mut impls = &[];
     for type1 in types {
         for type2 in types {
-            let body = if type1 == type2 {
-                quote { self }
-            } else if type1 == quote { bool } {
-                quote { self != 0 }
-            } else {
-                quote { self as $type1 }
-            };
+            // Uncomment one, the serialize_1 program compiles.
+            // Uncomment two, it doesn't compile anymore.
+            // Uncomment three, it compiles again.
+            // Uncomment four, it still compiles.
+            // let a1 = quote {};
+            // let a2 = quote {};
+            // let a3 = quote {};
+            // let a4 = quote {};
 
-            let as_primitive_impl = quote {
+            impls = impls.push_back(
+                quote {
                 impl AsPrimitive<$type1> for $type2 {
                     fn as_(self) -> $type1 {
-                        $body
+                        $crate::default::Default::default()
                     }
                 }
-            };
-
-            impls = impls.push_back(as_primitive_impl);
+            },
+            );
         }
     }
     impls.join(quote {})


asterite:
I was able to reproduce this with a much smaller code.

For this program:

```noir
trait Serialize {
    let Size: u32;

    fn serialize(self) -> [Field; Self::Size];
}

impl<A, B> Serialize for (A, B)
where
    A: Serialize,
    B: Serialize,
{
    let Size: u32 = <A as Serialize>::Size + <B as Serialize>::Size;

    fn serialize(self: Self) -> [Field; Self::Size] {
        let _ = self.0.serialize();
        let _ = self.1.serialize();
        [0; Self::Size]
    }
}

impl<T, let N: u32> Serialize for [T; N]
where
    T: Serialize,
{
    let Size: u32 = <T as Serialize>::Size * N;

    fn serialize(self: Self) -> [Field; Self::Size] {
        [0; Self::Size]
    }
}

impl Serialize for Field {
    let Size: u32 = 1;

    fn serialize(self) -> [Field; Self::Size] {
        [self]
    }
}

fn main() {
    let x = (((1, [2, 3, 4]), [5, 6, 7, 8]), 9);
    let _ = x.serialize();
}
```

If we change `noir_stdlib/src/lib.nr` to be empty, the program above compiles.

If we put this in `noir_stdlib/src/lib.nr` then it stops compiling:

```noir
pub fn foo() {
    let _ = 1;
    let _ = 1;
}
```

Hopefully this will make finding the issue much easier 🤞 

But also this means that it's likely unrelated to comptime stuff, as it happens with non-comptime stuff too (the `foo` above isn't comptime).

asterite:
It's also worth noting that this happens if `foo` above is in the standard library. If we move it from the standard library to the program, then it works fine.

asterite:
Oh, and it also seems to happen if we move `foo` to a dependency of the program. So it's not the standard library, it's any dependency.

asterite:
Closed in favor of this issue: https://github.com/noir-lang/noir/issues/8780

PR #8744, opened 2025-05-30
fix: delay associated constants resolution

### LINKED ISSUE #8548: Cannot resolve `TraitAsPath`s to define an associate constant in terms of another associated constant.
I'd like for my `Serialize` trait to be able to reference the associate constants of the impls for other types. As an example we have the program below.

```noir
trait Serialize {
    let N: u32;

    fn serialize(self) -> [Field; N];
}

// Testing using primitive types

impl Serialize for Field {
    let N: u32 = 1;

    fn serialize(self) -> [Field; Self::N] {
        [self]
    }
}

pub struct Foo {
    inner: Field,
}

impl Serialize for Foo {
    let N = <Field as Serialize>::N;

    fn serialize(self) -> [Field; Self::N] {
        [0; <Field as Serialize>::N]
    }
}


// Testing using no primitive types.

pub struct Bar {}

pub struct Baz {
    inner: Bar
}

impl Serialize for Bar {
    let N: u32 = 1;

    fn serialize(self) -> [Field; Self::N] {
        [1]
    }
}

impl Serialize for Baz {
    let N = <Bar as Serialize>::N;

    fn serialize(self) -> [Field; Self::N] {
        [0; <Bar as Serialize>::N]
    }
}

fn main() {}
```

This program fails typechecking with the below:

```
% nargo check                                                                                                                   ~/noir/temp ab/access_trait_associated_constants_via_self_general_solution tom-box
error: No matching impl found for `Field: Serialize<N = _>`
   ┌─ src/main.nr:20:23
   │
20 │     let N = <Field as Serialize>::N;
   │                       --------- No impl for `Field: Serialize<N = _>`
   │

error: No matching impl found for `Bar: Serialize<N = _>`
   ┌─ src/main.nr:43:21
   │
43 │     let N = <Bar as Serialize>::N;
   │                     --------- No impl for `Bar: Serialize<N = _>`
   │

Aborting due to 2 previous errors
```

asterite:
# Description

## Problem

Resolves #8548

## Summary

Now instead of directly resolving the type of an associated constant in `define_function_metas` of a trait impl, we set those types to type variables (inside named generics). Then, in `collect_trait_impl`, we solve the actual types and bind the type variables to those.

## Additional Context

This doesn't work (meaning: the type resolution isn't delayed in this case) if the associated type/constant is specified in the impl trait generics. However, that shouldn't be allowed in the first place. See https://github.com/noir-lang/noir/issues/8751

## Documentation

Check one:

# PR Checklist


noir_stdlib/src/convert.nr
```diff
@@ -175,15 +175,15 @@ comptime fn generate_as_primitive_impls(_: FunctionDefinition) -> Quoted {
...
                 impl AsPrimitive<$type1> for $type2 {
                     fn as_(self) -> $type1 {
                         self as $type1
                     }
                 }
-            },
-            );
+            };
+
+            impls = impls.push_back(trait_impl);
```
TomAFrench, line 186: This is listed as a workaround. Do we have an issue for the issue this is working around?

asterite: Kind of. It's this bug: https://github.com/noir-lang/noir/pull/8740
I'll create an issue for this.

asterite: https://github.com/noir-lang/noir/issues/8780

I'll debug it now, now that I found a way to reduce it a lot. We could hold a bit on merging this, then.

TomAFrench: :shrug: I think that this change is perfectly good. It's not an awkward workaround at all.

asterite: Sounds good!

PR #8743, opened 2025-05-30
fix(mem2reg): Keep last stores used in MakeArray

vezenovm:
# Description

## Problem\*

Resolves #8739 
Resolves #8755
Resolves #8761

## Summary\*

We have an optimization that runs through our remaining last stores and checks whether they are used anywhere. If not, we remove those last stores. This optimization was not accounting for references used in arrays. 

## Documentation\*

Check one:
- [X] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [X] I have tested the changes locally.
- [X] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #8745, opened 2025-05-30
fix(ssa): Experiment with removing the last stores optimization in mem2reg

vezenovm:
# Description

## Problem\*

As as an alternative to https://github.com/noir-lang/noir/pull/8743 I just want to see the effects of removing this optimization where we remove the remaining last stores.

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
There are several regressions but largest bump looks to be contained to `brillig_block_parameter_liveness`. Perhaps we can simplify this optimization, but we shouldn't remove it outright.

PR #8747, opened 2025-06-02
chore: bump external pinned commits

noirwhal:
Automated update of the pinned commit of [aztec-packages](https://github.com/AztecProtocol/aztec-packages) repository against which we run benchmarks.


70nyIT:
is it possible that with this latest merge, in the docs the left menu disappeared? Few hours age was there, now I Cannot see anymore. See for example https://noir-lang.org/docs/noir/concepts/data_types/references

PR #8757, opened 2025-06-02
chore: add regression test for purity analysis issue.

TomAFrench:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
@@ -2020,4 +2020,664 @@ mod test {
         }
         ");
     }
+
+    #[test]
+    fn unknown_issue_regression_test() {
+        // Regression test for an issue discovered in https://github.com/AztecProtocol/aztec-packages/pull/14492
+        // Marking `f1` as `predicate_pure` instead of `impure` results in an invalid storage write.
+        let src = r#"
+        g0 = u32 44
+        g1 = u32 13
+        g2 = u32 6
+        g3 = u32 3
+        
+        brillig(inline) impure fn constructor f0 {
+          b0(v4: Field, v5: Field, v6: Field):
+            v12 = allocate -> &mut u1
+            store u1 0 at v12
+            v14 = allocate -> &mut Field
+            store Field 0 at v14
+            v16 = allocate -> &mut Field
+            store Field 2 at v16
+            v19 = call f5() -> Field
+            v21 = call f6(v19) -> [(u1, Field); 1]
+            v23 = array_get v21, index u32 0 -> u1
+            v25 = array_get v21, index u32 1 -> Field
+            v26 = not v23
+            v27 = cast v23 as Field
+            v28 = cast v26 as Field
+            v29 = mul v27, v25
+            constrain v23 == u1 1
+            v32 = call f7(v19) -> [(u1, Field); 1]
+            v33 = array_get v32, index u32 0 -> u1
+            v34 = array_get v32, index u32 1 -> Field
+            v35 = not v33
+            v36 = cast v33 as Field
+            v37 = cast v35 as Field
+            v38 = mul v36, v34
+            constrain v33 == u1 1
+            v40 = call f8(u32 0, u32 1) -> [Field; 1]
+            v41 = array_get v40, index u32 0 -> Field
+            v42 = truncate v41 to 32 bits, max_bit_size: 254
+            v43 = allocate -> &mut u1
+            v44 = allocate -> &mut Field
+            v45 = allocate -> &mut Field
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -1960,4 +1961,354 @@ mod test {
...
+            v134 = array_get v129, index v11 -> Field
+            call f12(v133, v134)
+            v136 = unchecked_add v11, u32 1
+            jmp b13(v136)
+          b15():
+            v131 = call f5() -> Field
+            call f10(v131)
+            return v129
+        }
+        brillig(inline) predicate_pure fn new f1 {
```
TomAFrench, line 1979: Marking this as `impure` causes the test to pass.

TomAFrench:
@aakoshh I'm getting a flake on `noir_ast_fuzzer::smoke arb_program_can_be_executed` in this PR.

aakoshh:
Captured in https://github.com/noir-lang/noir/issues/8975

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -311,13 +311,24 @@ impl<'brillig> Context<'brillig> {
...
+                    // This also applies to calls that return arrays.
+                    if runtime_is_brillig
+                        && matches!(
+                            instruction,
+                            Instruction::MakeArray { .. } | Instruction::Call { .. }
+                        )
+                    {
+                        for &value in cached {
+                            let value_type = dfg.type_of_value(value);
+                            if value_type.contains_an_array() {
```
guipublic: if 'value' is a reference to an array, shouldn't we also increase the ref count of the array?

TomAFrench: This shouldn't be possible as returning a reference would prevent the function from being pure.

jfecher: Maybe we want something like "is an array" here?

IncrementRc below relies on us passing an array type, not something containing an array. Besides another array though, the only other type that can contain an array is a reference like you said, which can't be cached. It does seem more clear though if we made this check more explicit I think. It'd be easier to read since we'd be relying less on external reasoning.

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -1960,4 +1947,87 @@ mod test {
         }
         ");
     }
+
+    #[test]
+    fn unknown_issue_regression_test() {
+        // Regression test for an issue discovered in https://github.com/AztecProtocol/aztec-packages/pull/14492
+        // Marking `f1` as `predicate_pure` instead of `impure` results in an invalid storage write.
```
jfecher: Can this name be more descriptive?
Do we still not know what went wrong or is it definitely the fact we were deduplicating a function call without an inc_rc?

jfecher: Ah, sorry just realized you didn't ask for a review yet

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -712,18 +723,23 @@ impl<'brillig> Context<'brillig> {
...
+                return;
+            };
+
+            let instruction = match &function.dfg[*value] {
                 Value::Instruction { instruction, .. } => &function.dfg[*instruction],
                 _ => return,
             };
 
-            if matches!(instruction, Instruction::MakeArray { .. }) {
+            if matches!(instruction, Instruction::MakeArray { .. } | Instruction::Call { .. }) {
```
guipublic, line 742: Just noting this for reference:
An alternative would be to increase the reference count of an array referenced by a Store instruction, instead of preventing the call here to  be de-duplicated.
We would need to check if this cover more cases.

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
TomAFrench: Ah yeah, I forgot to update the name. Yeah it's definitely the fact that deduplicating function calls can result in an equivalent scenario to deduplicating make_array without an inc_rc.

guipublic (approved):
LGTM
Some names could be re-worked as suggested.

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
TomAFrench: Agreed, I've added an `is_array` function for this.

Issue #8754, opened 2025-06-02
SSA pass result changes from defunctionalize to inlining simple functions


aakoshh:
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8728 although it doesn't have anything to do with references. I think the change was that due to some oversight I did not generate mutable parameters in `main`.

The following program changes its output when evaluated with the SSA interpreter:

```rust
unconstrained fn main(mut b: [Field; 1]) -> pub [Field; 1] {
    b[0] = (-79127210503643033360885884348279364033 * b[0]);
    b
}
```

Inputs:
```toml
b = [
    "0x2303837d532325828bfaacc490e4343b65cc94a35dc7910110bfb11c6a1d890f",
]

```

### Expected Behavior

Results shouldn't change.

### Bug

```console
❯ cargo run -q -p nargo_cli -- interpret --ssa-pass Defunctional --ssa-pass "Inlining simple"
--- Interpreter result after Defunctionalization (step 2):
Ok([ArrayOrSlice(ArrayValue { elements: Shared(RefCell { value: [Numeric(Field(5981576186306977734223805834677592686111272874625127943106928542629024074054))] }), rc: Shared(RefCell { value: 1 }), element_types: [Numeric(NativeField)], is_slice: false })])
---
--- Interpreter result after Inlining simple functions (step 3):
Ok([ArrayOrSlice(ArrayValue { elements: Shared(RefCell { value: [Numeric(Field(-1749910088320338237443070351801816902038064909777356655622258341294090347381))] }), rc: Shared(RefCell { value: 1 }), element_types: [Numeric(NativeField)], is_slice: false })])
---
```

Note that `b[0]` is `5981576186306977734223805834677592686111272874625127943106928542629024074054` in the first and `-1749910088320338237443070351801816902038064909777356655622258341294090347381` in the second case.

### To Reproduce

See above.

### Workaround

None

### Project Impact

None

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+7e6d7cdb32c806aa0622c1c73e2255a79d93637d (git version hash: 393d8c496ace9fd30437ee3088cff6b8e78953d0, is dirty: true)

### Would you like to submit a PR for this Issue?

None

PR #8762, opened 2025-06-02
fix: Create SSA interpreter arguments from scratch for each invocation

aakoshh:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/8754

## Summary\*

The SSA interpreter `Value` is `Clone` but the arrays share their underlying storage with their clones. This meant that the inputs I prepared for it were modified after the first pass, affecting the second pass.

## Additional Context

This affects the `pass_vs_prev` fuzz target and the `nargo interpret` command, and comes out where `main` has a `mut` array parameter, which it modifies.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/interpreter/value.rs
+++ b/compiler/noirc_evaluator/src/ssa/interpreter/value.rs
@@ -13,6 +13,7 @@ use crate::ssa::ir::{
 
 use super::IResult;
 
+/// Be careful when using `Clone`: the `ArrayValue` is backed by a `Shared` data structure.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Value {
     Numeric(NumericValue),
--- a/tooling/ast_fuzzer/fuzz/src/lib.rs
+++ b/tooling/ast_fuzzer/fuzz/src/lib.rs
@@ -176,7 +176,7 @@ pub fn compare_results_interpreted(
         eprintln!(
             "---\nSSA Inputs:\n{}",
             inputs
-                .input_values
+                .input_values()
                 .iter()
                 .enumerate()
                 .map(|(i, v)| format!("{i}: {v}"))
--- a/tooling/ast_fuzzer/src/compare/interpreted.rs
+++ b/tooling/ast_fuzzer/src/compare/interpreted.rs
@@ -42,8 +42,7 @@ pub struct CompareInterpreted {
     /// We could generate random input for the SSA directly, but it would
     /// make it more difficult to use it with `nargo` if we find a failure.
     pub input_map: InputMap,
-    /// Inputs for the `main` function in the SSA.
-    pub input_values: Vec<ssa::interpreter::value::Value>,
+
     /// Options that influence the pipeline, common to both passes.
     pub options: CompareOptions,
     pub ssa1: ComparePass,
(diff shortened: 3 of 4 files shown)


compiler/noirc_evaluator/src/ssa/interpreter/value.rs
```diff
@@ -13,6 +13,7 @@ use crate::ssa::ir::{
 
 use super::IResult;
 
+/// Be careful when using `Clone`: the `ArrayValue` is backed by a `Shared` data structure.
```
jfecher: References are too

jfecher (commented):
Something which may be useful is that the interpreter itself handles a very similar case already: each invocation of brillig from acir code should receive conceptually fresh arrays with the rc set to 1, etc. The interpreter calls `reset_array_state` for this but if we want something to work across multiple runs of the interpreter we may have to handle references as well (the interpreter doesn't handle this since you cannot pass references to brillig from acir).

aakoshh:
@jfecher thanks reminding me of the `reset_array_state`. I don't think it would help me in its current form though, because it only resets the ref-count, whereas in my case the value itself was modified. This is the output after modifying `define` to print what it does:

```console
❯ cargo run -q -p nargo_cli -- interpret --ssa-pass Initial --ssa-pass "(step 1)" --show-ssa-pass Initial --show-ssa-pass "(step 1)"
After Initial SSA:
brillig(inline) fn main f0 {
  b0(v0: [u1; 1]):
    v2 = allocate -> &mut [u1; 1]
    store v0 at v2
    v3 = load v2 -> [u1; 1]
    v5 = array_get v3, index u32 0 -> u1
    jmpif v5 then: b1, else: b2
  b1():
    v7 = load v2 -> [u1; 1]
    v9 = array_set v7, index u32 0, value u1 0
    store v9 at v2
    jmp b3(Field 1)
  b2():
    jmp b3(Field 0)
  b3(v1: Field):
    return v1
}

DEF v0 := rc1 [u1 true]
DEF v2 := *v2 = None
DEF v3 := rc1 [u1 true]
DEF v5 := u1 true
DEF v7 := rc1 [u1 true]
DEF v9 := rc1 [u1 false]
DEF v1 := Field 1
Ok([Numeric(Field(1))])
---
After Removing Unreachable Functions (step 1):
brillig(inline) fn main f0 {
  b0(v0: [u1; 1]):
    v2 = allocate -> &mut [u1; 1]
    store v0 at v2
    v3 = load v2 -> [u1; 1]
    v5 = array_get v3, index u32 0 -> u1
    jmpif v5 then: b1, else: b2
  b1():
    v7 = load v2 -> [u1; 1]
    v9 = array_set v7, index u32 0, value u1 0
    store v9 at v2
    jmp b3(Field 1)
  b2():
    jmp b3(Field 0)
  b3(v1: Field):
    return v1
}

DEF v0 := rc1 [u1 false]
DEF v2 := *v2 = None
DEF v3 := rc1 [u1 false]
DEF v5 := u1 false
DEF v1 := Field 0
--- Interpreter result after Removing Unreachable Functions (step 1):
Ok([Numeric(Field(0))])
---
```
As you can see `v0` in the 2nd pass is defined as `false`, because it still carries the effect of `v9 = array_set v7, index u32 0, value u1 0` from the 1st pass. Since this information is destroyed, I don't think we can "reset" it unfortunately, although adding special "snapshot" method could be more convenient than having to re-map from the ABI inputs.

aakoshh:
@jfecher I added helper methods to `Value` PTAL.

PR #8765, opened 2025-06-02
feat(ssa): Initial validation module

vezenovm:
# Description

## Problem\*

Working towards https://github.com/noir-lang/noir/issues/8705. It has multiple sub-issues so once those are resolved I think we could close the issue. 

## Summary\*

This PR simply makes an initial validation module that moves over some of our assertions on `Function` to a separate `Validator` struct. This felt like the simplest solution for our current validation plans. We can then expand this validator in the future as needed (e.g. extra state like the CFG and DominatorTree). 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/validation/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/validation/mod.rs
@@ -1,0 +1,272 @@
+use fxhash::FxHashSet as HashSet;
+
+use crate::ssa::ir::instruction::TerminatorInstruction;
+
+use super::ir::{
+    function::Function,
+    instruction::{BinaryOp, Instruction, InstructionId},
+    types::{NumericType, Type},
+    value::ValueId,
+};
+
+struct Validator<'f> {
+    function: &'f Function,
+    // State for truncate-after-signed-sub validation
+    // Stores: Option<(bit_size, result)>
+    signed_binary_op: Option<(u32, ValueId)>,
+}
+
+impl<'f> Validator<'f> {
+    fn new(function: &'f Function) -> Self {
+        Self { function, signed_binary_op: None }
+    }
+
+    fn start_function(&mut self) {
+        self.validate_single_return_block();
+        // Reset any state needed at start, if any
+        self.signed_binary_op = None;
+    }
+
+    fn visit_instruction(&mut self, instruction: InstructionId) {
+        self.validate_truncate_after_signed_sub(instruction);
+    }
+
+    fn finish_function(&mut self) {
+        if self.signed_binary_op.is_some() {
+            panic!("ICE: Truncate must follow the result of a checked signed add/sub");
+        }
+    }
+
+    /// Validates that any checked signed add/sub is followed by the expected truncate.
+    fn validate_truncate_after_signed_sub(&mut self, instruction: InstructionId) {
+        let dfg = &self.function.dfg;
+        match &dfg[instruction] {
+            Instruction::Binary(binary) => {
+                self.signed_binary_op = None;
+
+                match binary.operator {
+                    // Only validating checked addition/subtraction
+                    BinaryOp::Add { unchecked: false } | BinaryOp::Sub { unchecked: false } => {}
+                    // Otherwise, move onto the next instruction
+                    _ => return,
+                }
+
+                // Assumes rhs_type is the same as lhs_type
+                let lhs_type = dfg.type_of_value(binary.lhs);
+                if let Type::Numeric(NumericType::Signed { bit_size }) = lhs_type {
+                    let results = dfg.instruction_results(instruction);
+                    self.signed_binary_op = Some((bit_size, results[0]));
+                }
+            }
+            Instruction::Truncate { value, bit_size, max_bit_size } => {
+                let Some((signed_op_bit_size, signed_op_res)) = self.signed_binary_op.take() else {
+                    return;
+                };
+                assert_eq!(
+                    *bit_size, signed_op_bit_size,
+                    "ICE: Correct truncate must follow the result of a checked signed add/sub"
+                );
+                assert_eq!(
+                    *max_bit_size,
+                    *bit_size + 1,
+                    "ICE: Correct truncate must follow the result of a checked signed add/sub"
+                );
+                assert_eq!(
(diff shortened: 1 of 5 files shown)


compiler/noirc_evaluator/src/ssa/validation/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+
+impl<'f> Validator<'f> {
+    fn new(function: &'f Function) -> Self {
+        Self { function, signed_binary_op: None }
+    }
+
+    fn start_function(&mut self) {
+        self.validate_single_return_block();
+        // Reset any state needed at start, if any
+        self.signed_binary_op = None;
```
aakoshh: This looks a little strange. Is calling `start_function` multiple times an actual use case? I thought we call `new` and then `start_function`, so all state should be as it is by default. 

Maybe the iteration over the instructions could be moved into the `Validator`, to make it self contained. We clearly assume that we will visit all instructions and then call finish, not the other way around, and start in between them.

TomAFrench: Agree that this is a little odd. I'm fine with us doing multiple passes over each function in order to perform all of the various validations as this validation shouldn't be performance critical. Better to optimize for simplicity.

TomAFrench:
One thing to note here is that there's various validations which make sense to check on the initial SSA (even if we're starting halfway through the SSA optimization pipeline) and validations which we expect to hold after each and every SSA pass. 

I mention this as we can have cases where we assert that the initial SSA must follow rules like "all arithmetic must be checked" however SSA optimizations may convert checked arithmetic to unchecked later on so it would be good to have the validator flexible enough to handle this.

compiler/noirc_evaluator/src/ssa/validation/mod.rs
vezenovm: Yeah agreed this is weird. I originally tried to make the validators re-use the same pass over the function and made this "start", "visit", "finish" logic. However, with the combined validator this isn't necessary and is leftover logic that has leaked here. I'm going to get rid of this and move the visiting of instructions to the `Validator`.

vezenovm:
> Would it make sense to run this between SSA passes if some CLI flag is on? Maybe in integration tests?

Yeah a CLI command for this would be useful. Although this verifier is just on the initial SSA at the moment. I think we can extend the CLI in a follow-up. To start I think this on its own provides a nice benefit to fuzzing as it helps check that the SSA generated is well formed.

vezenovm:
> One thing to note here is that there's various validations which make sense to check on the initial SSA (even if we're starting halfway through the SSA optimization pipeline) and validations which we expect to hold after each and every SSA pass.
> 
> I mention this as we can have cases where we assert that the initial SSA must follow rules like "all arithmetic must be checked" however SSA optimizations may convert checked arithmetic to unchecked later on so it would be good to have the validator flexible enough to handle this.

Should we instead try to move away from validations that hold on the initial SSA but not on the SSA after certain passes? e.g. in your example I don't think we should have a requirement "all arithmetic must be checked" in the initial SSA.

vezenovm:
> One thing to note here is that there's various validations which make sense to check on the initial SSA (even if we're starting halfway through the SSA optimization pipeline) and validations which we expect to hold after each and every SSA pass

I will add configurations to the `Validator` to enable handling validation of different stages of the compilation pipeline. I will build off of this branch. For now, this PR is ready for review again.

vezenovm:
> One thing to note here is that there's various validations which make sense to check on the initial SSA (even if we're starting halfway through the SSA optimization pipeline) and validations which we expect to hold after each and every SSA pass.
> 
> I mention this as we can have cases where we assert that the initial SSA must follow rules like "all arithmetic must be checked" however SSA optimizations may convert checked arithmetic to unchecked later on so it would be good to have the validator flexible enough to handle this.

I was starting to add some configuration settings but at the moment I wasn't sure what invariants we apply to the initial SSA that we don't apply after other passes. There are per pass invariants (e.g. no more function types in parameters after defunctionalize) and I started on marking different stages of the compilation pipeline in the validator. However, I was thinking this may be good to keep localized in each pass. We can then have a hybrid approach where we apply per-pass localized checks, while this `Validator` can be for invariants that should hold about the SSA form no matter the pass. I feel this may be a better separation of concerns.

compiler/noirc_evaluator/src/ssa/validation/mod.rs
```diff
@@ -0,0 +1,472 @@
...
+    signed_binary_op: Option<(u32, ValueId)>,
+}
+
+impl<'f> Validator<'f> {
+    fn new(function: &'f Function) -> Self {
+        Self { function, signed_binary_op: None }
+    }
+
+    /// Validates that any checked signed add/sub is followed by the expected truncate.
+    fn validate_truncate_after_signed_sub(&mut self, instruction: InstructionId) {
```
asterite: nit: after_signed_add_or_sub?

compiler/noirc_evaluator/src/ssa/validation/mod.rs
```diff
@@ -0,0 +1,472 @@
...
+        let dfg = &self.function.dfg;
+        match &dfg[instruction] {
+            Instruction::Binary(binary) => {
+                self.signed_binary_op = None;
```
asterite: I was just thinking, what if there are two consecutive binary ops? I think the first one would be allowed to have no truncation on it.

asterite (approved):
Looks good! I left a comment for something that might need a fix.

compiler/noirc_evaluator/src/ssa/validation/mod.rs
vezenovm: You're right. I actually fixed that here https://github.com/noir-lang/noir/pull/8798. Would you be able to review that PR and then I can merge it into this branch as to avoid multiple PRs in the merge queue?

compiler/noirc_evaluator/src/ssa/validation/mod.rs
vezenovm: Renamed to `validate_signed_op_overflow_pattern` in https://github.com/noir-lang/noir/pull/8798

Issue #8764, opened 2025-06-02
Document `#[allow(dead_code)]`


Savio-Sou:
https://github.com/noir-lang/noir/pull/8066 introduced `#[allow(dead_code)]`, which should be documented in e.g. https://noir-lang.org/docs/noir/concepts/comptime#attributes

PR #8766, opened 2025-06-02
chore: document `allow(dead_code)` and reorganize attributes

asterite:
# Description

## Problem

Resolves https://github.com/noir-lang/noir/issues/8764

## Summary

Given that `allow(dead_code)` can be put on top of functions but also on top of data types, and an "Attributes" doc section existed but only for functions, I decided to create another section called "Attributes" to explain all of them. This section is still linked from the "Functions" section.

I only documented the attributes that were documented, plus a few more (`allow(unused_variables)` and `fuzz`). There are some other attributes, like `abi` and `export`, which are explained but are in other sections. Should I at least list them in the main attributes page and link to their respective sections?

Then, we could maybe mention that comptime attributes exist, and link to the comptime section from there too. Let me know!

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/docs/docs/noir/concepts/attributes.md
+++ b/docs/docs/noir/concepts/attributes.md
@@ -1,0 +1,120 @@
+---
+title: Attributes
+description:
+  Learn how to use attributes in Noir. Attributes are metadata that can be applied to data types,
+  functions and variables.
+keywords: [Noir, attributes]
+sidebar_position: 15
+---
+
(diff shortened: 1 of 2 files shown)


Savio-Sou:
Rather than linking out from the new Attribute page, moving all descriptions of attributes into the new page + linking to it from other pages should be more intuitive to read.

That said, we can tackle it with a future Issue. Have some docs is already exponentially better than having none 👍

PR #8778, opened 2025-06-03
fix: use value predicated by range checks

### LINKED ISSUE #8617: ACIR != Brillig: ACIR fails where Brillig returns a value
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8610

This is the minimised version of the program in https://github.com/noir-lang/noir/issues/8616

The following program fails with an overflow in ACIR but passes and returns a value in Brillig:
```rust
fn main(a: i32, c: bool) -> pub i32 {
    if c {
        a
    } else {
        ((-a) * a)
    }
}
```
Inputs:
```toml
a = "-0x000000000000000000000000000000000000000000000000000000002df86759"
c = true
```

### Expected Behavior

ACIR and Brillig should agree. In this case Brillig is right: if `c` is `true` then it should just return `a`. Changing `c` to `false` causes an overflow in Brillig as well. 

### Bug

```
❯ cargo run -q -p nargo_cli -- execute --force-brillig
[regression_fuzz_testing] Circuit output: Field(3523713191)

error: Failed constraint
  ┌─ src/main.nr:5:10
  │
5 │         ((-a) * a)
  │          --------
  │
  = Call stack:
    1. src/main.nr:5:10

```

### To Reproduce

See above.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+b2bfdb7bb0fc661bfb27da365f108cf131771295 (git version hash: ef1c8dc, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #8617 

## Summary\*
Instead of disabling a range check using a predicate, when a branch is not taken, we predicate the value itself. As a result the range is the same as before, but now we are ensured that this new predicated value respect the range-check.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
@@ -130,8 +130,8 @@ impl DefunctionalizationContext {
             let parameters = block.take_parameters();
             for parameter in &parameters {
                 let typ = &func.dfg.type_of_value(*parameter);
-                if is_function_type(typ) {
-                    func.dfg.set_type_of_value(*parameter, replacement_type(typ));
+                if let Some(rep) = replacement_type(typ) {
+                    func.dfg.set_type_of_value(*parameter, rep);
                 }
             }
 
@@ -160,8 +160,8 @@ impl DefunctionalizationContext {
                 #[allow(clippy::unnecessary_to_owned)] // clippy is wrong here
                 for result in func.dfg.instruction_results(instruction_id).to_vec() {
                     let typ = &func.dfg.type_of_value(result);
-                    if is_function_type(typ) {
-                        func.dfg.set_type_of_value(result, replacement_type(typ));
+                    if let Some(rep) = replacement_type(typ) {
+                        func.dfg.set_type_of_value(result, rep);
                     }
                 }
 
@@ -243,28 +243,8 @@ fn remove_first_class_functions_in_instruction(
             *arg = map_value(*arg);
         }
     } else if let Instruction::MakeArray { typ, .. } = instruction {
-        match typ {
-            Type::Array(element_types, len) => {
-                let new_element_types =
-                    element_types
-                        .iter()
-                        .map(|typ| {
-                            if matches!(typ, Type::Function) { Type::field() } else { typ.clone() }
-                        })
-                        .collect::<Vec<_>>();
-                *typ = Type::Array(Arc::new(new_element_types), *len);
-            }
-            Type::Slice(element_types) => {
-                let new_element_types =
-                    element_types
-                        .iter()
-                        .map(|typ| {
-                            if matches!(typ, Type::Function) { Type::field() } else { typ.clone() }
-                        })
-                        .collect::<Vec<_>>();
-                *typ = Type::Slice(Arc::new(new_element_types));
-            }
-            _ => {}
+        if let Some(rep) = replacement_type(typ) {
+            *typ = rep;
         }
         instruction.map_values_mut(map_value);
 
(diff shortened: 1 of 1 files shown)


TomAFrench:
I'm concerned about how this interacts with other references to this value which aren't under the same predicate. Seems like we can inadvertently modify those as well?

guipublic:
> I'm concerned about how this interacts with other references to this value which aren't under the same predicate. Seems like we can inadvertently modify those as well?

I tried this and indeed the predicated value is also used in the other branch for instance.

guipublic:
I have added a unit test and also made sure the predicated values are not used outside of their branch.

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -781,8 +811,12 @@ impl<'f> Context<'f> {
                     // Condition needs to be cast to argument type in order to multiply them together.
                     let casted_condition =
                         self.cast_condition_to_value_type(condition, value, call_stack);
-                    let value = self.mul_by_condition(value, casted_condition, call_stack);
-                    Instruction::RangeCheck { value, max_bit_size, assert_message }
+                    let predicate_value =
+                        self.mul_by_condition(value, casted_condition, call_stack);
+                    // Issue #8617: update the value to be the predicated value.
+                    // This ensure the value has the correct bit size in all cases.
```
vezenovm:
```suggestion
                    // This ensures that the value has the correct bit size in all cases.
```
nit

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -652,6 +660,28 @@ impl<'f> Context<'f> {
...
+    /// to the 'predicated_values' map if not already stored.
+    fn predicate_value(&mut self, value: ValueId, predicated_value: ValueId) {
+        let conditional_context = self.condition_stack.last_mut().unwrap();
+
+        if let std::collections::hash_map::Entry::Vacant(e) =
+            conditional_context.predicated_values.entry(value)
+        {
+            let old_mapping = self.inserter.resolve(value);
+            e.insert(old_mapping);
+        }
```
vezenovm:
```suggestion
        conditional_context
            .predicated_values
            .entry(value)
            .or_insert_with(|| self.inserter.resolve(value));
```
We can simplify this here

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -652,6 +660,28 @@ impl<'f> Context<'f> {
...
+        self.inserter.map_value(value, predicated_value);
+    }
+
+    /// Restore the previous mapping of predicated values.
+    fn reset_predicated_values(&mut self, conditional_context: &mut ConditionalContext) {
+        for (value, old_mapping) in conditional_context.predicated_values.drain() {
+            self.inserter.map_value(value, old_mapping);
+        }
+        conditional_context.predicated_values.clear();
```
vezenovm:
```suggestion
```
Doesn't drain remove elements as we iterate so this `clear()` is redundant?

guipublic: Yes it should, but in that case clearing it does not hurt. However I removed it since it can be confusing. I also checked that drain removed the elements.

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs

vezenovm (approved):

Issue #8780, opened 2025-06-03
Impl not found when adding variables to a dependency


asterite:
### Aim

Given this program:

```noir
trait Serialize {
    let Size: u32;

    fn serialize(self) -> [Field; Self::Size];
}

impl<A, B> Serialize for (A, B)
where
    A: Serialize,
    B: Serialize,
{
    let Size: u32 = <A as Serialize>::Size + <B as Serialize>::Size;

    fn serialize(self: Self) -> [Field; Self::Size] {
        let _ = self.0.serialize();
        let _ = self.1.serialize();
        [0; Self::Size]
    }
}

impl<T, let N: u32> Serialize for [T; N]
where
    T: Serialize,
{
    let Size: u32 = <T as Serialize>::Size * N;

    fn serialize(self: Self) -> [Field; Self::Size] {
        [0; Self::Size]
    }
}

impl Serialize for Field {
    let Size: u32 = 1;

    fn serialize(self) -> [Field; Self::Size] {
        [self]
    }
}

fn main() {
    let x = (((1, [2, 3, 4]), [5, 6, 7, 8]), 9);
    let _ = x.serialize();
}
```

It compiles fine.

If we add a dependency to it and put this in the dependency `src/lib.nr` file:

```noir
pub fn foo() {
    let _ = 1;
    let _ = 1;
    let _ = 1;
}
```

then the program stops compiling:

```
error: No matching impl found for `(Field, [Field; 3]): Serialize<Size = (8 - _)>`
   ┌─ src/main.nr:15:17
   │
15 │         let _ = self.0.serialize();
   │                 ---------------- No impl for `(Field, [Field; 3]): Serialize<Size = (8 - _)>`
   │
   = Required by `((Field, [Field; 3]), [Field; 4]): Serialize<Size = 8>`
```

For a repro that's easier to debug, see https://github.com/noir-lang/noir/pull/8740#issuecomment-2936295359

### Expected Behavior

The program should always compile.

### Bug

The program stops compiling when seemingly unrelated code exists.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

asterite:
Some debug notes:
- If we let `Type::canonicalize()` return `self.follow_bindings`, the code always compiles
[...]

PR #8781, opened 2025-06-03
chore: use BTreeMap for TypeBindings

asterite:
# Description

## Problem

Related to https://github.com/noir-lang/noir/issues/8780

## Summary

This is just a PR to prove that if we change the implementation of `TypeBindings` to be sorted by `TypeVariableId`, many programs stops to compile. This means that the order of the type bindings apparently matters.

I'd suggest, if this doesn't degrade performance, we could keep this going forward to get predictable results... that is, once we find why the order here matters.

## Additional Context

Before this PR, `TypeBindings` are an alias of `FxHashMap`. It seems this map always has a consistent order given the same keys. For example if we insert the keys 0, 1 and 2, the order ends up being 0, 1 and 2. If we insert the keys 2, 3 and 4 the order ends up being 4, 2 and 3. That would explain why adding seemingly unrelated code changes the behavior: there are new type variables, and the order type variables end up in the map end up being different.

Another note: this seems to only matter for the serialize programs, which rely on associated constants. So maybe the issue is around how associated constants are related to bindings, and not with the order of bindings.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir_def/types.rs
+++ b/compiler/noirc_frontend/src/hir_def/types.rs
@@ -1,6 +1,9 @@
-use std::{borrow::Cow, cell::RefCell, collections::BTreeSet, rc::Rc};
-
-use fxhash::FxHashMap as HashMap;
+use std::{
+    borrow::Cow,
+    cell::RefCell,
+    collections::{BTreeMap, BTreeSet},
+    rc::Rc,
+};
 
 #[cfg(test)]
 use proptest_derive::Arbitrary;
@@ -334,7 +337,7 @@ pub enum QuotedType {
 /// A list of (TypeVariableId, Kind)'s to bind to a type. Storing the
 /// TypeVariable in addition to the matching TypeVariableId allows
 /// the binding to later be undone if needed.
-pub type TypeBindings = HashMap<TypeVariableId, (TypeVariable, Kind, Type)>;
+pub type TypeBindings = BTreeMap<TypeVariableId, (TypeVariable, Kind, Type)>;
 
 /// Represents a struct or enum type in the type system. Each instance of this
 /// rust struct will be shared across all Type::DataType variants that represent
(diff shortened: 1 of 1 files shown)


jfecher:
> I'd suggest, if this doesn't degrade performance, we could keep this going forward to get predictable results... that is, once we find why the order here matters.

I'm a bit confused that this implies that our previous scheme was inconsistent but your Additional Details section states:

> Before this PR, TypeBindings are an alias of FxHashMap. It seems this map always has a consistent order given the same keys.

which sounds like it is consistent across runs?

I haven't noticed e.g. the serialization tests only failing sometimes either. I would think if we were inconsistent and we have all these tests failing when the order is changed, that we should see these tests randomly fail sometimes.

I'm also of the opinion that since we want to enforce that the order of type bindings shouldn't matter (since if it did it would just be wrong) that we should have a non-deterministic implementation so that non-deterministic failures could clue us into there being a bug where ordering is being relied upon.

I mentioned this in slack as well but the main case I can think of where ordering may matter is when we combine separate TypeBindings maps which were created in different "contexts" (ie. with different type bindings applied). If we had a function of type `fn(A, B) -> C` and one map with `A -> B -> i32` (a bound to b bound to i32), and another with `B -> u32`, we'd get different results since the bindings conflict, and may get different results depending on where `.follow_all_bindings()` is called as well.

asterite:
> which sounds like it is consistent across runs?

Right, what I mean is that if you run the compiler for the same code, it will always either compile or not compile. But if you change the code, for example adding unrelated code that introduces new type bindings, then the code might stop compiling because the order of the keys inside a given `TypeBindings` might have been altered (which is what's happening here).

> I haven't noticed e.g. the serialization tests only failing sometimes either

If you change the `TypeBindings` alias to be `HashMap`, not `FxHashMap`, then the serialize test will sometimes compile and sometimes not. I think that's because `HashMap` by default won't store the same keys in the same order across runs.

> I'm also of the opinion that since we want to enforce that the order of type bindings shouldn't matter (since if it did it would just be wrong) that we should have a non-deterministic implementation so that non-deterministic failures could clue us into there being a bug where ordering is being relied upon.

Agreed! 👍

> I mentioned this in slack as well but the main case I can think of where ordering may matter is when we combine separate TypeBindings maps which were created in different "contexts"

I checked if when calling `insert` on `TypeBindings` we already had a key for it and I couldn't find a case where this happens... but maybe I forgot to check all cases.

asterite:
Closing because I found a way to write a frontend test for this.

asterite:
(plus this isn't going to be the solution anyway)

PR #8782, opened 2025-06-03
chore(fuzz): Remove unreachable functions in the AST fuzzer

aakoshh:
# Description

## Problem\*

Resolves an inconvenience of having to wade through unreachable functions while minimising a Noir AST.

## Summary\*

Changes the AST fuzzer to remove functions which aren't reachable from `main` before doing the recursion rewrite. 

## Additional Context

These functions would have been removed by the SSA generation anyway, so the work of discovering which one is reachable would have had to be carried out by something eventually. This should make it easier to glance at a printed AST and guess how "big" it really is.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/ast_fuzzer/src/program/rewrite/unreachable.rs
+++ b/tooling/ast_fuzzer/src/program/rewrite/unreachable.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeSet, HashSet, VecDeque};
+use std::collections::{BTreeSet, VecDeque};
 
 use im::HashMap;
 use noirc_frontend::monomorphization::ast::{Definition, Expression, FuncId, Ident, Program};
@@ -15,7 +15,7 @@ pub(crate) fn remove_unreachable_functions(ctx: &mut Context) {
     }
     let reachable = find_reachable_functions(ctx);
 
-    // We have to re-assign function IDs, because in the `Program` t
+    // We have to re-assign function IDs, because in the `Program`
     // the ID of the function is expected to match its position.
     let remap = reachable
         .into_iter()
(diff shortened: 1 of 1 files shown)


tooling/ast_fuzzer/src/program/rewrite/unreachable.rs
```diff
@@ -0,0 +1,86 @@
...
+};
+
+/// Remove functions that are unreachable from main.
+pub(crate) fn remove_unreachable_functions(ctx: &mut Context) {
+    if ctx.functions.is_empty() {
+        return;
+    }
+    let reachable = find_reachable_functions(ctx);
+
+    // We have to re-assign function IDs, because in the `Program` t
```
vezenovm:
```suggestion
    // We have to re-assign function IDs, because in the `Program`
```

tooling/ast_fuzzer/src/program/rewrite/unreachable.rs
```diff
@@ -0,0 +1,86 @@
...
+    let mut visited = HashSet::new();
+
+    // Start from main.
+    queue.push_back(Program::main_id());
+
+    // Find all global functions referred to by their identifier.
+    while let Some(id) = queue.pop_front() {
+        if !visited.insert(id) {
+            continue;
+        }
```
vezenovm:
```suggestion
        if !reachable.insert(id) {
            continue;
        }
```
I think we can just merge these sets, why do we need a second `visited` set?

vezenovm (approved):
LGTM just some nits

tooling/ast_fuzzer/src/program/rewrite/unreachable.rs
aakoshh: You're right, I added `visited` on auto pilot 👍

Issue #8741, opened 2025-05-30
Compiler crash: Cannot return references from an if expression (re-assigning mut &mut)


aakoshh:
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8728
The crash occurs after merging https://github.com/noir-lang/noir/pull/8731

The following program crashes the compiler:

```rust
fn main(c: bool) -> pub bool {
    let mut e: &mut bool = (&mut false);
    if c { e = (&mut true); };
    *e
}
```

### Expected Behavior

Some kind of compilation error instead of a crash.

### Bug

```
❯ cargo run -q -p nargo_cli -- compile
The application panicked (crashed).
Message:  Cannot return references from an if expression
Location: compiler/noirc_evaluator/src/ssa/opt/flatten_cfg/value_merger.rs:68
```

So technically we are not returning a value from this `if`, but it would be equivalent to 
```
e = if c { &mut true } else { e };
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

Not using `mut &mut`, or not reassigning its value in an `if`.

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

jfecher:
Ah, this one's going to be more difficult to fix. I don't think we can just ban re-assigning a value to a mutable reference variable.

Subway2023:
Here are multiple [fuzz_targets](https://github.com/noir-lang/noir/tree/master/tooling/ast_fuzzer/fuzz/fuzz_targets), may I ask which one was used to discover this bug?

aakoshh:
@Subway2023 it came out in https://github.com/noir-lang/noir/pull/8728 , running the [smoke test](https://github.com/noir-lang/noir/blob/master/tooling/ast_fuzzer/tests/smoke.rs), then manually minimising to figure out which part of the printed code causes the issue.

Subway2023:
May I ask how this piece of code was generated? My understanding is that the smoke test is only used to test this code, but not to generate it.

aakoshh:
@Subway2023 of course, you can replicate it like this:

```shell
git fetch origin
# git checkout origin/af/af/7939-ast-fuzz-ref
git checkout 6bf936b4c3
NOIR_ARBTEST_SEED=0x9da2a3f300100000 cargo test -p noir_ast_fuzzer --test smoke
```

It prints this:
<details>
<summary>Test output</summary>

```rust
---- arb_program_can_be_executed stdout ----
[...]

Issue #8748, opened 2025-06-02
Compiler crash: Cannot return references from an if expression (array of &mut)


aakoshh:
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8728

This is similar to https://github.com/noir-lang/noir/issues/8741 but instead of reassigning a `mut &mut`, it's assigning to an `[&mut; _]`, either whole or in parts.

The following program crashes the compiler:

```rust
fn main(c: bool) {
    let mut a: [&mut bool; 2] = [&mut false, &mut true];

    if (c) {
        a = [a[1], a[0]];
        a[0] = &mut true;
    };

    a = [&mut true, a[1]];
    a[1] = &mut false;
}
```

### Expected Behavior

Not to crash. 

### Bug

```console
❯ cargo run -q -p nargo_cli -- compile --force
The application panicked (crashed).
Message:  Cannot return references from an if expression
Location: compiler/noirc_evaluator/src/ssa/opt/flatten_cfg/value_merger.rs:68
```

### To Reproduce

See above.

We can comment out either of the two lines in in the `if`, and either of the two after it, and the crash will still occur, until both lines are commented out either inside or outside. So for example this variant does not crash:
```rust
fn main(c: bool) {
    let mut a: [&mut bool; 2] = [&mut false, &mut true];

    if (c) {
        a = [a[1], a[0]];
        a[0] = &mut true;
    };

    // a = [&mut true, a[1]];
    // a[1] = &mut false;
}
```

### Workaround

None

### Project Impact

None

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+7e6d7cdb32c806aa0622c1c73e2255a79d93637d (git version hash: 485a0341340038eff5a4ba7af183a65137dc863f, is dirty: true)

### Would you like to submit a PR for this Issue?

None

PR #8783, opened 2025-06-03
fix: Thread errors through remove_if_else instead of panicing when the value merger finds reference values

jfecher:
# Description

## Problem\*

Working towards https://github.com/noir-lang/noir/issues/8741
Unfortunately this does not 100% solve the issue. We issue a RuntimeError now but panic when reporting the error since the call stack is empty. It turns out none of the call stacks for then_value, else_value, then_condition, or else_condition have any call stacks. This PR is already quite large though so I thought I'd keep this one focused on just error handling.

## Summary\*

~~These changes are all just threading RuntimeError through more places in SSA to account for the fact we can now error when inserting instructions.~~ The above is still true but we only thread through remove_if_else now and fail to optimize slice_push_back instead of propagating up further.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/errors.rs
+++ b/compiler/noirc_evaluator/src/errors.rs
@@ -66,9 +66,13 @@ pub enum RuntimeError {
         "Could not resolve some references to the array. All references must be resolved at compile time"
     )]
     UnknownReference { call_stack: CallStack },
-    #[error("Cannot return references from an if or match expression, or assignment within these expressions")]
+    #[error(
+        "Cannot return references from an if or match expression, or assignment within these expressions"
+    )]
     ReturnedReferenceFromDynamicIf { call_stack: CallStack },
-    #[error("Cannot return a function from an if or match expression, or assignment within these expressions")]
+    #[error(
+        "Cannot return a function from an if or match expression, or assignment within these expressions"
+    )]
     ReturnedFunctionFromDynamicIf { call_stack: CallStack },
 }
 
@@ -180,8 +184,8 @@ impl RuntimeError {
             | RuntimeError::UnconstrainedSliceReturnToConstrained { call_stack }
             | RuntimeError::UnconstrainedOracleReturnToConstrained { call_stack }
             | RuntimeError::UnknownReference { call_stack } => call_stack,
-            | RuntimeError::ReturnedReferenceFromDynamicIf { call_stack } => call_stack,
-            | RuntimeError::ReturnedFunctionFromDynamicIf { call_stack } => call_stack,
+            RuntimeError::ReturnedReferenceFromDynamicIf { call_stack } => call_stack,
+            RuntimeError::ReturnedFunctionFromDynamicIf { call_stack } => call_stack,
         }
     }
 }
--- a/compiler/noirc_evaluator/src/errors.rs
+++ b/compiler/noirc_evaluator/src/errors.rs
@@ -16,6 +16,8 @@ use thiserror::Error;
 use crate::ssa::ir::types::NumericType;
 use serde::{Deserialize, Serialize};
 
+pub type RtResult<T> = Result<T, RuntimeError>;
+
 #[derive(Debug, PartialEq, Eq, Clone, Error)]
 pub enum RuntimeError {
     #[error(transparent)]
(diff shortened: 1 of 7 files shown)


aakoshh:
Just wanted to mention https://github.com/noir-lang/noir/issues/8748 as well as being relevant to this. 

Can you reiterate the end goal? As I understand you didn't want to deny the ability to reassign mutable references in `if` statements, so the SSA has to handle it, right? What is supposed to happen with these errors that now returned by all of these passes? Are they just becoming a compilation error caught by the backend, rather than the frontend?

aakoshh:
Out of curiosity: what stops the frontend from rejecting reassigning mutable references only if they happen in the context of an if statement?

jfecher:
> What is supposed to happen with these errors that now returned by all of these passes? Are they just becoming a compilation error caught by the backend, rather than the frontend?

Yes, it will be reported where we report other SSA errors like when we fail to unroll a loop because the bounds aren't constant.

> Out of curiosity: what stops the frontend from rejecting reassigning mutable references only if they happen in the context of an if statement?

We'd have to do it for `match` too (this was something I missed in the original monomorphization error PR too) but I'm not sure. It just didn't sit right with me at the end of the day. Banning them from returning references directly is one thing but banning any mutable variable containing a reference from being reassigned to within an if/match is much closer to banning all mutable references within mutable variables in general. Only erroring in SSA is a _little_ more lenient in that we're still allowing them in ifs/matches which can be removed at codegen time but is still somewhat restrictive.

Maybe we should bite the bullet and ban mutable references in mutable variables in general? But I'm also not confident this is the core issue. It's more likely to do with nested mutable references in general. I wonder how much code would break if we did this.

I'll add an error during monomorphization today in another PR to ban nested mutable references and mutable refs in mutable vars and see how much code breaks in that PR.

jfecher:
This PR doesn't seem to slow down our SSA passes so let's merge it to remove the panic I think. In case my monomorphization PR doesn't catch all errors (e.g. nested references without a reference in the rhs: `**foo = 3;`) this will stop SSA from panicking in those cases and in cases where e.g. a fuzzer or ssa parser produces bad ssa directly.

asterite:
Like Akosh, I wonder if this could be somehow detected by the frontend. 

For example this code in Rust, which is the snippet that's in #8741 

```rust
fn main() {
    let c = true;
    let mut e: &mut bool =
        &mut false;
    if c {
        e = &mut true;
    };
    println!("{}", *e);
}
```

Gives this error:

```
error[E0716]: temporary value dropped while borrowed
 --> src/main.rs:6:18
  |
6 |         e = &mut true;
  |                  ^^^^- temporary value is freed at the end of this statement
  |                  |
  |                  creates a temporary value which is freed while still in use
7 |     };
8 |     println!("{}", *e);
  |                    -- borrow later used here
  |
  = note: consider using a `let` binding to create a longer lived value
```

Now I know that implementing this might mean we'd need something like the borrow checker, so I understand why we don't do that right now.

I was wondering... it seems the error only happens in `merge_values`. That's used in two places:
- the `remove_if_else` optimization
- in `simplify_slice_push_back`

I wonder if we could then:
- have `merge_values` return a `Result`, like it is now
- change SSA passes to return a `Result`, like now
- in `simplify_slice_push_back`, if `merge_values` doesn't return `Ok`, we don't simplify the instruction. Then I guess it will eventually error somewhere else (not sure!)

At least that way `insert_instruction` and friends don't need to change to return `Result`.

jfecher:
> Like Akosh, I wonder if this could be somehow detected by the frontend.

@asterite there is a related PR here: https://github.com/noir-lang/noir/pull/8790 for erroring in the frontend during monomorphization. Rust's lifetime error is unrelated to the issue here. Rust doesn't have a flattening pass like we do and if it needed to merge pointer values for some reason in its IR it could potentially treat them as integers and merge the integers. Our backend doesn't support actual pointers as values though so we have to remove them before then.

vezenovm (commented):
The changes to thread errors look fine as is. However, I am wondering whether we could do this as part of a pre-flattening validation pass as to avoid thread an `RtResult<ValueId>` from `merge_values`. We would then only have to change to `flatten_cfg(mut self) -> RtResult<Ssa>`.

jfecher:
@vezenovm would that require adding unwraps? My worry for adding unwraps to prevent Results when adding instructions is that if results are ever added in more places after it'd be more difficult to track if we're still meeting the conditions to not implicitly panic

vezenovm:
> @vezenovm would that require adding unwraps? My worry for adding unwraps to prevent Results when adding instructions is that if results are ever added in more places after it'd be more difficult to track if we're still meeting the conditions to not implicitly panic

I don't think it would. I need to correct my comment earlier. A _post_ flattening (or pre remove_if_else) validation pass I think would be better. As @asterite pointed `merge_values` is only ever called in `remove_if_else` and during the slice push back simplification. After flattening we could run a deeper validation of IfElse that we are not attempting to merge a reference. Only the new post flattening validation pass and `flatten_cfg` would have to return a Result then.

jfecher:
Updated using the "don't optimize slice_push_back if we get an error" approach. It's much smaller now. Only the remove_if_else exposes an RtError now as expected (I kept the new type alias).

PR #8784, opened 2025-06-04
fix: Merge `replacement_type` and `is_function_type` in `defunctionalization`

aakoshh:
# Description

## Problem\*

Resolves the following audit issues:
* [Array and slice types with deeply nested function types are not fully updated](https://audithub.veridise.com/app/organizations/85/projects/277/project-viewer?bpt=1&version=433&issue=21)
* [Missing/incorrect cases when replacing function types](https://audithub.veridise.com/app/organizations/85/projects/277/project-viewer?bpt=1&version=433&issue=18)

## Summary\*

* Removes `is_function_type` and changes `replacement_type` to return an `Option`, so there is no need for guarding the latter with the former. 
* Makes `replacement_type` recursive. 
* Uses `replacement_type` to handle types in arrays and slices.
* Extends `replacement_type` to consider arrays/slices that contain types with functions, recursively.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
@@ -632,7 +632,28 @@ fn replacement_type(typ: &Type) -> Option<Type> {
         Type::Reference(typ) => {
             replacement_type(typ.as_ref()).map(|typ| Type::Reference(Arc::new(typ)))
         }
-        _ => None,
+        Type::Numeric(_) => None,
+        Type::Array(items, size) => {
+            replacement_types(items.as_ref()).map(|types| Type::Array(Arc::new(types), *size))
+        }
+        Type::Slice(items) => {
+            replacement_types(items.as_ref()).map(|types| Type::Slice(Arc::new(types)))
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
```diff
@@ -245,23 +245,17 @@ fn remove_first_class_functions_in_instruction(
...
         match typ {
             Type::Array(element_types, len) => {
-                let new_element_types =
-                    element_types
-                        .iter()
-                        .map(|typ| {
-                            if matches!(typ, Type::Function) { Type::field() } else { typ.clone() }
-                        })
-                        .collect::<Vec<_>>();
+                let new_element_types = element_types
```
vezenovm: Isn't this also done by the `replacement_types` helper? Could we use that here?

aakoshh: I wouldn't use the `replacement_types` helper purely to save an extra allocation of vector while it decides whether it needs to replace anything, but you are right, we could reuse more of those functions: I think we can use `replacement_type` itself to get rid of the entire `match`, and then we only overwrite the type if we have to. Give me a sec...

PR #8785, opened 2025-06-04
chore: add a failing test for #8780

asterite:
# Description

## Problem

Related to #8780

## Summary

Just adds a failing test, because I can't find what changes to cause a failure.

## Documentation

Check one:

# PR Checklist


TomAFrench:
@asterite can you change this test so that it asserts on the incorrect behaviour? The test should link to the related issue #8780 and make it clear that we expect that we want this code to succeed in future (at which point the test should be updated).

This avoids us having a bunch of open PRs with failing tests and will flag up if we inadvertently fix the issue.

TomAFrench:
^ This is assuming that you're not expecting to immediately address this

asterite:
Ah, I'm working on this. Maybe I shouldn't have open a PR for this yet 😅 

My intention it to fix it now, I'll change the PR title.

asterite:
In the end I turned it into a failing test and we can merge it like that. I can't find the cause for the bug.

compiler/noirc_frontend/src/tests/traits.rs
```diff
@@ -1628,3 +1628,47 @@ fn accesses_associated_type_inside_trait_using_self() {
     "#;
     assert_no_errors!(src);
 }
+
+#[named]
+#[test]
+fn serialize_test_with_a_previous_unrelated_definition() {
+    let src = r#"
+    // If you remove this Trait definition, the code compiles fine. That's the bug.
+    pub trait Trait {}
```
asterite, line 1637: Also if you add one more trait, say `pub trait Trait1 {}`, then the code compiles fine too.

PR #8789, opened 2025-06-04
chore(fuzz): Tool to minimize Noir programs with `cvise`

### LINKED ISSUE #8535: Make util script for using `cvise` to minimize Noir errors
### Problem

Minimizing input programs can require a lot of context and be tedious, but [C-Vise](https://github.com/marxin/cvise?tab=readme-ov-file) is a parallel implementation of C-reduce, that supports ~arbitrary languages. 

I used this to minimize a case where executing ACIR/Brillig gives different results:

`bash` script I used:

```bash
#!/bin/bash
# No -o pipefail; we don't want nargo failures to cause the script to fail
set -eu

~/.nargo/bin/nargo new test_project && \
  (yes | cp main.nr test_project/src) && \
  (cd test_project && echo 'a = "0x000000000000000000000000000000000000000000000000131640459367fd34"' > Prover.toml) && \
  (cd test_project && \
    ~/.nargo/bin/nargo execute --force --silence-warnings ; \
    ~/.nargo/bin/nargo execute --force --silence-warnings --force-brillig \
  ) | pcre2grep --multiline "$(cat ~/.txt)"
```

and then just

```bash
cvise --not-c ./input_has_differing_output.sh main.nr
```

is able to reduce a ~150-line program (from AST-based fuzzing) down to <10 lines in ~60 seconds.

### Happy Case

Make a generic script and/or `README.md` so that it's easier to use `cvise` with `nargo` for compile/execute or ACIR vs. Brillig errors.

This requires:

```bash
Please ensure that [the test script succeeds] not only in the directory where
you are invoking C-Vise, but also in an arbitrary temporary
directory containing only the files that are being reduced. In other
words, running these commands:

  DIR=`mktemp -d`
  cp /home/michael/coding/noir/test_case_with_differing_output/main.nr $DIR
  cd $DIR
  /home/michael/coding/noir/test_case_with_differing_output/input_has_differing_output.sh
  echo $?

should result in '0' being echoed to the terminal.
Please ensure that the test script takes no arguments; it should be hard-coded to refer
to the same file that is passed as an argument to C-Vise.

See 'cvise.py --help' for more information.
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #8535

## Summary\*

Add scripts to utilise `cvise` to minimise the Noir programs produced by the AST fuzzer. 

See the [README](https://github.com/noir-lang/noir/blob/af/8535-cvise/tooling/ast_fuzzer/minimizer/README.md) for how to use and an example. 

## Additional Context

`cvise` has binaries for Linux, but not for Mac. For other platforms, using Docker is an [option](https://github.com/marxin/cvise/blob/master/INSTALL.md#using-docker-or-podman). 

Since it works by calling a script to compile the input program, it needs access to `nargo`, so I made a `Dockefile` that installs `nargo` and `cvise` as well. 

The idea is that we will mount a Noir project that replicates some issue, and then tell it to minimise it. We have to rebuild the image if we need to test a new version of `nargo` itself, but at we don't have to worry about mounting `nargo` from the developer machine, or cross building.

## Documentation\*

Check one:

# PR Checklist\*


tooling/ast_fuzzer/minimizer/scripts/minimize.sh
```diff
@@ -0,0 +1,46 @@
...
+        usage "$PROVER_PATH is not a file"
+    else
+        # `compile` doesn't need a Prover.toml file, but to keep `docker run`
+        # simple we can create an empty one.
+        touch $PROVER_PATH
+    fi
+fi
+
+# Make a copy because the minimizer will modify the file in-place.
+cp $MAIN_PATH $MAIN_PATH.bkp
```
asterite, line 43: What if we copy `main.nr` to `main.minimized.nr` and have the tool work with the minimized one? Then `main.nr` is unchanged, the and minimized one is changed, and has a name that reflects what it is.

aakoshh: I found it was convenient to let it overwrite `main.nr` because it's immediately obvious what is going on: you can see the code changing shape without having to open another file. But, I made no effort to prevent a 2nd execution of the script overwrite the `.bkp` file, and from that perspective I like your suggestion better. 

I thought I could add a `minimise.sh accept` command to at least help with overwriting `main.nr` if `main.minimized.nr` looks good. 

OTOH I can imagine my flow would be to immediately overwrite `main.nr` with the contents of `main.minimized.nr` so that I can try it in `nargo` on the host, and minimise it further, and in that context again modifying `main.nr` makes sense, at which point I would lose my original. Well, it would still be available in some console output. 

Perhaps we could create `.bkp` only if it doesn't exist, otherwise return an error, and then have `accept` remove it or `reject` restore the original?

asterite: Ah, actually, I think the `.bkp` approach makes sense after your explanation. After all you don't really need that huge main file, if you end up with a reduction then it's good, and the backup is there in case you need it 👍

tooling/ast_fuzzer/minimizer/scripts/docker-entry.sh
```diff
@@ -0,0 +1,21 @@
...
+# Create a check script which creates a new project with the contents we want to test,
+# run a `nargo` command, and checks the presence of the error message in the output.
+# If it's not present, it means the latest reduction step was not interesting.
+cat > check.sh <<EOF
+nargo new test_project
+cp main.nr test_project/src
+cp /noir/Prover.toml test_project
+cd test_project
+nargo $CMD 2>&1 | grep "$MSG"
+EOF
```
aakoshh, line 19: I went back and forth whether this should be a script checked into source control, taking env vars, without having the message verbatim in it. But then I thought `cvise` docs say we should provide a script that takes no parameters, and this is more in line with that.

PR #8790, opened 2025-06-04
fix!: Error when re-assigning a mutable reference

jfecher:
# Description

## Problem\*

Follow-up for https://github.com/noir-lang/noir/pull/8783#issuecomment-2940005113
Seeing if anything external breaks from this error.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
@aakoshh this doesn't appear to break anything so I think lets merge this for now. It's a very unfortunate restriction I think but we can try lifting or relaxing it in the future.

PR #8793, opened 2025-06-04
chore: Improve access to data required to extract Noir

iamrecursion:
# Description

## Problem\*

Extracting Lampe requires access to some data that is currently private within the various crates making up the compiler and toolchain. This PR makes these things public.

## Summary\*

The [Lampe](https://github.com/reilabs/lampe) project aims to formally verify Noir code through a process of extraction and tactic-based proving. In order to correctly perform its extraction, it needs access to some internal data within the compiler. This PR makes this data accessible, either by adding read-only accessors or by making fields and functions public where this seems more appropriate.

## Additional Context

N/A

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/node_interner.rs
```diff
@@ -280,6 +280,18 @@ pub struct NodeInterner {
...
+    pub fn dependency_graph(&self) -> &DiGraph<DependencyId, ()> {
+        &self.dependency_graph
+    }
+
+    /// Gets the trait implementations from the node interner.
+    pub fn trait_implementations(&self) -> &HashMap<TraitImplId, Shared<TraitImpl>> {
+        &self.trait_implementations
+    }
+}
+
```
jfecher: Do you mind moving this further down inside the existing `impl NodeInterner`?

compiler/noirc_frontend/src/node_interner.rs
```diff
@@ -441,6 +459,12 @@ impl GlobalId {
     }
 }
 
+impl Into<usize> for GlobalId {
+    fn into(self) -> usize {
+        self.0
+    }
+}
```
jfecher: Can you change the `Into` impls into `as_usize()` functions or similar?

Also curious what all these into usize impls are for

compiler/noirc_frontend/src/node_interner.rs
iamrecursion: Happily! Done.

compiler/noirc_frontend/src/node_interner.rs
iamrecursion: Great catch. These are artefacts of older methods of doing things that I clearly missed removing when I was checking through our changes. The only one remaining (which is now an `as_index` is used for some iteration-based searching for modules.

iamrecursion:
I've addressed your comments in the latest commit!

jfecher (approved):

jfecher:
@iamrecursion sorry for the delay! We were having a CI issue the other day where an unrelated test was blocking some PRs which happened to affect this PR and the clippy fix PR. It should be resolved now though - can you merge master again to pick up these fixes?

iamrecursion:
Sorry for the wait here! I was away for a few days; merged the new changes now and will hopefully see this merged soon!

PR #8796, opened 2025-06-04
fix: unify infix expressions by isolating unbound type variables

asterite:
# Description

## Problem

Resolves https://github.com/noir-lang/noir/issues/8780

## Summary

The explanation of why this works is a bit tricky, but here it is:
- At one point in the compiler (`sort_commutative`) types are sorted
- The type `Type::TypeVariable` eventually holds a type variable ID, and that's how these types are sorted
- Depending on the other of type variables in a `TypeBindings`, these might affect the result of `sort_commutative`
- In the case where the compiler failed to compile, we ended up with an equation like `A + B = 3 - C`, and that doesn't unify

Huge thanks to @jfecher for helping me debug this. While debugging, we noticed that even though `A + B = 3 - C` didn't unify, it **could**. The "solution" is to say that `A = 3 - C - B` or that `B = 3 - C - A`, and that's what this PR does. Now that these unify, it doesn't matter anymore whether we end up with one type or another using `sort_commutative`, because in both cases we'll be able to unify them.

@jfecher Feel free to improve the above explanation 😅 

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
```diff
@@ -374,6 +374,99 @@ impl Type {
...
+    /// - `A + rhs = other` -> `A = other - rhs`
+    /// - `A - rhs = other` -> `A = other + rhs`
+    /// - `lhs + B = other` -> `B = other - lhs`
+    /// - `lhs - B = other` -> `B = lhs - other`
+    fn try_unify_by_isolating_an_unbound_type_variable_in_self(
+        &self,
+        other: &Type,
+        bindings: &mut TypeBindings,
+    ) -> Result<(), UnificationError> {
+        if let Type::InfixExpr(lhs_a, op_a, rhs_a, _) = self {
```
jfecher: The name `rhs_a` is a bit confusing when the comment above the function uses `B` for this value and `A` for `lhs_a`

jfecher (approved):

compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
asterite: Good point. Done!

PR #8798, opened 2025-06-04
fix(ssa): Validate signed mul overflow

vezenovm:
# Description

## Problem\*

Resolves #8088 

Builds upon https://github.com/noir-lang/noir/pull/8765.

## Summary\*

This adds validation similar to https://github.com/noir-lang/noir/pull/8706, except for signed multiplication. The semantics for which are seen here. We could be more strict with these checks by checking this code gen as well:
```rust
self.check_signed_overflow(result, lhs, rhs, operator, bit_size, location);
self.insert_safe_cast(result, result_type, location)
```
However, for now I have chosen to simply check up to the truncate point to simplify our "state machine". We also mainly want to guarantee that we have truncated the result of an operation appropriately so that someone cannot code gen an overflow.

## Documentation\*

Check one:
- [ ] No documentation needed.
- [X] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [X] I have tested the changes locally.
- [X] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #8799, opened 2025-06-04
fix(ssa): Validate field to integer cast

### LINKED ISSUE #8087: Casting without truncation causing differences between ACIR and Brillig
### Aim

+

### Expected Behavior

Both failed or both succeeded

### Bug

The following SSA fails in ACIR and succeeds in Brillig
```
  b0(v0: Field, v1: u64):
    v2 = cast v0 as u64
    v3 = mul v2, v1
    return v3
}
```

```
  b0(v0: Field, v1: u64):
    v2 = cast v0 as u64
    v3 = mul v2, v1
    return v3
}
```

With the following witness:
```
v0 = 295707996719639606511915798873678070407142200220841934848
v1 = 9242244106192460504
```

Brillig returns 0, 
Acir: Execution failed: Failed assertion

### To Reproduce

in tooling/ssa_fuzzer/lib.rs
```rs
mod tests {
    .........
    use noirc_evaluator::ssa::ir::types::NumericType;
    use noirc_evaluator::ssa::ir::instruction::BinaryOp;
    use crate::runner::execute_single;
    fn regression_untruncated_cast() {
        let mut acir_builder = FuzzerBuilder::new_acir();
        let mut brillig_builder = FuzzerBuilder::new_brillig();

        let field_var_acir_id_1 =
            acir_builder.insert_variable(ValueType::Field.to_ssa_type()).value_id;
        let u64_var_acir_id_2 = acir_builder.insert_variable(ValueType::U64.to_ssa_type()).value_id;
        let field_var_brillig_id_1 =
            brillig_builder.insert_variable(ValueType::Field.to_ssa_type()).value_id;
        let u64_var_brillig_id_2 =
            brillig_builder.insert_variable(ValueType::U64.to_ssa_type()).value_id;

        let casted_acir = acir_builder
            .builder
            .insert_cast(field_var_acir_id_1, NumericType::Unsigned { bit_size: 64 });
        let casted_brillig = brillig_builder
            .builder
            .insert_cast(field_var_brillig_id_1, NumericType::Unsigned { bit_size: 64 });

        let mul_acir = acir_builder.builder.insert_binary(
            casted_acir,
            BinaryOp::Mul { unchecked: false },
            u64_var_acir_id_2,
        );
        let mul_brillig = brillig_builder.builder.insert_binary(
            casted_brillig,
            BinaryOp::Mul { unchecked: false },
            u64_var_brillig_id_2,
        );

        acir_builder.builder.terminate_with_return(vec![mul_acir]);
        brillig_builder.builder.terminate_with_return(vec![mul_brillig]);

        let acir_program = acir_builder.compile().unwrap();
        let brillig_program = brillig_builder.compile().unwrap();

        let mut initial_witness = WitnessMap::new();
        initial_witness.insert(
            Witness(0),
            FieldElement::try_from_str("295707996719639606511915798873678070407142200220841934848")
                .unwrap(),
        );
        initial_witness.insert(Witness(1), FieldElement::from(9242244106192460504_u64));

        let result_witness = Witness(2);
        let compare_results = run_and_compare(
            &acir_program.program,
            &brillig_program.program,
            initial_witness,
            result_witness,
            result_witness,
        );
        println!("{:?}", compare_results);
        match compare_results {
            CompareResults::Agree(_result) => {
                return;
            }
            CompareResults::BothFailed(_acir_error, _brillig_error) => {
                return;
            }
            _ => panic!("Should fail or agree in both"),
        }
    }
}
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #8087 

Builds upon #8798 as I was originally worried about merge conflicts for building off of #8765.

## Summary\*

This simply adds validation for Field -> integer casting. Casting involving signed integer requires a much more involved validation check. 

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -130,6 +133,7 @@ impl Context<'_, '_, '_> {
             let overflow = self.insert_binary(rhs, BinaryOp::Lt, bit_size_var);
             let predicate = self.insert_cast(overflow, typ);
             let pow = self.pow(base, rhs);
+            let pow = self.insert_truncate(pow, typ.bit_size(), FieldElement::max_num_bits());
```
guipublic: This cast seems to be wrong indeed, but I think we can avoid the truncate, because 'pow' is not used when the cast is not safe.

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -179,6 +183,7 @@ impl Context<'_, '_, '_> {
             rhs,
         );
         let pow = self.pow(base, rhs);
+        let pow = self.insert_truncate(pow, lhs_typ.bit_size(), FieldElement::max_num_bits());
```
guipublic: This cast is safe, so we do not need the truncate.
Indeed rhs is ensured to be less than the bit size of lhs (see the 'Nullify rhs in case of overflow' comment).

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -205,6 +210,7 @@ impl Context<'_, '_, '_> {
         // Performs the division on the 1-complement (or the operand if positive)
         let shifted_complement = self.insert_binary(one_complement, BinaryOp::Div, pow);
         // Convert back to 2-complement representation if operand is negative
+        let lhs_sign = self.insert_truncate(lhs_sign, lhs_typ.bit_size(), lhs_typ.bit_size() + 1);
```
guipublic: This cast is also safe, so we do not need the truncate either. `lhs_sign` is a boolean.

vezenovm:
> I think the added truncates in remove_bit_shifts can be avoided.

I ultimately removed all of them @guipublic.

I have moved to only running validation after initial SSA gen for the time being. It used to live in `FunctionBuilder::finish` which was also used by the inlining passes. Making validation more generally aware of value ranges is overly complex for this PR and should be done as part of an entirely separate value range analysis work. The validator can then reference the map constructed by a separate value range analysis to determine whether a Cast is safe. As the validator is currently written it can only be run on the initial SSA (which in itself required adding the truncate simplification invariants).

guipublic:
Could you also fix the wrong 'cast' in remove_bit_shits.rs?
Line 133, remove the casting:  `let pow = self.insert_cast(pow, typ);`, and return a pow of type 'Field' instead, and implement the casting after:
`if max_bit <= bit_size`  then do the casting here, we know it is safe
'else': the result is already 'cast' after the truncate, so there is nothing to do. (just remove the casting of 'pow' to Field, as it is already a Field now).

vezenovm:
> (just remove the casting of 'pow' to Field, as it is already a Field now).

I also ended up casting the overflow predicate to a Field so that the `predicate * pow` multiplication had matching types.

guipublic (approved):
LGTM

PR #8801, opened 2025-06-05
fix(ssa): Keep last stores used in array returned from a function

### LINKED ISSUE #8786: Execution crash: `condition value is not a boolean: MismatchedBitSize`
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8728 with https://github.com/noir-lang/noir/pull/8743 merged into it. 
Tested just on top of @vezenovm 's fix in https://github.com/noir-lang/noir/pull/8743

The following program compiles, but crashes during execution:

```rust
unconstrained fn main() -> pub [i8; 2] {
    if (*func_5()[1].0) {
        [0, 1]
    } else {
        [2, 3]
    }
}
unconstrained fn func_5() -> [(&mut bool, str<2>, str<4>, str<4>); 2] {
    [((&mut true), "AB", "WTGS", "AELL"), ((&mut false), "CD", "PKHQ", "LDWE")]
}
```

### Expected Behavior

It should return `[2,3]`. 

### Bug

```console
Message:  condition value is not a boolean: MismatchedBitSize { value_bit_size: 254, expected_bit_size: 1 }
Location: /Users/aakoshh/Work/aztec/noir/acvm-repo/brillig_vm/src/lib.rs:455
```

The behaviour changes if we edit the code: 

If we return `i8` instead of `[i8; 2]` then we get an execution error instead of crash (it still compiles):

```console
error: Assertion failed: 'Bit size for value 254 does not match the expected bit size 1'
  ┌─ src/main.nr:2:10
  │
2 │     if (*func_5()[1].0) {
  │          -----------
  │
  = Call stack:
    1. src/main.nr:2:10
```

If we change `func_5` to not return one of the `str<4>`s the execution succeeds:

```rust
unconstrained fn func_5() -> [(&mut bool, str<2>, str<4>); 2] {
    [((&mut true), "AB", "WTGS"), ((&mut false), "CD", "PKHQ")]
}
```

```console
[regression_fuzz_testing] Circuit output: Vec([Field(2), Field(3)])
```

### To Reproduce

1. Check out the linked PR branch.
2. Execute the program.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+0ce143dc8a9d541f0f12c21c1d571e4b202cb549 (git version hash: d3c008f1702f5df4b7c2987acd49f6f046f38c49, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #8786

Builds upon #8743 

## Summary\*

In #8743, we switched to mark all aliases of a Call's arguments unknown. Similarly, in this PR we mark all aliases a return termiantor's arguments unknown. This then keeps the store that would otherwise be removed.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -652,10 +652,22 @@ impl<'f> PerFunctionContext<'f> {
...
+                let mut all_aliases = Vec::new();
+                for return_value in return_values {
+                    if let Some(expression) = references.expressions.get(return_value) {
+                        if let Some(aliases) = references.aliases.get(expression) {
+                            aliases.for_each(|alias| {
+                                all_aliases.push(alias);
+                            });
+                        }
+                    }
+                }
```
aakoshh: Maybe this, and the one [here](https://github.com/noir-lang/noir/blob/8580c15e123c222f49dc55a674f08f026242a525/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs#L527-L536) and [here](https://github.com/noir-lang/noir/blob/8580c15e123c222f49dc55a674f08f026242a525/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs#L462-L464) could be merged into a helper function that takes a list value IDs to iterate and a lambda that tells what to do with the alias. Perhaps it could be arranged into further helpers to make it more intuitive to the casual reader to connect "anything which acts on a reference in these values" to gathering aliases.

aakoshh:
I think the fuzzer still finds failures after this; I'll try to isolate them in new tickets.

EDIT: It was https://github.com/noir-lang/noir/issues/8803 which seems to be a different issue with mem2reg.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
jfecher: It doesn't look like this handles the cases where expressions or aliases may return `None` - e.g. expressions with unknown aliases.

Can you add a test for:
```
brillig(inline) fn foo f1 {
  b0(v0: &mut u1):
    v0 = call f2()
    store u1 1 at v0
    v4 = make_array [v0] : [&mut u1; 1]
    return v4
}

brillig(inline) fn get_ref f2 {
  b0():
    v0 = allocate -> &mut u1
    store u1 1 at v0
    return v0
}
```

jfecher (changes requested):

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
vezenovm: > It doesn't look like this handles the cases where expressions or aliases may return `None` - e.g. expressions with unknown aliases.

@jfecher I added that test. `v0` will still be included as an alias of v4, so we will not remove `store u1 1 at v0`.

vezenovm: > Maybe this, and the one [here](https://github.com/noir-lang/noir/blob/8580c15e123c222f49dc55a674f08f026242a525/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs#L527-L536) and [here](https://github.com/noir-lang/noir/blob/8580c15e123c222f49dc55a674f08f026242a525/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs#L462-L464) could be merged into a helper function that takes a list value IDs to iterate and a lambda that tells what to do with the alias

I chose to just repeat the code as I thought we only did this for Call and Return. I forgot we do it inside Store as well. I'll make a helper before merging.

vezenovm: @aakoshh There are already some helpers on`Block` for fetching all aliases and a for_each helper on `AliasSet`. I chose to switch to those.

vezenovm: Hmm that seems to have broken things. I'll have to make a new helper.

vezenovm:
In the parent PR https://github.com/noir-lang/noir/pull/8743 we were not appropriately labeling array aliases. To keep arrays containing references we updated the input to `mark_all_unknown` by collecting all aliases for every argument to a call or return. In fact, `mark_all_unknown` was just too restrictive. I have updated `mark_all_unknown` to access every alias from an array constant.

vezenovm:
This is ready for review again. cc @jfecher

PR #8805, opened 2025-06-05
fix: signed right shift overflows to 0 or -1

### LINKED ISSUE #8791: Right shift for signed integers
### Problem

When the right shift 'overflow' the bit size, i.e when we shift a n-bits signed integer of m bits to the right, with m>=n, it returns 0.

### Happy Case

We should return -1 if the shifted value is negative, and 0 if it is positive.
This will match the behaviour of C++

### Workaround

None

### Workaround Description

An alternative is to use Rust behaviour which 'overflows' in that case.

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #8791 

## Summary\*
When right shift overflows the bit size, we return 0 is the shifted value is positive, and -1 it is strictly negative.

## Documentation\*

Check one:

# PR Checklist\*


jfecher (approved):
The perf regressions are a bit unfortunate

PR #8804, opened 2025-06-05
fix!: indexing arrays with non-u32 is now an error

asterite:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/7795

## Documentation

Check one:
- [x] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #8807, opened 2025-06-05
fix: type unification tests, and try moving constants to the other side

### LINKED ISSUE #8780: Impl not found when adding variables to a dependency

asterite:
# Description

## Problem

Resolves #8780

## Summary

Note that the first commit just moves things around.

The main change here, besides the unit tests, is that we'll now try to unify `(..a..) + constant = (..b..)` by moving the constant to the other side, `(..a..) = (..b..) - constant`, then trying to unify those. The tricky part is not trying to move the constant back to where it was, and for that I introduced a `UnificationFlags` type.

This solves the changed serialize test, which is taking `serialize_1` and adding a `pub trait Trait {}` to it, but I wasn't sure if we wanted such a test. I did confirm that it still compiles if we do that change, though.

## Additional Context

Maybe the HashMap used in `TypeBindings` shouldn't be an `FxHashMap`? That way I think on each compilation type variables in a hashmap will be arranged in a different order, which will cause these bugs to surface more easily... though it would also means the chance of flakes is greater (if we still have bugs).

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/hir_def/types/unification.rs
```diff
@@ -0,0 +1,893 @@
...
+}
+
+/// When unifying types we sometimes need to adjust the algorithm a bit.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+enum UnificationFlags {
+    /// Nothing special to do.
+    None,
+    /// If the right-hand side is `expr op constant`, don't try to move the constant to the left-hand side.
+    DoNotMoveConstantsOnTheRight,
+}
```
jfecher, line 30: Could maybe use this in the future to not call `Type::canonicalize` on every recursive invocation of try_unify as well

compiler/noirc_frontend/src/hir_def/types/unification.rs
```diff
@@ -0,0 +1,893 @@
...
+        }
+
+        fn type_variable_with_kind(&mut self, kind: Kind) -> (Type, TypeVariableId) {
+            let id = TypeVariableId(self.next_type_variable_id);
+            self.next_type_variable_id += 1;
+            (Type::TypeVariable(TypeVariable::unbound(id, kind)), id)
+        }
+    }
+
+    fn num(value: u128) -> Type {
```
jfecher:
```suggestion
    fn constant(value: u128) -> Type {
```
`num` makes me think `i32`, `u8`, `Field`, etc

compiler/noirc_frontend/src/hir_def/types/unification.rs
```diff
@@ -0,0 +1,893 @@
...
+
+        Err(UnificationError)
+    }
+
+    /// Try to unify the following equations:
+    /// - `(..a..) + 1 = (..b..)` -> `(..a..) = (..b..) - 1`
+    /// - `(..a..) - 1 = (..b..)` -> `(..a..) = (..b..) + 1`
+    /// - `(..a..) = (..b..) + 1` -> `(..b..) = (..a..) - 1`
+    /// - `(..a..) = (..b..) - 1` -> `(..b..) = (..a..) + 1`
+    fn try_unify_by_moving_single_constant_term(
```
jfecher, line 442: This seems very similar to `try_unify_by_moving_constant_terms` - from the comments above both they seem identical - what is the difference? Do we need both?

asterite: I noticed that too, but `try_unify_by_moving_constant_terms` seems to be done only when both sides are infixes with the same op (but maybe that wasn't intentional?). Then I thought it wasn't exactly the same because it uses `inverted_infix_expr` but looking again they do seem to be the same. I'll refactor this.

asterite: Ah, also the other one uses `approx_inverse`, I wasn't sure if that was good to use here... but I'll try.

asterite: Done! I removed `try_unify_by_moving_constant_terms` because it was only doing that on one side and the `try_unify_by_moving_single_constant_term` does it on both sides while preventing recursion.

compiler/noirc_frontend/src/hir_def/types/unification.rs
asterite: Done!

compiler/noirc_frontend/src/hir_def/types/unification.rs
asterite: Oh, you mean, when we call `try_unify` from, say `try_unify_by_moving_constant_terms`, there would be no need to call `canonicalize` on `lhs` and `rhs`, right?

jfecher: Even when we just call try_unify itself recursively on e.g. `a + (b + c) = d + (e + f)` we'll recur to `b + c = e + f` but we already canonicalized and don't need to canonicalize again every time - hence why canonicalization can be exponential with the size of the type.

jfecher (approved):
Oh lord the .snap changes :sweat_smile:

jfecher:
Actually - I'm unsure why we're getting so many snap changes here - were they committed accidentally (seeing all the failing tests)? I wouldn't expect a type system change to result in so many of these.

asterite:
Ah, you are right. I don't understand what's going on. When I ran the tests locally it produced those changes, but on CI they don't. I'll try to figure out what's wrong...

asterite:
I have no idea what's going on. It seems that it's picking up the source code from poseidon v0.1.1 in some cases, but the toml says it's v0.1.0. Some were updated in https://github.com/noir-lang/noir/pull/8804 but that isn't merged yet 🤔 

In any case poseidon will need to be updated to v0.1.1 so I'm going to try to update it anywhere it's referenced. Maybe there's some caching issue...

asterite:
Note that the regressions are because of updating poseidon.

PR #8814, opened 2025-06-06
fix(comptime): Do not overflow on signed checked ops

vezenovm:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/pull/8806#discussion_r2129493175

## Summary\*

Update the comptime interpreter to match the side effect (overflow) semantics of binary operations in the SSA. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/tests.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/tests.rs
@@ -314,3 +314,133 @@ fn generic_functions() {
     let result = interpret(program);
     assert_eq!(result, Value::U8(2));
 }
+
+#[test]
+fn add_unsigned() {
+    let src = "
+    comptime fn main() -> pub u32 {
+        2 + 100
+    }
+    ";
+    let result = interpret(src);
+    assert_eq!(result, Value::U32(102));
+}
+
+#[test]
+fn add_signed() {
+    let src = "
+    comptime fn main() -> pub i32 {
+        2 + 100
+    }
+    ";
+    let result = interpret(src);
+    assert_eq!(result, Value::I32(102));
+}
+
+#[test]
+fn add_overflow_unsigned() {
+    let src = "
+    comptime fn main() -> pub u8 {
+        200 + 100
+    }
+    ";
+    let result = interpret_expect_error(src);
+    let InterpreterError::MathError { operator, .. } = result else {
+        panic!("Expected MathError");
+    };
+    assert_eq!(operator, "+");
+}
+
+#[test]
+fn sub_unsigned() {
+    let src = "
+    comptime fn main() -> pub u32 {
+        10101 - 101
+    }
+    ";
+    let result = interpret(src);
+    assert_eq!(result, Value::U32(10000));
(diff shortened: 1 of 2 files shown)


compiler/noirc_frontend/src/hir/comptime/tests.rs
```diff
@@ -314,3 +314,133 @@ fn generic_functions() {
...
+
+#[test]
+fn sub_underflow_signed() {
+    let src = "
+    comptime fn main() -> pub i8 {
+        -120 - 10
+    }
+    ";
+    let result = interpret(src);
+    assert_eq!(result, Value::I8(126));
```
aakoshh, line 398: My understanding of https://github.com/noir-lang/noir/pull/8806 was that the SSA interpreter should not error out on overflows because there will be an SSA operation following up that will check whether the overflow happened, and handle it by returning its own error. So an overflow _will_ cause failure, it's just deferred. 

However, in the comptime interpreter there are no follow up instructions. By allowing overflows for signed integers, aren't we deviating from what ACIR and Brillig do?

For example:
```rust 
fn main() -> pub i8 {
    -120 - 10
}
```

```console
  ┌─ src/main.nr:2:5
  │
2 │     -120 - 10
  │     ---------
  │
  = Call stack:
    1. src/main.nr:2:5

Failed assertion
```

I'm surprised the AST fuzzer doesn't fail. 

What am I missing?

aakoshh: Another demonstration:

```rust
unconstrained fn main() {
    let a: i8 = comptime { foo() };
    let b = bar();
    assert_eq(a, b)
}

comptime fn foo() -> i8 {
    -120 - 10
}

fn bar() -> i8 {
    -120 - 10
}
```

Fails with the same assertion.

aakoshh: Ah, the reason it `comptime_vs_brillig` did not fail is because we disabled the generation of overflowing operations, because they generated too many known bugs. 

If I change `avoid_overflow` and `avoid_negative_int_literals` then it does fail, but often just because an overflow fails to compile the entire program. I do get some instances where only Brillig fails, though. @rkarabut maybe we can change `comptime_vs_brillig` to sometimes allow these things, but in the comparison do not fail if the program failed to compile due to an overflow, only if the comptime worked, but Brillig did not.

vezenovm: > My understanding of #8806 was that the SSA interpreter should not error out on overflows because there will be an SSA operation following up that will check whether the overflow happened, and handle it by returning its own error. So an overflow _will_ cause failure, it's just deferred

Ah yes, good point. 

> However, in the comptime interpreter there are no follow up instructions. By allowing overflows for signed integers, aren't we deviating from what ACIR and Brillig do?

Good catch yes we would be deviating from ACIR and Brillig. cc @jfecher I think I'm going to close this PR then and we just need to update the interpreter as in #8806

PR #8818, opened 2025-06-06
chore!: disallow casting signed integers to field

guipublic:
# Description

## Problem\*

As discussed, we disallow casting signed integers to Field. This is because the current implementation is not coherent:
-1_i8 as Field resulted in 255 instead of -1.
However changing this behaviour has too many impact before the 1.0 release.

## Summary\*
Casting a signed integer to a Field is forbidden

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/types.rs
+++ b/compiler/noirc_frontend/src/elaborator/types.rs
@@ -1193,7 +1193,18 @@ impl Elaborator<'_> {
 
         match to {
             Type::Integer(sign, bits) => Type::Integer(sign, bits),
-            Type::FieldElement => Type::FieldElement,
+            Type::FieldElement => {
+                let is_signed = match from_follow_bindings {
+                    Type::Integer(Signedness::Signed, _) => true,
+                    Type::TypeVariable(ref var) => var.is_signed(),
+                    _ => false,
+                };
+                if is_signed {
+                    self.push_err(TypeCheckError::UnsupportedFieldCast { location });
+                }
+
+                Type::FieldElement
+            }
             Type::Bool => {
                 let from_is_numeric = match from_follow_bindings {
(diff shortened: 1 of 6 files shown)


compiler/noirc_frontend/src/elaborator/types.rs
```diff
@@ -1193,7 +1193,18 @@ impl Elaborator<'_> {
 
         match to {
             Type::Integer(sign, bits) => Type::Integer(sign, bits),
-            Type::FieldElement => Type::FieldElement,
+            Type::FieldElement => {
+                let is_signed = match from_follow_bindings {
+                    Type::Integer(Signedness::Signed, _) => true,
+                    Type::TypeVariable(ref var) => var.is_signed(),
+                    _ => false,
+                };
```
jfecher: Can you edit `Type::is_signed` to add the TypeVariable case and use that method here instead?

We should also edit `Type::is_signed` to use `follow_bindings_shallow` instead

jfecher (commented):
Looks good but quite a few external repos are failing. May need to update them first or override CI.

compiler/noirc_frontend/src/elaborator/types.rs
guipublic: Done

guipublic:
> Looks good but quite a few external repos are failing. May need to update them first or override CI.

I updated the external repos in order to not cast a signed integer into Field (#14855).
We should check again this PR when we upgrade to a new version of Aztec Packages.

TomAFrench (commented):
It doesn't look like we have any tests which enforce that this error is emitted.

guipublic:
> It doesn't look like we have any tests which enforce that this error is emitted.

yes, I just added one.

TomAFrench (approved):

PR #8825, opened 2025-06-06
fix: bind self generic type in trait calls via a concrete type

### LINKED ISSUE #8794: Unexpected `Object type is unknown in field access` error
Taking the reproduction from https://github.com/noir-lang/noir/issues/4943 and slightly updating it, we get:

```noir
use dep::std::convert::From;

struct MyU128 {
    lo: Field,
    hi: Field,
}

impl MyU128 {
    pub fn from_u64s_le(lo: u64, hi: u64) -> MyU128 {
        MyU128 { lo: lo as Field, hi: hi as Field }
    }
}

impl From<u64> for MyU128 {
    fn from(value: u64) -> MyU128 {
       MyU128::from_u64s_le(value, 0)
    }
}

impl From<u32> for MyU128 {
    fn from(value: u32) -> MyU128 {
       MyU128::from(value as u64)
    }
}

fn main(mut x: u64) {
    let mut small_int = MyU128::from(x);
    assert(small_int.lo == x as Field);
    let u32_3: u32 = 3; 
    assert(MyU128::from(u32_3).lo == 3);
}
```

This program fails to compile with

```
error: Object type is unknown in field access
   ┌─ src/main.nr:31:22
   │
31 │     assert(small_int.lo == x as Field);
   │                      -- Type must be known by this point
   │
   = Try adding a type annotation for the object type before this expression

error: Object type is unknown in field access
   ┌─ src/main.nr:33:32
   │
33 │     assert(MyU128::from(u32_3).lo == 3);
   │                                -- Type must be known by this point
   │
   = Try adding a type annotation for the object type before this expression
```

which is odd as all the types are known upfront as far as I can tell.

asterite:
# Description

## Problem

Resolves #8794

## Summary

The main issue here was that when you had a call like this:

```noir
SomeConcreteType::trait_method(...)
```

that would resolve to a trait method, but we'd throw away the information that this was called on `SomeConcreteType`.

## Additional Context

So in the example of #8794:
- when we elaborate `MyU128::from` we end up with a type `fn('4) -> '5`
- then when we elaborate the call, `MyU128::from(x)`, given `x : u64` we end up with `fn(u64) -> '5`

The problem is that the return type is still unknown at this point.

Checking with Rust, it could figure out the return type even if we didn't call the function, for example like this:

```rust
// Rust doesn't know the argument types, but it knows it returns MyU128
let f = MyU128::from;
```

We already have specific code to solve `SomeConcreteType::trait_method` as a reference to a trait method, but the information about the specific type, `SomeConcreteType`, was thrown away. Now it's captured, then when elaborating it we'd bind the "Self" generic type of a function (if there's any) to that type.

Or at least this is what I felt was missing. Let me know if this should be done some other way.

Now it behaves like Rust, so we can actually do this:

```noir
// Noir now knows f returns MyU128
let f = MyU128::from;
let _ = f(x);
```

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/enums.rs
+++ b/compiler/noirc_frontend/src/elaborator/enums.rs
@@ -713,8 +713,9 @@ impl Elaborator<'_> {
             | PathResolutionItem::PrimitiveType(_)
             | PathResolutionItem::Trait(_)
             | PathResolutionItem::ModuleFunction(_)
-            | PathResolutionItem::TypeAliasFunction(_, _, _)
-            | PathResolutionItem::TraitFunction(_, _, _)
+            | PathResolutionItem::TypeAliasFunction(..)
+            | PathResolutionItem::TraitFunction(..)
+            | PathResolutionItem::TypeTraitFunction(..)
             | PathResolutionItem::PrimitiveFunction(..) => {
                 // This variable refers to an existing item
                 if let Some(name) = name {
--- a/compiler/noirc_frontend/src/elaborator/path_resolution.rs
+++ b/compiler/noirc_frontend/src/elaborator/path_resolution.rs
@@ -35,12 +35,22 @@ pub(crate) enum PathResolutionItem {
     Trait(TraitId),
 
     // These are values
+    /// A reference to a global value.
     Global(GlobalId),
+    /// A function call on a module, for example `some::module::function()`.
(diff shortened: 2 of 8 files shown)


TomAFrench (approved):
LGTM

PR #8827, opened 2025-06-06
fix: bind self generic type in trait calls via a concrete type in more cases

asterite:
# Description

## Problem

Follow-up to https://github.com/noir-lang/noir/pull/8825

## Summary

Follow-up to the recent PR for handling some remaining cases, mainly when `self` is a type alias or a primitive type.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/tests/traits.rs
+++ b/compiler/noirc_frontend/src/tests/traits.rs
@@ -1239,32 +1239,6 @@ fn calls_trait_method_using_struct_name_when_multiple_impls_exist() {
 
 #[named]
 #[test]
-fn calls_trait_method_using_struct_name_when_multiple_impls_exist_and_errors_turbofish() {
-    let src = r#"
-    trait From2<T> {
-        fn from2(input: T) -> Self;
-    }
-    struct U60Repr {}
-    impl From2<[Field; 3]> for U60Repr {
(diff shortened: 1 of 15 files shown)


compiler/noirc_frontend/src/tests/traits.rs
```diff
@@ -1237,32 +1237,6 @@ fn calls_trait_method_using_struct_name_when_multiple_impls_exist() {
...
-        }
-    }
-    impl From2<Field> for U60Repr {
-        fn from2(_: Field) -> Self {
-            U60Repr {}
-        }
-    }
-    fn main() {
-        let _ = U60Repr::<Field>::from2([1, 2, 3]);
-                                        ^^^^^^^^^ Expected type Field, found type [Field; 3]
```
asterite, line 1260: This test now started failing because `U60Repr` doesn't have any generics. The way it (incorrectly) worked is that the turbofish was applied to the generics of the trait. That's why the error says it's expecting `Field` (taken from the turbofish). So this PR also solves this bug. The error now is that `U60Repr` doesn't take generics... hmm... I think I'll add a test for that scenario instead.

asterite: Done!

jfecher (approved):

PR #8828, opened 2025-06-06
fix(interpreter): Return -1 for shr signed overflow

vezenovm:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/pull/8806#issuecomment-2950199379 

Follow-up to https://github.com/noir-lang/noir/pull/8805. We updated the comptime interpreter 

## Summary\*

Fallback to -1 when we have an overflowing shift right in the SSA interpreter to match functionality in comptime and in ACIR/Brillig.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
@@ -1091,7 +1091,17 @@ impl Interpreter<'_> {
                 }
             }
             BinaryOp::Shr => {
-                let zero = || NumericValue::zero(lhs.get_type());
+                // let zero = || NumericValue::zero(lhs.get_type());
+                // let zero = || NumericValue::zero(lhs.get_type());
+
+                let fallback = || {
+                    if lhs.get_type().is_unsigned() {
+                        NumericValue::zero(lhs.get_type())
+                    } else {
+                        NumericValue::neg_one(lhs.get_type())
+                    }
+                };
+
                 let Some(rhs) = rhs.as_u8() else {
                     let rhs = rhs.to_string();
                     return Err(internal(InternalError::RhsOfBitShiftShouldBeU8 {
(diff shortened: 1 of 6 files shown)


compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
```diff
@@ -1091,7 +1091,17 @@ impl Interpreter<'_> {
...
             BinaryOp::Shr => {
-                let zero = || NumericValue::zero(lhs.get_type());
+                // let zero = || NumericValue::zero(lhs.get_type());
+                // let zero = || NumericValue::zero(lhs.get_type());
+
+                let fallback = || {
+                    if lhs.get_type().is_unsigned() {
+                        NumericValue::zero(lhs.get_type())
+                    } else {
+                        NumericValue::neg_one(lhs.get_type())
```
jfecher: It's not that we return `-1` if the type is signed, it's that we return `-1` if the type is signed _and_ `lhs < 0`, right?

Let's add a test for this case, signed, positive integer

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
```diff
@@ -1091,7 +1091,17 @@ impl Interpreter<'_> {
                 }
             }
             BinaryOp::Shr => {
-                let zero = || NumericValue::zero(lhs.get_type());
+                // let zero = || NumericValue::zero(lhs.get_type());
+                // let zero = || NumericValue::zero(lhs.get_type());
+
```
vezenovm:
```suggestion
```

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
vezenovm: > It's not that we return `-1` if the type is signed, it's that we return `-1` if the type is signed _and_ `lhs < 0`, right?

Ah right. Will update and add a test.

vezenovm: Done

jfecher (approved):

vezenovm:
I have some changes that fix the comptime vs brillig shr flake https://github.com/noir-lang/noir/actions/runs/15499196774/job/43643781238

tooling/nargo_cli/tests/snapshots/compile_success_no_bug/comptime_right_shift_no_overflow/execute__tests__expanded.snap
```diff
@@ -3,6 +3,6 @@ source: tooling/nargo_cli/tests/execute.rs
 expression: expanded_code
 ---
 fn main(x: Field) -> pub Field {
-    let y: Field = 1;
+    let y: Field = 0;
```
vezenovm, line 6: This test was made with the expectatio nthat an overflowing shr will be zero https://github.com/noir-lang/noir/pull/8380. So this brings us back to the correct functionality.

Issue #8305, opened 2025-05-01
Unrolling: `Could not determine loop bound at compile-time` with negative bound minus zero


aakoshh:
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8299

The following program does not compile:

```rust
fn main(a: i32) -> pub i32 {
    let mut r: i32 = 0;
    let s: i32 = -2;
    let e: i32 = -1;
    for _ in (s - 0)..(e + 0) {
        r = r + a;
    }
    r
}
```

### Expected Behavior

Adding or subtracting 0 from a number should not make a difference. 

### Bug

```console
$ RUST_BACKTRACE=1 cargo run -q -p nargo_cli -- compile --skip-underconstrained-check --skip-brillig-constraints-check --silence-warnings 
error: Could not determine loop bound at compile-time
  ┌─ src/main.nr:5:15
  │
5 │     for _ in (s - 0)..(e + 0) {
  │               ----- If attempting to fetch the length of a slice, try converting to an array. Slices only use dynamic lengths.
  │
  = Call stack:
    1. src/main.nr:5:15

Aborting due to 1 previous error
```

It works with `--force-brillig`, because it doesn't try to unroll this loop. 

Both `s-0` and `e+0` are enough to cause issues on their own.

The SSA looks like this when we only have `s` and `e` (which compiles):

```
After Loop Invariant Code Motion (1):
acir(inline) predicate_pure fn main f0 {
  b0(v0: i32):
    v2 = allocate -> &mut i32
    store i32 0 at v2
    v4 = cast v0 as u32
    v6 = lt v4, u32 2147483648
    jmp b1(i32 4294967239)
  b1(v1: i32):
    v9 = lt v1, i32 4294967241
    jmpif v9 then: b2, else: b3
  b2():
...
```

and as follows when it's just `s-0`:
```
After Loop Invariant Code Motion (1):
acir(inline) predicate_pure fn main f0 {
  b0(v0: i32):
    v2 = allocate -> &mut i32
    store i32 0 at v2
    v5 = cast u32 4294967239 as i32
    v6 = cast v0 as u32
    v8 = lt v6, u32 2147483648
    jmp b1(v5)
  b1(v1: i32):
    v10 = lt v1, i32 4294967241
    jmpif v10 then: b2, else: b3
```

Note that `v5` is passed to `b1` instead of a literal.

### To Reproduce

See above.

### Workaround

None

### Project Impact

None

### Nargo Version

[...]

PR #8843, opened 2025-06-09
fix: don't check for overflow when rhs is zero

### LINKED ISSUE #8833: bug: unnecessary subtraction not optimized
### Aim

The constant folding isn't working optimally for the subtraction of zero like `x - 0`.

### Expected Behavior

The subtraction of zero should be optimized out, so `x - 0` should be simplified to just `x`.

### Bug

The following two programs should produce the same SSA form, but they don't.

```rust
global a: i8 = -121;
unconstrained fn main() -> pub i8 {
    let mut b: u32 = 25;
    (-(-a))
}
```

```rust
global a: i8 = -121;
unconstrained fn main() -> pub i8 {
    let mut b: u32 = (25 - 0);
    ((-(-a)) - 0)
}
```

### To Reproduce

1. Create a new Noir project.
2. Add the provided code snippets to the main file.
3. Compile with `nargo compile --show-ssa` to show the SSA form.
4. Observe that the SSA form for the second program includes an unnecessary subtraction of zero.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8833

## Summary

There's already a simplification that will turn `x - 0` into `x`. However, when creating the initial SSA we insert an overflow check after some binary operations, even if those binary operations end up being simplified. I thought about not inserting the overflow check when the instruction is simplified, but we don't know what it was simplified to or, well, we could check if it's a binary, etc., but it would be a bit bug-prone to do that.

## Additional Context

I'm not sure how to test this.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
+++ b/compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
@@ -366,6 +366,13 @@ impl<'a> FunctionContext<'a> {
         operator: BinaryOpKind,
         location: Location,
     ) -> ValueId {
+        // If the rhs is zero there's never an overflow chance
+        let rhs_value = self.builder.current_function.dfg.get_numeric_constant(rhs);
+        let rhs_is_zero = rhs_value.is_some_and(|rhs| rhs.is_zero());
+        if rhs_is_zero {
+            return result;
+        }
+
         let result_type = self.builder.current_function.dfg.type_of_value(result).unwrap_numeric();
         match result_type {
             NumericType::Signed { bit_size } => {


vezenovm (changes requested):
I'm not sure this is the best solution, as I would expect our simplifications to handle this case. If the operation can be simplified I would expect the overflow checks to be simplified as well. 
This is our final SSA:
```
After Dead Instruction Elimination (3) (step 43):
g0 = i8 135

  b0():
    v2 = cast u8 135 as i8
    v3 = truncate v2 to 8 bits, max_bit_size: 9
    v4 = cast v3 as u8
    v6 = lt v4, u8 128
    constrain v6 == u1 0, "attempt to subtract with overflow"
    return v3
}
```
We know that we should be able to simplify this SSA. I think our [signed cast simplification](https://github.com/noir-lang/noir/blob/7083516c37c6cfa00a612dd9a9685ac8696de3c1/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify/cast.rs#L60) is incorrect. We are only simplifying if the signed cast is below 2^(bit_size - 1), however, this does not encompass fully how we represent signed integers. I think we could convert our unsigned type using `IntegerConstant`, and then we would also remove another place where we are duplicating field -> integer conversions.

asterite:
For this program:

```noir
unconstrained fn main(x: i8) -> pub i8 {
    x - 0
}
```

we get this initial SSA:

```
brillig(inline) fn main f0 {
  b0(v0: i8):
    v1 = truncate v0 to 8 bits, max_bit_size: 9
    v2 = cast v1 as u8
    v3 = cast v0 as u8
    v5 = lt v3, u8 128
    v6 = not v5
    v7 = lt v2, u8 128
    v8 = eq v7, v5
    v9 = unchecked_mul v8, v6
    constrain v9 == v6, "attempt to subtract with overflow"
    v10 = cast v1 as i8
    return v10
}
```

There are no constants anymore so I don't think the cast can be simplified. It still does the overflow check for the binary operation, but there's no need to do it because it's subtracting zero from it. Maybe zero is a special case, not sure... What I mean is, I'm not sure this can be solved by further simplifying the "check_overflow" function.

asterite:
Also this likely doesn't matter much because this only happens in the initial SSA codegen, so you'd really have to have `- 0` in the source code, which should never be needed, so...

vezenovm:
> There are no constants anymore so I don't think the cast can be simplified. It still does the overflow check for the binary operation, but there's no need to do it because it's subtracting zero from it. Maybe zero is a special case, not sure... What I mean is, I'm not sure this can be solved by further simplifying the "check_overflow" function.

Yes, that one cannot be simplified as it is dynamic, and thus why the instructions generated by `check_overflow` are not simplified in this case. However, I think this is a separate issue, and maybe this is something for which we want to special case. 

The snippet in the issue https://github.com/noir-lang/noir/issues/8833 still can be simplified, and points me to that the cast simplification is incorrect. I would also expect these programs to produce the same SSA:
Unsimplified subtraction:
```noir
global a: i8 = -121;
unconstrained fn main() -> pub i8 {
    let mut b: u32 = (25 - 0);
    ((-(-a)) - 1)
}
```
No subtraction:
```noir
global a: i8 = -122;
unconstrained fn main() -> pub i8 {
    let mut b: u32 = 25;
    (-(-a))
}
```
With the current PR we would get the same SSA we get in the `- 0` case when it should be able to simplify to returning a single value.

asterite:
I now changed it to now to an overflow check if the binary operation was simplified to a constant. I personally think this makes sense, as if we were able to simplify the operation to a constant it means we were able to do it and there's no chance of an overflow, and `lhs` and `rhs` aren't needed for this operation anymore.

I understand that there might be some things to fix or improve in signed cast operations, but I think we'd still want to avoid doing overflow checks when they aren't needed, right?

asterite:
Apparently this also fixed #8305. I don't know if we'll merge, but I'll add a regression test for that.

vezenovm:
> I now changed it to now to an overflow check if the binary operation was simplified to a constant. I personally think this makes sense, as if we were able to simplify the operation to a constant it means we were able to do it and there's no chance of an overflow, and `lhs` and `rhs` aren't needed for this operation anymore.

Yes if we can simplify to a constant there is no chance of an overflow. But that is why I would expect the overflow instructions to also be simplified out upon insertion. If overflow checks are not needed they should be simplified out. 

> I understand that there might be some things to fix or improve in signed cast operations, but I think we'd still want to avoid doing overflow checks when they aren't needed, right?

The main benefit I see from this vs fixing signed simplifications is that we do not spend any time generating and simplifying instructions for `check_overflow` in SSA gen. It will avoid doing overflow checks, but only for the initial SSA gen and also leaks our simplifier into SSA gen (e.g. what if the simplifier did not make overflow checks).

This does not fix the issue if later SSA passes simplify a signed binary operation to a constant that does not overflow. Fixing simplification does handle that though. 

Now that we have moved away from checking for a constant zero and just a constant result I am ok with this change as it could provide a smaller IR earlier on (although I am not sure about this with fixed signed simplification), it is very little actual extra logic in code gen, and could provide some minor performance benefits. Although keeping this logic contained to the simplifier still provides a better separation of concerns in my opinion as the performance benefit from this change should still be minor and rare.

I have pushed the signed cast fix here https://github.com/noir-lang/noir/pull/8862 and I contend that this PR is an optimization rather than a fix. I'd like to get other team members thoughts on whether we want this optimization as I don't feel too strongly either way.

asterite:
Ah, I see now. I couldn't understand what the bug was, but now I understand that `constant_uint < integer_modulus` didn't work for negative values. I missed that, sorry!

I agree, the other PR is the better way to solve this. I'm curious why the numbers aren't exactly the same in the "Changes to ..." sections, but the differences are minimal.

vezenovm:
> I'm curious why the numbers aren't exactly the same in the "Changes to ..." sections, but the differences are minimal.

I have not investigated but I imagine it is due to further simplifications being made possible by our SSA passes, while in this PR it is only for the initial SSA.

PR #8855, opened 2025-06-09
fix: handle unconditional break during SSA codegen in all cases

asterite:
# Description

## Problem

Resolves https://github.com/noir-lang/noir/issues/8836

## Summary

Alternative to #8849 that implements this: https://github.com/noir-lang/noir/pull/8849#issuecomment-2956858514

## Additional Context

In the end this is much simpler, and it's just a handful of cases that we need to handle.

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/errors.rs
```diff
@@ -74,6 +74,10 @@ pub enum RuntimeError {
         "Cannot return a function from an if or match expression, or assignment within these expressions"
     )]
     ReturnedFunctionFromDynamicIf { call_stack: CallStack },
+    /// This case is not an error. It's used during codegen to prevent inserting instructions after
+    /// code for a break or continue is generated.
```
asterite:
```suggestion
    /// code when a break or continue is generated.
```

Issue #8618, opened 2025-05-21
SSA Interpreter: Internal error on dead instruction known to overflow


aakoshh:
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8610
Similar to https://github.com/noir-lang/noir/issues/8615 and https://github.com/noir-lang/noir/issues/8616

A program that the compiler knows will cause an overflow causes an internal `ConstantDoesNotFitInType` error in the SSA interpreter. After DIE the overflowing instruction itself is removed, and we're left with the always-fail constraint. 

Maybe the `ConstantDoesNotFitInType` should be promoted to a non-internal error.

### Expected Behavior

Don't expect `Internal` error from the SSA interpreter. 

### Bug

```
---
SSA Inputs:
0: rc1 [i16 32764]
---
Options:
CompareOptions { inliner_aggressiveness: 9223372036854775807 }
---
SSA 1 after step 29 (Check u128 mul overflow):
acir(inline) predicate_pure fn main f0 {
  b0(v0: [i16; 1]):
    v1 = allocate -> &mut u32
    v3 = array_get v0, index u32 0 -> i16
    v4 = cast v3 as u16
    v164 = lt v4, u16 32768
    v7 = not v164
    v8 = cast v7 as u64
    v165 = unchecked_mul u64 18446744073709486080, v8
    v11 = cast v3 as u64
    v166 = unchecked_add v165, v11
    v13 = cast v166 as i64
    v18 = cast v7 as u32
    v167 = unchecked_mul u32 4294901760, v18
    v21 = cast v3 as u32
    v168 = unchecked_add v167, v21
    v23 = cast v168 as i32
    v151 = make_array [u1 0, u1 1, u1 0, u1 1, u1 0, u1 0, u1 1, u1 0] : [u1; 8]
    v126 = cast Field 18889465931478580854784 as i64
    v128 = cast v166 as Field
    constrain u1 0 == u1 1, "attempt to bit-shift with overflow"
    v140 = call f1(i16 0, Field -288486952671523352321000885529921143040, u32 25) -> [(i16, Field, Field, [u8; 3]); 4]
    return v0
}
brillig(inline_always) predicate_pure fn func_3_proxy f1 {
  b0(v0: i16, v1: Field, v2: u32):
    v3 = allocate -> &mut i16
    v4 = allocate -> &mut Field
    v5 = allocate -> &mut u32
    v6 = allocate -> &mut i16
    v7 = allocate -> &mut Field
[...]

PR #8866, opened 2025-06-10
fix(interpreter): Promote ConstantDoesNotFitInType to non-internal error

vezenovm:
# Description

## Problem\*

Builds upon https://github.com/noir-lang/noir/pull/8862 to fix some of the fuzzing errors. 

To reproduce take this example:
```
NOIR_ARBTEST_SEED=0x1b4496bb00100000 cargo test -p noir_ast_fuzzer_fuzz pass_vs_prev 
```
This will produce an AST that is expected to have a signed integer overflow. 

## Summary\*

Leaving as a draft to see if it fixes the fuzzing failures. However, I'm not sure this is the best approach, as `ConstantDoesNotFitInType` does make sense as an internal error. We may want to fix how Cast's are interpreted instead.

We could also have both an internal `ConstantDoesNotFitInType` and an external `ConstantDoesNotFitInType`. For casting it would be an external error, but for things like converting an `ir::value::Value::NumericConstant` to an `interpreter::Value` we would trigger an internal error.

Another option could be to delay these internal errors in some way to see if we run into an expected external error. 

## Additional Context

A similar case has occurred here https://github.com/noir-lang/noir/issues/8618.

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
Going to close this for now. As I am not a fan of this solution on its own, but we will have to think of something to avoid `ConstantDoesNotFitInType` errors on programs we expect to overflow.

PR #8867, opened 2025-06-10
fix: Count array copies for slice functions

jfecher:
# Description

## Summary\*

`--count-array-copies` wasn't counting slice copies made from slice functions like `.push_back`, `.push_front`, or `.insert`.

## Additional Context

I've also added `--count-array-copies` to the force recompile flags since if there's a cached program otherwise nothing will be printed.

I'd also like to get `--count-array-copies` tested by CI automatically. We handle stdout output currently but don't pass this flag so I couldn't add a regression test for the slice copy code. This should be a separate PR though.

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
I haven't investigated deeply but I think the fuzz failure in https://github.com/noir-lang/noir/actions/runs/15566297609/job/43880648547?pr=8867 is most likely the same as https://github.com/noir-lang/noir/pull/8866. We expect an overflow and for the later constrain to be triggered, but the constant does not fit in our SSA interpreter type.

PR #8869, opened 2025-06-10
fix: (SSA interpreter) check requires_acir_gen_predicate for enable_side_effects

### LINKED ISSUE #8864: SSA Interpreter: Changes return value after Constant Folding
### Aim

Reproduction of https://github.com/noir-lang/noir/actions/runs/15561857419/job/43816860771?pr=8551

The following program changes the outcome during some of the SSA passes:

```rust
fn main(b: [u128; 1]) -> pub [u128; 1] {
    [
        if (
            if unsafe { func_1() } {
                0
            } else {
                (b[0] as Field)
            }
                != 1
        ) {
            b[0]
        } else {
            b[0]
        },
    ]
}
unconstrained fn func_1() -> bool {
    (!false)
}
```

Input:
```toml
b = ["0x0000000000000000000000000000000000000000000000000000000000000001"]
```

### Expected Behavior

SSA output should not change.

### Bug

```console
❯ cargo run -q -p nargo_cli -- interpret --ssa-pass "step 23" --ssa-pass "step 22" --show-ssa-pass "step 23" --show-ssa-pass "step 22"
warning: Unsafe block must have a safety comment above it
  ┌─ src/main.nr:4:16
  │
4 │             if unsafe { func_1() } {
  │                ------ The comment must start with the "Safety: " word
  │

After Purity Analysis (step 22):
  b0(v0: [u128; 1]):
    v2 = call f1() -> u1
    enable_side_effects v2
    v3 = not v2
    enable_side_effects v3
    v5 = array_get v0, index u32 0 -> u128
    v6 = cast v5 as Field
    enable_side_effects u1 1
    v8 = cast v2 as Field
    v9 = cast v3 as Field
    v10 = mul v9, v6
    v12 = eq v10, Field 1
    v13 = not v12
    enable_side_effects v13
    v14 = array_get v0, index u32 0 -> u128
    enable_side_effects v12
    v15 = array_get v0, index u32 0 -> u128
    enable_side_effects u1 1
    v16 = cast v13 as u128
    v17 = cast v12 as u128
    v18 = unchecked_mul v16, v14
    v19 = unchecked_mul v17, v15
    v20 = unchecked_add v18, v19
    v21 = make_array [v20] : [u128; 1]
    return v21
}
brillig(inline) predicate_pure fn func_1 f1 {
  b0():
    return u1 1
}

--- Interpreter result after Purity Analysis (step 22):
Ok([ArrayOrSlice(ArrayValue { elements: Shared(RefCell { value: [Numeric(U128(1))] }), rc: Shared(RefCell { value: 1 }), element_types: [Numeric(Unsigned { bit_size: 128 })], is_slice: false })])
---
After Constant Folding (step 23):
  b0(v0: [u128; 1]):
    v2 = call f1() -> u1
    enable_side_effects v2
    v3 = not v2
    enable_side_effects v3
    v5 = array_get v0, index u32 0 -> u128
    v6 = cast v5 as Field
    enable_side_effects u1 1
    v8 = cast v2 as Field
    v9 = cast v3 as Field
    v10 = mul v9, v6
    v12 = eq v10, Field 1
    v13 = not v12
    enable_side_effects u1 1
    v14 = cast v13 as u128
    v15 = cast v12 as u128
    v16 = unchecked_mul v14, v5
    v17 = unchecked_mul v15, v5
    v18 = unchecked_add v16, v17
    v19 = make_array [v18] : [u128; 1]
    return v19
}
brillig(inline) predicate_pure fn func_1 f1 {
  b0():
    return u1 1
}

--- Interpreter result after Constant Folding (step 23):
Ok([ArrayOrSlice(ArrayValue { elements: Shared(RefCell { value: [Numeric(U128(0))] }), rc: Shared(RefCell { value: 1 }), element_types: [Numeric(Unsigned { bit_size: 128 })], is_slice: false })])
---

```

The output changes back after step 25:
```console
❯ cargo run -q -p nargo_cli -- interpret --ssa-pass "step 25" --show-ssa-pass "step 25"
warning: Unsafe block must have a safety comment above it
  ┌─ src/main.nr:4:16
  │
4 │             if unsafe { func_1() } {
  │                ------ The comment must start with the "Safety: " word
  │

After EnableSideEffectsIf removal (step 25):
  b0(v0: [u128; 1]):
    v2 = call f1() -> u1
    v3 = not v2
    v5 = array_get v0, index u32 0 -> u128
    v6 = cast v5 as Field
    enable_side_effects u1 1
    v8 = cast v2 as Field
    v9 = cast v3 as Field
    v10 = mul v9, v6
    v12 = eq v10, Field 1
    v13 = not v12
    v14 = cast v13 as u128
    v15 = cast v12 as u128
    v16 = unchecked_mul v14, v5
    v17 = unchecked_mul v15, v5
    v18 = unchecked_add v16, v17
    v19 = make_array [v18] : [u128; 1]
    return v19
}
brillig(inline) predicate_pure fn func_1 f1 {
  b0():
    return u1 1
}

--- Interpreter result after EnableSideEffectsIf removal (step 25):
Ok([ArrayOrSlice(ArrayValue { elements: Shared(RefCell { value: [Numeric(U128(1))] }), rc: Shared(RefCell { value: 1 }), element_types: [Numeric(Unsigned { bit_size: 128 })], is_slice: false })])
---


```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+8f838fd807c534aeb2944d4b6bb9ab13d3c3a9ae (git version hash: bc60e6a4e2a80697baaf5564041d18af91b171f6, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8864

## Summary

Also uncomments a couple of tests that now can be written.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
@@ -94,12 +94,15 @@ impl<'ssa> Interpreter<'ssa> {
         &self.call_stack.first().expect("call_stack should always be non-empty").scope
     }
 
-    fn current_function(&self) -> &'ssa Function {
+    fn try_current_function(&self) -> Option<&'ssa Function> {
         let current_function_id = self.call_context().called_function;
-        let current_function_id = current_function_id.expect(
+        current_function_id.map(|current_function_id| &self.ssa.functions[&current_function_id])
+    }
+
+    fn current_function(&self) -> &'ssa Function {
+        self.try_current_function().expect(
             "Tried calling `Interpreter::current_function` while evaluating global instructions",
-        );
-        &self.ssa.functions[&current_function_id]
+        )
     }
 
     fn dfg(&self) -> &'ssa DataFlowGraph {
@@ -377,10 +380,15 @@ impl<'ssa> Interpreter<'ssa> {
     }
 
     fn side_effects_enabled(&self, instruction: &Instruction) -> bool {
-        match self.current_function().runtime() {
+        let Some(current_function) = self.try_current_function() else {
+            // If there's no current function it means we are evaluating global instructions
+            return true;
+        };
+
+        match current_function.runtime() {
             RuntimeType::Acir(_) => {
                 self.side_effects_enabled
-                    || !instruction.requires_acir_gen_predicate(&self.current_function().dfg)
+                    || !instruction.requires_acir_gen_predicate(&current_function.dfg)
             }
             RuntimeType::Brillig(_) => true,
         }
--- a/compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
@@ -376,9 +376,12 @@ impl<'ssa> Interpreter<'ssa> {
         try_vecmap(ids, |id| self.lookup(*id))
     }
 
-    fn side_effects_enabled(&self) -> bool {
+    fn side_effects_enabled(&self, instruction: &Instruction) -> bool {
         match self.current_function().runtime() {
-            RuntimeType::Acir(_) => self.side_effects_enabled,
+            RuntimeType::Acir(_) => {
+                self.side_effects_enabled
+                    || !instruction.requires_acir_gen_predicate(&self.current_function().dfg)
+            }
             RuntimeType::Brillig(_) => true,
         }
     }
@@ -389,9 +392,11 @@ impl<'ssa> Interpreter<'ssa> {
         instruction: &Instruction,
         results: &[ValueId],
     ) -> IResult<()> {
+        let side_effects_enabled = self.side_effects_enabled(instruction);
+
         match instruction {
             Instruction::Binary(binary) => {
-                let result = self.interpret_binary(binary)?;
+                let result = self.interpret_binary(binary, side_effects_enabled)?;
                 self.define(results[0], result);
                 Ok(())
             }
@@ -410,7 +415,7 @@ impl<'ssa> Interpreter<'ssa> {
             Instruction::Constrain(lhs_id, rhs_id, constrain_error) => {
                 let lhs = self.lookup(*lhs_id)?;
                 let rhs = self.lookup(*rhs_id)?;
-                if self.side_effects_enabled() && lhs != rhs {
+                if side_effects_enabled && lhs != rhs {
                     let lhs = lhs.to_string();
                     let rhs = rhs.to_string();
                     let lhs_id = *lhs_id;
@@ -433,7 +438,7 @@ impl<'ssa> Interpreter<'ssa> {
             Instruction::ConstrainNotEqual(lhs_id, rhs_id, constrain_error) => {
                 let lhs = self.lookup(*lhs_id)?;
                 let rhs = self.lookup(*rhs_id)?;
-                if self.side_effects_enabled() && lhs == rhs {
+                if side_effects_enabled && lhs == rhs {
                     let lhs = lhs.to_string();
                     let rhs = rhs.to_string();
                     let lhs_id = *lhs_id;
@@ -453,10 +458,16 @@ impl<'ssa> Interpreter<'ssa> {
                 }
                 Ok(())
             }
-            Instruction::RangeCheck { value, max_bit_size, assert_message } => {
-                self.interpret_range_check(*value, *max_bit_size, assert_message.as_ref())
+            Instruction::RangeCheck { value, max_bit_size, assert_message } => self
+                .interpret_range_check(
+                    *value,
+                    *max_bit_size,
+                    assert_message.as_ref(),
+                    side_effects_enabled,
+                ),
+            Instruction::Call { func, arguments } => {
+                self.interpret_call(*func, arguments, results, side_effects_enabled)
             }
-            Instruction::Call { func, arguments } => self.interpret_call(*func, arguments, results),
             Instruction::Allocate => {
                 self.interpret_allocate(results[0]);
                 Ok(())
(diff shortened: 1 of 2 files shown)


compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
```diff
@@ -376,9 +379,17 @@ impl<'ssa> Interpreter<'ssa> {
         try_vecmap(ids, |id| self.lookup(*id))
     }
 
-    fn side_effects_enabled(&self) -> bool {
-        match self.current_function().runtime() {
-            RuntimeType::Acir(_) => self.side_effects_enabled,
+    fn side_effects_enabled(&self, instruction: &Instruction) -> bool {
```
jfecher, line 382: I don't think side effects being enabled should be conditional based on the instruction - we should fix the instruction in question (array get) not having the correct check instead.

This is also an aside but I'd like the interpreter logic to remain mostly independent from internal details of SSA like `requires_acir_gen_predicate` so that we have 2 somewhat independent implementations that can be tested against each other to find differences between them - like the original issue here. If there is a future bug in `requires_acir_gen_predicate` for example I wouldn't want it to be in the interpreter as well automatically, I'd rather there be a discrepancy which could maybe be found through testing the two implementations against each other.

I think the above point is a bit pedantic but could be possible since `requires_acir_gen_predicate` is currently very acir-gen focused where the interpreter needs to consider side effects for more cases. Maybe we do want `requires_acir_gen_predicate` to be the final say on what things have side-effects and use it in more cases - @vezenovm previously had a brillig use case for it - in that case we should probably rename it and make it more clear that it's not an implementation detail, it defines what is and isn't a side-effect.

jfecher (changes requested):
I think we should avoid changing `side_effects_enabled` and instead fix the checks on each instruction that we expect to be different manually.

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
```diff
@@ -1066,7 +1097,7 @@ impl Interpreter<'_> {
         }
 
         // Disable this instruction if it is side-effectful and side effects are disabled.
-        if !self.side_effects_enabled() && binary.requires_acir_gen_predicate(self.dfg()) {
+        if !side_effects_enabled {
```
asterite, line 1100: @jfecher I'll try changing side_effects_enabled() to not rely on `requires_acir_gen_predicate`, but I just wanted to know that the code kind of already relied on this in some cases.

Do you mean for array get we should check whether the index is a constant or a dynamic value, then do one thing or the other? (that is, replicating the logic of `requires_acir_gen_predicate`?). I guess we should also replicate that logic in this case too, right?

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
vezenovm: > I don't think side effects being enabled should be conditional based on the instruction - we should fix the instruction in question (array get) not having the correct check instead.

Having a more independent SSA interpreter and ACIR runtime is a fair point. I had pushed @asterite in the issue discussion to use `requires_acir_gen_predicate` as to avoid duplicated logic, but if we are ok with (or actually want) duplicated logic I agree we should update array get manually. In that case we just need to make array get internally always return a value for safe indices and not consider the side effects var in those cases @asterite. 

> @vezenovm previously had a brillig use case for it - in that case we should probably rename it and make it more
clear that it's not an implementation detail, it defines what is and isn't a side-effect.

It is used more generally for determining whether we have side effects based upon predicates, so I think it would be good to rename. But that can be done separate of this PR.

asterite: As a side note, I'm looking at the acir codegen for ArrayGet and `requires_acir_gen_predicate` isn't used in that case. I mean, it's used at the top of `convert_ssa_instruction` to define a `predicate` variable, but then `predicate` isn't used for ArrayGet. It might be that, in the end, the side effects var isn't used when the index is a constant, but it's hard to see that relationship... which makes me think that relying on `requires_acir_gen_predicate` on the interpreter might not be the best thing to do, as you mention.

vezenovm: > I mean, it's used at the top of `convert_ssa_instruction` to define a `predicate` variable, but then `predicate` isn't used for ArrayGet. It might be that, in the end, the side effects var isn't used when the index is a constant, but it's hard to see that relationship...

This is an oversight in the ACIR gen that can be cleaned up. If you look at `convert_array_operation_inputs` you will see that `self.current_side_effects_enabled_var` is used and we duplicate the internal logic of `requires_acir_gen_predicate` by checking `DataFlowGraph::is_safe_index`.

asterite: I wonder if there's a way to not having to call `requires_acir_gen_predicate` during acir-gen or the interpreter. For example if an array get has a safe index, could `enable_side_effects u1 1` be inserted right before it, then `enable_side_effects var` be restored afterwards? Then the side effects var will always have the correct value for each instruction.

Though... I guess with the interpreter running after any SSA pass that wouldn't work.

vezenovm: > For example if an array get has a safe index, could `enable_side_effects u1 1` be inserted right before it, then `enable_side_effects var` be restored afterwards? Then the side effects var will always have the correct value for each instruction.

That is an interesting idea and could be explored separately. But either way this is an issue where the interpreter is considering safe array accesses as side effectual when they are not.

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
jfecher: I just looked in the interpreter code again and I guess we do already rely on `requires_acir_gen_predicate`. This PR is fine then but I do think we should make that function a bit more obvious.

jfecher (approved):

PR #8875, opened 2025-06-10
chore(ssa): Simplify truncate when the bit size we are truncating to is greater than the value's max bit size

vezenovm:
# Description

## Problem\*

No issue just something I noticed when looking at the Truncate simplification.

## Summary\*

If we are truncating to a `bit_size` but the `max_bit_size` is less than that value, we can return the value.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify.rs
@@ -141,7 +141,7 @@ pub(crate) fn simplify(
             try_optimize_array_set_from_previous_get(dfg, *array_id, *index_id, *value)
         }
         Instruction::Truncate { value, bit_size, max_bit_size } => {
-            if bit_size == max_bit_size {
+            if bit_size >= max_bit_size {
                 return SimplifiedTo(*value);
             }
             if let Some((numeric_constant, typ)) = dfg.get_numeric_constant_with_type(*value) {
--- a/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify.rs
@@ -490,4 +490,25 @@ mod tests {
             assert_normalized_ssa_equals(ssa, &expected);
         }
     }
+
+    #[test]
+    fn truncate_to_bit_size_bigger_than_value_max_bit_size() {
+        let src = "
+        acir(inline) fn main f0 {
+          b0(v0: u8):
+            v1 = truncate v0 to 16 bits, max_bit_size: 8
+            v2 = cast v1 as u16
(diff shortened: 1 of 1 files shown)


TomAFrench:
I think that we should perhaps just add this as a validation check. I would not expect us to generate these instructions in the first place.

jfecher (approved):
I think it's fine to have this check. It seems obviously correct and I don't see a need to move it to a separate validation pass to enforce we never generate instructions like this

asterite:
Maybe add a small test? It could be done using `Ssa::from_str_strimplifying`

TomAFrench:
>  It seems obviously correct

I agree that the simplification is correct but the fact that we would generate such a case implies that something is not behaving as we'd expect which warrants investigation.

jfecher:
> but the fact that we would generate such a case implies that something is not behaving as we'd expect which warrants investigation.

I could see us in the future adding a pass to calculate e.g. smaller max sizes for truncate instructions by trying to thread through more information (e.g. two `u8`s which were cast to `u32` and added, then the resulting `u32` truncated to `u16` should actually be a free case) so I don't think us generating it would necessarily be indicative of something concerning unless we're generating incorrect max bit widths.

vezenovm:
The only place I think where this may be possible at the moment is in remove_bit_shifts (although I don't see any performance improvements except the compilation time for `rollup-base-public`). Take a look at this truncate https://github.com/noir-lang/noir/blob/12bf0dbfeae032cd4d09622259a99abbcff950ce/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs#L151
The `max_bit_size` can come from `DataFlowGraph::get_value_max_num_bits` which I believe could return a max_bit_size less than the bit size of the type itself. I don't think us having a max_bit_size less than the bit_size we are truncating to is inherently bad SSA form. Especially as we add more value range analysis logic in the future. 

I have added a small test as well using `Ssa::from_str_strimplifying` that would have previously not simplified.

vezenovm:
To Tom's point, the test I wrote would not be generated by our initial SSA gen. We have this match case in `insert_safe_cast`:
```noir
            (
                Type::Numeric(NumericType::Unsigned { bit_size: incoming_type_size }),
                NumericType::Unsigned { bit_size: target_type_size },
            ) => {
                // If target size is smaller, we do a truncation
                if target_type_size < *incoming_type_size {
                    value =
                        self.builder.insert_truncate(value, target_type_size, *incoming_type_size);
                }
                value
            }
```
However, with this simplification we would not need to special case whether we insert a truncate in the initial SSA gen and leave it entirely to our instruction simplifications, which I view as a better separation of concerns. For now, I have left SSA gen untouched though. I am going to merge as I don't think this simplification implies ill formed SSA.

Issue #8846, opened 2025-06-09
Reference counters causing vector push back to always copy inside a loop


sirasistant:
### Aim

The macros in Aztec are generating code that boils down to this:
```noir
pub trait Serialize<let N: u32> {
    fn serialize(self) -> [Field; N];
}

impl Serialize<1> for Field {
    fn serialize(self) -> [Field; 1] {
        [self]
    }
}

pub unconstrained fn main(arr: [Field; 300]) {
    let mut serialized_args: [Field] = &[];
    let arr_serialized: [[Field; 1]; 300] =
        arr.map(|x: Field| -> [Field; 1] Serialize::serialize(x));
    for i in 0..arr.len() {
        serialized_args = serialized_args.append(arr_serialized[i].as_slice());
    }
    let calldata = serialized_args.push_front(611740813);
    std::println(calldata);
}
```

### Expected Behavior

The number of executed instructions should be in the order of magnitude of thousands.

### Bug

The number of executed instructions for the above code is 360k. Attaching flamegraph

![Image](https://github.com/user-attachments/assets/4040ca43-c147-41c2-9324-ae524dbb7be6)

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

Nice-to-have

### Would you like to submit a PR for this Issue?

None

jfecher:
Closed https://github.com/noir-lang/noir/pull/8876 - work to fix this performance issue has been pushed back since the original code in aztec_packages has been reworked to use a more efficient algorithm based on arrays instead of slices.

PR #8876, opened 2025-06-10
feat(optimization): Try to optimize ownership of assign statements

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/8846

## Summary\*

With this we go from 300 array copies down to 3 (not scaling with loop iterations) for the given test program.
Note that the example program uses slices so https://github.com/noir-lang/noir/pull/8867 is necessary to count the slice copies.

## Additional Context

This is a draft while I verify some corner cases with the assign check.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
This PR should be ready for review now, I can't seem to find any bugs with the approach here.

compiler/noirc_frontend/src/ownership/mod.rs
```diff
@@ -320,6 +320,19 @@ impl Context {
     }
 
     fn handle_call(&mut self, call: &mut crate::monomorphization::ast::Call) {
+        // Hack: We don't have references yet but pretend the array length builtin
```
vezenovm:
```suggestion
        // Hack: We don't have references of `self` yet but pretend the array length builtin
```
A bit pedantic but would this not be more accurate?

Also do we have an issue for references to self we want to link here?

vezenovm (approved):

compiler/noirc_frontend/src/ownership/mod.rs
jfecher: That'd be less accurate - I mean we don't have immutable references in general currently. We do have `&mut self` which could be used but we don't want array len to take a mutable self reference. So it takes a `self` but we don't actually want to inc_rc just for a .len() call, hence this hack.

vezenovm: > we don't have immutable references in general currently.

Ah yes. I was confused with the wording as I was thinking we do have _mutable_ references.

jfecher:
Found a couple of issues with this after thinking of some more cases where the previous value should not be moved. Just checking the loop index is not sufficient since:

```noir
for _ in 0 .. 10 {
    foo(bar);
}
for _ in 0 .. 10 {
    bar = &[];
}
```
In the above example if we just check the loop index, it will be `1` in both cases but we don't want to make `foo(bar)` a move since on the next iteration we'd be using a moved value.

jfecher:
Abandoning this work for now to focus more on fixing bugs before 1.0.
This requires more than a day's work to fix it seems since I think we'll need some new tracking structures to track when this optimization is valid since the existing loop_index and branches structures are insufficient.

PR #8877, opened 2025-06-10
fix(mem2reg): Keep last store for reference in array used only in an array get

### LINKED ISSUE #8803: Execution crash: `condition value is not a boolean: MismatchedBitSize` (reassign `mut &mut`, no if-then)
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8728 on top of https://github.com/noir-lang/noir/pull/8801

The following program crashes during execution:

```noir
unconstrained fn main() {
    let mut b: &mut bool = (&mut true);
    b = b; // Remove this and it works.
    {
        let mut i = 0;
        // this causes a crash: while (*b)
        // this causes runtime error: while (*b > false)
        while (*b) {
            if i == 0 {
                break;
            }
        }
    };
}
```

### Expected Behavior

It should not crash. 

We are re-assigning a mutable reference variable, but not doing it conditionally. Apart from that this is the same program as https://github.com/noir-lang/noir/issues/8739 , which used to crash during compilation with the same error, but after @vezenovm's fixes that works. 

Also related to #8783 which might lead to denying the ability to reassign these variables completely. 

### Bug

```console
Message:  condition value is not a boolean: MismatchedBitSize { value_bit_size: 254, expected_bit_size: 1 }
Location: /Users/aakoshh/Work/aztec/noir/acvm-repo/brillig_vm/src/lib.rs:455
```

As the comments say, if we change the loop condition to `while ((*b) <= false) {` we get an error instead:
```console
error: Assertion failed: 'Bit size for rhs 254 does not match op bit size 1'
  ┌─ src/main.nr:8:16
  │
8 │         while (*b > false) {
  │                ----------
  │
  = Call stack:
    1. src/main.nr:8:16

Failed assertion
```

### To Reproduce

See above. 

Same behaviour on master at 7a252703d7f2c11409f9b8034df8224c438f768f 

If we go with the non-crashing version, `nargo interpret` shows it's somewhere in `mem2reg` still:

```console
❯ cargo run -q -p nargo_cli -- interpret --ssa-pass "step 6" --ssa-pass "step 7" --show-ssa-pass "step 6" --show-ssa-pass "step 7"
After Inlining (step 6):
brillig(inline) fn main f0 {
  b0():
    v0 = allocate -> &mut u1
    store u1 1 at v0
    v2 = allocate -> &mut &mut u1
    store v0 at v2
    v3 = load v2 -> &mut u1
    store v3 at v2
    v4 = allocate -> &mut Field
    store Field 0 at v4
    jmp b1()
  b1():
    v6 = load v2 -> &mut u1
    v7 = load v6 -> u1
    v9 = lt u1 0, v7
    jmpif v9 then: b2, else: b3
  b2():
    v10 = load v4 -> Field
    v11 = eq v10, Field 0
    jmpif v11 then: b4, else: b5
  b3():
    return
  b4():
    jmp b3()
  b5():
    jmp b1()
}

--- Interpreter result after Inlining (step 6):
Ok([])
---
After Mem2Reg (step 7):
brillig(inline) fn main f0 {
  b0():
    v0 = allocate -> &mut u1
    v1 = allocate -> &mut &mut u1
    store v0 at v1
    v2 = allocate -> &mut Field
    store Field 0 at v2
    jmp b1()
  b1():
    v4 = load v1 -> &mut u1
    v5 = load v4 -> u1
    v7 = lt u1 0, v5
    jmpif v7 then: b2, else: b3
  b2():
    v8 = load v2 -> Field
    v9 = eq v8, Field 0
    jmpif v9 then: b4, else: b5
  b3():
    return
  b4():
    jmp b3()
  b5():
    jmp b1()
}

--- Interpreter result after Mem2Reg (step 7):
Err(Internal(UninitializedReferenceValueLoaded { value: "*v0 = None" }))
---


```

We can see the `store u1 1 at v0` is gone. Maybe it's because `v2` is `&mut &mut u1` and `store v0 at v2` doesn't count as use?

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+b21a44cd5fff5261025fb52e091a931ece8a9c91 (git version hash: d9218d87382b4fec32f4d1a0b33ce81b648e41bb, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #8803 

## Summary\*

Similar to https://github.com/noir-lang/noir/pull/8801 we need to mark each alias of an array used in an array get as being unsafe for removal.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -476,6 +476,9 @@ impl<'f> PerFunctionContext<'f> {
             }
             Instruction::ArrayGet { array, .. } => {
                 let result = self.inserter.function.dfg.instruction_results(instruction)[0];
+                references.for_each_alias_of(*array, |_, alias| {
+                    self.instruction_input_references.insert(alias);
+                });
```
aakoshh: I think I tried this and got another test failure. 

What are the rules though, why is `ArrayGet` an "instruction" but `MakeArray` is not?

aakoshh: This is how to replicate a failing test on top of this PR:

```shell
git checkout origin/af/8728-test-fixes
NOIR_AST_FUZZER_SHOW_AST=1 NOIR_ARBTEST_SEED=0x37dabe9c00100000 cargo test -p noir_ast_fuzzer_fuzz min_vs_full
```

This is the program:
```rust
unconstrained fn main(
    a: pub bool,
    b: call_data(1) str<4>,
    c: (Field, bool, bool, bool, Field),
) -> pub str<2> {
    let mut ctx_limit: u32 = 25;
    { while (*func_3(c, (&mut ctx_limit))[0]) {} };
    "SO"
}
unconstrained fn func_3(
    mut a: (Field, bool, bool, bool, Field),
    _ctx_limit: &mut u32,
) -> [&mut bool; 1] {
    {
        let mut idx_b: u32 = 0;
        loop {
            if (idx_b == 1) {
                break
            } else {
                idx_b = (idx_b + 1);
                continue;
                break;
                println(a);
                {
                    let mut idx_c: u32 = 0;
                    loop {
                        if (idx_c == 9) {
                            break
                        } else {
                            idx_c = (idx_c + 1);
                            {
                                let mut idx_d: u32 = 0;
                                while false {
                                    if (idx_d == 8) {
                                        break
                                    } else {
                                        idx_d = (idx_d + 1);
                                        assert(((true >= a.3) >= true), "ATU");
                                        let e = a.0;
                                        continue;
                                        assert((false < a.1), "KXK");
                                    }
                                }
                            };
                            ()
                        }
                    }
                };
                continue;
            }
        }
    };
    [(&mut a.3)]
}

```

and the input:
```toml
a = true
b = "\f2\u0010]"
c = [
    "0x1914084e4226d1ed6dcb87b49f18142d70924c68541acfd7ce73701aee675979",
    true,
    true,
    true,
    "0x1eaa901d57690be79c297228e0ea210d616f3bb0cf697a7af6a469b203b94b82",
]
```

Sorry I didn't have time to minimise it by hand.

```
❯ cargo run -q -p nargo_cli -- execute --silence-warnings
Message:  condition value is not a boolean: MismatchedBitSize { value_bit_size: 254, expected_bit_size: 1 }
Location: /Users/aakoshh/Work/aztec/noir/acvm-repo/brillig_vm/src/lib.rs:455
```

aakoshh: Ah, unfortunately the minimiser script turned it into an infinite loop as well, so `interpret` doesn't finish on it. 

The original AST is this:
```rust
global G_A: Field = 169520100645089417367167447589555010165;
unconstrained fn main(a: pub bool, b: call_data(1) str<4>, c: (Field, bool, bool, bool, Field)) -> pub str<2> {
    let mut ctx_limit: u32 = 25;
    {
        let mut idx_d: u32 = 0;
        while (*func_3(c, (&mut ctx_limit))[0]) {
            if (idx_d == 8) {
                break
            } else {
                idx_d = (idx_d + 1);
                let mut e = c;
                {
                    let mut idx_f: u32 = 0;
                    while (true > c.1) {
                        if (idx_f == 1) {
                            break
                        } else {
                            idx_f = (idx_f + 1);
                            assert((c.0 != ((c.1 as Field) / G_A)), "PHM");
                            let mut g: str<2> = {
                                "MT"
                            };
                            ()
                        }
                    }
                };
            }
        }
    };
    "SO"
}
#[inline_always]
unconstrained fn func_3(mut a: (Field, bool, bool, bool, Field), _ctx_limit: &mut u32) -> [&mut bool; 1] {
    assert(((a.3 as Field) == (a.1 as Field)), "RAV");
    println(a);
    println(a);
    {
        let mut idx_b: u32 = 0;
        loop {
            if (idx_b == 1) {
                break
            } else {
                idx_b = (idx_b + 1);
                continue;
                break;
                println(a);
                {
                    let mut idx_c: u32 = 0;
                    loop {
                        if (idx_c == 9) {
                            break
                        } else {
                            idx_c = (idx_c + 1);
                            {
                                let mut idx_d: u32 = 0;
                                while false {
                                    if (idx_d == 8) {
                                        break
                                    } else {
                                        idx_d = (idx_d + 1);
                                        assert(((true >= a.3) >= true), "ATU");
                                        let e = a.0;
                                        continue;
                                        assert((false < a.1), "KXK");
                                    }
                                }
                            };
                            ()
                        }
                    }
                };
                continue;
            }
        }
    };
    [(&mut a.3)]
}
```

aakoshh: Here's a minimised version:

```rust
unconstrained fn main(c: bool) {
    let mut idx_d: u32 = 0;
    while (*func_3(c)[0]) {
        if (idx_d == 1) {
            break
        } else {
            idx_d = (idx_d + 1);
        }
    }
}
#[inline_always]
unconstrained fn func_3(mut a: bool) -> [&mut bool; 1] {
    let mut idx_b: u32 = 0;
    loop {
        if (idx_b == 1) {
            break
        } else {
            idx_b = (idx_b + 1);
        }
    }
    [&mut a]
}
```
with input
```toml
c = true
```

Here's the SSA where it changes:

```
❯ cargo run -q -p nargo_cli -- interpret --ssa-pass "(step 5)" --ssa-pass "(step 4)" --show-ssa-pass "(step 5)" --show-ssa-pass "(step 4)"
After Removing Paired rc_inc & rc_decs (step 4):
brillig(inline) fn main f0 {
  b0(v0: u1):
    v1 = allocate -> &mut u32
    store u32 0 at v1
    jmp b1()
  b1():
    v4 = call f1(v0) -> [&mut u1; 1]
    v5 = array_get v4, index u32 0 -> &mut u1
    v6 = load v5 -> u1
    jmpif v6 then: b2, else: b3
  b2():
    v7 = load v1 -> u32
    v9 = eq v7, u32 1
    jmpif v9 then: b4, else: b5
  b3():
    return
  b4():
    jmp b3()
  b5():
    v10 = load v1 -> u32
    v11 = add v10, u32 1
    store v11 at v1
    jmp b6()
  b6():
    jmp b1()
}
brillig(inline_always) fn func_3 f1 {
  b0(v0: u1):
    v1 = allocate -> &mut u1
    store v0 at v1
    v2 = allocate -> &mut u32
    store u32 0 at v2
    jmp b1()
  b1():
    v4 = load v2 -> u32
    v6 = eq v4, u32 1
    jmpif v6 then: b2, else: b3
  b2():
    jmp b4()
  b3():
    v7 = load v2 -> u32
    v8 = add v7, u32 1
    store v8 at v2
    jmp b5()
  b4():
    v9 = make_array [v1] : [&mut u1; 1]
    return v9
  b5():
    jmp b1()
}

--- Interpreter result after Removing Paired rc_inc & rc_decs (step 4):
Ok([])
---
After Preprocessing Functions (step 5):
brillig(inline) fn main f0 {
  b0(v0: u1):
    v1 = allocate -> &mut u32
    store u32 0 at v1
    jmp b1()
  b1():
    v4 = call f1(v0) -> [&mut u1; 1]
    v5 = array_get v4, index u32 0 -> &mut u1
    v6 = load v5 -> u1
    jmpif v6 then: b2, else: b3
  b2():
    v7 = load v1 -> u32
    v9 = eq v7, u32 1
    jmpif v9 then: b4, else: b5
  b3():
    return
  b4():
    jmp b3()
  b5():
    v10 = load v1 -> u32
    v11 = add v10, u32 1
    store v11 at v1
    jmp b6()
  b6():
    jmp b1()
}
brillig(inline_always) fn func_3 f1 {
  b0(v0: u1):
    v1 = allocate -> &mut u1
    v2 = allocate -> &mut u32
    store u32 0 at v2
    jmp b1()
  b1():
    v4 = load v2 -> u32
    v6 = eq v4, u32 1
    jmpif v6 then: b2, else: b3
  b2():
    v9 = make_array [v1] : [&mut u1; 1]
    return v9
  b3():
    v7 = load v2 -> u32
    v8 = add v7, u32 1
    store v8 at v2
    jmp b1()
}

--- Interpreter result after Preprocessing Functions (step 5):
Err(Internal(UninitializedReferenceValueLoaded { value: "*v1 = None" }))
---
```

The difference is that in the 2nd pass `store v0 at v1` is gone from `func_3`, despite `v1` being used in `v9 = make_array [v1] : [&mut u1; 1]`

aakoshh: Notably the `array_get` is in `f0` while the `make_array` is in `f1`, I'm not sure if that can play a role.

vezenovm: I was suspicious that the array get update was not enough for the case where we have the array get in a separate function to the make array. It looks like our marking of references unknown for return terminators is insufficient at the moment. Thanks for the minimized repro.

vezenovm: > What are the rules though, why is `ArrayGet` an "instruction" but `MakeArray` is not?

The `instruction_input_references` adds a gate to our last stores removal optimization at the end of mem2reg. While mem2reg tracks last stores and marks them for removal per block, that optimization looks to remove any hanging stores we missed across the entire function. You can see where the map is accessed in `is_store_alias_used`. We don't want to insert on `MakeArray` because if that array is never used we should still be able to remove any stores. 

Another fix for handling some of the ArrayGet cases could be to go through all possible expressions in a block to see whether the last store is an alias to any other values (inside of `is_store_alias_used`). However, this would only work for the case where the store and the `MakeArray` are contained to the same block (without extra logic that is, thus `instruction_input_references` comes into play or we would have to go through every expression in every block). If we were to insert every element in a `MakeArray` into `instruction_input_references` we would prevent removing stores from a MakeArray that is never used. I have added a test case that remove stores in this case as well.

For the case you listed we needed to mark the aliases of return values in `instruction_input_references`. However, we still need to mark the aliases of an array used in an `ArrayGet` in `instruction_input_references` as we have no other way to tell whether an array was used.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -225,6 +225,7 @@ impl<'f> PerFunctionContext<'f> {
     ) -> bool {
         let reference_parameters = self.reference_parameters();
 
+        // Check whether the store address has an alias that is cross the entry point boundary (e.g. a Call or Return)
```
vezenovm:
```suggestion
        // Check whether the store address has an alias that crosses an entry point boundary (e.g. a Call or Return)
```

aakoshh:
Stores are still removed in https://github.com/noir-lang/noir/issues/8909 for nested arrays, but if you think this should be merged on its own and that one tackled as a separate issue, that's fine with me!

vezenovm:
> Stores are still removed in #8909 for nested arrays, but if you think this should be merged on its own and that one tackled as a separate issue, that's fine with me!

Let's just merge this one first as they are related but separate fixes. I will also start looking at #8909.

aakoshh (approved):
Another bug bites the dust 🐛

PR #8882, opened 2025-06-11
chore: add return value to every execution_success program that produces an output

asterite:
# Description

## Problem

No issue.

## Summary

Given that the SSA interpreter might have some bugs, I thought it would be nice to run it against all "execution_success" test programs. Ideally we'd check that the circuit output produced by the SSA interpreter is the same as the one produced by `nargo execute`. Initially I though we could run `nargo execute`, get the output, the run the SSA interpreter and compare the output. However, it might be easier for the SSA interpreter to check the return value against the `return` value in "Prover.toml". Given that this is missing from most "execution_success" tests, this PR adds those as an initial step towards that.

## Documentation

Check one:

# PR Checklist


asterite:
@TomAFrench I forgot to commit a change in `execute.rs` that requires a Prover.toml (and a `return` value) to be there for programs that produce output... in case you'd like to review that last commit.

PR #8883, opened 2025-06-11
feat: SSA interpreter checks return value

asterite:
# Description

## Problem

No issue, but I think @aakoshh mentioned this as a potential improvements.

This will also be useful to run `nargo interpret` on all execution_success test programs, combined with #8882.

## Summary

For example, given this program:

```noir
fn main() -> pub Field {
    0
}
```

and this Prover.toml:

```toml
return = 1
```

running `nargo interpret` we get this output:

```
Ok([Numeric(Field(0))])
---
Error: interpreter produced an unexpected result.
Expected result: [Numeric(Field(1))]
Actual result:   [Numeric(Field(0))]
```

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/ast_fuzzer/src/compare/interpreted.rs
+++ b/tooling/ast_fuzzer/src/compare/interpreted.rs
@@ -194,7 +194,7 @@ pub fn input_values_to_ssa(abi: &Abi, input_map: &InputMap) -> Vec<Value> {
 /// Convert one ABI encoded input to what the SSA interpreter expects.
 ///
 /// Tuple types are returned flattened.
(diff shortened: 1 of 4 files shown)


aakoshh:
Although actually one of the bugs you fixed did correct itself, so we would have to carry the “any disagreed” flag to the end.

Never mind though, I’m happy either way.

asterite:
I think that even if the return value changes in an intermediate pass and then changes again later on to the correct value, it might mean there's a bug in the interpreter so maybe the value fixing itself is just a coincidence.

PR #8884, opened 2025-06-11
fix: do not hoist control dependent cast

### LINKED ISSUE #8841: ACIR != Brillig: `Failed constraint`, multiply with overflow on branch that never runs
### Aim

Discovered by the AST fuzzer

The following program behaves differently in Brillig and ACIR:

```rust
fn main(a: bool, c: i8) -> pub i16 {
    if unsafe { func_2(1) } {
        for _ in 0..1 {
            if a {
                assert(((c as i64) != (((c as i16) * 4877) as i64)), "NWL");
            };
        }
        2
    } else {
        3
    }
}
unconstrained fn func_2(a: u32) -> bool {
    print(a);
    false
}
```

Input:
```
a = false
c = "0x000000000000000000000000000000000000000000000000000000000000001b"
```

### Expected Behavior

It should return `3` because `func_2` returns `false`. 

### Bug

```console
$ cargo run -q -p nargo_cli -- execute --silence-warnings
error: Failed constraint
  ┌─ src/main.nr:5:41
  │
5 │                 assert(((c as i64) != (((c as i16) * 4877) as i64)), "NWL");
  │                                         -----------------
  │
  = Call stack:
    1. src/main.nr:5:41

1
$ cargo run -q -p nargo_cli -- execute --silence-warnings --force-brillig
1[fuzz_testing] Circuit witness successfully solved
[fuzz_testing] Circuit output: Field(3)
```

The SSA Interpreter gets internal error after flattening:
```
❯ cargo run -q -p nargo_cli -- interpret
...
--- Interpreter result after Mem2Reg (step 16):
Ok([Numeric(I16(3))])
---
--- Interpreter result after Flattening (step 17):
Err(Internal(ConstantDoesNotFitInType { constant: 131679, typ: Unsigned { bit_size: 16 } }))

```

### To Reproduce

See above. 

The problem goes away in either of the following cases:
* If we remove the `println(a)`
* If we remove the `if a` and just leave the `if` body
* If we remove the `for _ 0 .. 1` and just leave the loop body

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+8f838fd807c534aeb2944d4b6bb9ab13d3c3a9ae (git version hash: fd917e3feff0fcc70ab45bb476fe627e771c6e23, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #8841

## Summary\*
Cast instruction were marked as 'can be hoisted', even if they are control dependent, so loop invariant code motion were hoisting them out of their conditional statement.
However, range-check are properly marked as control-dependent, so when the cast is done after a range-check, and the cast is moved, the related range-check is not moved, making the cast not safe anymore.

## Documentation\*

Check one:

# PR Checklist\*


guipublic:
I close it because I believe I did not branch it off master. I will create another PR

PR #8885, opened 2025-06-11
fix: always error if integer literal doesn't fit its type on the fron…

### LINKED ISSUE #8873: Literals that cannot fit into a type sometimes error during monomorphization
For example this program:

```noir
fn main() {
    let upper = 123456789101112;
    for _ in 0..upper {}
}
```

When coding in VSCode you don't get any error. But when you run `nargo compile` you get:

```
error: The value `123456789101112` cannot fit into `u32` which has range `0..=4294967295`
  ┌─ src/main.nr:2:17
  │
2 │     let upper = 123456789101112;
  │                 ---------------
  │
  = Call stack:
    1. src/main.nr:2:17

```

It's good that we get an error, however it would be nice if the error was produced by the frontend.

One way to do this could be to keep track of each numeric literal in the current `FunctionContext`. After popping a function context, and after making type variable default to their default type, we could check if literals fit their associated type and error if not.

### LINKED ISSUE #8879: Compiler crash: `Function has no terminator in block`
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8872

The following program crashes during SSA validation:

```rust
fn main(a: i32, b: [[i32; 3]; 4]) {
    if (a <= 114940085) {
        let mut c = b;
        for idx_d in 507459766..507459775 {
            for _idx_e in -27254740..-27254731 {
                c[3] = b[1];
                c[1] = [idx_d, (idx_d * b[0][1]), -idx_d];
            }
        }
    }
}
```

I discovered it by accident trying to minimize a failure. When I replaced `-idx_e` with `-idx_d` in `c[1] = [..., ..., -idx_e]`, it crashed. 

### Expected Behavior

### Bug

```console
Message:  Function f0 has no terminator in block b5
Location: compiler/noirc_evaluator/src/ssa/validation/mod.rs:149
```

Things to note:
* When `idx_e` was part of `c[1] = ...` then the compiler knew that its type was `i32`. Now it thinks it's `u32` because it's "just" a loop index variable, but it does not reject the code. 
* If I comment out the `if (a <= 114940085) ` bit then the frontend _does_ reject the code with `The value `-27254740` cannot fit into `u32` which has range `0..=4294967295``, which is strange, given that `a` is a dynamic input.

Related to #8011 

### To Reproduce

See above. 

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+501dc08ed0756a69b309b904091adb22776d55c8 (git version hash: bf792ddb97efb688725732aab47df15fe4228b52, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
…tend side

# Description

## Problem

Resolves #8873
Resolves #8879

## Summary

This check was previously done in a few cases:
- When you had `let x: u8 = ...`, that is, on let statements
- During SSA-gen when trying to create a numeric value from an integer literal

But I think this wasn't always checked, as seen in #8879 where a compiler crash happens.

A nice thing about this is that these errors will now also show up in VSCode.

There are a couple of cases removed from the old `overflowing_int`, but I checked and unary error for prefix expressions when you use `-` on a u32 still happen, as there's another part of the compiler that does that check too. Additionally, all tests still pass.

## Additional Context

This change made `nargo debug` fail, because it was generating some debug functions where all arguments to it had the same `Index` type (a generic), but I think the intention was to have a different generic for each argument.

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/debug/mod.rs
```diff
@@ -606,21 +606,23 @@ pub fn build_debug_crate_file() -> String {
                 // The variable signature has to be generic as Noir supports using any polymorphic integer as an index.
                 // If we were to set a specific type for index signatures here, such as `Field`, we will error in
                 // type checking if we attempt to index with a different type such as `u8`.
+                let indexes =
```
jfecher:
```suggestion
                let indices =
```

compiler/noirc_frontend/src/elaborator/lints.rs
```diff
@@ -206,67 +206,52 @@ pub(super) fn unnecessary_pub_argument(
...
-                        ty: annotated_type.clone(),
+                        ty: typ.clone(),
                         range: format!("-{}..={}", min, max),
                         location,
                     });
                 }
             }
             _ => (),
         },
-        HirExpression::Prefix(expr) => {
```
jfecher, line 253: Why don't we handle prefix/infix expressions any more?

jfecher (approved):

compiler/noirc_frontend/src/debug/mod.rs
```diff
@@ -606,21 +606,23 @@ pub fn build_debug_crate_file() -> String {
...
+                    (0..n).map(|i| format!["Index{i}"]).collect::<Vec<String>>().join(", ");
                 let var_sig =
-                    (0..n).map(|i| format!["_v{i}: Index"]).collect::<Vec<String>>().join(", ");
+                    (0..n).map(|i| format!["_v{i}: Index{i}"]).collect::<Vec<String>>().join(", ");
                 let vars = (0..n).map(|i| format!["_v{i}"]).collect::<Vec<String>>().join(", ");
                 format!(
                     r#"
                 #[oracle(__debug_member_assign_{n})]
-                unconstrained fn __debug_oracle_member_assign_{n}<T, Index>(
+                unconstrained fn __debug_oracle_member_assign_{n}<T, {indexes}>(
```
asterite:
```suggestion
                unconstrained fn __debug_oracle_member_assign_{n}<T, {indices}>(
```

compiler/noirc_frontend/src/debug/mod.rs
```diff
@@ -606,21 +606,23 @@ pub fn build_debug_crate_file() -> String {
...
                     _var_id: u32, _value: T, {var_sig}
                 ) {{}}
-                unconstrained fn __debug_inner_member_assign_{n}<T, Index>(
+                unconstrained fn __debug_inner_member_assign_{n}<T, {indexes}>(
```
asterite:
```suggestion
                unconstrained fn __debug_inner_member_assign_{n}<T, {indices}>(
```

compiler/noirc_frontend/src/debug/mod.rs
```diff
@@ -606,21 +606,23 @@ pub fn build_debug_crate_file() -> String {
...
                     var_id: u32, value: T, {var_sig}
                 ) {{
                     __debug_oracle_member_assign_{n}(var_id, value, {vars});
                 }}
-                pub fn __debug_member_assign_{n}<T, Index>(var_id: u32, value: T, {var_sig}) {{
+                pub fn __debug_member_assign_{n}<T, {indexes}>(var_id: u32, value: T, {var_sig}) {{
```
asterite:
```suggestion
                pub fn __debug_member_assign_{n}<T, {indices}>(var_id: u32, value: T, {var_sig}) {{
```

PR #8887, opened 2025-06-11
chore: interpret all execution_success programs

asterite:
# Description

## Problem

No issue, but will help us improve the interpreter.

## Summary

Runs `nargo interpret` against all execution_success test programs, except a few that currently don't work well.

Also fixes one issue where the ABI was built in a way that doesn't match how it's built everywhere else (this made some tests fail, which were around 43, and then the number got down to 35).

I decided to not keep fixing issues here as there seem to be a few of them. Some of them might be easier to fix (like the "IntrinsicArgumentCountMismatch" ones), but in any case we can probably fix them in follow-up PRs.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_driver/src/abi_gen.rs
+++ b/compiler/noirc_driver/src/abi_gen.rs
@@ -23,7 +23,7 @@ use noirc_frontend::{
 
 /// Arranges a function signature and a generated circuit's return witnesses into a
 /// `noirc_abi::Abi`.
-pub(super) fn gen_abi(
+pub fn gen_abi(
     context: &Context,
     func_id: &FuncId,
     return_visibility: Visibility,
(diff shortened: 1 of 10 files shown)


jfecher (approved):
This is very nice to have. I only wish for the tests that are currently failing we could assert what we expect the error to be

asterite:
> It'd just help to remind us when these issues are fixed to un-ignore the relevant test.

I was thinking about this some minutes ago. Instead of ignoring these tests, it would be nice to still run them and have them pass if they fail, but have them fail if they pass. I'll try to do it in a follow-up PR for all ignored tests (at least in nargo_cli).

asterite:
> It'd just help to remind us when these issues are fixed to un-ignore the relevant test.

I just tried implementing this but it's tricky because for interpret tests, some pass normally but fail with `--force-brillig` so we'd have to ignore only those. The way I was doing it was putting `#[should_panic]` on tests that we ignore. The problem is that the tests are generated using `test_case` and there doesn't seem to be a way to say that just some test cases are expected to fail.

For now I'll abandon this, though I'll try to fix make the interpreter tests pass.

Issue #8779, opened 2025-06-03
Compiler crash on dynamic index of array of reference: Could not resolve some references to the array


vezenovm:
### Aim

```noir
fn main(x: u32) -> pub bool {
    let d: [&mut bool; 1] = [&mut true];
    *d[x]
}
```

### Expected Behavior

This code should compile successfully or error out cleanly.

### Bug

Running `nargo compile` will give the following error:
```
The application panicked (crashed).
Message:  Expected RuntimeError to have a location. Error message: Could not resolve some references to the array. All references must be resolved at compile time
Location: compiler/noirc_evaluator/src/errors.rs:214
```
This is the initial SSA:
```
After Initial SSA:
acir(inline) fn main f0 {
  b0(v0: u32):
    v1 = allocate -> &mut u1
    store u1 1 at v1
    v3 = make_array [v1] : [&mut u1; 1]
    v5 = eq v0, u32 0
    constrain v0 == u32 0, "Index out of bounds"
    v6 = array_get v3, index v0 -> &mut u1
    v7 = load v6 -> u1
    return v7
}
```
Before and after DIE:
```
After Check u128 mul overflow (1) (step 29):
acir(inline) predicate_pure fn main f0 {
  b0(v0: u32):
    v1 = allocate -> &mut u1
    store u1 1 at v1
    v3 = make_array [v1] : [&mut u1; 1]
    constrain v0 == u32 0, "Index out of bounds"
    v5 = load v1 -> u1
    return v5
}

After Dead Instruction Elimination (1) (step 30):
acir(inline) predicate_pure fn main f0 {
  b0(v0: u32):
    v1 = allocate -> &mut u1
    constrain v0 == u32 0, "Index out of bounds"
    v3 = load v1 -> u1
    return v3
}
```
That store got removed even though it was in use by the load. 

### To Reproduce

1. Copy snippet above
2. Run nargo compile

### Workaround

None

### Project Impact

Blocker

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+5984575435cde6c01df3454053da471769f3d856 (git version hash: 5984575435cde6c01df3454053da471769f3d856, is dirty: false)

### Would you like to submit a PR for this Issue?

None

vezenovm:
Seeing now this is a simplification of https://github.com/noir-lang/noir/issues/8750

vezenovm:
[...]

Issue #8750, opened 2025-06-02
Compiler crash: Expected RuntimeError to have a location. Could not resolve some references to the array.


aakoshh:
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8728

The following program crashes the compiler:
```rust
fn main(c: u32) -> pub bool {
    let b: [&mut bool; 2] = [&mut false, &mut true];
    *b[c % 2]
}
```

### Expected Behavior

Not to crash.

### Bug

```console
❯ cargo run -q -p nargo_cli -- compile --force
The application panicked (crashed).
Message:  Expected RuntimeError to have a location. Error message: Could not resolve some references to the array. All references must be resolved at compile time
Location: compiler/noirc_evaluator/src/errors.rs:214
```

### To Reproduce

See above.

### Workaround

None

### Project Impact

None

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+7e6d7cdb32c806aa0622c1c73e2255a79d93637d (git version hash: 8d9e69fd20a58b1528ab3403115b08e1b96647b6, is dirty: true)

### Would you like to submit a PR for this Issue?

None

vezenovm:
We may have to ban references in dynamic arrays until we code gen references in ACIR gen. 

This is the SSA before DIE:
```
After Check u128 mul overflow (1) (step 29):
acir(inline) predicate_pure fn main f0 {
  b0(v0: u32):
    v1 = allocate -> &mut u1
    store u1 0 at v1
    v3 = allocate -> &mut u1
    store u1 1 at v3
    v5 = make_array [v1, v3] : [&mut u1; 2]
    v6 = truncate v0 to 1 bits, max_bit_size: 32
    v8 = lt v6, u32 2
    constrain v8 == u1 1, "Index out of bounds"
    v9 = array_get v5, index v6 -> &mut u1
    v10 = load v9 -> u1
    return v10
}
```
During DIE we remove the stores as we assume we have finished flattening and all references have been resolved:
```
After Dead Instruction Elimination (1) (step 30):
acir(inline) predicate_pure fn main f0 {
  b0(v0: u32):
    v1 = allocate -> &mut u1
    v2 = allocate -> &mut u1
    v3 = make_array [v1, v2] : [&mut u1; 2]
    v4 = truncate v0 to 1 bits, max_bit_size: 32
    v6 = lt v4, u32 2
    constrain v6 == u1 1, "Index out of bounds"
[...]

PR #8888, opened 2025-06-11
chore: Ban dynamic array indices with reference elements

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/8750

## Summary\*

It's not clear how we'll support this in the future so let's ban any dynamic indexing of arrays which have reference elements. We can always lift this later when we're able. Since this panics at the moment it is not expected to break existing code.

## Additional Context

This does not fix https://github.com/noir-lang/noir/issues/8779 since the array get is optimized out beforehand there.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/validation/dynamic_array_indices.rs
+++ b/compiler/noirc_evaluator/src/ssa/validation/dynamic_array_indices.rs
@@ -1,0 +1,82 @@
+use crate::{
+    errors::{RtResult, RuntimeError},
+    ssa::{
+        ir::{function::Function, instruction::Instruction},
+        ssa_gen::Ssa,
+    },
+};
+
+impl Ssa {
+    /// Verifies there are no `array_get` or `array_set` instructions remaining
+    /// with dynamic indices where the element type may contain a reference type.
+    /// This effectively bans dynamic-indexing of arrays with reference elements
+    /// since we cannot guarantee we optimize references out of the program in that case.
+    ///
+    /// This pass expects to be run late in the Ssa pipeline such that all array indices
+    /// used are either constant or derived from an input to the program. E.g. we expect
+    /// optimizations from inlining, flattening, and mem2reg to already be complete.
+    pub(crate) fn verify_no_dynamic_indices_to_references(self) -> RtResult<Ssa> {
+        for function in self.functions.values() {
+            function.verify_no_dynamic_indices_to_references()?;
+        }
+        Ok(self)
+    }
+}
+
+impl Function {
+    pub(crate) fn verify_no_dynamic_indices_to_references(&self) -> RtResult<()> {
+        if self.runtime().is_brillig() {
+            return Ok(());
+        }
+
+        for block in self.reachable_blocks() {
+            for instruction in self.dfg[block].instructions() {
+                match &self.dfg[*instruction] {
+                    Instruction::ArrayGet { array, index, .. }
+                    | Instruction::ArraySet { array, index, .. } => {
+                        let array_type = self.dfg.type_of_value(*array);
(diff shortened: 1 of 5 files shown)


compiler/noirc_evaluator/src/ssa/validation/dynamic_array_indices.rs
```diff
@@ -0,0 +1,82 @@
...
+                constrain v6 == u1 1, "Index out of bounds"
+                v8 = array_get v3, index v4 -> &mut u1
+                v9 = load v8 -> u1
+                return v9
+            }"#;
+
+        let ssa = Ssa::from_str(src).unwrap();
+        let result = ssa.verify_no_dynamic_indices_to_references();
+        assert!(matches!(result, Err(RuntimeError::DynamicIndexingWithReference { .. })));
+    }
```
vezenovm, line 81: Can we add a test that we do not error in Brillig

docs/docs/noir/concepts/data_types/arrays.md
```diff
@@ -77,6 +77,42 @@ However, multidimensional slices are not supported. For example, the following c
...
+    // will be used as an index here
+    let _b = array[double(x)];
+}
+
+fn double(y: u32) -> u32 {
+    y * 2
+}
+```
+
+There is another restriction with constant indices: they cannot be used on arrays with
```
vezenovm:
```suggestion
There is another restriction with dynamic indices: they cannot be used on arrays with
```

vezenovm (approved):

docs/docs/noir/concepts/data_types/arrays.md
```diff
@@ -77,6 +77,42 @@ However, multidimensional slices are not supported. For example, the following c
 let slice : [[Field]] = &[];
 ```
 
+## Dynamic Indexing
+
+Using constant indices of arrays will often be more efficient at runtime in constrained code.
+Indexing an array with non-constant indices (indices derived from the inputs to the program) is also
```
aakoshh, line 83: I was thinking of how to implement this in the fuzzer and came up with a scheme to propagate the information about which variable and expression can be originated back to inputs to the program. 

I tried this program, on the code that doesn't have this PR:
```rust
fn main() -> pub bool {
    let b = [&mut true, &mut false];
    let mut c = false;
    for k in 0..=1 {
        c |= *b[k];
    }
    c
}
```
It crashes the compiler, even though there is no indexing derived from program inputs. 

Is there another way of phrasing this restriction that better reflects the rules? In practice I see it's anything that cannot be simplified to a constant by the compiler, which inputs certainly are, but they don't cover the whole story.

aakoshh: This code above actually works with the fix in https://github.com/noir-lang/noir/pull/8877

aakoshh: Okay then I assume this phrasing is correct and if the compiler for any reason cannot figure out a numeric constant for an index _not_ dependent on input parameters, then it's a bug.

jfecher: > Okay then I assume this phrasing is correct and if the compiler for any reason cannot figure out a numeric constant for an index not dependent on input parameters, then it's a bug.

Yeah, we _should_ guarantee references are removed from the program before acir. So the only way they can still be in is if they're used in some way we inherently cannot optimize out. In theory anything taking constant inputs only which does not have side-effects should be able to be optimized out. If there are more side-effectful places references can be used in they also may not be optimized out there but we already ban these (no references in println, oracles, acir->brillig calls)

PR #8891, opened 2025-06-12
fix: preserve functions which are used in `array_set` instructions

### LINKED ISSUE #8890: Missing ArraySet case in Removing Unreachable Functions pass
https://audithub.veridise.com/app/organizations/85/projects/277/project-viewer?bpt=1&version=433&issue=14&issues=14

> The Removing Unreachable Functions optimization pass is used in Noir's pass pipeline to remove functions that are guaranteed to not be called. The pass works in three steps: first, it collects a set of "used functions" to build a call graph, then finds all functions in the call graph that are reachable through the entry point functions, and finally it removes all functions that are not reachable.
>
> The "used functions" are collected by finding all function values in Store, Call, and MakeArray instructions. However, this is missing the case for ArraySet instructions.
>
> The used_functions() function in the Removing Unreachable Functions pass
>
>```rust
> fn used_functions(func: &Function) -> BTreeSet<FunctionId> {
>    let mut used_function_ids = BTreeSet::default();
>
>    let mut find_functions = |value| {
>        if let Value::Function(function) = func.dfg[value] {
>            used_function_ids.insert(function);
>        }
>    };
>
>    for block_id in func.reachable_blocks() {
>        let block = &func.dfg[block_id];
>
>        for instruction_id in block.instructions() {
>            let instruction = &func.dfg[*instruction_id];
>
>            if matches!(
>                instruction,
>                Instruction::Store { .. }
>                    | Instruction::Call { .. }
>                    | Instruction::MakeArray { .. }
>            ) {
>                instruction.for_each_value(&mut find_functions);
>            }
>        }
>
>        block.unwrap_terminator().for_each_value(&mut find_functions);
>    }
>
>    used_function_ids
>}
>```
>
># Impact
>
>If a function is written to an array, the function will not be considered to be a "used function" by this pass. >Consequently, the function may be incorrectly removed from the program.
>
>For example, consider the following Noir program:
>
>```noir
>fn main(x: Field, y: pub Field) {
>    assert(x != y);
>    let mut f = [my_fun];
>    f[0] = my_fun2;
>    let z = (f[0])(x);
>}
>
>fn my_fun(x: Field) -> Field {
>  x + 1
>}
>
>fn my_fun2(x: Field) -> Field {
>  x + 2
>}
>```
>
>When this program is compiled with nargo compile --show-ssa, the nargo executable crashes with the following message:
>```
>After Initial SSA:
>acir(inline) fn main f0 {
>  b0(v0: Field, v1: Field):
>    v2 = eq v0, v1
>    v3 = not v2
>    constrain v2 == u1 0
>    v6 = make_array [f1] : [function; 1]
>    v7 = allocate -> &mut [function; 1]
>    store v6 at v7
>    v8 = load v7 -> [function; 1]
>    v11 = array_set v8, index u32 0, value f2
>    store v11 at v7
>    v12 = load v7 -> [function; 1]
>    v13 = array_get v12, index u32 0 -> function
>    v14 = call v13(v0) -> Field
>    return
>}
>acir(inline) fn my_fun f1 {
>  b0(v0: Field):
>    v2 = add v0, Field 1
>    return v2
>}
>acir(inline) fn my_fun2 f2 {
>  b0(v0: Field):
>    v2 = add v0, Field 2
>    return v2
>}
>
>The application panicked (crashed).
>Message:  internal error: entered unreachable code: Unmapped function with id f2
>```
>
>Because the Removing Unreachable Functions pass is the first pass that is run, the above message suggests that f2 (corresponding to the my_fun2 stored to the array) was removed by Removing Unreachable Functions.
>
>Note that the crash does not occur when the Noir program is compiled with nargo compile, so this error would have occurred silently without the extra --show-ssa debug option.

TomAFrench:
# Description

## Problem\*

Resolves #8890 

## Summary\*

This PR resolves 8890 by adding `array_set` to the set of instructions which we find functions which need to be preserved in.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/nargo_cli/tests/snapshots/execution_success/regression_8890/execute__tests__force_brillig_false_inliner_-9223372036854775808.snap
+++ b/tooling/nargo_cli/tests/snapshots/execution_success/regression_8890/execute__tests__force_brillig_false_inliner_-9223372036854775808.snap
@@ -22,24 +22,30 @@ expression: artifact
         "visibility": "public"
       }
     ],
-    "return_type": null,
+    "return_type": {
+      "abi_type": {
+        "kind": "field"
+      },
+      "visibility": "public"
+    },
     "error_types": {}
   },
   "bytecode": [
     "func 0",
-    "current witness index : _2",
+    "current witness index : _3",
     "private parameters indices : [_0]",
     "public parameters indices : [_1]",
-    "return value indices : []",
-    "BRILLIG CALL func 0: inputs: [Single(Expression { mul_terms: [], linear_combinations: [(1, Witness(0)), (-1, Witness(1))], q_c: 0 })], outputs: [Simple(Witness(2))]",
-    "EXPR [ (1, _0, _2) (-1, _1, _2) -1 ]",
+    "return value indices : [_2]",
+    "BRILLIG CALL func 0: inputs: [Single(Expression { mul_terms: [], linear_combinations: [(1, Witness(0)), (-1, Witness(1))], q_c: 0 })], outputs: [Simple(Witness(3))]",
+    "EXPR [ (1, _0, _3) (-1, _1, _3) -1 ]",
+    "EXPR [ (-1, _0) (1, _2) -2 ]",
     "unconstrained func 0",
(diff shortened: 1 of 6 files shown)


compiler/noirc_evaluator/src/ssa/opt/remove_unreachable.rs
```diff
@@ -211,4 +212,46 @@ mod tests {
...
+        let src = r#"
+        acir(inline) fn main f0 {
+          b0(v0: Field, v1: Field):
+            v2 = eq v0, v1
+            v3 = not v2
+            constrain v2 == u1 0
+            v6 = make_array [f1] : [function; 1]
+            v7 = allocate -> &mut [function; 1]
+            store v6 at v7
+            v8 = load v7 -> [function; 1]
```
vezenovm: Can reduce the size of this test by removing these trivial stores/loads, but this is non-blocking

test_programs/execution_success/regression_8890/src/main.nr
```diff
@@ -0,0 +1,17 @@
+// Regression test for issue "V-NSCA-VUL-003: Missing ArraySet case in Removing Unreachable Functions pass"
+// found in Veridise Audit.
+
+fn main(x: Field, y: pub Field) {
+    assert(x != y);
+    let mut f = [my_fun];
+    f[0] = my_fun2;
+    let _ = (f[0])(x);
```
vezenovm: should we return the result here to make sure we also are executing the expected function

Issue #8897, opened 2025-06-12
Wrong signature looked up when replacing self-recursive call


jfecher:
From https://audithub.veridise.com/app/organizations/85/projects/277/project-viewer?bpt=1&version=433&issue=17&issues=17:

Severity: Low

Status: Open

Likelihood: Likely

Impact: Bad

Type: Logic Error
Affected Files

[Audit start]
./compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
:183-194
Description

In the defunctionalize helper function of the Defunctionalize pass, every instruction will be transformed as follows:

    All direct references to functions used as arguments of the instruction will be replaced with the IDs of those functions (e.g., if my_fun is a function with ID 2, then the instruction foo(my_fun) will be replaced with foo(2)).

    All of the types of the arguments and results of the instruction will have function types substituted with the type Field.

    If the instruction is a call instruction whose callee is a block parameter or instruction result, then the instruction will be rewritten as a call to the apply function that matches the function signature of the call.

    Relevant lines in defunctionalize

```rust
    let mut instruction = func.dfg[instruction_id].clone();
    let mut replacement_instruction = None;
    if remove_first_class_functions_in_instruction(func, &mut instruction) {
        func.dfg[instruction_id] = instruction.clone();
    }
    for result in func.dfg.instruction_results(instruction_id).to_vec() {
        let typ = &func.dfg.type_of_value(result);
        if is_function_type(typ) {
            func.dfg.set_type_of_value(result, replacement_type(typ));
        }
    }
    let (target_func_id, arguments) = match &instruction {
        Instruction::Call { func: target_func_id, arguments } => {
            (*target_func_id, arguments)
        }
        _ => {
            continue;
        }
    };
    match func.dfg[target_func_id] {
        Value::Param { .. } | Value::Instruction { .. } => {
            let mut arguments = arguments.clone();
            let results = func.dfg.instruction_results(instruction_id);
            let signature = Signature {
                params: vecmap(&arguments, |param| func.dfg.type_of_value(*param)),
                returns: vecmap(results, |result| func.dfg.type_of_value(*result)),
            };
            let Some(apply_function) =
                self.get_apply_function(signature, func.runtime())
            else {
                continue;
            };
```

[...]

PR #8902, opened 2025-06-12
fix: differentiate between function and field arguments properly in defunctionalize pass

### LINKED ISSUE #8896: ApplyFunctions map created with wrong function signatures
From https://audithub.veridise.com/app/organizations/85/projects/277/project-viewer?bpt=1&version=433&issue=16&issues=14:

Given a map of pairs of (function type signature, runtime) to the set of functions matching that pair, the create_apply_functions helper function creates an "apply function" for each type signature / runtime pair, and then it returns a map where the pair is the key and the apply function is the value. The implementation loops over the entries in the map, and for each entry, it:

Computes the type signature of the apply function to generate. This is performed by using the replacement_type helper function to replace occurrences of function types in the signature with the Field type.

Either generates an apply function by calling the helper function create_apply_function, or if there is only one possible callee for this type signature, simply using that single callee as the "apply" function.

Lastly, it stores the mapping of (signature, runtime) to the apply function inside of the final map to be returned.

Relevant lines in create_apply_functions, with comments removed.
```rust
for ((mut signature, runtime), variants) in variants_map.into_iter() {
    if variants.is_empty() {
        continue;
    }
    let dispatches_to_multiple_functions = variants.len() > 1;
    for param in &mut signature.params {
        if is_function_type(param) {
            *param = replacement_type(param);
        }
    }
    for ret in &mut signature.returns {
        if is_function_type(ret) {
            *ret = replacement_type(ret);
        }
    }
    let id = if dispatches_to_multiple_functions {
        create_apply_function(ssa, signature.clone(), runtime, variants)
    } else {
        variants[0]
    };
    apply_functions
        .insert((signature, runtime), ApplyFunction { id, dispatches_to_multiple_functions });
}
```

At the last step where the (signature, runtime) is inserted into the final apply_functions map, the signature has already been modified in-place so that its function types have been replaced with Field. Thus, the apply_functions will not contain the original function signatures. Furthermore, HashMap::insert will replace existing keys, so any existing apply function entries (inserted in previous iterations of the loops) will be overridden.

For example, the signature (function, Field) -> function will be transformed to (Field, Field) -> Field. This causes any existing apply function entry for (Field, Field) -> Field to be replaced with the one for (function, Field) -> function.

Impact

If there is a dynamic dispatch involving functions whose signatures are the same after type replacement, then all of those functions will be invoked with the same apply function, even if that apply function may be the wrong one to invoke.

Consider the following test case:
```noir
    let v1 = dispatch1(|_| 1);
    let v2 = dispatch1(|_| 2);
    let f = if x == 0 { fn1 } else { fn2 };
    v1 + v2 + f(x)
}

fn dispatch1(f: fn(fn () -> Field) -> Field) -> Field {
  f(|| 0) * 3
}

fn fn1(x: Field) -> Field {
  x + 1
}

fn fn2(x: Field) -> Field {
  x * 5
}

#[test]
fn test_main() {
  assert(main(0) == 10);
  assert(main(1) == 14);
}
```

The relevant parts of the SSA IR after the Defunctionalize pass are:
```
acir(inline) fn main f0 {
  b0(v0: Field):
    v4 = call f1(Field 2) -> Field
    v6 = call f1(Field 3) -> Field
    v8 = eq v0, Field 0
    jmpif v8 then: b1, else: b2
  b1():
    jmp b3(Field 4)
  b2():
    jmp b3(Field 5)
  b3(v1: Field):
    v11 = add v4, v6
    v13 = call f8(v1, v0) -> Field  // this is the call to the apply function
    v14 = add v11, v13
    return v14
}
acir(inline) fn dispatch1 f1 {
  b0(v0: Field):
    v3 = call f8(v0, Field 6) -> Field   // this is the call to the apply function
    v5 = mul v3, Field 3
    return v5
}
acir(inline) fn lambda f2 {
  b0(v0: Field):
    return Field 1
}
acir(inline) fn lambda f3 {
  b0(v0: Field):
    return Field 2
}
acir(inline) fn apply f8 {
  b0(v0: Field, v1: Field):
    v7 = eq v0, Field 2
    jmpif v7 then: b3, else: b2
  b1(v2: Field):
    return v2
  b2():
    constrain v0 == Field 3
    v12 = call f3(v1) -> Field
    jmp b5(v12)
  b3():
    v9 = call f2(v1) -> Field
    jmp b4(v9)
  b4(v3: Field):
    jmp b6(v3)
  b5(v4: Field):
    jmp b6(v4)
  b6(v5: Field):
    jmp b1(v5)
}
```

Here, both main and dispatch1 invoke the apply function f8, which dispatches to either the lambda |_| 1 or the lambda |_| 2. This code is semantically incorrect, because the apply function call in main should be applying a function of type fn(Field) -> Field; however, the apply function f8 that is actually invoked is applying a function of type fn(fn() -> Field) -> Field.

Note that this example code "successfully" compiles to the wrong code, as can be demonstrated when the nargo test command fails to generate a proof for the execution of the test function:

```
[testcases] Running 1 test function
[testcases] Testing test_main ... FAIL

error: Failed constraint
  ┌─ src/main.nr:5:15
  │
5 │     v1 + v2 + f(x)
  │               ----
  │
  = Call stack:
    1. src/main.nr:22:10
    2. src/main.nr:5:15


[testcases] Failures:
     test_main

[testcases] 1 test failed
```

Recommendation

Make a copy of signature before it is mutated, and use this copy when inserting into the apply functions map.

TomAFrench:
…

# Description

## Problem\*

Resolves #8896 

## Summary\*

Chucking this up as a draft, expecting tests to not pass in general. I think that the main defunctionalize loop needs some unpicking to ensure that we do things in the correct order.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
@@ -635,8 +635,11 @@ fn defunctionalize_post_check(func: &Function) {
                 let function = &func.dfg[*target_func];
 
                 assert!(
-                    matches!(function, Value::Function(_)),
-                    "Call instructions should only call functions, not function values. Got '{function:?}' in instruction {instruction_id} of block {block_id} in function {} {}",
+                    matches!(
+                        function,
+                        Value::Function(_) | Value::ForeignFunction(_) | Value::Intrinsic(_)
+                    ),
+                    "Call instructions should only call functions, foreign functions, and intrinsics, not function values. Got '{function:?}' in instruction {instruction_id} of block {block_id} in function {} {}",
                     func.name(),
                     func.id()
                 );


compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
```diff
@@ -1307,4 +1332,218 @@ mod tests {
...
+          b0(v0: Field):
+            v4 = call f1(Field 1) -> Field
+            v6 = eq v0, Field 0
+            jmpif v6 then: b1, else: b2
+          b1():
+            jmp b3(Field 2)
+          b2():
+            jmp b3(Field 3)
+          b3(v1: Field):
+            v9 = call v1(v0) -> Field
```
TomAFrench: hmm

jfecher:
This is related to https://github.com/noir-lang/noir/issues/8897 - I'm thinking we may have to collect all the types of functions before the pass is run and use those for the types of functions collected during the pass.

TomAFrench:
Yeah, I've been a bit leery of these `set_type_of_value` calls as it seems ripe for an earlier instruction to affect the signature we calculate for a later instruction. Having an initial pass to find all of the relevant instructions before we start mutating things seems sensible.

As I'm off tomorrow/monday, feel free to take over so we can submit these fixes earlier. I can pick this up on Tuesday otherwise.

jfecher:
Closing in favor of https://github.com/noir-lang/noir/pull/8916

PR #8905, opened 2025-06-12
fix: Inline global arrays with functions at their call site

### LINKED ISSUE #8898: Functions stored in global variables are missing in generated SSA
From https://audithub.veridise.com/app/organizations/85/projects/277/project-viewer?bpt=1&version=433&issue=19&issues=19: 

Severity: Low

Status: Open

Likelihood: Likely

Impact: Bad

Type: Logic Error
Affected Files

[Audit start]
compiler/noirc_evaluator/src/ssa/opt/remove_unreachable.rs:49
Description

Currently, Noir permits function references to be stored in global array or slice variables. However, during initial SSA generation, functions exclusively accessed through global variable references are omitted from the generated IR. Instead, only their function IDs appear in directly in the program's global variables, with their function declarations missing.

    Example Noir program that demonstrates function references in global arrays

```noir
    global FN_ARRAY: [fn() -> Field; 2] = [f1, f2];

    fn main(x: u32, y: Field) {
      assert(FN_ARRAY[x]() != y);
    }

    fn f1() -> Field {
      1
    }

    fn f2() -> Field {
      2
    }
```

In the example above, f1() and f2() are stored in a global array FN_ARRAY, then they are called by main(). Note that in the example, f1() and f2() are not directly referenced by the main function; rather, they are referenced through the global variable FN_ARRAY.

This causes problems during the Removing Unreachable Functions pass, as demonstrated by the output of running RUST_BACKTRACE=1 nargo build (truncated to show only relevant parts). Specifically, the crash occurs when constructing the call graph.

    Log of a crash that occurs when running the Removing Unreachable Functions pass on the example code.

```
    Message:  no entry found for key
    Location: compiler/noirc_evaluator/src/ssa/ir/call_graph.rs:104

                                    ⋮ 13 frames hidden ⋮
      14: core::panicking::panic_display::hfc37a1029a2b8dcb
          at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:261
      15: core::option::expect_failed::h8f1c2a48df7a64c4
          at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/option.rs:2024
      16: core::option::Option<T>::expect::h6df44b09a994a86a
          at /Users/vfaltings/.rustup/toolchains/1.85.0-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/option.rs:933
      17: <std::collections::hash::map::HashMap<K,V,S> as core::ops::index::Index<&Q>>::index::h5ada8eb5ccbb7c7f
          at /Users/vfaltings/.rustup/toolchains/1.85.0-aarch64-apple-darwin/lib/rustlib/src/rust/library/std/src/collections/hash/map.rs:1428
      18: noirc_evaluator::ssa::ir::call_graph::CallGraph::from_deps::he75c8b7877355529
          at /Users/vfaltings/noir_playground/noir/compiler/noirc_evaluator/src/ssa/ir/call_graph.rs:104
      19: noirc_evaluator::ssa::opt::remove_unreachable::<impl noirc_evaluator::ssa::ssa_gen::program::Ssa>::remove_unreachable_functions::hd95091c16527a30b
          at /Users/vfaltings/noir_playground/noir/compiler/noirc_evaluator/src/ssa/opt/remove_unreachable.rs:54
      20: noirc_evaluator::ssa::opt::preprocess_fns::<impl noirc_evaluator::ssa::ssa_gen::program::Ssa>::preprocess_functions::hc73e4a683d1c9da8
          at /Users/vfaltings/noir_playground/noir/compiler/noirc_evaluator/src/ssa/opt/preprocess_fns.rs:74
      21: noirc_evaluator::ssa::SsaPass::new::{{closure}}::h9c319ea5b4b8d097
          at <unknown source file>:<unknown line>
      22: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call::h9423a6453e365667
          at /Users/vfaltings/.rustup/toolchains/1.85.0-aarch64-apple-darwin/lib/rustlib/src/rust/library/alloc/src/boxed.rs:2007
      23: noirc_evaluator::ssa::SsaPass::run::h831004aa2c5e8b79
          at /Users/vfaltings/noir_playground/noir/compiler/noirc_evaluator/src/ssa.rs:128
      24: noirc_evaluator::ssa::SsaBuilder::run_passes::{{closure}}::hd31a64dadd554f99
          at /Users/vfaltings/noir_playground/noir/compiler/noirc_evaluator/src/ssa.rs:683

    Snippet in remove_unreachable.rs that constructs the call graph.

    pub(crate) fn remove_unreachable_functions(mut self) -> Self {
        // Identify entry points
        let entry_points = self.functions.iter().filter_map(|(&id, func)| {
            let is_entry_point =
                id == self.main_id || func.runtime().is_acir() && func.runtime().is_entry_point();
            is_entry_point.then_some(id)
        });

        // Build call graph dependencies using this passes definition of reachability.
        let dependencies =
            self.functions.iter().map(|(&id, func)| (id, used_functions(func))).collect();
        let call_graph = CallGraph::from_deps(dependencies);
```

The root cause of this crash is that the actual declarations of f1() and f2() are missing, having been dropped at some point during the building of the SSA. The behavior here appears to be a bug, because f1() and f2() appear as function values in the global variables of the program, alongside the function array g2. When this example program is compiled with nargo build --show-ssa, the following output shows that no code is generated for f1 and f2 (with some modifications made to the printing logic, as it would crash otherwise).

    The generated SSA, before any optimization passes

```
    After Initial SSA:
    Found function in globals: f1
    Found function in globals: f2
    g2 = make_array [f1, f2] : [function; 2]

    acir(inline) fn main f0 {
      b0(v3: u32, v4: Field):
        v5 = call f1() -> Field
        v6 = eq v5, v4
        v7 = not v6
        constrain v6 == u1 0
        v10 = lt v3, u32 2
        constrain v10 == u1 1, "Index out of bounds"
        v12 = array_get g2, index v3 -> function
        v13 = call v12() -> Field
        v14 = eq v13, v4
        v15 = not v14
        constrain v14 == u1 0
        v16 = call f1() -> Field
        v17 = eq v16, v4
        v18 = not v17
        constrain v17 == u1 0
        return
    }
```

Impact

If an array of functions is stored in a global variable, and that array is referenced in a function, then the Noir compiler will crash.

The root cause of this issue (as described above) is out of scope of this audit; however, even if the issue is fixed, the Veridise security analysts believe that there may still be additional issues in passes such as the Removing Unreachable Functions pass. Specifically, the Removing Unreachable Functions pass will never consider functions like f1 and f2 in the example program to be reachable, since there is no logic to handle global variables. Additionally, if this functionality is intended to be supported, these types of programs could be interesting test cases for the Defunctionalize pass; however, the Veridise analysts cannot fully investigate the interaction due to the crash.
Recommendation

It is not clear to the Veridise auditors whether the Noir language is intended to support global variables whose values are arrays or slices of functions. We recommend clearly defining the semantics of such global variables:

    If this functionality is intended to be supported, the logic of the SSA program generation should be updated to generate code for the functions stored in the global variables. Additionally, the Removing Unreachable Functions and Defunctionalize passes should be updated to account for the possibility of direct function references being stored in global variables.

    If not, the compiler should raise an error and abort when the type of a global variable may (recursively) contain a function type.

vezenovm:
# Description

## Problem\*

Resolves #8898

## Summary\*

I just expanded what we already do for closures with globals to functions and arrays of functions. 

## Additional Context

Follow-up https://github.com/noir-lang/noir/issues/8906 to validate that we do not have function pointers in globals.

## Documentation\*

Check one:

# PR Checklist\*


jfecher (approved):
LGTM. Looks like the snap files just need to be updated for the new test

PR #8912, opened 2025-06-13
chore(fuzz): Track dynamic inputs for array indexing

aakoshh:
# Description

## Problem\*

Follow up for https://github.com/noir-lang/noir/pull/8888 in https://github.com/noir-lang/noir/pull/8728

Separate PR for visibility. 

## Summary\*

Changes `FunctionContext` to track which variable is derived from the function input parameters. These variables are avoided when we are looking generate a random index to access an array which contains reference types in ACIR.

## Additional Context

Removed `drop` because it's not used and didn't want to think about it any more.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/ast_fuzzer/src/program/tests.rs
+++ b/tooling/ast_fuzzer/src/program/tests.rs
@@ -201,10 +201,12 @@ fn test_recursion_limit_rewrite() {
     let code = format!("{}", DisplayAstAsNoir(&program));
 
     insta::assert_snapshot!(code, @r"
+    #[inline_always]
     fn main() -> () {
         let mut ctx_limit: u32 = 25;
         foo((&mut ctx_limit))
     }
+    #[inline_always]
     fn foo(ctx_limit: &mut u32) -> () {
         if ((*ctx_limit) == 0) {
             ()
@@ -213,6 +215,7 @@ fn test_recursion_limit_rewrite() {
             unsafe { bar_proxy((*ctx_limit)) }
         }
     }
+    #[inline_always]
     unconstrained fn bar(ctx_limit: &mut u32) -> () {
         if ((*ctx_limit) == 0) {
             ()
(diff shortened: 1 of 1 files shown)


tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -319,15 +330,30 @@ impl<'a> FunctionContext<'a> {
...
+        let can_use_dyn = |id: LocalId| !self.in_no_dynamic || !self.is_dynamic(&id);
+
         // Check if we have something that produces this exact type.
         if u.ratio(7, 10)? {
-            if let Some(id) = self.locals.current().choose_producer(u, typ)? {
+            let producer = if self.in_no_dynamic {
+                self.locals.current().choose_producer_filtered(u, typ, |id, _| can_use_dyn(*id))?
+            } else {
+                self.locals.current().choose_producer(u, typ)?
+            };
```
vezenovm:
```suggestion
            let producer =
                self.locals.current().choose_producer_filtered(u, typ, |id, _| can_use_dyn(*id))?;
```
It looks like this check is unnecessary. If `self.in_no_dynamic` is false then `can_use_dyn` always returns true.

aakoshh: I'm using this helper in the other branch, but yes, it is redundant here. I'll use the method directly.

vezenovm: Yeah I saw it was used later, I was just saying that checking `self.in_no_dynamic` in this specific check was redundant. I wasn't advocating to remove `can_use_dyn` entirely.

aakoshh: It's okay, I thought about it too. originally I wanted to use it in both, but the `_filtered` version requires an extra vector collection, and since this is rarely needed I thought we should use the faster version.

PR #8918, opened 2025-06-13
fix: cast values before inserting them in the databus

### LINKED ISSUE #8913: call_data and return_data generate invalid SSA
### Aim

This code:

```noir
fn main(x: call_data(0) u32, y: call_data(0) u32) -> return_data u32 {
    x + y
}
```

Generates this SSA:

```
acir(inline) fn main f0 {
  b0(v0: u32, v1: u32):
    v2 = make_array [v0, v1] : [Field; 2]
    v3 = add v0, v1
    v4 = make_array [v3] : [Field; 1]
    return v4
}
```

Note how `u32` values are put inside a `Field` array without first casting them.

### Expected Behavior

The values should be casted first.

### Bug

The values are not casted.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8913

## Summary

Also adds a couple of validations to `make_array` in the SSA interpreter to make sure the element count matches that of the element types, and that the element value types are the expected ones.

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/ssa/function_builder/data_bus.rs
```diff
@@ -142,7 +142,12 @@ impl FunctionBuilder {
         assert!(databus.databus.is_none(), "initializing finalized call data");
         let typ = self.current_function.dfg[value].get_type().into_owned();
         match typ {
-            Type::Numeric(_) => {
+            Type::Numeric(numeric_type) => {
+                let value = if matches!(numeric_type, NumericType::NativeField) {
```
vezenovm, line 146: nit: Simplification should handle this as well.

compiler/noirc_evaluator/src/ssa/function_builder/data_bus.rs
```diff
@@ -142,7 +142,12 @@ impl FunctionBuilder {
...
+                    value
+                } else {
+                    self.insert_cast(value, NumericType::NativeField)
```
vezenovm, line 149: This is going to break if we have signed integers in the data bus after https://github.com/noir-lang/noir/pull/8818.

Not sure if we already ban that or if that will cause downstream issues cc @guipublic

asterite: Oooh... good point. I checked the current `insert_safe_cast` and it's fine casting anything to Field, but I forgot about #8818.

I wonder if that implies that signed integers can't be in the databus.

compiler/noirc_evaluator/src/ssa/function_builder/data_bus.rs
asterite: Which one?

vezenovm: In `simplify_cast` if we go Field -> Field we just return the value

compiler/noirc_evaluator/src/ssa/function_builder/data_bus.rs
vezenovm: > I wonder if that implies that signed integers can't be in the databus.

It seems it will, unless we make an exception in SSA validation for signed integers...

We should check with the Aztec team whether they use signed integers anywhere in the data bus. I think they will though because I believe this would restrict what users could use in call data...

asterite: In the end it seems this cast is fine. Guillaume suggested we first cast to the unsigned type of the same size, then to Field, but the first cast is a no-op.

jfecher:
@asterite what is the status of this PR?

asterite:
It's ready to review. There's [this discussion](https://github.com/noir-lang/noir/pull/8918#discussion_r2145974171) but the conclusion is that it should be safe to directly cast signed integers to Field.

jfecher (approved):

PR #8920, opened 2025-06-16
chore: bump external pinned commits

noirwhal:
Automated update of the pinned commit of [aztec-packages](https://github.com/AztecProtocol/aztec-packages) repository against which we run benchmarks.


TomAFrench (approved):
I guess blobs got merged

PR #8924, opened 2025-06-16
feat(fuzz): Add comptime_vs_brillig_direct target

rkarabut:
# Description

Reworks the AST fuzzer comptime runner to add the comptime_vs_brillig_direct target, bypassing Nargo to call the interpreter directly. This significantly reduces overhead speeding up the fuzzing process, but at the moment runs into a few problems arising from the elaborator, preventing from running certain code which Nargo is fine with.

## Documentation

Check one:

# PR Checklist\*


tooling/ast_fuzzer/fuzz/src/targets/comptime_vs_brillig_direct.rs
```diff
@@ -0,0 +1,83 @@
...
+    let result = inputs.exec_direct()?;
+
+    compare_results_comptime(&inputs, &result)
+}
+
+#[cfg(test)]
+mod tests {
+
+    /// ```ignore
+    /// NOIR_ARBTEST_SEED=0x6819c61400001000 \
```
aakoshh:
```suggestion
    /// NOIR_AST_FUZZER_SEED=0x6819c61400001000 \
```

tooling/ast_fuzzer/fuzz/src/targets/comptime_vs_brillig_direct.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// NOIR_AST_FUZZER_SHOW_AST=1 \
+    /// cargo test -p noir_ast_fuzzer_fuzz comptime_vs_brillig_direct
+    /// ```
+    #[test]
+    fn fuzz_with_arbtest() {
+        // TODO(#8870): Allow more tests when the bug is fixed (fails in the 128th case).
+        crate::targets::tests::fuzz_with_arbtest(super::fuzz, 127);
```
aakoshh:
```suggestion
        crate::targets::tests::fuzz_with_arbtest(super::fuzz, 200 * ?);
```
I suggest finding a number that fits into e.g. 5-10 seconds locally. I set the nargo based on to 200, so if you have an idea of how much faster the direct approach this, then try that?

tooling/ast_fuzzer/src/compare/comptime.rs
```diff
@@ -64,6 +75,83 @@ fn prepare_and_compile_snippet<W: std::io::Write + 'static>(
     Ok(((program, output), warnings))
 }
 
+/// Interpret source code using the elaborator, without
+/// parsing it with nargo.
+fn interpret(src: &str) -> Expression {
```
aakoshh: If we moved this function into the frontend, could we avoid having to make so many types and properties public? It seems to me that what it does, turning source code into a monomorphized AST expression, is a nice thing to have there, it's not too outlandish, and avoids having to think about leaky abstractions and whether things should have accessor methods, etc.

tooling/ast_fuzzer/src/compare/comptime.rs
```diff
@@ -64,6 +75,83 @@ fn prepare_and_compile_snippet<W: std::io::Write + 'static>(
...
+    );
+
+    let file_manager = FileManager::new(&PathBuf::new());
+    let parsed_files = ParsedFiles::new();
+    let mut context = Context::new(file_manager, parsed_files);
+    context.def_interner.populate_dummy_operator_traits();
+
+    let krate = context.crate_graph.add_crate_root(FileId::dummy());
+
+    let (module, errors) = parse_program(src, file);
```
aakoshh: > without parsing it with nargo.

I'm trying to understand where the speedup is coming from. 

The `prepare_snippet` function we have in `comptime_vs_brillig` seems to have to go through very similar steps; I see it calls `parse_program` through `parse_file` in a parallel context, so there can be some overhead I suppose.

Then I see the existing stuff uses `compile_main`, which calls `collect_defs` through `check_crate`, while you call it directly here. But `compile_main` returns fully fledged bytecode, while this version goes directly to elaborate `main` and call it through the interpreter.

I didn't know you can call `main` directly in the interpreter, I thought it will only have a `comptime` body. 

So the speedup is not from "not parsing with nargo", but rather not compiling all the way with nargo?

tooling/ast_fuzzer/src/compare/comptime.rs
```diff
@@ -64,6 +75,83 @@ fn prepare_and_compile_snippet<W: std::io::Write + 'static>(
...
+        match interpreter.call_function(main, Vec::new(), Default::default(), Location::dummy()) {
+            Err(e) => panic!("interpreter error: {:?}", e),
+            Ok(value) => match value.into_hir_expression(elaborator.interner, Location::dummy()) {
+                Err(e) => panic!("could not convert interpreter result into HIR: {:?}", e),
+                Ok(expr_id) => expr_id,
+            },
+        };
+
+    let mut monomorphizer = Monomorphizer::new(elaborator.interner, DebugTypeTracker::default());
+    monomorphizer.expr(expr_id).expect("monomorphization error while converting interpreter execution result, should not be possible")
```
aakoshh: I don't fully understand whether making the things you made public is the only way to do this. 

For example the [interpret_cmd](https://github.com/noir-lang/noir/blob/d068d38f359e9ae656ba7d02c4404cf6f583bbbc/tooling/nargo_cli/src/cli/interpret_cmd.rs#L170-L207) uses the already public `monomorphize` function to get a monomorphized AST. If we call that on a context which has been prepared with a code snippet, wouldn't it return the same thing as you get here, in that the `body` of the `main` function would be the `Expression` you're looking for? The `comptime` part will have been interpreted by the elaborator, no?

tooling/ast_fuzzer/src/compare/comptime.rs
```diff
@@ -72,10 +160,52 @@ pub struct CompareComptime {
...
+
+        let res1 = do_exec(&comptime_ssa.artifact.program);
+        let res2 = do_exec(&self.ssa.artifact.program);
+
+        CompareCompiledResult::new(
+            &self.abi,
+            &Default::default(),
+            &self.ssa.artifact.error_types,
+            (res1, "".into()),
+            (res2, "".into()),
```
aakoshh, line 127: It would be nice to be able to capture the output, though. 

I see that you moved the compilation of the source into the `arb`. I suppose this is ambiguous, because the AST _is_ compiled into SSA there, yet the existing `comptime_vs_brillig` compiles the Noir snippet into SSA in `exec`. 

I think I put the SSA generation into `arb` out of necessity: I wanted to do everything that requires `Unstructured` in `arb`, including generating the `InputMap`, which required an SSA, so it kinda made sense to hold on to it until execution.

But, unlike a normal program generated by the fuzzer, comptime code produces its print output during compilation, which is a side effect we want to capture, and the return value of `exec` contains those, while the the output of `arb` does not, so I think it makes sense for this side-effecting compilation to take place in `exec`, even if it produces SSA. 

Then, we can capture both output, as it is done by the current `comptime_vs_brillig`.

tooling/ast_fuzzer/src/program/mod.rs
```diff
@@ -70,6 +70,32 @@ pub fn arb_program_comptime(u: &mut Unstructured, config: Config) -> arbitrary::
     Ok(program)
 }
 
+/// Build a program with the single `main` function returning
+/// the result of a given expression (used for conversion of the
+/// interpreter execution results for comparison)
+pub fn program_comptime_wrap_expression(
+    u: &mut Unstructured,
```
aakoshh: It looks like this `Unstructured` isn't actually used. I'd remove it from the parameters and construct an empty one in the the function instead.

tooling/ast_fuzzer/src/program/mod.rs
```diff
@@ -70,6 +70,32 @@ pub fn arb_program_comptime(u: &mut Unstructured, config: Config) -> arbitrary::
     Ok(program)
 }
 
+/// Build a program with the single `main` function returning
+/// the result of a given expression (used for conversion of the
+/// interpreter execution results for comparison)
+pub fn program_comptime_wrap_expression(
```
aakoshh: Does it have to say `comptime` at all?

tooling/ast_fuzzer/fuzz/src/targets/comptime_vs_brillig_direct.rs
```diff
@@ -0,0 +1,83 @@
...
+        comptime_friendly: true,
+        // Force brillig, to generate loops that the interpreter can do but ACIR cannot.
+        force_brillig: true,
+        // Avoid overflows, divisions by zero and constraints for now, as we currently
+        // don't catch errors issued by the elaborator
+        avoid_overflow: true,
+        avoid_err_by_zero: true,
+        avoid_constrain: true,
+        // At the moment prints aren't recognized by elaborator
+        avoid_print: true,
```
aakoshh, line 35: How come? It does work with the existing `comptime_vs_brillig` setup?

rkarabut: No idea, I didn't dive in deep enough. But print is getting recognized as an undefined variable for some reason. I left that for later

tooling/ast_fuzzer/src/compare/comptime.rs
rkarabut: Yeah, I'll update this as soon as we get `print` actually working I suppose

tooling/ast_fuzzer/src/compare/comptime.rs
rkarabut: Yes, let me change the phrasing there. We don't have to compile anything at all now (besides the final literal conversion, for comparison)

tooling/ast_fuzzer/src/compare/comptime.rs
rkarabut: Probably a good idea, yes

tooling/ast_fuzzer/src/program/mod.rs
rkarabut: Ah you mean call it just `wrap_expression`? Yeah I suppose it doesn't, really

tooling/ast_fuzzer/src/compare/comptime.rs
rkarabut: Moved

tooling/ast_fuzzer/src/compare/comptime.rs
rkarabut: Should be irrelevant now, moved `interpret()` into the elaborator

tooling/ast_fuzzer/fuzz/src/targets/comptime_vs_brillig_direct.rs
rkarabut: Added https://github.com/noir-lang/noir/issues/8973

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -2409,3 +2409,83 @@ impl Visitor for RemoveGenericsAppearingInTypeVisitor<'_> {
...
+    let parsed_files = ParsedFiles::new();
+    let mut context = Context::new(file_manager, parsed_files);
+    context.def_interner.populate_dummy_operator_traits();
+
+    let krate = context.crate_graph.add_crate_root(FileId::dummy());
+
+    let (module, errors) = parse_program(src, file);
+    // Skip warnings
+    let errors: Vec<_> = errors.iter().filter(|e| !e.is_warning()).collect();
+    assert_eq!(errors.len(), 0);
```
aakoshh: Can we return a `Result` instead of panicking?

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -2409,3 +2409,83 @@ impl Visitor for RemoveGenericsAppearingInTypeVisitor<'_> {
...
+        collector.items,
+        ElaboratorOptions::test_default(),
+    );
+
+    // Skip compiler warnings
+    let errors: Vec<_> = elaborator.errors.iter().filter(|&e| e.is_error()).cloned().collect();
+    if !errors.is_empty() {
+        log::debug!("elaborator errors: {:?}", errors);
+    }
+    assert_eq!(errors.len(), 0);
```
aakoshh: Return errors instead of logging and panicking? Now that it's a public function, it would be nice to behave like the others.

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -2409,3 +2409,83 @@ impl Visitor for RemoveGenericsAppearingInTypeVisitor<'_> {
...
+    let mut interpreter = elaborator.setup_interpreter();
+
+    // The most straightforward way to convert the interpreter result into
+    // an acceptable monomorphized AST expression seems to be converting it
+    // into HIR first and then processing it with the monomorphizer
+    let expr_id =
+        match interpreter.call_function(main, Vec::new(), Default::default(), Location::dummy()) {
+            Err(e) => panic!("interpreter error: {:?}", e),
+            Ok(value) => match value.into_hir_expression(elaborator.interner, Location::dummy()) {
+                Err(e) => panic!("could not convert interpreter result into HIR: {:?}", e),
```
aakoshh: Return errors?

tooling/ast_fuzzer/Cargo.toml
```diff
@@ -32,7 +32,7 @@ nargo.workspace = true
 noirc_abi.workspace = true
 noirc_driver.workspace = true
 noirc_evaluator.workspace = true
-noirc_frontend.workspace = true
+noirc_frontend = { workspace = true, features = ["test_utils"] }
```
aakoshh, line 35: What do we use from test utils?

tooling/ast_fuzzer/src/compare/comptime.rs
aakoshh: It would mean there would be no need for the `pub comptime_ssa: Option<CompareArtifact>,` field and the panicky unwrapping in exec, which is why I'm saying it's one less exceptional case to think of.

tooling/ast_fuzzer/Cargo.toml
rkarabut: Right, shouldn't be needed anymore (if we don't move the new `interpret()` under test-utils though :thinking: )

tooling/ast_fuzzer/src/program/mod.rs
```diff
@@ -312,15 +336,15 @@ impl Context {
...
     }
 
     /// Generate function with a specified body generator.
     fn gen_function_with_body(
         &mut self,
-        u: &mut Unstructured,
+        u: Option<&mut Unstructured>,
         id: FuncId,
-        f: impl Fn(&mut Unstructured, FunctionContext) -> arbitrary::Result<Expression>,
+        f: impl Fn(Option<&mut Unstructured>, FunctionContext) -> arbitrary::Result<Expression>,
```
aakoshh: Oh, I didn't think an optional unstructured, just something like passing `&mut Unstructured::new(&[])` in `program_wrap_expression`.

tooling/ast_fuzzer/Cargo.toml
aakoshh: I had that thought too

tooling/ast_fuzzer/src/program/mod.rs
rkarabut: Ah okay, done

compiler/noirc_frontend/src/elaborator/mod.rs
rkarabut: Done

tooling/ast_fuzzer/Cargo.toml
rkarabut: Moved

tooling/ast_fuzzer/src/compare/comptime.rs
rkarabut: Right, moved everything into exec_direct, removed arb_direct wholesale

tooling/ast_fuzzer/src/compare/comptime.rs
```diff
@@ -75,7 +77,58 @@ pub struct CompareComptime {
...
+    pub fn exec_direct(
+        &self,
+        f_comptime: impl FnOnce(Program) -> arbitrary::Result<(SsaProgramArtifact, CompareOptions)>,
+    ) -> eyre::Result<CompareCompiledResult> {
+        // log source code before interpreting
+        log::debug!("comptime src:\n{}", self.source);
+
+        let comptime_res = match interpret(&format!("comptime {}", self.source)) {
+            Err(e) => {
+                panic!("elaborator error while interpreting generated comptime code: {:?}", e)
```
aakoshh, line 92: Would this panic if there was an overflow in the code? In the other target I tried to treat compilation errors as potential equivalents of runtime errors.

aakoshh: Ah, I see the comment in the config. It's something to think about in the future.

aakoshh (approved):
Tidy darts!

PR #8934, opened 2025-06-16
fix: Strange use of predicates in euclidian division

Rumata888:
# Description

## Problem\*

Predicate and one were mixed in 1 place and predicate was reused instead of one in another (technically both were still correct, due to a coincidence).

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/acir/acir_context/mod.rs
```diff
@@ -902,7 +902,7 @@ impl<F: AcirField, B: BlackBoxFunctionSolver<F>> AcirContext<F, B> {
         // `rhs` has the correct bit-size because either it is enforced by the overflow checks
         // or `rhs` is zero when the overflow checks are disabled.
         // Indeed, in that case, rhs is replaced with 'predicate * rhs'
-        self.bound_constraint_with_offset(remainder_var, rhs, predicate, max_rhs_bits, one)?;
+        self.bound_constraint_with_offset(remainder_var, rhs, one, max_rhs_bits, predicate)?;
```
aakoshh: I see line 882 and 897 also pass `one` as `predicate`, are those correct?

aakoshh: According to the comments it is 👍

Rumata888:
> Makes sense based on the docs of `self.bound_constraint_with_offset`.
> 
> How did you spot these?

I am going through the whole SSA->ACIR conversion workflow to see if we can generate some tooling to check it. So I saw it. A bit worrying that tests are failing now

aakoshh:
One source of test failure is the `insta` snapshots complaining about changes in the bytecode. I'll check in those changes and see if there are others where the return value changes.

aakoshh:
Okay there is definitely something wrong, about a dozen tests [no longer build](https://github.com/noir-lang/noir/actions/runs/15780474405/job/44484840977?pr=8934), for example:

```console
$ cd test_programs/execution_success/bit_and
$ cargo run -q -p nargo_cli -- execute --force-brillig
[bit_and] Circuit witness successfully solved
[bit_and] Witness saved to target/bit_and.gz

$ cargo run -q -p nargo_cli -- execute 
error: Failed constraint
  ┌─ src/main.nr:4:19
  │
4 │     let x_as_u8 = x as u8;
  │                   -------
  │
  = Call stack:
    1. src/main.nr:4:19

```

compiler/noirc_evaluator/src/acir/acir_context/mod.rs
```diff
@@ -942,7 +942,7 @@ impl<F: AcirField, B: BlackBoxFunctionSolver<F>> AcirContext<F, B> {
                 self.bound_constraint_with_offset(
                     remainder_var,
                     max_r_var,
-                    predicate,
+                    one,
```
aakoshh, line 973: I think this needs to be `zero`. I'll check and push if that works.

aakoshh:
Changed the condition on `r` to be `r <= max_r`, which I think is generally consistent with how I think about _max_. This way the tests pass locally 🤞 

I think this means when `r == max_r` then `q*b+r == F::modulus()`. Is that okay?

Rumata888:
> Changed the condition on `r` to be `r <= max_r`, which I think is generally consistent with how I think about _max_. This way the tests pass locally 🤞
> 
> I think this means when `r == max_r` then `q*b+r == F::modulus()`. Is that okay?

I don't think so. The maximum value shouldn't be a modulus, then you can cheat. That's why it's all so surprising.

aakoshh:
In the failing `bit_and` test above, we have `x` being `0x00` in the input, and the SSA is:
```rust
    v4 = truncate v0 to 8 bits, max_bit_size: 254
    v5 = cast v4 as u8
```
with this the variables are:
```
bit_size = 254
max_rhs_bits = 9
246 + 9 >= 254 - 1
modulus = 21888242871839275222246405745257275088548364400416034343698204186575808495617
rhs = 256
q0 = modulus / rhs = 85500948718122168836900022442411230814642048439125134155071110103811751936
max_r = modulus - q0 * rhs = 1
```
Indeed if `q` is `q0` then an `r` of 1 would result in `q*256+r == modulus`.

What I see in the ACIR is that we have added a `BLACKBOX::RANGE [(_10, 0)] []` at the bottom, a range constraint for a `u0` numeric type. With the `zero` offset this is instead `BLACKBOX::RANGE [(_10, 1)] []`, for a `u1` type. The `u0` doesn't seem to make sense; I'm trying to find what does this.

aakoshh:
Although, it does make sense: `max_r` is 1, and because of the `offset` of 1 it becomes `r <= 0` instead of `r <= 1`, except it applies the constraint with the `(q==q0)` predicate. How can that be non-zero?

aakoshh:
The whole ACIR I'm looking at is like this:
```
BRILLIG CALL func 0: inputs: [Single(Expression { mul_terms: [], linear_combinations: [(1, Witness(0))], q_c: 0 }), Single(Expression { mul_terms: [], linear_combinations: [], q_c: 256 })], outputs: [Simple(Witness(5)), Simple(Witness(6))]
BLACKBOX::RANGE [(_5, 246)] []
BLACKBOX::RANGE [(_6, 8)] []
EXPR [ (1, _0) (-256, _5) (-1, _6) 0 ]
EXPR [ (-1, _5) (-1, _7) 85500948718122168836900022442411230814642048439125134155071110103811751936 ]
BLACKBOX::RANGE [(_7, 246)] []
BRILLIG CALL func 1: inputs: [Single(Expression { mul_terms: [], linear_combinations: [(-1, Witness(5))], q_c: 85500948718122168836900022442411230814642048439125134155071110103811751936 })], outputs: [Simple(Witness(8))]
EXPR [ (-1, _5, _8) (85500948718122168836900022442411230814642048439125134155071110103811751936, _8) (1, _9) -1 ]
EXPR [ (-1, _5, _9) (85500948718122168836900022442411230814642048439125134155071110103811751936, _9) 0 ]
EXPR [ (1, _6, _9) (-1, _10) 0 ]
BLACKBOX::RANGE [(_10, 0)] []
EXPR [ (1, _4) (-1, _6) 0 ]
```

aakoshh:
I checked what happens in `solve_range_opcode` when we are processing `BLACKBOX::RANGE [(_10, 0)] []`. 

`w10` is `r * size_predicate`, but it failed when I changed to be just `r` or `size_predicate`. Checking the actual value in the VM, it turns out that the value is correctly `0`, and we want it to be `0` bits, but [here](https://github.com/noir-lang/noir/blob/d7030e8c4e05dcba0eb30adac7f0f18458b306f0/acvm-repo/acvm/src/pwg/blackbox/range.rs#L16) it says that `num_bits()` for `w_value = 0` is actually 1 bit.

aakoshh:
And the reason it didn't fail before was because `predicate` wasn't a constant `one`, so it didn't [this code](https://github.com/noir-lang/noir/blob/d7030e8c4e05dcba0eb30adac7f0f18458b306f0/compiler/noirc_evaluator/src/acir/acir_context/mod.rs#L1014), which turned the RHS from 1 to 0.

aakoshh:
There is an actual [test](https://github.com/noir-lang/noir/blob/d7030e8c4e05dcba0eb30adac7f0f18458b306f0/acvm-repo/acir_field/src/field_element.rs#L411) that says that `Field::zero().num_bits() == 1`. This contradicts [this calculation](https://github.com/noir-lang/noir/blob/d7030e8c4e05dcba0eb30adac7f0f18458b306f0/compiler/noirc_evaluator/src/acir/acir_context/mod.rs#L994-L995) apparently.

aakoshh:
@TomAFrench I'll change the definition of `bit_size_u128` to return 1 bit for the value of 0, but I suppose we could also change `solve_range_opcode` to validate `bit_size == 0` by requiring the `w_value == 0`. Is it conceivable that another `F` implementation would use 0 bits for 0 value?

aakoshh:
Hm, actually that sounds like the better solution, because `FieldElement` requires 1 bit for the value 1 as well, so a range constraint of 1 bit admits both 0 and 1, whereas we wanted 0.

aakoshh:
Okay, it looks like that special handling would have to be added in a few more places, or we go the other way and allow 1 bit. Or return 0 from FieldElemenent::num_bits() for zero.

acvm-repo/acir_field/src/field_element.rs
```diff
@@ -368,8 +368,9 @@ impl BitCounter {
     fn bits(&self) -> u32 {
         // If we don't have a non-zero byte then the field element is zero,
         // which we consider to require a single bit to represent.
+        // OTOH some operations produce range constraints with 0 bits.
```
TomAFrench:
```suggestion
        // If we don't have a non-zero byte then the field element is zero,
        // which we consider to require a zero bits to represent.
```

TomAFrench (approved):
LGTM

compiler/noirc_evaluator/src/acir/acir_context/mod.rs
guipublic: Using the predicate as an offset is a small optimisation, because if the predicate is true, then the offset is one and this assert that 'r<rhs', without using a predicate (because 'one' is given for the predicate argument).
If the predicate is false, then this will assert 'r<=rhs', for which 'r=0' is a valid assignment.
So I think we should keep the previous code.

guipublic (changes requested):
The first change should be reverted.

compiler/noirc_evaluator/src/acir/acir_context/mod.rs
aakoshh: > If the predicate is false, then this will assert 'r<=rhs', for which 'r=0' is a valid assignment.

Say in the case of the `bit_and` test, where the `rhs` is 256; if the predicate is false, this will assert that `r <= 256`. But `r = 256` would not be a valid assignment, would it?

guipublic: 256 would be a valid assignment in that case, I would say.

aakoshh: I thought that what we are looking for is something like `let r = x % 256;`, where `r` would be between 0 and 255, inclusive, but never 256.

guipublic: This is not true when the predicate is false. For instance `r=x %0` under a false predicate is not between 0 and -1, because this is not possible.

aakoshh: Thanks for the explanation in the huddle, I updated the comments and reverted the code, PTAL

guipublic (approved):
LGTM

Issue #8923, opened 2025-06-16
Full != Minimal: return value changes in SSA passes


aakoshh:
### Aim

Discovered by the `orig_vs_min` fuzzer. 

The following program returns a different value with `--minimal-ssa`:

```rust
unconstrained fn main(mut b: [u16; 1]) -> pub [u16; 1] {
    if false {
        [1]
    } else {
        let mut d = b;
        if (true) {
            d[0] = 2;
            b
        } else {
            [3]
        }
    }
}
```

input:
```toml
b = ["0x0000000000000000000000000000000000000000000000000000000000000000"]
```

### Expected Behavior

The return value should be `[0]`, at least I thought `let mut d = b;` would result in a copy. 

### Bug

```console
❯ cargo run -q -p nargo_cli -- execute --minimal-ssa --force 
[fuzz_testing] Circuit witness successfully solved
[fuzz_testing] Witness saved to target/fuzz_testing.gz
[fuzz_testing] Circuit output: Vec([Field(2)])

❯ cargo run -q -p nargo_cli -- execute --force
[fuzz_testing] Circuit witness successfully solved
[fuzz_testing] Witness saved to target/fuzz_testing.gz
[fuzz_testing] Circuit output: Vec([Field(0)])
```

The difference goes away with either of following changes:
* Remove the `if false`
* Remove the `if true`
None of those should have an effect, but they do. 

Using the `interpret` command, we can see that the initial SSA is already incorrect, but then it changes its mind:

```console
❯ cargo run -q -p nargo_cli -- interpret
--- Interpreter result after Initial SSA:
Ok([ArrayOrSlice(ArrayValue { elements: Shared(RefCell { value: [Numeric(U16(2))] }), rc: Shared(RefCell { value: 1 }), element_types: [Numeric(Unsigned { bit_size: 16 })], is_slice: false })])
...
--- Interpreter result after Check u128 mul overflow (step 29):
Ok([ArrayOrSlice(ArrayValue { elements: Shared(RefCell { value: [Numeric(U16(2))] }), rc: Shared(RefCell { value: 1 }), element_types: [Numeric(Unsigned { bit_size: 16 })], is_slice: false })])
---
--- Interpreter result after Dead Instruction Elimination (step 30):
[...]

PR #8935, opened 2025-06-16
fix: Fix if/match tracking in last uses pass

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/8923

## Summary\*

The `BranchesPath` in some situations could get confused as being the same as a prior if. E.g. in

```
if true {
    if true {
        foo(b);
    }
}

if true {
    if false {
    } else {
        foo(b);
    }
}
```

We tried to move `b` both times previously because we tracked `Branch0 { Branch0 { use } }` and left it intact until the next use which looks compatible, so they were merged into `Branch0 { Branch0 { use }, Branch1 { use } }`. To fix this I tagged each if and match with its own unique id.

## Additional Context

~~This is a draft currently because I need to add a test~~

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/ownership/last_uses.rs
```diff
@@ -145,29 +170,30 @@ impl Context {
...
     fn unbranch(&mut self) {
         let path =
             self.current_loop_and_branch.last_mut().expect("We should always have at least 1 path");
-        let rest = std::mem::replace(path, BranchesPath::Here);
-
-        match rest {
-            BranchesPath::Here => panic!("unbranch called without any branches"),
-            BranchesPath::IfOrMatch { branch_index: _, rest } => *path = *rest,
-        }
+        path.pop();
```
asterite: Is it worth adding an `.expect(...)` call here?

PR #8939, opened 2025-06-17
chore: disable noncritical workflow steps which fail on external prs

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

These parts of CI fail on external PRs as we restrict access to the relevant access token permissions. This results in a decent amount of confusion for external contributors so it's worth adding a couple of lines to just skip these steps in these situations rather than just letting them fail.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
I've rolled in some bumps to the memory limits which I missed before.

PR #8942, opened 2025-06-17
chore: address various clippy issues

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

I think something has changed in the stable version of cargo (which we use in rust-analyzer) which is resulting in me getting a large number of clippy warnings. This PR applies fixes for these or silences those which we don't want.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,6 +72,10 @@ trivial_numeric_casts = "warn"
 unused_import_braces = "warn"
 unused_qualifications = "warn"
 
+[workspace.lints.clippy]
+semicolon_if_nothing_returned = "warn"
+result_large_err = "allow"
+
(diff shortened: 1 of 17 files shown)


Cargo.toml
```diff
@@ -71,6 +71,11 @@ trivial_casts = "warn"
 trivial_numeric_casts = "warn"
 unused_import_braces = "warn"
 unused_qualifications = "warn"
+unreachable_pub = "warn"
+
+[workspace.lints.clippy]
+semicolon_if_nothing_returned = "warn"
+result_large_err = "allow"
```
asterite, line 80: Do these mean we don't need to put these on some files? If so, we could remove them from the files they are in.

PR #8943, opened 2025-06-17
feat: return unsolvable opcode from `CircuitSimulator`

### LINKED ISSUE #8774: Compiler crash: Expected all load instructions to be removed before acir_gen
### Aim

Discovered while trying to further reduce https://github.com/noir-lang/noir/issues/8761

This code:

```noir
fn main() -> pub bool {
    let d: [&mut bool; 0] = [];
    *d[0]
}
```

crashes with this:

```
Message:  internal error: entered unreachable code: Expected all load instructions to be removed before acir_gen
Location: compiler/noirc_evaluator/src/acir/mod.rs:637
```

We can see the final SSA is this one:

```
  b0():
    v0 = make_array [] : [&mut u1; 0]
    v4 = array_get v0, index u32 0 -> &mut u1
    v5 = load v4 -> u1
    return v5
}
```

### Expected Behavior

The code should compile, and probably on execution give an index out of bounds error.

### Bug

The compiler crashes.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #8774 

## Summary\*
The compiler crash is transformed into an error during execution by using non solvable witness to represent unresolved loads.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/acir/acir_context/mod.rs
```diff
@@ -1567,7 +1571,9 @@ impl<F: AcirField, B: BlackBoxFunctionSolver<F>> AcirContext<F, B> {
                 for value in dynamic_array_values {
                     self.initialize_array_inner(witnesses, value)?;
                 }
-            }
+            } // AcirValue::UninitializedVar(_) => {
+              //     // Uninitialized variables are not initialized, so we do nothing.
+              // }
```
TomAFrench: leftover commented code.

guipublic: Done

asterite:
I wonder, the SSA in the issue is this one:

```noir
  b0():
    v0 = make_array [] : [&mut u1; 0]
    v4 = array_get v0, index u32 0 -> &mut u1
    v5 = load v4 -> u1
    return v5
}
```

I guess it crashes because there's an array get of an empty array, then loading that value...

Could we instead have an SSA pass that recursively removes values that can't happen? I'm thinking that `v4` can't succeed regardless of the index because the array is empty, so we'd remove that instruction. Now that `v4` was removed, `v5` should be removed too because it uses that, etc. The tricky part is removing the `return`, or replacing it with a zeroed value I guess.

asterite:
Actually, this might be simpler. In this SSA:

```noir
  b0():
    v0 = make_array [] : [&mut u1; 0]
    v4 = array_get v0, index u32 0 -> &mut u1
    v5 = load v4 -> u1
    return v5
}
```

there's a `constrain` that will always fail. Then we can remove any subsequent instructions in that block and replace the terminator values with zeroed values (the "zeroed values" logic was introduced in a recent PR).

asterite:
I'll try the above in a separate PR, just because now I'm curious about it 😊

acvm-repo/acvm/src/compiler/simulator.rs
```diff
@@ -23,15 +23,15 @@ impl CircuitSimulator {
     /// Simulate a symbolic solve for a circuit by keeping track of the witnesses that can be solved.
     /// Returns false if the circuit cannot be solved
```
aakoshh:
```suggestion
    /// Returns the index of the opcode that cannot be solved, if any.
```

acvm-repo/acvm/src/compiler/optimizers/merge_expressions.rs
```diff
@@ -293,15 +293,15 @@ mod tests {
     use std::collections::BTreeSet;
 
     fn check_circuit(circuit: Circuit<FieldElement>) -> Circuit<FieldElement> {
-        assert!(CircuitSimulator::default().check_circuit(&circuit));
+        assert!(CircuitSimulator::default().check_circuit(&circuit).is_none());
```
aakoshh, line 296: Golang has a pattern of prefixing panicky methods with `must_`. I wonder if a `must_check_circuit` could be added so that the assertion message can be made a bit more meaningful than `None` was found.

compiler/noirc_evaluator/src/acir/acir_context/mod.rs
```diff
@@ -83,7 +83,11 @@ impl<F: AcirField, B: BlackBoxFunctionSolver<F>> AcirContext<F, B> {
             big_int_ctx: Default::default(),
             expression_width: Default::default(),
             warnings: Default::default(),
-        }
+        };
+
+        let uninitialized_var = context.add_data(AcirVarData::Witness(Witness(u32::MAX)));
+        assert_eq!(uninitialized_var, AcirVar::uninitialized());
```
aakoshh: Can you add some comments about why this is here?

compiler/noirc_evaluator/src/acir/types.rs
```diff
@@ -163,6 +163,20 @@ impl AcirValue {
...
+
+    /// Generates an uninitialized `AcirValue` of the given type.
+    pub(super) fn uninitialized(typ: AcirType) -> Self {
+        match typ {
+            AcirType::NumericType(_) => AcirValue::Var(AcirVar::uninitialized(), typ),
+            AcirType::Array(acir_types, len) => {
+                let values: im::Vector<AcirValue> =
+                    acir_types.iter().map(|t| AcirValue::uninitialized(t.clone())).collect();
+                let values =
+                    if values.len() == 1 { values[0].clone() } else { AcirValue::Array(values) };
```
aakoshh:
```suggestion
                    if values.len() == 1 { values.into_iter().next().unwrap() } else { AcirValue::Array(values) };
```

tooling/nargo/src/ops/check.rs
```diff
@@ -7,11 +7,23 @@ use noirc_errors::CustomDiagnostic;
...
+                let call_stack =
+                    compiled_program.debug[i].location_tree.get_call_stack(*call_stack);
+                CustomDiagnostic::from_message(
+                    &format!("Circuit \"{}\" is not solvable", compiled_program.names[i]),
+                    call_stack[0].file,
+                )
+                .with_call_stack(call_stack)
+            } else {
+                CustomDiagnostic::from_message(
+                    &format!("Circuit \"{}\" is not solvable", compiled_program.names[i]),
```
aakoshh: Maybe still worth mentioning the opcode?

guipublic: Done

compiler/noirc_evaluator/src/acir/types.rs
guipublic: Done

compiler/noirc_evaluator/src/acir/acir_context/mod.rs
guipublic: Done

acvm-repo/acvm/src/compiler/optimizers/merge_expressions.rs
guipublic: Done

acvm-repo/acvm/src/compiler/simulator.rs
guipublic: Done

TomAFrench:
This is superceded by #9008 right @asterite?

asterite:
@TomAFrench Yes. I didn't know if this PR was also wanted, though I guess now unresolved loads can't (shouldn't) happen anymore.

guipublic:
This PR is different than #9008 and removes a compiler crash. I don't think there is a 100% guarantee that loads will always be removed.

aakoshh:
About load removal: I added a panic if they are still present after DIE in https://github.com/noir-lang/noir/pull/9018

I guess this is something that could be turned into a warning if we find a case we cannot eliminate.

jfecher (commented):
Let's revert the unreachable on Instruction::Load but we can keep the callstack changes

guipublic:
I reverted the changes to unresolved loads.

PR #8944, opened 2025-06-17
fix: when macro parse error happens, discard warnings; also preserve unquoted token locations

asterite:
# Description

## Problem

No issue, just something confusing mentioned over Slack that could be improved.

## Summary

Two things here:
1. When a macro parse error happened we'd show the just first error. The issue was that in the list of "errors" warnings could exist too. So here we discard warnings and truly show the first error. (Maybe ideally we'd show all errors and warnings here but I'm not sure how we'd show that as a result of a macro expansion, so I did the minimum thing to improve the situation without introducing a large change).
2. The location of the `$expr` token ended up being the location of the `quote` expression that held it, making errors point to that entire quote (as can be seen in the second commit). This was likely done before everything started holding a Location instead of a Span. The second improvement here is to preserve those token locations. The third commit shows how the error improves.

In the actual code with the confusing error message, this is what we initially saw in VSCode:

image

This is the warning caused by not being able to parse the macro.

Once that was fixed (first commit), we got this error:

image

It actually continues above and below that: it's the entire `quote { ... }` that has the error.

With the third commit we get this error:

image

with the error being "Expected value, found built-in type `(resolved type)`". So now it's much clearer where's the error.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
@@ -554,7 +554,7 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
             HirExpression::Match(match_) => todo!("Evaluate match in comptime code"),
             HirExpression::Tuple(tuple) => self.evaluate_tuple(tuple),
             HirExpression::Lambda(lambda) => self.evaluate_lambda(lambda, id),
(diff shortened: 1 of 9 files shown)


TomAFrench:
Looks like the snapshot is incorrect.

PR #8946, opened 2025-06-17
fix(ssa): Swap Brillig index shift and DIE in minimal pipeline

### LINKED ISSUE #8941: Compiler crash: `Global value not found in cache`
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8938

The `execution_success/integer_array_indexing` test crashes with the `--minimal-ssa` option.

### Expected Behavior

It should execute and return the same result as with `--force-ssa`. 

### Bug

```console
❯ cargo run -q -p nargo_cli -- execute --minimal-ssa
Message:  ICE: Global value not found in cache v0
Location: compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs:1830
```

The globals are populated by the DIE step, so I thought that maybe the minimal SSA pipeline is missing a step. 
I modified `Ssa::dead_instruction_elimination_inner` to print the `result.used_globals`, and noticed that the fact that `--force-brillig` works might be a coincidence. 

In the first DIE step, no globals are used:

```rust
After Check u128 mul overflow (1) (step 29):
g0 = u32 3

  b0(v1: [Field; 3], v2: u32):
    v4 = array_get v1, index u32 2 -> Field
    v5 = allocate -> &mut Field
    v7 = array_get v1, index u32 0 -> Field
    v8 = add v4, v7
    v10 = array_get v1, index u32 1 -> Field
    v11 = add v8, v10
    v12 = cast v2 as Field
    v13 = add v11, v12
    return v13
}

USED GLOBALS:
	f0: []
After Dead Instruction Elimination (1) (step 30):
g0 = u32 3

  b0(v1: [Field; 3], v2: u32):
    v4 = array_get v1, index u32 2 -> Field
    v6 = array_get v1, index u32 0 -> Field
    v7 = add v4, v6
    v9 = array_get v1, index u32 1 -> Field
    v10 = add v7, v9
    v11 = cast v2 as Field
    v12 = add v10, v11
    return v12
}
```

In the second DIE step, it reports the use of `v0`:
```rust
After Brillig Array Get and Set Optimizations (1) (step 37):
g0 = u32 3

  b0(v1: [Field; 3], v2: u32):
    v3 = array_get v1, index u32 3 minus 1 -> Field
    v5 = array_get v1, index u32 1 minus 1 -> Field
    v6 = add v3, v5
    v8 = array_get v1, index u32 2 minus 1 -> Field
    v9 = add v6, v8
    v10 = cast v2 as Field
    v11 = add v9, v10
    return v11
}

USED GLOBALS:
	f0: [v0]
After Dead Instruction Elimination (2) (step 38):
g0 = u32 3

  b0(v1: [Field; 3], v2: u32):
    v3 = array_get v1, index u32 3 minus 1 -> Field
    v5 = array_get v1, index u32 1 minus 1 -> Field
    v6 = add v3, v5
    v8 = array_get v1, index u32 2 minus 1 -> Field
    v9 = add v6, v8
    v10 = cast v2 as Field
    v11 = add v9, v10
    return v11
}
```
I don't actually see any use of `v0` or `g0` in the SSA. 

I wondered if the `minus 1` had anything to do with it: the second `array_get` could see `index u32 1 minus 1` as `v0`. Indeed it looks like this is the case; with some further prints:

In the first DIE pass, we only mark the array as used:
```
MARKING RESULTS OF ArrayGet { array: Id(1), index: Id(6), offset: None }
MARK USED v1: Param { block: Id(0), position: 0, typ: Array([Numeric(NativeField)], 3) }
MARKING RESULTS OF ArrayGet { array: Id(1), index: Id(3), offset: None }
MARK USED v1: Param { block: Id(0), position: 0, typ: Array([Numeric(NativeField)], 3) }
USED GLOBALS:
	f0: []
```

In the second DIE pass, we one of the `index` happens to take the ID of `v0`, the other does not:
```
MARKING RESULTS OF ArrayGet { array: Id(1), index: Id(4), offset: Array }
MARK USED v1: Param { block: Id(0), position: 0, typ: Array([Numeric(NativeField)], 3) }
MARKING RESULTS OF ArrayGet { array: Id(2), index: Id(0), offset: Array }
MARK USED v2: Param { block: Id(0), position: 0, typ: Array([Numeric(NativeField)], 3) }
MARK USED v0: NumericConstant { constant: 3, typ: Unsigned { bit_size: 32 } }
USED GLOBALS:
	f0: [v0]
```

Note that there is currently a `After Dead Instruction Elimination (3) (step 44)` step, but it's ACIR only, which is why with `--force-brillig` it might appear as there are _again_ no globals used, but that's because we never go into the `f0` function with `--force-brillig`. 

### To Reproduce

```shell
cd test_programs/execution_success/integer_array_indexing
cargo run -q -p nargo_cli -- execute --minimal-ssa
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.7 noirc version = 1.0.0-beta.7+ac2b1236e2959c5a829f65a3203da0095786a2b6 (git version hash: d068d38f359e9ae656ba7d02c4404cf6f583bbbc, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #8941

## Summary\*

Changes the order of "Dead Instruction Elimination" and "Brillig Array Get and Set Optimizations" in the `--minimal-ssa` pipeline to do the DIE as the last step. 

The reason is the Brillig array index optimisations shift constant indexes by an offset, and then insert the new constants into the DFG; when that happens the DFG tries to reuse existing constants and return their IDs, and these constants can come from the global as well as local values. If it comes from a global, it can mark a global value as _used_ even when it wasn't otherwise referenced. 

The DIE pass gathers used global values, but in the minimal pipeline it came before the index offsets, so it worked with stale data. 

## Additional Context

I don't know if this is an intended reuse, I suppose it might save some allocations the Brillig VM if we reuse global values, although it seems dubious if this is the _only_ use of a global value. Nevertheless, swapping the order doesn't seem to have any detrimental effect.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa.rs
+++ b/compiler/noirc_evaluator/src/ssa.rs
@@ -228,7 +228,7 @@ pub fn secondary_passes(brillig: &Brillig) -> Vec<SsaPass> {
         // It could happen that we inlined all calls to a given brillig function.
         // In that case it's unused so we can remove it. This is what we check next.
         SsaPass::new(Ssa::remove_unreachable_functions, "Removing Unreachable Functions"),
-        SsaPass::new(Ssa::dead_instruction_elimination_acir, "Dead Instruction Elimination"),
+        SsaPass::new(Ssa::dead_instruction_elimination_acir, "Dead Instruction Elimination - ACIR"),
     ]
 }
 
--- a/compiler/noirc_evaluator/src/ssa.rs
+++ b/compiler/noirc_evaluator/src/ssa.rs
@@ -207,7 +207,10 @@ pub fn primary_passes(options: &SsaEvaluatorOptions) -> Vec<SsaPass> {
         // We can safely place the pass before DIE as that pass only removes instructions.
         // We also need DIE's tracking of used globals in case the array get transformations
         // end up using an existing constant from the globals space.
+        // This pass might result in otherwise unused global constant becoming used,
+        // because the creation of shifted index constants can reuse their IDs.
         SsaPass::new(Ssa::brillig_array_get_and_set, "Brillig Array Get and Set Optimizations"),
+        // Perform another DIE pass to update the used globals after offsetting Brillig indexes.
         SsaPass::new(Ssa::dead_instruction_elimination, "Dead Instruction Elimination"),
         // A function can be potentially unreachable post-DIE if all calls to that function were removed.
         SsaPass::new(Ssa::remove_unreachable_functions, "Removing Unreachable Functions"),
--- a/compiler/noirc_evaluator/src/ssa.rs
+++ b/compiler/noirc_evaluator/src/ssa.rs
@@ -249,8 +249,8 @@ pub fn minimal_passes() -> Vec<SsaPass<'static>> {
         // panics later, when we are looking for global allocations.
         SsaPass::new(Ssa::remove_unreachable_functions, "Removing Unreachable Functions"),
         // We need to add an offset to constant array indices in Brillig.
-        // This can change which globals are used, because it shifts index constants,
-        // which might result in the (re)use of global constants.
+        // This can change which globals are used, because constant creation might result
+        // in the (re)use of otherwise unused global values.
         SsaPass::new(Ssa::brillig_array_get_and_set, "Brillig Array Get and Set Optimizations"),
         // We need a DIE pass to populate `used_globals`, otherwise it will panic later.
         SsaPass::new(Ssa::dead_instruction_elimination, "Dead Instruction Elimination"),
--- a/compiler/noirc_evaluator/src/ssa.rs
+++ b/compiler/noirc_evaluator/src/ssa.rs
@@ -248,10 +248,12 @@ pub fn minimal_passes() -> Vec<SsaPass<'static>> {
         // which was called in the AST not being called in the SSA. Such functions would cause
         // panics later, when we are looking for global allocations.
         SsaPass::new(Ssa::remove_unreachable_functions, "Removing Unreachable Functions"),
+        // We need to add an offset to constant array indices in Brillig.
+        // This can change which globals are used, because it shifts index constants,
+        // which might result in the (re)use of global constants.
+        SsaPass::new(Ssa::brillig_array_get_and_set, "Brillig Array Get and Set Optimizations"),
         // We need a DIE pass to populate `used_globals`, otherwise it will panic later.
         SsaPass::new(Ssa::dead_instruction_elimination, "Dead Instruction Elimination"),
-        // We need to add an offset to constant array indices in Brillig.
-        SsaPass::new(Ssa::brillig_array_get_and_set, "Brillig Array Get and Set Optimizations"),
     ]
 }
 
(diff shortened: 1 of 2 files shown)


vezenovm:
> although it seems dubious if this is the _only_ use of a global value

It is not. Global values are declared by users. We just also use them within the compiler for sharing constants across functions. We may want to move to only sharing array constants in globals and always inlining simple numeric constants, but that would be a separate experiment.

compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -248,10 +248,12 @@ pub fn minimal_passes() -> Vec<SsaPass<'static>> {
         // which was called in the AST not being called in the SSA. Such functions would cause
         // panics later, when we are looking for global allocations.
         SsaPass::new(Ssa::remove_unreachable_functions, "Removing Unreachable Functions"),
-        // We need a DIE pass to populate `used_globals`, otherwise it will panic later.
-        SsaPass::new(Ssa::dead_instruction_elimination, "Dead Instruction Elimination"),
         // We need to add an offset to constant array indices in Brillig.
+        // This can change which globals are used, because constant creation might result
+        // in the (re)use of otherwise unused global values.
         SsaPass::new(Ssa::brillig_array_get_and_set, "Brillig Array Get and Set Optimizations"),
+        // We need a DIE pass to populate `used_globals`, otherwise it will panic later.
```
vezenovm, line 258: Could we add these comments to `brillig_array_get_and_set` in the primary passes as well? I do think I put DIE after `brillig_array_get_and_set` in the primary passes due to this reason as well, but it was an oversight to not leave any comments.

vezenovm (approved):

compiler/noirc_evaluator/src/ssa.rs
vezenovm: Oh I guess there was this comment:
```
        // We also need DIE's tracking of used globals in case the array get transformations
        // end up using an existing constant from the globals space.
```
The additional comment helps provide clarification either way.

aakoshh:
> It is not. Global values are declared by users. 

I didn't mean it was wrong to use it, or that it's artificially introduced into the globals, just wanted to say that it seems a bit random if e.g. a single array item ends up surviving from the globals _only_ because a shift ends up matching its value. In other words, it's not like we piggy back on a global that has its use on its own in the _current_ SSA, but this way we save an extra slot, but rather this is the only thing that keeps it being used. There is no mechanism to seek opportunities for sharing, by elevating constants into the global space. 

Anyway, thanks for confirming that it works as intended!

vezenovm:
> In other words, it's not like we piggy back on a global that has its use on its own in the _current_ SSA, but this way we save an extra slot, but rather this is the only thing that keeps it being used. There is no mechanism to seek opportunities for sharing, by elevating constants into the global space.

Ah I see. We do in fact have a mechanism for elevating local constants to global constants, it just occurs after SSA gen and as part of Brillig gen. You can look in `brillig_globals.rs` for more information and here is an example unit test https://github.com/noir-lang/noir/blob/d1e55d93e31cefc8331d8f1174f56db87a995374/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_globals.rs#L500

PR #8947, opened 2025-06-17
chore(validation): Ban function pointers in SSA globals 

### LINKED ISSUE #8906: SSA Validation: Do not allow function pointers in globals
### Problem

As a follow-up to https://github.com/noir-lang/noir/pull/8905 and more generally, we do not want to allow function pointers in globals. It leads to issues like https://github.com/noir-lang/noir/issues/8898 which would require modifying globals in SSA which is unideal. 

### Happy Case

The frontend should resolve all function pointers or inline globals containing function pointers. We should have a validation check that makes sure we have no function pointers in globals.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

This guarantees that someone cannot generate bad SSA which could trigger a compiler panic.

Yes

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #8906 

## Summary\*

1. We now have an additional `type_check_globals` method we run during validation. This simply fetches the SSA type of the global value and determines whether it contains a `Value::Function`. If it does contain a function value it panics. 
2. The SSA printer and parser had to be updated to handle function pointers in globals. This was necessary in order to be able to use the SSA parser for the validation test.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/types.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/types.rs
@@ -301,8 +301,9 @@ impl Type {
     /// True if this is a function type or if it is a composite type which contains a function.
     pub(crate) fn contains_function(&self) -> bool {
         match self {
+            Type::Reference(element_type) => element_type.contains_function(),
             Type::Function => true,
-            Type::Numeric(_) | Type::Reference(_) => true,
+            Type::Numeric(_) => false,
             Type::Array(elements, _) | Type::Slice(elements) => {
                 elements.iter().any(|elem| elem.contains_function())
             }
--- a/compiler/noirc_evaluator/src/ssa/ir/types.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/types.rs
@@ -297,6 +297,17 @@ impl Type {
             }
         }
     }
+
+    /// True if this is a function type or if it is a composite type which contains a function.
+    pub(crate) fn contains_function(&self) -> bool {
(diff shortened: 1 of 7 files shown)


compiler/noirc_evaluator/src/ssa/ir/types.rs
```diff
@@ -297,6 +297,17 @@ impl Type {
             }
         }
     }
+
+    /// True if this is a function type or if it is a composite type which contains a function.
+    pub(crate) fn contains_function(&self) -> bool {
+        match self {
+            Type::Function => true,
+            Type::Numeric(_) | Type::Reference(_) => true,
```
asterite:
```suggestion
            Type::Numeric(_) | Type::Reference(_) => false,
```

Though I don't know if for references it should check if the referenced value is a function.

jfecher: > Though I don't know if for references it should check if the referenced value is a function.

It should - so it should recur on the element type. I think it should be possible to use references in globals... we haven't tested that at all though so it scares me a bit for mem2reg.

vezenovm: doh 🤦 yes numeric types do not contain functions. We have to recur on the element type like Jake said so I updated this locally.

asterite (approved):
Looks good!

Issue #8933, opened 2025-06-16
Min != Full: copy made instead of reference to `Field` (mem2reg)


aakoshh:
### Aim

Discovered by the `min_vs_full` fuzzer. 

The following program returns different result when executed with `--minimal-ssa`:

```rust
fn main() -> pub Field {
    let mut b = 0;
    let mut d = (&mut b);
    for idx_e in 10..11 {
        b = (idx_e as Field);
        b = (*d);
    }
    b
}
```

### Expected Behavior

The result should be the same. 

I'd expect that `d` is a reference to `b`, so any time we dereference it it should return the current value of `b`. So, the circuit should return 10. 

### Bug

```console
❯ cargo run -q -p nargo_cli -- execute --minimal-ssa --force --silence-warnings; cargo run -q -p nargo_cli -- execute --force --silence-warnings
[fuzz_testing] Circuit witness successfully solved
[fuzz_testing] Witness saved to target/fuzz_testing.gz
[fuzz_testing] Circuit output: Field(10)
[fuzz_testing] Circuit witness successfully solved
[fuzz_testing] Witness saved to target/fuzz_testing.gz
[fuzz_testing] Circuit output: Field(0)
```

So the full circuit restores the initial value. It stops doing that if we just assign 10 instead of doing it in a loop.

The `interpret` command suggests it happens after `mem2reg`:

```console
❯ cargo run -q -p nargo_cli -- interpret --ssa-pass "step 6" --ssa-pass "step 7" --show-ssa-pass "step 6" --show-ssa-pass "step 7"
After Inlining (step 6):
acir(inline) fn main f0 {
  b0():
    v1 = allocate -> &mut Field
    store Field 0 at v1
    v3 = allocate -> &mut &mut Field
    store v1 at v3
    jmp b1(u32 10)
  b1(v0: u32):
    v6 = lt v0, u32 11
    jmpif v6 then: b2, else: b3
  b2():
    v8 = cast v0 as Field
    store v8 at v1
    v9 = load v3 -> &mut Field
    v10 = load v9 -> Field
    store v10 at v1
    v12 = unchecked_add v0, u32 1
    jmp b1(v12)
  b3():
    v7 = load v1 -> Field
    return v7
}

--- Interpreter result after Inlining (step 6):
Ok([Numeric(Field(10))])
---
After Mem2Reg (step 7):
acir(inline) fn main f0 {
  b0():
    v1 = allocate -> &mut Field
    store Field 0 at v1
[...]

PR #8955, opened 2025-06-17
Add a aliased check for last stores

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/8933

## Summary\*

Gates the last store removal behind an additional check for aliased values. We now only remove last stores if the value is not aliased.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
@@ -456,8 +456,10 @@ impl<'f> PerFunctionContext<'f> {
 
                 // If there was another store to this instruction without any (unremoved) loads or
                 // function calls in-between, we can remove the previous store.
-                if let Some(last_store) = references.last_stores.get(&address) {
-                    self.instructions_to_remove.insert(*last_store);
+                if !self.aliased_references.contains_key(&address) {
+                    if let Some(last_store) = references.last_stores.get(&address) {
+                        self.instructions_to_remove.insert(*last_store);
+                    }
                 }
 
                 references.for_each_alias_of(value, |_, alias| {
@@ -1511,6 +1513,40 @@ mod tests {
           b6(v1: u1):
             return v1
         }
+        ";
+        let ssa = Ssa::from_str(src).unwrap();
+        let ssa = ssa.mem2reg();
+        assert_normalized_ssa_equals(ssa, src);
+    }
+
+    #[test]
+    fn keep_last_stores_with_aliased_references() {
+        // Ensure `store v8 at v1` is not removed from the program
+        // just because there is a subsequent `store v10 at v1`.
+        // In this case `v1` is aliased to `*v3` so the store is significant.
+        let src = "
+            acir(inline) fn main f0 {
+              b0():
+                v1 = allocate -> &mut Field
+                store Field 0 at v1
+                v3 = allocate -> &mut &mut Field
+                store v1 at v3
+                jmp b1(u32 10)
+              b1(v0: u32):
+                v6 = lt v0, u32 11
+                jmpif v6 then: b2, else: b3
+              b2():
+                v8 = cast v0 as Field
+                store v8 at v1
+                v9 = load v3 -> &mut Field
+                v10 = load v9 -> Field
+                store v10 at v1
+                v12 = unchecked_add v0, u32 1
+                jmp b1(v12)
+              b3():
+                v7 = load v1 -> Field
+                return v7
+            }
         ";
         let ssa = Ssa::from_str(src).unwrap();
         let ssa = ssa.mem2reg();


compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -950,6 +950,7 @@ mod tests {
             v2 = allocate -> &mut &mut Field
             jmp b1()
           b1():
+            store Field 1 at v0
```
aakoshh, line 953: It's not clear to me why this store should be kept. All we do is return from this block.

jfecher: It's not required, it's just something we fail to optimize out now that we don't remove stores in mem2reg when a reference is aliased since before with the issue we were removing them even when they were aliased and still needed by that alias.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -456,8 +456,10 @@ impl<'f> PerFunctionContext<'f> {
 
                 // If there was another store to this instruction without any (unremoved) loads or
                 // function calls in-between, we can remove the previous store.
-                if let Some(last_store) = references.last_stores.get(&address) {
-                    self.instructions_to_remove.insert(*last_store);
+                if !self.aliased_references.contains_key(&address) {
```
aakoshh, line 459: So this keeps the store even if all instructions which end up using its effect are removed, right? Maybe there is a way to track that after `store v1 at v3` we have `v3` as an alias of `v1`, and thus `v9 = load v3` is using the value in `store v8 at v1`? So that we can somehow decide later whether to remove the last store in `is_store_alias_used`?

jfecher: I think adding tracking like that would be a more significant addition which may add more bugs I want to avoid.

This is a slight regression but does not seem large enough to actually affect any of our tests. Perhaps we just need more tests with reference aliasing though.

vezenovm:
> Gates the last store removal behind an additional check for aliased values. We now only remove last stores if the value is not aliased.

I'm not 100% on this solution as the `aliased_references` map was made to assist the full function last store removal. Working into the per block last store removal feels a little bit like mixing concerns. I wonder if there is some other way to determine whether we have an alias here or if we are simply not marking aliases correctly. However, as this does provide an easy fix I won't block on this. It does feel though like we are simply not marking aliases correctly if in the following SSA:
```
            acir(inline) fn main f0 {
              b0():
                v1 = allocate -> &mut Field
                store Field 0 at v1
                v3 = allocate -> &mut &mut Field
                store v1 at v3
                jmp b1(u32 10)
              b1(v0: u32):
                v6 = lt v0, u32 11
                jmpif v6 then: b2, else: b3
              b2():
                v8 = cast v0 as Field
                store v8 at v1
                v9 = load v3 -> &mut Field
                v10 = load v9 -> Field
                store v10 at v1
                v12 = unchecked_add v0, u32 1
                jmp b1(v12)
              b3():
                v7 = load v1 -> Field
                return v7
            }
```
v1 is aliased to `*v3` but we are not seeing that. Perhaps we just need something per block to more accurately track dereferenced aliases per block similar to `aliased_references`.

vezenovm (approved):

jfecher:
> v1 is aliased to *v3 but we are not seeing that. 

Yeah, the issue is that it is aliased to `*v3` but we don't actually have a separate ValueId for `*v3` yet.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -1516,4 +1517,38 @@ mod tests {
...
+              b0():
+                v1 = allocate -> &mut Field
+                store Field 0 at v1
+                v3 = allocate -> &mut &mut Field
+                store v1 at v3
+                jmp b1(u32 10)
+              b1(v0: u32):
+                v6 = lt v0, u32 11
+                jmpif v6 then: b2, else: b3
+              b2():
```
aakoshh, line 1537: Here's something I noticed while trying to find a different way to do this: 
* we process blocks in the order of `b0`, `b1`, `b3`, `b2`
* when we process `b0`, we insert the known value of `v1` for `v3` into `references`.
* when we consider `b1`, its predecessors are `b0` and `b2`, and out of these only `b0` has been processed, and for `b2` we use a `Block::default()` in `find_starting_references`. 
* in `Block::unify` we only keep the intersection for `references`, so we forget the known value for `v3`, even though it's not set in `b2`, it stays `v1`. 
* therefore `b1` is starting with an empty `references` map
* and because `b1` is the only predecessor of `b2`, when we process `b2` we don't remember what we know about `v3` when we load it.

I thought if we remembered when we analysed `v9 = load v3` that we know that `v3` is actually `v1`, then we would know not to remove `store v8 at v1`. Not sure if it could work over longer redirections, though.

aakoshh: I wonder if there could be some preliminary processing of the blocks that doesn't require their predecessors to be done already. For example we could have a look at `b2` and see which keys it stores to, set them as `Unknown`, and then if one predecessor has a `v1: Known(x), v2: Known(y)` and the other predecessors has `v2: Uknown`, then we could keep `v1: Known(x)`.

I suppose it's possible that it the same key is set via an alias 🤔 Never mind then.

jfecher: > in Block::unify we only keep the intersection for references, so we forget the known value for v3, even though it's not set in b2, it stays v1.

Yeah, this is by design. We deliberately have them unify to unknown in this case to handle loops correctly where they may be set in the future.

> I wonder if there could be some preliminary processing of the blocks that doesn't require their predecessors to be done already.

I've wondered this for a long time now as well. I've had a couple experiments with it but they all end up unsound with loops. See the ramblings here: https://hackmd.io/9bbW07mCSiWlS-i11RWKGA#Better-Loop-Optimizations

aakoshh: Donald Rumsfeld would be proud of the "Known Not Yet Knowns" 😄 

My suggestion to that exact SSA would be to recognise that `v1` is affected by the loop, so it should unify to Unknown. 

But when `v1` is really `&mut v0`, and only `v0` is affected, then I'm not sure this hold. I suppose we could say `v1` is still known to be `v0`, because one has to `v2 = load v1` and follow with `v3 = load v2` to get to `v0`, but any real underlying constant of `**v1` is unknown 🤷

jfecher:
I've thought about keeping this open since it seems it can be done in a less ad-hoc measure but the regression here seems small and it fixes the bug. It is also a small enough change we can come back to it later.

Issue #8870, opened 2025-06-10
Comptime crash: `-128 as i8`


aakoshh:
### Aim

Discovered testing #8868 

The following program crashes the compiler:

```rust
fn main() {
    comptime {
        let i: i8 = -128 as i8;
        println(i);
    }
}
```

### Expected Behavior

It should print `-128`.  At least it shouldn't crash.

I understand now that `as i8` changes the behaviour to a negation of a positive value, and in fact it gives a failure when executed as an ACIR program as well, but I did not expect this because of this technique mentioned as a workaround in https://github.com/noir-lang/noir/issues/8759

### Bug

```console
❯ cargo run -q -p nargo_cli -- compile
The application panicked (crashed).
Message:  attempt to negate with overflow
Location: compiler/noirc_frontend/src/hir/comptime/interpreter.rs:1599
```

If I remove the `as i8` bit then it works. I added that in https://github.com/noir-lang/noir/pull/8847 as a workaround for the ambiguity of literals.

### To Reproduce

See above.

If I change it to non-comptime, we get a failure:
```console
❯ cargo run -q -p nargo_cli -- execute
warning: Casting value of type Field to a smaller type (i8)
  ┌─ src/main.nr:3:18
  │
3 │     let i: i8 = -128 as i8;
  │                  --------- casting untyped value (128) to a type with a maximum size (127) that's smaller than it
  │

bug: Assertion is always false
  ┌─ src/main.nr:3:17
  │
3 │     let i: i8 = -128 as i8;
  │                 ---------- As a result, the compiled circuit is ensured to fail. Other assertions may also fail during execution
  │
  = Call stack:
    1. src/main.nr:3:17

error: Assertion failed: attempt to subtract with overflow
  ┌─ src/main.nr:3:17
  │
3 │     let i: i8 = -128 as i8;
  │                 ----------
  │
  = Call stack:
    1. src/main.nr:3:17

Failed assertion

```

### Workaround

None

### Project Impact

None

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+501dc08ed0756a69b309b904091adb22776d55c8 (git version hash: 501dc08ed0756a69b309b904091adb22776d55c8, is dirty: true)

### Would you like to submit a PR for this Issue?

None

asterite:
An interesting thing here is that it seems `-128 as i8` is parsed as `-(128 as i8)` 😮

That's why it fails: `128` doesn't fit in `i8`.

[...]

PR #8956, opened 2025-06-17
fix(parser): let `as` have a lower precedence

asterite:
# Description

## Problem

Related to #8870

## Summary

`-128 as i8` was being parsed as `-(128 as i8)` which might be a bit unintuitive, and it's also different from Rust. This PR changes that so it's parsed as `(-128) as i8`.

That makes things a bit less intuitive in some cases because `-128 as i16` is actually `-127`, but that's what you get if you also do `let x: Field = -128; x as i16`...

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/parser/parser/expression.rs
```diff
@@ -109,15 +136,19 @@ impl Parser<'_> {
         }
     }
 
-    /// AtomOrUnaryRightExpression
-    ///     = Atom UnaryRightExpression*
-    fn parse_atom_or_unary_right(&mut self, allow_constructors: bool) -> Option<Expression> {
+    /// AtomOrMemberAccessOrCallOrMethodCallOrIndexExpressoin
```
TomAFrench:
```suggestion
    /// AtomOrMemberAccessOrCallOrMethodCallOrIndexExpression
```

compiler/noirc_frontend/src/parser/parser/expression.rs
```diff
@@ -70,13 +70,40 @@ impl Parser<'_> {
...
 
     /// Term
-    ///    = UnaryOp Term
-    ///    | AtomOrUnaryRightExpression
+    ///    = UnaryOrAtomOrMemberAccessOrCallOrMethodCallOrIndexExpression
+    ///    | CatExpression
     pub(super) fn parse_term(&mut self, allow_constructors: bool) -> Option<Expression> {
         let start_location = self.current_token_location;
 
+        let mut term = self.parse_unary_or_atom_or_memeber_access_or_call_or_method_call_or_index(
```
TomAFrench:
```suggestion
        let mut term = self.parse_unary_or_atom_or_member_access_or_call_or_method_call_or_index(
```

compiler/noirc_frontend/src/parser/parser/expression.rs
```diff
@@ -70,13 +70,40 @@ impl Parser<'_> {
...
+                break;
+            }
+        }
+
+        Some(term)
+    }
+
+    /// UnaryOrAtomOrMemberAccessOrCallOrMethodCallOrIndexExpression
+    ///    = UnaryOp* AtomOrMemberAccessOrCallOrMethodCallOrIndexExpression
+    fn parse_unary_or_atom_or_memeber_access_or_call_or_method_call_or_index(
```
TomAFrench:
```suggestion
    fn parse_unary_or_atom_or_member_access_or_call_or_method_call_or_index(
```

compiler/noirc_frontend/src/parser/parser/expression.rs
```diff
@@ -70,13 +70,40 @@ impl Parser<'_> {
...
+        &mut self,
+        allow_constructors: bool,
+    ) -> Option<Expression> {
+        let start_location = self.current_token_location;
+
         if let Some(operator) = self.parse_unary_op() {
-            let Some(rhs) = self.parse_term(allow_constructors) else {
+            let Some(rhs) = self
+                .parse_unary_or_atom_or_memeber_access_or_call_or_method_call_or_index(
```
TomAFrench:
```suggestion
                .parse_unary_or_atom_or_member_access_or_call_or_method_call_or_index(
```

compiler/noirc_frontend/src/parser/parser/expression.rs
```diff
@@ -70,13 +70,40 @@ impl Parser<'_> {
     }
 
     /// Term
-    ///    = UnaryOp Term
-    ///    | AtomOrUnaryRightExpression
+    ///    = UnaryOrAtomOrMemberAccessOrCallOrMethodCallOrIndexExpression
+    ///    | CatExpression
```
aakoshh:
```suggestion
    ///    | CastExpression
```
😺

compiler/noirc_frontend/src/parser/parser/expression.rs
```diff
@@ -128,12 +159,12 @@ impl Parser<'_> {
         Some(atom)
     }
 
-    /// UnaryRightExpression
+    /// MemberAccessOrCallOrMethodCallOrIndexExpression
     ///     = CallExpression
     ///     | MemberAccessOrMethodCallExpression
     ///     | CastExpression
```
aakoshh:
```suggestion
```

compiler/noirc_frontend/src/parser/parser/expression.rs
```diff
@@ -247,7 +273,7 @@ impl Parser<'_> {
         }
     }
 
-    /// CastExpression = Atom 'as' Type
+    /// CastExpression = UnaryOrAtomOrMemberAccessOrCallOrMethodCallOrIndexExpression 'as' Type
```
aakoshh: I was just wondering if there really is no shorter way to phrase this. I took the letters "U A MA C MC I" and rearranged them into "AMcIMaUC", which if you say it sounds a bit like a garbled "a mickey mouse" 🐭 
I recommend we take the mickey over this long name 😉

compiler/noirc_frontend/src/parser/parser/expression.rs
asterite: Oh, now I feel like I want to introduce a cat expression to the language, using that emoji. We just have to think what it would do...

aakoshh: Catch Mickey Mouse!

compiler/noirc_frontend/src/parser/parser/expression.rs
asterite: Hahaha, definitely! The names were nice before this PR, like "term" and "atom or unary right". But now there's two "unary right" types so I wasn't sure what name to use... I ended up using verbose and long names to avoid having to think of better names (but I don't like those names 😢)

jfecher:
> That makes things a bit less intuitive in some cases because -128 as i16

Once we add types directly on integer literals this will be less of an issue at least

compiler/noirc_frontend/src/parser/parser/expression.rs
```diff
@@ -70,13 +70,40 @@ impl Parser<'_> {
...
+                break;
+            }
+        }
+
+        Some(term)
+    }
+
+    /// UnaryOrAtomOrMemberAccessOrCallOrMethodCallOrIndexExpression
+    ///    = UnaryOp* AtomOrMemberAccessOrCallOrMethodCallOrIndexExpression
+    fn parse_unary_or_atom_or_member_access_or_call_or_method_call_or_index(
```
jfecher:
```suggestion
    /// Unary = UnaryOp* Atom
    fn parse_unary(
```

compiler/noirc_frontend/src/parser/parser/expression.rs
```diff
@@ -109,15 +136,19 @@ impl Parser<'_> {
         }
     }
 
-    /// AtomOrUnaryRightExpression
-    ///     = Atom UnaryRightExpression*
-    fn parse_atom_or_unary_right(&mut self, allow_constructors: bool) -> Option<Expression> {
+    /// AtomOrMemberAccessOrCallOrMethodCallOrIndexExpression
+    ///     = Atom MemberAccessOrCallOrMethodCallOrIndexExpression*
+    fn parse_atom_or_member_access_or_call_or_method_call_or_index(
```
jfecher:
```suggestion
    /// Atom = Quark AtomRhs*
    fn parse_atom(
```
Alternatively we can use `high_prec_postfix_op` or similar to distinguish it from `Cast` which is lower prec

compiler/noirc_frontend/src/parser/parser/expression.rs
```diff
@@ -109,15 +136,19 @@ impl Parser<'_> {
...
+        allow_constructors: bool,
+    ) -> Option<Expression> {
         let start_location = self.current_token_location;
         let mut atom = self.parse_atom(allow_constructors)?;
         let mut parsed;
 
         loop {
-            (atom, parsed) = self.parse_unary_right(atom, start_location);
+            (atom, parsed) =
+                self.parse_member_access_or_call_or_method_call_or_index(atom, start_location);
```
jfecher: Can't comment on it above since it wasn't changed but we can change `parse_atom` to `parse_quark` since I already used atom above.

Atom is a standard name, quark is not but it continues the theme and makes sense as an indivisible parselet

compiler/noirc_frontend/src/parser/parser/expression.rs
```diff
@@ -128,12 +159,11 @@ impl Parser<'_> {
...
     }
 
-    /// UnaryRightExpression
+    /// MemberAccessOrCallOrMethodCallOrIndexExpression
     ///     = CallExpression
     ///     | MemberAccessOrMethodCallExpression
-    ///     | CastExpression
     ///     | IndexExpression
-    fn parse_unary_right(
+    fn parse_member_access_or_call_or_method_call_or_index(
```
jfecher: Let's rename this one too.

This can be `atom_rhs` or `high_prec_postfix_op`, etc depending on the name you choose for my previous comment

jfecher (commented):
Let's just simplify the parser names.

They don't need to contain every rule. Usually a parser rule X includes X and any base cases. E.g. `term` includes not just multiplication terms but also atoms in other langs but they don't name it term_or_atom, etc.

compiler/noirc_frontend/src/parser/parser/expression.rs
asterite: Done!

compiler/noirc_frontend/src/parser/parser/expression.rs
asterite: Done!

compiler/noirc_frontend/src/parser/parser/expression.rs
asterite: In the end I went with `atom_rhs`

compiler/noirc_frontend/src/parser/parser/expression.rs
asterite: In the end I went with `atom_rhs`

asterite:

PR #8960, opened 2025-06-18
fix(mem2reg): Keep store when any aliased reference is kept

### LINKED ISSUE #8926: Execution crash: `value is not typed as brillig usize`
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8925

The following program crashes during execution
```rust
unconstrained fn main(b: ([Field; 2], str<4>)) {
    func_2();
}
unconstrained fn func_2() -> (str<4>) {
    let mut a = (&mut "CQVV");
    a = a;
    let mut idx_b = 0;
    loop {
        if (idx_b == 0) {
            break
        }
    }
    (*a)
}

```

input:
```toml
b = [
    [
        "0x2622702b40f5fab7ad82e8ca3c9edfef03120af6ff694bf1626c182915fd090d",
        "0x20bc958d7fb1d94821586da3a91ce82ed1361e83a3d393d6e5b9882b647d268a",
    ],
    "(x$f",
]
```

### Expected Behavior

Shouldn't crash. A similar bug was fixed in https://github.com/noir-lang/noir/issues/8727

### Bug

```console
warning: Unused expression result of type str<4>
  ┌─ src/main.nr:2:5
  │
2 │     func_2();
  │     --------
  │

warning: unused variable b
  ┌─ src/main.nr:1:23
  │
1 │ unconstrained fn main(b: ([Field; 2], str<4>)) {
  │                       - unused variable
  │

Message:  value is not typed as brillig usize
Location: /Users/aakoshh/Work/aztec/noir/acvm-repo/brillig_vm/src/memory.rs:62

```

The crash goes away in either of these cases:
* change the type of the unused `b` parameter to `b: (str<4>)` or `b: ([Field; 2])`
* remove `a = a;`
* remove the `loop` that does nothing

### To Reproduce

See above.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.7 noirc version = 1.0.0-beta.7+ac2b1236e2959c5a829f65a3203da0095786a2b6 (git version hash: 2e36b3098e765809227ac9685eded019674bf203, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #8926 

## Summary\*

Changes `is_store_alias_used` in `mem2reg` to consider a store used if there is any instance in `aliased_references` that isn't in `instructions_to_remove`. Previously it only kept it if _all_ aliases were not to be removed.

## Additional Context

In the integration tests added, we started with the following SSA:
```rust
After Inlining (step 6):
brillig(inline) fn main f0 {
  b0():
    v3 = make_array b"ABC"
    v4 = allocate -> &mut [u8; 3]
    store v3 at v4
    v5 = allocate -> &mut &mut [u8; 3]
    store v4 at v5
    v6 = load v5 -> &mut [u8; 3]
    store v6 at v5
    v7 = allocate -> &mut Field
    store Field 0 at v7
    jmp b1()
  b1():
    v9 = load v7 -> Field
    v10 = eq v9, Field 0
    jmpif v10 then: b2, else: b3
  b2():
    jmp b4()
  b3():
    jmp b1()
  b4():
    v11 = load v5 -> &mut [u8; 3]
    v12 = load v11 -> [u8; 3]
    inc_rc v12
    return v12
}
```

Then `mem2reg` removed the store of "ABC":
```rust
After Mem2Reg (step 7):
brillig(inline) fn main f0 {
  b0():
    v3 = make_array b"ABC"
    v4 = allocate -> &mut [u8; 3]
    v5 = allocate -> &mut &mut [u8; 3]
    store v4 at v5
    v6 = allocate -> &mut Field
    store Field 0 at v6
    jmp b1()
  b1():
    v8 = load v6 -> Field
    v9 = eq v8, Field 0
    jmpif v9 then: b2, else: b3
  b2():
    jmp b4()
  b3():
    jmp b1()
  b4():
    v10 = load v5 -> &mut [u8; 3]
    v11 = load v10 -> [u8; 3]
    inc_rc v11
    return v11
}
```

The way it analysed instructions was:
* `store v4 at v5` added `v4` to the `aliased_references`, because it's used in this store instruction
* `store v6 at v5` might have marked `store v4 at v5` to be removed, but also added another entry to `aliased_references` for `v4` (which it realised `v6` equals), so it was now used in two instructions
* when it was time to decide whether to keep the last store to `v4`, which is `store v3 at v4`, `is_store_alias_used` retrieved `[store v4 at v5, store v6 at v5]` from `aliased_references` of `v4`, and had `[v6 = load v5, store v6 at v5]` in `instructions_to_remove` (these are basically `a = a;`, which is correctly recognised as a no-op). 
* then it wanted the two to be _disjoint_, which it wasn't, so `store v3 at v4` was removed

I think it makes sense that if there is at least one instruction which we keep, and we use the value in that instruction, then we have to keep the store to it.

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
Does this resolve some of those those other mut &mut issues you recently posted?

aakoshh:
I don't know, I only checked the test in https://github.com/noir-lang/noir/pull/8955 but it did not fix that.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -252,15 +252,15 @@ impl<'f> PerFunctionContext<'f> {
                     return true;
                 }
 
-                let all_aliases_set_for_removal = aliases.any(|alias| {
+                let has_alias_not_marked_for_removal = aliases.any(|alias| {
```
vezenovm, line 257: Could we just add a comment above here? Something like "Check whether there are any aliases whose instructions are not all marked for removal. If there is any alias marked to survive, we should not remove its last store."

vezenovm (approved):
Good find! LGTM

Issue #8963, opened 2025-06-18
Integer type not correctly inferred when unrelated trait impl exists


asterite:
### Aim

This code:

```noir
trait From2<T> {
    fn from2(input: T) -> Self;
}

struct U60Repr {}

impl From2<[i32; 3]> for U60Repr {
    fn from2(_: [i32; 3]) -> Self {
        U60Repr {}
    }
}

// Comment this impl and it works fine
impl From2<i32> for U60Repr {
    fn from2(_: i32) -> Self {
        U60Repr {}
    }
}

fn main() {
    let x = [1, 2, 3];
    let _ = U60Repr::from2(x);}
```

doesn't compile: it inferrs that `x` is `[Field; 3]` when it should infer that it's `[i32; 3]`, because that's the only matching impl.

If the impl `From2<i32>` is removed, the code compiles fine.

### Expected Behavior

The code should compile.

### Bug

The code doesn't compile.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

asterite:
The problem seems to be that we resolve `U60Repr::from2` to a trait method (we don't know the impl yet), then we default the integer types to `Field`, then we verify the trait constraints so they fail. But I feel that with the call `U60Repr::from2(x)` the compiler should be able to discard the non-array impl since the argument doesn't match... though I don't know how that can be done (yet).

PR #8965, opened 2025-06-18
feat(experimental!): let the default integer type be i32 instead of Field

asterite:
# Description

## Problem

No reported issue, but having Field be the default integer type might lead to confusion.

## Summary

Note: this is just an experiment to see how much code breaks with this change. There's also [this bug](https://github.com/noir-lang/noir/issues/8963) that, when fixed, should lead to even less breakage (I'll try to fix that next).

The main issue with having integer literals default to Field is that negative literals are allowed, but Field is mainly an unsigned type. It's okay to allow negative literals for it, but that should only probably be allowed with an explicit type annotation, or if the literal destination is a Field. For example right now `(-56) as i8` gives `-55`, because the field `-56` is actually a very big positive number that, when casted to `i8`, gives `-55`. With the change in this PR, `-56` is an `i32`, and casting that to `i8` preserves the value `-56`. But regardless of this issue, it might make more sense for the default type to be a signed type because literals can be positive or negative.

The reasoning here is that because Noir, like Rust, requires type annotations in function signatures, struct types, etc., changing the default from `Field` to `i32` shouldn't have a big impact, because if existing code expects `Field`, it will still use `Field`, and code would still compile.

For example:

```noir
fn foo(x: Field) { ... }

fn main() {
    foo(1)
}
```

Here it doesn't matter what that `1` defaults to, because the type will be bound to the call argument. It's the same case if that's assigned to a struct field, etc. The only issue right now is #8963 where this isn't true, but that's a bug that we can fix.

Then in isolated code the type can change. For example here:

```noir
let x = 1;
println(x);
```

That used to print `0x01`, but now it prints `1` because `x` is `i32` instead of `Field`. But that code is mostly used in tests. For example, only one change was required to be done to `noir_stdlib` to make it compile again, and it was in some test code related to macros.

In any case, this is just experimental! I understand that Field is more efficient than any other integer type, but my hope is that this change doesn't actually impact performance.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir_def/types/unification.rs
+++ b/compiler/noirc_frontend/src/hir_def/types/unification.rs
@@ -41,7 +41,10 @@ impl Kind {
             | (Kind::Integer | Kind::SignedIntegerOrField, Kind::Normal) => true,
 
             // Kind::Integer unifies with Kind::IntegerOrField
-            (Kind::Integer | Kind::SignedIntegerOrField, Kind::Integer | Kind::SignedIntegerOrField) => true,
+            (
+                Kind::Integer | Kind::SignedIntegerOrField,
+                Kind::Integer | Kind::SignedIntegerOrField,
+            ) => true,
 
             // Kind::IntegerOrField unifies with Kind::Numeric(_)
             (Kind::SignedIntegerOrField, Kind::Numeric(_typ))
@@ -140,12 +143,11 @@ impl Type {
                         })
                     }
                 }
-                TypeBinding::Unbound(_id, Kind::SignedIntegerOrField) => {
-                    other.try_unify_to_type_variable(var, flags, bindings, |bindings| {
+                TypeBinding::Unbound(_id, Kind::SignedIntegerOrField) => other
+                    .try_unify_to_type_variable(var, flags, bindings, |bindings| {
                         let only_integer = false;
                         other.try_bind_to_polymorphic_int(var, bindings, only_integer)
-                    })
-                }
+                    }),
                 TypeBinding::Unbound(_id, Kind::Integer) => {
                     other.try_unify_to_type_variable(var, flags, bindings, |bindings| {
                         let only_integer = true;


jfecher (changes requested):
I agree that `(-56) as i8 = -55` is confusing but I think defaulting to Field is still sensible from an efficiency standpoint as well as a "what type is most commonly used?" standpoint.

I also think we should keep allowing negative literals for field since negatives are still consistent within field operations. E.g. `-5 + 7 = 2`. (If anything I think field division is more confusing).

If we don't want `(-56) as i8 = -55` as the expected semantics then I'd lean toward altering the field to signed int cast instead. We could add a branch in the cast which would make it slower but perhaps more expected. This may be fine for the signed integer types which are rarely used today:
```rs
// This would be built-in, but just to show the logic:
fn field_to_i8(x: Field) -> Option<i8> {
    // Absolute value of i8::MIN
    let abs_min = pow(2, 7);
    if x >= -abs_min {
        truncate(x + abs_min)
    } else {
        truncate(x)
    }
}
```

asterite:
> If we don't want (-56) as i8 = -55 as the expected semantics then I'd lean toward altering the field to signed int cast instead.

I agree, we could do that. However, I'd find it a bit odd that `(-56) as i8` would give `-56` but `21888242871839275222246405745257275088548364400416034343698204186575808495561 as i8` would also give `-56`... when both are the "same" number, except that the latter would be considered the negative version.

Maybe the odd thing is that printing the Field `-1` shows `0x30644e72e131a029b85045b68181585d2833e84879b9709143e1f593efffffc9`? That is, we allow negative fields literals, we even have some branching logic in our codebase that field values above a certain value are considered negative, but then `-1` isn't actually shown like that. If we changed Field to be like a signed type, maybe that would solve these issues? For example casting the i8 `-1` to `Field` should give you `-1`, not 0 (as I think you proposed a while ago).

asterite:
The above would also mean that this could shouldn't compile:

```noir
fn main() {
    let x = 21888242871839275222246405745257275088548364400416034343698204186575808495616;
    println(x);
}
```

as `21888242871839275222246405745257275088548364400416034343698204186575808495616` is currently the max field value, but if we changed it to be treated like a signed integer then the maximum value would be about half of that.

jfecher:
@asterite I don't think we should dedicate bit patterns in field to be exclusively negative unless we wanted to add it as a separate signed field type.

For the printing discrepancy, I just think we're going to have to draw the line somewhere.
- It could be "negative field values are actually positive, so when printing they appear as large positive numbers,"
- It could be "large/negative field values _cannot_ be cast to signed integers (they give a range constraint error)"
- or somewhere else

asterite:
Thanks!

I'll close this in the meantime. It doesn't seem this change breaks a lot of things, which is what I wanted to know when creating this PR. The compilation errors are:
1. Large integer literals that were previously inferred to `Field` now don't fit and would need an explicit cast or type annotation
2. Issues because of the bug in #8963

asterite:
> I agree that (-56) as i8 = -55 is confusing but I think defaulting to Field is still sensible from an efficiency standpoint as well as a "what type is most commonly used?" standpoint.

@jfecher  I kept thinking about this and wondering if this is still true today:

https://github.com/noir-lang/noir/pull/10096

I understand that Field is the most efficient type, but according to the above PR we want to discourage users from using Field.

PR #8969, opened 2025-06-18
fix: check "negate with overflow" in comptime code

### LINKED ISSUE #8870: Comptime crash: `-128 as i8`

asterite:
# Description

## Problem

Resolves #8870

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/errors.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/errors.rs
@@ -126,8 +126,11 @@ pub enum InterpreterError {
         operator: &'static str,
         location: Location,
     },
-    MathError {
+    InfixMathError {
         operator: &'static str,
+        location: Location,
+    },
+    PrefixMathError {
         location: Location,
     },
     CastToNonNumericType {
@@ -301,7 +304,8 @@ impl InterpreterError {
             | InterpreterError::TypeUnsupported { location, .. }
             | InterpreterError::InvalidValueForUnary { location, .. }
             | InterpreterError::InvalidValuesForBinary { location, .. }
-            | InterpreterError::MathError { location, .. }
+            | InterpreterError::InfixMathError { location, .. }
+            | InterpreterError::PrefixMathError { location, .. }
             | InterpreterError::CastToNonNumericType { location, .. }
             | InterpreterError::NonStructInConstructor { location, .. }
             | InterpreterError::NonEnumInConstructor { location, .. }
@@ -504,7 +508,7 @@ impl<'a> From<&'a InterpreterError> for CustomDiagnostic {
                 let msg = format!("No implementation for `{lhs}` {operator} `{rhs}`");
                 CustomDiagnostic::simple_error(msg, String::new(), *location)
             }
-            InterpreterError::MathError { operator, location } => {
+            InterpreterError::InfixMathError { operator, location } => {
                 let msg = if *operator == "/" {
                     "Attempt to divide by zero".to_string()
                 } else {
@@ -518,6 +522,10 @@ impl<'a> From<&'a InterpreterError> for CustomDiagnostic {
                     };
                     format!("Attempt to {operator} with overflow")
                 };
+                CustomDiagnostic::simple_error(msg, String::new(), *location)
+            }
+            InterpreterError::PrefixMathError { location } => {
+                let msg = "Attempt to negate with overflow".to_string();
                 CustomDiagnostic::simple_error(msg, String::new(), *location)
             }
             InterpreterError::CastToNonNumericType { typ, location } => {
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
@@ -1595,10 +1595,22 @@ fn evaluate_prefix_with_value(rhs: Value, operator: UnaryOp, location: Location)
     match operator {
         UnaryOp::Minus => match rhs {
             Value::Field(value) => Ok(Value::Field(-value)),
-            Value::I8(value) => Ok(Value::I8(-value)),
-            Value::I16(value) => Ok(Value::I16(-value)),
-            Value::I32(value) => Ok(Value::I32(-value)),
-            Value::I64(value) => Ok(Value::I64(-value)),
+            Value::I8(value) => value
+                .checked_neg()
+                .map(Value::I8)
+                .ok_or_else(|| InterpreterError::PrefixMathError { location }),
+            Value::I16(value) => value
+                .checked_neg()
+                .map(Value::I16)
+                .ok_or_else(|| InterpreterError::PrefixMathError { location }),
+            Value::I32(value) => value
+                .checked_neg()
+                .map(Value::I32)
+                .ok_or_else(|| InterpreterError::PrefixMathError { location }),
+            Value::I64(value) => value
+                .checked_neg()
+                .map(Value::I64)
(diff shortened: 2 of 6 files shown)


compiler/noirc_frontend/src/hir/comptime/errors.rs
```diff
@@ -520,6 +524,10 @@ impl<'a> From<&'a InterpreterError> for CustomDiagnostic {
                 };
                 CustomDiagnostic::simple_error(msg, String::new(), *location)
             }
+            InterpreterError::PrefixMathError { location } => {
+                let msg = "Attempt to negate with overflow".to_string();
+                CustomDiagnostic::simple_error(msg, String::new(), *location)
+            }
```
jfecher: Can you rename this to `NegateWithOverflow`?
I'd prefer that over the somewhat ambiguous more general name, especially when the error isn't actually as general as the name makes it sound.

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
```diff
@@ -1595,10 +1595,22 @@ fn evaluate_prefix_with_value(rhs: Value, operator: UnaryOp, location: Location)
...
+                .ok_or_else(|| InterpreterError::PrefixMathError { location }),
+            Value::I32(value) => value
+                .checked_neg()
+                .map(Value::I32)
+                .ok_or_else(|| InterpreterError::PrefixMathError { location }),
+            Value::I64(value) => value
+                .checked_neg()
+                .map(Value::I64)
+                .ok_or_else(|| InterpreterError::PrefixMathError { location }),
             Value::U8(value) => Ok(Value::U8(0 - value)),
```
jfecher: Do we allow unary negation for unsigned integers?

asterite: Hm, good point! Apparently it's forbidden in non-comptime code (it gives an error) but in comptime code it crashes the compiler:

```noir
fn main() {
    comptime {
        let i: u8 = 1;
        let _ = -i;
    }
}
```

```
Message:  attempt to subtract with overflow
Location: compiler/noirc_frontend/src/hir/comptime/interpreter.rs:1614
```

Maybe the error would still be reported later on but it ends up being evaluated at comptime 🤔 

I'll try to fix it in this PR too.

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
```diff
@@ -1595,15 +1595,29 @@ fn evaluate_prefix_with_value(rhs: Value, operator: UnaryOp, location: Location)
...
+                .ok_or_else(|| InterpreterError::NegateWithOverflow { location }),
+            Value::I32(value) => value
+                .checked_neg()
+                .map(Value::I32)
+                .ok_or_else(|| InterpreterError::NegateWithOverflow { location }),
+            Value::I64(value) => value
+                .checked_neg()
+                .map(Value::I64)
+                .ok_or_else(|| InterpreterError::NegateWithOverflow { location }),
+            Value::U8(_) => Err(InterpreterError::CannotApplyMinusToType { location, typ: "u8" }),
```
jfecher, line 1631: I think we're missing `Value::U1` here (and below), can you add it for consistency?

jfecher (approved):

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
asterite: I made it error. The only issue is that we now get the "same" error twice: once when type-checking, another one when evaluating comptime code. I thought about not erroring and just going on with the same value, but then it could be confusing to continue with an incorrect value if that incorrect value leads to more errors later on.

But then, if you have this code:

```noir
    comptime {
        let i: u8 = 1 + 300;
    }
```

you also get two errors on "300". The messages are slightly different, but they are essentially the same error. So maybe double erroring is fine (or at least it's not a new thing of this PR).

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
asterite: I added it locally, then noticed it's also missing when evaluating `for`, so I went to try it and I get this error message:

```
The type `u1` is currently unsupported in comptime expressions
```

Do you know why that is?

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
jfecher: The double error is definitely fine and is already the case for many type errors when run in comptime code

compiler/noirc_frontend/src/hir/comptime/interpreter.rs
jfecher: > Do you know why that is?

I think it is just that it was added afterward and wasn't supported initially but not all match cases were changed to account for this.

asterite: Done! I added support for u1 and fixed a few things.

compiler/noirc_frontend/src/hir/comptime/value.rs
```diff
@@ -609,6 +610,9 @@ impl Value {
             Self::I16(value) => (*value >= 0).then_some((*value as u128).into()),
             Self::I32(value) => (*value >= 0).then_some((*value as u128).into()),
             Self::I64(value) => (*value >= 0).then_some((*value as u128).into()),
+            Self::U1(value) => {
+                Some(if *value { FieldElement::one() } else { FieldElement::zero() })
```
TomAFrench:
```suggestion
                Some(FieldElement::from(*value))
```

Issue #8968, opened 2025-06-18
Variables can be lexed as part of an integer only if the variable only contains a, b, c, d, e or f


jfecher:
When we see `1234abc` it is lexed as a single token:

```
error: Invalid integer literal
  ┌─ src/main.nr:2:5
  │
2 │     1234abc;
  │     -------  1234abc is not an integer
  │

```

But if we had `1234jqp` it is treated as two separate tokens:

```
error: cannot find `jqp` in this scope
  ┌─ src/main.nr:2:9
  │
2 │     1234jqp;
  │         --- not found in this scope
  │
```

Issue #8389, opened 2025-05-07
u128 division overflow in comptime


rkarabut:
### Aim

The program

```
fn main() -> pub u128 {
    comptime {
        340282366920938463463374607431768211455
    }
}
```

errors out with

```
error: The value `10944121435919637611123202872628637544444323383668486403580789397003788353536` cannot fit into `u128` which has range `0..=340282366920938463463374607431768211455`
  ┌─ src/main.nr:2:5
  │  
2 │ ╭     comptime {
3 │ │         340282366920938463463374607431768211455 / 2
4 │ │     }
  │ ╰─────'
  │  
  = Call stack:
    1. src/main.nr:2:5
```

### Expected Behavior

Should return 170141183460469231731687303715884105727.

### Bug

```
nargo execute --force
error: The value `10944121435919637611123202872628637544444323383668486403580789397003788353536` cannot fit into `u128` which has range `0..=340282366920938463463374607431768211455`
  ┌─ src/main.nr:2:5
  │  
2 │ ╭     comptime {
3 │ │         340282366920938463463374607431768211455 / 2
4 │ │     }
  │ ╰─────'
  │  
  = Call stack:
    1. src/main.nr:2:5
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

TomAFrench:
Looks like we're doing field division instead of integer division here. I've tracked down the root cause.

jfecher:
After github.com/noir-lang/noir/pull/8970 the code in this snippet now errors:
```
error: expected type u128, found type Field
   ┌─ src/main.nr:13:18
   │  
13 │   fn main() -> pub u128 {
   │                    ---- expected u128 because of return type
14 │ ╭     comptime {
15 │ │         340282366920938463463374607431768211455 / 2
16 │ │     }
   │ ╰─────' Field returned here
   │  
```
We still do field division but the compiler is more strict in asserting the output stays a field. To get it working as intended you can add a `_u128` suffix to either integer literal to perform u128 division. It is a bit awkward (the compiler should be able to push down the correct return type) but the miscompilation is fixed at least.

PR #8970, opened 2025-06-18
feat: Implement type suffixes

### LINKED ISSUE #8759: Add integer type suffixes
### Problem

There's no way to specify on the integer literal itself the type of the integer literal. Workarounds include adding an unnecessary cast `-1 as i8` or creating a let binding `let x: i8 = -1;`

### Happy Case

We should enable integer type suffixes to write the type of the integer directly on the integer itself: `-1_i8`, `-1_i32`, `-1_Field`, etc.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

jfecher:
# Description

## Problem\*

Resolves #8759 
Resolves https://github.com/noir-lang/noir/issues/8968

## Summary\*

Does what it says on the tin. 90% of changes are .snap changes.

## Additional Context

To keep `nargo expand` tests working, we print out the type suffix on every integer literal there since they are normally not stored in the Hir.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/lexer/lexer.rs
```diff
@@ -403,32 +405,22 @@ impl<'a> Lexer<'a> {
...
-        // We're not concerned with an underscore at the beginning of a decimal literal
-        // such as `_5` as this would be lexed into an ident rather than an integer literal.
-        let invalid_underscore_location = integer_str.ends_with('_');
-        let consecutive_underscores = integer_str.contains("__");
-        if invalid_underscore_location || consecutive_underscores {
-            return Err(LexerErrorKind::InvalidIntegerLiteral {
-                location: self.location(Span::inclusive(start, end)),
-                found: integer_str,
-            });
-        }
```
asterite, line 428: I guess we lost that? Though checking Rust they are more permissive than this (they do allow double underscore) so this is probably fine.

compiler/noirc_frontend/src/lexer/lexer.rs
```diff
@@ -403,32 +405,22 @@ impl<'a> Lexer<'a> {
...
-
         // Underscores needs to be stripped out before the literal can be converted to a `FieldElement.
-        let integer_str = integer_str.replace('_', "");
+        let mut integer_str = integer_str.replace('_', "");
```
asterite: Really minor, but if you write "1_hello" then you get this error:

```noir
error: Invalid integer literal
  ┌─ src/main.nr:2:13
  │
2 │     let x = 1_hello;
  │             -------  1hello is not an integer
```

Note that it says "1hello" because the "_" were stripped. We could keep the original string and show that for errors.

(the error in Rust says that "_hello" is not a valid integer literal suffix, which is a bit nicer, but could be done in a follow-up PR is we wanted)

docs/docs/noir/concepts/data_types/integers.md
```diff
@@ -95,7 +97,7 @@ A similar error would happen with signed integers:
 ```rust
 fn main() -> i8 {
     let x: i8 = -118;
-    let y: i8 = -11;
+    let y = -11;
```
asterite, line 100: Was this type removed for a reason?

asterite:
This PR also fixes some past inconsistencies. For example this code:

```noir
fn main() {
    let x = comptime {
        let x: i8 = -1;
        x
    };
    println(x);
}
```

used to print "0x30644e72e131a029b85045b68181585d2833e84879b9709143e1f593f0000000" because the value that came out of the comptime block was just "1" without any suffix, but now it's "1_u8".

On the other hand, this code used to compile:

```noir
fn foo() -> Field {
    comptime {
        let x: i8 = -1;
        x
    }
}
```

but now it doesn't, so this PR could be considered a breaking change (but it's fine: this is a bug fix).

All of these seem unrelated to the CI failures, though! It must be something with numeric kinds getting a type suffix or something...

compiler/noirc_frontend/src/lexer/lexer.rs
jfecher: Yeah, I found this and couldn't remember us ever wanting such a check. I went back to the PR it was added in and couldn't find justification for it either. Since rust doesn't do this check and it made adding suffixes more difficult I just removed it.

docs/docs/noir/concepts/data_types/integers.md
jfecher: I went through our docs to ensure we showed the default polymorphic-ness of integers better. In this case `x` already has a type and `y` is used with it so I wanted to show it isn't required.

jfecher:
> On the other hand, this code used to compile

Yeah, locally the error I'm seeing is with `comptime_return_minus_one_as_i32`:
```
fn main() -> pub i32 {
    comptime {
        -1
    }
}
```
This used to compile but now actually errors with Field != i32

asterite:
The errors on CI seems to be related to this not compiling:

```noir
fn main() {
    let _ = [0; 10_u32];
}
```

nor this:

```noir
fn main() {
    let _: [Field; 10_u32] = [0; 10];
}
```

(sorry if you were already investigating that, I got curious 😊)

asterite:
The error you mention is also related to this issue: https://github.com/noir-lang/noir/issues/8389

I wonder if we could pass down the expected type of the `comptime` block. 

In this code:

```noir
fn main() -> pub i32 {
    comptime {
        -1
    }
}
```

we could pass the type `i32` to the comptime block, which would act like its return type. Then `-1` would get the type `i32`.

It would be similar in this code:

```noir
fn main() -> pub i32 {
    comptime {
        let x = -1;
        x
    }
}
```

I think we are already passing the target type in the elaborator, but not using it in all cases.

That said... I think comptime blocks are eagerly evaluated before the function finishes so in something like this it might not work:

```noir
fn main() -> pub i32 {
    let y = comptime {
        let x = -1;
        x
    };
    y
}
```

Just an idea :-)

jfecher:
Found another failure:
```noir
    trait Serialize<let N: u32> {
        fn serialize() {}
    }

    #[attr]
    pub fn foobar() {}

    comptime fn attr(_f: FunctionDefinition) -> Quoted {
        let serialized_len = 1;
        // We are testing that when we unquote $serialized_len, it's unquoted
        // as the token `1` and not as something else that later won't be parsed correctly
        // in the context of a generic argument.
        quote {
            impl Serialize<$serialized_len> for Field {
                fn serialize() { }
            }
        }
    }

    fn main() {}
```
This now fails where we substitute `let serialized_len = 1;` in for `impl Serialize<$serialized_len> for Field {` because Serialize expects a u32 and now with the type suffix it is seeing `1_Field`

Edit: Actually we still get the same error even with specifying "_u32"

aakoshh:
Could you update the [monomorphized AST printer](https://github.com/noir-lang/noir/blob/f05c1f3cae0bf2b5658337b1258ed018f5926626/compiler/noirc_frontend/src/monomorphization/printer.rs#L256-L258) as well? 🙏

asterite:
Maybe if we don't carry the suffix out of comptime code then it would work? We'd lose the fixes I mentioned above, but that's how it was working anyway.

jfecher:
@asterite yeah, it may be worth reverting that portion to get this change in. The bug in that case would still be logged to be fixed later.

compiler/noirc_frontend/src/lexer/lexer.rs
jfecher: fixed :slightly_smiling_face:

jfecher:
PR should be ready for review - with the `into_tokens` revert and some snap file updates all tests are now passing

test_programs/compile_success_no_bug/comptime_right_shift_no_overflow/src/main.nr
```diff
@@ -1,4 +1,4 @@
 fn main(x: Field) -> pub Field {
     let y = comptime { 1 >> 32 };
-    x + y
+    x + y as Field
```
asterite, line 3: Can this be reverted?

test_programs/compile_success_no_bug/noirc_frontend_tests_metaprogramming_unquoted_integer_as_integer_token/src/main.nr
```diff
@@ -7,7 +7,7 @@
     pub fn foobar() {}
 
     comptime fn attr(_f: FunctionDefinition) -> Quoted {
-        let serialized_len = 1;
+        let serialized_len = 1u32;
```
asterite: Can this too be reverted? Mainly to make sure that it works the way it used to work.

asterite (approved):
Looks great! I left a couple of comments.

test_programs/compile_success_no_bug/comptime_right_shift_no_overflow/src/main.nr
jfecher: No - looks like we still remember `y` is a `u32` in this case.
Imo this is correct behavior anyway as the result of a right shift operation

test_programs/compile_success_no_bug/noirc_frontend_tests_metaprogramming_unquoted_integer_as_integer_token/src/main.nr
jfecher: Reverted

PR #8977, opened 2025-06-19
fix(expand): show references to ModuleDefId recursing on parents

asterite:
# Description

## Problem

Part of https://github.com/noir-lang/noir/issues/8281

## Summary

`nargo expand` didn't correctly produce references to items in some cases. To do this it relied on a function that didn't have access to the expander (a function used mainly for LSP) so it didn't know about existing imports, etc. Actually, in the end some logic was moved to be done more generally (previously it was only done if the item to reference wasn't visible).

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/nargo_cli/src/cli/expand_cmd/printer.rs
+++ b/tooling/nargo_cli/src/cli/expand_cmd/printer.rs
@@ -1074,10 +1074,7 @@ impl<'context, 'string> ItemPrinter<'context, 'string> {
         if module_def_id_parent_module != Some(self.module_id) {
             if let Some(module_def_id_parent_module) = module_def_id_parent_module {
                 let visibility = self
-                    .interner
-                    .try_module_attributes(&module_def_id_parent_module)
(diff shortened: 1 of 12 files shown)


tooling/nargo_cli/src/cli/expand_cmd/printer.rs
```diff
@@ -1061,63 +1056,33 @@ impl<'context, 'string> ItemPrinter<'context, 'string> {
...
-                    if !full_path.is_empty() {
-                        self.push_str(&full_path);
-                        self.push_str("::");
-                    }
-                } else {
-                    self.push_str(&full_path);
-                    self.push_str("::");
-                }
+        // Recurse on the parent module, but only if the parent module isn't the current module
+        // (if so, we can already reach the definition just by priting its name)
```
TomAFrench:
```suggestion
        // (if so, we can already reach the definition just by printing its name)
```

TomAFrench (approved):

PR #8979, opened 2025-06-19
fix: add a remove_unreachable_instructions SSA pass

### LINKED ISSUE #8774: Compiler crash: Expected all load instructions to be removed before acir_gen

asterite:
# Description

## Problem

Resolves #8774

## Summary

Alternative to #8943, or maybe something to do in addition to that (unsure yet).

I _think_ the main issue in #8774 is that it ends up generating a `load` instruction for a value that can never happen, because a previous `constrain` is always false, and it's trying to get a value from an empty array. The solution here is to remove any instruction that comes after an "always fails" constrain in a block, also replacing the block's terminator with a zeroed value.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa.rs
+++ b/compiler/noirc_evaluator/src/ssa.rs
@@ -219,6 +219,7 @@ pub fn primary_passes(options: &SsaEvaluatorOptions) -> Vec<SsaPass> {
             Ssa::verify_no_dynamic_indices_to_references,
             "Verifying no dynamic array indices to reference value elements",
         ),
+        SsaPass::new(Ssa::remove_unreachable_instructions, "Remove unreachable instructions"),
     ]
 }
 
@@ -232,6 +233,7 @@ pub fn secondary_passes(brillig: &Brillig) -> Vec<SsaPass> {
         // In that case it's unused so we can remove it. This is what we check next.
         SsaPass::new(Ssa::remove_unreachable_functions, "Removing Unreachable Functions"),
         SsaPass::new(Ssa::dead_instruction_elimination_acir, "Dead Instruction Elimination - ACIR"),
+        SsaPass::new(Ssa::remove_unreachable_instructions, "Remove unreachable instructions"),
     ]
 }
 
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
@@ -1,0 +1,159 @@
+//! This module defines an SSA pass to remove instructions that are unreachable.
+//! For example, if an instruction in a block is `constrain u1 0 == u1 1`,
+//! any subsequent instructions in that block will never be executed. This pass
+//! then removes those subsequent instructions and replaces the block's terminator
+//! values with zeroed values of the appropriate type.
+use std::sync::Arc;
+
+use acvm::{AcirField, FieldElement};
+use noirc_errors::call_stack::CallStackId;
+
+use crate::ssa::{
+    ir::{
+        basic_block::BasicBlockId, function::Function, instruction::Instruction, types::Type,
+        value::ValueId,
+    },
+    ssa_gen::Ssa,
+};
+
+impl Ssa {
+    pub(crate) fn remove_unreachable_instructions(mut self) -> Ssa {
+        for function in self.functions.values_mut() {
+            function.remove_unreachable_instructions();
+        }
+        self
+    }
+}
+
+impl Function {
+    fn remove_unreachable_instructions(&mut self) {
+        let mut current_block_id = None;
+        let mut current_block_instructions_are_unreachable = false;
+        let mut blocks_with_unreachable_terminators = Vec::new();
+
+        self.simple_reachable_blocks_optimization(|context| {
+            let block_id = context.block_id;
+
+            if current_block_id.is_none() {
+                current_block_id = Some(block_id);
+            } else if current_block_id != Some(block_id) {
+                current_block_instructions_are_unreachable = false;
+                current_block_id = Some(block_id);
+            }
+
+            if current_block_instructions_are_unreachable {
+                context.remove_current_instruction();
+            }
+
+            let instruction = context.instruction();
+            match instruction {
+                Instruction::Constrain(lhs, rhs, _) => {
+                    let Some(lhs_constant) = context.dfg.get_numeric_constant(*lhs) else {
+                        return;
+                    };
+                    let Some(rhs_constant) = context.dfg.get_numeric_constant(*rhs) else {
+                        return;
+                    };
+                    if lhs_constant != rhs_constant {
+                        current_block_instructions_are_unreachable = true;
+                        blocks_with_unreachable_terminators.push(block_id);
+                    }
+                }
+                Instruction::ConstrainNotEqual(lhs, rhs, _) => {
+                    let Some(lhs_constant) = context.dfg.get_numeric_constant(*lhs) else {
+                        return;
+                    };
+                    let Some(rhs_constant) = context.dfg.get_numeric_constant(*rhs) else {
+                        return;
+                    };
+                    if lhs_constant == rhs_constant {
+                        current_block_instructions_are_unreachable = true;
+                        blocks_with_unreachable_terminators.push(block_id);
+                    }
+                }
+                _ => (),
+            }
+        });
+
+        for block_id in blocks_with_unreachable_terminators {
+            let mut terminator = self.dfg[block_id].take_terminator();
+            terminator.map_values_mut(|value_id| {
+                let typ = self.dfg.type_of_value(value_id);
+                zeroed_value(self, block_id, &typ)
+            });
+            self.dfg[block_id].set_terminator(terminator);
+        }
+    }
+}
+
+fn zeroed_value(function: &mut Function, block_id: BasicBlockId, typ: &Type) -> ValueId {
+    match typ {
+        Type::Numeric(numeric_type) => {
+            function.dfg.make_constant(FieldElement::zero(), *numeric_type)
+        }
+        Type::Array(element_types, len) => {
+            let mut array = im::Vector::new();
+            for _ in 0..*len {
+                for typ in element_types.iter() {
+                    array.push_back(zeroed_value(function, block_id, typ));
+                }
+            }
+            let instruction = Instruction::MakeArray { elements: array, typ: typ.clone() };
+            let stack = CallStackId::root();
+            function.dfg.insert_instruction_and_results(instruction, block_id, None, stack).first()
+        }
+        Type::Slice(_) => {
+            let array = im::Vector::new();
+            let instruction = Instruction::MakeArray { elements: array, typ: typ.clone() };
+            let stack = CallStackId::root();
+            function.dfg.insert_instruction_and_results(instruction, block_id, None, stack).first()
+        }
+        Type::Reference(element_type) => {
+            let instruction = Instruction::Allocate;
+            let reference_type = Type::Reference(Arc::new((**element_type).clone()));
+            function
+                .dfg
+                .insert_instruction_and_results(
+                    instruction,
+                    block_id,
+                    Some(vec![reference_type]),
+                    CallStackId::root(),
+                )
+                .first()
+        }
+        Type::Function => {
+            // We can have the function return itself. It's fine because the terminator is unreachable anyway.
+            function.dfg.import_function(function.id())
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use crate::{assert_ssa_snapshot, ssa::ssa_gen::Ssa};
+
+    #[test]
+    fn removes_unreachable_instructions() {
+        let src = r#"
+        acir(inline) predicate_pure fn main f0 {
+          b0():
+            v0 = make_array [] : [&mut u1; 0]
+            constrain u1 0 == u1 1, "Index out of bounds"
+            v4 = array_get v0, index u32 0 -> &mut u1
+            v5 = load v4 -> u1
+            return v5
+        }
+        "#;
+        let ssa = Ssa::from_str(src).unwrap();
+        let ssa = ssa.remove_unreachable_instructions();
+
+        assert_ssa_snapshot!(ssa, @r#"
+        acir(inline) predicate_pure fn main f0 {
+          b0():
+            v0 = make_array [] : [&mut u1; 0]
+            constrain u1 0 == u1 1, "Index out of bounds"
+            return u1 0
+        }
+        "#);
+    }
+}
--- a/compiler/noirc_evaluator/src/ssa/opt/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/mod.rs
@@ -31,6 +31,7 @@ mod remove_enable_side_effects;
 mod remove_if_else;
 mod remove_truncate_after_range_check;
 mod remove_unreachable;
+mod remove_unreachable_instructions;
 mod simple_optimization;
 mod simplify_cfg;
 mod unrolling;
--- a/test_programs/compile_success_with_bug/regression_8774/Nargo.toml
+++ b/test_programs/compile_success_with_bug/regression_8774/Nargo.toml
@@ -1,0 +1,6 @@
+[package]
+name = "regression_8774"
+type = "bin"
+authors = [""]
+
+[dependencies]
--- a/test_programs/compile_success_with_bug/regression_8774/src/main.nr
+++ b/test_programs/compile_success_with_bug/regression_8774/src/main.nr
@@ -1,0 +1,4 @@
+fn main() -> pub bool {
+    let d: [&mut bool; 0] = [];
+    *d[0]
+}
--- a/tooling/nargo_cli/tests/snapshots/compile_success_with_bug/regression_8774/execute__tests__expanded.snap
+++ b/tooling/nargo_cli/tests/snapshots/compile_success_with_bug/regression_8774/execute__tests__expanded.snap
@@ -1,0 +1,8 @@
+---
+source: tooling/nargo_cli/tests/execute.rs
+expression: expanded_code
+---
+fn main() -> pub bool {
+    let d: [&mut bool; 0] = [];
+    *d[0]
+}


asterite:
This is trickier than what I initially thought.

It's easy to remove all instructions after we find a constrain that'll always fail. It's also easy to zero that block's terminator.

But then we need to consider that block's successors as unreachable too, but if they can only be reached from that block. And this is the part that's hard to implement.

compiler/noirc_evaluator/src/ssa/ir/function.rs
```diff
@@ -192,6 +193,23 @@ impl Function {
         blocks
     }
 
+    /// Collects all the reachable blocks of this function and returns them in pre-order.
```
aakoshh: Are you sure we don't already have somewhere something like this? 

In any case, it would be good to spell out what we mean by _pre-order_. 

Say we have a graph like this:
```
b0 - b1 ---------- b5
   \              /
     b2 - b3 - b4 
```
It looks to me like the code below will yield `[b0, b1, b2, b5, b3, b4]`. Since `b5` comes before some of its predecessors, I'm not sure it can be considered pre-order, which I believe generally means "the parent node is visited before its children".

Maybe some tests would be helpful.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -0,0 +1,430 @@
...
+/// Adds all of a block's successors to the `unreachable_blocks` set if they are dominated by that block.
+fn add_successors(
+    block_id: BasicBlockId,
+    dfg: &DataFlowGraph,
+    dom: &mut DominatorTree,
+    unreachable_blocks: &mut HashSet<BasicBlockId>,
+) {
+    // First compute the set of all successors
+    let mut all_successors = HashSet::default();
+    all_successors.insert(block_id);
```
aakoshh: Why do we consider _self_ a successor?

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -0,0 +1,430 @@
...
+                let typ = self.dfg.type_of_value(value_id);
+                zeroed_value(self, block_id, &typ)
+            });
+            self.dfg[block_id].set_terminator(terminator);
+        }
+    }
+}
+
+/// Adds all of a block's successors to the `unreachable_blocks` set if they are dominated by that block.
+fn add_successors(
```
aakoshh: It's not obvious from this description that the algorithm will visit all "descendants" and not just the immediate successors. Maybe you could break it up into `blocks_reachable_from` and then another one to add dominated blocks to the unreachable set. Or `blocks_dominated_by` to return the list of all dominated blocks in the graph, with a second step to add them to the unreachable?

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -0,0 +1,430 @@
...
+          b2():
+            jmp b1()
+          b3():
+            return Field 0
+        }
+        "#);
+    }
+
+    #[test]
+    fn removes_unreachable_instructions_following_block_with_no_instructions() {
```
aakoshh, line 406: Why was a block without instructions considered a special case to be tested? Isn't `b1` in the previous test also without instructions?

compiler/noirc_evaluator/src/ssa/ir/function.rs
aakoshh: In `mem2reg` its:
```rust
        // Iterate each block in reverse post order = forward order
        let mut block_order = PostOrder::with_function(self.inserter.function).into_vec();
        block_order.reverse();
```
Would that be what we're looking for?

In `PostOrder` the blocks are returned in an order such that each block appears before any of its predecessors do. When we reverse that, we get an order in which every block appears before any of its successors. In the above DAG it would mean `b5` appears first in `PostOrder` and last in its reverse.

compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -212,6 +212,7 @@ pub fn primary_passes(options: &SsaEvaluatorOptions) -> Vec<SsaPass> {
         SsaPass::new(Ssa::brillig_array_get_and_set, "Brillig Array Get and Set Optimizations"),
         // Perform another DIE pass to update the used globals after offsetting Brillig indexes.
         SsaPass::new(Ssa::dead_instruction_elimination, "Dead Instruction Elimination"),
+        SsaPass::new(Ssa::remove_unreachable_instructions, "Remove Unreachable Instructions"),
```
aakoshh, line 217: Can we do this higher up? Maybe start with it, since the compiler already inserts some failures where we know something overflows at compile time. 

What else can introduce constant constraints? Maybe `mem2reg` can turn something into a known value.

compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -212,6 +212,7 @@ pub fn primary_passes(options: &SsaEvaluatorOptions) -> Vec<SsaPass> {
...
         // A function can be potentially unreachable post-DIE if all calls to that function were removed.
```
aakoshh:
```suggestion
        // A function can be potentially unreachable post-DIE if all calls to that function were removed, 
        // or after the removal of unreachable instructions.
```

aakoshh (commented):
This feels like a breath of fresh air, to clear out the SSA and never have to think about it again 😌 👏

aakoshh:
Just a note that the reason the integration test did not fail with `--minimal-ssa`, which doesn't have this pass, is because this problem only affected ACIR, and `--minimal-ssa` implies `--force-brillig`.

compiler/noirc_evaluator/src/ssa/ir/function.rs
asterite: You are right! Yes, I saw we had PostOrder, looked up graph traversals, a question about post and pre order showed up and I must have looked at it wrong and concluded this is pre-order. I think what I implemented is BFS (Breadth-first search). I wonder if the above code returns the same thing as what I implemented, though (because the logic of `PostOrder` seems a bit more complex than that).

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: Ah, this is a left-over of another thing I was trying before this.

compiler/noirc_evaluator/src/ssa/ir/function.rs
aakoshh: I don't think it will return the same thing you implemented, you're right yours is BFS. The `PostOrder` looks like an elegant way of ensuring all descendants are already done before their parents, so I _think_ if you reverse it you get what you want.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: It was initially considered a special case because originally I only added immediate successors of a block when it was determined it had some unreachable instructions. Then when reaching the successor I'd add it's successors, etc. The issue was that `simple_reachable_blocks_optimizations` never visits blocks which have no instructions.

To be honest, most of the tests might be redundant and exist because I found things that didn't work as I implemented them. Maybe I can just name them 1, 2, 3, as just interesting cases, and add a comment at the top of each explaining what's interesting about them?

compiler/noirc_evaluator/src/ssa.rs
asterite: I think for example if you do `array_get` of an empty array, that constrain is added. Not sure if that can happen later on in the pipeline. But it's a good point, if it can be done earlier it would theoretically speed up things later on.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -0,0 +1,430 @@
...
+          b1():
+            jmp b2()
+          b2():
+            return Field 0
+        }
+        "#);
+    }
+
+    #[test]
+    fn does_not_zeroes_terminator_of_previously_seen_blocks() {
```
aakoshh:
```suggestion
    fn does_not_zeroes_terminator_of_non_dominated_blocks() {
```

I commented on this but it got lost: I think "previously seen" assumes we are much familiar with the traversal algorithm, rather than stating an invariant. For example here I think we will visit blocks in the order of `b0, b1, b2, b3, b4`; I guess the test name highlights that we won't visit `b1` again because we have already seen it, and therefore it keeps its `add`, but `b4` also keeps the `add`, despite being visited after `b3`. The real reason both get to keep their instructions is they are both not dominated by `b3`. 

I realise this must be one of the tests you created earlier, just wanted to point it out that it could be renamed.

compiler/noirc_evaluator/src/ssa/ir/function.rs
asterite: In the end the `simple_reachable_blocks_optimization` logic was good for this PR too. I think I had problems before understanding we need to use dominance to implement this. So now the order of blocks traversal doesn't matter much as long as it goes in a way that if we visit a block, then blocks dominated by it will come next (this is guaranteed by the current logic though it's not explicitly documented).

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: I made a few renames here.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: I ended up renaming this method.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: I ended up renaming some tests and adding some comments to them. I left all of them just in case, it doesn't hurt to have a few more tests on this 😊

compiler/noirc_evaluator/src/ssa.rs
asterite: Done!

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: I ended up renaming this test.

compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -913,10 +913,14 @@ impl<'a> Context<'a> {
         terminator: &TerminatorInstruction,
         dfg: &DataFlowGraph,
     ) -> usize {
+        let no_return_values = Vec::new();
         let return_values = match terminator {
             TerminatorInstruction::Return { return_values, .. } => return_values,
+            TerminatorInstruction::Unreachable { .. } => &no_return_values,
```
aakoshh: Why not avoid the allocation of a vector and return 0 here?

compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -930,12 +934,16 @@ impl<'a> Context<'a> {
...
         let (return_values, call_stack) = match terminator {
             TerminatorInstruction::Return { return_values, call_stack } => {
                 (return_values, *call_stack)
             }
             // TODO(https://github.com/noir-lang/noir/issues/4616): Enable recursion on foldable/non-inlined ACIR functions
-            _ => unreachable!("ICE: Program must have a singular return"),
+            TerminatorInstruction::JmpIf { .. } | TerminatorInstruction::Jmp { .. } => {
+                unreachable!("ICE: Program must have a singular return")
+            }
+            TerminatorInstruction::Unreachable { call_stack } => (&no_return_values, *call_stack),
```
aakoshh: We know that we will return `(vec![], vec![])` from the method, no?

compiler/noirc_evaluator/src/ssa/opt/die/prune_dead_parameters.rs
```diff
@@ -128,6 +128,9 @@ impl Function {
                     TerminatorInstruction::Return { .. } => {
                         unreachable!("ICE: A return block should not be a predecessor");
                     }
+                    TerminatorInstruction::Unreachable { .. } => {
+                        unreachable!("ICE: An unreachable block should not be a predecessor");
```
aakoshh: Is there something now to prune unreachable blocks?

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -0,0 +1,461 @@
+//! This module defines an SSA pass to remove instructions that are unreachable.
+//! For example, if an instruction in a block is `constrain u1 0 == u1 1`,
+//! any subsequent instructions in that block will never be executed. This pass
+//! then removes those subsequent instructions and replaces the block's terminator
+//! values with zeroed values of the appropriate type. If the block has successors
```
aakoshh, line 5:
```suggestion
//! then removes those subsequent instructions and replaces the block's terminator
//! with a special `unreachable` value. If the block has successors
```

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -0,0 +1,461 @@
...
+        // The current block we are currently processing
+        let mut current_block_id = None;
+
+        // Whether the current block instructions were determined to be unreachable
+        let mut current_block_instructions_are_unreachable = false;
+
+        // Blocks that can't be reached because their dominator has an always failing instruction.
+        let mut unreachable_blocks = HashSet::default();
+
+        self.simple_reachable_blocks_optimization(|context| {
```
aakoshh: Convoluted example, but here's a setup which I think the `reachable_blocks` implementation wouldn't handle:

```

   
           b0
         /    \ 
        b1     b2*
        |      |
        b3*    |
         \    /
           b4
```
What if you know that both `b3` and `b2` contain always failing instructions? You could remove `b4`, if the definition was "all predecessors are unreachable". It wouldn't have to do anything with the dominator tree, but it would need to be based on the "reverse post-order", so by the time you consider a block, all its predecessors have been marked.

compiler/noirc_evaluator/src/ssa/opt/die/prune_dead_parameters.rs
asterite: I think no. Based on the comment on "return", the logic here should happen on a predecessor of a block, and `unreachable` can't be a predecessor of any block.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -0,0 +1,461 @@
...
+            };
+
+            if always_failing_instruction {
+                current_block_instructions_are_unreachable = true;
+
+                add_dominated_blocks(block_id, context.dfg, &mut dom, &mut unreachable_blocks);
+                let terminator = context.dfg[block_id].take_terminator();
+                let call_stack = terminator.call_stack();
+                context.dfg[block_id]
+                    .set_terminator(TerminatorInstruction::Unreachable { call_stack });
```
aakoshh: You will do this for each instruction in the block. Maybe it can be done only if there was a switch?

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -0,0 +1,461 @@
...
+            }
+        });
```
aakoshh, line 109: At the end, should all unreachable non-entry blocks be removed? The CFG will never reach them, after all, since the structure of the graph will have been forgotten.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: I just checked and no, because we set `current_block_instructions_are_unreachable = true`, then this code above will be hit first:

```rust
            if current_block_instructions_are_unreachable {
                context.remove_current_instruction();
                return;
            }
```

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -0,0 +1,461 @@
...
+        self.simple_reachable_blocks_optimization(|context| {
+            let block_id = context.block_id;
+
+            if current_block_id != Some(block_id) {
+                current_block_id = Some(block_id);
+                current_block_instructions_are_unreachable = unreachable_blocks.contains(&block_id);
+            }
+
+            if current_block_instructions_are_unreachable {
+                context.remove_current_instruction();
```
aakoshh, line 77: If we already know the block is unreachable, because it was marked by `add_dominated_blocks`, then this will return without changing the block's terminator to `unreachable`. Not sure if that is intended, or only the first such block should be terminated like that. If the ones added (but not the one that triggered it) would be removed then it wouldn't matter.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: I think there's no way to remove blocks. They are stored in a dense map in the DFG.

compiler/noirc_evaluator/src/ssa.rs
aakoshh: I think we should do at at least once at the top as well.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: It's kind of intended. If we find the block is unreachable then it won't be reached by any other block so it will not show up in the final SSA (though it will still remain in the DFG, but all blocks do).

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: This is a good point! I added a test for this. It ends up working because no block can reach `b4` so it doesn't show up in the final SSA.

compiler/noirc_evaluator/src/ssa/opt/die/prune_dead_parameters.rs
aakoshh: Yeah, I didn't see anything. I know that this won't be hit, I was just wondering whether replacing `Jmp` and `JmpIf` with this will mean we'll have blocks that we can't reach, and should drop, or is it okay to leave them hanging there. Maybe they naturally get removed by something.

compiler/noirc_evaluator/src/ssa.rs
asterite: You mean right before the very first `remove_unreachable_functions`?

asterite: It could be worth it, but I'm not sure. For example if we have this:

```noir
fn main() -> pub bool {
    foo();
    true
}

fn foo() {
    let x: [u32; 0] = [];
    let _ = x[0];
}
```

then `foo` will get `unreachable`, but it's only after we inline `foo` inside `main` that `main` will get an `unreachable`... so maybe doing it in the end is the same.

The main goal of this pass is to avoid getting to acir/brillig with instructions that can't happen, like the `load` that was crashing the compiler. That's why I put it almost right next to the end of the pipeline. But... it could be done earlier too for performance reasons, though I don't know if it will make things slower if we do it near the beginning.

compiler/noirc_evaluator/src/acir/mod.rs
asterite: Done!

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: Done

compiler/noirc_evaluator/src/ssa.rs
aakoshh: I linked an issue where the initial SSA already contains a known overflow with an always failing constraint , and another where the compiler knows but doesn’t currently generate a constraint. 

Removing such instructions ASAP would make all the other passes disappear, saving a lot of work in fuzzing. 

You are right that in normal code it would probably just be another step that mostly doesn’t do anything

asterite: I just tried it and got this:

```
Message:  Expected block to have a terminator
Location: compiler/noirc_evaluator/src/ssa/ir/basic_block.rs:121
```

I'll debug it a bit more, not sure why it would panic...

asterite: I think it's inlining that doesn't work well now. That is, the logic of inlining a function or block that is unreachable.

asterite: Yup, it's inlining. To make it work we'd need to handle unreachable blocks, which currently were not very well handled.

I actually don't know how to handle this case so for now I put a panic there. Actually, I'm sure if I look at it a bit more I can figure it out, but maybe it would be better to do it in a follow-up PR. What do you think?

aakoshh: I wasn’t sure whether this new terminator is something that any subsequent code generation can handle properly. I liked the elegance of returning default values from unreachable blocks, like Maxim generated dummy functions in defunctionalize, because they won’t be called but also don’t need anyone to handle special cases. This approach looks more invasive, obviously. 

I think if tests are panicking it means we introduced something that isn’t well specified, so we should probably spend more time on it up front. 

I’d also ask Jake and the others what they think

asterite: I'll try to make it work in this PR then.

aakoshh: Would the previous approach with the defaulta, or the one where each unreachable, not just the dominated, gets emptied, work better?

asterite: I think so... but I really like the `unreachable` thing 😊

aakoshh: I can imagine the code that collects all blocks in a loop is not prepared for a CFG that never returns to the header. With the defaults it would, hopefully without losing track of indexes.

asterite: I guess not introducing a new `unreachable` terminator is also less risky regarding introducing bugs. I'll revert to the commit that worked like that then.

asterite: Done!

aakoshh: I think GitHub may have rejected your last push, at least I can't see any revert.

asterite: Ah, yes, I said "Done" but I forgot to push 😅

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -0,0 +1,508 @@
...
+
+    // Now add them to `unreachable_blocks` if they are dominated by the block
+    for reachable_block in reachable_blocks {
+        if dom.dominates(block_id, reachable_block) {
+            unreachable_blocks.insert(reachable_block);
+        }
+    }
+}
+
+fn zeroed_value(function: &mut Function, block_id: BasicBlockId, typ: &Type) -> ValueId {
```
aakoshh, line 134: I forgot how much work it was to produce this zeroed value, I thought it was supported out of the box. 

Nicely done! I had a thought that if you wanted to just pretend to return a value, then you could insert something like this:
```
v0 = allocate -> &mut <return-type>
v1 = load v0 -> <return-type>
return v1
```
Since the block is unreachable, it should never actually load anything, but AFAIK references are typeless. This is a bug the `mem2reg` pass often introduced. Not sure if anything would choke on it. 

In any case, your construction looks great!

asterite:
Hmmmmmm....

Now I want to go back to using `unreachable` for these reasons:
1. The case you showed where you had two branches with always failing instructions jumping to a same block no longer "works". That is, both blocks jump to that block and continue from there, instead of that block being removed. That said, it doesn't cause any issues... I tried this code:
    ```
    fn main(x: bool) -> pub bool {
        let z = if x {
            let d: [&mut bool; 0] = [];
            d[0]
        } else {
            let d: [&mut bool; 0] = [];
            d[1]
        };
        *z
    }
    ```
    My intention was to load a value in a branch that is unreachable because all of its predecessors are unreachable. But this fails to compile with "Cannot return a reference type from an if or match expression". That's why it's not an issue... but stil..
2. If we don't do this new pass right at the beginning, this PR still fixes #8994
3. I think with `unreachable` this led to better performance.

With the above points, I feel it's okay to do this pass almost near the end, and assume unreachable terminators don't happen in previous passes (panic if they do). And we can always try to do this pass right at the beginning while also handling unreachable in the inlining pass as a follow-up PR if we wanted to.

What do you think?

aakoshh:
I would try to order blocks in reverse post order and check if all predecessors have been marked as unreachable.

asterite:
I'll open a separate PR with the `unreachable` terminator so we can more easily see the differences.

aakoshh:
Otherwise I feel you are erecting barriers to where this can be executed based on issues that are down to preferring one approach over another

asterite:
> I would try to order blocks in reverse post order and check if all predecessors have been marked as unreachable.

Ah, this is a good one. I'll try it.

asterite:
In any case, `arb_ssa_roundtrip` is failing with a panic that I don't understand yet, but it goes away if we don't run this new pass as the first pass in the pipeline. I'll look at it tomorrow...

asterite:
This code seems to crash the compiler now:

```noir
fn main() {
    unsafe { func_2(true) };
}

    loop {
        if cond {
            break
        }

        assert(false);

        loop {
            if cond {
                break
            }
        }
    }
}
```

We end up with this SSA after purity analysis, then it crashes:

```
  b0():
    call f1(u1 1)
    return
}
  b0(v0: u1):
    jmpif v0 then: b1, else: b2
  b1():
    return
  b2():
    constrain u1 0 == u1 1
    jmp b3()
  b3():
    jmp b3()
}
```

The error is:

```
Message:  internal error: entered unreachable code: Post order for unreachable block is undefined
Location: compiler/noirc_evaluator/src/ssa/ir/dom.rs:79
```

Not sure why it errors... maybe because it keeps looping forever in `b3`?

I could try to fix it, but... I really feel like having an `unreachable` terminator is the more correct way to do this. I understand that it doesn't work if we run that pass first in the pipeline, but it doesn't work either without `unreachable`. Then, my idea with this pass was to run it right before acir/brillig so that those don't have to deal with undefined values (I think in LLVM these are poison values, not sure!). It leads to cleaner SSA and I think the only thing missing would be to handle these during inlining, but it could be done later on, if really needed.

Also, with `unreachable` we can determine that a function never returns, if it doesn't have any `return`. This could happen if a branch with a `return` has an instruction that always fails, then its terminator is replaced with `unreachable`. Then when inlining that function into other functions we can know that calls to it are unreachable and keep pruning code. If we use zeroed values we can't do that.

It also aligns with at least LLVM, which has an unreachable terminator: https://llvm.org/docs/LangRef.html#unreachable-instruction

So I feel like reverting back to [6e4e138](https://github.com/noir-lang/noir/pull/8979/commits/6e4e138c825adbf90344a3d3e3e61098d04a0bf3)

aakoshh:
I'll have a look.

aakoshh:
Right, because of the zeroed values, what starts out as this:
```
  b5(): 
    jmpif v0 then: b7, else: b8 // The 2nd  `if cond { break }`
```
is replaced with this:
```
  b5():
    jmpif u1 0 then: b7, else: b8
```
and a subsequent step optimises out the `break` and now we have a tight loop, which during LICM breaks the dominator tree. 

I'll try if my earlier suggestion of using an undefined reference could prevent these optimisations, but this crash can be induced without your pass, so I'd consider it a standalone bug as well that needs to be investigated. I'll create a ticket about that.

aakoshh:
I hear what you are saying about focusing on fixing the ACIR generation, and not leaving an SSA that defies expectations. Thanks for pointing out that LLVM has `unreachable` as well. 

I do wonder though: if we just want to make sure ACIR generation doesn't go into undefined territory, whether the detection of always-failing constraints should be placed there, instead of preparing an SSA that is difficult for other passes to process. That is, to quit producing ACIR for a block after the constraint has been encountered.

aakoshh:
Your example program that crashed above works now:

```rust
After Remove Unreachable Instructions (1) (step 1):
acir(inline) fn main f0 {
  b0():
    call f1(u1 1)
    return
}
brillig(inline) fn func_2 f1 {
  b0(v0: u1):
    jmp b1()
  b1():
    jmpif v0 then: b3, else: b4
  b2():
    return
  b3():
    jmp b2()
  b4():
    constrain u1 0 == u1 1
    jmp b5()
  b5():
    v3 = allocate -> &mut u1
    v4 = load v3 -> u1
    jmpif v4 then: b7, else: b8
  b6():
    jmp b1()
  b7():
    jmp b6()
  b8():
    jmp b5()
}
...
After Dead Instruction Elimination - ACIR (1) (step 48):
  b0():
    return
}

```

asterite:
> That is, to quit producing ACIR for a block after the constraint has been encountered.

We could do that. But then if in Brillig we also have a case of not wanting some instructions past an always failing instruction we'd have to do the same logic. In my mind SSA should be prepared so that ACIR and brillig do the minimal thing.

I've been also thinking that with `unreachable` the logic of this pass is simpler than what I wrote: when we change a block's terminator to `unreachable` then automatically all dominated blocks become unreachable... because they won't be reached anymore! So we don't need to do this with the dom.

I think I'll open a separate PR with the `unreachable` logic, as I mentioned before, but I'll try again to make it work with inlining. It shouldn't be that hard: if we try to inline a call that has no returns then it's unreachable. Then we should stop inserting instructions past that call, change the terminator to unreachable... but then we'd need to avoid visiting dominated blocks, and I think here we'd need to use the Dom type. I'll check :-)

aakoshh:
> I've been also thinking that with unreachable the logic of this pass is simpler than what I wrote: when we change a block's terminator to unreachable then automatically all dominated blocks become unreachable... because they won't be reached anymore! 

Having seen both, I don't know this pass will be that much simpler; it wasn't easy to spot the differences, as you remember my questions. But your fantastic test coverage provides examples that cover the whole range 👍 

The flip side of any simplification here is potentially making other passes more complex, because we have to think about a loop that starts but never finishes because one of its blocks is has an unreachable terminator. 

Definitely do explore this option! I just wanted to avoid yolo'ing into it without updating the other passes to handle it.

asterite:
> Definitely do explore this option! I just wanted to avoid yolo'ing into it without updating the other passes to handle it.

And you are still not in favor of introducing the `unreachable` terminator in a pass near the end, avoiding the need to handle it during inlining, and having that be done in a later PR?

aakoshh:
Well, having worked on this for a while now, I came around to your position that having an explicit `unreachable` would be better than what I'm doing now, which is to try to construct an SSA that doesn't break anything, because it seems very tricky to do so.

I'd ask around whether everyone is on board with having it left panicky in earlier passes.

aakoshh:
I found the instance where unrolling doesn't work.

Say we start with this:
```rust
  b13():
    constrain u1 0 == u1 1, "OSA"
    jmp b15(u32 1338103543)
  b15(v34: u32):
    v68 = lt v34, u32 1338103546
    jmpif v68 then: b16, else: b17
```
after this pass, it becomes:
```rust
  b13():
    constrain u1 0 == u1 1, "OSA"
    jmp b15(u32 1338103543)

  b15(v34: u32):
    v67 = allocate -> &mut u1
    v68 = load v67 -> u1
    jmpif v68 then: b16, else: b17
```
But unrolling expects to find an `lt` and look at its arguments to find the upper bound.

asterite:
Closed in favor of #9008

PR #8987, opened 2025-06-20
chore: remove unreachable functions after inlining simple functions

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

In aztec-packages we have a glut of small functions which hang around in the SSA for a long time after being inlined. This PR integrates the `remove_unreachable_functions` pass into the `inline_simple_functions` pass so we can remove these asap.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/inline_simple_functions.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/inline_simple_functions.rs
@@ -71,7 +71,7 @@ impl Ssa {
             (*id, function.inlined(&self, &should_inline_call))
         });
 
-        self
+        self.remove_unreachable_functions()
     }
 }
 
@@ -112,10 +112,6 @@ mod test {
           b0(v0: Field):
             v1 = add v0, v0
             return v1
-        }
-        acir(inline) fn foo f1 {
-          b0(v0: Field):
-            return v0
         }
         ");
     }
@@ -160,14 +156,10 @@ mod test {
         assert_ssa_snapshot!(&mut ssa, @r"
         acir(inline) fn main f0 {
           b0(v0: Field):
-            v2 = call f2(v0) -> Field
+            v2 = call f1(v0) -> Field
             return v2
         }
-        acir(inline) fn foo f1 {
-          b0(v0: Field):
-            return v0
-        }
-        acir(inline) fn bar f2 {
+        acir(inline) fn bar f1 {
           b0(v0: Field):
             v1 = add v0, v0
             return v1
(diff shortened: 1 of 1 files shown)


TomAFrench:
huh?

TomAFrench:
Ah, looks like this affects the weights applied to different functions as the simple functions which are now removed would have still contributed to the weight of the functions which they called.

compiler/noirc_evaluator/src/ssa/opt/inline_simple_functions.rs
```diff
@@ -71,7 +71,7 @@ impl Ssa {
             (*id, function.inlined(&self, &should_inline_call))
         });
 
-        self
+        self.remove_unreachable_functions()
```
aakoshh: Why not add it as a separate pass?

TomAFrench: This ensures that it'll always be applied after the inlining and avoids printing a whole extra pass in the `show-ssa` output.

TomAFrench: We have some examples of running simple passes within another pass to clean up the output already such as in https://github.com/noir-lang/noir/pull/8895. Also the preprocess functions pass is similar to this.

aakoshh: Yes, the preprocess is a collection of other passes. 

On the other hand the `remove_unreachable_functions` appears in the primary pipeline with comments [here](https://github.com/noir-lang/noir/blob/256c67899ade021c0140a204dd00801eaebda055/compiler/noirc_evaluator/src/ssa.rs#L201-L202) and [here](https://github.com/noir-lang/noir/blob/256c67899ade021c0140a204dd00801eaebda055/compiler/noirc_evaluator/src/ssa.rs#L215-L216) explaining why we add them following another step. 

On the third hand the more passes we have the less chance that `pass_vs_prev` finds something interesting. 

Maybe there could be an `SsaPass` combinator constructor that runs 2 passes, so it's visible in the pipeline why it's there, leaves individual passes free to focus on their own concerns, but only prints one message?

aakoshh: Another way to look at it is that the inlining of simple functions has been audited on its own. Not saying it’s a stamp of approval, but in general if passes start calling each other, it’s less clear what the scope of auditing such a pass is.

TomAFrench: hmm, yeah I can go along with that. I'll replace this with a standalone pass for now and we can perhaps add the combinator in future.

aakoshh (approved):

PR #8989, opened 2025-06-21
feat: blake2b

ordian:
blake2b is used in Kusama and Polkadot and noir is [considered](https://forum.polkadot.network/t/make-kusama-chaotic-again/11123/32) to be integrated with Kusama, so this is a necessary step towards that integration


TomAFrench:
Hey @ordian! Before we can merge this we'd need an implementation of blake2b to be added to barretenberg, otherwise noir programs which use this new opcode would not be provable.

Barretenberg is currently being prepped for audit so I'm not sure this will be a priority in the short term but I'd recommend opening an issue on https://github.com/AztecProtocol/aztec-packages.

TomAFrench:
I'm going to close this PR as we cannot move forwards without a backend implementation.

PR #8993, opened 2025-06-23
fix: replace public key with curve generators in inactive branches

TomAFrench:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
@@ -131,6 +131,8 @@
 //!   v11 = mul v4, Field 12
 //!   v12 = add v10, v11
 //!   store v12 at v5         (new store)
+use std::sync::Arc;
+
 use fxhash::{FxHashMap as HashMap, FxHashSet as HashSet};
 
 use acvm::{FieldElement, acir::AcirField, acir::BlackBoxFunc};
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -830,69 +832,158 @@ impl<'f> Context<'f> {
...
+                                condition,
+                                call_stack,
+                            );
+
+                            Instruction::Call { func, arguments }
+                        }
+
+                        BlackBoxFunc::RecursiveAggregation => {
+                            todo!("Conditional recursive aggregation is not supported yet")
+                        }
```
TomAFrench:
```suggestion
                        // TODO: determine what changes are necessary here.
                        BlackBoxFunc::RecursiveAggregation => Instruction::Call { func, arguments },
```

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -830,69 +832,157 @@ impl<'f> Context<'f> {
...
+                                call_stack,
+                            );
+
+                            let instruction = Instruction::MakeArray { elements, typ };
+                            let array = self.insert_instruction(instruction, call_stack);
+                            arguments[points_array_idx] = array;
+                            Instruction::Call { func, arguments }
+                        }
+
+                        BlackBoxFunc::EcdsaSecp256k1 => {
```
TomAFrench, line 907:
```suggestion
                        // The ECDSA blackbox functions will fail to prove inside barretenberg in the situation where
                        // the public key doesn't not sit on the relevant curve.
                        //
                        // We then replace the public key with the generator point if the constraint is inactive to avoid
                        // invalid public keys from causing constraints to fail.
                        BlackBoxFunc::EcdsaSecp256k1 => {
```

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -830,69 +832,162 @@ impl<'f> Context<'f> {
...
+                                arguments[1],
+                                GENERATOR_Y,
+                                condition,
+                                call_stack,
+                            );
+
+                            Instruction::Call { func, arguments }
+                        }
+
+                        // TODO: determine what changes are necessary here.
```
guipublic: This comment could be improved; either link to an issue or mention why changes may be required.

guipublic (approved):
I mentioned a comment that can be improved, but apart from this, LGTM.

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
TomAFrench:
```suggestion
                        // TODO: https://github.com/noir-lang/noir/issues/8998
```

TomAFrench: Done, we have an issue already but I've added a link.

Issue #8995, opened 2025-06-23
Misleading error on dynamic indexing: operation known to overflow causes "only constant index" error


aakoshh:
### Aim

Reproduce one of the errors in https://github.com/noir-lang/noir/actions/runs/15814544181/job/44570940834

The following code works:
```rust
fn main() -> pub bool {
    let b: [(bool, &mut u1); 2] = [(false, &mut 0), (true, &mut 1)];
    b[((0 - 1) as u32 % 2)].0
}
```

What looks like an overflow is not, because 0-1 is a `Field`, which is then cast to `u32` to be some large value.

However in the AST the literals were of a different type. If we annotate them like so, we get a misleading error:
```
fn main() -> pub bool {
    let b: [(bool, &mut u1); 2] = [(false, &mut 0), (true, &mut 1)];

    let x: u32 = 0;
    let y: u32 = 1;
    let z: u32 = 2;

    b[((x - y) % z)].0
}
```
The error says "Only constant indices are supported when indexing an array containing reference values".

### Expected Behavior

Based on the discussion in documentation and discussion in https://github.com/noir-lang/noir/pull/8888/files#diff-c4074234ede131aa0635431511a53793853f6ee7a470f63159f2300849764c7dR83 I expected the ban on dynamic indexing to be based on whether something is coming from program input:

> Indexing an array with non-constant indices (indices derived from the inputs to the program) is also is also called "dynamic indexing"
> There is another restriction with dynamic indices: they cannot be used on arrays with elements which contain a reference type

Since `x`, `y` and `z` are all constants in the code, I'd expect the compiler to be able to use them, which it can _iff_ the `x - y` does not overflow, but it cannot if it overflows. So for example with `let y = 0;` it works, and that suggests the compiler knows that the expression whether it's going to work, and therefore it should present a better error message. 

### Bug

```console
❯ cargo run -q -p nargo_cli -- execute --show-ssa-pass Initial
After Initial SSA:
acir(inline) fn main f0 {
  b0():
    v0 = allocate -> &mut u1
    store u1 0 at v0
    v2 = allocate -> &mut u1
    store u1 1 at v2
    v4 = make_array [u1 0, v0, u1 1, v2] : [(u1, &mut u1); 2]
    v7 = sub u32 0, u32 1
    v8 = truncate v7 to 1 bits, max_bit_size: 32
    v10 = lt v8, u32 2
    constrain v10 == u1 1, "Index out of bounds"
[...]

PR #8996, opened 2025-06-23
fix(fuzz): Avoid overflowing binary ops in "no dynamic" mode

aakoshh:
# Description

## Problem\*

Workaround for https://github.com/noir-lang/noir/issues/8995

## Summary\*

When the AST code generator is in "no dynamic" mode, avoid binary operations that can fail, otherwise we will get a compilation error saying that we are trying to index an array with a non-constant expression, which is true from the compiler perspective, but fell outside the mechanism implemented to prevent this kind of indexing, which only tracked whether variables came from either program inputs or unconstrained environment.

## Additional Context

To replicate the error in the ticket:
```shell
NOIR_AST_FUZZER_SHOW_AST=1 NOIR_AST_FUZZER_SEED=0xcfb2f7da00100000 cargo test -p noir_ast_fuzzer_fuzz min_vs_full
```

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/ast_fuzzer/src/program/func.rs
+++ b/tooling/ast_fuzzer/src/program/func.rs
@@ -726,11 +726,14 @@ impl<'a> FunctionContext<'a> {
         max_depth: usize,
     ) -> arbitrary::Result<Option<TrackedExpression>> {
         // Collect the operations can return the expected type.
+        // Avoid operations that can fail in no-dynamic mode, otherwise they will be considered non-constant indexes.
         let ops = BinaryOp::iter()
             .filter(|op| {
                 types::can_binary_op_return(op, typ)
-                    && (!self.ctx.config.avoid_overflow || !types::can_binary_op_overflow(op))
-                    && (!self.ctx.config.avoid_err_by_zero || !types::can_binary_op_err_by_zero(op))
+                    && (!self.ctx.config.avoid_overflow && !self.in_no_dynamic
+                        || !types::can_binary_op_overflow(op))
+                    && (!self.ctx.config.avoid_err_by_zero && !self.in_no_dynamic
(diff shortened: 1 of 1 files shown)


tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -726,11 +726,14 @@ impl<'a> FunctionContext<'a> {
...
+        // Avoid operations that can fail in no-dynamic mode, otherwise they will be considered non-constant indexes.
         let ops = BinaryOp::iter()
             .filter(|op| {
                 types::can_binary_op_return(op, typ)
-                    && (!self.ctx.config.avoid_overflow || !types::can_binary_op_overflow(op))
-                    && (!self.ctx.config.avoid_err_by_zero || !types::can_binary_op_err_by_zero(op))
+                    && (!self.ctx.config.avoid_overflow && !self.in_no_dynamic
+                        || !types::can_binary_op_overflow(op))
+                    && (!self.ctx.config.avoid_err_by_zero && !self.in_no_dynamic
+                        || !types::can_binary_op_err_by_zero(op))
```
asterite: Can parentheses be added to make it explicit what the `||` applies to? (though I don't know if clippy will complain about this, but we can silence clippy in that case)

aakoshh: Sure, that's how I started, but I thought this was a bit more readable.

aakoshh: The idea was that this way it's `(!<should-avoid> || !<would-cause-problem>)`.

Changed to `!(<should-avoid-foo> || <should-avoid-bar>) || !<would-cause-problem>)`

asterite (approved):

PR #9007, opened 2025-06-24
fix: add limit to brillig vm execution trace

### LINKED ISSUE #5026: Recusion / AVM trace is unlimited
### Aim

I was testing a function on an edgecase, but it kept running forever.

### Expected Behavior

There should be a limit on the recursion and separately on the AVM trace length to avoid unbounded computation and draining resources.

### Bug

Computation is unbounded.

### To Reproduce

The following code allows to DoS the prover.
```rust
unconstrained fn recurse() {
        recurse();
    }
    #[test]
    fn test_recurse() {
        recurse();
    }
```
Run this test

### Project Impact

Nice-to-have

### Impact Context

_No response_

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Installation Method

Compiled from source

### Nargo Version

nargo version = 0.29.0 noirc version = 0.29.0+b776d287db096b657e3afa12da9ae18a803a72ea (git version hash: e7e0fa5a54317e09a1c82c93c2609a688a188a88, is dirty: true)

### NoirJS Version

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #9006: Compiler hangs: infinite loop in `Inlining Brillig Calls`
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8979

The following code causes the compiler to hang:

```rust
fn main() {
    unsafe { func_2() };
}

unconstrained fn func_2() {
    loop {
        if false {
            break
        }
    }
}
```

or like this:
```rust
fn main() {
    unsafe { func_2(false) };
}

    loop {
        if cond {
            break
        }
    }
}

```

### Expected Behavior

Expected it to compile, despite the infinite loop. 

### Bug

```console
$ cargo run -q -p nargo_cli -- compile --show-ssa
After Verifying no dynamic array indices to reference value elements (1) (step 41):
  b0():
    call f1(u1 0)
    return
}
  b0(v0: u1):
    jmp b1()
  b1():
    jmpif v0 then: b2, else: b3
  b2():
    return
  b3():
    jmp b1()
}

^C⏎                                                                                                                                                                                                                                                  
```
That is the last step of the primary pipeline. If we change the program to remove the infinite loop, we can see that the next step would be `After Inlining Brillig Calls (1) (step 42):`, which involves running the Brillig VM with constant inputs. In this case it means the compiler never returns, because it gets stuck in the infinite loop. 

Maybe there could be some limit to how far we try to evaluate such calls at compile time.

### To Reproduce

See above.

### Workaround

None

### Workaround Description

Not having any constant arguments that cause infinite loops in Brillig functions. 

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.7 noirc version = 1.0.0-beta.7+45246337701fe88d7f97416ee0715d23acf50ab2 (git version hash: d7030e8c4e05dcba0eb30adac7f0f18458b306f0, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

TomAFrench:
# Description

## Problem\*

Resolves #5026 
Resolves #9006 

## Summary\*

This PR adds an optional limit to the number of brillig opcodes to execute before halting execution and handling control back to the caller. I've set this to a limit of 10M opcodes (chosen based on execution trace of `rollup_block_root` when compiled to brillig being 317,303,405).

I'm still consdering how we want this interface to look.
 
## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/acvm-repo/brillig_vm/src/lib.rs
+++ b/acvm-repo/brillig_vm/src/lib.rs
@@ -641,7 +641,7 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>> VM<'a, F, B> {
         self.set_program_counter(self.program_counter + 1)
     }
 
-    /// Sets the program counter by `value`.
+    /// Sets the program counter to `value`.
     /// If the program counter no longer points to an opcode
     /// in the bytecode, then the VMStatus reports halted.
     fn set_program_counter(&mut self, value: usize) -> VMStatus<F> {


acvm-repo/brillig_vm/src/lib.rs
```diff
@@ -618,7 +641,7 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>> VM<'a, F, B> {
         self.set_program_counter(self.program_counter + 1)
     }
 
-    /// Increments the program counter by `value`.
+    /// Sets the program counter by `value`.
```
TomAFrench:
```suggestion
    /// Sets the program counter to `value`.
```

Issue #8994, opened 2025-06-23
Misleading error message: Could not resolve array reference instead of Index Out Of Bounds


aakoshh:
### Aim

Trying to replicate one of the errors in https://github.com/noir-lang/noir/actions/runs/15814544181/job/44570940834

The following programs errors saying references need to be resolved at compile time, but there is nothing the compiler doesn't know:

```rust
fn main() -> pub bool {
    let b: [(bool, &mut u1); 1] = [(false, &mut 0)];
    b[1].0
}
```

### Expected Behavior

It should say that `b[1]` is out of bounds, since the array size is 1.

### Bug

```console
❯ cargo run -q -p nargo_cli -- compile
error: Could not resolve some references to the array. All references must be resolved at compile time
  ┌─ src/main.nr:2:49
  │
2 │     let b: [(bool, &mut u1); 1] = [(false, &mut 0)];
  │                                                 -
  │
  = Call stack:
    1. src/main.nr:2:49

```

Note that it's pointing at `0`, but that's not where the problem is.
If I change `&mut u1` to just `u1` then the error instead moves to `u[1]`, so something about the reference interferes with the error. 

The final SSA shows that we know that the index will overflow:
```
After Dead Instruction Elimination - ACIR (1) (step 44):
acir(inline) predicate_pure fn main f0 {
  b0():
    v0 = allocate -> &mut u1
    v2 = make_array [u1 0, v0] : [(u1, &mut u1); 1]
    constrain u1 0 == u1 1, "Index out of bounds"
    v5 = array_get v2, index u32 2 -> u1
    return v5
}
```

Perhaps this, too, would be solved by https://github.com/noir-lang/noir/pull/8979

### To Reproduce

See above.

### Workaround

None

### Project Impact

None

### Nargo Version

nargo version = 1.0.0-beta.7 noirc version = 1.0.0-beta.7+5fff72ae09112e5f067507cdcd06e99151eed748 (git version hash: d7030e8c4e05dcba0eb30adac7f0f18458b306f0, is dirty: true)

### Would you like to submit a PR for this Issue?

None

jfecher:
Same issue as https://github.com/noir-lang/noir/issues/8995#issuecomment-2996705048.

This is because the operation is an index out of bounds and can't be optimized out so it appears non-constant to any code including the one issueing the error here

aakoshh:
The index here is an actual constant, though. 

[...]

PR #9008, opened 2025-06-24
fix: Add a remove_unreachable_instructions SSA pass (with unreachable terminator)

asterite:
# Description

## Problem

Resolves https://github.com/noir-lang/noir/issues/8774
Resolves https://github.com/noir-lang/noir/issues/8994

## Summary

Alternative to #8979 that introduces a new `unreachable` terminator.

## Documentation

Check one:

# PR Checklist


asterite:
@jfecher @TomAFrench I'm interested in your opinion on this PR. I started this by zeroing out terminators but then I figured that having a new `unreachable` terminator might be better as it makes the code clearer. Plus there's precedence for such a thing ([LLVM has an unreachable terminator](https://llvm.org/docs/LangRef.html#unreachable-instruction)). The only issue with this new terminator is that it currently isn't implemented in the inlining pass. However, the pass that introduces unreachable terminators currently runs way after inlining so it's not an issue. We could do this work in a future PR if it's really needed.

TomAFrench:
> The only issue with this new terminator is that it currently isn't implemented in the inlining pass. However, the pass that introduces unreachable terminators currently runs way after inlining so it's not an issue.

I don't think this is a showstopper. We've got a number of other passes which are required to run before or after another pass so this wouldn't be unique.

compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -915,8 +915,11 @@ impl<'a> Context<'a> {
     ) -> usize {
         let return_values = match terminator {
             TerminatorInstruction::Return { return_values, .. } => return_values,
+            TerminatorInstruction::Unreachable { .. } => return 0,
```
jfecher, line 918: In what cases do we reach this in acir?
I'm assuming we do since this used to be a catch-all for  `unreachable!`

compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -935,7 +938,10 @@ impl<'a> Context<'a> {
                 (return_values, *call_stack)
             }
             // TODO(https://github.com/noir-lang/noir/issues/4616): Enable recursion on foldable/non-inlined ACIR functions
-            _ => unreachable!("ICE: Program must have a singular return"),
+            TerminatorInstruction::JmpIf { .. } | TerminatorInstruction::Jmp { .. } => {
+                unreachable!("ICE: Program must have a singular return")
+            }
+            TerminatorInstruction::Unreachable { .. } => return Ok((vec![], vec![])),
```
jfecher, line 944: Same question here

compiler/noirc_evaluator/src/ssa/interpreter/errors.rs
```diff
@@ -58,6 +58,8 @@ pub enum InterpreterError {
     ToRadixFailed { field_id: ValueId, field: FieldElement, radix: u32 },
     #[error("Failed to solve blackbox function {name}: {reason}")]
     BlackBoxError { name: String, reason: String },
+    #[error("Reached the unreachable")]
+    ReachedTheUnreachable,
```
jfecher, line 62: This sounds like an inspiring quote

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -0,0 +1,465 @@
...
+                let call_stack = terminator.call_stack();
+                context.dfg[block_id]
+                    .set_terminator(TerminatorInstruction::Unreachable { call_stack });
+            }
+        });
+    }
+}
+
+/// Returns `true` if the given instruction will always produce an asertion failure.
+fn always_fails(dfg: &DataFlowGraph, instruction: &Instruction) -> bool {
```
jfecher, line 60: Nice - we can extend this in the future to things like array get/sets with indices which are known to overflow

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -0,0 +1,465 @@
...
+            jmp b4()
+          b4():
+            v3 = add Field 1, Field 2
+            return v3
+        }
+        "#);
+    }
+
+    #[test]
+    fn does_not_zeroes_terminator_of_non_dominated_block_4() {
```
jfecher: Looks like some tests names still mention the old approach for zeroing terminators

jfecher (approved):
LGTM. I think I prefer this approach to the old one. It makes it easier to track what happened previously and I like that it models our intention of the code more closely.

I agree with Tom that the ordering constraint on this pass is not the end of the world.

compiler/noirc_evaluator/src/acir/mod.rs
asterite: Good question. Yes, I think one of the tests failed so I handled this case. Here's a code that triggers this:

```noir
fn main() -> pub bool {
    assert(false);
    true
}
```

The final SSA is:

```noir
  b0():
    constrain u1 0 == u1 1
    unreachable
}
```

compiler/noirc_evaluator/src/ssa/interpreter/errors.rs
asterite: 😆 

I think it's the message you get when in LLVM you somehow reach the unreachable terminator, if you compile things in debug mode. In release mode unreachable produces no code. When developing Crystal we'd sometimes reach this "Reached the unreachable" and laugh (initially, then we'd have to chase down the bug 😅). I think LLVM's message is actually "Reached the unreachable!", with a bang, so even funnier.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: True! I was thinking a range check could also always fail, but I didn't include it in this PR for simplicity.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: Good catch, I just renamed them.

In reality most of these tests now are redundant. Previously when zeroing out things I had to make sure dominated blocks were also cleared out. Now that we replace terminators with `unreachable`, if a block was dominated by something that became unreachable then it naturally can't be reached anymore and won't show up in the final SSA.

But... I guess it's fine to keep these tests to avoid regressions.

PR #9013, opened 2025-06-24
feat: show why an assertion will always fail if it's a static string

asterite:
# Description

## Problem

In https://github.com/noir-lang/noir/issues/8995#issuecomment-3001528393 I noticed that when the compiler produced a `bug: Assertion is always false` warning it didn't show the assertion failure message if it's known.

## Summary

So for example this program:

```noir
fn main() -> pub u32 {
    let x: [u32; 0] = [];
    x[0]
}
```

used to produce this error:

```
  ┌─ src/main.nr:3:7
  │
3 │     x[0]
  │       - As a result, the compiled circuit is ensured to fail. Other assertions may also fail during execution
  │
  = Call stack:
    1. src/main.nr:3:7
```

Now the error is:

```
bug: Assertion is always false: Index out of bounds
  ┌─ src/main.nr:3:7
  │
3 │     x[0]
  │       - As a result, the compiled circuit is ensured to fail. Other assertions may also fail during execution
  │
  = Call stack:
    1. src/main.nr:3:7
```

Additionally, I made `compile_success_with_bug` produce snapshots for stderr, which were missing. I added this because the second commit is a proof that this PR actually shows the assertion failure message :-)

## Documentation

Check one:

# PR Checklist


aakoshh (approved):
Ace!

PR #9017, opened 2025-06-25
chore: add idempotency check to `remove_unreachable_functions`

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

~I noticed in #9016 that we start our SSA pipeline with `remove_unreachable_functions` which I would expect to do nothing as the initial SSA codegen should not generate any unreachable functions afaik. This PR adds a check to enforce this.~

Pivoting this PR to just add an idempotency check to this SSA pass.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_functions.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_functions.rs
@@ -21,6 +21,8 @@
 
 use std::collections::BTreeSet;
 
+use fxhash::FxHashSet as HashSet;
+
 use crate::ssa::{
     ir::{
         call_graph::CallGraph,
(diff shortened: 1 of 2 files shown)


TomAFrench:
>  the initial SSA codegen should not generate any unreachable functions afaik.

Looks like this can happen when assigning functions to values.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_functions.rs
```diff
@@ -34,29 +36,66 @@ use crate::ssa::{
...
+/// Post-check condition for [Ssa::remove_unreachable_functions].
+///
+/// Succeeds if:
+///   - `ssa` contains no unreachable functions, i.e., all functions are reachable from the entry points.
+///
+/// Otherwise panics.
+///
+/// ## Note
+/// We reuse the same logic for checking reachability as in the main pass, so this function will not
+/// catch any bugs in the pass itself, but it will ensure that the pass is idempotent.
```
aakoshh, line 60: I'm not sure _idempotent_ is the right word here. After all removing an unreachable function does change the SSA.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_functions.rs
```diff
@@ -34,29 +36,66 @@ use crate::ssa::{
...
 
+/// Post-check condition for [Ssa::remove_unreachable_functions].
+///
+/// Succeeds if:
+///   - `ssa` contains no unreachable functions, i.e., all functions are reachable from the entry points.
+///
+/// Otherwise panics.
+///
+/// ## Note
+/// We reuse the same logic for checking reachability as in the main pass, so this function will not
```
aakoshh, line 59: Shouldn't we think of another criteria to test with? The way it's implemented seems like it's testing the `retain` function itself. 

For example we could assert something like: every remaining function should be used by something. It wouldn't detect a completely isolated cycle of functions, but could work for finding counter examples without having to rely on exactly the same logic.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_functions.rs
TomAFrench: A function is idempotent if applying it N times has the same effect as applying it once. Checking that there's no reachable functions to be removed in later invocations is correct to ensure idempotency.

aakoshh (approved):

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_functions.rs
TomAFrench: Ensuring that the same algo returns no unreachable functions checks that the pass is idempotent as a whole for very little effort. Considering the simplicity of `used_functions`, we'd end up including the same logic in the suggested check.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_functions.rs
aakoshh: Of course, that was an embarrassing brain fart 🤦‍♂️

PR #9018, opened 2025-06-25
fix(ssa): Perform `mem2reg` before DIE

### LINKED ISSUE #8779: Compiler crash on dynamic index of array of reference: Could not resolve some references to the array

aakoshh:
# Description

## Problem\*

Resolves #8779 

## Summary\*

Fixes the SSA pipeline to perform `mem2reg` before the DIE pass, as per the comments regarding the `Store` instruction in `can_be_eliminated_if_unused`.

Also added a post-check to DIE to see that no `Load` or `Store` instructions are left in it if the conditions indicate that `Store` can be removed.

## Additional Context

The program in the bug ticket had this SSA:

```rust
After Check u128 mul overflow (1) (step 30):
  b0(v0: u32):
    v1 = allocate -> &mut u1
    store u1 1 at v1
    v3 = make_array [v1] : [&mut u1; 1]
    constrain v0 == u32 0, "Index out of bounds"
    v5 = load v1 -> u1                            	// src/main.nr:3:6
    return v5
}

After Dead Instruction Elimination (1) (step 31):
  b0(v0: u32):
    v1 = allocate -> &mut u1
    constrain v0 == u32 0, "Index out of bounds"
    v3 = load v1 -> u1                            	// src/main.nr:3:6
    return v3
}
```

The problem was that `store u1 1 at v1` was removed by the DIE pass, even though it's used in `v3 = load v1`. The reason for this is:

> Store instructions must be removed by DIE in acir code, any load
> instructions should already be unused by that point.
>
> Note that this check assumes that it is being performed after the flattening
> pass and after the last mem2reg pass. This is currently the case for the DIE
> pass where this check is done, but does mean that we cannot perform mem2reg
> after the DIE pass.

So DIE assumes that `mem2reg` has figured out and eliminated any `store` followed by `load` and replaced e.g. `v3` with the value in `u1 1`. However that wasn't the case here, and by removing the `store`, it prevented the `mem2reg` that followed from figuring it out, which meant the `load` survived until the end, and caused a panic during ACIR gen.

The array by this point was gone, but the error mentioned it anyway, but that's just because `UnknownReference` has this error message.

I don't know if there could be longer chains that would require multiple `mem2reg` passes, which is why I added the post-check. But since there already was a `mem2reg` after DIE, which is hampered by the removal of `Store`, I thought moving it up should only improve things. 

### Other moves

I wasn't sure if `simplify_cfg` shouldn't also be moved up: 
* On one hand it says it removes blocks with no instructions, and I thought if DIE removed the last unused instruction, we might get more opportunities to simplify things.
* On the other hand, when we unify two blocks in `mem2reg`, if one is yet to be visited, it will make it forget all known values from the visited block, even if the one it merges with doesn't modify them. If simplification reduces the number of blocks, `mem2reg` has a better chance of keeping known values. 

In the end I moved it as well, because otherwise the `references` test grows by 25% in Brillig opcode size.

I also had to move the new `remove_unreachable_instructions` before the DIE pass, so that the #8774 example doesn't fail on the new post-check that asserts that no `Load` should be left after DIE: in that case it's impossible to load from an empty array, so `mem2reg` couldn't get rid of it.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -610,6 +610,38 @@ impl RcTracker {
...
 }
 
+/// Store instructions must be removed by DIE in acir code, any load
+/// instructions should already be unused by that point.
+///
+/// Note that this check assumes that it is being performed after the flattening
+/// pass and after the last mem2reg pass. This is currently the case for the DIE
+/// pass where this check is done, but does mean that we cannot perform mem2reg
+/// after the DIE pass.
+fn should_remove_load_and_store(func: &Function, flattened: bool) -> bool {
```
jfecher:
```suggestion
fn should_remove_store(func: &Function, flattened: bool) -> bool {
```
Doesn't look like we call this for loads

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -431,16 +440,7 @@ fn can_be_eliminated_if_unused(
...
-        // instructions should already be unused by that point.
-        //
-        // Note that this check assumes that it is being performed after the flattening
-        // pass and after the last mem2reg pass. This is currently the case for the DIE
-        // pass where this check is done, but does mean that we cannot perform mem2reg
-        // after the DIE pass.
-        Store { .. } => {
-            flattened && function.runtime().is_acir() && function.reachable_blocks().len() == 1
-        }
+        Store { .. } => should_remove_load_and_store(function, flattened),
```
jfecher:
```suggestion
        Store { .. } => should_remove_store(function, flattened),
```

jfecher (approved):
LGTM.
And yes, the original expectation was for there to be no more loads before DIE runs and thus DIE could just freely remove all Stores.

compiler/noirc_evaluator/src/ssa/opt/die.rs
aakoshh: I named it like this because in the post check I look for the absence of both loads and stores, that is, stores should be removed because loads should be removed because loads should be unused. We don't call it for that, but that's what should be happening, they should go out together.

PR #9019, opened 2025-06-25
fix: compute the dominance frontier of the reverse cfg using the extended cfg

### LINKED ISSUE #9005: Compiler crash: `Post order for unreachable block is undefined`
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8979

The following code causes the compiler to crash:
```rust
fn main() {
    unsafe { func_2(true) };
}

    loop {
        if cond {
            break;
        }
        loop {
            if false {
                break
            }
        }
    }
}
```

### Expected Behavior

Yes, there is an infinite loop in the code, but the compiler should not crash. 

### Bug

```console
❯ cargo run -q -p nargo_cli -- compile --show-ssa
After Initial SSA:
acir(inline) fn main f0 {
  b0():
    call f1(u1 1)
    return
}
brillig(inline) fn func_2 f1 {
  b0(v0: u1):
    jmp b1()
  b1():
    jmpif v0 then: b3, else: b4
  b2():
    return
  b3():
    jmp b2()
  b4():
    jmp b5()
  b5():
    jmp b5()
}
...
After Purity Analysis (1) (step 14):
  b0():
    call f1(u1 1)
    return
}
  b0(v0: u1):
    jmpif v0 then: b1, else: b2
  b1():
    return
  b2():
    jmp b3()
  b3():
    jmp b3()
}

Message:  internal error: entered unreachable code: Post order for unreachable block is undefined
Location: compiler/noirc_evaluator/src/ssa/ir/dom.rs:79
```

### To Reproduce

See above.

### Workaround

None

### Workaround Description

Make sure there is a viable condition for `break`. 

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.7 noirc version = 1.0.0-beta.7+45246337701fe88d7f97416ee0715d23acf50ab2 (git version hash: d7030e8c4e05dcba0eb30adac7f0f18458b306f0, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #9005

## Summary\*
The reverse CFG is not a CFG, i.e it does not always have a single entry node. As a result using the dominance frontier algorithm for the reverse cfg can fail.
We extend the CFG for this computation by ensuring there is a unique exit node.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/cfg.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/cfg.rs
@@ -181,10 +181,10 @@ impl ControlFlowGraph {
     /// b3   b4          b3   b4              
     /// |    |           ^    ^             
     /// v    v           |    |             
-    /// b2   b5 <-|      b2*  b5 <-|             
+    /// b2   b5 <-|      b2   b5 <-|             
     /// \    /\___|      ^    ^\___|             
     ///  v  v            \    /
-    ///  exit             exit
+    ///  exit             exit*
     pub(crate) fn extended_reverse(func: &mut Function) -> Self {
         let mut cfg = Self::with_function(func);
         // Exit blocks are the ones having no successor


compiler/noirc_evaluator/src/ssa/ir/cfg.rs
```diff
@@ -148,6 +148,48 @@ impl ControlFlowGraph {
...
+        let exit_nodes: Vec<BasicBlockId> =
+            cfg.data.keys().filter(|&&block| cfg.successors(block).len() == 0).copied().collect();
+        // Traverse the reverse CFG from the exit blocks
+        let reverse = cfg.reverse();
+        let post_order = crate::ssa::ir::post_order::PostOrder::with_cfg(&reverse);
+        // Extract blocks that are not reachable from the exit blocks
+        let dead_blocks: Vec<BasicBlockId> = cfg
+            .data
+            .keys()
+            .filter(|&block| !post_order.as_slice().contains(block))
```
aakoshh: Perhaps you could collect into a `HashSet` before, to avoid having `O(n)` lookups?

compiler/noirc_evaluator/src/ssa/ir/cfg.rs
```diff
@@ -148,6 +148,48 @@ impl ControlFlowGraph {
     pub(crate) fn compute_entry_blocks(&self) -> Vec<BasicBlockId> {
         self.data.keys().filter(|&&block| self.predecessors(block).len() == 0).copied().collect()
     }
+
+    /// Computes the reverse graph of the extended CFG.
+    /// The extended CFG is the CFG with an additional unique exit node (if there is none)
+    /// such that there is a path from every block to the exit node.
+    pub(crate) fn extended_reverse(func: &mut Function) -> Self {
```
aakoshh, line 188: It would be really useful to have an ASCII diagram depicting a graph with dead blocks or multiple exit nodes.

compiler/noirc_evaluator/src/ssa/ir/cfg.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/ir/cfg.rs
aakoshh: Here's one:
```
forward          reverse
-------          -------
  b0*              b0              
  |                ^             
  v                |             
  b1               b1             
 /  \             ^  ^             
v    v           /    \             
b3   b4          b3   b4              
|    |           ^    ^             
v    v           |    |             
b2   b5 <-|      b2*  b5 <-|             
      \___|            \___|                      
```
It's from the integration test. Shows the entry points with `*` and then we can mention that `b4` and `b5` in the reverse CFG are dead.

compiler/noirc_evaluator/src/ssa/ir/cfg.rs
```diff
@@ -148,6 +149,43 @@ impl ControlFlowGraph {
...
+        let mut cfg = Self::with_function(func);
+        // Exit blocks are the ones having no successor
+        let exit_nodes: Vec<BasicBlockId> =
+            cfg.data.keys().filter(|&&block| cfg.successors(block).len() == 0).copied().collect();
```
aakoshh, line 192: Just to make sure I understand: both the forward and the reverse CFG in the example only have 1 exit node, however, it's not reachable from every block. That is, `b5` is not an exit node, because it has a successor: itself. 

Not sure if you found a case with multiple exit nodes, but in case you didn't, then perhaps instead of adding a new unique exit node to the DFG and connecting all dead and exit nodes to it, we could connect dead nodes in the reverse post order graph to the single exit node of the original? Or would that distort the frontier algorithm?

Never mind, it probably would.

guipublic: If the original exit node is not empty, connecting b5 to it will not be correct. I don't know if this is a pb, but I prefer to use a new empty block.

aakoshh: I guess `unreachable` terminators would be the multiple exit node examples.

compiler/noirc_evaluator/src/ssa/ir/cfg.rs
guipublic: Done, thank you for the sample diagram.

compiler/noirc_evaluator/src/ssa/ir/cfg.rs
```diff
@@ -148,6 +149,75 @@ impl ControlFlowGraph {
...
+    ///   b1               b1             
+    ///  /  \             ^  ^             
+    /// v    v           /    \             
+    /// b3   b4          b3   b4              
+    /// |    |           ^    ^             
+    /// v    v           |    |             
+    /// b2   b5 <-|      b2*  b5 <-|             
+    /// \    /\___|      ^    ^\___|             
+    ///  v  v            \    /
+    ///  exit             exit
```
aakoshh:
```suggestion
    /// b2   b5 <-|      b2   b5 <-|             
    /// \    /\___|      ^    ^\___|             
    ///  v  v            \    /
    ///  exit             exit*
```

guipublic: Done

Issue #9020, opened 2025-06-25
Trait default methods should only be typed once


asterite:
### Aim

This is almost a duplicate of #8632 but instead of stating the problem it states how it should work.

Right now, when a trait has a default method, that default method is cloned and added to each of its trait impls. This has a few issues:
1. Paths are always solved related to the trait, so if they trait impl is in another module it doesn't work
2. If a default method refers to a trait generic, that generic doesn't exist anymore in the trait impl
3. If there's an error in the default implementation it's reported multiple times (once in the trait, then once more per trait impl)
4. (minor) we end up typing the same method multiple times, when it shouldn't be necessary

However, to implement this we would first need to allow referring to associated constants and types of a trait. That is, in this code:

```noir
trait Trait {
  let N: u32;

  fn foo() -> Self::N {
    Self::N
  }
}
```

`Self::N` should resolve to that trait's constant, but then it should be replaced by the one in each impl during monomorphization. In a way, implementing that is similar to implementing #8797

### Expected Behavior

Trait default methods should only be typed once.

### Bug

Trait default methods are typed multiple times.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

Issue #8797, opened 2025-06-04
Can't access associated constants via `AsTraitPath` in expressions


asterite:
### Aim

This code:

```noir
trait Serialize {
    let N: u32;
}

impl Serialize for Field {
    let N: u32 = 1;
}

fn main() {
    let x = <Field as Serialize>::N;
    println(x);
}
```

gives this error:

```
error: Trait `Serialize` has no method named `N`
   ┌─ src/main.nr:10:35
   │
10 │     let x = <Field as Serialize>::N;
   │                                   -
```

However, it should compile.

### Expected Behavior

The code should compile.

### Bug

It doesn't compile.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #9021, opened 2025-06-25
fix: correctly solve paths in trait default methods

### LINKED ISSUE #8632: Import errors on partial trait implementations

asterite:
# Description

## Problem

Resolves #8632

## Summary

This PR implements a workaround for solving #8632. The correct solution would be the one described in https://github.com/noir-lang/noir/issues/9020 but, as noted there, it's not trivial to do so. But this workaround would unblock users that use trait default methods (which, I think, is a very useful feature).

## Documentation

Check one:

# PR Checklist


jfecher:
Not sure about implementing a workaround when I'm also unsure how difficult the full fix would be. I'd like to take another stab at https://github.com/noir-lang/noir/issues/8797 which could maybe open us up to a different implementation of this where we elaborate once on the trait and use that instead of copying beforehand.

Issue #9022, opened 2025-06-25
Shift-left inconsistency


asterite:
### Aim

This program:

```noir
fn main() {
    let x: u8 = 1;
    println(x << 7);
}
```

prints 128. This is correct.

This program:

```noir
fn main() {
    let x: u8 = 1;
    println(x << 8);
}
```

gives a bug during compilation and fails to execute:

```
bug: Assertion is always false: attempt to bit-shift with overflow
  ┌─ src/main.nr:3:13
  │
3 │     println(x << 8);
  │             ------ As a result, the compiled circuit is ensured to fail. Other assertions may also fail during execution
```

But this program:

```noir
fn main() {
    let x: u8 = 2;
    println(x << 7);
}
```

prints zero.

So the overflow condition seems to happen exclusively if the right-hand side is guaranteed to overflow the target bitsize, but otherwise an overflow leads to zero.

What's the expected result here?

### Expected Behavior

It should either always overflow or always be zero.

### Bug

There's an inconsistency.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

guipublic:
Note that this is related to the question I asked in slack; we need first to clarify the expected behaviour that we want.

guipublic:
The behaviour we agreed on is the following:

At the Noir user level:
- The rhs of a bit shift will no longer be required to be a u8 and will instead match the type of the lhs
 -All bit shifts will error if their rhs is not in 0 <= rhs < lhs_bitsize, e.g. 3_u8 << 10 will be an error
- Signed types specifically will error if their rhs is not in 0 <= rhs < lhs_bitsize - 1` to account for the sign bit (note that the largest signed bit size is 64 which means this also guarantees rhs is non-negative for all bit sizes)
- Shifting signed types will use arithmetic shifts to preserve their sign, while shifting unsigned types will use logical shifts
If a user wants logical shifts on signed types they can always cast to unsigned, shift, then cast back.

Note:
- there is no distinction between constant and runtime bit shifts, we expect their observable behavior to be the same
- Disallowing 3u8 << 8  allows us to do a 3-bit range constraint which should be more efficient than if we had to allow 8 which would not be a useful shift anyway.

At the SSA/implementation level:
Likely implementation for this is for each shr and shl instruction to be preceeded by a range check. Thus each shift can assume the rhs is within range
[...]

PR #9023, opened 2025-06-25
fix: put constraint failure after binary operations that overflow

### LINKED ISSUE #8995: Misleading error on dynamic indexing: operation known to overflow causes "only constant index" error

asterite:
# Description

## Problem

Resolves #8995

## Summary

The "remove unreachable instructions" now does one more thing: it replaces binary operations that are guaranteed to fail with a constrain failure, then it removes subsequent instructions from that block and sets its terminator to unreachable.

Some compile bugs will start showing up now during compilation instead of runtime, but I think this is consistent with `assert(false)` and `[](0)` doing the same thing.

## Additional Context

It seems with this change this program stops compiling:

```noir
global G_A: bool = false;
fn main() {
    unsafe { func() };
}

unconstrained fn func() {
    let mut idx_c: u32 = 0;
    while G_A {
        if (idx_c == 3) {
            break
        } else {
            ()
        }
    }
    let mut f = 1 / 0;
}
```

The final SSA we get is:

```noir
g0 = u1 0

  b0():
    call f1()                                     	// src/main.nr:3:14
    return
}
brillig(inline) predicate_pure fn func f1 {
  b0():
    constrain u1 0 == u1 1, "attempt to divide by zero"	// src/main.nr:15:17
    unreachable
}
```

and for some reason it panics with this:

```
Message:  ICE: Global value not found in cache v0
Location: compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs:1833
```

I need to investigate why this happens...

## Documentation

Check one:

# PR Checklist


asterite:
Here's a smaller code that reproduces the issues:

```noir
global CONST: bool = false;

fn main() {
    let _ = CONST;
    unsafe { func() };
}

unconstrained fn func() {
    let _ = 1 / 0;
}
```

We get:

```
Message:  ICE: Global value not found in cache v0
Location: compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs:1834
```

`v0` refers to `CONST`. For some reason, because `1 / 0` is replaced with `constrain u1 0 == u1 1`, that `u1 0` reuses the global value and then ir crashes 🤔 But if I replace `let _ = 1 / 0` with `assert(false)` it works fine. Very strange!

jfecher:
> v0 refers to CONST. For some reason, because 1 / 0 is replaced with constrain u1 0 == u1 1, that u1 0 reuses the global value and then ir crashes 🤔 But if I replace let _ = 1 / 0 with assert(false) it works fine. Very strange!

This sounds like something inlining used to do: re-using the same ValueId from an integer in one function in a new function without properly mapping the id. So the same ID gets used for a global instead of an integer. I'm not sure if the culprit is inlining, but it sounds similar. It could be maybe that `func` doesn't use any globals so it doesn't get the GlobalId included but we expect globals to have the same ValueId everywhere so when we do some inlining (?) or something from `func` we end up checking against a global ValueId that func doesn't use or vice-versa.

asterite:
Ah, `used_globals` ended up with an incorrect value. We need to run DIE once more. But I think it would be better if DIE didn't compute that, and instead we'd compute it in a separate pass or maybe when needed before acir/brillig... I'll check.

compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -243,6 +243,8 @@ pub fn primary_passes(options: &SsaEvaluatorOptions) -> Vec<SsaPass> {
             // A function can be potentially unreachable post-DIE if all calls to that function were removed,
             // or after the removal of unreachable instructions.
             .and_then(Ssa::remove_unreachable_functions),
+        // Another DIE pass is needed to recompute `used_globals`.
+        SsaPass::new(Ssa::dead_instruction_elimination, "Dead Instruction Elimination"),
```
aakoshh: I am aware of the other PR, but couldn't we just swap the order of the removal of unreachable instruction and dead instructions?

compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
```diff
@@ -116,13 +122,36 @@ pub(crate) fn eval_constant_binary_op(
...
+
             // Check for overflow
             if result != 0 && result.ilog2() >= bit_size {
-                return None;
+                if let BinaryOp::Shl = operator {
+                    return None;
+                }
+
+                if let BinaryOp::Shr = operator {
+                    return Some(Ok((FieldElement::zero(), operand_type)));
```
aakoshh: This doesn't look right: if `result != 0` then we shouldn't return `FieldElement::zero()`. It was more like an overflowing multiplication, ending up with more bit size than what fits the type, so this is an `Err` I think.

compiler/noirc_evaluator/src/ssa.rs
asterite: I thought about it but now I can't remember why I didn't do that 😅 

In any case I'll remove it once the other PR gets merged.

compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
asterite: In this case it's a shift-right, and I _think_ that when that "overflows" we return zero... but I might be wrong. Shifting has been very confusing lately 😅

asterite: Ah, for unsigned integers I just checked and this is zero. But for unsigned integers it can be 0 or -1 so I'll at least fix that on the other branch.

aakoshh: Ah, so `result != 0` because it turned into a large number on the overflow, gotcha.

compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
```diff
@@ -168,7 +203,22 @@ pub(crate) fn eval_constant_binary_op(
...
+    Some(Ok((value, operand_type)))
+}
+
+fn binary_op_function_name(op: BinaryOp) -> &'static str {
+    match op {
+        BinaryOp::Add { .. } => "add",
+        BinaryOp::Sub { .. } => "subtract",
+        BinaryOp::Mul { .. } => "multiply",
+        BinaryOp::Div => "divide",
+        BinaryOp::Mod => "divide",
```
aakoshh:
```suggestion
        BinaryOp::Mod => "modulo",
```

asterite: Ah... here I left it so the message is "Attempt to divide by zero" because when taking the modulo you need to divide. I also thought that was Rust said, but now I see it says "attempt to calculate the remainder with a divisor of zero" so I'll update the code to do that.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -31,20 +41,88 @@ impl Function {
...
+                }
+                Instruction::Binary(binary @ Binary { lhs, operator, rhs }) => {
+                    let requires_acir_gen_predicate =
+                        binary.requires_acir_gen_predicate(context.dfg);
+                    if requires_acir_gen_predicate {
+                        let predicate = context.dfg.get_numeric_constant(side_effects_condition);
+                        match predicate {
+                            Some(predicate) => {
+                                if predicate.is_zero() {
+                                    return;
```
aakoshh: Do you think this part could be supported with some comments?

compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
```diff
@@ -116,13 +122,36 @@ pub(crate) fn eval_constant_binary_op(
...
             // and the operation should be handled by ACIR generation.
             if matches!(operator, BinaryOp::Div | BinaryOp::Mod) && rhs == 0 {
-                return None;
+                return Some(Err("attempt to divide by zero".to_string()));
             }
-            let result = function(lhs, rhs)?;
+
+            let Some(result) = function(lhs, rhs) else {
+                if let BinaryOp::Shl = operator {
+                    return None;
```
aakoshh: Why is SHL not an overflow?

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -56,34 +134,54 @@ impl Function {
...
+        | BinaryOp::Xor
+        | BinaryOp::Shl
+        | BinaryOp::Shr => (),
+    };
+
+    let rhs_value = context.dfg.get_numeric_constant(rhs)?;
+
+    if matches!(operator, BinaryOp::Div | BinaryOp::Mod) && rhs_value.is_zero() {
+        return Some("attempt to divide by zero".to_string());
+    }
```
aakoshh: Is this not covered by `eval_constant_binary_op`?

compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
asterite: Because of this: https://github.com/noir-lang/noir/issues/9022

I'll add a comment linking to that issue.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: No. When I was coding this I tried a code like `x / 0` and it wasn't changing to a constrain failure. I then realized `eval_constant_binary_op` expects two field values, so it only works if both operands are constants.

asterite: I guess we could change `eval_constant_binary_op` to have both operands as `Option<ValueId>`, then it would get their values, etc. But the other part where `eval_constant_binary_op` is used is when simplifying binaries and there it never gets simplified to a `Constrain` (that's what I tried first but things break horribly because `insert_binary` always expects one result but `Constrain` has no results).

compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
asterite: In the end this message will never show up here, I think, because "modulo" can never overflow. I changed it anyway because it's more correct. Then I updated the message I thinking about in another places.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: Done!

compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
asterite: Done!

asterite:
The initial performance boosts were apparently because of a bug I introduced (incorrect handling of `eval_constant_binary_op` which did `eval_constant_binary_op(...).is_some()` and the compiler didn't tell me to update it because it was still returning an `Option`. I eventually handled it, but now I changed `eval_constant_binary_op` to return an enum as it's clearer than trying to remember what's `None` vs. `Some(Ok)` vs. `Some(Err)`.

The tiny performance regression is, I think, one new assertion message is now hardcoded in brillig (previously it would fail without a message). But this should be a constant as it's always the same message(s).

aakoshh:
I narrowed down one of the fuzzer failures to the following. 

This program succeeds:
```rust
unconstrained fn main() {
    let c = (122_i8 << 5);
}
```
This one does not:
```rust
unconstrained fn main() {
    let c = (122_i8 << 5) * 1;
}
```

On `master` both succeed. The SSA of the first one goes like this:
```rust
After Removing Paired rc_inc & rc_decs (1) (step 4):
brillig(inline) fn main f0 {
  b0():
    v2 = shl i8 122, u8 5                         	// src/main.nr:2:14
    v3 = truncate v2 to 8 bits, max_bit_size: 9   	// src/main.nr:2:14
    return
}

After Preprocessing Functions (1) (step 5):
brillig(inline) fn main f0 {
  b0():
    return
}
```
While the second one preserves the `shl` because it's used in `mul`:
```rust
After Dead Instruction Elimination - ACIR (1) (step 42):
  b0():
    v2 = shl i8 122, u8 5                         	// src/main.nr:2:14
    v3 = truncate v2 to 8 bits, max_bit_size: 9   	// src/main.nr:2:14
    v4 = cast v3 as u8                            	// src/main.nr:2:13
    v6 = lt v4, u8 128                            	// src/main.nr:2:13
    v7 = not v6                                   	// src/main.nr:2:13
    v8 = cast v3 as Field                         	// src/main.nr:2:13
    v9 = cast v6 as Field                         	// src/main.nr:2:13
    v10 = mul v9, v8                              	// src/main.nr:2:13
    v12 = sub Field 256, v8                       	// src/main.nr:2:13
    v13 = cast v7 as Field                        	// src/main.nr:2:13
    v14 = mul v13, v12                            	// src/main.nr:2:13
    v15 = add v10, v14                            	// src/main.nr:2:13
    range_check v15 to 8 bits, "attempt to multiply with overflow"	// src/main.nr:2:13
    v16 = cast v15 as u8                          	// src/main.nr:2:13
    v17 = cast v7 as u8                           	// src/main.nr:2:13
    v18 = unchecked_add u8 128, v17               	// src/main.nr:2:13
    v19 = lt v16, v18                             	// src/main.nr:2:13
    constrain v19 == u1 1, "attempt to add with overflow"	// src/main.nr:2:13
    return
}
```

The fact that this does not fail means that `shl` is not meant to fail at the moment. 

But on this branch, it does, because the introduction of `* 1` means it survives long enough to be turned into an always-fail:
```rust
After Mem2Reg (4) (step 30):
  b0():
    v2 = shl i8 122, u8 5                         	// src/main.nr:2:14
    v3 = truncate v2 to 8 bits, max_bit_size: 9   	// src/main.nr:2:14
    v4 = cast v3 as u8                            	// src/main.nr:2:13
    v6 = lt v4, u8 128                            	// src/main.nr:2:13
    v7 = not v6                                   	// src/main.nr:2:13
    v8 = cast v3 as Field                         	// src/main.nr:2:13
    v9 = cast v6 as Field                         	// src/main.nr:2:13
    v10 = mul v9, v8                              	// src/main.nr:2:13
    v12 = sub Field 256, v8                       	// src/main.nr:2:13
    v13 = cast v7 as Field                        	// src/main.nr:2:13
    v14 = mul v13, v12                            	// src/main.nr:2:13
    v15 = add v10, v14                            	// src/main.nr:2:13
    range_check v15 to 8 bits, "attempt to multiply with overflow"	// src/main.nr:2:13
    v16 = cast v15 as u8                          	// src/main.nr:2:13
    v17 = cast v7 as u8                           	// src/main.nr:2:13
    v18 = unchecked_add u8 128, v17               	// src/main.nr:2:13
    v19 = lt v16, v18                             	// src/main.nr:2:13
    constrain v19 == u1 1, "attempt to add with overflow"	// src/main.nr:2:13
    return
}

After Remove Unreachable Instructions (1) (step 31):
  b0():
    constrain u1 0 == u1 1, "attempt to shift left with overflow"	// src/main.nr:2:14
    unreachable
}
```

The humble metamorphic test caught a regression 🕺

aakoshh:
OTOH another test which failed is this:
```shell
NOIR_AST_FUZZER_SEED=0x309cb9e20000f131 cargo test -p noir_ast_fuzzer_fuzz min_vs_full
```
and it that one the initial SSA already contains an always-fail constraint after the `shl`:


```rust
After Initial SSA:
brillig(inline) fn main f0 {
  ...
  b20():
    v37 = load v32 -> u32                         	// src/main.nr:39:33
    v38 = add v37, u32 1                          	// src/main.nr:42:42
    store v38 at v32                              	// src/main.nr:42:42
    v40 = shl v5, u8 210                          	// src/main.nr:43:51
    constrain u1 0 == u1 1, "attempt to bit-shift with overflow"	// src/main.nr:43:51
    ...
```
But this is because the RHS of the SHL is not less than the bit size.

aakoshh:
In the above test, the LICM moves the `shl` to the header of a loop, but leaves the failing constraint. Then the new pass turns the `shl` in the header into an always fail.

Here's a minimised program:
```rust
unconstrained fn main(mut b: bool) -> pub str<2> {
    for idx_f in 220538393_i32..220538394_i32 {
        let mut idx_h: u32 = 0_u32;
        while b {
            if (idx_h == 3_u32) {
                break
            } else {
                idx_h = (idx_h + 1_u32);
                assert((idx_f != (idx_f << 210_u8)), "FVM");
            }
        }
    }
    "ok"
}
```
called with `b = false`. 

And here's the LICM in action:
```rust
After Purity Analysis (1) (step 11):
...
  b8():
    v15 = load v9 -> u32                          	// src/main.nr:5:17
    v17 = add v15, u32 1                          	// src/main.nr:8:26
    store v17 at v9                               	// src/main.nr:8:26
    v19 = shl v1, u8 210                          	// src/main.nr:9:35
    constrain u1 0 == u1 1, "attempt to bit-shift with overflow"	// src/main.nr:9:35
    v22 = truncate v19 to 32 bits, max_bit_size: 33	// src/main.nr:9:35
    v23 = eq v1, v22                              	// src/main.nr:9:25
    constrain v23 == u1 0, "FVM"                  	// src/main.nr:9:25
    jmp b4()
}

After Loop Invariant Code Motion (1) (step 12):
...
  b2():
    v9 = allocate -> &mut u32                     	// src/main.nr:3:30
    store u32 0 at v9                             	// src/main.nr:3:30
    v12 = shl v1, u8 210                          	// src/main.nr:9:35
    v13 = truncate v12 to 32 bits, max_bit_size: 33	// src/main.nr:9:35
    v14 = eq v1, v13                              	// src/main.nr:9:25
    jmp b4()
...
  b8():
    v19 = load v9 -> u32                          	// src/main.nr:5:17
    v21 = add v19, u32 1                          	// src/main.nr:8:26
    store v21 at v9                               	// src/main.nr:8:26
    constrain u1 0 == u1 1, "attempt to bit-shift with overflow"	// src/main.nr:9:35
    constrain v14 == u1 0, "FVM"                  	// src/main.nr:9:25
    jmp b4()
}
```

I reported this in the past in https://github.com/noir-lang/noir/issues/8615 although it was the other way around, where the `shl` stayed put and the constraint moved 🙃 Maxim says there that he wasn't aware of `shl` failing, only the subsequent constraints are expected to fail, inserted by codegen.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -31,20 +41,94 @@ impl Function {
...
+                    }
+
+                    if let Some(message) =
+                        binary_operation_always_fails(*lhs, *operator, *rhs, context)
+                    {
+                        let zero = context.dfg.make_constant(0_u128.into(), NumericType::bool());
+                        let one = context.dfg.make_constant(1_u128.into(), NumericType::bool());
+                        let message = Some(ConstrainError::StaticString(message));
+                        let instruction = Instruction::Constrain(zero, one, message);
+                        context.remove_current_instruction();
```
aakoshh: Here's an idea. If we know that this binary instruction will always fail, we don't have to remove it and replace it with a generic constraint, rather we can leave it alone, and just remove all subsequent instructions. 

This would better preserve the reason of what is failing, like in the SSA interpreter we would actually see the values, and also make it easier to see in a large SSA block which was the instruction that caused things to shortcut, because the culprit would still be there.

aakoshh: As a stopgap solution to debug things like the `shl` not causing a constraint failure on its own, we could leave the offending instruction _and_ follow up with a backstop always fail constraint, to make sure we didn't just _think_ it would fail, and then it doesn't.

asterite:
@aakoshh Wow, thank you for looking into the fuzzer failures! 🙏 

> The fact that this does not fail means that shl is not meant to fail at the moment.

I just looked into the code and for unsigned integers we don't produce an overflow on `shl`, but for signed integers we do. Maybe if we don't overflow in this case (in `eval_constant_binary_op`) the issue will be solved? I'll try it after breakfast :-)

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: I like this idea!

The only "issue" with this is that it won't produce a `bug: ...` at compile time, but maybe that's fine as that's how it's been working so far. I'll try it later.

jfecher: > Here's an idea. If we know that this binary instruction will always fail, we don't have to remove it and replace it with a generic constraint, rather we can leave it alone, and just remove all subsequent instructions.

Something else to note is that if we leave in the instruction we will still get the arrayget/arrayset with references using dynamic indices error. I suppose we could add a check for that when that error is issued though to ensure the indices are not just known out of bounds, but it does sound like repeated work.

aakoshh: > if we leave in the instruction we will still get the arrayget/arrayset with references using dynamic indices error

Hm, I must be missing something: if we leave in e.g. a `mul` known to fail, but remove subsequent instructions, wouldn't that include the `array_get` or `array_set` as well that would cause the dynamic indices error?

jfecher: I was thinking for when we wanted to expand this check to include other instructions that are known to fail like arrayget/set with out of bounds indices

asterite: I implemented this. I think it also makes the `remove_unreachable_instructions` simpler as it's only job is to determine if an instruction will always fail, but doesn't replace instructions with other instructions. Let's see what CI says now...

asterite: Hmm... in the end I'm not sure about doing it this way. It's still a bit inconsistent. For example this code:

```noir
fn main(b: [u32; 1]) {
    func_1(3032078109, b);
}
fn func_1(a: u32, mut b: [u32; 1]) {
    b[0] = (b[0] % (869374236 - a));
}
```

In master it gives a compile bug:

```
bug: Assertion is always false: attempted to divide by constant larger than operand type: 254 > 32
  ┌─ src/main.nr:5:13
  │
5 │     b[0] = (b[0] % (869374236 - a));
  │             ---------------------- As a result, the compiled circuit is ensured to fail. Other assertions may also fail during execution
```

That's because eventually some operation is transformed to something else and that operation is guaranteed to fail. But with the latest change in this PR it doesn't because the subtraction is guaranteed to overflow so any subsequent instructions are removed.

I really think that if an operation is guaranteed to fail it should give a compile bug in all cases. Or, put another way, it should never say "compile bug". But if it's "sometimes yes, sometimes no" it's hard to set expectations... for users, and for our tests 😅 

For now I'll revert the last commit.

asterite:
Now one fuzz test fails with this:

```
DivisionByZero { lhs_id: Id(7), lhs: "-249012260880455945185240087545684409864", rhs_id: Id(8), rhs: "0" }
ConstrainEqFailed { lhs: "u1 false", lhs_id: Id(0), rhs: "u1 true", rhs_id: Id(1), msg: ", \"attempt to divide by zero\"" }
```

@aakoshh I guess we could consider both equal, what do you think?

aakoshh:
> I guess we could consider both equal,

Yes, I started adding these to https://github.com/noir-lang/noir/pull/9023/files#diff-a753c80e2ec2247a4783fe08945017ebebc906aac26aea4504813c6d8f01d7eb and https://github.com/noir-lang/noir/pull/9023/files#diff-169e9b287d9548e14121b798c589ea051cc12a9c8621fe4baa1c848acfb75f13

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
aakoshh: Just so I understand: on `master` we have a number of instructions after the `sub`, and one of them generates this `bug: `. When you removed those and left just the `sub` doomed to fail, the `bug: `disappears, to be replaced by a runtime failure. 

With the revert, even the `sub` will disappear, to be replaced with a constraint failure. Is that right? 
So even if we left both the `sub` and added the `constrain`, we wouldn't have the `bug: `?

tooling/ast_fuzzer/src/compare/interpreted.rs
```diff
@@ -192,6 +198,18 @@ impl Comparable for ssa::interpreter::errors::InterpreterError {
...
+                RangeCheckFailedWithMessage { message: msg1, .. },
+                ConstrainEqFailed { msg: msg2, .. },
+            ) => {
+                // The removal of unreachable instructions evaluates constant binary operations and can replace
+                // e.g. a `mul` followed by a `range_check` with a `constrain true == false, "attempt to multiple with overflow"`
+                msg2.contains(msg1)
+            }
+            (DivisionByZero { .. }, ConstrainEqFailed { msg, .. })
+            | (ConstrainEqFailed { msg, .. }, DivisionByZero { .. }) => {
+                msg == "attempt to divide by zero"
```
aakoshh: Two things to note:
1. With the `RangeCheckFailedWithMessage` above I only considered going from a the `range_check` to `constrain`, because it's not going to be reversed. 
2. The `msg` of `ConstrainEqFailed` and `ConstrainNeFailed` I think would actually be `", \"attempt to divide by zero\""`, which is why I used `msg2.contains(msg1)`.

aakoshh: This was an unfortunate hack I did in so `#[error("constrain {lhs_id} == {rhs_id}{msg} failed:\n    {lhs} != {rhs}")]` looks right, although I'm sure `this_error` has some trick to substitute a function call, so that the `msg` field can be nice and the display can be formatted.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite: Correct!

I just checked and in master we end up with this SSA:

```
  b0(v0: [u32; 1]):
    v2 = array_get v0, index u32 0 -> u32         	// src/main.nr:5:13
    v5 = sub u32 869374236, u32 3032078109        	// src/main.nr:5:21
    v6 = mod v2, v5                               	// src/main.nr:5:13
    return
}
```

and it seems it's that `mod` that eventually does a division that will always fail and triggers the `bug`. The failure message in this case is "attempted to divide by constant larger than operand type: 254 > 32" which is kind of cryptic 😅

tooling/ast_fuzzer/src/compare/interpreted.rs
asterite: Ah, yes, I remember seeing such commit. All good, it's not a big deal for now :-)

asterite:
CI is green 😮 🎉

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
aakoshh: Okay, but then does reverting to the previous commit or not make any difference?

asterite: Do you mean, if we leave the `sub`, remove every instruction after that and put an `unreachable` terminator? Then the program doesn't "bug" anymore, it only fails at runtime. That's what I didn't think it was right because previously it used to "bug".

aakoshh: Yes, and we just discussed that it does the same thing if you remove the `sub`. No “bug” either way

asterite: But in the state of the current PR the `sub` is removed and a `constrain` that always fails is added. That produces the "bug". Was that not desired?

aakoshh: I'm sorry, I didn't get that with the `sub` removed and `constrain` added we still have the "bug". I thought we only got it from the subsequent `mod`, which was removed in both cases. 

So just for my understanding, if we left the `sub` in place _and_ added a `constrain` after it, would we get a bug?

asterite: Aaaah... yes, in that case we'd still get a bug. I missed that possibility!

Is it worth changing it to that? I think it would be a bit redundant, though I understand it will make the interpreter errors more consistent... but I'm not sure.

aakoshh: I see that it’s not exactly straight forward with the way the simple block traversal routine works. I thought it would have been helpful when I was staring at a block that didn’t fit my screen, looking for which instruction is the last in the previous SSA that I can no longer see in the new one, to divine where the constant failure instruction was. Not a blocker, just wanted to throw it out there

aakoshh: Another argument I could bring is that this program:
```rust
fn main() {
    let _ = 1_u32 << 32;
}
```
Turns into this SSA:
```rust
❯ cargo run -q -p nargo_cli -- compile --show-ssa-pass Initial
After Initial SSA:
acir(inline) fn main f0 {
  b0():
    v2 = shl u32 1, u8 32                         	// src/main.nr:2:13
    constrain u1 0 == u1 1, "attempt to bit-shift with overflow"	// src/main.nr:2:13
    v5 = truncate v2 to 32 bits, max_bit_size: 33 	// src/main.nr:2:13
    return
}

bug: Assertion is always false: attempt to bit-shift with overflow
  ┌─ src/main.nr:2:13
  │
2 │     let _ = 1_u32 << 32;
  │             ----------- As a result, the compiled circuit is ensured to fail. Other assertions may also fail during execution
  │
  = Call stack:
    1. src/main.nr:2:13
```
So the `shl` is left there by the compiler, and it emits an always-failing `constrain` as well. It could do just the latter, but this way it's easier to understand. 

Not sure about which instruction is responsible for the `bug: ` in this case. Maybe we have to make it so that any always-fail `constrain` is showed up as a bug? 

The only counter argument I can think of is saving 1 instruction, but if it fails anyway and it's buggy, then I wouldn't worry too much about the bytecode size.

asterite: Sounds good, I'll try to make it work like that.

asterite: Done!

Issue #8910, opened 2025-06-13
ACIR != Brillig: both fail but differently


aakoshh:
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/8728 but unrelated

The following program fails differently in ACIR and Brillig. 

```rust
fn main(a: u32, b: u1) {
    assert((((b as u32) - 1619421681) == (a << 214)), "THE");
}
```

input:
```toml
a = "0x0000000000000000000000000000000000000000000000000000000012106f69"
b = "0x0000000000000000000000000000000000000000000000000000000000000001"
```

### Expected Behavior

They are both right to fail, but the fuzz tests assume that they should fail with the same message, which was also the assumption @vezenovm voiced in https://github.com/noir-lang/noir/issues/8615#issuecomment-2901953589 

I know @guipublic disagrees and thinks we should not expect equivalence between failures. 

### Bug

```console
❯ cargo run -q -p nargo_cli -- execute --silence-warnings
error: Assertion failed: attempt to bit-shift with overflow
  ┌─ src/main.nr:2:43
  │
2 │     assert((((b as u32) - 1619421681) == (a << 214)), "THE");
  │                                           --------
  │
  = Call stack:
    1. src/main.nr:2:43

Failed assertion

❯ cargo run -q -p nargo_cli -- execute --silence-warnings --force-brillig
error: Assertion failed: attempt to subtract with overflow
  ┌─ src/main.nr:2:14
  │
2 │     assert((((b as u32) - 1619421681) == (a << 214)), "THE");
  │              -----------------------
  │
  = Call stack:
    1. src/main.nr:2:14

Failed assertion

```

### To Reproduce

See above. 

The only difference I see in the SSA is in the way the "remove bit shifts" pass works on the two. 

The SSA before is identical:
```
After Flattening (1) (step 17):
acir(inline) predicate_pure fn main f0 {
  b0(v0: u32, v1: u1):
    v2 = cast v1 as u32
    v4 = sub v2, u32 1619421681
    v6 = shl v0, u8 214
    constrain u1 0 == u1 1, "attempt to bit-shift with overflow"
    constrain v4 == v6, "THE"
    return
}
```

In ACIR the `shl` is removed:
```
After Removing Bit Shifts (1) (step 18):
acir(inline) predicate_pure fn main f0 {
  b0(v0: u32, v1: u1):
    v2 = cast v1 as u32
    v4 = sub v2, u32 1619421681
[...]

PR #9028, opened 2025-06-26
feat(ssa): Handle `println` in the SSA interpreter

aakoshh:
# Description

## Problem\*

Resolves the missing implementation of `call_print` in the SSA interpreter and the inability to compare output between passes in the fuzz tests.

## Summary\*

Implements `call_print` in the SSA interpreter. It is disabled in the `Ssa::interpret` command, but enabled during `nargo interpret`. 

The AST fuzzer captures and compares the output in `pass_vs_prev`, which should give us more reassurance that if we consider different errors as equivalent (see the discussion in https://github.com/noir-lang/noir/issues/8910), then we do so without ignoring any potential difference in side effects.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_printable_type/src/lib.rs
+++ b/compiler/noirc_printable_type/src/lib.rs
@@ -334,8 +334,7 @@ pub fn decode_printable_value<F: AcirField>(
             // we want to consume the fields from the environment, but for now they are not actually printed
             let _env = decode_printable_value(field_iterator, env);
             let func_id = field_iterator.next().unwrap();
-            let func_ref = PrintableValue::Field(func_id);
-            func_ref
+            PrintableValue::Field(func_id)
         }
         PrintableType::Reference { typ, .. } => {
             // we decode the reference, but it's not really used for printing


compiler/noirc_evaluator/src/ssa/interpreter/intrinsics.rs
```diff
@@ -690,3 +772,73 @@ fn new_embedded_curve_point(
...
+            }
+            Value::Function(id) => {
+                fields.push(FieldElement::from(id.to_u32()));
+                // We should push the fields in the `env` of the function, but I'm not sure how to get them.
+            }
+            Value::Intrinsic(_) => {
+                todo!("convert intrinsic to fields")
+            }
+            Value::ForeignFunction(_) => {
+                todo!("convert foreign function to fields")
```
asterite: Should we capture issues for these?

aakoshh: I wouldn't know what to do with them to be honest, like, I don't even know if they can come up.

asterite: Makes sense. I think its fine to leave this like that then, if it eventually comes up (unlikely!) we can fix it.

aakoshh: I'll turn them into panic

aakoshh: Added a test for lambdas as well, which revealed that the SSA passes things in the opposite order as `decode_printable_value` expected for functions, although it didn't matter as nothing was displayed.

PR #9029, opened 2025-06-26
chore: don't compute used globals during DIE

asterite:
# Description

## Problem

No issue, just something I noticed while working on #9023

## Summary

Before this PR, the `Ssa` `used_globals` field was computed, and re-computed, on each DIE pass. I guess it was done like that because DIE also keeps track of used values, so the set of used globals ends up being the set of used values that are globals.

The main issue with the previous approach is that if a pass after the last DIE ended up reusing a global (`make_constant` can do that) then the of `used_globals` will be incorrect. To fix this we'd need to run DIE once again, but it's unclear that you have to do this. With this PR everything works fine automatically.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa.rs
+++ b/compiler/noirc_evaluator/src/ssa.rs
@@ -230,8 +230,6 @@ pub fn primary_passes(options: &SsaEvaluatorOptions) -> Vec<SsaPass> {
             .and_then(Ssa::remove_unreachable_functions),
         SsaPass::new(Ssa::dead_instruction_elimination, "Dead Instruction Elimination"),
         SsaPass::new(Ssa::array_set_optimization, "Array Set Optimizations"),
-        // The Brillig globals pass expected that we have the used globals map set for each function.
-        // The used globals map is determined during DIE, so we should duplicate entry points before a DIE pass run.
         SsaPass::new(Ssa::brillig_entry_point_analysis, "Brillig Entry Point Analysis")
             // Remove any potentially unnecessary duplication from the Brillig entry point analysis.
             .and_then(Ssa::remove_unreachable_functions),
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_globals.rs
```diff
@@ -443,11 +439,9 @@ mod tests {
         let ssa = Ssa::from_str(src).unwrap();
         // Need to run SSA pass that sets up Brillig array gets
         let ssa = ssa.brillig_array_get_and_set();
-        // Need to run DIE to generate the used globals map, which is necessary for Brillig globals generation.
-        let mut ssa = ssa.dead_instruction_elimination();
+        dbg!(ssa.used_globals_in_brillig_functions());
```
TomAFrench:
```suggestion
```

aakoshh:
Awesome! Could you remove DIE from here?

https://github.com/noir-lang/noir/blob/3c8f827c1aa63f2193904a3e99cea7bcb99d5ed9/compiler/noirc_evaluator/src/ssa.rs#L295

asterite:
@aakoshh Done! I also removed it from the main passes which was done for a similar reason.

There's still this comment which I think became outdated now but I don't know what should change:

```rust
        // The Brillig globals pass expected that we have the used globals map set for each function.
        // The used globals map is determined during DIE, so we should duplicate entry points before a DIE pass run.
        SsaPass::new(Ssa::brillig_entry_point_analysis, "Brillig Entry Point Analysis")
            // Remove any potentially unnecessary duplication from the Brillig entry point analysis.
```

asterite:
Hmmm... that change introduced regression. Maybe we can still run DIE but not because of `used_globals`?

aakoshh:
> @aakoshh Done! I also removed it from the main passes which was done for a similar reason.
> 
> There's still this comment which I think became outdated now but I don't know what should change:
> ```rust
>         // The used globals map is determined during DIE, so we should duplicate entry points before a DIE pass run.
>         SsaPass::new(Ssa::brillig_entry_point_analysis, "Brillig Entry Point Analysis")
> ```

As far as I understand that pass, it duplicates brillig functions so that the global addresses don't get mixed up. 
Because of that, DIE had to run after this, to correctly determine which globals are used in which entry point.
With the removal of the used globals routine from DIE, it shouldn't matter which order they run in.

aakoshh:
> that change introduced regression.

Which change exactly? The removal of DIE from the minimal pipeline? What is the regression?

asterite:
> Which change exactly? The removal of DIE from the minimal pipeline? What is the regression?

I went ahead and remove one DIE from the main pipeline which was done for the same purpose... or so I thought! I shouldn't have done that 😅 Now it's undone, I think there shouldn't be any regressions.

PR #9040, opened 2025-06-27
fix(ssa): Remove array from cache in constant folding if it's an argument to a `Call`

### LINKED ISSUE #9037: Min != Full: Global Array Modified
### Aim

Reproduction of the nightly failure in https://github.com/noir-lang/noir/actions/runs/15917440273/job/44897565113

The following program returns a different result for `--minimal-ssa` than the full pipeline:

```rust
global G_B: ([bool; 1],) = ([false],);
unconstrained fn main() -> pub [[bool; 1]; 1] {
    let _ = func_3(([true],), [G_B.0], (&mut ctx_limit));
    [G_B.0]
}
unconstrained fn func_3(a: ([bool; 1],), mut c: [[bool; 1]; 1], ctx_limit: &mut u32) -> bool {
    if ((*ctx_limit) > 0_u32) {
        if func_3(G_B, [a.0], ctx_limit) {
            c[0_u32] = {
                {
                    let mut idx_d: u32 = 0_u32;
                    loop {
                        if (idx_d == 3_u32) {
                            break
                        } else {
                            idx_d = (idx_d + 1_u32);
                            ()
                        }
                    }
                };
                a.0
            };
        }
    }
    true
}
```

### Expected Behavior

They should agree. 

We recently had https://github.com/noir-lang/noir/pull/9029 touch global calculations.

### Bug

```console
❯ cargo run -q -p nargo_cli -- execute --minimal-ssa --silence-warnings; cargo run -q -p nargo_cli -- execute --force --silence-warnings
[fuzz_testing] Circuit output: Vec([Vec([Field(0)])])
[fuzz_testing] Circuit output: Vec([Vec([Field(1)])])
```

### To Reproduce

See above. 

Notably before constant folding the SSA returns a fresh array made from the global, but after it returns the one it passed to the function:

```rust
After Purity Analysis (2) (step 21):
g0 = u1 0
g1 = make_array [u1 0] : [u1; 1]

brillig(inline) impure fn main f0 {
  b0():
    v2 = allocate -> &mut u32                     	// src/main.nr:3:30
    store u32 25 at v2                            	// src/main.nr:3:30
    v5 = make_array [u1 1] : [u1; 1]              	// src/main.nr:3:30
    inc_rc g1                                     	// src/main.nr:3:30
    v6 = make_array [g1] : [[u1; 1]; 1]           	// src/main.nr:3:30
    v8 = call f1(v5, v6, v2) -> u1                	// src/main.nr:4:13
    inc_rc g1                                     	// src/main.nr:4:13
    v9 = make_array [g1] : [[u1; 1]; 1]           	// src/main.nr:4:13
    return v9
}
...
After Constant Folding (1) (step 22):
g0 = u1 0
g1 = make_array [u1 0] : [u1; 1]

brillig(inline) impure fn main f0 {
  b0():
    v2 = allocate -> &mut u32                     	// src/main.nr:3:30
    store u32 25 at v2                            	// src/main.nr:3:30
    v5 = make_array [u1 1] : [u1; 1]              	// src/main.nr:3:30
    inc_rc g1                                     	// src/main.nr:3:30
    v6 = make_array [g1] : [[u1; 1]; 1]           	// src/main.nr:3:30
    v8 = call f1(v5, v6, v2) -> u1                	// src/main.nr:4:13
    inc_rc g1                                     	// src/main.nr:4:13
    inc_rc v6                                     	// src/main.nr:4:13
    return v6
}
...
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.7 noirc version = 1.0.0-beta.7+5480889c9c336c93cc6a6ff941004502f889bad9 (git version hash: 2203e14f139db2264a9a87dea3a4822d1397e6e9, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9037 

## Summary\*

Constant folding has a `remove_possibly_mutated_cached_make_arrays` method to remove `MakeArray` and `Call` instructions from the cache used to deduplicate instructions and reuse their results, if the array returned by those instructions is used in a `Store` or an `ArraySet`. That is, if we know we write to the array, we don't use it to replace an identical `MakeArray` call, or if we know that we store it in a reference, we treat it as opaque, like anything can happen to it after. 

This PR extends this to arrays passed to other functions as `Call` arguments. To avoid being too heavy handed, this is only done for Brillig functions, and only if the callee is an impure user function, which should help with e.g. `println`. 

## Additional Context

This rule is stricter than necessary. We could make it more permissive by tracking whether the array returned by the previous instructions has a positive reference count in the current block. If it does, then any modifications to it it should modify a copy, leaving the original safe to reuse. 

## Documentation\*

Check one:

# PR Checklist\*


aakoshh:
The type error seems to be `nargo expand` not rendering `([false], )` back as tuple, but rather just `([false])`
```
thread 'tests::nargo_expand_execution_success::test_regression_9037' panicked at /home/runner/.rustup/toolchains/1.85.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5:
Unexpected failure.
code=1
stderr=``````
error: Expected type ([bool; 1],), found type [bool; 1]
  ┌─ /tmp/.tmp76O4Rd/src/main.nr:1:28
  │
1 │ global G_B: ([bool; 1],) = ([false]);
  │                            ---------
  │

```
@asterite 

I'll change it here to add some dummy second element.

tooling/ast_fuzzer/src/program/mod.rs
```diff
@@ -572,9 +572,7 @@ impl std::fmt::Display for DisplayAstAsNoir<'_> {
         printer.show_type_in_let = true;
         // Most of the time it doesn't affect testing, except the comptime tests where
         // we parse back the code. For that we use `DisplayAstAsNoirComptime`.
-        // Using it also inserts an extra `cast` in the SSA,
-        // which at the moment for example defeats loop unrolling.
-        printer.show_type_of_int_literal = false;
+        printer.show_type_of_int_literal = true;
```
aakoshh:
```suggestion
        printer.show_type_of_int_literal = false;
```
So this another PR. Already setting it in Ary’s, but there it causes a test failure. Accidental check in

asterite:
@aakoshh Mind if I fix that nargo expand issue in this PR? Here's the diff to fix it:

```diff
diff --git a/tooling/nargo_expand/src/printer.rs b/tooling/nargo_expand/src/printer.rs
index 18d9d111b0..33d3db47a1 100644
--- a/tooling/nargo_expand/src/printer.rs
+++ b/tooling/nargo_expand/src/printer.rs
@@ -807,6 +807,9 @@ impl<'context, 'string> ItemPrinter<'context, 'string> {
                     }
                     self.show_value(value);
                 }
+                if values.len() == 1 {
+                    self.push(',');
+                }
                 self.push(')');
             }
             Value::Struct(fields, typ) => {
```

jfecher (approved):

jfecher:
Pulled this from the merge queue in case you wanted to apply the nargo expand patch

PR #9045, opened 2025-06-30
fix(fuzz): enable print in comptime_vs_brillig_direct

### LINKED ISSUE #8973: Allow the comptime_vs_brillig_direct target of the AST fuzzer to use `print`
### Problem

At the moment the faster comptime fuzzing target, `comptime_vs_brilig_direct` (introduced in https://github.com/noir-lang/noir/pull/8924) must avoid `print` statements, as the elaborator doesn't recognize them. 

### Happy Case

`print` statements should work.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

rkarabut:
# Description

Turns out to use print bypassing nargo we just need to bring in the part of stdlib correctly declaring print oracle harnesses.

## Problem\*

Resolves #8973 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/interpreter.rs
@@ -249,7 +249,7 @@ impl<'local, 'interner> Interpreter<'local, 'interner> {
     ) -> IResult<Value> {
         let attributes = self.elaborator.interner.function_attributes(&function);
         let func_attrs = &attributes.function()
-            .expect("all builtin functions must contain a function  attribute which contains the opcode which it links to").kind;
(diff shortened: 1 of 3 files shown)


TomAFrench:
huh, the fuzzing test doesn't seem to be a flake here. Not sure why it's seeing two imports of `print`.

aakoshh:
I think the reason it fails is because `comptime_vs_brillig_nargo` adds print on its own, while `comptime_vs_brillig_direct` doesn't, for some reason.

aakoshh:
So this doesn't need to happen for the traditional comptime fuzzing where it already worked, just the direct one, which bypasses certain steps. Maybe not loading the standard library is on of the things that makes it faster.

Issue #8874, opened 2025-06-10
SSA Interpreter `array_get` returns `Value::uninitialized()` when disabled while acir returns `array[0]`


jfecher:
When we have an `array_get` instruction with an unknown index in code with disabled side effects, acir-gen returns `array[0]` while the SSA interpreter returns `Value::uninitialized()`.

We should investigate to see if this can cause any differences between the two runtimes. One possible difference would be if we have an empty array, acir gen may fail but the interpreter would not.

jfecher:
@guipublic do you think there's any way we could have acir not perform `array[0]` for an array get in code where the predicate is false?

For example, the code:
```noir
// i = 5
// print = false
fn main(print: bool, i: u32) {
    let array: [u32; 0] = [];

    if print {
        println(array[i]);
    }
}
```
Gives:
```
error: Index out of bounds, array has size 0, but index was 0
   ┌─ src/main.nr:17:17
   │
17 │         println(array[i]);
   │                 --------
   │
   = Call stack:
     1. src/main.nr:17:17

Failed to solve program: 'Index out of bounds, array has size 0, but index was 0'
```
Is there a way we can get it to not error for arrays of length zero?

guipublic:
If the array length is known at compile time, and is 0, then we can safely replace the `array[0]` by `{assert(false); 0 }`
The assert being under the condition, the program will solve successfully when `print` is false.

jfecher:
@guipublic Hmm, yeah I suppose since any index would be out of bounds we could just always replace zero-length indexing with an assert false with the same error message.

PR #9050, opened 2025-06-30
chore: replace array operation on 0-len arrays with side effect assertion

guipublic:
# Description

## Problem\*

Related to #8874

## Summary\*
Array operations on 0-length arrays are replaced by an assertion on the 'side-effects-enabled'

## Additional Context
I tried to do it with a ssa pass, but it seemed overkill for an uncommon case (it cannot use the 'simple optimisation' pass) . I end up doing this transformation directly in acir-gen.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/acir/arrays.rs
+++ b/compiler/noirc_evaluator/src/acir/arrays.rs
@@ -105,6 +105,28 @@ impl Context<'_> {
             unreachable!("ICE: expected array or slice type");
         };
 
+        match &array_typ {
+            // Replace Array operations on 0-length arrays by asserting that enable_side_effects are false
+            Type::Array(_, len) => {
+                let zero_var = self.acir_context.add_constant(FieldElement::zero());
+                let result_ids = dfg.instruction_results(instruction);
+                if result_ids.len() == 1 && *len == 0 {
+                    let acir_typ: AcirType = dfg.type_of_value(result_ids[0]).into();
+                    let zero_value = AcirValue::Var(zero_var, acir_typ);
+                    self.define_result(dfg, instruction, zero_value);
+                    let msg = format!("Index out of bounds, array has size 0");
+                    let msg = self.acir_context.generate_assertion_message_payload(msg);
+                    return self.acir_context.assert_eq_var(
+                        self.current_side_effects_enabled_var,
+                        zero_var,
+                        Some(msg),
+                    );
+                }
+            }
+            Type::Slice(_) => (),
+            _ => unreachable!("ICE: expected array or slice type"),
+        }
+
         if self.handle_constant_index_wrapper(instruction, dfg, array, index, store_value)? {
             return Ok(());
         }
--- a/test_programs/execution_success/regression_8874/Nargo.toml
+++ b/test_programs/execution_success/regression_8874/Nargo.toml
@@ -1,0 +1,6 @@
+[package]
+name = "regression_8874"
+type = "bin"
+authors = [""]
+
+[dependencies]
--- a/test_programs/execution_success/regression_8874/Prover.toml
+++ b/test_programs/execution_success/regression_8874/Prover.toml
@@ -1,0 +1,2 @@
+print = false
+i = 5
--- a/test_programs/execution_success/regression_8874/src/main.nr
+++ b/test_programs/execution_success/regression_8874/src/main.nr
@@ -1,0 +1,7 @@
+fn main(print: bool, i: u32) {
+    let array: [u32; 0] = [];
+
+    if print {
+        println(array[i]);
+    }
+}


jfecher:
> I tried to do it with a ssa pass, but it seemed overkill for an uncommon case (it cannot use the 'simple optimisation' pass) . I end up doing this transformation directly in acir-gen.

Maybe in Instruction::simplify we can check if the array argument of ArrayGet/Set's length is 0, and if so replace it with an assert false there? Then again I think that wouldn't work either since the result would be a different type.

Per Tom's comment, the SSA interpreter is just meant to match the SSA so when we fix this we could apply it to the interpreter as well. E.g. return `array[0]` there unless the length is zero.

guipublic:
Indeed, that's why I said that this PR relates to #8874 instead of resolving it, because it did not resolve it.

I just updated the interpreter in order to have a similar behaviour than what acir-gen is doing.
I think it is not worth to try to get the exact same behaviour, but using a similar strategy is good nevertheless.
What I did is to use a valid index when the index is out-of-range and side-effects are disabled, except when the array length is zero, since there is no valid index in this case.
I believe that with this we can say that #8874 is resolved?

compiler/noirc_evaluator/src/acir/arrays.rs
```diff
@@ -105,6 +105,28 @@ impl Context<'_> {
             unreachable!("ICE: expected array or slice type");
         };
 
+        match &array_typ {
+            // Replace Array operations on 0-length arrays by asserting that enable_side_effects are false
+            Type::Array(_, len) => {
+                let zero_var = self.acir_context.add_constant(FieldElement::zero());
+                let result_ids = dfg.instruction_results(instruction);
+                if result_ids.len() == 1 && *len == 0 {
```
jfecher: Why is this only done if `result_ids.len() == 1`?
Can we not loop over all the results and zero them?

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
```diff
@@ -875,15 +875,33 @@ impl<'ssa, W: Write> Interpreter<'ssa, W> {
...
-            let index = self.lookup_u32(index, "array get index")?;
-            let index = index - offset.to_u32();
-            array.elements.borrow()[index as usize].clone()
-        } else {
+        let array = self.lookup_array_or_slice(array, "array get")?;
+        let index = self.lookup_u32(index, "array get index")?;
+        let mut index = index - offset.to_u32();
+        // An array_get with false side_effects_enabled is replaced
+        // by a load at a valid index during acir-gen.
+        if !side_effects_enabled && array.elements.borrow().len() <= index as usize {
```
jfecher: It looks like this still doesn't quite match acir. Acir does:
```rust
if array length is 0 {
    assert(!predicate)
} else {
    array[i * predicate]
}
```
Where this does:
```rust
if !predicate && array len <= index {
    if no array elements {
        define(result, 0);
        return;
    }
    index = find_index_with_matching_type(array, result_type);
}
array[index]
```

jfecher (changes requested):

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
guipublic: Indeed, in a previous comment I said:
> I think it is not worth to try to get the exact same behaviour

The behaviour in acir-gen is quite complex, but in the case of an empty array it will be similar: use 0 value is predicate is false, and fail if the predicate is true.
Should I replicate the full behaviour?

jfecher: I think it would be better to try to replicate it more closely, yes. Unless there is something more fundamental that prevents us from doing so

compiler/noirc_evaluator/src/acir/arrays.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
guipublic: The problem is that acir-gen behaviour depends on constant vs non-constant values, while the ssa interpreter depends on ssa values, which can become constant on a later stage (e.g after optimisations), so the behaviour would not be identical depending on which ssa representation is used.

jfecher: Can you explain how acir-gen depends on the const-ness of values? For example, I'm wondering where my mental model of:
```rs
if array length is 0 {
    assert(!predicate)
} else {
    array[i * predicate]
}
```
goes wrong

jfecher: E.g. can we change the interpreter to do the above with the length check == 0 then we assert the predicate is false, else we can do the i * predicate behavior?

jfecher:
related https://github.com/noir-lang/noir/pull/9195

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
guipublic: The complications in acir-gen are due to the flatenning of arrays, which we do not need to do here, so yes I agree it is simpler and we can use your suggestion.
I updated the code.

guipublic: In fact some flattening pass will also flatten the arrays in SSA, so we can't just do i*predicate. I restored the search for a compatible index.

jfecher (approved):
Thank you!

jfecher:
@TomAFrench need a re-review here

PR #9052, opened 2025-06-30
chore: move a couple of test_utils functions into a module

asterite:
# Description

## Problem

The current code would give a warning saying that the `InterpreterError` import was unused, because it was only used if `test_utils` was enabled. I considered changing the usage to a fully-qualified path but maybe it's better to put everything inside its own module so that imports need to be defined again and they don't clash with the non-gated module.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/mod.rs
+++ b/compiler/noirc_frontend/src/elaborator/mod.rs
@@ -24,7 +24,7 @@ use crate::{
     graph::CrateId,
     hir::{
         Context,
-        comptime::{ComptimeError, InterpreterError},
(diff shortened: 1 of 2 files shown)


compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -24,7 +24,7 @@ use crate::{
     graph::CrateId,
     hir::{
         Context,
-        comptime::{ComptimeError, InterpreterError},
```
asterite, line 27: This was the unused import.

PR #9055, opened 2025-06-30
chore!: remove protobuf

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

Usage of protobuf was rejected from barretenberg so there's no real point in continuing to support this. It is also causing flakes in CI so I think it's sensible to remove this code.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -66,7 +66,7 @@ dependencies = [
  "bn254_blackbox_solver",
  "brillig_vm",
  "criterion",
- "indexmap 2.10.0",
+ "indexmap 2.11.0",
  "insta",
  "num-bigint",
  "pprof",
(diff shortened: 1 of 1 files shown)


aakoshh:
For the record, my suggestion for how we could use this was to use [prost-reflect](https://crates.io/crates/prost-reflect) to achieve the following:
* Have our Noir models as pure rust
* Generate DTO model from the protobuf schemas with prost, as we do today
* Map the Noir models to the protobuf DTOs
* Serialise the protobuf DTOs to the protobuf data structures used by prost-reflect
* Traverse those data structures and map them to msgpack, using the protobuf field IDs instead of strings for field names, for a more compact, but backwards compatible format
* Generate the msgpack deserialisation code based on the protobuf descriptors, instead of the serde reflect

This would achieve:
* Separation of Noir model from DTOs
* Compact serialisation that doesn't need to serialise strings
* Backwards compatibility based on the protobuf schema evolution rules 

It wouldn't achieve Adam's vision of immutability, which is that if an opcode needs to change then we instead need to introduce a new opcode.

TomAFrench:
Marking draft for me to have a think about the above.

TomAFrench:
Protobuf is causing significant issues with CI stability atm and I think that if we go down the above route then we can pull this work out of the git history relatively easily.

aakoshh:
I thought I added property based testing for serialisation but can't find it now 👀 
Would be nice to keep that for the msgpack stuff.

Found [them](https://github.com/noir-lang/noir/pull/9055/changes#diff-14705c2518812ea77f89edb9719a79ac1876e404e3a3a931ff75a10fdcf650cb) 👍

aakoshh (dismissed):
👋 🥹

PR #9060, opened 2025-06-30
fix(ssa): Do not fail for static assertions in a map over empty arrays

### LINKED ISSUE #7710: `static_assert`/`assert_constant` fail in empty loops over empty dynamic array as slice
### Aim

Attempted to ensure that code in the loop body of empty arrays/slices is never run:

```noir
fn main(x: [Field; 0]) {
    x.as_slice().map(|_| {
        // these work:
        assert(false, "test");
        panic(f"test");

        // these fail:
        // std::static_assert(false, "test");
        // assert_constant(x);
    });

    for _ in x.as_slice() {
        // these work:
        assert(false, "test");
        panic(f"test");

        // these fail:
        // std::static_assert(false, "test");
        std::assert_constant(x);
    }

    // these all work:

    x.map(|_| {
        std::static_assert(false, "test");
        assert(false, "test");
        panic(f"test");
        assert_constant(x);
    });

    for _ in x {
        std::static_assert(false, "test");
        assert(false, "test");
        panic(f"test");
        assert_constant(x);
    }

    let empty_array: [Field; 0] = [];
    for _ in empty_array.as_slice() {
        std::static_assert(false, "test");
        assert(false, "test");
        panic(f"test");
        assert_constant(x);
    }

    for _ in 0..0 {
        std::static_assert(false, "test");
        assert(false, "test");
        panic(f"test");
        assert_constant(x);
    }
}
```

### Expected Behavior

Expected compilation and execution to succeed.

### Bug

`static_assert` and `assert_constant` each fail when looping over `[].as_slice()` when the `[]` is dynamic:

```bash
❯ ~/.nargo/bin/nargo execute
error: Argument is not constant
   ┌─ src/main.nr:19:9
   │
19 │         std::assert_constant(x);
   │         -----------------------
   │
   = Call stack:
     1. src/main.nr:19:9

```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.3 noirc version = 1.0.0-beta.3+c30bcd7e753d835442187bb27a6c14231041ed77 (git version hash: c30bcd7e753d835442187bb27a6c14231041ed77, is dirty: false)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #7710 

## Summary\*

This change ended up being more involved than expected:
1. If we are in an empty loop (e.g. the upper bound is not greater than the lower bound), we skip evaluation a static assertion and do not insert it back into the SSA. For now I chose to still error out if the static assertion is inside of a dynamic loop. However, this may not be sufficient on its own, perhaps we should add a warning to not place static assertions under dynamic loops.
2. With non-aggressive inlining we can have loop bounds from function parameters that would otherwise resolve to constants if the function was inlined. For example, the regression test added in this PR would fail with the minimum inliner aggressiveness and only the change from (1). I chose to simply always inline functions which contains calls to static assertions.
3. We can still have nested functions calls where we pass along a constant call argument through many calls. To fully handle (2) I switched inlining to operate with a feedback loop until a fixed point. This type of logic is common in modern compilers and is something we already do in multiple SSA passes.

## Additional Context

A full fix to handling constants call arguments would be https://github.com/noir-lang/noir/issues/8321. However, this is quite an involved change so I decided (2) and (3) above provided a nice way to iteratively fix this bug and propagate constants for static assertions.  

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/inlining.rs
```diff
@@ -51,16 +51,42 @@ impl Ssa {
...
+            let num_functions_before = self.functions.len();
+
+            let call_graph = CallGraph::from_ssa_weighted(&self);
+            let inline_infos = compute_inline_infos(
+                &self,
+                &call_graph,
+                inline_no_predicates_functions,
+                aggressiveness,
+            );
+            let contains_static_assertion =
```
vezenovm: I gated re-inlining as to avoid large regressions in the test suite durations. e.g. we see increases like such:

> test_report_AztecProtocol_aztec-packages_noir-projects_aztec-nr | 101 s | 75 s | 1.35
> -- | -- | -- | --
> test_report_AztecProtocol_aztec-packages_noir-projects_noir-contracts | 150 s | 127 s | 1.18

Unfortunately, this looks to have not reverted the timing regressions. So I would like to remove this extra logic. 

However, with this check we saw one test suite decrease in time:
> test_report_zkpassport_noir-ecdsa_ | 93 s | 115 s | 0.81

I leave this logic in place for the reviewer, but I lean towards removing this check. If we are unhappy with these timing regressions I can investigate other options to reduce compilation time increases from this feedback loop.

vezenovm: Another option could be to just make `contains_static_assertion` more strict and have it only apply when the call is within a loop. As this is already an iterative fix perhaps this is the best option as to not increase compilation for common cases.

vezenovm: Perhaps these test suite timings are off... some additional context https://github.com/noir-lang/noir/pull/9064#issuecomment-3020489459. Will investigate further tomorrow.

TomAFrench:
merging master to run benchmarks again

vezenovm:
> merging master to run benchmarks again

Running once more without the re-inlining gate https://github.com/noir-lang/noir/pull/9060#discussion_r2175661673 after https://github.com/noir-lang/noir/pull/9069

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
vezenovm: https://github.com/noir-lang/noir/pull/9060#issuecomment-3023269658

vezenovm: I ended up leaving gating of re-inlining out of this feedback loop. The times mostly look ok except for some of the test suite (although some looked to have decreased in time). If the reviewers are unhappy with these timing regressions I can further experiment with running other optimizations passes between inlining re-runs.

TomAFrench:
@vezenovm Looks like there's still issues with the merge.

compiler/noirc_evaluator/src/ssa/opt/inlining/inline_info.rs
```diff
@@ -158,22 +160,37 @@ fn compute_function_should_be_inlined(
         return; // Already processed
     }
 
+    let function = &ssa.functions[&func_id];
+    let contains_static_assertion = function.reachable_blocks().iter().any(|block| {
+        function.dfg[*block].instructions().iter().any(|instruction| {
+            let assert_constant_id = function.dfg.get_intrinsic(Intrinsic::AssertConstant).copied();
+            let static_assert_id = function.dfg.get_intrinsic(Intrinsic::StaticAssert).copied();
```
TomAFrench:
```suggestion
        let assert_constant_id = function.dfg.get_intrinsic(Intrinsic::AssertConstant).copied();
        let static_assert_id = function.dfg.get_intrinsic(Intrinsic::StaticAssert).copied();
        function.dfg[*block].instructions().iter().any(|instruction| {
```

Swapping the order of these so that we don't look up these function ids on every instruction (we could pull it out of the outer loop as well but :shrug:.

compiler/noirc_evaluator/src/ssa/opt/assert_constant.rs
```diff
@@ -68,14 +97,27 @@ impl Function {
...
+        return Ok(true);
+    }
+
     match &function.dfg[instruction] {
         Instruction::Call { func, arguments } => {
-            if *func == assert_constant_id {
+            let is_assert_constant = Some(*func) == assert_constant_id.copied();
+            let is_static_assert = Some(*func) == static_assert_id.copied();
+
+            // Skip known assertions inside empty loops
```
TomAFrench:
```suggestion
            // Skip assertions inside known empty loops
```

This sentence seems weird. Is this what you meant?

compiler/noirc_evaluator/src/ssa/opt/assert_constant.rs
```diff
@@ -169,3 +211,112 @@ fn append_foreign_call_param(
...
+          b3():
+            v11 = load v7 -> u32
+            v12 = load v9 -> [Field]
+            return
+        }
+        brillig(inline) fn foo f1 {
+          b0():
+            v0 = make_array [] : [Field; 0]
+            return v0
+        }
```
TomAFrench: Seems like we can minimise this more. A lot of the code is unrelated to what we're testing.

compiler/noirc_evaluator/src/ssa/opt/assert_constant.rs
```diff
@@ -41,14 +45,39 @@ impl Function {
     pub(crate) fn evaluate_static_assert_and_assert_constant(
         &mut self,
     ) -> Result<(), RuntimeError> {
+        let loops = Loops::find_all(self);
+
+        let cfg = ControlFlowGraph::with_function(self);
+        let mut blocks_within_empty_loop = HashSet::default();
+        for loop_ in loops.yet_to_unroll {
+            let Ok(pre_header) = loop_.get_pre_header(self, &cfg) else {
+                // If the loop does not have a preheader we skip hoisting loop invariants for this loop
```
TomAFrench: Looks like this comment has been copy-pasted from elsewhere and no longer is relevant.

TomAFrench (approved):
Looks good, there's a few nits plus a bad merge.

compiler/noirc_evaluator/src/ssa/opt/assert_constant.rs
vezenovm: Yeah, that is what I meant.

compiler/noirc_evaluator/src/ssa/opt/assert_constant.rs
vezenovm: I will remove the unneeded code.

compiler/noirc_evaluator/src/ssa/opt/inlining/inline_info.rs
vezenovm: I ended up doing this manually as we can pull it out of the outer loop. I was hoping rust would be smart enough to pull it out, but better to just do it.

compiler/noirc_evaluator/src/ssa/opt/assert_constant.rs
vezenovm: I updated to reflect what we skip now.

Issue #9047, opened 2025-06-30
Min != Full: Should never execute, yet degenerates into always-fail


aakoshh:
### Aim

Replace one of the nightly failures from https://github.com/noir-lang/noir/actions/runs/15963166617/job/45018853663

The following program succeeds with `--minimal-ssa` but fails with the default pipeline:
```rust
fn main() -> pub str<2> {
    if unsafe { func_5_proxy() } {
        for idx_d in -179503052_i32..-179503049_i32 {
            assert(((idx_d * idx_d) == -1301407079_i32), "NUX");
        }
    };
    "ok"
}

unconstrained fn func_5_proxy() -> bool {
    false
}
```

### Expected Behavior

They should agree. 

### Bug

```console
❯ cargo run -q -p nargo_cli -- execute --silence-warnings; cargo run -q -p nargo_cli -- execute --silence-warnings --minimal-ssa
error: Assertion failed: attempt to multiply with overflow
  ┌─ src/main.nr:4:22
  │
4 │             assert(((idx_d * idx_d) == -1301407079_i32), "NUX");
  │                      -------------
  │
  = Call stack:
    1. src/main.nr:4:22

Failed assertion
[fuzz_testing] Circuit witness successfully solved
[fuzz_testing] Witness saved to target/fuzz_testing.gz
[fuzz_testing] Circuit output: String("ok")
```

Looking at the SSA, it looks like `mul` was not supposed to overflow right there and then, but https://github.com/noir-lang/noir/pull/9023 treats it like so.

```
After Mem2Reg (4) (step 30):
acir(inline) predicate_pure fn main f0 {
  b0():
    v1 = call f1() -> u1                          	// src/main.nr:2:17
    v3 = mul i32 4115464244, i32 4115464244       	// src/main.nr:4:22
    v4 = cast v3 as u64                           	// src/main.nr:4:22
    v5 = truncate v4 to 32 bits, max_bit_size: 64 	// src/main.nr:4:22
    v6 = cast v1 as Field                         	// src/main.nr:4:22
    v8 = mul Field 32221345677314704, v6          	// src/main.nr:4:22
    range_check v8 to 32 bits, "attempt to multiply with overflow"	// src/main.nr:4:22
    constrain u1 0 == v1, "attempt to add with overflow"	// src/main.nr:4:22
    v10 = cast v5 as i32                          	// src/main.nr:4:22
    v12 = mul i32 4115464245, i32 4115464245      	// src/main.nr:4:22
[...]

PR #9061, opened 2025-06-30
chore(fuzz): Print AST with numeric type suffixes

aakoshh:
# Description

## Problem\*

We disabled generating ASTs with large or negative integers because replicating failures in Noir with `nargo` was extremely cumbersome. Now that we have numeric suffixes though, we can print code with those and match the AST unambiguously. 

## Summary\*

* Change `DisplayAstAsNoir` to print numeric type suffixes.
* Changes fuzz targets to enable the types that can be make unambiguous with suffixes.
* Omit the type suffix for `Field` literals, since they are the default and it's a bit noisy.

Some tests seem fail with overflows now, potentially related to https://github.com/noir-lang/noir/issues/9047

## Documentation\*

Check one:

# PR Checklist\*


tooling/ast_fuzzer/fuzz/src/targets/comptime_vs_brillig_direct.rs
```diff
@@ -16,16 +16,14 @@ use noir_ast_fuzzer::rewrite::change_all_functions_into_unconstrained;
...
-        // Also avoid negative integers, because the frontend rejects them for loops.
-        avoid_negative_int_literals: true,
         // Avoid break/continue
         avoid_loop_control: true,
         // Has to only use expressions valid in comptime
         comptime_friendly: true,
         // Force brillig, to generate loops that the interpreter can do but ACIR cannot.
         force_brillig: true,
+        // Allow overflows half the time.
+        avoid_overflow: u.arbitrary()?,
```
rkarabut, line 26: Hmm what is the reasoning for doing it half the time? Why not just allow them?

aakoshh: I did this in all tests thinking that an arithmetic overflow is so easy to trigger that with enough arithmetic thrown into a program, we might end up with nothing but errors. So, half the time we avoid those, to allow other bugs to shine.

PR #9063, opened 2025-06-30
chore: error on infinite recursion

### LINKED ISSUE #9033: Replace RECURSION_LIMIT panic in ssa inlining with an actual error
### Aim

Replace https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/inlining.rs#L233 with a returned Err variant

### Expected Behavior

Nargo should not panic for any user code

### Bug

Nargo may panic for infinitely (or just too-largely) recursive user code. We should return an error instead.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #9033 

## Summary\*
Returns a Runtime Error when reaching recursion limit.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/inlining.rs
```diff
@@ -91,13 +95,20 @@ impl Ssa {
...
-            let new_function = function.inlined(&self, &should_inline_call);
-            (entry_point, new_function)
-        });
+            let inlined = function.inlined(&self, &should_inline_call);
+            if let Ok(inlined_function) = inlined {
+                new_functions.insert(entry_point, inlined_function);
+            } else {
+                let err = inlined.expect_err("");
+                return Err(err);
+            }
```
asterite:
```suggestion
            let inlined = function.inlined(&self, &should_inline_call)?;
            new_functions.insert(entry_point, inlined);
```

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
```diff
@@ -996,4 +1008,24 @@ mod test {
...
+        }
+        acir(inline) fn foo f1 {
+        b0(v0: u32):
+            v2 = call f1(v0) -> u32
+            v3 = add v2, v0
+            return v3
+        }";
+        let ssa = Ssa::from_str(src).unwrap();
+        let ssa = ssa.inline_functions(i64::MIN);
+        assert!(ssa.is_err());
```
asterite: Mainly so that the error is clear on this test:

```suggestion
        let Err(err) = ssa else {
            panic!("Expected an error");
        };
        assert_snapshot!(err.to_string(), @"Attempted to recurse more than 1000 times during inlining function 'foo'");```

jfecher: Instead of `let Err(err) = ssa else { panic!("...") }` I assume we can also do `let err = ssa.unwrap_err();`

asterite: I tried that first but it requires `Ssa` to implement `Debug` (which we could too)

jfecher: Ah, right. An alternative I usually use in that case is `unwrap_or_else` then panic inside... maybe slightly simpler. We could also just implement Debug for Ssa though and defer to the Display impl

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
guipublic: Done, I used the first suggestion because `unwrap_or_else` would required to return a dummy ssa object and also to handle the proper path in the error case, which seems lees clear to me.

jfecher: I was suggesting `.unwrap_or_else(|| panic!(...)` which should be equivalent but this  way is fine too

PR #9064, opened 2025-06-30
chore: Switch to rustc-hash

jfecher:
# Description

## Problem\*

We depend on `fxhash` which is no longer maintained and has been superseded by `rustc-hash`

## Summary\*

This PR:
- Replaces uses of `fxhash` with `rustc-hash`
- Adds `util::hash64` to provide a utility function missing from `rustc-hash` which we use in several crates.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
Seems like some fairly large test suite duration regressions

vezenovm:
> Seems like some fairly large test suite duration regressions

Some of these times such as for aztec-nr and types seem to line up with what I see in my PR https://github.com/noir-lang/noir/pull/9060#pullrequestreview-2971872209. Makes me think that perhaps something else is up.

vezenovm:
> Some of these times such as for aztec-nr and types seem to line up with what I see in my PR [#9060 (review)](https://github.com/noir-lang/noir/pull/9060#pullrequestreview-2971872209). Makes me think that perhaps something else is up.

Yeah something is definitely up with the alerts https://github.com/noir-lang/noir/pull/9068#pullrequestreview-2974247123. I am very surprised to see an alert on #9068. cc @TomAFrench

TomAFrench:
I was just taking a look at this. I'm not entirely sure what's up as there's been no bumps to the pinned commit in the period from when these branches were made.

TomAFrench:
Huh, actually it looks like it's using a commit from ~2 weeks ago as the baseline.

TomAFrench:
@vezenovm the fix is in https://github.com/noir-lang/noir/pull/9069

Cargo.toml
```diff
@@ -41,12 +41,13 @@ members = [
...
     # Utility crates
     "utils/iter-extended",
+    "utils/hash64",
 
     "tooling/ssa_fuzzer",
     "tooling/ssa_fuzzer/fuzzer",
     "tooling/ssa_verification",
     "tooling/ssa_executor",
-    "utils/protobuf",
+    "utils/protobuf", "utils/hash64",
```
TomAFrench: This is doubled.

utils/hash64/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+/// Utility to return the 64-bit hash of a value using rustc-hash.
+pub fn hash64<T: std::hash::Hash + ?Sized>(v: &T) -> u64 {
+    use std::hash::BuildHasher;
+    rustc_hash::FxBuildHasher.hash_one(v)
+}
```
TomAFrench, line 5: We can get this down to an import + one-liner if we use `BuildHasher`. I'm not clear on the benefit of adding an extra crate for this tbh, especially when we're going to end up needing to publish it at some point just to be able to publish the other crates.

jfecher: Ah - I thought there was such a function but couldn't find it. I'll remove the util crate.

TomAFrench:
Closing this as I've made this change in #9750

PR #9001, opened 2025-06-23
feat(debug): Print ssa locations along with ssa

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/8983

## Summary\*

Prints the top (last) function in the callstack and the source line along with each SSA instruction to help finding where instructions originate from when printing larger files.

This requires we pass a `FileManager` when printing SSA now if you want the debug comments. For ease of debugging, you can still print SSA without this though and we just won't print out the source lines.

## Additional Context

Before this PR (lambda_from_dynamic_if test):

```
After Dead Instruction Elimination - ACIR (1) (step 45):
acir(inline) impure fn main f0 {
  b0(v0: u32):
    v2 = eq v0, u32 0
    v3 = not v2
    v5 = eq v0, u32 1
    v6 = unchecked_mul v3, v5
    v7 = not v5 // src/main.nr:12
    v8 = unchecked_mul v3, v7
    v10 = make_array b"!"
    constrain u1 0 == v8, data v10, Field 0
    v13 = cast v6 as Field
    v15 = mul v13, Field 9
    enable_side_effects u1 1
    v17 = cast v2 as Field
    v18 = cast v3 as Field
    v20 = mul v17, Field 8
    v21 = mul v18, v15
    v22 = add v20, v21
    v23 = eq v22, Field 8
    v26 = make_array b"hi"
    enable_side_effects v23
    call f3(u1 1, v26)
    v28 = not v23
    v29 = eq v22, Field 9
    v30 = unchecked_mul v28, v29
    v34 = make_array b"bye"
    enable_side_effects v30
    ...
```

After this PR:

```
After Dead Instruction Elimination - ACIR (1) (step 45):
acir(inline) impure fn main f0 {
  b0(v0: u32):
    v2 = eq v0, u32 0           // src/main.nr:10
    v3 = not v2         // src/main.nr:10
    v5 = eq v0, u32 1           // src/main.nr:12
    v6 = unchecked_mul v3, v5           // src/main.nr:12
    v7 = not v5         // src/main.nr:12
    v8 = unchecked_mul v3, v7           // src/main.nr:12
    v10 = make_array b"!"               // src/main.nr:12
    constrain u1 0 == v8, data v10, Field 0             // std/panic.nr:2
    v13 = cast v6 as Field
    v15 = mul v13, Field 9
    enable_side_effects u1 1
    v17 = cast v2 as Field
    v18 = cast v3 as Field
    v20 = mul v17, Field 8
[...]

PR #9066, opened 2025-06-30
chore(debug): Toggle SSA locations when printing SSA 

vezenovm:
# Description

## Problem\*

No issue, just something I would like post https://github.com/noir-lang/noir/pull/9001.

Sometimes I like to write small programs in Noir so that I can print their SSA and then transfer it to SSA unit tests. To do this always having SSA locations adds some extra friction, especially when indents don't perfectly line up. 

## Summary\*

I simply added a new compiler options `no_ssa_locations` to determine whether we pass a file manager during `compile_no_check`. 

I debated expanding our `SsaLogging` enum to a struct to contain more settings, but as this print detail was already being toggled through an optional file manager I felt adding a compile option was simpler and touched less code. 

Running on `test_programs/assert_statement`:
The default `nargo compile --show-ssa`:
```
  b0(v0: Field, v1: Field):
    constrain v0 == v1, "x and y are not equal"         // src/main.nr:5:12
    return
}
```
With `nargo execute --show-ssa --no-ssa-locations`:
```
  b0(v0: Field, v1: Field):
    constrain v0 == v1, "x and y are not equal"
    return
}
```

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_driver/src/lib.rs
+++ b/compiler/noirc_driver/src/lib.rs
@@ -79,6 +79,11 @@ pub struct CompileOptions {
     #[arg(long, hide = true)]
     pub show_ssa_pass: Vec<String>,
 
+    /// Do not emit source file locations when emitting debug information for the SSA IR to stdout.
+    /// By default, source file locations will be shown.
+    #[arg(long)]
+    pub no_ssa_locations: bool,
+
     /// Only show the SSA and ACIR for the contract function with a given name.
     #[arg(long, hide = true)]
     pub show_contract_fn: Option<String>,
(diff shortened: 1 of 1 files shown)


compiler/noirc_driver/src/lib.rs
```diff
@@ -79,6 +79,11 @@ pub struct CompileOptions {
     #[arg(long, hide = true)]
     pub show_ssa_pass: Vec<String>,
 
+    /// Do not emit source file locations when emitting debug information for the SSA IR to stdout.
+    /// By default, source file locations will be shown.
+    #[arg(long)]
+    pub no_ssa_locations: bool,
```
jfecher:
```suggestion
    #[arg(long, hide = true)]
    pub no_ssa_locations: bool,
```
Since `show_ssa` is hidden, this should be too

vezenovm: Ah yes I overlooked that `show_ssa` is hidden.

PR #9072, opened 2025-07-01
fix(fuzz): Do not access arrays with `&mut` inside `if` using dynamic condition

### LINKED ISSUE #9035: DIE post-check crash: `load` left after mem2reg because of dynamic indexing
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/9023

The following program crashes the compiler on the DIE post check added in #9018 

```rust
fn main(mut a: bool) {
    for idx_b in 1..3 {
        let c: [&mut bool; 3] = [(&mut true), (&mut false), (&mut true)];
        for idx_d in 1..3 {
            if a {
                a = (*c[(if (*c[0_u32]) { idx_b } else { idx_d })]);
            }
            println(a);
        }
    }
}
```

### Expected Behavior

The assumption was that all `load` instructions will be eliminated after `mem2reg`, and if not that's a bug.

### Bug

```console
$ cargo run -q -p nargo_cli -- compile --show-ssa --force
Message:  not expected to have Load or Store instruction after DIE in an ACIR function: main f0 / b0 / 13: Load { address: Id(19) }
```

If I remove the post-check, we get a different error:
```console
error: Only constant indices are supported when indexing an array containing reference values
  ┌─ src/main.nr:7:27
  │
7 │                     a = (*c[(if (*c[0_u32]) { idx_b } else { idx_d } % 3_u32)]);
  │                           ----------------------------------------------------
  │
  = Call stack:
    1. src/main.nr:7:27
```

The problem goes away if we either if we remove the `if a`.

### To Reproduce

See above.

### Workaround

None

### Workaround Description

I guess I can just remove the post-check assertion, or move it to the end of the SSA pipeline like the verification of no dynamic indexing, but first we should investigate why it's considered dynamic.

I might also have to tweak the fuzzer to avoid generating such code. So far dynamic indexing meant: 1) not using input parameters 2) not using values returned form Brillig and 3) not using values that can overflow.

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.7 noirc version = 1.0.0-beta.7+5480889c9c336c93cc6a6ff941004502f889bad9 (git version hash: 2203e14f139db2264a9a87dea3a4822d1397e6e9, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9035 

## Summary\*

* Add an `in_dynamic` flag to the AST generator which is turned on when we are inside the _then_ or the _else_ of an `if` statement where the _condition_ is using a dynamic variable. Under such circumstances anything we do inside is under some predicate that will tie the outcome to the circuit inputs in ACIR. 
* Do not use arrays with references in them if we are in an `in_dynamic` region.
* Move the DIE post-check to happen as the last step of the SSA pipeline, to let the dynamic index access verifier provide better feedback than what the panic does.

Waiting on #9047

## Additional Context

Instead of panicking on leftover loads, we can defer it to runtime using https://github.com/noir-lang/noir/pull/8943
Or we can keep panicking and try to turn those panics into better user errors, like the `verify_no_dynamic_indices_to_references`, once we have examples of them.

## Documentation\*

Check one:

# PR Checklist\*


tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -337,6 +341,11 @@ impl<'a> FunctionContext<'a> {
         self.dynamics.contains(id)
     }
 
+    /// Check if a type can be used inside a dynamic input context.
+    fn can_be_used_in_dynamic(&self, typ: &Type) -> bool {
+        self.unconstrained() || !(types::contains_reference(typ) && types::is_array_or_slice(typ))
```
vezenovm: What if we have a tuple that contains an array of references?

aakoshh: Thanks, you're right, I had to add filtering to another method which derives a target type from a source expression. This way it's okay to use the tuple with the array of references, as long as we don't traverse into the array.

tooling/ast_fuzzer/src/program/func.rs
```diff
@@ -337,6 +341,16 @@ impl<'a> FunctionContext<'a> {
...
     }
 
+    /// Check if a source type can be used inside a dynamic input context to produce some target type.
+    fn can_be_used_in_dynamic(&self, src_type: &Type, tgt_type: &Type) -> bool {
+        // Dynamic inputs are restricted only in ACIR
+        self.unconstrained()
+        // If we are looking for the exact type, it's okay.
+            || src_type == tgt_type
+            // But we can't index an array with references.
+            || !(types::is_array_or_slice(src_type) && types::contains_reference(src_type) )
```
asterite, line 375: Is `tgt_type` the index? Or why `tgt_type` isn't used in this line?

asterite (approved):
Looks good! I left a question but it's likely me misunderstanding.

tooling/ast_fuzzer/src/program/func.rs
aakoshh: This can be confusing without seeing the context in which it’s used. 

So we know that type source type can be used to produce a set of other target types, for example an array of tuples can produce itself, the tuple, or any of the tuple member types, etc. 

Here we check whether we can dig down to one of those produced target types, starting from the source, and say that if we are dealing with an array that contains references, then we can’t. 

If the condition was on the target type, then we would say it’s okay to try to get a tuple member from the array of tuple references, if that member itself is eg. a primitive type, even if we have to go through the refs to get it, but this isn’t the case. 

It might actually be more restrictive than necessary, like maybe it would be safe to access an array that has a tuple with a primitive and a reference member, as long as the reference part isn’t touched.

asterite: Thanks!

PR #9059, opened 2025-06-30
fix: mark all signed binary ops as unchecked

TomAFrench:
# Description

## Problem\*

Resolves #9047 
## Summary\*

Shows a fix for #9047. I think that we probably want to go a bit further and push the overflow checks for the signed operations inside of the actual binary instruction. This can then have an SSA pass which expands these into the current format but we should avoid the jankiness which currently exists around signed types.

## Documentation\*

Check one:
- [ ] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [ ] I have tested the changes locally.
- [ ] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

Issue #8088, opened 2025-04-15
Multiplication without `range_check` causes difference between ACIR and Brillig


jewelofchaos9:
### Aim

.

### Expected Behavior

Both failed or both succeeded

### Bug

The following SSA fails in ACIR and succeeds in Brillig
```
acir(inline) predicate_pure fn main f0 {
  b0(v0: Field):
    v1 = truncate v0 to 16 bits, max_bit_size: 254
    v2 = cast v1 as i16
    v3 = mul v2, v2
    v4 = div v3, v2
    return v4
}
```
```
brillig(inline) predicate_pure fn main f0 {
  b0(v0: Field):
    v1 = truncate v0 to 16 bits, max_bit_size: 254
    v2 = cast v1 as i16
    v3 = mul v2, v2
    v4 = div v3, v2
    return v4
}
```

With `v0 = 47811`

Brillig returned 0
ACIR: Execution failed: Failed to solve program: 'Cannot satisfy constraint'

### To Reproduce

in `tooling/ssa_fuzzer/lib.rs`
```rs
mod tests {
    .........
    use noirc_evaluator::ssa::ir::types::NumericType;
    use noirc_evaluator::ssa::ir::instruction::BinaryOp;
   
    #[test]
    fn regression_multiplication_without_range_check() {
        let mut acir_builder = FuzzerBuilder::new_acir();
        let mut brillig_builder = FuzzerBuilder::new_brillig();

        let field_acir_var = acir_builder.insert_variable(ValueType::Field.to_ssa_type()).value_id;
        let field_brillig_var =
            brillig_builder.insert_variable(ValueType::Field.to_ssa_type()).value_id;

        let truncated_acir = acir_builder.builder.insert_truncate(field_acir_var, 16, 254);
        let truncated_brillig = brillig_builder.builder.insert_truncate(field_brillig_var, 16, 254);

        let field_casted_i16_acir =
            acir_builder.builder.insert_cast(truncated_acir, NumericType::Signed { bit_size: 16 });
        let field_casted_i16_brillig = brillig_builder
            .builder
            .insert_cast(truncated_brillig, NumericType::Signed { bit_size: 16 });

        let casted_pow_2_acir = acir_builder.builder.insert_binary(
            field_casted_i16_acir,
            BinaryOp::Mul { unchecked: false },
            field_casted_i16_acir,
        );
        let casted_pow_2_brillig = brillig_builder.builder.insert_binary(
            field_casted_i16_brillig,
[...]

PR #9074, opened 2025-07-01
feat: push overflow checks inside of signed binary ops

TomAFrench:
# Description

## Problem\*

Supercedes https://github.com/noir-lang/noir/pull/9059
Resolves https://github.com/noir-lang/noir/issues/9047

## Summary\*

This PR aims to put signed integer instruction in SSA on a similar footing to unsigned integer operations, i.e. unless explicitly marked as unchecked then it internally contains an appropriate overflow check in isolation.

I've basically just moved the relevant code from the ssa_codegen step and moved it into the new `expand_signed_checks` ssa pass. There's now a good deal of code duplication between this pass and `remove_bit_shifts` which implies we should push some of these methods into the `simple_optimization` module to be shared.

I've removed the SSA validation around signed overflow checks (cc @vezenovm) as this is no longer necessary as we can rely on the testing of this new SSA pass to provide the same guarantees.

## Additional Context

I've looked into the changes in `higher_order_functions` a little as I initially expected that this would not affect any of the build artifacts. It seems like the cause is that because the signed operation is marked as unchecked now, the LICM pass picks it up now which unlocks some extra optimization.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Code is relatively messy atm, I'll be doing a little bit of cleaning before marking as ready for review.

compiler/noirc_evaluator/src/ssa.rs
```diff
@@ -164,6 +164,7 @@ pub struct ArtifactsAndWarnings(pub Artifacts, pub Vec<SsaReport>);
 /// something we take can advantage of in the [secondary_passes].
 pub fn primary_passes(options: &SsaEvaluatorOptions) -> Vec<SsaPass> {
     vec![
+        SsaPass::new(Ssa::expand_signed_checks, "expand signed checks"),
```
TomAFrench: This name is temporary but not sure on what would be best. The only thing that comes to mind is `expand_signed_overflow_checks`, but I'm open to suggestions.

TomAFrench: Having this at the very beginning of the SSA pipeline is intentional as I want to ensure that we replicate the current behaviour. In a followup PR we can push this further down so that we can ideally mark more signed operations as unchecked.

compiler/noirc_evaluator/src/ssa/opt/expand_signed_checks.rs
```diff
@@ -0,0 +1,561 @@
...
+
+    /// Insert a constrain instruction at the end of the current block.
+    fn insert_constrain(
+        &mut self,
+        lhs: ValueId,
+        rhs: ValueId,
+        assert_message: Option<ConstrainError>,
+    ) {
+        self.context.insert_instruction(Instruction::Constrain(lhs, rhs, assert_message), None);
+    }
```
TomAFrench, line 440: These should perhaps be pushed onto the inner context as `remove_bit_shifts` has a lot of the same methods. Something for a followup however.

compiler/noirc_evaluator/src/ssa/interpreter/errors.rs
```diff
@@ -14,15 +14,15 @@ pub enum InterpreterError {
     /// These errors are all the result from malformed input SSA
     #[error("{0}")]
     Internal(InternalError),
-    #[error("constrain {lhs_id} == {rhs_id}{message} failed:\n    {lhs} != {rhs}", message = constraint_message(.msg))]
+    #[error("constrain {lhs_id} == {rhs_id}, {message} failed:\n    {lhs} != {rhs}", message = constraint_message(.msg))]
```
aakoshh: `constraint_message` [takes care of printing the comma](https://github.com/noir-lang/noir/blob/5344ab759ae01821f5fb799d4441c1164271caa0/compiler/noirc_evaluator/src/ssa/interpreter/errors.rs#L195), and only do so if the message is present, to match what `constrain` would look like in the SSA.

TomAFrench:
```suggestion
    #[error("constrain {lhs_id} == {rhs_id}{message} failed:\n    {lhs} != {rhs}", message = constraint_message(.msg))]
```

compiler/noirc_evaluator/src/ssa/interpreter/errors.rs
```diff
@@ -55,7 +55,7 @@ pub enum InterpreterError {
     )]
     IncRcRevive { value_id: ValueId, value: String },
     #[error("An overflow occurred while evaluating {instruction}")]
-    Overflow { instruction: String },
+    Overflow { operator: BinaryOp, instruction: String },
```
asterite, line 58: Was this meant to be used in the error message?

asterite:
> I've removed the SSA validation around signed overflow checks (cc @vezenovm) as this is no longer necessary as we can rely on the testing of this new SSA pass to provide the same guarantees.

There was [this issue](https://github.com/noir-lang/noir/issues/8088) that got closed when validations were added. I guess the code in that issue would start to fail again after this PR. I'm not sure how this can be solved... maybe the overflow checks should be added by ACIR/brillig so one can't craft a faulty SSA?

vezenovm:
> There was [this issue](https://github.com/noir-lang/noir/issues/8088) that got closed when validations were added. I guess the code in that issue would start to fail again after this PR. I'm not sure how this can be solved... maybe the overflow checks should be added by ACIR/brillig so one can't craft a faulty SSA?

Oh hmm when I saw the validation checks were removed I forgot about the regression tests added in https://github.com/noir-lang/noir/pull/8901. I see that those tests were removed entirely in this PR. Can we instead just update them to make sure we do not have an ACIR vs. Brillig difference as specified in https://github.com/noir-lang/noir/issues/8088?

compiler/noirc_evaluator/src/ssa/interpreter/tests/instructions.rs
```diff
@@ -76,7 +76,7 @@ fn add_overflow_signed() {
         }
     ",
     );
-    assert_eq!(value, Value::Numeric(NumericValue::I8(-127)));
+    assert!(matches!(error, InterpreterError::Overflow { .. }));
```
aakoshh, line 79: Initially I thought the PR was just moving SSA code generation into a pass, so I was a bit surprised to see these tests change in their expected outcome. 

Where are the overflow rules written down? IIUC previously overflows on signed integers wrapped around, while unsigned ones gave an error, but now they both give an error. Why was it okay to differ earlier, what changed? 

Or are we _not_ generating some follow-up checks now, because they become inherent in the operation?

TomAFrench: > Or are we not generating some follow-up checks now, because they become inherent in the operation?

Yes, this is correct.

> IIUC previously overflows on signed integers wrapped around, while unsigned ones gave an error, but now they both give an error. Why was it okay to differ earlier, what changed?

Previously both checked and unchecked signed integer operations were both really unchecked. A "checked signed integer operation" would be this big stack of operations which would be created during SSA codegen.

```
v2 = add v0, v1  <- Note that this doesn't generate an overflow check.
v3 = truncate v2 to 32 bits, max_bit_size: 33
v4 = cast v3 as u32
v5 = cast v0 as u32
v6 = cast v1 as u32
v8 = lt v5, u32 2147483648
v9 = lt v6, u32 2147483648
v10 = eq v8, v9
v11 = lt v4, u32 2147483648
v12 = eq v11, v8
v13 = unchecked_mul v12, v10
constrain v13 == v10, "attempt to add with overflow" <- the real overflow check.
v14 = cast v3 as i32
```

This PR changes it so that `v2 = add v0, v1` implies this big stack of instructions so that it's a proper checked addition. The SSA pass then unzips this into the current behaviour but with an explicitly unchecked binary operator.

---

wrt the linked tests, it's now correct to have an overflow error as I'm changing the behaviour of this instruction from being identical to the unchecked instruction to including an overflow check.

compiler/noirc_evaluator/src/ssa/interpreter/errors.rs
TomAFrench: No, we use this to link interpreter errors between an overflowing checked instruction and the range check/ constraint created when we expand this instruction into multiple instructions in the `expand_signed_checks` pass.

It would be good to replace the `String` with more structured data at some point (e.g. `lhs`, `rhs`) like the other error variants but that can happen in a future PR.

compiler/noirc_evaluator/src/ssa/interpreter/tests/instructions.rs
aakoshh: Thanks for the explanation. My only question about this is: what would happen if we didn't expand an `add` into the above series of stuff, plus an `unchecked_add`? I see from this test that the interpreter would be allowed to fail, but what would ACIR and Brillig do? Why couldn't we just leave it like `add`, now that it's expected to fail?

TomAFrench: > Why couldn't we just leave it like add, now that it's expected to fail?

I haven't changed anything in acirgen or brillig gen so the compiler would treat it exactly the same as it would before, except we wouldn't have the instructions which come after which perform an explicit overflow check. We'd then have a bug where overflow checks weren't being applied.

TomAFrench: Now we _could_ move all of this overflow logic down into acirgen and brilliggen and I'm open to considering this. I chose to go this route as it starts enforcing overflow checks on the SSA level while having the minimum impact on the build output (so easier to show correctness)

TomAFrench:
> Can we instead just update them to make sure we do not have an ACIR vs. Brillig difference as specified in https://github.com/noir-lang/noir/issues/8088?

:cry: all these tests assume `u64`s

TomAFrench:
Yeah, I can (and probably should) add these but it's gonna basically require a complete revamp of these tests.

compiler/noirc_evaluator/src/ssa/interpreter/tests/instructions.rs
aakoshh: Right, that makes sense that currently the ACIR and Brillig codegen are not updated to handle it. 

Just wanted to understand from a "specification" perspective, that the SSA interpreter, which I thought could be our source of truth about what we expect from an `add` for example, does not, in fact, match the others. 

Is it correct to assume then that `add` should never currently reach the final codegen, always to be replaced by `unchecked_add`, with or without the optional follow up checks? If so, should we panic if `add` gets through, if we don't do it already?

compiler/noirc_evaluator/src/ssa/interpreter/errors.rs
```diff
@@ -22,7 +22,7 @@ pub enum InterpreterError {
         rhs_id: ValueId,
         msg: Option<String>,
     },
-    #[error("constrain {lhs_id} != {rhs_id}{message} failed:\n    {lhs} == {rhs}", message = constraint_message(.msg))]
+    #[error("constrain {lhs_id} != {rhs_id}, {message} failed:\n    {lhs} == {rhs}", message = constraint_message(.msg))]
```
aakoshh:
```suggestion
    #[error("constrain {lhs_id} != {rhs_id}{message} failed:\n    {lhs} == {rhs}", message = constraint_message(.msg))]
```
I thought I sent a comment on this already 🤔 
`constraint_message` inserts he commands if there is a non-empty message.

compiler/noirc_evaluator/src/ssa/ir/types.rs
```diff
@@ -155,6 +155,11 @@ impl Type {
         matches!(self, Type::Numeric(NumericType::Unsigned { .. }))
     }
 
+    /// Returns whether the `Type` represents an signed numeric type.
+    pub fn is_signed(&self) -> bool {
+        matches!(self, Type::Numeric(NumericType::Signed { .. }))
+    }
```
aakoshh, line 161: Just a thought, but it would be helpful to have `op.is_checked()` defined for `BinaryOp`, because, as Ary has also pointed out in the past, reading `unchecked: false` and translating it into `checked: true` is curiously unintuitive.

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -370,26 +336,9 @@ impl<'a> FunctionContext<'a> {
...
-                    BinaryOpKind::Add | BinaryOpKind::Subtract => {
-                        // Result is computed modulo the bit size
-                        let result = self.builder.insert_truncate(result, bit_size, bit_size + 1);
-                        let result = self.insert_safe_cast(
-                            result,
-                            NumericType::unsigned(bit_size),
-                            location,
-                        );
-
-                        self.check_signed_overflow(result, lhs, rhs, operator, bit_size, location);
```
aakoshh, line 382: The docstring of this method, `check_overflow`, talks extensively about delegating to `check_signed_overflow`, which is now removed below.

compiler/noirc_evaluator/src/ssa/opt/expand_signed_checks.rs
```diff
@@ -0,0 +1,606 @@
...
+            // We remove the current instruction, as we will need to replace it with multiple new instructions.
+            context.remove_current_instruction();
+
+            let old_result = *context.dfg.instruction_results(instruction_id).first().unwrap();
+
+            let mut simple_context = Context { context };
+            let new_result = match operator {
+                BinaryOp::Add { .. } => simple_context.insert_add(lhs, rhs),
+                BinaryOp::Sub { .. } => simple_context.insert_sub(lhs, rhs),
+                BinaryOp::Mul { .. } => simple_context.insert_mul(lhs, rhs),
```
aakoshh:
```suggestion
            let mut expansion_context = Context { context };
            let new_result = match operator {
                BinaryOp::Add { .. } => expansion_context.insert_add(lhs, rhs),
                BinaryOp::Sub { .. } => expansion_context.insert_sub(lhs, rhs),
                BinaryOp::Mul { .. } => expansion_context.insert_mul(lhs, rhs),
```
I see `context` is a `SimpleOptimizationContext` type, which is where I think the name `simple_context` comes from the wrapper, but I'm not sure why this `Context` type is considered _simple_.

compiler/noirc_evaluator/src/ssa/opt/expand_signed_checks.rs
```diff
@@ -0,0 +1,606 @@
...
+            positive_predicate,
+            BinaryOp::Add { unchecked: true },
+            negative_predicate,
+        )
+    }
+
+    /// Inserts a cast instruction at the end of the current block and returns the results
+    /// of the cast.
+    ///
+    /// Compared to `self.builder.insert_cast`, this version will automatically truncate `value` to be a valid `typ`.
```
aakoshh:
```suggestion
    /// Compared to `self.insert_cast`, this version will automatically truncate `value` to be a valid `typ`.
```

compiler/noirc_evaluator/src/ssa/validation/mod.rs
```diff
@@ -41,110 +38,9 @@ struct Validator<'f> {
...
-        Self { function, signed_binary_op: None, range_checks: HashMap::default() }
-    }
-
-    /// Validates that any checked signed add/sub/mul are followed by the appropriate instructions.
-    /// Signed overflow is many instructions but we validate up to the initial truncate.
-    ///
-    /// Expects the following SSA form for signed checked operations:
-    /// Add/Sub -> Truncate
-    /// Mul -> Cast -> Truncate
-    fn validate_signed_op_overflow_pattern(&mut self, instruction: InstructionId) {
```
aakoshh, line 62: Could we salvage this validation method as a debug-only post-check for the expansion pass?

TomAFrench: Yeah, I think this is sensible.

compiler/noirc_evaluator/src/ssa/ir/types.rs
TomAFrench: I thought about this but I wasn't sure on how to handle the operations outside of add/sub/mul. I think we can handle this in a follow up however.

aakoshh: Sure. I left the comment here, because of how similar this method is to that situation: `is_signed` is defined on all `Type`, even though it only returns `true` for numeric signed types, and `false` for everything else, including those for those for which it doesn't make any sense. I suppose it would look something like `matches!(op, Add | Sub | Div) && op.is_checked()`.

TomAFrench:
> Yeah, I can (and probably should) add these but it's gonna basically require a complete revamp of these tests.

I'm going to punt this as it's forcing basically a complete rewrite of this test suite.

compiler/noirc_evaluator/src/ssa/validation/mod.rs
TomAFrench: added in [aad9b03](https://github.com/noir-lang/noir/pull/9074/commits/aad9b036ffaaa806fe1d8d646f9541f82ffcef31)

TomAFrench: I've removed this again as it was failing on preexsiting unchecked ops (which wouldn't have an overflow check added)

PR #9075, opened 2025-07-01
chore: graceful handling of SIGPIPE

### LINKED ISSUE #8534: `nargo` panics when pipe is broken
### Aim

Attempted to pipe `nargo`'s output to a failing command

### Expected Behavior

Expected `nargo` to exit gracefully when the pipe is broken

### Bug

```bash
❯ nargo execute | echo "${missing_cmd}"

Message:  failed printing to stdout: Broken pipe (os error 32)
Location: library/std/src/io/stdio.rs:1123

```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.6 noirc version = 1.0.0-beta.6+e796dfd67726cbc28eb9991782533b211025928d (git version hash: e796dfd67726cbc28eb9991782533b211025928d, is dirty: false)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #8534

## Summary\*
handle SIGPIPE when printing to stdout

## Additional Context
I tried 'execute', 'compile' and 'info', but some other println may need to be replaced.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Wouldn't this require us to wrap every instance of `println`?

guipublic:
> Wouldn't this require us to wrap every instance of `println`?

Probably, compile with output is failing, I will try to fix it and see if is a lot of refactoring of not.

TomAFrench:
If this is something which we're need to wrap all instances of `println` then I'd say it's probably not worth going forwards with this. It's going to be annoying to guarantee that we don't have any unwrapped `println`s and we don't really unlock much with this anyway so it's probably not worth the maintenance burden.

guipublic:
For 'nargo compile', it was just 3 `println!` to replace.
So far the impact is small and it's not critical if not exhaustif I believe.

jfecher:
I wonder what `cargo` does internally for this. It doesn't seem to panic in the same way when I run e.g. `cargo run | echo "${missing_cmd}"`

compiler/noirc_driver/src/lib.rs
```diff
@@ -445,8 +445,8 @@ pub fn compile_main(
     }
 
     if options.print_acir {
-        println!("Compiled ACIR for main (unoptimized):");
-        println!("{}", compiled_program.program);
+        noirc_errors::print_to_stdout(format_args!("Compiled ACIR for main (unoptimized):\n"));
+        noirc_errors::print_to_stdout(format_args!("{}\n", compiled_program.program));
```
aakoshh: It would be nice if this was a bit less verbose, and that we wouldn't have to remember to put a `\n` in it.

guipublic: Done

PR #9079, opened 2025-07-01
fix(codegen): look at struct args (generics) when generating structs

asterite:
# Description

## Problem

Resolves #9044

## Summary

The fix in the previous PR was wrong. Nested structs were generated. It's just that at one point generic struct field types get replaced with bindings, and the bindings (possibly structs) end up in an `args` field. So what was missing is taking a look at `args` to find more structs to generate.

## Additional Context

I added `rm -rf ./test/test_lib/export` to the test script because otherwise when you made changed to the Noir code some old code might remain there and error, and it can be a bit misleading.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/noir_codegen/package.json
+++ b/tooling/noir_codegen/package.json
@@ -35,7 +35,7 @@
     "dev": "tsc-multi --watch",
     "build": "tsc",
     "test": "yarn test:codegen && yarn test:node && yarn test:clean",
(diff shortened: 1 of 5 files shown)


asterite:
Even though this seems correct it ends up generating duplicated structs in some cases. I'm not going to work on this for now so I'll close this.

PR #9083, opened 2025-07-02
chore(fuzz): Test AST print and parse roundtrip

### LINKED ISSUE #9058: Fuzz test the AST generation and parse roundtrip
Followup for https://github.com/noir-lang/noir/pull/9056

The AST fuzzer generates monomorphized AST, bypassing the frontend, which has some undocumented assumptions. Besides #9057 and test failures, we could have a fuzz test where an AST is generated, printed, parsed back into an AST, printed again, and compared to the original print: if it looks different, then we have some discrepancy that could lead to potential differences in the effects.

```
let program1 = arb_program();
let src1 = format!("{}", DisplayAstAsNoir(&program1));
let program2 = parse(&src1);
let src2 = format!("{}", DisplayAstAsNoir(&program2));
similar_equals::assert_equal(src1, src2);
```

aakoshh:
# Description

## Problem\*

Resolves #9058 

## Summary\*

Adds a new prop test to the AST fuzzer to ensure that the monomorphized AST it generates matches the one the compiler would create from same code. 

```shell
cargo test -p noir_ast_fuzzer --test mono
```

## Additional Context

The can be subtle differences, for example rendering `Unary { op: Minus, rhs: Literal(1) }` could be parsed back as `Literal(-1)`. These don't cause a test failure, because we only compare the printed Noir source, and I made some changes to the printer to ensure they appear identical. Other instances that cause no harm such as `--1` being parsed as `1` are handled in the test itself. 

The test did highlight some actual semantic differences, such as a misunderstanding in the `return_type` of a unary reference, which resulted in an extra `&mut` in the printed output. Similarly it uncovered that the separation of array indexes into a temporary variable had a different definition of "side effect" than the fuzzer. 

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
Looks great!

PR #4675, opened 2024-03-28
chore!: Add `as_array` and remove `_slice` variants of hash functions

jfecher:
# Description

## Problem\*

Resolves an issue raised in a slack thread:

```rs
let slice1 = &[1, 2, 3];
let slice2 = slice1.pop_back().0.push_back(3);

assert(slice1 == slice2); // fails
```

## Summary\*

Removes `_slice` functions from the stdlib and adds `as_array` method to convert a slice to an array.

## Additional Context

Currently failing because the pedersen hasher uses the pedersen_slice variant.

## Documentation\*

Check one:
- [ ] No documentation needed.
- [x] Documentation included in this PR.
  - Documentation is removed currently and I'll need to add separate docs for `as_array`
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [ ] I have tested the changes locally.
- [ ] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #9098, opened 2025-07-02
fix!: remove `hash_to_field` from stdlib

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR removes the insecure function `hash_to_field` from the stdlib. It looks like this came while trying to avoid a breaking change in https://github.com/noir-lang/noir/pull/4675 as it's no longer possible to pass a slice to `blake2s`.

Any fix for this is going to be a breaking to the point of being a completely different function (at the minimum it can't take a slice as input, it also doesn't really make sense for it to accept `Field`s as I would just use a poseidon hash in that case). I think it's then worth just removing the function entirely and we can re-add a fixed implementation in future if necessary

## Documentation\*

Check one:

# PR Checklist\*


jfecher (approved):
Need to update all .snap files now that the stdlib changed

TomAFrench:
:melting_face:

jfecher:
I'm running the tests locally now, I'll push a commit when they finish

Savio-Sou:
Projects hosted on GitHub that relies on the library: https://github.com/search?q=path%3A*.nr+AND+%22hash_to_field%22&type=code

For similar removal of features in the future, we should at the minimum add an error that i) recommends an alternative (e.g. alternative hashing scheme in this case) for our users, and ii) prompts them to create an Issue if the alternative is not helpful.

PR #9103, opened 2025-07-03
fix: forbid self-referencing type aliases

### LINKED ISSUE #8530: Type alias referring to itself crashes the compiler
### Aim

Given this code:

```noir
pub type X = X;

fn main() {
    let x: X = 1;
}
```

### Expected Behavior

It should not compile, saying that a cycle was detected while solving `X`.

### Bug

The compiler gives a stack overflow.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #8530

## Summary\*
Check for cyclic type aliases when checking the type kind.

## Documentation\*

Check one:

# PR Checklist\*


guipublic:
This PR is a draft because it does not use the dependency graph as suggested in the issue. Therefore I'd like to validate this approach.
When investigating the issue, I found out that it happened when retrieving the type kind for a self-referencing alias. That's why I added the check here.

compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -1501,6 +1502,36 @@ impl Type {
...
+            }
+            Type::Alias(def, generics) => {
+                let alias_id = def.borrow().id;
+                if aliases.contains(&alias_id) {
+                    true
+                } else {
+                    aliases.insert(alias_id);
+                    def.borrow().get_type(generics).has_cyclic_alias(aliases)
+                }
+            }
```
jfecher, line 1622: I don't think this is quite right since we can still have the same alias id nested within the same alias with different generics. E.g. consider an alias that is just a wrapper over some type:
```noir
type Id<T> = T;
```
Now if we had a `Id<Id<Field>>` it would register as a cyclic type due to the `Id<Field>` inside the first `Id`.

jfecher (changes requested):
Could use some more robust tests to catch more cases

compiler/noirc_frontend/src/hir_def/types.rs
guipublic: I tried and this gives an error: `Could not determine the type of the generic argument `T``
I added a unit test for this case.

jfecher:
When I run the program:
```noir
type A<T> = A<T>;

fn main() {
    let x: A<A<Field>> = 0;
}
```
I get no error with this PR despite `A` referring to itself forever

guipublic:
> I get no error with this PR despite `A` referring to itself forever

This is because you are not using x so it gets simplified. Without this PR, it gives a stack overflow.
If you want to have an error, you need to use x. E.g:
```
type A<T> = A<T>;

fn main() -> pub A<A<Field>> {
    let x: A<A<Field>> = 0;
    x
}
```

compiler/noirc_frontend/src/elaborator/types.rs
```diff
@@ -777,6 +777,9 @@ impl Elaborator<'_> {
         expected_kind: &Kind,
         location: Location,
     ) -> Type {
+        if typ.has_cyclic_alias(&mut HashSet::default()) {
+            return Type::Error;
```
jfecher, line 782: Can we issue an error here in addition to returning an error type?
As-is if we do code like:
```noir
type Rec = Rec;

fn main() {
    let z: Rec = 2;
    println(z);
}
```
We get:
```
error: Type annotation needed
   ┌─ src/main.nr:5:5
   │
11 │     println(z);
   │     ------- Could not determine the type of the generic argument `T` declared on the function `println`
   │
```
which may confuse users since `z` already has a type

guipublic: I added a cyclic type error, the result is a bit strange but at least the cyclic issue is obvious now.
```
error: Binding `A<Field>` here to the `_` inside would create a cyclic type
  ┌─ src/main.nr:3:20
  │
3 │ fn main() -> pub A<A<Field>> {
  │                    - Cyclic types have unlimited size and are prohibited in Noir
  │

error: Binding `A<error>` here to the `_` inside would create a cyclic type
  ┌─ src/main.nr:3:18
  │
3 │ fn main() -> pub A<A<Field>> {
  │                  - Cyclic types have unlimited size and are prohibited in Noir
  │

error: Binding `A<Field>` here to the `_` inside would create a cyclic type
  ┌─ src/main.nr:4:14
  │
4 │     let x: A<A<Field>> = 0;
  │              - Cyclic types have unlimited size and are prohibited in Noir
  │

error: Binding `A<error>` here to the `_` inside would create a cyclic type
  ┌─ src/main.nr:4:12
  │
4 │     let x: A<A<Field>> = 0;
  │            - Cyclic types have unlimited size and are prohibited in Noir
  │

Aborting due to 4 previous errors
```

jfecher (approved):
Thanks!

PR #9108, opened 2025-07-03
feat(fuzz): Generate Match expressions and statements

### LINKED ISSUE #7926: Generate arbitrary Match expressions
Follow up https://github.com/noir-lang/noir/pull/7890 with arbitrary `Match` expressions

aakoshh:
# Description

## Problem\*

Resolves #7926 

## Summary\*

Implement `match` statements in the AST generator.

## Additional Context

We are generating the what a match statement looks like in the monomorphized AST, but doing so without having in mind a full match statement in Noir, rather, just doing a single layer. 

For example take this program:
```rust
global G: (u32, u32) = (2, 3);
fn main() -> pub u8 {
  match G {
    (2, _) => 1, 
    (_, 3) => 2, 
    (4, _) => 3, 
    _ => 4
  }
}
```
It turns into the following AST:
```
global G$g0: (u32, u32) = (2, 3);
fn main$f0() -> pub u8 {
    {
        let internal variable$l0 = G$g0;
        match $0 {
            ($1, $2) => match $1 {
                2 => {
                    let _$l3 = internal_match_variable_1$l2;
                    1
                },
                4 => match $2 {
                    3 => {
                        let _$l4 = internal_match_variable_0$l1;
                        2
                    },
                    _ => {
                        let _$l5 = internal_match_variable_1$l2;
                        3
                    },
                },
                _ => match $2 {
                    3 => {
                        let _$l6 = internal_match_variable_0$l1;
                        2
                    },
                    _ => {
                        let _$l7 = internal variable$l0;
                        4
                    },
                },
            },
        }
    }
}
```

So, it's a series of simpler match statements, on one variable at a time. The AST generator generates only one random match, which can be:
* a bool variable against `true`, `false`, `_`
* a numeric variable against some random values and `_` (wanted to do ranges but the frontend doesn't parse them)
* a tuple gets matched and individual fields introduced as local variables

I added support for variable names in the `Match` so that I can print a parseable AST.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs
@@ -774,7 +774,7 @@ impl FunctionContext<'_> {
     }
 
     fn codegen_match(&mut self, match_expr: &ast::Match) -> Result<Values, RuntimeError> {
-        let variable = self.lookup(match_expr.variable_to_match);
+        let variable = self.lookup(match_expr.variable_to_match.0);
 
         // Any matches with only a single case we don't need to check the tag at all.
         // Note that this includes all matches on struct / tuple values.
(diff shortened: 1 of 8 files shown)


aakoshh:
Latest test failures need https://github.com/noir-lang/noir/pull/9195

PR #9112, opened 2025-07-03
chore: test that associated function and constant with the same name collide

### LINKED ISSUE #9111: Test that trait associated constants and functions collide by name
Add a test that this code errors:

```noir
pub trait Foo {
    let N: u32;

    fn N() {}
}

fn main() {}
```

Right now it errors but the message could be improved:

```
error: Duplicate definitions of trait associated function with name N found
  ┌─ src/main.nr:2:9
  │
2 │     let N: u32;
  │         - First trait associated function found here
3 │
4 │     fn N() {}
  │        - Second trait associated function found here
```

asterite:
# Description

## Problem

Resolves #9111

## Summary

Also always uses "trait associated item" in error messages. 

I checked Rust and their error message is:

```
error[E0428]: the name `N` is defined multiple times
 --> src/main.rs:3:5
  |
2 |     fn N() {}
  |     ------ previous definition of the value `N` here
3 |     fn N() {}
  |     ^^^^^^ `N` redefined here
  |
  = note: `N` must be defined only once in the value namespace of this trait
```

so they also try to avoid saying "function" or "constant".

## Documentation

Check one:

# PR Checklist


TomAFrench (approved):
LGTM

PR #9113, opened 2025-07-03
chore: refactor ssa-fuzzer tests

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR does a bunch of the refactoring to make adding fuzz tests in #9074 easier.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/ssa_fuzzer/src/lib.rs
+++ b/tooling/ssa_fuzzer/src/lib.rs
@@ -14,12 +14,10 @@ mod tests {
     use acvm::FieldElement;
     use acvm::acir::native_types::{Witness, WitnessMap};
     use noirc_driver::{CompileOptions, CompiledProgram};
-    use rand::RngCore;
+    use rand::Rng;
 
     use noirc_evaluator::ssa::ir::instruction::BinaryOp;
     use noirc_evaluator::ssa::ir::types::NumericType;
-
-    const NUMBER_OF_VARIABLES_INITIAL: u32 = 7;
 
     struct TestHelper {
         acir_builder: FuzzerBuilder,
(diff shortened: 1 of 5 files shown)


TomAFrench:
Also it seems like the signed int tests in that module fail nondeterministically?

tooling/ssa_fuzzer/src/lib.rs
```diff
@@ -55,21 +57,19 @@ mod tests {
         }
     }
 
-    /// Generates a random array with config::NUMBER_OF_VARIABLES_INITIAL elements
-    fn generate_values() -> Vec<u64> {
+    fn generate_values<T>() -> (T, T)
```
aakoshh: I never knew why the number of constant `NUMBER_OF_VARIABLES_INITIAL` was used in so many places, but perhaps it would be good to include a docstring about why a generating random _pair_ of values has a dedicated method. 

I'd consider renaming to `generate_pair`, and potentially having another `generate_values<T, const N: usize>() -> [T; N]`.

tooling/ssa_fuzzer/src/lib.rs
```diff
@@ -81,149 +81,169 @@ mod tests {
...
-        values[1] %= 64;
-        let noir_res =
-            run_instruction_double_arg(FuzzerBuilder::insert_shr_instruction, values.clone());
-        compare_results(values[0] >> values[1], noir_res);
+        let (lhs, mut rhs): (u64, u64) = generate_values();
+        rhs %= 64;
+        let noir_res = run_instruction_double_arg(
+            FuzzerBuilder::insert_shr_instruction,
+            (lhs.into(), ValueType::U64),
+            (rhs.into(), ValueType::U64),
```
aakoshh, line 263: I thought shifting happens by `u8` 🤔 
I was wondering if we have a case for the two arguments to have different types, because the test seems to have always assumed homogenous values so far, and I thought this was it, but apparently not? 

I know it's just a refactoring of exactly how it worked, so this is just a comment, not something to be fixed here and now.

aakoshh (approved):
LGTM 👍

tooling/ssa_fuzzer/src/lib.rs
TomAFrench: Yeah, this is a prexisting issue in the ssa fuzzer. I'm going to leave this as out of scope for this PR however.

PR #9115, opened 2025-07-04
fix: use enable_side_effects for u128 multiplication overflow checks

### LINKED ISSUE #9099: ACIR vs Brillig: failed assertion on unreached branch; constant u128 multiplication overflow
### Aim

The following Noir program failed in ACIR and solved in Brillig
```noir
fn main(condition: bool) -> pub bool {
  if condition {
    let a: u128 = 0;
    let b = !a;
    let _ = b * b;
  }
  condition
}
```
with Prover.toml
```toml
"condition" = 0
```

### Expected Behavior

ACIR solved

### `nargo execute`
```
nargo execute
  ┌─ /home/defkit/work/hui/src/main.nr:5:13
  │
5 │     let _ = b * b;
  │             -----
  │
  = Call stack:
    1. /home/defkit/work/hui/src/main.nr:5:13

Failed assertion
```

### `nargo execute --force-brillig`
```
nargo execute --force-brillig
[hui] Circuit witness successfully solved
[hui] Circuit output: Field(0)
[hui] Witness saved to /home/defkit/work/hui/target/hui.gz
```

By the way, when I use `--show-ssa` I also receive assertion is always false for ACIR
```
  b0(v0: u1):
    enable_side_effects v0
    constrain u128 1 == u128 0, "attempt to multiply with overflow"
    enable_side_effects u1 1
    return v0
}

  ┌─ src/main.nr:5:13
  │
5 │     let _ = b * b;
  │             ----- As a result, the compiled circuit is ensured to fail. Other assertions may also fail during execution
  │
  = Call stack:
    1. src/main.nr:5:13

  ┌─ /home/defkit/work/hui/src/main.nr:5:13
  │
5 │     let _ = b * b;
  │             -----
  │
  = Call stack:
    1. /home/defkit/work/hui/src/main.nr:5:13

Failed assertion
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #9099 

## Summary\*
u128 overflow checks take the current 'enable_side_effects' into account so that the check is disabled if there if side effects are disabled as well.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/simple_optimization.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/simple_optimization.rs
@@ -1,3 +1,5 @@
+use acvm::FieldElement;
+
 use crate::{
     errors::RtResult,
     ssa::ir::{
@@ -5,6 +7,7 @@ use crate::{
         dfg::DataFlowGraph,
         function::Function,
         instruction::{Instruction, InstructionId},
+        types::NumericType,
         value::{ValueId, ValueMapping},
     },
 };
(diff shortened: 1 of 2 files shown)


compiler/noirc_evaluator/src/ssa/opt/simple_optimization.rs
```diff
@@ -95,6 +101,7 @@ pub(crate) struct SimpleOptimizationContext<'dfg, 'mapping> {
     pub(crate) block_id: BasicBlockId,
     pub(crate) instruction_id: InstructionId,
     pub(crate) dfg: &'dfg mut DataFlowGraph,
+    pub(crate) enable_side_effects: ValueId,
```
asterite, line 104: Nice! We could use this in `remove_unreachable_instructions` as well. No need to do it in this PR, I can do it in a follow-up PR.

guipublic:
> Looks great! Could you add a regression test?

I already added a unit test, do you think it is better to have an additional test in test_programs?

asterite:
Ooooh... I missed the unit test. I saw the issue had Noir code and couldn't find it in this PR, but like this it's fine, sorry!

asterite:
But, hmm... it might be good to have the failing file as a regression test as it's a more complete scenario. Up to you.

asterite (approved):

PR #9123, opened 2025-07-04
chore: nicer way to show nargo interpret results

asterite:
# Description

## Problem

## Summary

For this program:

```noir
fn main() -> pub [(Field, Field); 2] {
    [(1, 2), (3, 4)]
}
```

`nargo interpret` would show this:

```
--- Interpreter result after Verifying no dynamic array indices to reference value elements (step 39):
Ok([ArrayOrSlice(ArrayValue { elements: Shared(RefCell { value: [Numeric(Field(1)), Numeric(Field(2)), Numeric(Field(3)), Numeric(Field(4))] }), rc: Shared(RefCell { value: 1 }), element_types: [Numeric(NativeField), Numeric(NativeField)], is_slice: false })])
```

which is... fine! I was about to change it by implementing a nicer `Display` for `Value` but that was already implemented, we just weren't using it. Together with improving how arrays are shown, the new output is:

```
--- Interpreter result after Verifying no dynamic array indices to reference value elements (step 39):
Ok(rc1 [(Field 1, Field 2), (Field 3, Field 4)])
```

In the case of an empty array, it's shown as `[(Field, Field); 0]` so the element types can be seen (when there are values, the values are grouped according to element types, if needed, so the element types can be inferred from those values).

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/interpreter/tests/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/interpreter/tests/mod.rs
@@ -3,6 +3,7 @@
 use std::sync::Arc;
 
 use acvm::{AcirField, FieldElement};
+use insta::assert_snapshot;
 
 use crate::ssa::{
     interpreter::value::NumericValue,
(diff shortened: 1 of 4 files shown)


aakoshh (approved):
Lovely 👏

PR #9128, opened 2025-07-07
fix(ssa): Recursively invalidate cached array instruction in constant folding

### LINKED ISSUE #9102: Corruption of literal defined array in Brillig
### Aim

This program:
```noir
fn main() -> pub [bool; 2] {
    unsafe {
        func_1([[true, true], [true, false]])
    }
}

unconstrained fn func_1(
    mut a: [[bool; 2]; 2],
) -> [bool; 2] {
        let mut i: u32 = 1;
        loop {
            if (i == 1) {
                break
            }
        }
        a[1][1] = true;
        [true, false]
}
```

returns [1, 1].

Running it in comptime correctly produces [1, 0]; removing either the loop or `a` assignment, or making `i` immutable all lead to correct [1, 0]. 

### Expected Behavior

The output should be [1, 0].

### Bug

```
$ nargo execute --force
[test1] Circuit output: Vec([Field(1), Field(1)])
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9102 

## Summary\*

Fixes `remove_possibly_mutated_cached_make_arrays` to handle multi-dimensional arrays.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench (commented):
We apply `inc_rc` instructions on cached`make_array` instructions in `fold_constants_into_instruction`. Looks like we may need to update this to handle nested arrays as well?

aakoshh:
@TomAFrench I'll have a look, but in this particular test, and the bug I fixed last time, there was no `inc_rc` in the SSA that could have informed anything before the reuse:
```
After Initial SSA:
brillig(inline) fn main f0 {
  b0():
    v1 = make_array [u1 1, u1 1] : [u1; 2]
    v3 = make_array [u1 1, u1 0] : [u1; 2]
    v4 = make_array [v1, v3] : [[u1; 2]; 2]
    v6 = call f1(v4) -> [u1; 2]                   	// src/main.nr:4:9
    return v6
}
  b0(v0: [[u1; 2]; 2]):
    v1 = allocate -> &mut [[u1; 2]; 2]            	// src/main.nr:4:9
    store v0 at v1                                	// src/main.nr:4:9
    v2 = allocate -> &mut u32                     	// src/main.nr:9:22
    store u32 1 at v2                             	// src/main.nr:9:22
    jmp b1()
  b1():
    v4 = load v2 -> u32                           	// src/main.nr:9:22
    v5 = eq v4, u32 1                             	// src/main.nr:11:13
    jmpif v5 then: b3, else: b4
  b2():
    v6 = load v1 -> [[u1; 2]; 2]                  	// src/main.nr:11:13
    v7 = array_get v6, index u32 1 -> [u1; 2]     	// src/main.nr:15:5
    v9 = array_set v7, index u32 1, value u1 1    	// src/main.nr:15:5
    v10 = array_set v6, index u32 1, value v9     	// src/main.nr:15:5
    store v10 at v1                               	// src/main.nr:15:5
    v12 = make_array [u1 1, u1 0] : [u1; 2]       	// src/main.nr:15:5
    return v12
  b3():
    jmp b2()
  b4():
    jmp b1()
}

...

After Purity Analysis (2) (step 21):
  b0():
    v1 = make_array [u1 1, u1 1] : [u1; 2]
    v3 = make_array [u1 1, u1 0] : [u1; 2]
    v4 = make_array [v1, v3] : [[u1; 2]; 2]
    v5 = allocate -> &mut [[u1; 2]; 2]            	// src/main.nr:4:9
    store v4 at v5                                	// src/main.nr:4:9
    v6 = allocate -> &mut u32                     	// src/main.nr:9:22
    store u32 1 at v6                             	// src/main.nr:9:22
    jmp b1()
  b1():
    v8 = load v6 -> u32                           	// src/main.nr:9:22
    v9 = eq v8, u32 1                             	// src/main.nr:11:13
    jmpif v9 then: b2, else: b3
  b2():
    v10 = load v5 -> [[u1; 2]; 2]                 	// src/main.nr:11:13
    v11 = array_get v10, index u32 1 -> [u1; 2]   	// src/main.nr:15:5
    v12 = array_set v11, index u32 1, value u1 1  	// src/main.nr:15:5
    v13 = array_set v10, index u32 1, value v12   	// src/main.nr:15:5
    store v13 at v5                               	// src/main.nr:15:5
    v14 = make_array [u1 1, u1 0] : [u1; 2]       	// src/main.nr:15:5
    return v14
  b3():
    jmp b1()
}
```

aakoshh:
@TomAFrench I see [which one you mean now](https://github.com/noir-lang/noir/blob/090d939071b69c288233fa17080e27e2057c85b6/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs#L315). 

You may be right that it needs to insert `inc_rc` for nested arrays, so that when we do something like this:
```
    v1 = make_array [u1 1, u1 1] : [u1; 2]
    v3 = make_array [u1 1, u1 0] : [u1; 2]
    v4 = make_array [v1, v3] : [[u1; 2]; 2]
    ...
    v10 = load v5 -> [[u1; 2]; 2]                 	// src/main.nr:11:13
    v11 = array_get v10, index u32 1 -> [u1; 2]   	// src/main.nr:15:5
    v12 = array_set v11, index u32 1, value u1 1  	// src/main.nr:15:5
    v13 = array_set v10, index u32 1, value v12   	// src/main.nr:15:5
```
then the first `array_set` doesn't modify a sub-array if there is a chance that we have a cache specifically for that. 

OTOH if we have reused `v3` from the cache, it will have its own `inc_rc v3` inserted, which should prevent writing to `v11` from modifying it. And if we reused `v4` from the cache, then we will have inserted `inc_rc v4`, so even if `v11` is written to, as long as we go through `v4`, the write to `v10` should have resulted in a copy, leaving `v4` in tact, right? I'm not sure if that's actually how it works. 

But, the problem with these inserted `inc_rc` instructions is that they don't take into account the modification that happens _before_ we decide to reuse something from the cache. The `inc_rc` would potentially have to be inserted retroactively after the creation:

Say we start with this:
```rust
v1 = make_array [1, 2, 3]
v2 = array_set v1, 0, 4
v3 = make_array [1, 2, 3]
```
disregarding the `remove_potentially_modified_array`, currently we end up with this:
```rust
v1 = make_array [1, 2, 3]
v2 = array_set v1, 0, 4
v3 = v1
inc_rc v1
```
but what would be safe is this:
```rust
v1 = make_array [1, 2, 3]
inc_rc v1
v2 = array_set v1, 0, 4
v3 = v1
```

TomAFrench:
> the problem with these inserted inc_rc instructions is that they don't take into account the modification that happens before we decide to reuse something from the cache.

This is fine as [we remove any instructions from the cache if we see an `array_set` instruction acting on it](https://github.com/noir-lang/noir/blob/090d939071b69c288233fa17080e27e2057c85b6/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs#L750-L752). We could go back and push an `inc_rc` but just aborting is fine for now (and safer).

aakoshh:
That's right. What I did in this PR is to remove from the cache the nested arrays used by the instruction as well, not just the array it created. The argument about inserting the `inc_rc` for those nested array values has the same problem as the result array itself: by the time we decide to reuse them, they have potentially already been modified.

I'll check if the `inc_rc` on the container array would have prevented them from being modified further down the line.

TomAFrench:
Ah, I think I misunderstood you. Yes, we should be injecting `inc_rc` for the subarrays already when we have two `make_array` instructions which will get deduplicated. No change is necessary then.

aakoshh:
To check what would happen, I used a test like this:
```rust
#[test]
fn reuse_array() {
    let values: Vec<Value> = expect_values(
        r#"
  b0():
    v1 = make_array [u1 0, u1 1] : [u1; 2]
    v2 = make_array [v1] : [[u1; 2]; 1]
    inc_rc v2
    v3 = allocate -> &mut [[u1; 2]; 2]
    store v2 at v3
    v6 = load v3 -> [[u1; 2]; 1]
    v7 = array_get v6, index u32 0 -> [u1; 2]
    v8 = array_set v7, index u32 0, value u1 1
    v9 = array_set v6, index u32 0, value v8
    store v9 at v3
    return v1
}
    "#,
    );
    let u1 = NumericType::unsigned(1);
    let zero = || from_constant(0u128.into(), u1.clone());
    let one = || from_constant(1u128.into(), u1.clone());
    let arr0 = || Value::array(vec![zero(), one()], vec![Type::Numeric(u1)]);
    assert_eq!(values[0], arr0());
}

```
It fails: the `inc_rc v2` does not prevent the write to `v1` from happening, so an `inc_rc` on the nested arrays is necessary. 

But I couldn't construct a Noir program where during cache reuse it did not already insert an `inc_rc` for the nested component, so I agree, we would just be duplicating them.

PR #9041, opened 2025-06-27
feat: Allow TraitAsType syntax to refer to associated constants in expressions

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/8797

## Summary

Implements `<Type as Trait>::Constant` in an expression position.

This also makes `T::Constant` work in a context where we have `T: Trait`.

## Additional Context

I wanted to implement this without adding new concepts or things for the compiler to track. I ended up generalizing `TraitMethodId` into a `TraitItemId` by changing the `method_index: usize` into a `definition_id: DefinitionId` field and giving associated constants a definition id. I did end up adding some new things to track these new definition ids though so it isn't perfect.

## Documentation

Check one:
- [x] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #9133, opened 2025-07-07
chore: simplify associated constants lookup

asterite:
# Description

## Problem

## Summary

After #9041 there were a couple of code paths that were no longer needed. That is, #9041 was the correct way to implement associated constants, the other code wasn't a workaround but it didn't account for all cases.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/patterns.rs
+++ b/compiler/noirc_frontend/src/elaborator/patterns.rs
@@ -569,12 +569,6 @@ impl Elaborator<'_> {
 
     pub(super) fn elaborate_variable(&mut self, variable: Path) -> (ExprId, Type) {
         let variable = self.validate_path(variable);
-        if let Some((expr_id, typ)) =
-            self.elaborate_variable_as_self_method_or_associated_constant(&variable)
-        {
-            return (expr_id, typ);
-        }
-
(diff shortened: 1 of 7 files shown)


compiler/noirc_frontend/src/hir_def/traits.rs
```diff
@@ -231,7 +231,7 @@ impl Trait {
         let ordered = vecmap(&self.generics, |generic| generic.clone().as_named_generic());
         let named = vecmap(&self.associated_types, |generic| {
             let name = Ident::new(generic.name.to_string(), location);
-            NamedType { name, typ: generic.clone().as_named_generic() }
+            NamedType { name, typ: Type::TypeVariable(generic.type_var.clone()) }
```
asterite, line 234: @jfecher When I initially started this PR some tests were failing. Comparing with `master` I noticed that when we elaborate `AsTraitPath` the `TraitGenerics` we produce there have type variables for named arguments. Here we were returning `NamedGeneric` and that didn't work (they don't unify the same way as type variables). Given that the `AsTraitPath` was working well, I decided to change this code to also return `TypeVariable`. That fixed the issue but I'm not 100% sure this is fine.

PR #9134, opened 2025-07-07
fix: check for maximum limbs during radix decomposition

### LINKED ISSUE #9073: ACIR vs. Brillig: Byte decomposition for field larger than num bytes
### Aim

The tests below will pass (taken from test added in https://github.com/noir-lang/noir/pull/9068).
```noir
fn constrained_le_decompose(x: Field) -> [u8; 16] {
    let x_bytes: [u8; 16] = x.to_le_bytes();
    println(x);
    println(x_bytes);
    x_bytes
}
unconstrained fn brillig_le_decompose(x: Field) -> [u8; 16] {
    constrained_le_decompose(x)
}
#[test(should_fail_with = "Field failed to decompose into specified 16 limbs")]
fn test_acir_field_too_big() {
    // One greater than `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF`
    let _ = constrained_le_decompose(0x100000000000000000000000000000000);
}
#[test]
fn test_brillig() {
    // Safety: testing context
    let _ = unsafe { brillig_le_decompose(0x100000000000000000000000000000000) };
}
```

### Expected Behavior

I expect the behavior between ACIR and Brillig to match.

### Bug

ACIR will fail for a field larger than the specified num bytes, while for Brillig it pass.

### To Reproduce

1. Copy the test above
2. Run `nargo test`

### Workaround

Yes

### Workaround Description

Manually asserting in Brillig on the limbs 

### Additional Context

_No response_

### Project Impact

Blocker

### Blocker Context

Block for ACIR vs. Brillig correctness

### Nargo Version

nargo version = 1.0.0-beta.7 noirc version = 1.0.0-beta.7+d6920017a3b1ae0a1c927e50b01ca748d901e9ce (git version hash: d6920017a3b1ae0a1c927e50b01ca748d901e9ce, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #9073

## Summary\*
In unconstrained mode, add a check that the input fits in the limbs during radix decomposition.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/test_programs/noir_test_success/to_radix_error/src/main.nr
+++ b/test_programs/noir_test_success/to_radix_error/src/main.nr
@@ -20,9 +20,9 @@ fn test_acir_field_too_big() {
     let _ = constrained_le_decompose(0x100000000000000000000000000000000);
 }
 
-// We do not constrain a decomposition in Brillig. We instead just write up to the number of specified limbs,
-// thus this does not fail in Brillig where the number of limbs determines the correctness of the byte decomposition.
-#[test]
+// We also check in Brillig that the input field fits in the limbs decomposition
+// so that acir and brillig behavior is consistent.
+#[test(should_fail_with = "Field failed to decompose into specified 16 limbs")]
 fn test_brillig() {
     // Safety: testing context
     let _ = unsafe { brillig_le_decompose(0x100000000000000000000000000000000) };


TomAFrench:
I'm thinking that we'd want to have this check in the brillig VM. It'll be much faster and I'm sure that the AVM constrains this already.

guipublic:
> I'm thinking that we'd want to have this check in the brillig VM. It'll be much faster and I'm sure that the AVM constrains this already.

I looked for doing this in the brillig VM but the issue is that to_radix is a brillig building so the VM does not know when we are doing it, and so do it cannot do the check.
I reduced the scope to only to_bits and to_bytes, and this limit the increase in the brillig bytecode.

An alternative would be to create a brillig blackbox function for to_radix so that the brillig VM can handle it.

What do you think is best?

TomAFrench:
I'm not sure what you mean here. We can add an assertion to check that we have enough limbs to hold the result in https://github.com/noir-lang/noir/blob/5d58460a9605d63a4d1672cbba7e535fe5827a41/acvm-repo/brillig_vm/src/black_box.rs#L337-L383.

guipublic:
> I'm not sure what you mean here. We can add an assertion to check that we have enough limbs to hold the result in

Yes I got confused with the to_radix brillig directive which is used by acir gen.
I made the change to use the Brillig VM

Issue #8485, opened 2025-05-13
Panic when using associated types


jfecher:
### Aim

```noir
pub trait Deserialize {
    let N: u32;

    fn deserialize(fields: [Field; Self::N]) -> Self;
}

pub struct PrivateCallInterface<T> {
  return_type: T
}

pub struct PrivateVoidCallInterface {
  return_type: ()
}

trait SuperCallInterface where Self::T: Deserialize<N = Self::M> {
    type T;
    let M: u32;

    fn call(self) -> Self::T;
}

impl<T> SuperCallInterface for PrivateCallInterface<T>
    where T: Deserialize
{
    type T = T;
    type M = <T as Deserialize>::N;

    fn call(self) -> T {
        let preimage: [Field; Self::M] = std::mem::zeroed();

        let returns: T = T::deserialize(preimage);
        returns
    }
}

// Removing this impl fixes the panic despite this type being unused
impl SuperCallInterface<()> for PrivateVoidCallInterface {
    let M: u32 = 0;

    fn call(self) -> () {
        ()
    }
}

unconstrained fn main() {
  let _should_be_field = PrivateCallInterface {return_type: 1}.call();
}
```

### Expected Behavior

No panic

### Bug

```
The application panicked (crashed).
Message:  while substituting (unbound): expected kind of unbound TypeVariable (Normal) to match the kind of its bin
ding (Numeric(u32))
Location: compiler/noirc_frontend/src/hir_def/types.rs:2422

This is a bug. We may have already fixed this in newer versions of Nargo so try searching for similar issues at htt
ps://github.com/noir-lang/noir/issues/.
If there isn't an open issue for this bug, consider opening one at https://github.com/noir-lang/noir/issues/new?lab
els=bug&template=bug_report.yml
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

jfecher:
>     type M = <T as Deserialize>::N;

The panic comes from this line, `M` is correctly created with Kind Normal so its the later unification that shouldn't panic. We should issue a kind error before that point.

jfecher:
Actually, this oddly still panics even if we declare it as `let M: u32 = ...`

PR #9137, opened 2025-07-07
fix: Impl associated types can get out of order

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/8485

## Summary\*

In some cases the associated types of an impl can get out of order, particularly in the presence of programs with other errors in them such as just not having all associated types to begin with (Although they are filled in with Type::Error this can be out of order). It seemed better to sort them when we need them then continue to rely on them being in order everywhere.

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
Looks good!

I wonder if things should be stored as `HashMap` instead, but maybe this isn't generally needed.

Issue #9067, opened 2025-06-30
Stack overflow during `skip_then_find_join_point`


defkit:
### Aim

The following SSA code crashes compiler with stack overflow error
```text
acir(inline) fn main f0 {
  b0(v0: u1, v1: u64, v2: i32, v3: u1, v4: Field, v5: u1, v6: u1):
    jmpif v6 then: b1, else: b2
  b1():
    jmp b14()
  b2():
    jmp b5(u32 67)
  b3():
    jmp b8(u32 198)
  b4():
    jmp b14()
  b5(v7: u32):
    v11 = lt v7, u32 232
    jmpif v11 then: b3, else: b4
  b6():
    v18 = add v7, u32 1
    v19 = truncate v18 to 32 bits, max_bit_size: 33
    v21 = call f1(v4, v4, v4, v4, v4, v0, v0) -> Field
    jmp b5(v19)
  b7():
    jmpif v6 then: b10, else: b11
  b8(v8: u32):
    v14 = lt v8, u32 90
    jmpif v14 then: b7, else: b6
  b9():
    v16 = add v8, u32 1
    v17 = truncate v16 to 32 bits, max_bit_size: 33
    jmp b8(v17)
  b10():
    jmp b12()
  b11():
    jmp b12()
  b12():
    jmp b9()
  b13():
    return v0
  b14():
    jmp b13()
}
acir(inline) fn f1 f1 {
  b0(v0: Field, v1: Field, v2: Field, v3: Field, v4: Field, v5: u1, v6: u1):
    jmp b3(u32 54)
  b1():
    jmpif v6 then: b5, else: b6
  b2():
    jmp b8()
  b3(v7: u32):
    v10 = lt v7, u32 207
    jmpif v10 then: b1, else: b2
  b4():
    v12 = add v7, u32 1
    v13 = truncate v12 to 32 bits, max_bit_size: 33
    jmp b3(v13)
  b5():
    jmp b7()
  b6():
    jmp b7()
  b7():
    jmp b4()
  b8():
    return v4
}
```

And does not crash without `v21 = call f1(v4, v4, v4, v4, v4, v0, v0) -> Field`

![Image](https://github.com/user-attachments/assets/810ee909-3e65-4be3-bc64-563bc2dca701)
0x7317a44 is `noirc_evaluator::ssa::opt::flatten_cfg::branch_analysis::Context::skip_then_find_join_point`

### Expected Behavior

It doesn't crash?

### Bug

.

### To Reproduce

[...]

PR #9139, opened 2025-07-08
fix(ssa): Iterative branch analysis

aakoshh:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/9067

## Summary\*

Changes the `find_join_point_of_branches` implementation in `branch_analysis` from the recursive version to an iterative one, to deal with the large CFG that the test generates. 

## Additional Context

I'm not exactly sure how this CFG was generated, but it does parse and can be evaluated with the SSA interpreter. There are certain empty blocks in it which branch off and rejoin immediately and unconditionally. The resulting CFG has over 75k blocks, so it is reasonable that the recursive algorithm, which looks for where a branching block is rejoined by calling itself on the left and right branches, would at some point hit a stack overflow. 

I managed to induce a stack overflow from Noir with the following program, but only at the end, in the unit test I added I used the program from the ticket. I added an integration test based on the one below.
```rust
fn main(c: bool) -> pub Field {
    let mut s = 0;
    if c {
        for i in 0..232 {
            for j in 0..207 {
                if c {
                    s += (j as Field);
                } else {
                    s += (j as Field);
                }
            }
        }
    }
    s
}
```

The iterative algorithm I implemented is sadly not as intuitive as the recursive version, but I suppose that is to be expected.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/flatten_cfg/branch_analysis.rs
```diff
@@ -72,54 +109,169 @@ impl<'cfg> Context<'cfg> {
     ) -> BasicBlockId {
         let left = successors.next().unwrap();
         let right = successors.next().unwrap();
+        let mut visited = HashSet::new();
+
+        // Kick off the stack from the starting branch. It doesn't have a parent.
+        self.stack.push((start, left, right));
```
asterite, line 115: Is the stack cleared after this method returns? I'm seeing that this method is called in a `loop` so it might be called again and I'm not sure it will start with a fresh state.

aakoshh: Yes, the `while` loop goes until it’s empty. At some point I started with an assertion that it’s empty, but I thought since there is no early return it’s easy to see that this is the case.

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg/branch_analysis.rs
```diff
@@ -72,54 +109,169 @@ impl<'cfg> Context<'cfg> {
...
+                continue;
+            }
+            let left = self.find_next_point(left, false);
+            let right = self.find_next_point(right, false);
 
-        assert_eq!(left_join, right_join, "Expected two blocks to join to the same block");
-        self.branch_ends.insert(start, left_join);
+            if let Some(mut join) = self.maybe_join(branch, &left, &right) {
+                // If we managed to join the branches immediately, then we know where this branch ends,
+                // and we can check if we can complete any parent levels.
```
asterite, line 126: I'm trying to follow the algorithm, it's pretty complex 😅 But, as you say, it's probably expected for going from recursive to iterative.

What does "complete" mean here?

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg/branch_analysis.rs
asterite: Aaah... I missed that the `while` doesn't have a `break`. I saw a break but it was inside the nested loop.

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg/branch_analysis.rs
aakoshh: Yeah, sorry!

Say we have a CFG like this:
```
                   b11
                  /   \
      b5        b9     b13 
     /  \      /  \   /   \
   b1    b7--b8    b12     b15  
  /  \   /     \          /   \
b0    b6        b10----b14     b4
  \                           /  
   b2-----------------------b3 
```

When we consider the branch `b9` into `b11` and `b12`, we get the "next" points on both as `Join(b13)`, so we "immediately" know that `b9` ends at `b13`. By contrast when we considered the branch `b8`, we got `Join(b15)` on one side, and `Branch(b9, b11, b12)` on the other, so it was not an "immediate" join, and all we could do was to take a _pending_ note that `b8` probably ends at `b15`, and then enqueue `b9`. 

Once we know that `b9` ends at `b13`, we have to look forward what is coming after `b13`: it's `Join(b15)`. The parent of `b9` is `b8`, and we already have a pending entry for `b8` finishing at `b15`, so we can "complete" `b8` and write it into `branch_ends`. 

Then, we check whether we can "complete" the parent of `b8`, which is `b0`. Parent here means a branch with a lesser branching factor, like how far are we: we start with `b0`, then increase the level at `b1`, decrease it at `b7`, increase it again at `b8`, and with `b15->b4` we "return to the parent level". 

This parent relationship was nicely done on the stack with the recursive algorithm, whereas here we trace our steps back using the `branch_parents` map.

aakoshh:
Ugh, 500k lines for the integration test snapshots. I think I'll move it to be just a `compile_success`.

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg/branch_analysis.rs
aakoshh: I should probably move that `loop` out from under the `while` into its own method.

aakoshh: Done, also included the illustration from below.

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg/branch_analysis.rs
```diff
@@ -72,54 +109,194 @@ impl<'cfg> Context<'cfg> {
...
+                    let right = successors.next().unwrap();
+                    return Point::Branch(block, left, right);
+                }
+                1 => {
+                    skip = false;
+                    block = successors.next().unwrap();
+                }
+                0 => {
+                    unreachable!(
+                        "return encountered before a join point was found. This can only happen if early-return was added to the language without implementing it by jmping to a join block first"
```
asterite: Just a tiny thing, but mainly to keep cSpell in zero for Rust 😊 
```suggestion
                        "return encountered before a join point was found. This can only happen if early-return was added to the language without implementing it by jumping to a join block first"
```

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg/branch_analysis.rs
asterite: This is great, thank you!

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg/branch_analysis.rs
asterite: This was very useful! I ended up writing a test for this and adding some `println` statements to see the algorithm in action to make sure I understood it well 😊 

Here's the test in case you think it's useful:

```noir
    #[test]
    fn test_find_branch_ends_with_documented_example() {
        //                    b11
        //                   /   \
        //       b5        b9     b13
        //      /  \      /  \   /   \
        //    b1    b7--b8    b12     b15
        //    / \  /      \          /   \
        //  b0   b6        b10----b14    b4
        //    \                          /
        //    b2-----------------------b3
        let src = "
        acir(inline) fn main f0 {
          b0():
            jmpif u1 0 then: b1, else: b2
          b1():
            jmpif u1 0 then: b5, else: b6
          b2():
            jmp b3()
          b3():
            jmp b4()
          b4():
            return
          b5():
            jmp b7()
          b6():
            jmp b7()
          b7():
            jmp b8()
          b8():
            jmpif u1 0 then: b9, else: b10
          b9():
            jmpif u1 0 then: b11, else: b12
          b10():
            jmp b14()
          b11():
            jmp b13()
          b12():
            jmp b13()
          b13():  
            jmp b15()
          b14():
            jmp b15()
          b15():
            jmp b4()
        }
        ";
        let main = ssa.main();
        let cfg = ControlFlowGraph::with_function(main);
        let ends = find_branch_ends(main, &cfg);

        let b = |n| BasicBlockId::new(n);
        assert_eq!(ends[&b(0)], b(4));
        assert_eq!(ends[&b(1)], b(7));
        assert_eq!(ends[&b(8)], b(15));
        assert_eq!(ends[&b(9)], b(13));
    }
```

aakoshh: Thanks for this, I put it in 🙏

Issue #9136, opened 2025-07-07
AsTraitPath forgets associated constant value


asterite:
### Aim

In this code:

```noir
struct Params<let N: u32> {}

trait Trait {
    let N: u32;
    fn params() -> Params<N>;
}

pub struct Foo {}

impl Trait for Foo {
    let N: u32 = 1;

    fn params() -> Params<1> {
        Params {}
    }
}

fn main() {
    let params = <Foo as Trait>::params();
    foo(params);
}

fn foo<let N: u32>(_: Params<N>) {}
```

we get an error:

```
error: Type annotation needed
   ┌─ src/main.nr:20:5
   │
20 │     foo(params);
   │     --- Could not determine the value of the generic argument `N` declared on the function `foo`
```

LSP shows `let params: Params<_>` which means that the `N` couldn't be figured out. In reality it should show `let params: Params<1>` because in the trait impl `impl Trait for Foo` `N` is set to 1.

This is blocking #9133. There it's not exactly the same, but when we have `Self::params()` inside a trait impl method that is resolved in a similar way to `AsTraitPath` and we get the same failure.

### Expected Behavior

The above code should compile fine.

### Bug

The code doesn't compile.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

asterite:
The issue seems to be about generics, not numeric generics. This code also gives the same error:

```noir
struct Params<T> {}

trait Trait {
    type T;
    fn params() -> Params<T>;
}

pub struct Foo {}

impl Trait for Foo {
    type T = i32;

    fn params() -> Params<i32> {
        Params {}
    }
}

fn main() {
    let params = <Foo as Trait>::params();
    foo(params); // Could not determine the type of the generic argument `T` declared on the function `foo`
}

fn foo<T>(_: Params<T>) {}
```

PR #9140, opened 2025-07-08
fix: try to use existing trait constraint when solving AsTraitPath expression

### LINKED ISSUE #9130: Compiler seems to "remember previous T" after modifying Deserialize to use associated constant
### Aim

See [this discussion](https://aztecprotocol.slack.com/archives/C04QF64EDNV/p1751892132327219?thread_ts=1751891691.214659&cid=C04QF64EDNV) on slack for context

### Expected Behavior

...

### Bug

...

### To Reproduce

1. git clone https://github.com/benesjan/noir_slow.git
2. cd noir_slow
3. nargo compile

You should get the following error:

Image

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.7 noirc version = 1.0.0-beta.7+0000000000000000000000000000000000000000

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9130

## Summary

I spent quite some time debugging this. I eventually found that the second test I added, which is very similar to the first one, did compile. The difference there was that in the case of `T::N` that `TraitItem` ended up with a constraint that already exists for `T` because of the where clause. In the case of `<T as Trait>::N` a new trait constraint, which had new type variables for the implicitly added named generics, was used. The solution here is to treat this `<T as Trait>::N` the same way `T::N` is treated, by reusing the existing trait constraint.

That said, as with everything related to trait constraints and type bindings, I'm not 100% sure this is the right way to do this.

## Documentation

Check one:

# PR Checklist


jfecher:
Going to keep this open until I find a solution for https://github.com/noir-lang/noir/issues/9136 to see if that'd fix this as well since the two issues seem related

jfecher:
Closing in favor of https://github.com/noir-lang/noir/pull/9146

PR #9146, opened 2025-07-08
fix: Make `<T as Trait>` no longer an assumed impl

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/9136
Resolves https://github.com/noir-lang/noir/issues/9130

## Summary\*

This ended up being a fairly simple solution. 

How I found it in case it is useful to anyone:
I was looking through the elaboration for TraitAsType and was wondering why we were losing bindings (during type checking, not monomorphization, which eliminates instantiation_bindings as a problem). @asterite's changes to look for existing constraints made me wonder why solving the trait at the end of scope wasn't sufficient. Even if it had unbound type variables those should be bound when the existing impl (even if it is an Assumed one) is found. It turns out the constraint made by `TraitAsType` was itself `assumed: true` and thus we didn't even try solving for it. Changed it to not be assumed and that's all that was needed. 

## Documentation\*

Check one:

# PR Checklist\*


asterite:
> It turns out the constraint made by TraitAsType was itself assumed: true and thus we didn't even try solving for it

Aaaah... I saw this in the code but I assumed (no pun intended) it was correct for some reason.

Would this one too need to be `assumed: false`?

https://github.com/noir-lang/noir/blob/4742c0286e17d9810860b9e7aba96b96b93dc5fd/compiler/noirc_frontend/src/elaborator/types.rs#L742

And I think there are a few more, might be good to double-check them.

jfecher:
> Would this one too need to be assumed: false?

I don't think so - I get several tests failing when changing it. That check is for when we are inside a trait as well so I think we can assume if we're inside a trait default function that the trait itself is implemented. I don't think type variables there would be an issue either since they should stay abstract in the trait itself, but perhaps this could be something to revisit when we add default associated types/constants in the future since to ensure it still works with any values of those and not just the defaults.

PR #9171, opened 2025-07-10
fix(parser): allow type expressions in turbofish named arguments

### LINKED ISSUE #8179: Associated numerics can't be specified in where clause
### Aim

This code compiles fine:

```noir
trait Bar {
    type N;
}

impl Bar for Field {
    type N = i32;
}

fn foo<B>()
where
    B: Bar<N = i32>,
{}
```

However, if we change `N` to be a numeric:

```noir
trait Bar {
    let N: u32;
}

impl Bar for Field {
    let N: u32 = 1;
}

fn foo<B>()
where
    B: Bar<N = 1>,
{}
```

the code fails to compile:

```
error: Bar expects 0 generics but 1 was given
   ┌─ src/main.nr:11:8
   │
11 │     B: Bar<N = 1>,
   │        ---
   │

error: Expected a type but found '1'
   ┌─ src/main.nr:11:16
   │
11 │     B: Bar<N = 1>,
   │                -
   │

error: Expected a `,` separating these two generic parameters
   ┌─ src/main.nr:11:16
   │
11 │     B: Bar<N = 1>,
   │                -
```

### Expected Behavior

### Bug

It doesn't compile. The reason is that the parser actually doesn't allow numbers at that position (`<N = 1>`), it only allows types.

In Rust it compiles (when using `const` instead of `let`) but it requires `#![feature(associated_const_equality)]`.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8179

## Documentation

Check one:

# PR Checklist


jfecher (approved):
Nice!

PR #9172, opened 2025-07-10
fix: disallow trait associated constants to have a default value

### LINKED ISSUE #9138: Default values for associated constants are unimplemented
### Aim

We currently parse:
```
trait Foo {
    let N: u32 = 5;
}
```
but we don't actually implement handling of `= 5` anywhere.

### Expected Behavior

A default should allow this associated constant to be omitted from an impl to use the default automatically.

### Bug

`let N: u32 = 5;` behaves exactly like `let N: u32;`

Suggestion: This would be some additional work to implement (citation needed) but since we also don't support this for associated types we should simply not support this before 1.0 to limit scope. So we should remove parsing support for it until then or issue an error when it is parsed.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9138

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/parser/parser/traits.rs
+++ b/compiler/noirc_frontend/src/parser/parser/traits.rs
@@ -213,12 +213,9 @@ impl Parser<'_> {
             UnresolvedType { typ: UnresolvedTypeData::Unspecified, location }
         };
 
-        let default_value =
-            if self.eat_assign() { Some(self.parse_expression_or_error()) } else { None };
-
         self.eat_semicolon_or_error();
 
-        Some(TraitItem::Constant { name, typ, default_value })
+        Some(TraitItem::Constant { name, typ })
     }
 
     /// TraitFunction = Modifiers Function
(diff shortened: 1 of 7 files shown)


compiler/noirc_frontend/src/parser/parser/traits.rs
```diff
@@ -213,12 +213,9 @@ impl Parser<'_> {
             UnresolvedType { typ: UnresolvedTypeData::Unspecified, location }
         };
 
-        let default_value =
-            if self.eat_assign() { Some(self.parse_expression_or_error()) } else { None };
```
TomAFrench, line 217: Worth keeping this and emitting an error saying that default values are not support here?

asterite: I wasn't sure... but sounds good.

asterite: Done!

PR #9173, opened 2025-07-10
fix(ssa): Do not unroll loops in Brillig with a  constant back-edge value

### LINKED ISSUE #9165: `-Zenums`: Compiler hangs
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/9108

The following program causes the compiler to hang:
```rust
unconstrained fn main() {
    for idx_a in 4048979427..4048979431 {
        assert(
            (
                idx_a
                    == match idx_a {
                        3011893503 => idx_a,
                        198384605 => idx_a,
                        _ => 3138276502,
                    }
            ),
            "PHP",
        );
    }
}

```

### Expected Behavior

It should compile. In fact it does if we remove the `unconstrained`. 

### Bug

```console
❯ cargo run -p nargo_cli -- compile --force -Zenums 
    Finished `dev` profile [optimized + debuginfo] target(s) in 0.51s
     Running `/Users/aakoshh/Work/aztec/noir/target/debug/nargo compile --force -Zenums`
^C⏎   
```

With `--show-ssa` we can see the last successful passes:
```rust
After EnableSideEffectsIf removal (1) (step 25):
  b0():
    jmp b1(u32 4048979427)
  b1(v0: u32):
    v3 = lt v0, u32 4048979431                    	// src/main.nr:2:30
    jmpif v3 then: b2, else: b3
  b2():
    constrain v0 == u32 3138276502, "PHP"         	// src/main.nr:5:17
    v6 = unchecked_add v0, u32 1                  	// src/main.nr:5:17
    jmp b1(v6)
  b3():
    return
}

After Constraint Folding using constraints (1) (step 26):
  b0():
    jmp b1(u32 4048979427)
  b1(v0: u32):
    v3 = lt v0, u32 4048979431                    	// src/main.nr:2:30
    jmpif v3 then: b2, else: b3
  b2():
    constrain v0 == u32 3138276502, "PHP"         	// src/main.nr:5:17
    jmp b1(u32 3138276503)
  b3():
    return
}
```
What seems to happen is that `3138276502` comes from the default arm of the `match` statement, which then because of the `constrain` informs the constraint folding pass to replace `v6 = unchecked_add v0, u32 1` with the constant `3138276503`. That would cause an infinite loop, with a constraint failure in the next loop, but some pass seems to pick up on the structure of the loop and get stuck at compile time. 

The next pass in the pipeline at this point is `make_constrain_not_equal_instructions`.

### To Reproduce

```
NOIR_AST_FUZZER_SHOW_AST=1 CI=1 cargo test -p noir_ast_fuzzer_fuzz min_vs_full -- --nocapture
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.7 noirc version = 1.0.0-beta.7+464125e13369206fd554fc3062627163c353b1c7 (git version hash: 464125e13369206fd554fc3062627163c353b1c7, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9165 

## Summary\*

Fixes the `unrolling` to inspect the induction variable in the back-edge of loops: if the value is a numeric constant, then trying to unroll would be futile. In Brillig we can skip such loops, while in ACIR we return an error.

## Additional Context

The way this happens is when during constraint folding we have a constraint on the induction variable that equates it to a known constant, for example:
```rust
  b0():
    jmp b1(u32 10)
  b1(v0: u32):
    v3 = lt v0, u32 20  
    jmpif v3 then: b2, else: b3
  b2():
    constrain v0 == u32 1
    v6 = unchecked_add v0, u32 1
    jmp b1(v6)
  b3():
    return
}
```
If the `constrain` passes, then we know that the value of `v0` is 1, so the compiler replaces `v6` with 2:
```
  b2():
    constrain v0 == u32 1
    jmp b1(u32 2)
```
This does not make the program invalid in any way: _if_ the constraint was true, then 2 _is_ the next value for the loop variable, even thought the actual bounds were `[10, 20)`, which just means that at runtime (or even at compile time in a later pass), the constraint will fail, before we can run a loop with 2 as the induction value. 

However, the unrolling tracks the value of the induction and evaluates `lt v0, u32 20` at compile time to decide whether it needs to unroll another loop, and in that context, a `v0` that goes `10, 2, 2, 2, 2, ...` results in an infinite loop. 

In Brillig we can leave the loop as it is and let it fail at runtime. 

In ACIR this would be a problem: we _have_ to unroll the loop and we _could_ unroll it, if only constraint folding hadn't got rid of the `unchecked_add` which would increase the value. At the moment there is another unrolling pass before constraint folding though, so hopefully this does not happen in ACIR, because we'd have already unrolled the loops we were able to, but if it does, then we would have to come up with a way to limit what constraint folding can simplify.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
@@ -225,6 +225,18 @@ impl Loops {
             if function.runtime().is_brillig() && !next_loop.is_small_loop(function, &self.cfg) {
                 continue;
             }
+
+            if next_loop.has_const_back_edge_induction_value(function) {
+                // Don't try to unroll this.
+                self.failed_to_unroll.insert(next_loop.header);
+                // If this Brillig, we can still evaluate this loop at runtime.
+                if function.runtime().is_acir() {
+                    unroll_errors
+                        .push(RuntimeError::UnknownLoopBound { call_stack: CallStack::new() });
+                }
+                continue;
+            }
+
             // If we've previously modified a block in this loop we need to refresh the context.
             // This happens any time we have nested loops.
             if next_loop.blocks.iter().any(|block| self.modified_blocks.contains(block)) {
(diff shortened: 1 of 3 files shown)


asterite (approved):
Looks good! The explanation was very helpful.

PR #9176, opened 2025-07-11
fix: handle MakeArray values in DIE for finding mutated array types

### LINKED ISSUE #9160: Nested array corruption in Brilllig
### Aim

The following program:
```noir
fn main() -> pub [[bool; 1]; 1] {
    let mut ctx_limit: u32 = 1;
    unsafe {
        func_1(
            [[true]],
            [[true]],
            ctx_limit,
        )
    }
}
unconstrained fn func_1(
    a: [[bool; 1]; 1],
    mut b: [[bool; 1]; 1],
    mut ctx_limit: u32,
) -> [[bool; 1]; 1] {
    if (ctx_limit == 0) {
        [[true]]
    } else {
        b[0][0] = false;
        println(a);
        b = func_1(a,b,
            ctx_limit - 1,
        );
        a
    }
}
```

returns `[[1]]` when compiled normally, and `[[0]]` with `--force-brillig`. As `a` shouldn't be touched at all, the Brillig version is wrong. If there is no inner call to `func_1`, `a` does not get corrupted even though it happens before the call.

### Expected Behavior

The program prints out `[[true]]` and returns `[[1]]` when ran with `--force-brillig`.

### Bug

```
$ nargo execute --force
[[true]]
[test1] Circuit output: Vec([Vec([Field(1)])])
```

```
$ nargo execute --force --force-brillig
[[false]]
[test1] Circuit output: Vec([Vec([Field(0)])])
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9160

## Summary

In this program:

```
  b0():
    v1 = make_array [u1 1] : [u1; 1]
    v2 = make_array [v1] : [[u1; 1]; 1]
    inc_rc v1
    inc_rc v2 
    return v2
}
```

DIE would previously remove `inc_rc v1`. The reason is that `v1` doesn't show up in an instruction where we track mutated array types (we do this in call arguments, store, return, and some other instructions). For example `inc_rc v2` isn't removed because `v2` is in a `return`. However, since `v2` is a `make_array` that holds `v1`, `inc_rc v1` shouldn't be removed either (I think! At least not removing it fixes the issue).

In this PR, when we analyze a value to see if it has mutated arrays, we also check if the value itself is a `make_array` and, if so, do the same check recursively for each of its values.

I thought about whether we'd need to check other instructions, but I'm not sure. For example what if an element in a `make_array` is a `load`? But in that case the loaded value must have been stored and we would already have tracked that. 

## Documentation

Check one:

# PR Checklist


jfecher (approved):
Good catch.
Cases like this are why I wanted to add `MakeArray` as an instruction instead of having `Array` as a `Value` variant. It is much easier to remember to check this case when having it as an instruction and gives us a concrete time in the program when the array should be initialized as well.

PR #9177, opened 2025-07-11
fix(fuzz): Avoid suffixing slice length modulo blocks with an index access

### LINKED ISSUE #9175: Tests are failing on ARM64
The Rust tests for the Noir compiler are currently failing on ARM64.

Check the [Test (ARM64)](https://github.com/noir-lang/noir/actions/runs/16216193296) workflow for details.

This issue was raised by the workflow `Test (ARM64)`

aakoshh:
# Description

## Problem\*

Resolves #9175 

## Summary\*

Changes the handling of `Slice` sources in `gen_expr_from_source` so that instead of something like this:
```rust
{
    let s: [[u8; 4]] = ...;
    s[1234 % s.len()]
}[5678 % 4]
```
it generates:
```rust
{
    let s: [[u8; 4]] = ...;
    s[1234 % s.len()][5678 % 4]
}
```

## Additional Context

When it comes to slices, unlike with arrays, we don't know their length until runtime, so in order to access them with a chance of not getting _Index out of bounds_ all the time, we take a modulo with their `.len()`. To do so, the generator first assigns the slice to a temporary variable, because the slice might be coming from a function call for example, or a literal we don't want to repeat. Because we have to return an `Expression`, the only way to do this is to return a `Block`. 

Then, the generator considered how it can derive the actual target type from the returned `Block`. However, not every expression that forms a valid AST can be parsed back by the frontend, and it turns out it doesn't like parsing `{...}[...]` as indexing an array returned by the block, but rather as a block whose result is unused, followed by an `[u32; 1]` array. 

To avoid this, we can put the final derivation inside the block as well, where it should line up with the accessing of the slice member.

## Documentation\*

Check one:

# PR Checklist\*


tooling/ast_fuzzer/tests/parser.rs
```diff
@@ -76,16 +76,25 @@ fn arb_ssa_roundtrip() {
...
                 let msg = passes.last().map(|p| p.msg()).unwrap_or("Initial SSA");
                 print_ast_and_panic(&format!(
                     "Could not parse SSA after step {last_pass} ({msg}): \n{e:?}"
                 ))
             });
 
+        ssa2.normalize_ids();
+
+        // println!("{}", ssa1.print_without_locations());
+        // println!("{}", ssa2.print_without_locations());
```
asterite: Are these changes needed?

asterite:
I guess another way to fix this is, when printing `obj[index]`, if `obj` is a block then surround it by parentheses.

tooling/ast_fuzzer/tests/parser.rs
aakoshh: They helped identify the issue but no, removed.

aakoshh:
> is a block then surround it by parentheses.

Thanks, I didn't think about that. Added it in the comments 🙏

jfecher (approved):

PR #9183, opened 2025-07-11
fix(SSA): validate MakeArray instruction

### LINKED ISSUE #9158: ACIR vs Brillig; array of bigger size than defined
### Aim

The following SSA failed in ACIR and solved in brillig
```
            fn main f0 {
              b0():
                text = make_array [u8 0, u8 12, u8 3, u8 15] : [u8; 4]
                iv   = make_array [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 16]
                key  = make_array [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 16]
                ct   = call aes128_encrypt(text, iv, key) -> [u8; 16]
                return ct
            }
```
Note that `iv` length is 17 

### Expected Behavior

Both failed

### Bug

Acir failed with: SSA compilation failed: SSA compilation failed: RuntimeError(InternalError(General { message: "Panic occurred: Panic: Compiler should generate correct size inputs: [FunctionInput { input: Constant(0), num_bits: 8 }, FunctionInput { input: Constant(0), num_bits: 8 }, FunctionInput { input: Constant(0), num_bits: 8 }, FunctionInput { input: Constant(0), num_bits: 8 }, FunctionInput { input: Constant(0), num_bits: 8 }, FunctionInput { input: Constant(0), num_bits: 8 }, FunctionInput { input: Constant(0), num_bits: 8 }, FunctionInput { input: Constant(0), num_bits: 8 }, FunctionInput { input: Constant(0), num_bits: 8 }, FunctionInput { input: Constant(0), num_bits: 8 }, FunctionInput { input: Constant(0), num_bits: 8 }, FunctionInput { input: Constant(0), num_bits: 8 }, FunctionInput { input: Constant(0), num_bits: 8 }, FunctionInput { input: Constant(0), num_bits: 8 }, FunctionInput { input: Constant(0), num_bits: 8 }, FunctionInput { input: Constant(0), num_bits: 8 }, FunctionInput { input: Constant(0), num_bits: 8 }] at compiler/noirc_evaluator/src/acir/acir_context/generated_acir/mod.rs:194", call_stack: [] })), brillig succeeded

### To Reproduce

In ssa_executor tests

```rs
    #[test]
    fn foo() {
        let ssa_without_runtime = "
            (inline) fn main f0 {
              b0():
                text = make_array [u8 0, u8 12, u8 3, u8 15] : [u8; 4]
                iv   = make_array [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 16]
                key  = make_array [u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0, u8 0] : [u8; 16]
                ct   = call aes128_encrypt(text, iv, key) -> [u8; 16]
                return ct
            }
        ";
        let acir_ssa = "acir".to_string() + ssa_without_runtime;
        let brillig_ssa = "brillig".to_string() + ssa_without_runtime;
        let witness_map = WitnessMap::new();
        let acir_result = execute_ssa(acir_ssa.to_string(), witness_map.clone(), CompileOptions::default());
        let brillig_result = execute_ssa(brillig_ssa.to_string(), witness_map.clone(), CompileOptions::default());
        match (acir_result, brillig_result) {
            (Err(acir), Ok(_brillig)) => panic!("Acir failed with: {}, brillig succeeded", acir),
            (Ok(_acir), Err(brillig)) => panic!("Acir succeeded, brillig failed: {}", brillig),
            (Err(acir), Err(brillig)) => println!("ACIR {} Brillig {} ", acir, brillig),
            _ => {}
        }
    }
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
Blocked by #8918 as currently the databus puts values of the wrong type in MakeArray.

# Description

## Problem

Resolves #9158

## Summary

With the validation in some `multi_scalar_mul` simplification tests started to fail. It turned out that some `MakeArray` types were incorrect there so this PR fixes that too.

## Additional Context

While doing this I realized call argument types aren't checked against the target function so I might do that in a separate PR.

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/ssa/validation/mod.rs
```diff
@@ -258,6 +260,36 @@ impl<'f> Validator<'f> {
                     );
                 }
             }
+            Instruction::MakeArray { elements, typ: _ } => {
+                let results = dfg.instruction_results(instruction);
+                assert_eq!(results.len(), 1, "MakeArray must return exactly one value");
+
+                let result_type = dfg.type_of_value(results[0]);
+                if !matches!(result_type, Type::Slice(_)) {
```
aakoshh: This is great! But isn't there something we can check for `Slice` as well? 

For example we could take the `composite_type` without the `length`, and check that `elements.len() % composite_type.len() == 0`, that is, we don't have partial entries. 

The iteration over the elements for type checking also seems like it could work for slices, it doesn't need the `length`.

asterite: Good ideas! Done.

compiler/noirc_evaluator/src/ssa/validation/mod.rs
```diff
@@ -258,6 +260,50 @@ impl<'f> Validator<'f> {
...
+                        composite_type
+                    }
+                    _ => {
+                        panic!("MakeArray must return an array or slice type, not {result_type}");
+                    }
+                };
+
+                for (index, element) in elements.iter().enumerate() {
+                    let element_type = dfg.type_of_value(*element);
+                    let expected_type = &composite_type[index % composite_type.len()];
```
aakoshh: Not sure if `composite_type.len()` is cheap to compute, but maybe it could be lifted out of the loop? Or maybe the compiler does it on its own? Just a thought.

compiler/noirc_evaluator/src/ssa/validation/mod.rs
```diff
@@ -603,4 +649,49 @@ mod tests {
...
+    fn make_array_has_incorrect_element_type() {
+        let src = "
+        acir(inline) fn main f0 {
+          b0():
+            v0 = make_array [u8 1, Field 2, u8 3, u8 4] : [(u8, u8); 2]
+            return v0
+        }
+        ";
+        let _ = Ssa::from_str(src).unwrap();
+    }
```
aakoshh, line 712: Maybe add a test for slices as well, just to make sure it doesn't get uncovered in the future?

aakoshh (approved):
Great stuff 👏

aakoshh:
Are the circuit size changes expected?

asterite:
> Are the circuit size changes expected?

Good question. I hope so! 😄

asterite:
> Are the circuit size changes expected?

The `multi_scalar_multiplication` simplification was returning incorrect types. For example instead of `[(Field, Field, u1); N]` it was returning `[Field; 6]` which I guess later on has some different implications compared to the correct type. But... I'll compare the SSAs to see what changed to understand this more.

asterite:
I just checked the SSA difference and it seems now we can reuse existing arrays. Previously they would get an incorrect type after simplification so they didn't match previous arrays. Or at least that's one difference...

compiler/noirc_evaluator/src/ssa/validation/mod.rs
asterite: Done!

compiler/noirc_evaluator/src/ssa/validation/mod.rs
asterite: Done!

asterite:
@TomAFrench Could you check if this change is okay? Mainly to make sure the performance improvements are correct and that I didn't break anything 😅

vezenovm (approved):
I looked over the improvements and as you mentioned they look to be happening as simplifications revealed themselves once we accurately marked the array types. When looking at things to optimize in the past I was always thinking the `embedded_curve_add` was more expensive than expected. I didn't investigate too deeply where the exact simplification occurs that has knockdown benefits, but the types you marked look correct.

PR #9185, opened 2025-07-11
chore: add acir-gen unit tests per ssa instruction (2)

guipublic:
# Description

## Problem\*

Same PR as #8084 
I re-do it due to issues with PR review.
## Summary\*
Add a framework for acir-gen unit tests.
It generates SSA for a single instruction, converts it using acir-gen and execute it.
Then we compare the result with a SSA evaluation using constant folding.
All 'simple' instructions are covered, in particular all the binary instructions.
More tests with different types or missing instructions such as array operations and calls could be added in future.

## Documentation\*

Check one:

# PR Checklist\*


jfecher (approved):
The bigcurve failures are unrelated and are happening on every other PR and master currently

compiler/noirc_evaluator/src/acir/tests/mod.rs
```diff
@@ -924,3 +925,208 @@ fn do_not_overflow_with_constant_constrain_neq() {
...
+        // SSA interpreter will emit an error but not ACVM
+        // "and",
+        // "xor",
+        "unchecked_add",
+        "unchecked_sub",
+        "unchecked_mul",
+        "range_check 32",
+        "truncate 32 254",
+    ];
+    let inputs = [FieldElement::from(1_usize), FieldElement::from(2_usize)];
```
TomAFrench: We're only testing a single input here which means we're not going to cover a lot of interesting inputs. Can you look at tying this into proptest so we can compile the circuits once and then throw a lot of inputs into it?

guipublic: Done, although it does compile the circuits per input. I don't know how to avoid this since the inputs are inputs to the test function. However, the circuit is only a few lines (one circuit per instruction) so I don't think it matters.

TomAFrench (approved):

Issue #9193, opened 2025-07-14
Side effects of reading from a zero-length array escapes into other branches.


TomAFrench:
Consider the program below:

```noir
fn main(a: bool) {
    if a {
        1 / [][0];
    };
}
```

If we pass `a = false` then this program passes in brillig but will fail to execute in ACIR. In this case the final SSA is

```
acir(inline) predicate_pure fn main f0 {
  b0(v0: u1):
    v1 = make_array [] : [Field; 0]                     // src/main.nr:4:9
    constrain u1 0 == v0, "Index out of bounds"         // src/main.nr:4:16
    enable_side_effects u1 0
    v4 = array_get v1, index u32 0 -> Field             // src/main.nr:4:13
    v6 = div Field 1, v4                                // src/main.nr:4:9
    enable_side_effects u1 1
    return
}
```

As the array_get and div instructions are under `enable_side_effects u1 0` and their results are unused then we should remove them from the program entirely as they should not affect execution at all. The final program would then be

```
acir(inline) predicate_pure fn main f0 {
  b0(v0: u1):
    constrain u1 0 == v0, "Index out of bounds"         // src/main.nr:4:16
    return
}
```

guipublic:
Closing this as 'no repro'. I believe this has been indirecly fixed.

PR #9195, opened 2025-07-14
fix(acir-gen): Do not generate ACIR array opcodes when the length is zero

### LINKED ISSUE #9174: ACIR != Brillig: Index Out Of Bounds vs Divide By Zero
### Aim

Discovered testing https://github.com/noir-lang/noir/pull/9061/
Potentially related to https://github.com/noir-lang/noir/issues/9159

The following program fails on different parts of the code in ACIR and Brillig:

```rust
global G_A: [Field] = &[];
fn main(a: bool, b: bool, mut c: call_data(2) (i64, Field, Field)) -> pub (Field, Field) {
    if a {
        (
            {
                let i = G_A;
                i[(2050918985_u32 % i.len())]
            },
            unsafe {
                func_1_proxy(
                    if a {
                        if b {
                            (unsafe { func_1_proxy(0) } as u32)
                        } else {
                            766482358_u32
                        }
                    } else {
                        1568313658_u32
                    },
                )
            },
        )
    } else {
        (
            {
                let k = G_A;
                k[(779011912_u32 % k.len())]
            }, c.2,
        )
    }
}

unconstrained fn func_1_proxy(_a: u32) -> Field {
    0
}

```

Inputs:
```toml
a = false
b = false
c = [
    "0x000000000000000000000000000000000000000000000000566516d121a04755",
    "0x0bf4513bb0a3a24efbb0decbd8995eba23fa0e9236156abd0b1c4b1f88e19ddf",
    "0x08c9c54133258070e36902247f2c89f746736c2abeaba3283c66587b49e9b512",
]
```

### Expected Behavior

They should both fail with "divide by zero" accessing `k` because they are taking a modulo with the length of an empty slice, and `a` is `false`. 

### Bug

```console
❯ cargo run -q -p nargo_cli -- execute --force --silence-warnings; cargo run -q -p nargo_cli -- execute --force-brillig --silence-warnings
  ┌─ src/main.nr:7:17
  │
7 │                 i[(2050918985_u32 % i.len())]
  │                 -----------------------------
  │
  = Call stack:
    1. src/main.nr:7:17

error: Assertion failed: 'Attempted to divide by zero'
   ┌─ src/main.nr:27:20
   │
27 │                 k[(779011912_u32 % k.len())]
   │                    -----------------------
   │
   = Call stack:
     1. src/main.nr:27:20

Failed assertion

```

ACIR fails with `Index out of bounds` on `i`, which is in a branch that should not even run. 

### To Reproduce

Check out the PR branch at `9a317c8cd88756b95310686f845f3e432784519b`, then run:
```shell
NOIR_AST_FUZZER_SEED=0x90fce797000a0a98 cargo test -p noir_ast_fuzzer_fuzz acir_vs_brillig 
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.7 noirc version = 1.0.0-beta.7+464125e13369206fd554fc3062627163c353b1c7 (git version hash: 9a317c8cd88756b95310686f845f3e432784519b, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9174 

## Summary\*

Changes `handle_array_operations` in ACIR generation to skip generating opcodes for accessing an array when know that the array/slice has a zero size. 

## Additional Context

The code in question has a global zero sized slice, which it accesses in both branches of an _if_ expression:

```rust
global G_A: [Field] = &[];
fn main(a: bool, b: bool, c: Field) -> pub (Field, Field) {
    if a {
        (
            {
                let i = G_A;
                i[(2050918985_u32 % i.len())]
            },
            ...,
        )
    } else {
        (
            {
                let k = G_A;
                k[(779011912_u32 % k.len())]
            }, c,
        )
    }
}
```

The final SSA began like this:
```rust
After Verifying no dynamic array indices to reference value elements (1) (step 40):
g0 = u32 0
g1 = make_array [] : [Field]

  b0(v2: u1, v3: u1, v4: Field):
    enable_side_effects v2                          // If `a` is true
    v6 = mod u32 2050918985, u32 0                	// 2050918985_u32 % i.len(), but only fail if `a` is true
    constrain u1 0 == v2, "Index out of bounds"   	// always fail if `a` is true, because the length is 0
    v8 = make_array [] : [Field]                  	// initialize an empty array for the slice
    v9 = array_get v8, index v6 -> Field          	// access the array, but by now we know we will have failed
    enable_side_effects u1 0
...
```
And the generated ACIR:
```rust
func 0                                              // main
...                            
private parameters indices : [_0, _1, _2]           // a, b, c
...
BLACKBOX::RANGE [(_1, 1)] []                        // `b` is bool
EXPR [ (1, _0) 0 ]                                  // because the modulo is known to fail if `a` is true, `a` _must_ be false for the circuit to succeed
INIT (id: 0, len: 0, witnesses: [])                 // initialize an empty array for the slice
EXPR [ (-1, _5) 0 ]                                 // create a witness for the index, reading from index 0
MEM (id: 0, read at: EXPR [ (1, _5) 0 ], value: EXPR [ (1, _6) 0 ]) // read an item from the empty array
EXPR [ -1 ]                                         // fail unconditionally
...
```

The program failed with the following error:
```
  ┌─ src/main.nr:7:17
  │
7 │                 i[(2050918985_u32 % i.len())]
```
So this branch should never have run, but the `MEM` opcode fails anyway, because even though it looks at the side effect variable to decide whether to get/set items, it always expected a non-empty array.

With the changes in the PR, this is now eliminated:
```
func 0
...
BLACKBOX::RANGE [(_1, 1)] []
EXPR [ (1, _0) 0 ]
EXPR [ -1 ]
...
```

In the original approach of the PR the elimination only happened on the top branch, where it was obvious that the code would never run, because `v2` was constrained to be 0. Then I changed it to happen whenever we know the array/slice is empty, which includes the `else` branch:

```rust
g0 = u32 0
g1 = make_array [] : [Field]

  b0(v2: u1, v3: u1, v4: Field):
...
    v17 = not v2
...
    enable_side_effects v17
    v27 = mod u32 779011912, u32 0                	// src/main.nr:27:20
    constrain v2 == u1 1, "Index out of bounds"   	// src/main.nr:27:20
    v29 = array_get v8, index v27 -> Field        	// src/main.nr:27:17
    enable_side_effects u1 1
...
    return v34, v37
}
```
in ACIR this no longer has memory operations either:
```rust
...
BRILLIG CALL func 1: inputs: [Single(Expression { mul_terms: [], linear_combinations: [], q_c: 779011912 }), Single(Expression { mul_terms: [], linear_combinations: [], q_c: 0 })], outputs: [Simple(Witness(5)), Simple(Witness(6))]
BLACKBOX::RANGE [(_5, 33)] []
BLACKBOX::RANGE [(_6, 0)] []
EXPR [ (-1, _7) 0 ]
BLACKBOX::RANGE [(_7, 0)] []
EXPR [ (-1, _6) 779011912 ]
EXPR [ -1 ]
EXPR [ 1 ]
EXPR [ (1, _3) 0 ]
EXPR [ (-1, _2) (1, _4) 0 ]
...
```
So we call Brillig to calculate the modulo (we can look separately into why the compiler did not eliminate this call), then fail unconditionally. 

While the reason for the presence of the Brillig call is still something we can work out, the fix should unblock the PRs that fail due to the difference in ACIR and Brillig.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/nargo_cli/tests/snapshots/execution_success/regression_9193/execute__tests__force_brillig_false_inliner_-9223372036854775808.snap
+++ b/tooling/nargo_cli/tests/snapshots/execution_success/regression_9193/execute__tests__force_brillig_false_inliner_-9223372036854775808.snap
@@ -17,6 +17,10 @@ expression: artifact
     ],
     "return_type": null,
     "error_types": {
+      "12639702852095440652": {
+        "error_kind": "string",
+        "string": "Index out of bounds, array has size 0"
+      },
       "14225679739041873922": {
         "error_kind": "string",
         "string": "Index out of bounds"
--- a/tooling/nargo_cli/tests/snapshots/execution_success/regression_9193/execute__tests__force_brillig_false_inliner_0.snap
+++ b/tooling/nargo_cli/tests/snapshots/execution_success/regression_9193/execute__tests__force_brillig_false_inliner_0.snap
@@ -17,6 +17,10 @@ expression: artifact
     ],
     "return_type": null,
     "error_types": {
+      "12639702852095440652": {
+        "error_kind": "string",
+        "string": "Index out of bounds, array has size 0"
+      },
       "14225679739041873922": {
         "error_kind": "string",
         "string": "Index out of bounds"
--- a/tooling/nargo_cli/tests/snapshots/execution_success/regression_9193/execute__tests__force_brillig_false_inliner_9223372036854775807.snap
+++ b/tooling/nargo_cli/tests/snapshots/execution_success/regression_9193/execute__tests__force_brillig_false_inliner_9223372036854775807.snap
@@ -17,6 +17,10 @@ expression: artifact
     ],
     "return_type": null,
     "error_types": {
+      "12639702852095440652": {
+        "error_kind": "string",
+        "string": "Index out of bounds, array has size 0"
+      },
       "14225679739041873922": {
         "error_kind": "string",
         "string": "Index out of bounds"


TomAFrench:
Looks like this covers #9193, ~shall we swap out the regression test to have something simpler?~ ah looks like the case is slightly different as the indexing is in both branches.

TomAFrench:
> While the reason for the presence of the Brillig call is still something we can work out, the fix should unblock the PRs that fail due to the difference in ACIR and Brillig.

Yeah, I agree that this isn't the best solution for this issue but it's relatively light touch so if it unblocks master then I'm happy to merge it as a stopgap.

I think the ideal solution on this is to handle this in SSA where I think we can be more aggressive on marking certain instructions as unreachable.

aakoshh:
I checked locally that it does unblock https://github.com/noir-lang/noir/pull/9108 and https://github.com/noir-lang/noir/pull/9061/

I agree it would be nice to make unreachable instruction removal take care of this. As discussed it has a bit of cascading effect at the moment: maybe we could replace variables with defaults that are defined until the next side effect variable, or do it earlier, but then it disrupted flattening. @asterite was looking at it.

TomAFrench (approved):

jfecher:
> #9174 

Apologies for missing this earlier, but it looks like this PR https://github.com/noir-lang/noir/pull/9050 also tackles the problem of changing acri-gen's behavior to not do `array[index * predicate]` for 0-length arrays

> Changes handle_array_operations in ACIR generation to skip generating opcodes for accessing an array when i) we know that the side effect variable is zero and ii) the array/slice has a zero size.

We should also change the behavior not to do `array[index * predicate]` even if the predicate is not known to be zero

aakoshh:
@jfecher thanks for the pointer to Guillaume's PR, I changed it to work like his by inserting an assertion that the side effect variable is false, and doing it whenever the `flattened_size` is zero, which covers slices and arrays as well. Not sure if I'm not inviting differences with the interpreter this way, let's see.

jfecher:
> which covers slices and arrays as well. Not sure if I'm not inviting differences with the interpreter this way, let's see.

maybe, but we can leave it to the other PR to resolve those differences at least

PR #9200, opened 2025-07-14
feat: only inject "out of bounds" checks in brillig

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This is a refinement to https://github.com/noir-lang/noir/pull/7827 which omits the out-of-bounds check in situations where we can rely on the implicit range check from the underlying memory block to prevent out-of-bounds accesses

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
There's a couple of outstanding issues here:
- `unused_array_get_known_index_out_of_bounds` is now passing rather than halting. It looks like the array access is being removed in DIE rather than being replaced with an out-of-bounds check.
- `regression_8774` fails with "not expected to have Load or Store instruction after DIE in an ACIR function". This is due to the out-of-bounds check which previously constrained the index to be 0 being removed so we're not resolving the reference immediately like we were before.

TomAFrench:
Merging master into this branch atm.

vezenovm (approved):

Issue #9201, opened 2025-07-14
Trait constraint type inference regression with Self type


jfecher:
### Aim

```noir
pub trait BigCurve<B> {
    fn one() -> Self;
}

struct Bn254 {}

impl<B> BigCurve<B> for Bn254 {
    fn one() -> Self { Bn254 {} }
}

fn main() {
    let _ = <Bn254 as BigCurve<()>>::one();
}
```

### Expected Behavior

The code to compile

### Bug

```
error: Type annotation needed
   ┌─ src/main.nr:12:14
   │
12 │     let _ = <Bn254 as BigCurve<()>>::one();
   │              ----------------- Could not determine type of generic argument
   │

Aborting due to 1 previous error
```
This is due to us not always binding Self in https://github.com/noir-lang/noir/blob/11d86ef432ffc23a2f90b55ba12833ded4c467a3/compiler/noirc_frontend/src/elaborator/types.rs#L2283.
However, removing the `if assumed {` to always bind it causes us to expect `&mut &mut PoseidonHasher` (etc) where `&mut PoseidonHasher` should be correct. This is because when doing method calls we generate a trait constraint with the object type directly, which may be `&mut T` while the method may be declared as `fn method(&mut self)` so `Self` should be `T`, not `&mut T` - so we wind up with the `&mut &mut T`.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #9202, opened 2025-07-14
fix: Fix type inference regression with trait constraints and the Self type

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/9201

## Summary\*

This is an issue with the Self type not being bound in certain trait constraints.
1. In https://github.com/noir-lang/noir/blob/11d86ef432ffc23a2f90b55ba12833ded4c467a3/compiler/noirc_frontend/src/elaborator/types.rs#L2283 we only bind self `if assumed` instead of always
2. If we remove that if we run into issues where the `constraint.typ` doesn't match the `Self` type when calling methods. For example, `constraint.typ = &mut T` but the method is `fn method(&mut self)`, so `Self` should be `T`, not `&mut T`.
3. We don't match against the function and object type until _after_ the trait constraint and HirIdent is generated from a method call so I just added a hack for now.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/expressions.rs
+++ b/compiler/noirc_frontend/src/elaborator/expressions.rs
@@ -629,12 +629,13 @@ impl Elaborator<'_> {
 
                 let location = object_location.merge(method_name_location);
 
-                let (function_id, mut function_name) = method_ref.clone().into_function_id_and_name(
-                    object_type.clone(),
-                    generics.clone(),
-                    location,
-                    self.interner,
-                );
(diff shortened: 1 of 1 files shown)


jfecher:
Looks like this is causing us to actually bind the original Self type in some cases. E.g. `nargo expand` output shows we change:
```noir
trait Default {
    fn default2() -> Self;
}
```
into
```noir
trait Default {
    fn default2() -> Field;
}
```
for example, with one field impl.

PR #9203, opened 2025-07-14
feat: optimize array reads from length 1 arrays

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This optimization deals with a couple of avoidable issues which arose in #9200. If we have a length 1 array then we can just optimistically read from index 0 and inject an assertion that the index will be zero.

This allows us to simplify some later instructions however the results are not shown against master due to #7827 currently injecting constraints which turn all of these array accesses into constant index accesses currently.

## Documentation\*

Check one:

# PR Checklist\*


guipublic (approved):
LGTM

PR #9208, opened 2025-07-15
fix(acir-gen): Use the array length for 0 check, not the storage size

### LINKED ISSUE #9206: ACIR != Brillig: Index out of bounds, array has size 0
### Aim

One of the nightly failures in https://github.com/noir-lang/noir/actions/runs/16283361382/job/45977181286
Probably introduced by https://github.com/noir-lang/noir/pull/9195

The following program fails with _Index out of bounds, array has size 0_ in ACIR, passes in Brillig:

```rust
fn main(b: u8) -> pub u8 {
    let _d = unsafe { func_3(func_3([], b)[1], 0) };
    b
}
unconstrained fn func_3(_a: [u128; 0], _b: u8) -> [[u128; 0]; 2] {
    G_A
}
```

inputs:
```toml
b = 0
```

### Expected Behavior

They should both pass. 

### Bug

```console
❯ cargo run -q -p nargo_cli -- execute --force --silence-warnings -Zenums ; cargo run -q -p nargo_cli -- execute --force --silence-warnings -Zenums --force-brillig
error: Assertion failed: Index out of bounds, array has size 0
  ┌─ src/main.nr:3:30
  │
3 │     let _d = unsafe { func_3(func_3([], b)[1], 0) };
  │                              ----------------
  │
  = Call stack:
    1. src/main.nr:3:30

Failed assertion
[fuzz_testing] Circuit output: Field(0)
```

There is some mixup with the array sizes, because `func_3` returns `[_; 2]` so `func_3(...)[1]` should not fail. The fact that the items in the array are zero sized should not cause IOOB, but it looks like somehow the ACIR generation considers it has to access the array to pass it to `func_3` again. 

This works: `let _d = unsafe { func_3([], b)[1] };` 

The bug also goes away if we stop passing `b` to the inner call. 

I reckon this has to do with the way I use `flattened_size` to check if the array or slice has 0 size. The outer array has size 2, but the inner one has 0, so overall the data structure doesn't take up any space. This can easily be fixed for arrays, I'll check how it works for slices with empty arrays inside.

### To Reproduce

```
NOIR_AST_FUZZER_SEED=0x035b8fb600100000 cargo test -p noir_ast_fuzzer_fuzz acir_vs_brillig 
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.8 noirc version = 1.0.0-beta.8+fe44a1e273ee55ae9cd713cc5baf43e5b51601d7 (git version hash: 49c49ce89ff8a5fbd76df25686ffe60d3ff77c4a, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9206

## Summary\*

Fixes the ACIR codegen for array operations to do the 0-length check using `Array::Type(_, length)` rather than `flattened_size`, which can be 0 for arrays that contain data that takes up no space.

## Additional Context

I think `flattened_size` works for slices for this check, at least so far I couldn't find a counter example.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/acir/arrays.rs
```diff
@@ -814,24 +814,42 @@ impl Context<'_> {
...
+    ///
+    /// This is different from `flattened_size` in that a non-zero length
+    /// array containing zero length arrays has zero size, but we can still
+    /// access its elements.
+    fn has_zero_length(&mut self, array: ValueId, dfg: &DataFlowGraph) -> bool {
+        if let Type::Array(_, size) = dfg.type_of_value(array) {
+            size == 0
+        } else {
+            match &dfg[array] {
+                Value::NumericConstant { constant, .. } => constant.is_zero(),
```
TomAFrench: This is unreachable, no? An array can't be a numeric constant.

compiler/noirc_evaluator/src/acir/arrays.rs
```diff
@@ -814,24 +814,42 @@ impl Context<'_> {
...
+                Value::Instruction { .. } | Value::Param { .. } => {
                     let array_acir_value = self.convert_value(array, dfg);
-                    size += flattened_value_size(&array_acir_value);
+                    match array_acir_value {
+                        AcirValue::DynamicArray(AcirDynamicArray { len, .. }) => len == 0,
+                        AcirValue::Var(_, _) => false,
```
TomAFrench: This should also be unreachable.

compiler/noirc_evaluator/src/acir/arrays.rs
aakoshh: I'm not sure, I took it from `flattened_size` where the parameter naming suggests we are dealing with arrays. I thought maybe it was somehow related to the separate size variable for slices.

compiler/noirc_evaluator/src/acir/arrays.rs
aakoshh: I took this from `flattened_value_size`, but I suppose there it's about the contents of the array 👍

compiler/noirc_evaluator/src/acir/arrays.rs
aakoshh: Removed this match so it's `unreachable!` now.

TomAFrench (approved):
LGTM!

PR #9211, opened 2025-07-15
feat(ssa): Following an always failing binary, replace instructions with defaults until the next predicate

aakoshh:
# Description

## Problem\*

Follow up for https://github.com/noir-lang/noir/pull/9195#issuecomment-3069923577

## Summary\*

Adds new logic to `remove_unreachable_instructions` to handle always failing binary instructions which are under a predicate by entering an `UnreachableUnderPredicate` mode, in which instructions that have side effects get removed, and their results replaced by default values. This mode persists until the next `enable_side_effects` instruction, or the end of the block.

The reasoning to do so is that if the binary operation is known to fail, then until side effect conditions change, it doesn't matter if it was enabled or not, the subsequent operations cannot have an effect, so we can get rid of them. We keep their results, because within the same block those might be used, but we replace them with default values that may not even appear in the SSA as instructions, such as numeric constants.

## Additional Context

Looking at the SSA of the program from the predecessor ticket, it changes as follows:

```console
❯ cargo run -q -p nargo_cli -- execute --force --show-ssa --silence-warnings --show-ssa-pass "step 31" --show-ssa-pass "step 32"
After Mem2Reg (4) (step 31):
g0 = u32 0
g1 = make_array [] : [Field]

acir(inline) predicate_pure fn main f0 {
  b0(v2: u1, v3: u1, v4: Field):
    enable_side_effects v2
    v6 = mod u32 2050918985, u32 0                	// src/main.nr:7:20
    constrain u1 0 == v2, "Index out of bounds"   	// src/main.nr:7:20
    v8 = make_array [] : [Field]                  	// src/main.nr:7:20
    v9 = array_get v8, index v6 -> Field          	// src/main.nr:7:17
    enable_side_effects u1 0
    v11 = call f1(u32 0) -> Field                 	// src/main.nr:14:29
    v12 = truncate v11 to 32 bits, max_bit_size: 254	// src/main.nr:14:29
    v13 = cast v12 as u32                         	// src/main.nr:14:29
    v14 = not v3
    v15 = unchecked_mul v2, v14
    v16 = cast v15 as u32                         	// src/main.nr:14:29
    v18 = unchecked_mul v16, u32 766482358        	// src/main.nr:14:29
    v19 = not v2
    v20 = unchecked_mul v2, v19
    v21 = cast v2 as u32                          	// src/main.nr:14:29
    v22 = cast v20 as u32                         	// src/main.nr:14:29
    v23 = unchecked_mul v21, v18                  	// src/main.nr:14:29
[...]

PR #9212, opened 2025-07-15
feat: mark instructions after invalid array operations as unreachable

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

Currently we don't mark array operations as "always failing" under any circumstances. We can however do this if we know that the index points past the end of the array at compile time or that the array has length zero.

This PR then marks these instructions as always failing.

## Documentation\*

Check one:

# PR Checklist\*


jfecher (approved):

TomAFrench:
Let's merge this after #9211 as it'll cause conflicts and that one has more of an immediate impact.

TomAFrench:
This merge was janky and likely incorrect.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -173,7 +173,35 @@ impl Function {
...
+                            Some(predicate) => predicate.is_one(),
+                            None => false, // The predicate is a variable
+                        };
+
+                    let array_or_slice_type = context.dfg.type_of_value(*array);
+                    match array_or_slice_type {
+                        Type::Slice(_) => (),
+                        array_type @ Type::Array(_, len) => {
+                            if len == 0 {
+                              current_block_reachability = Reachability::Unreachable;
```
aakoshh: I think this shouldn't necessarily mean `Unreachable`. I would treat it the same way as the `Binary` operations:
* Ceteris paribus, does it always fail?
* If so, are we under a predicate? 
  * If we are under a constant one predicate, the whole block will be unreachable
  * If we are under a zero or variable predicate, subsequent side effecting instructions can be zeroed out

TomAFrench: My reasoning for marking this as `unreachable` is that any array read from a zero length array will fail as there's no value to read.

Granted that's complicated a bit by https://github.com/noir-lang/noir/pull/9195 as this prevent ACIR from failing. Perhaps we should push this to a future PR which prevents acirgen/brilliggen from emitting zero-length arrays entirely.

aakoshh: Just to be clear in the PR you linked, what we are saying is that if the array has 0 length _and_ the side effect variable is [enabled](https://github.com/noir-lang/noir/pull/9195/files#diff-f6eddbb80ca1ba7234edd355f9d0cac33c577309265a48da4cd0a3f1555c2f51R122) then we fail the circuit with an error message, rather than attempt the memory operation. If the side effects are off, then nothing happens and we move on, which was impossible with the way the disabled array operations worked. 

Here I am just pointing out that you are right, if the array has 0 length then it _would_ fail, but it should only mean the end of the block if we know that the side effect variable is a constant 1. Otherwise it may or may not be enabled, and we should use `UnreachableUnderPredicate` instead. 

So I would reorganise it like this:
```rust
let array_op_always_fails = match instruction { 
  Instruction::ArrayGet { array, index, offset } => ... len == 0 || index - offset >= len
};
if array_op_always_fails {
  let is_predicate_constant_one = ...;
  current_block_reachability = if is_predicate_constant_one { Unreachable } else { UnreachableUnderPredicate };
}
```
The lower part can probably be factored out from the binary and this.

aakoshh: Maybe what's confusing in the merge is that in Ary's version if the predicate was not 1, the logic `return`d without coming to inspect the array, but now we inspect the array/binary first, and then reason about the scope of what needs to become unreachable.

PR #9215, opened 2025-07-16
fix(fuzz): `Value::snapshot` needs to be recursive

aakoshh:
# Description

## Problem\*

Resolves one of the nightly failures in https://github.com/noir-lang/noir/actions/runs/16309675831/job/46062679969

## Summary\*

Changes `Value::snapshot` in the SSA interpreter to recurse into arrays, rather than do a shallow snapshot. 

## Additional Context

The problem manifested itself like this:
```console
❯ NOIR_AST_FUZZER_SEED=0x758873ee00100000 cargo test -p noir_ast_fuzzer_fuzz pass_vs_prev
    Finished `test` profile [optimized + debuginfo] target(s) in 0.20s
     Running unittests src/lib.rs (target/debug/deps/noir_ast_fuzzer_fuzz-6990cb9d73ef591d)

running 1 test
test targets::pass_vs_prev::tests::fuzz_with_arbtest ... FAILED

failures:

---- targets::pass_vs_prev::tests::fuzz_with_arbtest stdout ----
---
Comparison failed:
both programs passed, but disagree on printed output:
---
[[true], [true], [false]]
[[true], [true], [true]]

--- != ---
[[true], [true], [true]]
[[true], [true], [true]]

---
---
AST:
unconstrained fn main(mut a: pub [[bool; 1]; 3]) -> pub [i8; 2] {
    println(a);
    {
        let idx_b: u32 = (3558586064_u32 % 3_u32);
        a[idx_b][0_u32] = ...;
    };
    println(a);
    ...
}
...

---
ABI Inputs:
a = [[true], [true], [false]]
...

thread 'targets::pass_vs_prev::tests::fuzz_with_arbtest' panicked at tooling/ast_fuzzer/fuzz/src/targets/mod.rs:76:18:
called `Result::unwrap()` on an `Err` value: both programs passed, but disagree on printed output:
---
[[true], [true], [false]]
[[true], [true], [true]]

--- != ---
[[true], [true], [true]]
[[true], [true], [true]]
```

The second invocation of the interpreter saw the change on the input enacted by the first invocation.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/ast/expression.rs
+++ b/compiler/noirc_frontend/src/ast/expression.rs
@@ -67,24 +67,45 @@ pub type UnresolvedGenerics = Vec<UnresolvedGeneric>;
 
 #[derive(Debug, PartialEq, Eq, Clone, Hash)]
 pub enum UnresolvedGeneric {
-    Variable(Ident, Vec<TraitBound>),
-    Numeric {
-        ident: Ident,
-        typ: UnresolvedType,
-    },
+    Variable(IdentOrQuotedType, Vec<TraitBound>),
+    Numeric { ident: IdentOrQuotedType, typ: UnresolvedType },
+}
+
+#[derive(Debug, PartialEq, Eq, Clone, Hash)]
+pub enum IdentOrQuotedType {
+    Ident(Ident),
 
     /// Already-resolved generics can be parsed as generics when a macro
     /// splices existing types into a generic list. In this case we have
     /// to validate the type refers to a named generic and treat that
     /// as a ResolvedGeneric when this is resolved.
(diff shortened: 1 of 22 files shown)


TomAFrench:
Looks good, can you pull out a regression test so we don't need to rely on the fuzzer for this?

TomAFrench (approved):
Good to merge with the above testcase

aakoshh:
Sure, I'll add a unit test for `snapshot`. This method was actually added specifically for the fuzzer, which reuses the input between passes. It just so happens that `Value::clone` is not enough to make a detached copy.

PR #9219, opened 2025-07-16
feat: Add `compiler_unstable_features` to `Nargo.toml`

### LINKED ISSUE #9164: Enable experimental features in `Nargo.toml`
There seems to be currently no way to tell the LSP that to enable an experimental feature. The VS Code plugin has a field for CLI arguments, but putting `-Zenums` there has no effect. 

This makes it harder to work with Noir programs that have a `match` statements, as the LSP highlights them as error, asking for the feature to be enabled. 

It would be a helpful if we could put into the `Nargo.toml` file a list of features the project requires to compile, so `nargo` can pick it up from there as well as the CLI args.

aakoshh:
# Description

## Problem\*

Resolves #9164 

## Summary\*

Adds a new `compiler_unstable_features` field to `Nargo.toml` under `[package]`, which can be used by the current package or its dependencies to require an unstable feature to be enabled during compilation for files in that crate. 

This allows the LSP to pick up whether the currently edited package does or doesn't need a feature, without having to edit the plugin settings where CLI arguments can be passed. 

The requirement is on a crate-by-crate basis, so if dependency `foo` requires `enums`, but another dependency `bar` does not state that it does, then it would be an error in `bar` to use a `match`, unless the feature is enabled globally via CLI with `-Zenums`.

Required unstable features are ignored if features are explicitly enabled on the CLI, e.g. if `foo` requires `enums`, but we pass `-Zownership`, then enums are still disabled, because we interpret `-Z` as the developer taking control of what exactly is acceptable globally.

The PR also adds a `--no-unstable-features` CLI argument to ignore all requirements in manifests, in case someone finds it important to make sure no unstable features sneaked in from dependencies.

## Additional Context

Example:
```toml
[package]
name = "fuzz_testing"
type = "bin"
authors = [""]
compiler_unstable_features = ["enums"]

[dependencies]
```

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/elaborator/options.rs
```diff
@@ -1,6 +1,6 @@
 use std::str::FromStr;
 
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
```
aakoshh:
```suggestion
#[derive(Copy, Clone, Debug, PartialEq, Eq)]
```

rkarabut (approved):

PR #9230, opened 2025-07-17
fix(ssa): Change constraint message to "multiply"

### LINKED ISSUE #9225: Unexpected error message in "expand signed checks"
### Aim

Replicate one of the nightly failures in https://github.com/noir-lang/noir/actions/runs/16335526975/job/46146707738

The following program reports overflow on `mul` in the Initial SSA, then switches to overflow on `add` during the expansion of signed checks. 

```rust
fn main(a: i32) -> pub i32 {
    (1231707443_i32 * a)
}
```

Input:
```toml
a = "-0x0000000000000000000000000000000000000000000000000000000000000002"
```

### Expected Behavior

I expect the message to talk about multiplication, which is also suggested by how the [error equivalence](https://github.com/noir-lang/noir/blob/7cf7a89ca9593893dffe5175a5a4bf55110fb00a/tooling/ast_fuzzer/src/compare/interpreted.rs#L189-L202) logic is set up.

### Bug

```console
❯ cargo run -q -p nargo_cli -- interpret  --force --silence-warnings -Zenums --ssa-pass Initial --ssa-pass expand --show-ssa-pass Initial --show-ssa-pass expand
After Initial SSA:
acir(inline) fn main f0 {
  b0(v0: i32):
    v2 = mul i32 1231707443, v0                   	// src/main.nr:2:6
    return v2
}

Err(An overflow occurred while evaluating `mul i32 1231707443, v0` (mul i32 1231707443, i32 -2))
---
After expand signed checks (step 1):
acir(inline) fn main f0 {
  b0(v0: i32):
    v2 = unchecked_mul i32 1231707443, v0         	// src/main.nr:2:6
    v3 = cast v2 as u64                           	// src/main.nr:2:6
    v4 = truncate v3 to 32 bits, max_bit_size: 64 	// src/main.nr:2:6
    v5 = cast v0 as u32                           	// src/main.nr:2:6
    v7 = lt v5, u32 2147483648                    	// src/main.nr:2:6
    v8 = not v7                                   	// src/main.nr:2:6
    v9 = cast v0 as Field                         	// src/main.nr:2:6
    v10 = cast v7 as Field                        	// src/main.nr:2:6
    v11 = mul v10, v9                             	// src/main.nr:2:6
    v13 = sub Field 4294967296, v9                	// src/main.nr:2:6
    v14 = cast v8 as Field                        	// src/main.nr:2:6
    v15 = mul v14, v13                            	// src/main.nr:2:6
    v16 = add v11, v15                            	// src/main.nr:2:6
    v18 = mul Field 1231707443, v16               	// src/main.nr:2:6
    range_check v18 to 32 bits, "attempt to multiply with overflow"	// src/main.nr:2:6
    v19 = truncate v18 to 32 bits, max_bit_size: 254	// src/main.nr:2:6
    v20 = cast v19 as u32                         	// src/main.nr:2:6
    v21 = not v7                                  	// src/main.nr:2:6
    v22 = cast v21 as u32                         	// src/main.nr:2:6
    v23 = unchecked_add u32 2147483648, v22       	// src/main.nr:2:6
    v24 = lt v20, v23                             	// src/main.nr:2:6
    constrain v24 == u1 1, "attempt to add with overflow"	// src/main.nr:2:6
    v26 = cast v4 as i32                          	// src/main.nr:2:6
    return v26
}

--- Interpreter result after expand signed checks (step 1):
Err(constrain v24 == v25, "attempt to add with overflow" failed:
    u1 0 != u1 1)
---
```

As we can see the SSA contains both "attempt to multiply with overflow" and "attempt to add with overflow". Maybe despite being an add, the error message on the latter should carry the name of the operation of the original that caused it to be inserted, or maybe the first constraint should have failed.

### To Reproduce

```shell
NOIR_AST_FUZZER_SEED=0xe5c7258700100000 cargo test -p noir_ast_fuzzer_fuzz pass_vs_prev
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.8 noirc version = 1.0.0-beta.8+ef3a9fde86ae56a68d81bfedcd87b2fa88f54635 (git version hash: 7cf7a89ca9593893dffe5175a5a4bf55110fb00a, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9225 

## Summary\*

Changes the error message of the constraints inserted by the `expand_signed_checks` pass to match the operation. 

## Additional Context

`check_signed_overflow` handles `Add`, `Sub` and `Mul`, but the `message` only depended on whether the operation was `Sub` in which case it was "subtract", otherwise it was "add". The `Mul` strategy had its own error message with "multiply", and used the sub/add message after an addition operation; but that addition was part of the overall multiplication overflow check, so I think it's an implementation detail, and potentially causes confusion. It's not clear whether this distinction was deliberate, or the result of merging strategies. 

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/expand_signed_checks.rs
```diff
@@ -163,13 +166,13 @@ impl Context<'_, '_, '_> {
...
+        if matches!(operator, BinaryOp::Sub { .. }) {
             // lhs - rhs = lhs + (-rhs)
             rhs_sign = self.insert_not(rhs_sign);
-            "attempt to subtract with overflow".to_string()
-        } else {
-            "attempt to add with overflow".to_string()
-        };
+        }
+
+        let message = format!("attempt to {operation} with overflow");
```
vezenovm, line 174:
```suggestion
        let message = format!("attempt to {operator} with overflow");
```
We have a Display impl on `BinaryOp`

compiler/noirc_evaluator/src/ssa/opt/expand_signed_checks.rs
```diff
@@ -163,13 +166,13 @@ impl Context<'_, '_, '_> {
         let rhs_as_unsigned = self.insert_safe_cast(rhs, NumericType::unsigned(bit_size));
         let lhs_sign = self.insert_binary(lhs_as_unsigned, BinaryOp::Lt, half_width);
         let mut rhs_sign = self.insert_binary(rhs_as_unsigned, BinaryOp::Lt, half_width);
-        let message = if is_sub {
+        if matches!(operator, BinaryOp::Sub { .. }) {
```
vezenovm, line 169: We should keep this change of not returning the message, but we can utilize the `BinaryOp` display impl. We can then remove the need to pass an `operation` parameter to this function.

compiler/noirc_evaluator/src/ssa/opt/expand_signed_checks.rs
aakoshh: True, but it says `sub` and `mul` instead of `subtract` and `multiply`, which is what the tests expect. I suppose in written error messages the full verb reads better, but not sure why `Display` uses the abbreviated forms.

aakoshh: It actually uses `unchecked_add` as well, so it's not meant to appear in English text.

vezenovm: Hmm, the shortened word is probably to reduce clutter in the SSA display. It seems that the current `Display` impl should in fact be a Debug impl. We could then update the Display impl to "pretty prints" the operators with their long form words. I won't block on this, I just thought it would be nice to avoid the extra parameter.

aakoshh: I don't have a strong opinion on it, I got used to `Debug` being derived, full of details but difficult to read. For these we have verbs that conveniently go into that sentence, but to change `Display` I would need to come up with something suitable for operations like `Eq`, `And` and `Lt`, which are less obvious. The verbs used here are probably the exception, rather than the rule.

PR #9232, opened 2025-07-17
fix(ssa): Do not remove array operations that can be OOB during DIE 

### LINKED ISSUE #9223: Array operations which would fail execution are removed by DIE if the result is unused
If we take the below program and compile it then execution will succeed despite the fact that we perform a read past the end of the array. 

```
          b0(v0: [Field; 1]):
            v1 = array_get v0, index u32 2 -> Field
            return v0
        }
```

This is incorrect as this program should fail. I've opened a PR which implements these changes here: https://github.com/noir-lang/noir/pull/9220

vezenovm:
# Description

## Problem\*

Resolves #9223 

## Summary\*

Builds off of https://github.com/noir-lang/noir/pull/9200

This replaces https://github.com/noir-lang/noir/pull/9220 which branches off master. The motivation to have these changes branch off one another is that we have several test failures triggerd by https://github.com/noir-lang/noir/pull/9200 (the original parent PR). We can more easily get a passing CI but having any fixes built off one another.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
@@ -172,8 +172,9 @@ impl Ssa {
         }
 
         for function in self.functions.values_mut() {
-            #[cfg(debug_assertions)]
-            flatten_cfg_pre_check(function);
+            // Disabled due to failures
+            // #[cfg(debug_assertions)]
+            // flatten_cfg_pre_check(function);
 
             flatten_function_cfg(function, &no_predicates);
 
@@ -192,6 +193,7 @@ impl Ssa {
 ///
 /// Otherwise panics.
 #[cfg(debug_assertions)]
+#[allow(dead_code)]
 fn flatten_cfg_pre_check(function: &Function) {
     function.reachable_blocks().iter().for_each(|block| {
         if let TerminatorInstruction::JmpIf { condition, .. } =


TomAFrench:
I took a look at the cause of the regressions in this PR and it seems to be down to us now having a lot of instructions would have been removed by die which are sitting under a constant zero predicate.

It's a bit annoying to deal with these in die (as we'd need to do lookahead to determine the predicate). Chucking a `simplify_cfg` in front of flatten_cfg should either fix it or give us a thread to pull on.

vezenovm:
> I took a look at the cause of the regressions in this PR and it seems to be down to us now having a lot of instructions would have been removed by die which are sitting under a constant zero predicate

Which ones do you see are sitting under a constant zero?

Some of the circuit regressions look to be unavoidable without extra handling in DIE. We may have to bring back some form of the OOB checks inserted during DIE that were removed in https://github.com/noir-lang/noir/pull/7995.

For example this program taken from `regression_7612`:
```noir
pub struct Data {
    fields: [Field; 1],
    counter: u32,
}

fn main(array: [Data; 1], x: bool) {
    let index = if x { 0 } else { 1 };
    if index != 0 {
        assert(array[index - 1].counter < 3);
    }
}
```
Produces this SSA:
```
After Dead Instruction Elimination - ACIR (1) (step 43):
  b0(v0: [([Field; 1], u32); 1], v1: u1):
    v2 = not v1
    enable_side_effects u1 1
    v4 = cast v2 as u32
    v6 = eq v4, u32 0
    v7 = not v6
    enable_side_effects v7
    v9 = sub v4, u32 1
    v11 = unchecked_mul v9, u32 2
    v12 = array_get v0, index v11 -> [Field; 1]
    v13 = unchecked_add v11, u32 1
    v14 = array_get v0, index v13 -> u32
    v16 = lt v14, u32 3
    v17 = unchecked_mul v16, v7
    constrain v17 == v7
    enable_side_effects u1 1
    return
}
```
Previously we would remove `v12 = array_get v0, index v11 -> [Field; 1]` and be confident that the separate OOB kept us safe. However, we can still safely remove this `v12` result as it is unused and part of the same array access (if next array get would fail if it was OOB). We would only check this for ACIR. It would complicate DIE again though by bringing back some form of what was removed in #7995.

TomAFrench:
~I found the regression in `no_predicates_numeric_generic_poseidon`, I get constant condition jmpifs appearing in the mem2reg just before flattening.~

More details in #9275, tl;dr is that this is a no_predicates issue.

vezenovm:
I'm looking at bring back removing unused array get groups during DIE.

vezenovm:
> I'm looking at bring back removing unused array get groups during DIE.

This is showing to be promising.

TomAFrench:
huh, sorry I seem to have broken this.

PR #9236, opened 2025-07-17
fix: bind self when type-checking AsTraitPath

### LINKED ISSUE #9201: Trait constraint type inference regression with Self type

asterite:
# Description

## Problem

Resolves #9201

## Documentation

Check one:

# PR Checklist


jfecher (approved):
LGTM - thank you

We should document how type bindings work for different ast nodes in general before auditing the elaborator

PR #9237, opened 2025-07-17
chore: bump linked bb version

TomAFrench:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/integration-tests/package.json
+++ b/compiler/integration-tests/package.json
@@ -13,7 +13,7 @@
     "lint": "NODE_NO_WARNINGS=1 eslint . --max-warnings 0"
   },
   "dependencies": {
-    "@aztec/bb.js": "0.87.0",
+    "@aztec/bb.js": "1.0.0-staging.6",
     "@noir-lang/noir_js": "workspace:*",
     "@noir-lang/noir_wasm": "workspace:*",
     "@nomicfoundation/hardhat-chai-matchers": "^2.0.8",
--- a/scripts/install_bb.sh
+++ b/scripts/install_bb.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-VERSION="0.87.0"
+VERSION="1.0.0-staging.6"
 
 BBUP_PATH=~/.bb/bbup
 
--- a/yarn.lock
+++ b/yarn.lock
@@ -238,20 +238,20 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@aztec/bb.js@npm:0.87.0":
-  version: 0.87.0
-  resolution: "@aztec/bb.js@npm:0.87.0"
+"@aztec/bb.js@npm:1.0.0-staging.6":
+  version: 1.0.0-staging.6
+  resolution: "@aztec/bb.js@npm:1.0.0-staging.6"
   dependencies:
     comlink: "npm:^4.4.1"
     commander: "npm:^12.1.0"
-    debug: "npm:^4.3.4"
-    fflate: "npm:^0.8.0"
+    idb-keyval: "npm:^6.2.1"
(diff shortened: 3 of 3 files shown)


TomAFrench:
@saleel, it looks like there's been changes to the bb CLI + bb.js. Do you know why these are now failing?

saleel:
I can think of:
- Verification key `-k` is now expected to be provided to the `bb prove` command
- `UltraHonkZK` flavor is the default now in the CLI, but it was not implemented in wasm when the above release was made  😔 (so you have different proof size on cli vs js unless you add `--disable-zk` flags). 
- `KeyHash` is now enforced on recursion - not sure if that was included in the above release

I would say we update to version 1.2.0 when that comes out

asterite:
A couple of tests fail with:

> Contract code size is 25518 bytes and exceeds 24576 bytes

I don't know how the size can be reduced. I tried using a newer Solidity compiler, or running more optimizations, but it led to even more code.

TomAFrench:
We can tell hardhat to ignore the max contract size limit: https://v2.hardhat.org/hardhat-network/docs/reference#allowunlimitedcontractsize

asterite:
Ah, I saw that but I didn't know if it was okay to bypass that restriction.

In any case now it fails with this:

> Error: VM Exception while processing transaction: reverted with custom error 'ProofLengthWrong()'

I'll check if there's something that can be done...

saleel:
> Error: VM Exception while processing transaction: reverted with custom error 'ProofLengthWrong()'

@asterite  Can you try changing [this](https://github.com/noir-lang/noir/blob/master/compiler/integration-tests/test/node/smart_contract_verifier.test.ts#L50) to `keccakZK: true`

ZK variant is the default in cli now iirc. `keccak: true` means use the non-ZK version of solidity verifier - which is what we had before (these config names needs to be improved)

asterite:
@saleel Thanks! Is it just in that line or also a few lines below that?

- If I change only that line I get "Proof fails verification in JS"
- If I change both I get "VM Exception while processing transaction: reverted with custom error 'SumcheckFailed()'"

saleel:
It should be both those lines. 
Let me give it a try now.

asterite:
I feel like the issue is on barretenberg's side.

There are a bunch of imports like this:

```ts
import { pino } from "pino";
```

However, I'm seeing that [recent tutorials](https://betterstack.com/community/guides/logging/how-to-install-setup-and-use-pino-to-log-node-js-applications/) say to import it like this:

```ts
import pino from "pino";
```

Maybe pino exposes both forms of imports? Not sure if that's possible. I also don't know how it would work on barretenberg's CI but not here. I wanted to try that change but `./bootstrap.sh` doesn't work for me because it seems it doesn't work on Macs.

asterite:
Or maybe not... at least that change didn't work in Aztec-Packages.

asterite:
This seems to always fail when `UltraHonkBackend` is used. But it also fails if we just have this code:

```ts
let backend = UltraHonkBackend;
```

That is, just referring to that type will produce a failure. That's why maybe mocking pino doesn't work (though it is strange that we get a different error).

saleel:
Yes, the error is different now. I changed the config to use the relative path `rootDir: fileURLToPath(new URL('./', import.meta.url)),`
and now it seems to go to a different error  - "Buffer is not defined" (which is a polyfill issue, and Buffer is being used in bb.js)

Trying to polyfill that, but not sure how it was working until now with the `  rootDir: fileURLToPath(new URL('./../..', import.meta.url))` path

TomAFrench:
Thanks for addressing this @asterite and @saleel!

PR #9238, opened 2025-07-17
feat(ssa): Remove all non side effectual instructions from unreachable blocks during DIE

vezenovm:
# Description

## Problem\*

This helps resolve a failure with `regression_8994` in https://github.com/noir-lang/noir/actions/runs/16346960365/job/46184088603?pr=9232. I decided to pull it out as it should apply more generally.

No issue but I noticed after https://github.com/noir-lang/noir/pull/9008 that we can be significantly more aggressive about remove instructions with `Unreachable` terminators. 

## Summary\*

If we have a block that terminates with `unreachable`, we can remove all instructions in the block (not just those after the unreachable), if the instruciton:
- Does not have side effects
- Is not the trap that triggered the unreachable terminator
- And does not produce results used in later instructions 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/die.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/die.rs
@@ -245,7 +245,9 @@ impl Context {
         for instruction_id in block.instructions().iter().rev() {
             let instruction = &function.dfg[*instruction_id];
 
-            if self.is_unused(*instruction_id, function) {
+            if self.is_unused(*instruction_id, function)
+                || self.can_remove_with_unreachable_terminator(function, block, *instruction_id)
+            {
                 self.instructions_to_remove.insert(*instruction_id);
             } else {
                 // We can't remove rc instructions if they're loaded from a reference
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -1179,4 +1206,60 @@ mod test {
...
     }
+
+    #[test]
+    fn remove_all_instructions_in_unreachable_block_but_final_trap() {
+        let src = "
+        acir(inline) predicate_pure fn main f0 {
+          b0():
+            v0 = allocate -> &mut u1                            
+            v2 = make_array [u1 0, v0] : [(u1, &mut u1); 1]     
+            v4 = array_get v2, index u32 2 -> u1
```
aakoshh, line 1217: Shouldn't the current DIE pass (without the changes in this PR) already remove these as unused?

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -290,6 +292,31 @@ impl Context {
...
+            return false;
+        };
+
+        let last_instruction = block.instructions().last().copied();
+        let is_last = Some(instruction_id) == last_instruction;
+        if is_last {
+            return false;
+        }
+
+        self.is_unused(instruction_id, function)
```
aakoshh, line 317: I was wondering if we can find a way to refactor this so that `self.is_unused` isn't called twice (`self.is_unused(...) || self.can_remove_with_unreachable_terminator(...)` where the latter calls the former again). 

I must be missing something, but when do we return from this construct anything but what `is_unsed` would have returned already? All we return is `false` from the early returns, and then end up in the same case as what the `||` already checked.

compiler/noirc_evaluator/src/ssa/opt/die.rs
vezenovm: Yes it would. I can update the test, but after https://github.com/noir-lang/noir/pull/9232 none of these will be removed.

compiler/noirc_evaluator/src/ssa/opt/die.rs
vezenovm: Yeah good point, I'll refactor this

vezenovm:
I worked on this a little fast. After fixing the logic there are some other bugs I am getting with uninstantiated variables. Converting back to a draft for now.

compiler/noirc_evaluator/src/ssa/opt/die.rs
vezenovm: I am going to branch off of #9232 as this is mostly relevant for that case, and I'm not sure if this is the best solution anymore.

PR #9249, opened 2025-07-18
fix(licm): Ensure that all nested loops the current block is part of are guaranteed to execute

### LINKED ISSUE #9243: Min != Full: Assertion failure in loop that never runs
### Aim

Replicates a nightly failure in https://github.com/noir-lang/noir/actions/runs/16361498650/job/46230095734

The following program fails with the full pipeline, but passes with `--minimal-ssa`:

```rust
fn main() {
    for idx_b in 119718805667097306623756199175472805749_u128..119718805667097306623756199175472805750_u128 {
        let start = 169386661893075804622328296976770135180_u128 % 9_u128;
        let end = idx_b % 9_u128;
        // println((start, end));
        for _ in start..end {
            assert((idx_b <= 115517568315105269094017991206916906366_u128), "UHL");
        }
    }
}
```

### Expected Behavior

I expect them to agree. If I uncomment the `println` then we can see the loop boundaries are `[4,2)`, which means this loop should never execute. (The `println` also prevents the bug from materialising, so it's left commented out).

### Bug

```console
❯ cargo run -q -p nargo_cli -- execute --force --silence-warnings; cargo run -q -p nargo_cli -- execute --force --silence-warnings --minimal-ssa
error: Assertion failed: UHL
  ┌─ src/main.nr:7:21
  │
7 │             assert((idx_b <= 115517568315105269094017991206916906366_u128), "UHL");
  │                     -----------------------------------------------------
  │
  = Call stack:
    1. src/main.nr:7:21

Failed assertion

```

It fails if main is `unconstrained` as well (or with `--force-brillig`)

### To Reproduce

```shell
NOIR_AST_FUZZER_SEED=0xe8d35a6300100000 cargo test -p noir_ast_fuzzer_fuzz min_vs_full
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.8 noirc version = 1.0.0-beta.8+ef3a9fde86ae56a68d81bfedcd87b2fa88f54635 (git version hash: 1719bc3a7977248e792816df3a3e907f2d006a95, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9243 

## Summary\*

Adds more book keeping to Loop Invariant Code Motion:
* In the beginning determine any known fixed bounds of each loop
* When we are making decisions whether to hoist an invariant out of a block into the pre-header of a loop that it's part of, visit all nested loops between the current loop and the block, and check whether they too are guaranteed to execute.

## Additional Context

Say we have loops like this:
```rust
'foo: for i in 10 .. 20 {
    'bar: for j in 30 .. 40 {
      'baz: for k in 5 .. 0 {
        assert(i < 10);
      }
    }
}
```
When we are processing `'bar`, we have:
* `current_induction_variables` as ` { j: [100, 110] }`
* `outer_induction_variables` as `{ i: [0, 10] }`

When we make a decision about whether to hoist `assert(i < 10)` from the body of `'baz` into the pre-header of `'bar`, the current logic only looked at `current_induction_variables` to check that the `'bar` loop executes, but it did not consider whether the `'baz` loop itself executes or not. 

It did consider whether the loops nested in `'bar` that the block from which the instruction would be hoisted is control dependent on something, but in the case of the SSA in the program in the integration test, the Post Dominance Frontier was completely empty. Since we couldn't use the CFG to inform us about the fact that `'baz` may or may not execute, we have to explicitly look at the loop bounds of all nested loops we go through to get to the block.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -2,9 +2,18 @@
 # Do not cancel the test run on the first failure.
 fail-fast = false
 
+# We do not run `ram_blowup_regression` by default as it's very slow
+default-filter = "not test(ram_blowup_regression)"
+
+[profile.ci-master]
+# Do not cancel the test run on the first failure.
+fail-fast = false
+
+default-filter = "all()"
+
 [profile.merge-queue]
 # fail fast to kick from merge queue faster.
 fail-fast = true
 
 # Disable fuzzing to avoid flakiness
-default-filter = "not (package(noir_ast_fuzzer_fuzz) or test(arb_program_freqs_in_expected_range))"
+default-filter = "not test(ram_blowup_regression) or not (package(noir_ast_fuzzer_fuzz) or test(arb_program_freqs_in_expected_range))"
--- a/.github/workflows/test-rust-workspace-arm64.yml
+++ b/.github/workflows/test-rust-workspace-arm64.yml
@@ -79,7 +79,7 @@ jobs:
           RUST_MIN_STACK=8388608 \
           cargo nextest run --archive-file nextest-archive-arm64.tar.zst \
             --partition count:${{ matrix.partition }}/4 \
-            --profile ci
+            --profile ci-master
 
   # This is a job which depends on all test jobs and reports the overall status.
   # This allows us to add/remove test jobs without having to update the required workflows.
--- a/.github/workflows/test-rust-workspace-msrv.yml
+++ b/.github/workflows/test-rust-workspace-msrv.yml
@@ -93,7 +93,7 @@ jobs:
           RUST_MIN_STACK=8388608 \
           cargo nextest run --archive-file nextest-archive.tar.zst \
             --partition count:${{ matrix.partition }}/4 \
-            --profile ci
+            --profile ci-master
 
   # This is a job which depends on all test jobs and reports the overall status.
   # This allows us to add/remove test jobs without having to update the required workflows.
--- a/.github/workflows/test-rust-workspace.yml
+++ b/.github/workflows/test-rust-workspace.yml
@@ -83,7 +83,7 @@ jobs:
             --partition count:${{ matrix.partition }}/4 \
             --profile $NEXTEST_PROFILE
         env:
-          NEXTEST_PROFILE: ${{ (github.event_name == 'merge_group' && 'merge-queue') || 'ci' }}
+          NEXTEST_PROFILE: ${{ (github.event_name == 'merge_group' && 'merge-queue') || (github.ref == 'refs/heads/master' && 'ci-master') || 'ci' }}
 
   # This is a job which depends on all test jobs and reports the overall status.
   # This allows us to add/remove test jobs without having to update the required workflows.
--- a/EXTERNAL_NOIR_LIBRARIES.yml
+++ b/EXTERNAL_NOIR_LIBRARIES.yml
@@ -113,7 +113,7 @@ libraries:
     repo: AztecProtocol/aztec-packages
     ref: *AZ_COMMIT
     path: noir-projects/noir-protocol-circuits/crates/rollup-lib
-    timeout: 700
+    timeout: 800
     critical: false
     # Use 1 test threads for rollup-lib because each test requires a lot of memory, and multiple ones in parallel exceed the maximum memory limit.
     nargo_args: "--test-threads 1"
--- a/compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
@@ -877,7 +877,7 @@ impl<'ssa, W: Write> Interpreter<'ssa, W> {
         let index = self.lookup_u32(index, "array get index")?;
         let mut index = index - offset.to_u32();
 
-        let element = if array.elements.borrow().len() == 0 {
+        let element = if array.elements.borrow().is_empty() {
             // Accessing an array of 0-len is replaced by asserting
             // the branch is not-taken during acir-gen and
             // a zeroed type is used in case of array get
--- a/tooling/nargo_toml/src/lib.rs
+++ b/tooling/nargo_toml/src/lib.rs
@@ -266,6 +266,7 @@ impl PackageConfig {
 /// Contains all the information about a package, as loaded from a `Nargo.toml`.
 #[derive(Debug, Deserialize, Clone)]
 #[serde(untagged)]
+#[allow(clippy::large_enum_variant)]
 pub enum Config {
     /// Represents a `Nargo.toml` with package fields.
     Package {
--- a/tooling/ssa_fuzzer/fuzzer/Cargo.toml
+++ b/tooling/ssa_fuzzer/fuzzer/Cargo.toml
@@ -8,6 +8,9 @@ license.workspace = true
 [package.metadata]
 cargo-fuzz = true
 
+[lints]
+workspace = true
+
 [dependencies]
 libfuzzer-sys = { version = "0.4.0", features = ["arbitrary-derive"] }
 noirc_evaluator.workspace = true
--- a/tooling/ssa_fuzzer/fuzzer/src/fuzz_lib/fuzz_target_lib.rs
+++ b/tooling/ssa_fuzzer/fuzzer/src/fuzz_lib/fuzz_target_lib.rs
@@ -52,7 +52,7 @@ pub(crate) fn fuzz_target(data: FuzzerData, options: FuzzerOptions) -> Option<Fi
 
     let mut fuzzer = Fuzzer::new(data.instruction_blocks, values, options);
     for func in data.functions {
-        log::debug!("initial_witness: {:?}", witness_map);
+        log::debug!("initial_witness: {witness_map:?}");
         log::debug!("commands: {:?}", func.commands);
         fuzzer.process_function(func, types.clone());
     }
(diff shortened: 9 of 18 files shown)


compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -155,17 +162,17 @@ struct LoopInvariantContext<'f> {
     current_induction_variables: HashMap<ValueId, (IntegerConstant, IntegerConstant)>,
     // Maps outer loop induction variable -> fixed lower and upper loop bound
     // This will be used by inner loops to determine whether they
-    // have safe operations reliant upon an outer loop's maximum induction variable.
+    // have safe operations reliant upon an outer loop's maximum induction variable
     outer_induction_variables: HashMap<ValueId, (IntegerConstant, IntegerConstant)>,
+    // All induction variables collected up front.
+    all_induction_variables: HashMap<ValueId, (IntegerConstant, IntegerConstant)>,
```
vezenovm, line 168: At this point could we get rid of the current and outer induction variables maps and just have an all encompassing induction variables map associated per loop no that we are collecting all variables up front?

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -123,9 +127,12 @@ impl Loop {
     ///     v5 = lt v1, u32 4           // Upper bound
     ///     jmpif v5 then: b3, else: b2
     /// ```
-    /// In the example above, `v1` is the induction variable
-    fn get_induction_variable(&self, function: &Function) -> ValueId {
-        function.dfg.block_parameters(self.header)[0]
+    /// In the example above, `v1` is the induction variable.
+    ///
+    /// There is an example in the tests where a loop does not have an induction variable,
+    /// but rather loads a reference in the header, in which case this will return `None`.
```
vezenovm, line 133: Which test is this?

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: I realize this may be a separate refactor though, but just curious if you looked at doing this.

vezenovm: Ah this looks like it may be a pain for the outer induction variables.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -370,6 +387,55 @@ impl<'f> LoopInvariantContext<'f> {
...
+            return false;
+        }
+
+        // If the block is part of any nested loop, they have to execute as well.
+        for nested in all_loops.iter() {
+            if !nested.blocks.contains(&block) {
+                continue;
+            }
+            let Some(induction_variable) = self.get_induction_variable(nested) else {
+                // If we don't know what the induction variable is, we can't say if it executes.
```
vezenovm, line 423: Can you point to when this happens?

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
aakoshh: I tried that originally: to get rid of `outer_induction_variables` and use the pre-calculated `all_induction_variables` instead. The unit tests started failing unfortunately, which I could only explain by concluding that the various reads from `outer_induction_variables` are not supposed to have access to that information until after the loop has been processed. 

Here's an example of a failed test:
```
Snapshot: do_not_simplify_constraint
Source: compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs:2322
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Expression: ssa_string
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
    3     3 │     store v0 at v4
    4     4 │     v6 = eq v1, u32 4
    5     5 │     jmp b1(u32 0)
    6     6 │   b1(v3: u32):
    7       │-    v9 = lt v3, u32 5
    8       │-    jmpif v9 then: b2, else: b3
          7 │+    jmpif u1 1 then: b2, else: b3
    9     8 │   b2():
   10     9 │     jmpif u1 1 then: b4, else: b5
   11    10 │   b3():
   12       │-    v19 = load v4 -> u32
   13       │-    v20 = lt v1, v19
   14       │-    constrain v20 == u1 1
         11 │+    v17 = load v4 -> u32
         12 │+    v18 = lt v1, v17
         13 │+    constrain v18 == u1 1
   15    14 │     return
   16    15 │   b4():
   17       │-    v11 = load v4 -> u32
   18       │-    v13 = add v11, u32 1
   19       │-    store v13 at v4
         16 │+    v9 = load v4 -> u32
         17 │+    v11 = add v9, u32 1
         18 │+    store v11 at v4
   20    19 │     jmp b5()
   21    20 │   b5():
   22       │-    v15 = lt u32 2, v3
   23       │-    v16 = unchecked_mul v6, v15
   24       │-    jmpif v16 then: b3, else: b6
         21 │+    v13 = lt u32 2, v3
         22 │+    v14 = unchecked_mul v6, v13
         23 │+    jmpif v14 then: b3, else: b6
   25    24 │   b6():
   26       │-    v17 = lt v3, u32 4
   27       │-    constrain v17 == u1 1
   28       │-    v18 = unchecked_add v3, u32 1
   29       │-    jmp b1(v18)
         25 │+    v15 = lt v3, u32 4
         26 │+    constrain v15 == u1 1
         27 │+    v16 = unchecked_add v3, u32 1
         28 │+    jmp b1(v16)
   30    29 │ }
────────────┴───────────────────────────────────────
```

It's probably easy to get rid of the `current_induction_variables` though.

vezenovm: We can just do it in a later PR, there isn't large regressions anywhere so this is fine as is imo.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
aakoshh: `do_not_hoist_non_control_dependent_div_in_non_executed_loop` has this SSA:

```rust
        brillig(inline) fn main f0 {
          b0(v0: u32):
            v1 = allocate -> &mut Field
            store Field 0 at v1
            jmp b1()
          b1():
            v3 = load v1 -> Field
            v4 = eq v3, Field 0
            jmpif v4 then: b2, else: b3
          b2():
            return
          b3():
            v6 = div u32 5, v0
            jmp b1()
        }
```

This does not cause an issue currently because for this loop we never query the induction variable. In the following code, `get_const_bounds` returns `None` because it calls the module level `get_induction_variable(function, pre_header).ok()?` under the hood, so we never get to `loop_.get_induction_variable` which would panic:
```rust
fn set_induction_var_bounds(&mut self, loop_: &Loop, current_loop: bool) {
        let bounds = loop_.get_const_bounds(self.inserter.function, self.pre_header());
        if let Some((lower_bound, upper_bound)) = bounds {
            let induction_variable = loop_.get_induction_variable(self.inserter.function);
```

guipublic:
>  in the case of the SSA in the program in the integration test, the Post Dominance Frontier was completely empty.

Do you mean that looking at control dependence is not enough or that in your case the post dominance frontier was not computed for some reason?

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
aakoshh: It's the `do_not_hoist_non_control_dependent_div_in_non_executed_loop` test.

aakoshh:
@guipublic the PDF was calculated, but it was completely empty.

I asked Maxim if this makes sense, and he kindly calculated it by hand:

> Just to write it out:
> * A node X post dominates Y if every path from Y to the exit goes through X
> * A PDF for a node X is the set of nodes S that post dominates a successor of X but do not strictly dominate X (e.g. A != B)

> The first node we would look at for the outer loop is b2

> I do actually think the post dominance frontiers are correct. If we draw out our post dominance tree I think we have the following:
```
b3
- - b1
-   - - b0
-   - - b6
-
- - b4  
    - - b2
    - - b5
```
> From this tree, the PDFs would be empty.

Here's the SSA for reference:
```rust
  b0():
    jmp b1(u128 119718805667097306623756199175472805749)
  b1(v0: u128):
    v4 = lt v0, u128 119718805667097306623756199175472805750	
    jmpif v4 then: b2, else: b3
  b2():
    v6 = mod v0, u128 9                           	
    jmp b4(u128 4)
  b3():
    return
  b4(v1: u128):
    v8 = lt v1, v6                               
    jmpif v8 then: b5, else: b6
  b5():
    v12 = lt u128 115517568315105269094017991206916906366, v0	
    constrain v12 == u1 0, "UHL"               
    v14 = unchecked_add v1, u128 1      
    jmp b4(v14)
  b6():
    v10 = unchecked_add v0, u128 1         
    jmp b1(v10)
}
```

guipublic:
An alternative could be to hoist only what is at the current loop-level. This would be simpler, but it also requires to process the loops starting from the inner-most loop, which is not the current behaviour (so it is not simpler in fact).

PR #9234, opened 2025-07-17
feat(ssa): Mark instructions after OOB array op as unreachable

vezenovm:
# Description

## Problem\*

Builds off of https://github.com/noir-lang/noir/pull/9232 to resolve https://github.com/noir-lang/noir/issues/9217

## Documentation\*

Check one:
- [X] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [X] I have tested the changes locally.
- [X] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #9252, opened 2025-07-18
feat(ssa): Transform OOB array op to dummy array when marking an unreachable block

vezenovm:
# Description

## Problem\*

Resolves bugs from https://github.com/noir-lang/noir/pull/9234 such as with regression_8994 when we fail to resolve arrays.

After #9234 we have the following SSA:
```
          b0():
            v0 = allocate -> &mut u8                      
            store u8 0 at v0                               
            v2 = make_array [u8 0, v0] : [(u8, &mut u8); 1]
            v4 = array_get v2, index u32 2 -> u8         
            v6 = array_get v2, index u32 3 -> &mut u8      
            return v4
        }
```
That will ultimately be sent to ACIR gen like such:
```
          b0():
            v0 = allocate -> &mut u1                            
            v2 = make_array [u1 0, v0] : [(u1, &mut u1); 1]     
            v4 = array_get v2, index u32 2 -> u1     
            unreachable
        }
```
We want the array get to trigger a failure, but we have no way to remove that allocation as we never load the reference.

## Summary\*

We transform failing array OOB accesses to a dummy array OOB access. We then expect follw up passes such as DIE to remove the allocations. 

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
> Mildly gross but LGTM

Yeah this is not ideal. A full solution would most likely require handling references in ACIR. I can gate this logic to just be for ACIR. Perhaps that would be best for a follow-up once we get these changes merged.

Issue #9248, opened 2025-07-18
Noir fails to resolve implementation of a trait for an array with generics


benesjan:
### Aim

...

### Expected Behavior

...

### Bug

The following code:

```rust
pub trait Deserialize {
    let N: u32;

    fn deserialize(fields: [Field; N]) -> Self;
}

impl Deserialize for Field {
    let N: u32 = 1;

    fn deserialize(fields: [Field; Self::N]) -> Self {
        fields[0]
    }
}

impl<T, let M: u32> Deserialize for [T; M]
where
    T: Deserialize,
{
    let N: u32 = <T as Deserialize>::N * M;

    fn deserialize(fields: [Field; Self::N]) -> Self {
        let mut result: [T; M] = std::mem::zeroed();

        for i in 0..M {
            let mut element_fields = [0; <T as Deserialize>::N];
            for j in 0..<T as Deserialize>::N {
                element_fields[j] = fields[i * <T as Deserialize>::N + j];
            }
            result[i] = T::deserialize(element_fields);
        }

        result
    }
}

pub struct Log<let M: u32> {
    pub fields: [Field; M],
}

impl<let M: u32> Deserialize for Log<M> {
    let N: u32 = M;

    fn deserialize(serialized: [Field; Self::N]) -> Self {
        let fields = <[Field; M] as Deserialize>::deserialize(serialized);
        Self { fields }
    }
}
```

fails to compile with:

```
error: No matching impl found for `[Field; M]: Deserialize<N = M>`
   ┌─ src/main.nr:44:23
   │
44 │         let fields = <[Field; M] as Deserialize>::deserialize(serialized);
   │                       ------------------------- No impl for `[Field; M]: Deserialize<N = M>`
   │

Aborting due to 1 previous error
```

If we make the implementation of Deserialize specific to `[Field; M]` instead of `[T; M]`:

```rust
impl<let M: u32> Deserialize for [Field; M] {
    let N: u32 = <Field as Deserialize>::N * M;

    fn deserialize(fields: [Field; Self::N]) -> Self {
        let mut result: [Field; M] = std::mem::zeroed();
        for i in 0..M {
            let mut element_fields = [0; <Field as Deserialize>::N];
            for j in 0..<Field as Deserialize>::N {
                element_fields[j] = fields[i * <Field as Deserialize>::N + j];
            }
            result[i] = Field::deserialize(element_fields);
        }
        result
    }
}
```

The code compiles.

### To Reproduce

1.
2.
3.
[...]

PR #9255, opened 2025-07-18
feat: simplify adding a type variable to itself

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR adds some handling for @asterite's repro in #9248 however it doesn't solve the full issue.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Interesting that this started causing failures in other tests.

jfecher (commented):
I don't know if we always do want to change `N + N` into `2 * N`. For one, multiplication can be more difficult to simplify out due to its inverse being integer division. For another, if we have `N + N + N` we would end up changing it to `2*N + N` currently instead of `3 * N`.

TomAFrench:
Good point on `3 * N`. I don't think that we want to merge this as is, I was mostly experimenting around the behaviour Ary highlighted in https://github.com/noir-lang/noir/issues/9248#issuecomment-3090130116.

PR #9258, opened 2025-07-21
fix: parse AsTraitPath in type expressions

### LINKED ISSUE #5794: Allow AsTraitPaths in arithmetic generics
### Problem

As mentioned in https://github.com/noir-lang/noir/pull/5739, AsTraitPaths (`<T as Trait>::Type`) currently aren't allowed in arithmetic generics.

### Happy Case

We should allow these in arithmetic generics to allow for code like:
```rs
impl<A, B> Serialize for (A, B) where A: Serialize, B: Serialize {
    let Size = <A as Serialize>::Size + <B as Serialize>::Size;
    ...
}
```
Once we allow for implicit associated types as well.

### Workaround

These generics can be made explicit instead:
```rs
impl<A, B, let AS: u32, let BS: u32> Serialize for (A, B) where A: Serialize<Size = AS>, B: Serialize<Size = BS> {
    let Size = AS + BS;
    ...
}
```

### Blocker Context

Blocks @benesjan from finishing [this PR](https://github.com/AztecProtocol/aztec-packages/pull/15834)

asterite:
# Description

## Problem

Resolves #5794

## Summary

I don't know if I missed this while porting the old parser.

That said, this currently doesn't work if `AsTraitPath` is used in a function signature. That is, this doesn't compile:

```noir
use std::mem::zeroed;

trait Trait {
    let N: u32;
}

impl Trait for Field {
    let N: u32 = 10;
}

fn main() {}

fn foo(x: [Field; <Field as Trait>::N]) {} // No matching impl found for `Field: Trait<N = _>`
```

I think the reason is that function signatures are checked before impls are processed, not sure. In any case this PR at least makes the use case in the linked issue work.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/parser/parser/type_expression.rs
+++ b/compiler/noirc_frontend/src/parser/parser/type_expression.rs
@@ -143,6 +143,7 @@ impl Parser<'_> {
     /// AtomTypeExpression
     ///     = ConstantTypeExpression
     ///     | VariableTypeExpression
+    ///     | AsTraitPathTypeExpression
     ///     | ParenthesizedTypeExpression
     fn parse_atom_type_expression(&mut self) -> Option<UnresolvedTypeExpression> {
         if let Some(type_expr) = self.parse_constant_type_expression() {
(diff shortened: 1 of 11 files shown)


TomAFrench (approved):
LGTM

PR #5810, opened 2024-08-23
fix(acir_gen): Nested dynamic array initialization

vezenovm:
# Description

## Problem\*

Resolves #5782 

## Summary\*

We are using our recursive `initialize_array_inner` function to also check whether the outer AcirValue we are trying to initialize is a dynamic array. We should be able to initialize a new AcirValue::Array with internal AcirValue::DynamicArray types. 

## Documentation\*

Check one:
- [X] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [X] I have tested the changes locally.
- [X] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #9259, opened 2025-07-21
fix(acir_gen): Bail out of `handle_constant_index` when it encounters `DynamicArray`

### LINKED ISSUE #9207: Compiler crash: `Dynamic array should already be initialized`
The Rust tests for the Noir compiler are currently failing on ARM64.

Check the [Test (ARM64)](https://github.com/noir-lang/noir/actions/runs/16288994263) workflow for details.

This issue was raised by the workflow `Test (ARM64)`

To reproduce:
```shell
NOIR_AST_FUZZER_SEED=0xf8a978a600100000 cargo test -p noir_ast_fuzzer --test smoke
```

Version:
```console
❯ git rev-parse master
1719bc3a7977248e792816df3a3e907f2d006a95
```

Here's a more minimal version of the code generated by the fuzzer:
```rust
fn main(a: [[bool; 2]; 3], mut b: [[bool; 2]; 3]) -> pub bool {
    for _ in 0..2 {
        let idx_e: u32 = (if (0 != (a[2][0] as Field)) { 0 } else { 1 });

        b[2][idx_e] = (!a[0][1]);

        for idx_g in 65_u8..70_u8 {
            let idx_j = ((idx_g as u32) % 2);
            let idx_k = ((idx_g as u32) % 3);
            b[1][idx_j] = a[idx_k][1];
        }
    }
    b[0][0]
}
```

Failing with:
```console
❯ cargo run -q -p nargo_cli -- compile --force --silence-warnings
Message:  internal error: entered unreachable code: Dynamic array should already be initialized
Location: compiler/noirc_evaluator/src/acir/acir_context/mod.rs:1559
```

aakoshh:
# Description

## Problem\*

Resolves #9207 

## Summary\*

Returns `false` from `handle_constant_index` if the value it would return is a `DynamicArray`, so that we allow such values to go through the regular array operation steps, which do take care of initialising the array member values.

## Additional Context

In the new integration test, we had a nested array, where we retrieve a nested array item, update it, then write it back into its wrapper:
```
    v30 = array_set mut v28, index u32 1, value v29	// src/main.nr:10:13
    v31 = array_get v30, index u32 2 -> [u1; 2]   	// src/main.nr:5:9
    v32 = array_set mut v31, index v13, value v12 	// src/main.nr:5:9
    v33 = array_set mut v30, index u32 2, value v32	// src/main.nr:5:9
```

The current program had this ACIR value for `v30`:
```
v30 = Array([
  Array([Var(AcirVar(13), ...), Var(AcirVar(15), ...)]), 
  Array([Var(AcirVar(11), ...), Var(AcirVar(3), ...]), 
  DynamicArray(id: 2, len: 2, element_type_sizes: None)
])
```

Because the index was a constant, `array_get v30, index u32 2 -> [u1; 2] ` returned the `DynamicArray` as-is for `v31`, which then failed in the `array_set mut v31` step, as it wasn't initialised. This is because we bypassed [these steps](https://github.com/noir-lang/noir/blob/566de3f20199259e1594d778e47ad3550aa344cd/compiler/noirc_evaluator/src/acir/arrays.rs#L473-L485), which would have read those values index by index.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/acir/arrays.rs
```diff
@@ -241,6 +243,13 @@ impl Context<'_> {
             // as if the predicate were true. This is as if the predicate were to resolve to false then
             // the result should not affect the rest of circuit execution.
             let value = array[index].clone();
+            // An `Array` might contain a `DynamicArray`, however if we define the result this way,
+            // we would bypass the array initialization and the handling of dynamic values that
+            // happens in `array_get_value`. Rather than repeat it here, let the non-special-case
+            // handling take over by returning `false`.
+            if matches!(value, AcirValue::DynamicArray(_)) {
+                return Ok(false);
```
vezenovm, line 251: Why not just initialize it here? The logic feels easier to follow then and less scattered.

vezenovm (commented):
I'm wondering if we can just get rid of the panic added here https://github.com/noir-lang/noir/pull/5810/files? Did you consider this? It seems to have just been an overly conservative assumption.

compiler/noirc_evaluator/src/acir/arrays.rs
aakoshh: I felt the opposite: by returning a flag it felt like it was looking for a quick win without trying too hard. It flat out returns false for top level DynamicArray for example. I thought if we add another place where arrays can be initialised it increases the scattering of logic.

That said I would be in support for revisiting this part of the code because it’s a bit sparsely documented.

vezenovm: > I thought if where arrays can be initialised it adds to the scattering of logic.

Yes, that is a fair point. We can leave as is for now. This code could use revisiting in general to streamline the logic.

aakoshh:
> Did you consider this? It seems to have just been an overly conservative assumption.

I always start assuming that it’s there for a reason, which is in this case that array_get should have taken care of initializing the array, so it’s right to panic in a way. 

OTOH I do find it a bit confusing that it initializes DynamicArrays inside Array, but not if they are on the top, so I’m in support of refactoring, as a follow up PR

PR #9261, opened 2025-07-21
lets see the effect

vezenovm:
# Description

## Problem\*

I want to see the effect of removing the panic in `initialize_array` as https://github.com/noir-lang/noir/pull/9259 looks to have led to some timing regressions. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/acir/acir_context/mod.rs
+++ b/compiler/noirc_evaluator/src/acir/acir_context/mod.rs
@@ -1555,9 +1555,6 @@ impl<F: AcirField, B: BlackBoxFunctionSolver<F>> AcirContext<F, B> {
             }
             Some(optional_value) => {
                 let mut values = Vec::new();
-                if let AcirValue::DynamicArray(_) = optional_value {
-                    unreachable!("Dynamic array should already be initialized");
-                }
                 self.initialize_array_inner(&mut values, optional_value)?;
                 values
             }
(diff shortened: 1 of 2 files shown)


TomAFrench:
I'm not sure that we want to have an explicit call to `initialize_array` with a `DynamicArray` tbf.

`initialize_array` is responsible for laying down the `MemInit` opcode but if we have an initialized `DynamicArray` then that should only occur in the situation where we've done a read/write to an array at a dynamic index (as otherwise we'll be performing the read/write at compile time and the array will stay as an `Array` type).

The only situation where we need to initialize an array from a `DynamicArray` is when we're doing a non-mutable dynamic write into it. In this situation we need to do a complete copy of the array by reading every element out of it and then intializing it from the read elements. But by doing these array reads we've constructed a regular non-dynamic `Array`!

TomAFrench:
hmm, true. Looked back at this and the handling of internal dynamic arrays is odd in comparison.

vezenovm:
> I'm not sure that we want to have an explicit call to `initialize_array` with a `DynamicArray` tbf.

Yeah I agree it is odd, and it does seem indicative of something else being off. 

> initialized `DynamicArray` then that should only occur in the situation where we've done a read/write to an array at a dynamic index (as otherwise we'll be performing the read/write at compile time and the array will stay as an `Array` type).

We will also have a dynamic array when operating over arrays that come from block parameters, even with compile time indices. This is what happens in the `regression_9207` case. After looking at the bug closer I am good with the solution in #9259. It looks like the cleanest solution aside streamlining our array ops more generally. I'm going to close this PR.

PR #9263, opened 2025-07-21
chore(optimization): Experiment with changing ordering of node ids

jfecher:
# Description

## Summary\*

Experiment with finding any possible compiler performance improvement with reordering how we assign node ids.

## Additional Context

The theory: we should have fewer cache misses if the node ordering in our node arena: `[expr1, expr2, expr3, ...]` generally follows the order we perform recursion on nodes.

Currently we allocate and push a node when we finish creating it which leads to a post ordering: `[child_node1, child_node2, parent_node]`. This PR reserves the space for parent nodes before their children to reverse this.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/expressions.rs
+++ b/compiler/noirc_frontend/src/elaborator/expressions.rs
@@ -55,48 +55,76 @@ impl Elaborator<'_> {
         expr: Expression,
         target_type: Option<&Type>,
     ) -> (ExprId, Type) {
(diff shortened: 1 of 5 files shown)


jfecher:
Pushed a change to address the memory usage increase but from some basic testing it looks like this may be slightly slower than master

jfecher:
This PR is currently slightly slower and I think it is because we retrieve expressions with `interner.expression(id)` which copies the expression, and thus takes it out of the same cache line.

I'll update this PR so that that returns a reference instead but then I also need to compare that not against master but against master with the same reference change but not the ID reordering.

PR #9264, opened 2025-07-21
Remove useless use of cat in plot-bytecode-size.sh

Galoretka:
Replaced the use of cat $IN | jq ... with the more idiomatic jq ... \"$IN\" in the plot-bytecode-size.sh script.
Using cat to pipe a file into jq is considered a "useless use of cat" anti-pattern, since jq can read the file directly. This change improves script readability and efficiency.


TomAFrench:
This works for our purposes as this code is used for adhoc purposes by the noir team internally.

PR #9265, opened 2025-07-21
fix: solve object type and where clause before args when looking up trait impl

### LINKED ISSUE #9245: Trait implementation fails to get resolved
### Aim

...

### Expected Behavior

...

### Bug

The code in https://github.com/benesjan/noir_slow.git fails to compile with the following error:

Image

### To Reproduce

1. git clone https://github.com/benesjan/noir_slow.git
2. cd noir_slow
3. nargo compile

### Workaround

None

### Workaround Description

As Tom mentioned in [this discussion](https://aztecprotocol.slack.com/archives/C04QF64EDNV/p1752830606411429) on slack this can be worked around by calling `Deserialize::deserialize(fields)` instead of `<[Field; O]>::deserialize(fields)`.

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.8 noirc version = 1.0.0-beta.8+0000000000000000000000000000000000000000

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9245

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/node_interner.rs
+++ b/compiler/noirc_frontend/src/node_interner.rs
@@ -1648,6 +1648,28 @@ impl NodeInterner {
 
             let mut fresh_bindings = type_bindings.clone();
 
+            if !object_type.try_unify(&existing_object_type, &mut fresh_bindings).is_ok() {
+                continue;
+            }
+
+            if let TraitImplKind::Normal(impl_id) = impl_kind {
+                let trait_impl = self.get_trait_implementation(*impl_id);
+                let trait_impl = trait_impl.borrow();
+
+                if let Err(error) = self.validate_where_clause(
+                    &trait_impl.where_clause,
+                    &mut fresh_bindings,
+                    &instantiation_bindings,
+                    recursion_limit,
+                ) {
+                    // Only keep the first errors we get from a failing where clause
+                    if where_clause_error.is_none() {
+                        where_clause_error = Some(error);
+                    }
+                    continue;
+                }
+            }
+
             let mut check_trait_generics =
                 |impl_generics: &[Type], impl_associated_types: &[NamedType]| {
                     let generics_unify = trait_generics.iter().zip(impl_generics).all(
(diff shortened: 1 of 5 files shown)


jfecher (changes requested):
I think unifying the object type before checking generics is cleaner (and can't remember why it was reversed to begin with but I do think it was done on purpose - maybe just a bad quick fix). So I think that is good but also checking the where clause earlier seems arbitrary to me. You mentioned this is to bind it against a type variable before the same variable gets overwritten with another incorrect type so it sounds like to me the correct solution would be figuring out why this other type variable is being (permanently? temporarily?) bound to an incorrect type.

asterite:
@jfecher I asked for a review but I'm still not sure this is the right way to fix this. For example #9248 still fails with this change. Like I commented on the other issue, `<T as Deserialize>::N` doesn't get the constant value it should have, and maybe it gets that at some point in `validate_where_clause` but I don't know if calling that method is the way to bind `T` to `Field` (in the example code).

asterite:
@jfecher This is ready for review again, and changes now are more likely to be correct (checking bindings in `canonicalize` did help 🎉 )

compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
```diff
@@ -63,72 +67,113 @@ impl Type {
...
+                }
+
+                // See if this is `X + 0` or `X - 0`, in which case we can simplify it to `X`
+                if matches!(op, BinaryTypeOperator::Addition | BinaryTypeOperator::Subtraction) {
+                    if let Ok(rhs_value) = evaluate(rhs) {
+                        if rhs_value.is_zero() {
+                            return *lhs.clone();
+                        }
+                    }
+                }
```
asterite, line 122: I don't know if the two new rules above should happen here or somewhere else.

jfecher:
Looks like the failing proptest is asserting on exact equality of the types which is different now. I think we can instead change it to assert that the types are equal after canonicalizing them (again)

compiler/noirc_frontend/src/hir_def/types/arithmetic.rs
jfecher: Missed this part in the review - can you move these checks after the `let lhs = lhs.canonicalize_helper(...);` calls?

Currently if rhs is one or zero we're returning `lhs` but we're not canonicalizing it recursively in that case so the expression as a whole may not be in canonical form.

PR #9275, opened 2025-07-22
feat: simply cfg immediately before flattening

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR adds another simplification pass immediately before cfg flattening. I've also added a pre-check to flattening which asserts that there are no constant condition jmpifs.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -172,12 +172,46 @@ impl Ssa {
...
+        {
+            if function
+                .dfg
+                .get_numeric_constant(*condition)
+                .is_some_and(|c| c.is_zero() || c.is_one())
+            {
+                // We have this check here as if we do not, the flattening pass will generate groups of opcodes which are
+                // under a zero predicate. These opcodes are not always removed by the die pass and so bloat the code.
+                //
+                // We could add handling for this inside of the pass itself but it's simpler to just run `simplfy_cfg`
```
vezenovm:
```suggestion
                // We could add handling for this inside of the pass itself but it's simpler to just run `simpilfy_cfg`
```

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -172,12 +172,46 @@ impl Ssa {
...
+                // immmediately before flattening.
```
vezenovm:
```suggestion
                // immediately before flattening.
```

vezenovm (approved):
I didn't see much benefit for the `no_predicates_numeric_generic_poseidon` test you had mentioned in https://github.com/noir-lang/noir/pull/9232 and it looks like the reports confirm that. However, I am good with this change as it still provides benefits and simplify CFG is a more lightweight pass.

TomAFrench:
I'm not seeing a change in `no_predicates_numeric_generic_poseidon` so it seems like this is a result of using `no_predicates`.

TomAFrench:
Ok, the issue is that we have control flow in the `no_predicates` function which depends on the arguments to the function. We perform flattening in this function with the understanding the both branches are reachable, however once we inline this into the calling function we end up discovering that we'll never go down one of these branches.

This then results in us having a `enable_side_effects u1 0` popping up, except now it's happening later on in the compilation pipeline.

PR #9276, opened 2025-07-22
feat: add better handling for follow-on optimizations in `simplify_cfg`

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR modifies the `simplify_cfg` pass so it's more likely to catch further simplifications which are uncovered. This is done by pushing the current block back onto the stack if we've successfully simplified it.

I've also added an explicit check for unreachable blocks which may exist on the stack (e.g. an orphaned block from a constant condition jmpif) so we can ensure that its successors are removed and we don't consider it any further.
## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/simplify_cfg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/simplify_cfg.rs
@@ -95,7 +95,7 @@ impl Function {
                 // optimizations performed after this point on the same block should check if
                 // the inlining here was successful before continuing.
                 if try_inline_into_predecessor(self, &mut cfg, block, predecessor) {
-                    stack.push(block);
+                    stack.push(predecessor);
                 };
             } else {
                 drop(predecessors);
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/simplify_cfg.rs
```diff
@@ -771,4 +805,69 @@ mod test {
...
+        }
+        "#;
+
+        let ssa = Ssa::from_str(src).unwrap();
+        let ssa = ssa.simplify_cfg();
+
+        assert_ssa_snapshot!(ssa, @r"
+        brillig(inline) impure fn main f0 {
+          b0():
+            jmp b1()
```
vezenovm, line 867: I guess this also could be further simplified to just a return in b0

vezenovm (approved):
Looks good

compiler/noirc_evaluator/src/ssa/opt/simplify_cfg.rs
TomAFrench: Yep, I'd like to address this at some point but we want to make sure that we don't end up creating multiple return blocks.

PR #9279, opened 2025-07-22
chore(LSP): simplify code lens request handling

asterite:
# Description

## Problem

## Summary

While I was recently looking at some LSP code I noticed that whenever we type-check a package we compute code lenses for the changed file, then when code lenses are requested from the client we'd try to use that cache or compute them again. This code exists since before I started making changes to LSP and now there's no need to compute code lenses when type-checking, nor there's a need to cache that information. This makes that request work more like every other LSP request handler and simplifies some code.

## Additional Context

I checked that after this change the time it took to respond to a code lens request was still fast (for example it's less than a millisecond in `token_contract`).

I also changed a few `BTreeMap<CrateId, CrateDefMap>` to the `DefMaps` alias, and had to some some small refactors to use some functions without having a `Context`, when all that was used was a `NodeInterner` plus `DefMaps`.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/mod.rs
+++ b/compiler/noirc_frontend/src/hir/mod.rs
@@ -9,6 +9,7 @@ use crate::ast::UnresolvedGenerics;
 use crate::debug::DebugInstrumenter;
 use crate::elaborator::UnstableFeature;
 use crate::graph::{CrateGraph, CrateId};
+use crate::hir::def_map::DefMaps;
 use crate::hir_def::function::FuncMeta;
 use crate::node_interner::{FuncId, NodeInterner, TypeId};
 use crate::parser::ParserError;
@@ -141,17 +142,7 @@ impl Context<'_, '_> {
     }
 
     pub fn fully_qualified_function_name(&self, crate_id: &CrateId, id: &FuncId) -> String {
-        let def_map = self.def_map(crate_id).expect("The local crate should be analyzed already");
-
-        let name = self.def_interner.function_name(id);
-
-        let module_id = self.def_interner.function_module(*id);
-        let module = self.module(module_id);
-
-        let parent =
-            def_map.get_module_path_with_separator(module_id.local_id, module.parent, "::");
-
-        if parent.is_empty() { name.into() } else { format!("{parent}::{name}") }
+        fully_qualified_function_name(*crate_id, *id, &self.def_interner, &self.def_maps)
     }
 
     /// Returns a fully-qualified path to the given [TypeId] from the given [CrateId]. This function also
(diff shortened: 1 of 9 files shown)


compiler/noirc_frontend/src/hir/mod.rs
```diff
@@ -303,3 +279,49 @@ impl Context<'_, '_> {
...
+                FunctionNameMatch::Contains(patterns) => patterns
+                    .iter()
+                    .any(|pattern| fully_qualified_name.contains(pattern))
+                    .then_some((fully_qualified_name, test_function)),
+            }
+        })
+        .collect()
+}
+
+pub fn fully_qualified_function_name(
```
TomAFrench:
```suggestion
fn fully_qualified_function_name(
```

compiler/noirc_frontend/src/hir/mod.rs
```diff
@@ -303,3 +279,49 @@ impl Context<'_, '_> {
         self.interpreter_output = Some(output);
     }
 }
+
+pub fn get_all_test_functions_in_crate_matching(
```
TomAFrench, line 283:
```suggestion
fn get_all_test_functions_in_crate_matching(
```

tooling/lsp/src/requests/code_lens_request.rs
```diff
@@ -50,192 +62,180 @@ fn on_code_lens_request_inner(
...
+    package: &Package,
+    range: async_lsp::lsp_types::Range,
+) -> CodeLens {
+    let compile_command = Command {
+        title: with_arrow(COMPILE_CODELENS_TITLE),
+        command: COMPILE_COMMAND.into(),
+        arguments: Some(package_selection_args(workspace, package)),
+    };
+    CodeLens { range, command: Some(compile_command), data: None }
+}
```
TomAFrench: duplicate

tooling/lsp/src/requests/code_lens_request.rs
```diff
@@ -50,192 +62,180 @@ fn on_code_lens_request_inner(
...
+            if file_id == current_file {
+                let range =
+                    byte_span_to_range(files, file_id, location.span.into()).unwrap_or_default();
+                lenses.push(compile_code_lens(workspace, package, range));
+
+                let internal_command_lenses = [
+                    (INFO_CODELENS_TITLE, INFO_COMMAND),
+                    (EXECUTE_CODELENS_TITLE, EXECUTE_COMMAND),
+                    (DEBUG_CODELENS_TITLE, DEBUG_COMMAND),
+                ];
```
TomAFrench: Looks like you were meaning to replace these?

tooling/lsp/src/requests/code_lens_request.rs
asterite: Thanks! I had a feeling I duplicated this function but Rust didn't complain... I didn't realize it was because I chose two different names 😅

compiler/noirc_frontend/src/hir/mod.rs
asterite: Thanks, done!

compiler/noirc_frontend/src/hir/mod.rs
asterite: This one is used by the LSP crate so I can't change it.

tooling/lsp/src/requests/code_lens_request.rs
asterite: I wasn't sure but I just did :-)

TomAFrench (approved):

PR #9286, opened 2025-07-23
fix: throw error if foreign call returns the wrong number of fields

TomAFrench:
# Description

## Problem\*

Resolves https://aztecprotocol.slack.com/archives/C04QF64EDNV/p1753216370658889

## Summary\*

This PR adds more validation on foreign call return results to throw an error if we get a differing number of values from the user.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/acvm-repo/brillig/src/opcodes.rs
+++ b/acvm-repo/brillig/src/opcodes.rs
@@ -84,6 +84,18 @@ impl HeapValueType {
     pub fn field() -> HeapValueType {
         HeapValueType::Simple(BitSize::Field)
     }
+
+    pub fn flattened_size(&self) -> usize {
+        match self {
+            HeapValueType::Simple(_) => 1,
+            HeapValueType::Array { value_types, size } => {
+                value_types.iter().map(|t| t.flattened_size()).sum::<usize>() * size
+            }
+            HeapValueType::Vector { value_types } => {
+                value_types.iter().map(|t| t.flattened_size()).sum()
+            }
+        }
+    }
 }
 
 /// A fixed-sized array starting from a Brillig memory location.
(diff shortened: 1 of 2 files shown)


Savio-Sou:
Breaking change (i.e. needs an `!`)?

test_programs/noir_test_success/regression_4561/src/main.nr
```diff
@@ -60,13 +60,20 @@ struct TestTypeFoo {
...
+        TestTypeFoo {
+            a: 7,
+            b: [
+                [[1, 2, 3], [4, 5, 6], [7, 8, 9], [10, 11, 12]],
+                [[13, 14, 15], [16, 17, 18], [19, 20, 21], [22, 23, 24]],
+            ],
+            c: [[1, 2, 3], [4, 5, 6]],
+            d: [1, 2, 3],
+        },
     ));
```
TomAFrench, line 75: This mocked result was misshapen so I'm fixing this here.

acvm-repo/brillig/src/opcodes.rs
```diff
@@ -84,6 +84,18 @@ impl HeapValueType {
...
     }
+
+    pub fn flattened_size(&self) -> usize {
+        match self {
+            HeapValueType::Simple(_) => 1,
+            HeapValueType::Array { value_types, size } => {
+                value_types.iter().map(|t| t.flattened_size()).sum::<usize>() * size
+            }
+            HeapValueType::Vector { value_types } => {
+                value_types.iter().map(|t| t.flattened_size()).sum()
```
asterite: Is this valid for vectors/slices? The length is unknown so I'm not sure how this can work. Maybe you can add a test with vectors to see it working? That said, I don't know if slices are allowed in foreign calls.

TomAFrench: Yeah, this is incorrect for vectors. I've updated this to return `None` in this case.

I've also added extra handling to ensure that the length of the returned data for vectors are constrained to be a multiple of the element size.

PR #9287, opened 2025-07-23
fix: Special handling for slices in printing

### LINKED ISSUE #9271: ACIR != Brillig: Printing slice arg with varying length garbled in ACIR
### Aim

Replicates a nightly failure from https://github.com/noir-lang/noir/actions/runs/16433887822/job/46440148153

The following program prints different values for `a` in ACIR and Brillig:

```rust
fn main(b: bool) -> pub bool {
    unsafe {
        func_5((if b { &[0, 0] } else { &[0, 0, 0, 0] }, true, &[100]))
    }
}
fn func_5(a: ([u1], bool, [u32])) -> bool {
    println(a);
    a.1
}
```

Input:
```toml
b = true
```

### Expected Behavior

They should both print the same thing.

### Bug

```console
❯ cargo run -q -p nargo_cli -- execute --force --silence-warnings; cargo run -q -p nargo_cli -- execute --force --silence-warnings --force-brillig
(&[0, 0], false, &[])
[fuzz_testing] Circuit output: Field(1)
(&[0, 0], true, &[100])
[fuzz_testing] Circuit output: Field(1)
```

Note that despite printing `false` and `true`, the returned value of `a.1` is `true` for both ACIR and Brillig.

The problem goes away if we make the two branches of `if b` have equal length, so it might be related to how slice length are passed around, perhaps how they are passed to print.

### To Reproduce

```
NOIR_AST_FUZZER_SEED=0x63f407cc00100000 cargo test -q -p noir_ast_fuzzer_fuzz acir_vs_brillig
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.8 noirc version = 1.0.0-beta.8+eb0982ee8859640b18a1874b1d92b29d0e2c0499 (git version hash: eb0982ee8859640b18a1874b1d92b29d0e2c0499, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9271 

## Summary\*

Adds special flattening to support printing slices where the capacity is different than the length. With this handling slices and arrays in foreign call parameters are prefixed by their length, which the decoder expects and uses to consume the extra items, before allowing the next parser to consume the rest of the data. 

There are some caveats with this:
* Revert data contains already flattened data, so we can't do this trick, however at the moment it cannot contain slices, so that's okay.
* There is only one prefix per parameter, so if it had nested data structures, those would not be prefixed, however nested slices are not allowed, so that's okay.
* We prefix both arrays and slices, even though this is only needed for slices, and arrays _can_ be nested, so we have to take care not to expect the prefix for them. It would be a bit tricky to only do this for slices in the foreign call handler, because a Struct could have a Slice field among others, and would be passed as multiple parameters, but just 1 corresponding type descriptor, so it's not trivial to tell from M values and N types which value needs to be prefixed. 
* Debug values won't be prefixed, but they are separate variables, so they shouldn't need them.
* Return values from foreign calls are not prefixed, but we don't expect slices to be returned.

Because of these reason I'm considering changing `brillig_vm::VM::get_memory_values` to prefix `HeapVector` specifically, so we don't need special casing anywhere, and can expect any `Slice` that appears anywhere in `ForeignCallParam` to have a prefix, and nothing else.

## Documentation\*

Check one:

# PR Checklist\*


aakoshh:
Closing in favour of https://github.com/noir-lang/noir/pull/9321

PR #9288, opened 2025-07-23
fix!: Prefix `HeapVector` with its `size` in `ForeignCallParams`

aakoshh:
# Description

## Problem\*

https://github.com/noir-lang/noir/pull/9287 tries to handle slices only in the printing process, which is a bit clunky and full of caveats. This is an alternative approach which prefixes slices when the Brillig VM reads them from memory.

## Summary\*

Changes `brillig_vm::VM::get_memory_values` to prefix `HeapVector` data by its size, which should be the _capacity_ of slices. By doing it in `get_memory_values`, before the values are converted to `ForeignCallParam::Array`, we can easily limit this handling to just slices, but not `HeapArray`, where the length is known, and then nested types are not a problem during decoding.

This is a breaking change because any foreign call handler decoding slices would need to expect 1 extra field in the data.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_printable_type/src/lib.rs
+++ b/compiler/noirc_printable_type/src/lib.rs
@@ -316,14 +316,10 @@ pub fn decode_printable_value<F: AcirField>(
                 .expect("expected slice to have at least 1 field for capacity")
                 .to_u128() as usize;
 
-            // Capacity is in terms of flattened members.
-            // For example a `[(u128, bool, [u32; 2])]` slice with 2 items is represented
-            // as an Array with 8 fields, but it has a capacity of 6, because the array
-            // only takes 1 slot. It's the tuple that has 3 members, and we have 2 tuples.
-            // To know how many items we need to consume, we need to know the number of
-            // members in the type that is contained in the slice.
-            let capacity = match printable_type_length(typ) {
-                0 => flattened_capacity,
+            // Capacity is in terms of flattened members. To figure out how many items are
+            // in the slice, we have to divide that with how wide the type is.
+            let capacity = match printable_type_width(typ) {
+                0 => length,
                 n => flattened_capacity / n,
             };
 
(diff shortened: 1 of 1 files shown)


aakoshh:
Okay, so I misunderstood the `size` in `get_memory_values`. Initially I thought it's the capacity, but then it turns out it's a bit more involved, because for a slice of tuples, it's going to be capacity*item_size, e.g. `[(1,2), (3,4)]` has a size of 4. Then I wanted implement `flattened_size` for the `PrintableType`, and divide to get the number of items I need to consume, however I didn't know what to do for `Enum` where the number of fields depends on the tag at runtime, and it turns out it crashes if I want to print it in https://github.com/noir-lang/noir/issues/9294

Then for some reason I thought the `size` would correspond to the number of _fields_ in the data (or actually inherited that approach from https://github.com/noir-lang/noir/pull/9287), but that's not true: `[[1,2], [3,4]]` has a size of 2, ie. 2 arrays in the slice, and `[(0, true, [1, 2]), (3, false, [5, 6]),]` has a size of 6, despite all 8 fields being in the slice: the arrays only contribute once.

To formula I came up with is to derive the number of items in the slice involves calculating a type-dependent flattened size, where tuples are structs are flattened, but array length and content is treated as 1. We use this length to divide the capacity inserted in the VM, which gives us the number of items we need to consume in total to clear the slice from the data.

compiler/noirc_printable_type/src/lib.rs
```diff
@@ -376,47 +363,28 @@ where
...
+            let end = {
                 let capacity = field_iterator
                     .next()
-                    .expect("expected length prefixed slice to have at least 1 element for the capacity")
+                    .expect("expected slice to have at least 1 field for capacity")
                     .to_u128() as usize;
-                Some(field_iterator.consumed() + capacity)
-            } else {
-                None
+                field_iterator.consumed() + capacity
```
vezenovm: We can remove the block `{ ... }` this expression is contained within right?

compiler/noirc_printable_type/src/lib.rs
```diff
@@ -374,59 +310,50 @@ where
...
-            // Number of fields in the slice. Items consume varying number of fields.
-            let end = if with_length_prefix {
-                let capacity = field_iterator
-                    .next()
-                    .expect("expected length prefixed slice to have at least 1 element for the capacity")
-                    .to_u128() as usize;
-                Some(field_iterator.consumed() + capacity)
-            } else {
-                None
+            let flattened_capacity = field_iterator
```
vezenovm, line 314: I definitely prefer this approach to that of the parent. It didn't make sense to me that we should have to toggle `with_length_prefix` for both arrays and slices. As we have the types we should know that a slice always has a capacity prefix and an array never has that prefix.

compiler/noirc_printable_type/src/lib.rs
aakoshh: Yes I think I added some extra block that wasn't needed, but in this PR this has been removed now, I found a better way.

compiler/noirc_printable_type/src/lib.rs
aakoshh: Let me explain the difficulties that I saw with fully reasoning out which parameter need to be padded and which one doesn't. It might be that there is a way to do it, I haven't fully thought it through. 

So we have a number of `ForeignCallParam`s, and where it's an `Array` then it might come from a `HeapVector` or a `HeapArray`, we don't know without looking at the type. However, the `PrintableType` we have are of a different granularity: say we have a `Tuple` as a parameter for the print, and it happens to contain a `Slice` and an `Array` field, interspersed with some others, then these are going to appear as separate `ForeignCallParam`s, but to tell which one is the `Slice` and which one is an `Array`, we have to unpack the `Tuple` and match it to the parameters. 

Now, I wasn't sure if one of the `Tuple` fields was itself a `Tuple`, then I think it gets flattened out as well, so we can't just iterate over the parameters, we have to recursively unpack the types until they match the fields. It seemed messy. 

`decode_` does it by recursing over the type tree, while consuming flattened data, so I thought if I prefix the vectors, then I would know where to expect it and where not to. But it turned out to be rather artificial, as the `Array` itself can contain complex data, and only the top one prefixed. 

Then I tried to prefix at the source, and by doing so it makes it so much more sense: we attach not some artificial "byte size" but an actual type-dependent vector capacity. That made me finally understand what this capacity even means, which isn't trivial, but at least it depends on the types, not the size of the data, which is easier to attach in the interpreter for example.

aakoshh:
> So it seems you ultimately decided to pass the capacity (when reading the foreign call inputs in the VM) due to the call inputs all being flattened?

The call inputs are not all flattened, although I kinda wish they would be, because it's a bit hard to understand still why something is or isn't flattened. Maybe it's how the variables in the SSA are? 

For example the members of a tuple are passed as separate parameters, but a slice of tuples are passed as a single array. 

And then there is the revert data, which is already flattened when the printer sees it, but cannot contain slices. The output of a foreign call also comes flattened. Debug variable content is stored flattened.

I thought if we include the capacity at the source in the VM we can limit this special handling to slices in a way that is easy to explain, without having to radically rewrite the decoding part to match types with fields precisely. Not sure if this is an acceptable break in backwards compatibility, but I wanted to throw it out there. Doing this limits the scope of this change, for example debug variables automatically get it.

aakoshh:
Closing in favour of https://github.com/noir-lang/noir/pull/9321

Issue #9274, opened 2025-07-22
format strings cannot be printed


benesjan:
### Bug

This code:

```rust
comptime fn my_macro(s: TypeDefinition) -> Quoted {
    let name = s.name();
    let typ = s.as_type();

    quote {
        impl $typ {
            fn random_func(self)  {
                let note_name = f"$name";
                std::static_assert(false, f"{note_name}");
            }
        }
    }
}

#[my_macro]
struct MyStruct {
    a: u32,
    b: u32,
}

fn main() -> pub Field {
    let my_struct = MyStruct { a: 1, b: 2 };
    my_struct.random_func();
    0
}
```

fails to compile with:

```
The application panicked (crashed).
Message:  internal error: entered unreachable code: format strings cannot be printed
Location: compiler/noirc_frontend/src/hir_def/types.rs:2716

This is a bug. We may have already fixed this in newer versions of Nargo so try searching for similar issues at https://github.com/noir-lang/noir/issues/.
If there isn't an open issue for this bug, consider opening one at https://github.com/noir-lang/noir/issues/new?labels=bug&template=bug_report.yml
```

### Nargo Version

nargo version = 1.0.0-beta.8

asterite:
Reduced:

```noir
fn main() {
    let x = f"";
    println(f"{x}");
}
```

(same thing with `static_assert(false, f"{x}")`)

My guess is that this should be a compile error rather than a panic but I'm not sure.

TomAFrench:
Yeah, I would say that this should be updated to a compiler error.

benesjan:
I was trying to interpolate the fmt string in another string when trying to get get the type information printed in the error message [here](https://github.com/AztecProtocol/aztec-packages/blob/0a3f96547a995e1ae888ac576258ac51ca4166b9/noir-projects/aztec-nr/aztec/src/macros/notes.nr#L46). Instead of `Note packed length exceeds the maximum...` I would like to have there `SpecificNoteTypeName packed length exceeds the maximum...`.

So if this is only updated to a compiler error it would be nice to get some feedback on how that could be done. Strings and fmt strings don't seem to play well with quotes.

asterite:
@benesjan This seems to work:

```noir
use std::meta::ctstring::AsCtString;

comptime fn my_macro(s: TypeDefinition) -> Quoted {
    let name = s.name();
    let message: str<_> =
        f"{name} packed length exceeds the maximum".as_ctstring().as_quoted_str!();
[...]

PR #9293, opened 2025-07-23
chore: Add a helper function `fmtstr::as_quoted_str`

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/9274#issuecomment-3108164960

## Summary\*

Adds a helper function to make this behavior more discoverable for users.

## Additional Context

`as_quoted_str` is not actually required here since lowering a format string will lower it into a normal string, per the example in the metaprogramming docs:

```noir
fn main() {
    let foo = comptime {
        let i = 2;
        f"i = {i}"
    };
    assert_eq(foo, "i = 2");
}
```
But adding the method makes this behavior easier to find for users and removes the need for an extra comptime block sometimes.

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):
Nice!

PR #9295, opened 2025-07-23
fix: top-level item in dependency isn't always visible

### LINKED ISSUE #9277: Add an `unreachable_pub` system
We've started enforcing visibility rules but there's an important case which is not being enforced. It's currently possible to reach inside of private modules to pull out internals which are marked public, e.g:

```noir

mod foo { 
  pub fn bar() {
    // This function should not be accessible from outside `foo`
  }
}

fn main() {
    foo::bar();
}
```

This program should not compile but it does. We should update this to display a visibility warning when trying to access `foo::bar`.

Ideally we'd have some form of lint which displays items in a module with a stricter visibility than the item itself.

@asterite, interested?

asterite:
# Description

## Problem

Resolves #9277

## Summary

When there's a path `foo::bar::baz`, it's guaranteed that `foo` is always visible from where we are at. That's what the code did. However, that logic of "is this is first segment?" was also triggered when dependencies were solved. For example if we have `some_dependency::foo::bar::baz`, the code would first solve `some_dependency`, then solve `foo::bar::baz` starting from that dependency. In that case it's no longer true that the first segment is always visible.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/path_resolution.rs
+++ b/compiler/noirc_frontend/src/elaborator/path_resolution.rs
@@ -468,7 +468,11 @@ impl Elaborator<'_> {
             });
         }
 
-        let plain_or_crate = matches!(path.kind, PathKind::Plain | PathKind::Crate);
+        let first_segment_is_always_visible = match path.kind {
+            PathKind::Crate => true,
+            PathKind::Plain => importing_module == starting_module,
+            PathKind::Dep | PathKind::Super | PathKind::Resolved(_) => false,
+        };
 
         // The current module and module ID as we resolve path segments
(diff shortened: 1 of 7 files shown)


asterite:
It seems we'd need to change the bignum library so that `mod bignum` is actually `pub mod bignum`...

@TomAFrench Do you think we need to wait for that to happen or should we merge this now and eventually fix that?

TomAFrench:
Looks like there's issues with the semaphore circuit as well. I think it'ld be quicker to just push out some fixes in the libraries rather than going through the effort of disabling these and re-enabling them again.

TomAFrench (approved):

Issue #9251, opened 2025-07-18
comptime != acir: storing in an explicit reference to a struct field


jfecher:
### Aim

Related to https://github.com/noir-lang/noir/issues/9154

Executing:
```
fn main() {
    let ct = comptime { store_through_pointer(); }
    let acir = store_through_pointer();
    assert(ct);
    assert(acir); // fails here
}

fn store_through_pointer() -> bool {
    let a = &mut (false,);
    let b = &mut a.0;
    *a = (true,);
    println(*b);
    *b
}
```
Note: After https://github.com/noir-lang/noir/pull/9250, this test is already present in test_programs/execution_success/struct_assignment_with_shared_ref_to_field

### Expected Behavior

All asserts should pass

### Bug

The `assert(acir);` fails

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

PR #9307, opened 2025-07-24
fix: Make Ssa-gen use existing reference when compiling `&mut foo.bar.baz`

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/9251
This fixes the final `false` part of the table in https://github.com/noir-lang/noir/pull/9250

## Summary\*

When you did `&mut foo.bar.baz` previously this would translate to `&mut (*foo).bar.baz` internally due to `.` performing auto dereferencing. When we have a leading `&mut` however we do not want this behavior and want `.` to implicitly perform as an offset instead. We didn't find this before because it was only observable when you had another referenced aliased inside the struct, as in the linked issue.

To fix this I've:
- Created a new function `elaborate_reference` which is only called when elaborating the rhs of `&mut rhs`
- `elaborate_reference` passes a flag to elaborate member access telling it not to automatically dereference and return a boolean if it did pass through a reference
- If this returned flag is true we need to skip the `&mut` because the member access is already returning a reference, so we do not get a `&mut &mut T`.
- Instead of replacing the unary op with just its rhs we need to keep it with a new `skip: bool` flag so that `nargo expand` still prints the `&mut` without removing it, and `--show-monomorphized` does as well.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/debugger/ignored-tests.txt
+++ b/tooling/debugger/ignored-tests.txt
@@ -17,3 +17,4 @@ mutate_array_copy
 regression_8229
 regression_8926
 to_radix_error
+reference_cancelling


asterite (approved):
Looks good!

PR #9309, opened 2025-07-24
fix: allow nested fmtstr

### LINKED ISSUE #9274: format strings cannot be printed

asterite:
# Description

## Problem

Resolves #9274

## Summary

I noticed that nested fmtstr did work in comptime so I checked if it would be possible to actually support this at runtime. It turns out that it is (after spending some time figuring out how things are encoded).

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/hir_def/types.rs
```diff
@@ -2746,7 +2746,13 @@ impl From<&Type> for PrintableType {
                     .expect("Cannot print variable sized strings");
                 PrintableType::String { length: size }
             }
-            Type::FmtString(_, _) => unreachable!("format strings cannot be printed"),
+            Type::FmtString(size, typ) => {
+                let dummy_location = Location::dummy();
+                let size = size
+                    .evaluate_to_u32(dummy_location)
+                    .expect("Cannot print variable sized strings");
```
jfecher, line 2753: Maybe we should issue an actual error here instead of moving the panic to a less common situation?
Although the error message would use the dummy location it seems. I wonder if we can forward the location from the println somehow.

jfecher (commented):
Nice! Just one concern on a potential new panic

compiler/noirc_frontend/src/hir_def/types.rs
asterite: Do you know if it's possible for `evaluate_to_u32` to fail here? My thinking is that this happens after monomorphization and if was dynamic it would have errored by now.

That said, I'll try to get a location here.

asterite: Hm, it seems monomorphization type handling doesn't happen for these arguments so what I said above is wrong.

asterite: Changing this would require some form of Result, and at least in one part of the code there's no Result returned from a function that calls this, so...
Before making this change, it would be nice to find a way to trigger this failure. A similar panic exists for the string case but I don't know if it's possible to end up with `str<N>` or `fmtstr<N, T>` where `N` is not a constant.

vezenovm: > A similar panic exists for the string case but I don't know if it's possible to end up with `str<N>` or `fmtstr<N, T>` where `N` is not a constant.

Strings are treated as arrays so `N` should always be constant. We do the same things for strings, moving away from this panic feels like a separate issue to this PR.

PR #9320, opened 2025-07-24
fix(ssa_interpreter): `push_back` and `pop_back` to slices with padding

### LINKED ISSUE #9312: SSA interpreter returns wrong result when popping the back of merged slice
### Aim

Inspired by https://github.com/noir-lang/noir/issues/9271 and https://github.com/noir-lang/noir/issues/9267

The following program returns wrong result in the SSA interpreter:

```rust
fn main(a: bool) -> pub (u32, Field) {
    let s: [Field] = if a { &[1, 2] } else { &[3, 4, 5] };
    let (s, f) = s.pop_back();
    (s.len(), f)
}
```

input:
```
a = true
```

### Expected Behavior

It should return `[1, 20]`. It works in ACIR and Brillig.

### Bug

```console
❯ cargo run -q -p nargo_cli -- interpret --force --silence-warnings 
...
--- Interpreter result after Inlining (step 21):
Ok(u32 1, Field 20)
---
--- Interpreter result after Remove IfElse (step 22):
Ok(u32 1, Field 0)
---
...
```

The problem is that the interpreter pops the end of the physical slice, but it ignores the actual length of the slice, which is less then its merged capacity.
```
Ok(u32 1, Field 20)
---
After Remove IfElse (step 22):
  b0(v0: u1):
    enable_side_effects v0
    v3 = make_array [Field 10, Field 20] : [Field]	// src/main.nr:2:35
    v4 = not v0
    enable_side_effects v4
    v8 = make_array [Field 30, Field 40, Field 50] : [Field]	// src/main.nr:2:58
    enable_side_effects u1 1
    v10 = cast v0 as u32
    v11 = cast v4 as u32
    v13 = unchecked_mul v10, u32 2
    v15 = unchecked_mul v11, u32 3
    v16 = unchecked_add v13, v15
    v17 = cast v0 as Field
    v18 = cast v4 as Field
    v19 = mul v17, Field 10
    v20 = mul v18, Field 30
    v21 = add v19, v20
    v22 = cast v0 as Field
    v23 = cast v4 as Field
    v24 = mul v22, Field 20
    v25 = mul v23, Field 40
    v26 = add v24, v25
    v27 = cast v0 as Field
    v28 = cast v4 as Field
    v29 = mul v28, Field 50
    v30 = make_array [v21, v26, v29] : [Field]
    v32, v33, v34 = call slice_pop_back(v16, v30) -> (u32, [Field], Field)	// src/main.nr:3:18
    return v32, v34
}


```

### To Reproduce

See above. 

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.8 noirc version = 1.0.0-beta.8+ee66c987f6392c05e6b46cdd9560d38251b0fb24 (git version hash: f0731641b20bea7ed4bd52194b235b034022439a, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9312 

## Summary\*

Change `slice_push_back` and `slice_push_front` to take into account any potential extra items in the slice which are beyond its `length`. Push and pop must operate before this region. This is achieved by removing any potential extras, since they are never accessed by the interpreter.

## Additional Context

This situation arises when we have something like this code:
```rust
let a: [u32] = [1, 2];
let b: [u32] = [3, 4, 5, 6]; 
let c: [u32] = if p { a } else { b };
let (d, e) = c.pop_back();
let f = c.push_back(7);
```
The SSA pass that removes the `IfElse` expressions creates a slice like this:
```
c_len = p * 2 + (1 - p) * 4
c_data = [p * a[0] + (1-p) * b[0], p * a[1] + (1-p) * b[1], b[2], b[3]]
```
If `p` is true, `c` ends up looking like `[1,2,5,6]`, and after pushing we want it to be `(3, [1,2,7,5,6])`, or if we pop then it should return `(1, [1,5,6], 2)` (length, data, elem). 

Initially I appended the extra items back onto the slice, but recognised some alternatives:
* We could drop the extras; at least I can't think of a situation outside of printing where they would come into play. 
* We could overwrite the extras in `push_back`, rather than grow the slice, to reclaim space; not like the interpreter is necessarily built to be efficient, just to maybe match what ACIR or Brillig might do. 

I checked what ACIR passes to the print oracle after a push/pop, but it doesn't looks like something the interpreter aims to match:

```rust
fn main(a: bool) {
    let s0: [u32] = if a { &[10, 20] } else { &[30, 40, 50] };
    println(s0);
    let (s1, _) = s0.pop_back();
    println(s1);
    let s2 = s0.push_back(60);
    println(s2);
}
```
The values passed to the foreign function look like this:
```
args[0] = Single(2)
args[1] = Array([10, 20, 0])
&[10, 20]
args[0] = Single(1)
args[1] = Array([10, 20, 0])
&[10]
args[0] = Single(3)
args[1] = Array([10, 20, 60, 60])
&[10, 20, 60]
```
So popping didn't make the array shorter, just decreased the length, and pushing both filled the extra element _and_ added a new one.

For this reason I concluded that the extra elements shouldn't be relied on to have any particular content, and it's best if we truncate the slices in the interpreter during pushing and popping, rather than go the extra mile to keep the extra elements. I didn't update other slice operations though.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/interpreter/intrinsics.rs
+++ b/compiler/noirc_evaluator/src/ssa/interpreter/intrinsics.rs
@@ -511,12 +511,11 @@ impl<W: Write> Interpreter<'_, W> {
         let element_types = slice.element_types.clone();
 
         // The slice might contain more elements than its length.
-        // Insert the new values before the extras.
-        let mut padding = new_elements.split_off(element_types.len() * length as usize);
+        // We need to either insert before the extras, overwrite, or remove them.
+        new_elements.truncate(element_types.len() * length as usize);
         for arg in args.iter().skip(2) {
             new_elements.push(self.lookup(*arg)?);
         }
-        new_elements.append(&mut padding);
 
         let new_length = Value::Numeric(NumericValue::U32(length + 1));
         let new_slice = Value::slice(new_elements, element_types);
@@ -553,11 +552,11 @@ impl<W: Write> Interpreter<'_, W> {
         check_slice_can_pop_all_element_types(args[1], &slice)?;
 
         // The slice might contain more elements than its length.
-        // We want the last valid element, then we can put the extras back.
-        let mut padding = slice_elements.split_off(element_types.len() * length as usize);
+        // We want the last valid element, ignoring any extras following it.
+        // We don't ever access the extras, so we might as well remove any.
+        slice_elements.truncate(element_types.len() * length as usize);
         let mut popped_elements = vecmap(0..element_types.len(), |_| slice_elements.pop().unwrap());
         popped_elements.reverse();
-        slice_elements.append(&mut padding);
 
         let new_length = Value::Numeric(NumericValue::U32(length - 1));
         let new_slice = Value::slice(slice_elements, element_types);


jfecher:
> and pushing both filled the extra element and added a new one.

I wonder if this could be related to any of the performance issues with slices.

Maybe this could happen if e.g. you have a `if foo { slice = slice.push_back(..) }` which causes us to merge a slice of length N with one of length N + 1, so you'd get these empty spaces pushed every loop effectively and not used by the next push back.

jfecher (approved):

PR #9321, opened 2025-07-24
fix(brillig_vm): Remove slice padding for foreign call inputs

vezenovm:
# Description

## Problem\*

Alternative to https://github.com/noir-lang/noir/pull/9288

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/acvm-repo/brillig/src/opcodes.rs
+++ b/acvm-repo/brillig/src/opcodes.rs
@@ -84,6 +84,31 @@ impl HeapValueType {
     pub fn field() -> HeapValueType {
         HeapValueType::Simple(BitSize::Field)
     }
+
+    /// Returns the total number of field elements required to represent this type in memory
+    pub fn flattened_size(&self) -> usize {
+        match self {
+            HeapValueType::Simple(_) => 1,
+
+            HeapValueType::Array { value_types, size } => {
+                // Recursively compute the flattened size of all value types
+                let element_size: usize = value_types
+                    .iter()
+                    .map(|ty| ty.flattened_size())
+                    .sum();
+
+                element_size * size
+            }
+
+            // The size of a vector is stored separately and should be handled by the caller of this method
+            HeapValueType::Vector { value_types } => {
+                value_types
+                    .iter()
+                    .map(|ty| ty.flattened_size())
+                    .sum()
+            }
+        }
+    }
 }
 
 /// A fixed-sized array starting from a Brillig memory location.
--- a/acvm-repo/brillig_vm/src/lib.rs
+++ b/acvm-repo/brillig_vm/src/lib.rs
@@ -525,10 +525,39 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>> VM<'a, F, B> {
                     // resolved inputs back to the caller. Once the caller pushes to `foreign_call_results`,
                     // they can then make another call to the VM that starts at this opcode
                     // but has the necessary results to proceed with execution.
+                    let mut previous_input: Option<ForeignCallParam<F>> = None;
                     let resolved_inputs = inputs
                         .iter()
                         .zip(input_value_types)
-                        .map(|(input, input_type)| self.get_memory_values(*input, input_type))
+                        .map(|(input, input_type)| {
+                            let mut input = self.get_memory_values(*input, input_type);
+
+                            match input_type {
+                                HeapValueType::Vector { .. } => {
+                                    if let Some(ForeignCallParam::Single(value)) = previous_input {
+                                        // Cut off slices with their semantic length, 
+                                        // They may have a capacity which could be higher than their semantic length.
+                                        let flattened_size = input_type.flattened_size();
+                                        let fields = input.fields();
+                                        let value_as_usize = value.to_u128() as usize;
+                                        input = ForeignCallParam::Array(fields[0..(value_as_usize * flattened_size)].to_vec());
+                                    }
+
+                                    previous_input = None;
+                                }
+                                HeapValueType::Simple(BitSize::Integer(IntegerBitSize::U32)) => {
+                                    // If we have a single u32 we may have a slice representation, so store this input.
+                                    // On the next iteration, if we have a vector then we know we have the dynamic length 
+                                    // for that slice. 
+                                    previous_input = Some(input.clone());
+                                }
+                                _ => {
+                                    // Do nothing otherwise
+                                    previous_input = None;
+                                }
+                            }
+                            input
+                        })
                         .collect::<Vec<_>>();
                     return self.wait_for_foreign_call(function.clone(), resolved_inputs);
                 }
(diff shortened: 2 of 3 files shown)


aakoshh:
I like this idea. I have a similar conundrum in https://github.com/noir-lang/noir/pull/9320 where I can't think for a reason not to truncate the extra elements during slice operations.

vezenovm:
> I like this idea. I have a similar conundrum in #9320 where I can't think for a reason not to truncate the extra elements during slice operations.

I'll be able to take a deeper look at #9320 tomorrow. For the interpreter I agree that we should be able to remove the padding as we have access to the actual dynamic length.

aakoshh:
I checked on the parent of this PR what ACIR passes to the oracle when the hybrid slice is being popped/pushed (appended to the PR description in the interpreter fix), and it's definitely not what the interpreter would pass, so these values should not be relied on by anybody. Removing would be the cleanest approach 👏

compiler/noirc_printable_type/src/lib.rs
```diff
@@ -311,29 +311,12 @@ pub fn decode_printable_value<F: AcirField>(
...
-                n => flattened_capacity / n,
-            };
-
             let mut array_elements = Vec::with_capacity(length);
 
             for _ in 0..length {
                 array_elements.push(decode_printable_value(field_iterator, typ));
             }
 
-            // Consume padding.
```
aakoshh: The interpreter needs to be updated as well, in `intrinsics.rs`, not to append the capacity.

acvm-repo/brillig_vm/src/lib.rs
```diff
@@ -525,11 +525,50 @@ impl<'a, F: AcirField, B: BlackBoxFunctionSolver<F>> VM<'a, F, B> {
...
+                    // but it is cumbersome, and the cleanest solution is not to send the extra empty
+                    // items at all. To do this, however, we need infer which input is the vector length.
+                    let mut vector_length: Option<usize> = None;
+
                     let resolved_inputs = inputs
                         .iter()
                         .zip(input_value_types)
-                        .map(|(input, input_type)| self.get_memory_values(*input, input_type))
+                        .map(|(input, input_type)| {
+                            let mut input = self.get_memory_values(*input, input_type);
```
asterite, line 542: It seems `get_memory_values` also pairs simple values with vectors? And then it's done again here below? I'm almost sure that's right but I just don't understand why.

compiler/noirc_evaluator/src/ssa/interpreter/intrinsics.rs
```diff
@@ -776,37 +774,56 @@ fn new_embedded_curve_point(
...
-            Value::ArrayOrSlice(array_value) => {
-                for value in array_value.elements.borrow().iter() {
-                    go(value, fields);
+                Value::ArrayOrSlice(array_value) => {
+                    let length = match vector_length {
+                        Some(length) if array_value.is_slice => {
+                            length * array_value.element_types.len()
+                        }
+                        _ => array_value.elements.borrow().len(),
+                    };
```
asterite, line 809: Should `vector_length` be set to `None` here? Alternative there's `Option::take` we could use (and in the other places too).

compiler/noirc_evaluator/src/ssa/interpreter/intrinsics.rs
```diff
@@ -776,37 +774,56 @@ fn new_embedded_curve_point(
...
+                    go(array_value.elements.borrow().iter().take(length), fields);
+                }
+                Value::Function(id) => {
+                    // Based on `decode_printable_value` it will expect consume the environment as well,
+                    // but that's catered for the by the SSA generation: the env is passed as separate values.
+                    fields.push(FieldElement::from(id.to_u32()));
```
asterite, line 815: I think `vector_length` must also be set to `None` here.

acvm-repo/brillig_vm/src/lib.rs
aakoshh: It is a bit confusing, for example I wasn't sure if vectors can only be top level, because here we have their length as a separate variable, clearly mapped to a separate `ForeignCallParam`, but recursively this isn't obvious. 

If I understand your question, you're pointing out that we seem to be doing things twice. I think we are not doing that, because:
* The `inputs` here are coming from the `Opcode::ForeignCall`, so we expect that the call parameters, whoever put them together, included the length and the corresponding vector one after the other. 
* Then we call `get_memory_value` on _individual_ parameters to get the flattened contents by calling `read_slice_of_values_from_memory` to recursively read the contents of arrays and vectors
* `read_slice_of_values_from_memory` deals with _multiple_ values, like the handling of `inputs` from the `Opcode`: these are the fields that sit next to each other in an array of tuples for example. Here we can again see length followed by the corresponding vector pointer, so we do the pruning again, but these are not the same values as we did it on above. 

So it's like this:
```
Opcode::ForeignCall(inputs = [array1, length2, vector2, param3])
                              |                |
                              |                [size, start]
                              |                       |
                              |                       [data0, data1, ...]
                              |
                              [length0, vector0, length1, vector1]
                                            |             |
                                            |             [size, start]
                                            |                    |
                                            |                    [data0, data1, ...]
                                            [size, start]
                                                   |
                                                   [data0, data1, ...]
```
When we call `get_memory_values(vector2)` we don't get the pruning because `length2` is in the `inputs`, and `vector2` doesn't carry that information, but when we call `get_memory_value(array1)`, then internally we do pair up `length{i}` to `vector{i}`.

compiler/noirc_evaluator/src/ssa/interpreter/intrinsics.rs
aakoshh: It happens for all cases on line 821: we either have a number, and then set the length to that, or we don't, because we just used it or had something else, either way it's being reset.

compiler/noirc_evaluator/src/ssa/interpreter/intrinsics.rs
aakoshh: I liked the `Option::take` idea, but alas I think it would be more dangerous: if only call `take` in this `ArrayOrSlice` arm, then we would fail to clear it out if the field following a `u32` is not a vector, so we would have to start by taking it before even looking at what value we are dealing with with. So we can either unconditionally `take` in the beginning, or unconditionally set it at the end, which is what happens now.

acvm-repo/brillig_vm/src/lib.rs
asterite: Thank you! That ascii diagram is very helpful! 😍

compiler/noirc_evaluator/src/ssa/interpreter/intrinsics.rs
asterite: Aaaaah... I missed that last bit of code. Thanks!

TomAFrench:
I imagine this could still be breaking if there's code in aztec-packages which is assuming that it needs to do this truncation itself.

aakoshh:
I thought that wouldn’t be a problem, because if the slices are passed as a separate ForeignCallParam then they would just truncate something that is already shorter, which is a no-op. 

If they instead dealt with flattened data, then they would have had to hardcode some value for the capacity on their side, since nothing indicates it in the data, which would be tight coupling to an implementation detail that I imagine wasn’t meant to be part of the interface, and would have been fragile anyway.

acvm-repo/brillig/src/opcodes.rs
```diff
@@ -84,6 +84,25 @@ impl HeapValueType {
...
+
+                element_size * size
+            }
+
+            // The size of a vector is stored separately and should be handled by the caller of this method
+            HeapValueType::Vector { value_types } => {
+                value_types.iter().map(|ty| ty.flattened_size()).sum()
+            }
+        }
+    }
```
TomAFrench: This is a bit wonky as it's returning the full type size for simple and array types but only the element size for vectors. I've had to implement a similar method in a different PR but I return `None` here if we encounter a vector.

aakoshh: I'll use `Either` for it to make a distinction.

aakoshh: Or maybe just move this function as a private one into the module where it's called, as an Array of Vector is not handled by it.

TomAFrench (approved):
LGTM

PR #9324, opened 2025-07-25
fix: some nargo expand fixes

asterite:
# Description

## Problem

For https://github.com/AztecProtocol/aztec-packages/issues/15971

## Summary

This mainly fixes an issue I knew about, but didn't consider it a priority to fix: calls generated by macros could refer to traits that are not in scope, if they are done using macro calls like `get_trait_impl`, then `methods`, etc. `nargo expand` will now check if the trait related to the call is in scope and, if not, use `<Type as Trait>::method` to call it.

While doing this I noticed that `nargo expand` would also not put unquote markers, so I fixed that too.

## Additional Context

I think we should put a disclaimer somewhere that `nargo expand` is mostly a debugging tool and that it will try its best to produce code that compiles, but it won't always be possible (even if we fix all bugs). For example if a macro inserts a struct value, `nargo expand` will show it as `Struct { field1: value1, field2: value2, ... }` even though those fields might be private.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir/comptime/display.rs
+++ b/compiler/noirc_frontend/src/hir/comptime/display.rs
@@ -36,7 +36,7 @@ pub(super) fn display_quoted(
         writeln!(f, "quote {{")?;
         let indent = indent + 1;
         write!(f, "{}", " ".repeat(indent * 4))?;
-        TokensPrettyPrinter { tokens, interner, indent }.fmt(f)?;
+        TokensPrettyPrinter { tokens, interner, indent, preserve_unquote_markers: false }.fmt(f)?;
         writeln!(f)?;
         let indent = indent - 1;
         write!(f, "{}", " ".repeat(indent * 4))?;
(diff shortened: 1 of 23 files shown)


TomAFrench:
Are there any ignored `nargo expand` tests which need to be removed from the list of tests in the build.rs file?

> I think we should put a disclaimer somewhere that nargo expand is mostly a debugging tool and that it will try its best to produce code that compiles, but it won't always be possible (even if we fix all bugs). For example if a macro inserts a struct value, nargo expand will show it as Struct { field1: value1, field2: value2, ... } even though those fields might be private.

Agreed that we should highlight that this is done on a best-efforts basis.

TomAFrench (approved):
LGTM, only question is whether we can take some more tests off of the ignore list.

asterite:
@TomAFrench Good point. I changed the tests so we can now when they start passing and, indeed, some more tests passed.

PR #9328, opened 2025-07-25
fix: substitute bindings in type before canonicalization

asterite:
# Description

## Problem

Resolves https://github.com/noir-lang/noir/issues/9326

Alternative to #9327

## Summary

This was another issue related to type variables, their ordering, and `sort_commutative`.

The idea of `sort_commutative` is to transform an expression like `(1 + a) + 2` into `(a + 1) + 2`, that is, always place constants on the right-hand side (that way when we need to simplify the addition above we can rely on the constants always being in those places).

For types that aren't constants, they are rearranged according to the order given by `impl Ord for Type`, which, for type variables, depends on their ID.

The problem was that we ended up with an expression like `(a + b) - 1` where, depending on the entire code to compile, the same expression could end up having `a: '1, b: '2` or `a: '2, a: '1`, that is, both are type variables but they got different IDs.

(Why they get different IDs? I didn't check, but the code should work fine regardless of this, so...)

Now imagine `b` is bound to a constant in `bindings`. `sort_commutative` will either leave it like `(a + b) - 1` or `(b + a) - 1`. That means that, when eventually we try to simplify the above we'll end up with `(a + 1) - 1` or `(1 + a) -1`): we'll be able to simplify it in the first scenario but not in the second one.

What this PR does is to eagerly replace `b` with `1` way before we reach `sort_commutative`. We do this before type canonicalization.

This partially reverts what #9265 did, greatly simplifying the code (no need to pass `bindings` all around).

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/hir_def/types/unification.rs
+++ b/compiler/noirc_frontend/src/hir_def/types/unification.rs
@@ -272,6 +272,7 @@ impl Type {
 
             (Constant(value, kind), other) | (other, Constant(value, kind)) => {
                 let dummy_location = Location::dummy();
+                let other = other.substitute(&bindings);
                 if let Ok(other_value) = other.evaluate_to_field_element(kind, dummy_location) {
                     if *value == other_value && kind.unifies(&other.kind()) {
                         Ok(())
@@ -287,7 +288,7 @@ impl Type {
                             rhs.clone(),
                         );
 
-                        new_type.try_unify(lhs, bindings)?;
+                        new_type.try_unify(&lhs, bindings)?;
                         Ok(())
                     } else {
                         Err(UnificationError)
@@ -472,6 +473,7 @@ impl Type {
             if let Some(lhs_op_inverse) = lhs_op.approx_inverse() {
                 let kind = lhs_lhs.infix_kind(lhs_rhs);
                 let dummy_location = Location::dummy();
+                let lhs_rhs = lhs_rhs.substitute(bindings);
                 if let Ok(value) = lhs_rhs.evaluate_to_field_element(&kind, dummy_location) {
                     let lhs_rhs = Box::new(Type::Constant(value, kind));
                     let new_rhs =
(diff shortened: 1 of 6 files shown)


compiler/noirc_frontend/src/hir_def/types/unification.rs
```diff
@@ -272,9 +272,7 @@ impl Type {
 
             (Constant(value, kind), other) | (other, Constant(value, kind)) => {
                 let dummy_location = Location::dummy();
-                if let Ok(other_value) =
-                    other.evaluate_to_field_element(kind, bindings, dummy_location)
-                {
+                if let Ok(other_value) = other.evaluate_to_field_element(kind, dummy_location) {
```
jfecher, line 276: I think this may be a similar issue: I think we'd need to substitute bindings on other before evaluating to field element. E.g. what if other is a type variable bound to `2` in the bindings?

compiler/noirc_frontend/src/hir_def/types/unification.rs
```diff
@@ -474,9 +472,7 @@ impl Type {
             if let Some(lhs_op_inverse) = lhs_op.approx_inverse() {
                 let kind = lhs_lhs.infix_kind(lhs_rhs);
                 let dummy_location = Location::dummy();
-                if let Ok(value) =
-                    lhs_rhs.evaluate_to_field_element(&kind, bindings, dummy_location)
-                {
+                if let Ok(value) = lhs_rhs.evaluate_to_field_element(&kind, dummy_location) {
```
jfecher, line 477: Same here

jfecher (approved):
Looks good - I found a couple spots we could do more substitutions for possibly related issues

compiler/noirc_frontend/src/hir_def/types/unification.rs
asterite: Done! I was thinking that maybe we can call `substitute` right at the beginning of `try_unify` regardless of the expression being an infix one, though maybe that'll be too heavy. For now I applied your suggestions.

compiler/noirc_frontend/src/hir_def/types/unification.rs
asterite: Done!

PR #9338, opened 2025-07-28
fix: add missing js_sys import in foreign_call outputs module

Galoretka:
Add missing js_sys import to fix compilation errors in foreign_call outputs module.
The module was using js_sys::Array without importing the js_sys crate.


TomAFrench:
This should compile fine as we compile this in CI on every PR and on master. This new import doesn't seem necessary to me.

PR #9340, opened 2025-07-28
fix(acir_gen): Keep the first range check using the lowest bit size

### LINKED ISSUE #9335: ACIR != Brillig: ACIR prints result of overflow and fails on other constraint
### Aim

Reproduce a nightly failure from https://github.com/noir-lang/noir/actions/runs/16559377316/job/46826043334

The following program fails correctly in Brillig and the SSA interpreter, but behaves in a strange way in ACIR:

```rust
global G_B: u64 = 2288862572;
fn main(a: u64) -> pub u64 {
    let b = a * 18259827970668162041_u64;
    println(f"b = {b}");
    assert(b == G_B, "YHV");
    b
}
```

Input:
```toml
a = "0x000000000000000000000000000000000000000000000000000000006f3c2bda"
```

### Expected Behavior

They should agree on the outcome, which is an overflow in the multiplication.

### Bug

```console
b = 16372204933514773770
error: Assertion failed: YHV
  ┌─ src/main.nr:5:13
  │
5 │     assert(((b as u128) == (G_B as u128)), "YHV");
  │             ----------------------------
  │
  = Call stack:
    1. src/main.nr:5:13

Failed assertion
```

Note how it even printed the value of `b`. But if we comment out the `assert`, or run it with `--force-brillig`, we get a different failure:
```console
  ┌─ src/main.nr:3:13
  │
3 │     let b = (a as u64) * 18259827970668162041_u64;
  │             -------------------------------------
  │
  = Call stack:
    1. src/main.nr:3:13

Failed assertion
```

I see nothing suspicious in the SSA:
```
After Initial SSA:
g0 = u64 2288862572

acir(inline) fn main f0 {
  b0(v1: u64):
    v3 = mul v1, u64 18259827970668162041         	// src/main.nr:3:13
    v9 = make_array b"b = {b}"                    	// src/main.nr:3:13
    call f1(v9, Field 1, v3)                      	// src/main.nr:4:5
    v12 = eq v3, u64 2288862572                   	// src/main.nr:5:12
    constrain v3 == u64 2288862572, "YHV"         	// src/main.nr:5:12
    return v3
}
acir(inline) fn println f1 {
  b0(v1: [u8; 7], v2: Field, v3: u64):
    call f2(u1 1, v1, v2, v3)                     	// std/lib.nr:40:9
    return
}
brillig(inline) fn print_unconstrained f2 {
  b0(v1: u1, v2: [u8; 7], v3: Field, v4: u64):
    v24 = make_array b"{\"kind\":\"unsignedinteger\",\"width\":64}"	// std/lib.nr:40:9
    call print(v1, v2, v3, v4, v24, u1 1)         	// std/lib.nr:34:5
    return
}
```

Since we can actually observe a different side effect (the print) in ACIR, we can't consider these equivalent failures.

### To Reproduce

```
NOIR_AST_FUZZER_SEED=0x2422539c00100000 cargo test -q -p noir_ast_fuzzer_fuzz acir_vs_brillig
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.9 noirc version = 1.0.0-beta.9+e7fb7f3bf083de9fc6d430e7c30a571ea94534e4 (git version hash: e7fb7f3bf083de9fc6d430e7c30a571ea94534e4, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9335

## Summary\*

Changes the `redundant_range` optimisation so that instead of only keeping the range check which matches the minimum bit size, it keeps the first range check, but updates it to use the minimum overall bit size for the witness. 

## Additional Context

In the integration test we started with this ACIR:
```
  BLACKBOX::RANGE [(_0, 64)] []
  EXPR [ (18259827970668162041, _0) (-1, _2) 0 ]
  BLACKBOX::RANGE [(_2, 64)] []
  EXPR [ (1, _2) -2288862572 ]
  EXPR [ (1, _1) -2288862572 ]
```
The `redundant_range` optimisation looked at the opcodes and saw that if `_2` must be equal to `-2288862572` then it must have less than 64 bits (31 apparently), so it removed the `  BLACKBOX::RANGE [(_2, 64)] []` opcode, because 64 is not 31. There was no range constraint with 31, so we were left with the `EXPR`. 

The problem was that it did this even if between the RANGE it removed and the EXPR there was a CALL to Brillig, where side effects saw the invalid value. 

With the change in this PR, the new ACIR opcodes look like this:
```
BLACKBOX::RANGE [(_0, 64)] []
EXPR [ (18259827970668162041, _0) (-1, _2) 0 ]
BLACKBOX::RANGE [(_2, 31)] []
EXPR [ (1, _2) -2288862572 ]
EXPR [ (1, _1) -2288862572 ]
```

So one extra opcode. I suppose we could try to detect the need for whether we should keep the first, or let it go until it finds the minimum, and save that extra opcode when we can, but I went with the simpler solution for now.

## Documentation\*

Check one:

# PR Checklist\*


acvm-repo/acvm/src/compiler/optimizers/redundant_range.rs
```diff
@@ -141,18 +161,82 @@ impl<F: AcirField> RangeOptimizer<F> {
...
+                // If any one of them is an input to a side effecting opcode later on,
+                // we should keep the constraints of all the others, so that if they
+                // are invalid, we don't let any indirect side effect get through.
+                // For example: [RANGE(a, 32); Assert(b < a); BrilligCall(print, b); Assert(a == 1)]
+                // we should keep the RANGE on `a`, otherwise we might let a 64 bit a go through and print b.
+                Opcode::AssertZero(expr) => {
+                    if let Some(idx) = effects.min_idx(expr) {
+                        effects.mark_expr(expr, idx);
+                    }
+                }
```
aakoshh: If we remove this, then we would only keep range constraints for witnesses which directly go to a Brillig call. I'm not sure if there is a way to attack that, I haven't been able to come up with an obvious counter example where we:
1. have an overflow on `x`
2. put `x` in an assertion together with `y`
3. print `y`
4. assert `x` with a constant
and we see `y` printed before we fail on the 4th, except in trivial cases like `y = x` or `assert(x != 10 * y)`. Everything else I tried involved calling Brillig with `x` just to produce the 2nd result. 

Nevertheless these are the tests where we would have fewer opcodes if we disabled this transitive logic:
```
a_6_array
arithmetic_binary_operations
binary_operator_overloading
bit_shifts_runtime
brillig_uninitialized_arrays
nested_array_dynamic
regression_8519
signed_arithmetic
slice_dynamic_index
u128_type
u16_support
wrapping_operations
```

aakoshh: In `regression_8519` we have this:
```
BRILLIG CALL func 0: inputs: [EXPR [ (-1, _0) 0 ], EXPR [ 340282366920938463463374607431768211456 ]], outputs: [_2, _3]
BLACKBOX::RANGE [(_2, 126)] []
BLACKBOX::RANGE [(_3, 126)] []
EXPR [ (-1, _3) (-1, _4) 340282366920938463463374607431768211455 ]
BLACKBOX::RANGE [(_4, 129)] []
EXPR [ (-1, _0) (-340282366920938463463374607431768211456, _2) (-1, _3) 0 ]
EXPR [ (1, _2) (-1, _5) 20746827117051438823981594372716013474 ]
BLACKBOX::RANGE [(_5, 126)] []
BRILLIG CALL func 1: inputs: [EXPR [ (-1, _2) 64323764613183177041862057485226039389 ]], outputs: [_6]  
EXPR [ (-1, _2, _6) (64323764613183177041862057485226039389, _6) (1, _7) -1 ]
EXPR [ (-1, _2, _7) (64323764613183177041862057485226039389, _7) 0 ]
EXPR [ (-1, _3, _7) (53438638232309528389504892708671455233, _7) (-1, _8) 0 ]
BLACKBOX::RANGE [(_8, 129)] []
EXPR [ (1, _3) -53438638232309528388129535304893203149 ]
EXPR [ (1, _1) -53438638232309528388129535304893203149 ]
```

* `_2` and `_3` are both outputs of the first Brillig call
* `_2` goes into the second Brillig call
* `EXPR [ (-1, _0) (-340282366920938463463374607431768211456, _2) (-1, _3) 0 ]` adds a common constraint on `_2` and `_3`
* because of this `BLACKBOX::RANGE [(_3, 126)] []` is kept, even though we have `EXPR [ (1, _3) -53438638232309528388129535304893203149 ]` later

So this is an example where an assert binds two witnesses without one of them being an input of a Brillig call itself.

aakoshh: This makes me think that maybe we should treat outputs the same way as `Assert`: if one output is going to be an input to Brillig calls, then the other outputs shouldn't get their range constraints delayed either. 

But that makes me think that maybe _any_ Brillig call should mean range constraints from _before_ the call should not be delayed to _after_ the call, regardless of whether they are inputs 🤔

aakoshh: Indeed:
```rust
global G_B: u64 = 2288862572;
fn main(a: u64) -> pub u64 {
    let b = a * 18259827970668162041_u64;
    println("Now you see me");
    assert(b == G_B, "YHV");
    b
}
```
With `--force-brillig` we get a failure with no prints, with ACIR we see the print.

guipublic:
In the example, you say:

> We keep the range constraint on _2, but we use 32 bits, as implied by the later assertion.

however I don't believe this is correct, because in the original program, if you have` a `such that `b` does not overflow 64 bits but is more than 32 bits, you should get:
print(b)
assertion failed: YHV

with the 'remove redundant ranges', you will get
assertion fail: b is not 32 bits.

aakoshh:
@guipublic you are right. 

I wanted to keep the basic premise of the pass, which that there will be at most one range constraint kept. I did think about whether I should keep the original range constraint, but at some point I thought that might mean keeping a whole range of them, whatever is the minimum _at the point_ of the constraint, then to keep every subsequent stricter constraint 🙈 

I also had a version where I kept every range as it was _until_ we meet the minimum, _iff_ they are needed for side effects. I guess that's what I'd have to go back to (this was never committed), with a deduplication based on seen-so-far, so maybe we keep `range(64), range(32), assert(0)`, but if we see 32 first, we skip 64.

vezenovm:
I'm not sure if changing the range optimizer is the best fix here. As per the issue this snippet:
```noir
global G_B: u64 = 2288862572;
fn main(a: u64) -> pub u64 {
    let b = a * 18259827970668162041_u64;
    println(f"b = {b}");
    assert(b == G_B, "YHV");
    b
}
```
produces the following _unoptimized_ ACIR:
```
current witness index : _2
private parameters indices : [_0]
return value indices : [_1]
BLACKBOX::RANGE [(_0, 64)] []
EXPR [ (18259827970668162041, _0) (-1, _2) 0 ]
BRILLIG CALL func 0: inputs: [EXPR [ 1 ], [EXPR [ 98 ], EXPR [ 32 ], EXPR [ 61 ], EXPR [ 32 ], EXPR [ 123 ], EXPR [ 98 ], EXPR [ 125 ]], EXPR [ 1 ], EXPR [ (1, _2) 0 ]], outputs: []
EXPR [ (1, _2) -2288862572 ]
EXPR [ (1, _1) -2288862572 ]
```
It looks like we are already missing a range check before we ever try to transform the program using the ACVM compiler methods.

aakoshh:
@vezenovm as it turns out we call this optimisation twice: the "unoptimised" is already half-optimised, then it goes through it again later. Ask me how I know.

vezenovm:
> the "unoptimised" is already half-optimised, then it goes through it again later. Ask me how I know.

Nice lol, where does this happen? 

EDIT: Ah I see we call out to `acvm::compiler::optimize` when building the artifacts in `ssa/mod.rs`

aakoshh:
First it happens at https://github.com/noir-lang/noir/blob/7d98dc2ec806c49e41c98fe54b314f85c37b5473/compiler/noirc_evaluator/src/ssa/mod.rs#L518 which is called from https://github.com/noir-lang/noir/blob/7d98dc2ec806c49e41c98fe54b314f85c37b5473/compiler/noirc_evaluator/src/ssa/mod.rs#L400

Second time it's at https://github.com/noir-lang/noir/blob/7d98dc2ec806c49e41c98fe54b314f85c37b5473/tooling/nargo_cli/src/cli/compile_cmd.rs#L252

vezenovm:
A more accurate name would be `non-transformed` instead of `unoptimized`

acvm-repo/acvm/src/compiler/optimizers/redundant_range.rs
```diff
@@ -345,12 +413,56 @@ mod tests {
...
         let acir_opcode_positions = circuit.opcodes.iter().enumerate().map(|(i, _)| i).collect();
-        let optimizer = RangeOptimizer::new(circuit);
+        let optimizer = RangeOptimizer::new(circuit, &Default::default());
         let (optimized_circuit, _) = optimizer.replace_redundant_ranges(acir_opcode_positions);
         assert_eq!(optimized_circuit.opcodes.len(), 1);
         assert_eq!(optimized_circuit.opcodes[0], Opcode::AssertZero(Witness(1).into()));
     }
 
+    #[test]
+    fn potential_side_effects() {
```
vezenovm, line 395: We can use the ACIR parser now

aakoshh:
Updated the logic to apply the strictest known constraint if there are no even stricter ones before the next side effect, or the end of the circuit.

acvm-repo/acvm/src/compiler/optimizers/redundant_range.rs
```diff
@@ -139,20 +164,64 @@ impl<F: AcirField> RangeOptimizer<F> {
                 continue;
             };
 
+            let next_side_effect = next_side_effects[idx];
```
vezenovm: This feels a little roundabout with `collect_side_effects`. Could we not pass `brillig_side_effects: &BTreeMap<BrilligFunctionId, bool>,` to `collect_ranges` match on an `Opcode::BrilligCall` here and return `next_side_effect` directly from the match based upon the `BrilligFunctionId`. We could then avoid enumerating this loop.

vezenovm (commented):
I think you have gone down the correct route of adding side effects awareness. Just some minor style comments.

acvm-repo/acvm/src/compiler/optimizers/redundant_range.rs
aakoshh: The interaction between `collect_side_effects` and `collect_ranges` certainly evolved. Originally `collect_side_effects` tracked inputs and asserts, and I didn't want to burden `collect_ranges` with that. 

However, I don't see immediately how we can avoid enumerating this loop. Here we go forward in the opcodes, while `collect_side_effects` provides information about higher indexes, opcodes this enumeration hasn't seen yet, to answer whether there will be things before the next thing happens. 

`collect_ranges` could build up two data structures, but I'm not sure it would be easier to understand that way.

Happy to revisit this, or if you can simplify it. Just want to get it right, and ~still need to look at the recursion in the Brillig opcodes~ 😬

aakoshh: I removed `collect_side_effects`. Instead `next_side_effect` is maintained during `replace_redundant_ranges`, which now goes backwards. Also simplified the switch points to just be a set of indexes where we might want range constraints to be kept.

guipublic:
In fact in BB we are also removing redundant range-checks, so this pass does not matter much. I just added as well 
'the implied ranges from array access'.

vezenovm:
> In fact in BB we are also removing redundant range-checks, so this pass does not matter much. I just added as well 'the implied ranges from array access'.

Oh I didn't realize we were removing redundant range checks in BB. In theory this pass would only benefit execution times a bit by not executing redundant range checks. Either way if we have this pass as part of the compiler it needs to be aware of side effects for execution.

> **strings**	+4 ❌	**+13.79%**	+52 ❌	**+157.58%**

However, it does look like we still have some regressions when we are not removing certain range checks. Perhaps this is simply due to some missing functionality though (e.g., the 'the implied ranges from array access').

aakoshh:
> In fact in BB we are also removing redundant range-checks, so this pass does not matter much.

Sorry, what do you mean by "this pass" here? This fix is for ACIR.

aakoshh:
> However, it does look like we still have some regressions when we are not removing certain range checks.

For the `strings` test the range checks we did not remove were for the string input:
```rust
fn main(message: pub str<11>, y: Field, hex_as_string: str<4>, hex_as_field: Field) {
...
std::println(10);
...
assert(hex_as_string == "0x41");
...
}
```
They range constraints to restrict all 4 characters in `hex_as_string` to 8 bits were left in because before the assertion on the string contents there was a side effecting print. 
```
❯ cargo run -q -p nargo_cli -- info --force --silence-warnings --print-acir
Compiled ACIR for main (non-transformed):
func 0
current witness index : _16
private parameters indices : [_11, _12, _13, _14, _15, _16]
public parameters indices : [_0, _1, _2, _3, _4, _5, _6, _7, _8, _9, _10]
return value indices : []
BLACKBOX::RANGE [(_12, 8)] []
BLACKBOX::RANGE [(_13, 8)] []
BLACKBOX::RANGE [(_14, 8)] []
BLACKBOX::RANGE [(_15, 8)] []
...
BRILLIG CALL ...
...
EXPR [ (1, _12) -48 ]
EXPR [ (1, _13) -120 ]
EXPR [ (1, _14) -52 ]
EXPR [ (1, _15) -49 ]
EXPR [ (1, _16) -65 ]
```

I moved the assertion to the top of the program to let it remove the range checks.

aakoshh:
I investigated the regression on `lambda_from_array` as well. 

We had for example this code:
```rust
fn main(x: u32) {
    lambdas_in_array_literal(x - 1);
}
fn lambdas_in_array_literal(x: u32) {
    let xs = [|| println("hi"), || println("bye"), || println("wow"), || println("big")];
    (xs[x])();
}
```
Resulting in this ACIR:
```
Compiled ACIR for main (non-transformed):
func 0
current witness index : _485
private parameters indices : [_0]
return value indices : []
BLACKBOX::RANGE [(_0, 32)] []
EXPR [ (1, _0) (-1, _1) -1 ]
...
BLACKBOX::RANGE [(_1, 2)] []
EXPR [ (-1, _164) 18 ]
EXPR [ (-1, _165) 19 ]
EXPR [ (-1, _166) 20 ]
EXPR [ (-1, _167) 21 ]
INIT (id: 1, len: 4, witnesses: [_164, _165, _166, _167])
MEM (id: 1, read at: EXPR [ (1, _1) 0 ], value: EXPR [ (1, _168) 0 ]) 
BRILLIG CALL func 3: inputs: [EXPR [ (1, _168) -7 ]], outputs: [_169]
...
```
So the array has 4 elements and we got a RANGE constraint to limit `x - 1` as witness `_1` to 2 bits. We might expect that MEM would make this RANGE redundant, because it also implies indexing the array of size 4, however this did not happen for two reasons:
* Only `AssertZero` was considered as _implied_, which is the only kind of constraint considered to make an equal length RANGE redundant. Not sure why, but this is how it worked in the original. 
* The implied size from INIT was 3, not 2, so even if it was treated like `AssertZero`, it was considered less strict than the RANGE.

I changed this so MEM is allowed to take precedence over RANGE, and by limiting the number of bits to what is required to represent the maximum index value, rather than the array size.

(In the original code `x - 1` is also used to index an array of size 1, which results in a `EXPR [ (1, _1) 0 ]` appear in the code, and for that reason all RANGE constraints were removed).

jfecher:
@aakoshh can you bump up the time for rollup-block-root-empty a bit?

CC @vezenovm for review

vezenovm:
> CC @vezenovm for review

Will be taking a look today.

acvm-repo/acvm/src/compiler/optimizers/redundant_range.rs
```diff
@@ -141,83 +158,101 @@ impl<F: AcirField> RangeOptimizer<F> {
...
+
         (Circuit { opcodes: optimized_opcodes, ..self.circuit }, new_order_list)
     }
 }
 
+/// Calculate the maximum number of bits required to index a memory block of a certain size.
 fn memory_block_implied_max_bits(init: &[Witness]) -> u32 {
     let array_len = init.len() as u32;
-    32 - array_len.leading_zeros()
+    let max_index = array_len.saturating_sub(1);
```
vezenovm, line 249: Nice find

acvm-repo/acvm/src/compiler/optimizers/redundant_range.rs
```diff
@@ -345,12 +384,69 @@ mod tests {
...
+        let optimizer = RangeOptimizer::new(circuit, &brillig_side_effects);
         let (optimized_circuit, _) = optimizer.replace_redundant_ranges(acir_opcode_positions);
         assert_eq!(optimized_circuit.opcodes.len(), 1);
         assert_eq!(optimized_circuit.opcodes[0], Opcode::AssertZero(Witness(1).into()));
     }
 
+    #[test]
+    fn potential_side_effects() {
+        // The optimizer should not remove range constraints if doing so might allow invalid side effects to go through.
+        let mut circuit = test_circuit(vec![(Witness(1), 32)]);
```
vezenovm, line 397: I'm assuming we didn't use the ACIR parser here as it would require changing some dependencies? I know the ACIR parser is currently under a `#[cfg(test)]`. We may want to make the ACIR parser its own crate so we can then import it into separate crates like `acvm` as a `dev-dependency`. This can be separate work though.

aakoshh: I just didn't want to deviate from the style of the existing unit tests, but if we were writing some new pass I'd much rather use the parser. Happy to have a follow up PR to rewrite all tests to use the parser.

aakoshh: https://github.com/noir-lang/noir/issues/9577

Issue #9270, opened 2025-07-22
ACIR != Brillig: Brillig modifies global array


aakoshh:
### Aim

Reproduces a nightly failure from https://github.com/noir-lang/noir/actions/runs/16433887822/job/46440148153

The following program returns different result in ACIR and Brillig:

```rust
global G_B: [[str<2>; 1]; 2] = [["DS"], ["EN"]];
fn main(b: pub bool) -> pub [[str<2>; 1]; 2] {
    unsafe { func_2(G_B) };
    G_B
}
unconstrained fn func_2(mut b: [[str<2>; 1]; 2]) {
    let mut idx_d: u32 = 0;
    loop {
        if (idx_d == 1) {
            break
        } else {
            idx_d = (idx_d + 1);
            b[0][0] = G_B[1][0];
        }
    }
}
```

### Expected Behavior

It should return `[["DS"], ["EN"]]` because global variables are supposed to be immutable.

### Bug

```console
❯ cargo run -q -p nargo_cli -- execute --force --silence-warnings; cargo run -q -p nargo_cli -- execute --force --silence-warnings --force-brillig
[fuzz_testing] Circuit witness successfully solved
[fuzz_testing] Witness saved to target/fuzz_testing.gz
[fuzz_testing] Circuit output: Vec([Vec([String("DS")]), Vec([String("EN")])])
[fuzz_testing] Circuit witness successfully solved
[fuzz_testing] Witness saved to target/fuzz_testing.gz
[fuzz_testing] Circuit output: Vec([Vec([String("EN")]), Vec([String("EN")])])
```

The Brillig called returns `[["EN"], ["EN"]]`

### To Reproduce

NOIR_AST_FUZZER_SEED=0x58ccbf4c00100000 cargo test -q -p noir_ast_fuzzer_fuzz min_vs_full

### Workaround

None

### Project Impact

None

### Nargo Version

nargo version = 1.0.0-beta.8 noirc version = 1.0.0-beta.8+eb0982ee8859640b18a1874b1d92b29d0e2c0499 (git version hash: eb0982ee8859640b18a1874b1d92b29d0e2c0499, is dirty: true)

### Would you like to submit a PR for this Issue?

None

PR #9346, opened 2025-07-28
chore: try printing byte arrays as strings in the SSA interpreter

asterite:
# Description

## Problem

No issue.

## Summary

While debugging https://github.com/noir-lang/noir/issues/9270 with `nargo interpret` it was hard to remember which values were correct or incorrect by looking at numbers. With this PR byte arrays that are likely strings will be shown as such, like we do everywhere else when it comes to SSA. For example:

```
Ok(rc3 [rc1 [rc2 b"EN"], rc2 [rc2 b"EN"]])
```

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/interpreter/value.rs
+++ b/compiler/noirc_evaluator/src/ssa/interpreter/value.rs
@@ -442,19 +442,24 @@ impl std::fmt::Display for ArrayValue {
         {
             const FORM_FEED: u8 = 12; // This is the ASCII code for '\f', which isn't a valid escape sequence in strings
             let printable = self.elements.borrow().iter().all(|value| {
-                matches!(value, Value::Numeric(NumericValue::U8(byte)) 
-                  if *byte != FORM_FEED && 
-                    (byte.is_ascii_alphanumeric() 
-                      || byte.is_ascii_punctuation() 
+                matches!(value, Value::Numeric(NumericValue::U8(byte))
+                  if *byte != FORM_FEED &&
+                    (byte.is_ascii_alphanumeric()
+                      || byte.is_ascii_punctuation()
                       || byte.is_ascii_whitespace()))
             });
             if printable {
-                let bytes = self.elements.borrow().iter().map(|value| {
-                    let Value::Numeric(NumericValue::U8(byte)) = value else {
-                        panic!("Expected U8 value in array, found {value}");
-                    };
-                    *byte
-                }).collect::<Vec<_>>();
+                let bytes = self
+                    .elements
+                    .borrow()
+                    .iter()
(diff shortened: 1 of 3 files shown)


compiler/noirc_evaluator/src/ssa/interpreter/value.rs
```diff
@@ -434,7 +434,34 @@ impl std::fmt::Display for ArrayValue {
...
+        write!(f, "rc{rc} {is_slice}")?;
+
+        // Check if the array could be shown as a string literal
+        if self.element_types.len() == 1
+            && matches!(self.element_types[0], Type::Numeric(NumericType::Unsigned { bit_size: 8 }))
+        {
+            const FORM_FEED: u8 = 12; // This is the ASCII code for '\f', which isn't a valid escape sequence in strings
+            let printable = self.elements.borrow().iter().all(|value| {
+                matches!(value, Value::Numeric(NumericValue::U8(byte)) 
+                  if *byte != FORM_FEED &&
```
vezenovm:
```suggestion
                  *byte != FORM_FEED && 
```
do we need this if?

compiler/noirc_evaluator/src/ssa/interpreter/value.rs
```diff
@@ -434,7 +434,34 @@ impl std::fmt::Display for ArrayValue {
...
+
+        // Check if the array could be shown as a string literal
+        if self.element_types.len() == 1
+            && matches!(self.element_types[0], Type::Numeric(NumericType::Unsigned { bit_size: 8 }))
+        {
+            const FORM_FEED: u8 = 12; // This is the ASCII code for '\f', which isn't a valid escape sequence in strings
+            let printable = self.elements.borrow().iter().all(|value| {
+                matches!(value, Value::Numeric(NumericValue::U8(byte)) 
+                  if *byte != FORM_FEED && 
+                    (byte.is_ascii_alphanumeric()
```
vezenovm: Could we just do `is_ascii()`?

compiler/noirc_evaluator/src/ssa/interpreter/value.rs
asterite: It's the `if` for the `matches!` guard. The logic is also in the evaluator. I thought we couldn't use that but I just pushed a commit to do it.

compiler/noirc_evaluator/src/ssa/interpreter/value.rs
asterite: I think that will return true for the entire ascii range, so an array like `[1, 2, 3]` would give true for that, but that likely isn't a string.

PR #9350, opened 2025-07-29
fix: ignore implicit unbound named generic when searching trait impls

### LINKED ISSUE #9168: Trait bound not satisified when it has associated constants
### Aim

This code:

```noir
pub trait Other {
    let N: u32;
}

pub trait Trait<T>
where
    T: Other,
{}

impl Other for Field {
    let N: u32 = 1;
}

impl Trait<Field> for i32 {}

fn main() {}
```

Gives an error:

```
error: The trait bound `Field: Other<N = <T as Other>::N>` is not satisfied
   ┌─ src/main.nr:7:8
   │
 7 │     T: Other,
   │        ----- required by this bound in `Trait`
   ·
14 │ impl Trait<Field> for i32 {}
   │                       --- The trait `Other<N = <T as Other>::N>` is not implemented for `Field`
```

It's saying that the trait `Other` isn't implemented for `Field` when it is.

The error message might hint that this isn't working as expected because of the implicitly added named generics... not sure!

If we remove the associated constants from the code it works fine.

The issue also happens with associated types.

Found in #9167 because a trait constraint was starting to take effect in external repositories which then triggered this bug.

### Expected Behavior

### Bug

It doesn't compile.

### To Reproduce

1.
2.
3.
4.

### Workaround

Yes.

### Workaround Description

Use the second snippet, or the one in the comment.

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9168

## Summary

I'm unsure about this change but in the small code snippet for the test the only thing that was failing was the trait impl lookup. This is the code:

```noir
pub trait Other {
    let N: u32;
}
pub trait Trait<T>
where
    T: Other,
{}
impl Other for Field {
    let N: u32 = 1;
}
impl Trait<Field> for i32 {}
```

When analyzing `impl Trait<Field> for i32` we check if `Field: Other` we try to find an impl for `Field` that matches `Other<N = implicitly_added_named_generic>`. It will then try to unify that implicitly added named generic with `1` and fail.

Maybe an alternative is to replace those implicitly added named generics here with fresh type variables? In any case these bindings aren't committed in this lookup....

## Documentation

Check one:

# PR Checklist


asterite:
Closed in favor of #9352

PR #9352, opened 2025-07-29
Ab/trait parent check 9168 alternative

asterite:
# Description

## Problem

Resolves https://github.com/noir-lang/noir/issues/9168

Alternative to #9350

## Summary

I'm also unsure about this, but more sure than the implementation in #9350 as this change is localized to this check instead of to every trait implementation lookup.

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1396,13 +1396,28 @@ impl<'context> Elaborator<'context> {
             let trait_constraint_type = trait_constraint.typ.substitute(&bindings);
             let trait_bound = &trait_constraint.trait_bound;
 
+            // Replace implicitly added named generics with fresh type variables
+            let named_generics = trait_bound.trait_generics.named.iter();
+            let named_generics = named_generics.map(|named_type| {
```
jfecher: This is a bit lengthy - can you make a helper method for it instead?

compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1396,13 +1396,28 @@ impl<'context> Elaborator<'context> {
             let trait_constraint_type = trait_constraint.typ.substitute(&bindings);
             let trait_bound = &trait_constraint.trait_bound;
 
+            // Replace implicitly added named generics with fresh type variables
```
jfecher: Should mention here that the impl uses implicitly added named generics to represent its associated types and we need to replace them if they are unbound because they cannot be bound over until monomorphization, so they would not match anything otherwise. If they are bound we do not replace them because we want to ensure the binding matches.

compiler/noirc_frontend/src/elaborator/mod.rs
asterite: Done! I ended up with how it is because of the borrow checker (I can't pass `&trait_bound.trait_generics.named` to a helper function as it needs to take `&mut self` to create a new type variable, but `trait_bound` is a reference from self).

compiler/noirc_frontend/src/elaborator/mod.rs
asterite: Done!

PR #9353, opened 2025-07-29
chore: remove unused `compile_workspace`

asterite:
# Description

## Problem

No issue.

## Summary

I noticed that this function is unused. Can it be removed? I don't know if it was intended to be used as an API though.

## Documentation

Check one:

# PR Checklist


asterite:
It seems the `compile_workspace` removed here just type-checks the programs and contracts, monomorphizing, etc., but doesn't write the resulting programs and contracts to disk. The other one that remains does do that, so they are a bit different.

PR #9354, opened 2025-07-29
fix: push definition trait constraints after trait item constraint

### LINKED ISSUE #9349: Object type unknown in method call when trait generic has bound
### Aim

This code:

```noir
trait BigNum {}

trait BigCurve<B>
where
    B: BigNum,
{
    fn new() -> Self;
}

pub fn foo<B: BigNum, Curve: BigCurve<B>>() {
    let _: Curve = BigCurve::new();
}

fn main() {}
```

gives this error:

```
error: Object type is unknown in method call
   ┌─ src/main.nr:11:20
   │
11 │     let _: Curve = BigCurve::new();
   │                    ------------- Type must be known by this point to know which method to call
   │
   = Try adding a type annotation for the object type before this method call
```

However, if we remove the `where` clause:

```noir
trait BigNum {}

trait BigCurve<B> {
    fn new() -> Self;
}

pub fn foo<B: BigNum, Curve: BigCurve<B>>() {
    let _: Curve = BigCurve::new();
}

fn main() {}
```

the code compiles fine.

Note that both snippets compile fine in Rust.

This is blocking #9168 because once we fix the fact that `trait Trait<B: Constraint>` doesn't translate to `trait Trait<B> where B: Constraint` we then start getting the above error in the bigcurve library.

### Expected Behavior

The first snippet should compile fine.

### Bug

The first snippet doesn't compile.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9349

## Summary

I still don't fully understand how everything is tied together, but [this comment](https://github.com/noir-lang/noir/issues/9349#issuecomment-3133303723) from @jfecher made me wonder if it's indeed an ordering issue.

Inside `type_check_variable_with_bindings` we potentially push these constraints:
1. a constraint for the trait method itself
2. constraints for any where clause on the method

Before this PR we were pushing 2 before 1. There was also a comment about why 1 needs to come before 2, though since that comment was written the `select_impl` argument was added so I think that comment became obsolete at that point (since the constraints for the where clause won't lead to selecting a different impl).

Specifically, for the concrete example in the test we got that the type of the `BigCurve::new` call is:

```
fn() -> '7
```

Then we have this constraint for point 1:

```
'7: BigCurve<'6>
```

and this one for point 2:

```
'6: BigNum
```

We were checking point 2 before point 1, and it was failing because `'6` was unbound.

When we flip the order of operations, when we solve `'7: BigCurve<'6>` we get these type bindings applied:

```
{
    '6: (
        '6,
        Normal,
        B'4,
    ),
}
```

At that point `'6` is `B'4` (which I guess has an assumed impl for `BigNum`, according to the function signature), then the constraint `'6: BigNum` will pass.

## Additional Context

Thank you @jfecher ! I merely tried flipping the order here, the debugged things a bit more for the above explanation (it was hard to see that the bindings produced by `verify_trait_constraint` would make the second check work because when it's flipped we don't get to check the other one). The conclusion is that you solved it, I merely sent the code change 😅 

(well, ~we still have to see if CI passes...~ I do wonder if flipping the order would cause a problem in other scenarios)

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/elaborator/patterns.rs
```diff
@@ -1004,6 +1004,22 @@ impl Elaborator<'_> {
         let (typ, bindings) =
             self.instantiate(t, bindings, generics, function_generic_count, location);
 
+        if let ImplKind::TraitItem(mut method) = ident.impl_kind {
```
jfecher, line 1007: Can you add a comment either above this if or above line 1025 explaining why this constraint needs to be pushed before any subsequent ones?

asterite: Done! I used the same code for the example to try to explain it. Let me know if it's fine. I feel like it could be explained better, but I'm not sure how.

In any case I think because of the new regression test the order can't be reverted back.

Issue #9097, opened 2025-07-02
Bug: value is not typed as brillig usize when using `--show-ssa`


asterite:
### Aim

This program:

```noir
fn main() {
    // Safety:
    let result = unsafe { foo([0]) };
    println(result);
}

unconstrained fn foo(bytes: [u8; 1]) -> Field {
    for _ in 0..bytes.len() {
        let _ = bytes[0];
    }
    0
}
```

Compiles fine, but when it's compiled with `--show-ssa` it panics:

```noir
The application panicked (crashed).
Message:  value is not typed as brillig usize: true: u1
Location: /Users/asterite/Projects/noir/acvm-repo/brillig_vm/src/memory.rs:62
```

It might be that ID normalization is breaking something, because when normalization isn't done it works fine.

### Expected Behavior

The code should compile fine with or without `--show-ssa`.

### Bug

The code fails to compile with `--show-ssa`

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

asterite:
One thing I'm seeing is that function purity is lost when we normalize IDs.

PR #9355, opened 2025-07-29
fix: perserve purities after SSA normalization

asterite:
# Description

## Problem

Found while trying to solve #9097 but isn't the solution for that.

## Summary

Function purities are an `Arc` value that's set to every function dfg's in the SSA. The purities map a function ID to its purity.

In SSA normalization we traverse existing functions and create new functions which have value IDs that start from zero, then set those new functions as the SSA functions. For example we could have two functions with IDs 1 and 2 (which ended up being the only reachable ones), then after normalization we'll end up with functions 0 and 1. If we had a purity for the original function 1 we'll end up with that purity in the new function 1, which is actually the old function 2, which is incorrect.

Here we build a new purities map and use that going forward.

## Additional Context

This also fixes a tiny bug regarding function purities: in `defunctionalize` some pure dummy functions are created, but this purity is only tracked in those functions associated purities. This PR fixes that so that these purities are set to all functions.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/normalize_value_ids.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/normalize_value_ids.rs
@@ -61,8 +61,6 @@ impl Context {
         else {
             return;
         };
-            return;
-        };
         let mut new_purities = HashMap::default();
 
         for (id, function) in functions {


compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
```diff
@@ -1,4 +1,4 @@
-//! This module defines the defunctionalization pass for the SSA IR.
+//! This module defines the defunctionalization iass for the SSA IR.
```
jfecher:
```suggestion
//! This module defines the defunctionalization pass for the SSA IR.
```

compiler/noirc_evaluator/src/ssa/opt/normalize_value_ids.rs
```diff
@@ -57,12 +57,29 @@ struct IdMaps {
 
 impl Context {
     fn populate_functions(&mut self, functions: &BTreeMap<FunctionId, Function>) {
+        if functions.is_empty() {
+            return;
+        }
+
+        let old_purities = &functions.iter().next().unwrap().1.dfg.function_purities;
```
jfecher:
```suggestion
        let Some(old_purities) = &functions.iter().next().map(|f| &f.1.dfg.function_purities) else {
            return;
        };
```
The lone `is_empty` check just felt a bit odd to me when I think every program should have an entry point at least.

compiler/noirc_evaluator/src/ssa/opt/normalize_value_ids.rs
```diff
@@ -57,12 +57,27 @@ struct IdMaps {
 
 impl Context {
     fn populate_functions(&mut self, functions: &BTreeMap<FunctionId, Function>) {
+        let Some(old_purities) = &functions.iter().next().map(|f| &f.1.dfg.function_purities) else {
```
asterite:
```suggestion
        let Some(old_purities) = &functions.iter().next().map(|f| &f.1.dfg.function_purities)
        else {
            return;
        };
```

compiler/noirc_evaluator/src/ssa/opt/normalize_value_ids.rs
```diff
@@ -57,12 +57,30 @@ struct IdMaps {
 
 impl Context {
     fn populate_functions(&mut self, functions: &BTreeMap<FunctionId, Function>) {
+        let Some(old_purities) = &functions.iter().next().map(|f| &f.1.dfg.function_purities)
+        else {
+            return;
+        };
+            return;
+        };
```
asterite:
```suggestion
        else {
            return;
        };
```

PR #9357, opened 2025-07-30
chore: store function purities on the SSA

asterite:
# Description

## Problem

Made on top of #9355, just trying something.

## Summary

This is just an experiment to see how the code looks like if we avoid storing function purities on every function dfg and instead store it once in the SSA and pass it around as needed.

I'm not 100% this is worth it, but it does seem to make the code a bit easier to follow (we can't forget to set purities in some functions, etc.), so I'm mainly interested in your opinion.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
@@ -1,4 +1,4 @@
-//! This module defines the defunctionalization iass for the SSA IR.
+//! This module defines the defunctionalization pass for the SSA IR.
 //! Certain IR targets (e.g., Brillig and ACIR) do not support higher-order functions directly.
 //!
 //! The pass eliminates higher-order functions (a function which accepts function values as arguments or returns functions)


compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -126,9 +127,10 @@ impl Function {
     }
 }
 
-struct Context<'a> {
+struct Context<'brillig, 'purities> {
+    purities: &'purities FunctionPurities,
     use_constraint_info: bool,
-    brillig_info: Option<BrilligInfo<'a>>,
+    brillig_info: Option<BrilligInfo<'brillig>>,
```
TomAFrench: I think we can get away with the single lifetime here.

compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
```diff
@@ -1,4 +1,4 @@
-//! This module defines the defunctionalization pass for the SSA IR.
+//! This module defines the defunctionalization iass for the SSA IR.
```
TomAFrench:
```suggestion
//! This module defines the defunctionalization pass for the SSA IR.
```

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
asterite: Sorry, what do you mean?

jfecher: Presumably that the additional `'purities` lifetime is not necessary here

asterite: Thanks, done!

TomAFrench:
This looks pretty stale at this point. @asterite shall we close this or is it worth reviving?

asterite:
Let's close this. We could always revive it later. I feel like it might be worth it because there was at least one bug where we forgot to copy purities.

PR #9358, opened 2025-07-30
chore: enhance trait constraint comment

asterite:
# Description

## Problem

Follow up to #9354

## Summary

Yesterday I didn't feel satisfied with the comment I wrote. Apparently my brain kept thinking about this and now I understand a bit more how things work, so I decided to write it so things are clearer.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_frontend/src/elaborator/patterns.rs
+++ b/compiler/noirc_frontend/src/elaborator/patterns.rs
@@ -1035,7 +1035,7 @@ impl Elaborator<'_> {
         //     fn new() -> Self;
         // }
         //
-        // fn foo<O: One, T: Two<O>>() {
+        // fn foo<X: One, T: Two<X>>() {
         //     let _: T = Two::new();
         // }
         // ```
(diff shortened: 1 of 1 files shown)


compiler/noirc_frontend/src/elaborator/patterns.rs
```diff
@@ -1035,16 +1035,21 @@ impl Elaborator<'_> {
...
         //     let _: T = Two::new();
         // }
         // ```
         //
         // when type-checking `Two::new` we'll have a return type `'2` which is constrained by `'2: Two<'1>`.
         // Then the definition for `new` has a constraint on it, `O: One`, which translates to `'1: One`.
         //
-        // If we try to find a trait implementation for `'1` before finding one for `'2` we'll never find it:
-        // once we find one for `'2`, `'1` will become bound and a trait implementation will be found.
+        // Becuase of the explicit type in the `let`, `'2` will be unified with `T`.
```
asterite:
```suggestion
        // Because of the explicit type in the `let`, `'2` will be unified with `T`.
```

PR #9360, opened 2025-07-30
fix: allow constants in the type-system to be negative

### LINKED ISSUE #8649: Associated constants can't be negative
### Aim

I don't know if this is intentional. If so, there are a few issues with how it's currently implemented.

```noir
trait Foo {
    let N: i32;
}

impl Foo for i32 {
    let N: i32 = -1;
}
```

gives this error:

```
error: Invalid type expression: '-1'. Only unsigned integer constants up to `u32`, globals, generics, +, -, *, /, and % may be used in this context.
  ┌─ src/main.nr:6:18
  │
6 │     let N: i32 = -1;
  │                  --
```

On the other hand this code gives a different error:

```noir
trait Foo {
    let N: i32;

    fn n() -> i32 {
        Self::N
    }
}

impl Foo for i32 {
    let N: i32 = 1 - 10;
}

fn main() {
    println(i32::n());
}
```

It says:

```
error: Could not determine the value of associated constant `N`, encountered error: `The value `-9` cannot fit into `numeric i32` which has a maximum size of `2147483647``
  ┌─ src/main.nr:5:9
  │
5 │         Self::N
  │         -------
```

which is a bit misleading. It **could** determine the value, and it actually fits the range of `i32`.

### Expected Behavior

Not sure.

### Bug

A misleading error message, and it's unclear why signed numbers are banned from associated constants.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #8649

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/elaborator/enums.rs
```diff
@@ -795,7 +795,7 @@ impl Elaborator<'_> {
 
         let value = match constant {
             Value::Bool(value) => SignedField::positive(value),
-            Value::Field(value) => SignedField::positive(value),
```
asterite, line 798: This was a bug. Because `SignedField` is `impl Into<FieldElement>` it was transformed into a `FieldElement`, then that was sent to `positive`. The problem with that is that if we have a `SignedField(1, negative: true)` it's going to be passed as the largest field value into positive, effectively changing the meaning of that value (it's the same value for field elements but it could be a different one if we then need to convert that to another type).

That's why I removed `impl From<SignedField> for FieldElement`, which wasn't used in a lot of places (and most of the places it was used it was incorrect like here).

compiler/noirc_frontend/src/signed_field.rs
```diff
@@ -39,6 +39,14 @@ impl SignedField {
         self.is_negative
     }
 
+    pub fn is_zero(&self) -> bool {
+        self.field.is_zero()
+    }
+
+    pub fn is_one(&self) -> bool {
+        self.field.is_one()
```
jfecher:
```suggestion
        self.field.is_one() && !self.is_negative
```

compiler/noirc_frontend/src/signed_field.rs
```diff
@@ -183,15 +199,33 @@ impl PartialOrd for SignedField {
...
-        if value.is_negative { -value.field } else { value.field }
+impl From<u128> for SignedField {
+    fn from(value: u128) -> Self {
+        Self::new(value.into(), false)
+    }
+}
+
+impl From<i128> for SignedField {
+    fn from(value: i128) -> Self {
+        if value < 0 { Self::new((-value).into(), true) } else { Self::new(value.into(), false) }
```
jfecher, line 222: This will panic when `value = -128`

compiler/noirc_frontend/src/signed_field.rs
```diff
@@ -98,6 +106,14 @@ impl SignedField {
     pub fn to_field_element(self) -> FieldElement {
         if self.is_negative { -self.field } else { self.field }
     }
+
+    pub fn to_u128(self) -> u128 {
+        self.to_field_element().to_u128()
+    }
+
+    pub fn to_i128(self) -> i128 {
+        self.to_field_element().to_i128()
```
jfecher: Does this handle negative values?

jfecher (commented):
Good catch with the conversion bug

asterite:
> Good catch with the conversion bug

I actually implemented things first and they compiled, but the program in #8649 errored during SSA-gen. I eventually figured it out but I was forced to find it 😅

compiler/noirc_frontend/src/signed_field.rs
asterite: Good catch!

compiler/noirc_frontend/src/signed_field.rs
asterite: Good catch! (I guess you meant `i128::MIN`, not sure! I added a couple of tests)

asterite: Ah, no, I was thinking about `to_i128`, nevermind...

compiler/noirc_frontend/src/signed_field.rs
asterite: No, but now it does. I added a couple of tests. Thanks!

compiler/noirc_frontend/src/signed_field.rs
jfecher: Oops, yes I saw `i128` and thought `i8::MIN = -128` for some reason, but yes the minimum value.

asterite: Was this fine? I'll add a few tests for it tomorrow just in case.

PR #9365, opened 2025-07-30
fix(ssa): Constant fold Brillig calls using the SSA interpreter

### LINKED ISSUE #9015: Consider replacing usage of the brillig VM in `fold_constants_with_brillig` with the SSA interpreter
We currently perform brillig gen halfway through the SSA pipeline so that we can continue processing some ACIR functions further while having the ability to execute any brillig functions for which the inputs are known at compile time.

We recently added the [SSA interpreter](#6826) which allows us to execute the SSA IR directly. This would allow us to avoid doing early brillig generation and just interpret the SSA in any situation where we're currently running the brillig bytecode.

### LINKED ISSUE #9097: Bug: value is not typed as brillig usize when using `--show-ssa`

vezenovm:
# Description

## Problem\*

Resolves #9015 
Resolves #9097 

This also resolves a panic from a [Noir sync](https://github.com/AztecProtocol/aztec-packages/pull/15899) when attempting to debug log an array value (the same error as #9097)

Let's take the snippet from #9097:
```noir
fn main() {
    // Safety:
    let result = unsafe { foo([0]) };
    println(result);
}

unconstrained fn foo(bytes: [u8; 1]) -> Field {
    for _ in 0..bytes.len() {
        let _ = bytes[0];
    }
    0
}
```

We can see the following SSA when we fold constants with Brillig: 
<p>

```
After Inlining Brillig Calls (1) (step 42):
  b0():
    v1 = make_array [u8 0] : [u8; 1]                    // src/main.nr:3:32
    call f2(u1 1, Field 0)                              // std/lib.nr:40:9
    return
}
  b0(v0: [u8; 1]):
    inc_rc v0                                           // src/main.nr:8:17
    return Field 0
}
brillig(inline) impure fn print_unconstrained f2 {
  b0(v0: u1, v1: Field):
    v13 = make_array b"{\"kind\":\"field\"}"            // std/lib.nr:40:9
    call print(v0, v1, v13, u1 0)                       // std/lib.nr:34:5
    return
}

Remove Unreachable Instructions (2) (step 43): 0 ms
After Remove Unreachable Instructions (2) (step 43):
  b0():
    v1 = make_array [u8 0] : [u8; 1]                    // src/main.nr:3:32
    call f1(u1 1, Field 0)                              // std/lib.nr:40:9
    return
}
brillig(inline) predicate_pure fn print_unconstrained f1 {
  b0(v0: u1, v1: Field):
    v13 = make_array b"{\"kind\":\"field\"}"            // std/lib.nr:40:9
    call print(v0, v1, v13, u1 0)                       // std/lib.nr:34:5
    return
}
```

We can see that one Brillig function was folding into `f0`. However, the print remains. However, the Brillig context structure contains all the artifacts already has a set [mapping for function ids to Brillig artifacts](https://github.com/noir-lang/noir/blob/7d48355fa729a7e3d862e7afb193494c6a9d5aed/compiler/noirc_evaluator/src/brillig/mod.rs#L51). Thus, when we remove functions after Brillig gen and change their IDs (e.g., like when normalizing) we end up fetching an artifact for the wrong function when it comes times to generate an entry point. So we were fetching 

</p>
</details> 

## Summary\*

I realized this could have been fixed for #9097 by simply updating the IDs point to the Brillig artifacts as part of normalization. However, this felt brittle and I felt the problem laid out above still feels like it could come up elsewhere. 

Aside the benefit of avoiding extra Brillig compilation, switching to using the SSA interpreter for constant folding Brillig entry points also avoids these kinds of situations where Brillig context needs to be updated due to secondary SSA passes post Brillig gen. 

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
> **regression_7451**	+2 ❌	**+200.00%**	+2,848 ❌	**+16752.94%**
> **inline_decompose_hint_brillig_call**	+15 ❌	**+1500.00%**	+74 ❌	**+435.29%**
> **strings**	+1 ❌	**+3.45%**	+3 ❌	**+9.09%**

Hmm looks like we are getting some regressions. I was not expecting any regression so I will have to investigate before we can merge.

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
```diff
@@ -49,6 +49,8 @@ struct Interpreter<'ssa, W> {
 pub struct InterpreterOptions {
     /// If true, the interpreter will trace its execution.
     pub trace: bool,
+    /// If true, the interpreter treats calls all foreign function calls (e.g., `print`) as unknown
```
asterite:
```suggestion
    /// If true, the interpreter treats all foreign function calls (e.g., `print`) as unknown
```

compiler/noirc_evaluator/src/ssa/opt/constant_evaluation.rs
```diff
@@ -0,0 +1,474 @@
...
+        instruction::{Instruction, InstructionId},
+        types::Type,
+        value::{Value as IrValue, ValueId},
+    },
+    ssa_gen::Ssa,
+};
+
+impl Ssa {
+    /// See [`fold_constant_brillig_calls`][self] module for more information.
+    pub(crate) fn fold_constant_brillig_calls(mut self) -> Ssa {
```
asterite: Is the name of this function okay? If so, the doc comment has a broken link (not sure if it should be `constant_evaluation` or `constant_folding`)

vezenovm: Ah I this doc link is broken. I originally had the name `constant_evaluation` but changed it to make it more specific to what it does. Do you prefer this name or `constant_evaluation`? I would change the module to `constant_folding_brillig_calls` otherwise.

compiler/noirc_evaluator/src/ssa/opt/constant_evaluation.rs
```diff
@@ -0,0 +1,474 @@
...
+            .functions
+            .iter()
+            .map(|(id, function)| (*id, function.constant_evaluation(&self)))
+            .collect::<HashMap<_, _>>();
+
+        // Apply constant folding for each function based on the interpreter results
+        for function in self.functions.values_mut() {
+            let Some(constant_evaluations) = constant_evaluations.remove(&function.id()) else {
+                continue;
+            };
```
asterite: Is there a case where this returns `None`? Based on the above loop it seems every function ID will get a HashMap, or at least an empty one.
(maybe `constant_evaluation` could return `Option<HashMap<..>>`?)

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -286,6 +280,7 @@ impl Context {
                 // side effects ordering remains correct.
                 if function.runtime().is_acir()
                     && insert_out_of_bounds_checks
+                // if insert_out_of_bounds_checks
```
asterite: Extra comment

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
vezenovm: I have this removed locally so resolving

compiler/noirc_evaluator/src/ssa/opt/die.rs
vezenovm: locally removed

compiler/noirc_evaluator/src/ssa/opt/constant_evaluation.rs
asterite: I think `constant_folding_brillig_calls` is better, as you say.

compiler/noirc_evaluator/src/ssa/opt/constant_evaluation.rs
vezenovm: We are already returning an `Option` by fetching from a map. Yes, there shouldn't be a case where this is `None`. I could panic here if we feel that is safer. I just wanted to avoid a panic because if there isn't a value in `constant_evaluations` we don't break correctness by not attempting to constant fold the call.

compiler/noirc_evaluator/src/ssa/mod.rs
```diff
@@ -183,6 +180,9 @@ pub fn primary_passes(options: &SsaEvaluatorOptions) -> Vec<SsaPass> {
             // Remove any potentially unnecessary duplication from the Brillig entry point analysis.
             .and_then(Ssa::remove_unreachable_functions),
         SsaPass::new(Ssa::remove_truncate_after_range_check, "Removing Truncate after RangeCheck"),
+        SsaPass::new(Ssa::fold_constant_brillig_calls, "Constant Fold Brillig Calls")
+            .and_then(Ssa::remove_unreachable_functions),
+        SsaPass::new(Ssa::fold_constants, "Constant Folding"),
```
asterite: I _think_ repeating these two passes again fixes the regression issues (or at least for `array_oob_regression_7965` it leads back to the original final SSA).

It might be because the two passes were just one before, so while folding constants we'd also evaluate brillig calls and use that in the folding.

Not sure if it would make sense to merge the two passes again or do something else.

asterite: I also noticed that in secondary passes were were doing "remove unreachable instructions" + DIE and that's gone now, not sure if it's needed too.

vezenovm: Yeah I'm experimenting with the ordering of the passes here. Trying to find the right balance as I don't want to repeat passes if we can avoid it. 

> "remove unreachable instructions"

Ah I forgot we were doing this though. I will try to add that back before the final DIE.

vezenovm: Regular constant folding before `fold_constant_brillig_calls ` solved regressions for `array_oob_regression_7965`. However, for some of the other regressions it was due to not tracking the constant results of folded Brillig calls. For example if we a Brillig call whose constant result was used as the input to another Brillig call.

compiler/noirc_evaluator/src/ssa/opt/constant_folding_brillig_calls.rs
```diff
@@ -0,0 +1,526 @@
...
+                    unreachable!("ICE: Function {func_id} does not exist, yet it is being called");
+                };
+
+                // Skip calls to non-Brillig functions
+                if !func.runtime().is_brillig() {
+                    continue;
+                }
+
+                let mut all_constants = true;
+                let mut interpreter_args = Vec::new();
```
asterite: Not sure how it affects performance but I wonder if splitting this into two things:
1. check if all arguments are constant
2. creating the `interpreter_args` array

would be better. For example if we have a brillig call with a large constant array as the first argument but a dynamic second argument we'd go through `const_ir_value_to_interpreter_value`, etc.
This is a minor thing, though.

compiler/noirc_evaluator/src/ssa/mod.rs
asterite: Looks good!

I see there's still a tiny regression. I would prefer if this constant evaluation happened inside `constant_folding`, like before, but I see that it would be pretty hard to do it as at that point we have a `&mut Function` reference so I think we wouldn't be able to also have access to a `&Ssa` to interpret functions.

There was a recent Slack thread about sha512 not performing well so not sure if this regression will be impactful.

In any case I agree with merging this as it's a bug fix.

asterite (approved):
Looks good!

compiler/noirc_evaluator/src/ssa/opt/constant_folding_brillig_calls.rs
vezenovm: That is a good point. I'll switch to that.

vezenovm:
> **sha512_100_bytes**	+221 ❌	**+1.68%**	+443 ❌	**+1.12%**

Looks like we still have a minor regression here. Most likely due to us not folding Brillig calls as we constant fold all the other instructions. I'm going to look into a fixpoint loop quickly to see if it removes the regression. Otherwise, I'm going to merge as this regression is small and we in fact see improvements in other tests. Optimizing this can then come in a separate PR as it would most likely require changing the interpreter interface to not borrow `Ssa` so that we can evaluate calls during the main `constant_folding` pass.

compiler/noirc_evaluator/src/ssa/opt/constant_folding_brillig_calls.rs
vezenovm: done

vezenovm:
I ended up making a fixed point loop in `constant_folding`. The loop is contained in a top-level method in `constant_folding` and `fold_constant_brillig_calls` is now a submodule of `constant_folding`.

compiler/noirc_evaluator/src/ssa/mod.rs
vezenovm: > I would prefer if this constant evaluation happened inside `constant_folding`, like before, but I see that it would be pretty hard to do it as at that point we have a `&mut Function` reference so I think we wouldn't be able to also have access to a `&Ssa` to interpret functions.

Sorry missed this comment earlier. Yes, I was initially hoping for the same but was prevented by us needing to borrow the `Ssa` object for the interpreter. We could change the interpreter interface but this feels like it still may require a lot of cloning and we would probably have to do some analysis to make sure we aren't cloning a bunch of unnecessary information. It felt like a lot for this bug fix. I did switch to the constant evaluation being a submodule of `constant_folding` and using it in a fixed point loop. https://github.com/noir-lang/noir/pull/9365#issuecomment-3141314899

vezenovm:
> I ended up making a fixed point loop in `constant_folding`. The loop is contained in a top-level method in `constant_folding` and `fold_constant_brillig_calls` is now a submodule of `constant_folding`.

ugh, we seem to be getting large time regressions in the sha512 library https://github.com/noir-lang/noir/actions/runs/16660549208/job/47156670574?pr=9365. Perhaps we do just have to bite the bullet and change the interpreter interface.

compiler/noirc_evaluator/src/ssa/mod.rs
vezenovm: > There was a recent Slack thread about sha512 not performing well so not sure if this regression will be impactful.

The regression is only ~1-2% which isn't horrible. The previous performance regression was much greater that was resolved here https://github.com/noir-lang/noir/pull/9200#issuecomment-3070767165.

vezenovm:
> ugh, we seem to be getting large time regressions in the sha512 library https://github.com/noir-lang/noir/actions/runs/16660549208/job/47156670574?pr=9365. Perhaps we do just have to bite the bullet and change the interpreter interface.

If we want to get this fix in and are ok with a [small circuit size regression](https://github.com/noir-lang/noir/pull/9365#issuecomment-3141314899) I will just revert to [e83fe33](https://github.com/noir-lang/noir/pull/9365/commits/e83fe33289ab9bc4cc962b02d29edad894345a8a) where everything is green. Otherwise, I will take a look at how we can change the interpreter interface to avoid borrowing the entire `Ssa` for this case.

compiler/noirc_evaluator/src/ssa/opt/constant_folding/brillig_calls.rs
```diff
@@ -0,0 +1,532 @@
+//! Evaluates calls to Brillig functions at compile time when all arguments are constant.
+//!
+//! This acts as a partial evaluator for Brillig entry points called from ACIR functions.
+//! For any Brillig function call where all arguments are constant, we interpret the
+//! function using the [SSA interpreter][crate::ssa::interpreter] and replace the call with the resulting constants.
+//!
+//! This improves constant folding across function boundaries, particularly for small utility
+//! Brillig functions that can be inlined
+//!
+//! Note: Only calls to Brillig functions from ACIR functions (Brillig entry points) are evaluated here.
```
TomAFrench, line 10: This doesn't seem correct based on the changes to brillig bytecode sizes shown in the benchmarls.

vezenovm:
Superseded by #9655

PR #9369, opened 2025-07-31
fix: trait where clause check fixes

### LINKED ISSUE #7024: Trait definition where clause doesn't support associated type constraints

### LINKED ISSUE #9367: Trait bound not satisfied when constraint refers to two generics
### Aim

I wasn't sure what title to use.

This code:

```noir
pub trait Foo<U> {}

pub trait Baz<T, U>
where
    T: Foo<U>,
{}

pub struct HasFoo1 {}
impl Foo<()> for HasFoo1 {}

pub struct HasBaz1 {}
impl Baz<HasFoo1, ()> for HasBaz1 {}

fn main() {}
```

Gives this error:

```
error: The trait bound `HasFoo1: Foo<U>` is not satisfied
   ┌─ src/main.nr:5:8
   │
 5 │     T: Foo<U>,
   │        --- required by this bound in `Baz`
   ·
12 │ impl Baz<HasFoo1, ()> for HasBaz1 {}
   │                           ------- The trait `Foo<U>` is not implemented for `HasFoo1`
```

However, that code compiles fine in Rust and it should also compile fine in Noir because `HasFoo: Foo<()>` holds because of the first impl.

### Expected Behavior

The code should compile.

### Bug

It doesn't compile.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9367
Resolves #7024

### First issue

The first issue has this code:

```noir
pub trait Foo<U> {}

pub trait Baz<T, U>
where
    T: Foo<U>,
{}

pub struct HasFoo1 {}
impl Foo<()> for HasFoo1 {}

pub struct HasBaz1 {}
impl Baz<HasFoo1, ()> for HasBaz1 {}

fn main() {}
```

- For `impl Baz<HasFoo1, ()> for HasBaz1 {}` we'll need to see if `T: Foo<U>` holds, where `T` is `HasFoo1` and `U` is `()`.
- The constraint to solve, when debugging this, is `T'3: Foo<U'4>`
- We can also see that after binding the ordered generics in `Baz<HasFoo1, ()>` we end up with these bindings: `'4: (), '3: HasFoo1`
- We were correctly applying bindings to the constraint type (`T`) so it became `HasFoo1`, but we didn't do so for the entire constraint, meaning that the constraint remained as `HasFoo1: Foo<U'4>` and failed as there is no such generic impl.
- The fix was to also apply bindings (well, substitute) to the trait bound (and generally to the entire constraint) so that the constraint becomes `HasFoo1: Foo<()>` which now holds.

### Second issue

This is the code:

```noir
pub trait BarTrait {}

pub trait Foo {
    type Bar;
}

pub trait Baz<T>
where
    T: Foo,
    <T as Foo>::Bar: BarTrait,
{}

pub struct HasBarTrait1 {}
impl BarTrait for HasBarTrait1 {}

pub struct HasFoo1 {}
impl Foo for HasFoo1 {
    type Bar = HasBarTrait1;
}

pub struct HasBaz1 {}
impl Baz<HasFoo1> for HasBaz1 {}

fn main() {}
```

- Here we start by checking `Baz<HasFoo1>`. For that we need to check that `T: Foo` and `<T as Foo>::Bar: BarTrait` hold.
- We first check `T'4: Foo` where we having `'4: HasFoo1` in `bindings`, so we end up checking `HasFoo1: Foo` which succeeds
- Actually, because `Foo` has an associated type we check `HasFoo1: Foo<Bar = <T as Foo>::Bar'6>` where `<T as Foo>::Bar'6` is an implicitly added named generic
- In a recent PR we started replacing these implicitly added named generics with type variables so that the trait impl lookup succeeds. In this case we end up replacing `<T as Foo>::Bar'6` with `'10`, so we are really checking `HasFoo1: Foo<Bar = '10>`, which succeeds.
- Next we check the next constraint, `<T as Foo>::Bar'6: BarTrait`. The problem here is that this will fail as the constraint type is an unbound named generic. It's actually the implicitly added named generic!
- The solution here is to first remember that we replaced `<T as Foo>::Bar'6` with `'10`. We do this by storing it in `bindings`: `'6: '10`.
- Next, after we check `HasFoo1: Foo<Bar = '10>`, `try_lookup_trait_implementation` will return these bindings: `'10: HasBarTrait1`. We were ignoring these returned bindings but not anymore: we add them to the existing `bindings`.
- Finally, when we need to solve `<T as Foo>::Bar'6: BarTrait` we apply bindings. Because we have `'6: '10, '10: HasBarTrait1` we end up solving `HasBarTrait1: BarTrait` which now holds! 🎉 

Sorry for the long explanation but since I'm starting to understand how bindings work I wanted to convince myself that the code I'm writing makes sense.

## Additional Context

As a bonus, if the where clause has this order:

```noir
where
    <T as Foo>::Bar: BarTrait,
    T: Foo,
```

it would fail in Noir but actually succeed in Rust. I don't know how they do it but I implemented something for that here: we reorder the where clause so that "simpler" constraint types come before more complex ones. In this case `T` is simpler than `<T as Foo>::Bar` so we end up putting before it, and then it works.

## Documentation

Check one:

# PR Checklist


compiler/noirc_frontend/src/elaborator/mod.rs
```diff
@@ -1407,43 +1410,56 @@ impl<'context> Elaborator<'context> {
...
-                });
+            match self.interner.try_lookup_trait_implementation(
+                &trait_constraint_type,
+                trait_bound.trait_id,
+                &trait_bound.trait_generics.ordered,
+                &named_generics,
+            ) {
+                Ok((_, impl_bindings, impl_instantiation_bindings)) => {
+                    bindings.extend(impl_bindings);
+                    bindings.extend(impl_instantiation_bindings);
```
asterite, line 1430: I'm not sure about also using `impl_instantiation_bindings` but given that ignoring them in the past usually meant a bug somewhere, and also because these end up just being local bindings in this Rust function, I think it should be fine.

compiler/noirc_frontend/src/hir/def_collector/dc_mod.rs
```diff
@@ -517,6 +521,7 @@ impl ModCollector<'_> {
             for item in &mut trait_definition.items {
                 if let TraitItem::Function { generics, where_clause, .. } = &mut item.item {
                     desugar_generic_trait_bounds(generics, where_clause);
+                    reorder_where_clause(where_clause);
```
jfecher: I see a lot of these checks - would it be easier to add a constructor for where clauses instead and do this once there?

jfecher (approved):
:tada:

compiler/noirc_frontend/src/hir/def_collector/dc_mod.rs
asterite: I thought about it but `desugar_generic_trait_bounds` will transform a line like `fn foo<T: Trait>(x: T) where <T as Foo>::Bar: Baz` to `fn foo<T>(x: T) where <T as Foo>::Bar: Baz, T: Trait`. That is, we don't have the entire where clause when it's parsed.

I could inline `reorder_where_clause` inside `desugar_generic_trait_bounds`, though, it's always called right after that. Or maybe have another method that does both things. What do you think?

asterite: I went with a method that does both things.

PR #9373, opened 2025-08-01
chore!: new semantic for bit-shifts 

### LINKED ISSUE #9022: Shift-left inconsistency

guipublic:
# Description

## Problem\*

Resolves #9022 

## Summary\*
Bit-shifts now requires that rhs and lhs have the same type (instead of u8 for rhs). This is a breaking change.
If rhs > bit_size of rhs type (or bit_size -1 for signed types), then bit-shift will overflow.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
Looks like this is timing out all the debug tests locally

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -341,25 +340,12 @@ impl<'a> FunctionContext<'a> {
...
-                        if let Some(rhs_const) = dfg.get_numeric_constant(rhs) {
-                            let bit_shift_size = rhs_const.to_u128() as u32;
-
-                            if max_lhs_bits + bit_shift_size <= bit_size {
-                                // `lhs` has been casted up from a smaller type such that shifting it by a constant
-                                // `rhs` is known not to exceed the maximum bit size.
-                                return result;
-                            }
-                        }
-
```
TomAFrench, line 362: Why was this removed?

compiler/noirc_evaluator/src/ssa/ir/dfg/simplify/binary.rs
```diff
@@ -294,19 +292,8 @@ pub(super) fn simplify_binary(binary: &Binary, dfg: &mut DataFlowGraph) -> Simpl
...
-            if let Some(rhs_const) = rhs_value {
-                if rhs_const >= FieldElement::from(lhs_type.bit_size() as u128) {
-                    // Shifting by the full width of the operand type, any unsigned `lhs` goes to zero.
-                    if lhs_type.is_unsigned() {
-                        let zero = dfg.make_constant(FieldElement::zero(), lhs_type);
-                        return SimplifyResult::SimplifiedTo(zero);
-                    }
-                }
-                return SimplifyResult::SimplifiedToInstruction(simplified);
-            }
```
TomAFrench, line 309: Why was this removed?

test_programs/execution_success/shift_right_overflow/src/main.nr
```diff
@@ -1,5 +1,5 @@
 fn main(x: u8) {
-    // This would previously overflow in ACIR. Now it returns zero.
-    let value = 1 >> x;
+    // 1>>8 and 1>>9 would previously return zero, now they overflow.
+    let value = 1 >> 7;
```
TomAFrench, line 3: ~~I don't think this matches with what was discussed in slack: https://aztecprotocol.slack.com/archives/C0183F0V42V/p1753720749271979?thread_ts=1749745678.481559&cid=C0183F0V42V~~

~~I would expect this to return zero as 7 < 8 which is the bit size of the lhs type.~~

I can't read.

TomAFrench (commented):
~~I think we've got some undesirable behaviour here which means that `4u8>>2u8` is fine but `1u8>>2u8` is disallowed. I would expect the latter to result in zero.~~ I was misreading comments.

TomAFrench:
> | Program | Brillig opcodes (+/-) | % |
> |:-|-:|-:|
> | **bit_shifts_runtime_inliner_max** | +677 ❌ | **+298.24%** |
> | **bit_shifts_runtime_inliner_min** | +677 ❌ | **+298.24%** |
> | **bit_shifts_runtime_inliner_zero** | +677 ❌ | **+298.24%** |

Strange we're getting such a large regression here as while there's been a few changes to this program it shouldn't result in ~700 new opcodes afaict.

Edit: due to the removals above?

Edit Edit: We're now doing more runtime signed bitshifts so the overflow checks could be the cause of this.

compiler/noirc_evaluator/src/ssa/ir/dfg/simplify/binary.rs
TomAFrench: We probably want to just gate this behind a check to make sure it's an unsigned bitshift.

compiler/noirc_evaluator/src/ssa/interpreter/errors.rs
```diff
@@ -115,8 +115,8 @@ pub enum InternalError {
         "Invalid bit size of `{bit_size}` given to truncate, maximum size allowed for unsigned values is {MAX_UNSIGNED_BIT_SIZE}"
     )]
     InvalidUnsignedTruncateBitSize { bit_size: u32 },
-    #[error("Rhs of `{operator}` should be a u8 but found `{rhs_id} = {rhs}`")]
-    RhsOfBitShiftShouldBeU8 { operator: &'static str, rhs_id: ValueId, rhs: String },
+    #[error("Operands of `{operator}` should have the same type`")]
+    BitShiftShouldHaveSameType { operator: &'static str },
```
TomAFrench: This can be replaced with usage of `MismatchedTypesInBinaryOperator` now.

compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
```diff
@@ -286,10 +295,14 @@ pub(crate) fn convert_signed_integer_to_field_element(int: i128, bit_size: u32)
...
-        // We add an offset of `bit_size` bits to shift the negative values into the range [2^(bitsize-1), 2^bitsize)
-        assert!(bit_size < 128, "{bit_size} is too large");
-        let offset_int = (1i128 << bit_size) + int;
-        FieldElement::from(offset_int)
+        let two_complement = match bit_size {
+            8 => ((int as i8) as u8) as u128,
+            16 => ((int as i16) as u16) as u128,
+            32 => ((int as i32) as u32) as u128,
+            64 => ((int as i64) as u64) as u128,
+            _ => unreachable!("ICE - invalide bit size {bit_size} for signed integer"),
```
TomAFrench:
```suggestion
            _ => unreachable!("ICE - invalid bit size {bit_size} for signed integer"),
```

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -518,13 +515,229 @@ mod tests {
...
             v84 = mul v83, v79
             v85 = add v82, v84
-            v86 = mul v4, v85
-            v87 = cast v0 as Field
-            v88 = mul v87, v86
-            v89 = truncate v88 to 32 bits, max_bit_size: 254
-            v90 = cast v89 as u32
-            v91 = truncate v90 to 32 bits, max_bit_size: 33
-            return v90
+            v87 = array_get v7, index u32 23 -> u1
```
TomAFrench: This extra bit decomposition seems unnecessary. If any more than 8 bits are used then we're going to overflow the field.

compiler/noirc_evaluator/src/ssa/ir/instruction/binary.rs
guipublic: done

compiler/noirc_evaluator/src/ssa/interpreter/errors.rs
guipublic: done

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
guipublic: The overflow check is different now, it checks that rhs< bit_size(type of lhs), which will be known at constant time in this case. So handling this case here is not needed.

compiler/noirc_evaluator/src/ssa/ir/dfg/simplify/binary.rs
guipublic: If rhs is more than the bitesize of lhs type, then it is now an overflow, not zero.

jfecher (commented):
LGTM - looks like external repos are failing to build though

PR #9376, opened 2025-08-01
fix(ssa): Do not check ArrayGet/Set as unreachable for Brillig

vezenovm:
# Description

## Problem\*

Resolves a panic (` Assertion failed: Bit size for lhs 16 does not match op bit size 32`) from a [Noir sync](https://github.com/AztecProtocol/aztec-packages/pull/15899) that is currently blocking any more syncs.

## Summary\*

It looks like the Aztec end-to-end tests pass with the an ACIR runtime gate just on `Instruction::ArrayGet`/`Instruction::ArraySet`. I thought the core issue is most likely with the dummy array (thinking it was due to some mismatched types but I have not found the exact issue there yet). However, gating just the dummy array logic was not enough to fix the Brillig panic. Gating the entire array get/set check was the only thing that removed the panic. 

However, we can do a hot fix here by only allowing this optimization for ACIR.

I only gated `Instruction::ArrayGet` and `Instruction::ArraySet` in case we still wanted to mark unreachable instructions for failing constrains in Brillig. Looking back at the history of when this pass was created it does look like it was mainly made to avoid ACIR compilation errors. So perhaps for the sake of simplicity it would be better to simply gate the entire pass to only be performed on ACIR. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
@@ -30,9 +30,9 @@ use crate::ssa::{
 impl Ssa {
     pub(crate) fn remove_unreachable_instructions(mut self) -> Ssa {
         for function in self.functions.values_mut() {
-            if function.runtime().is_brillig() {
-                continue;
-            }
+            // if function.runtime().is_brillig() {
+            //     continue;
+            // }
             function.remove_unreachable_instructions();
         }
         self
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
@@ -904,4 +904,23 @@ mod test {
         }
         ");
     }
+
+    #[test]
+    fn do_not_transform_failing_array_access_in_brillig() {
+        let src = "
+        brillig(inline) predicate_pure fn main f0 {
+          b0():
+            v0 = allocate -> &mut u8                      
+            store u8 0 at v0                               
+            v2 = make_array [u8 0, v0] : [(u8, &mut u8); 1]
+            v4 = array_get v2, index u32 2 -> u8         
+            v6 = array_get v2, index u32 3 -> &mut u8      
+            return v4
+        }
+        ";
+
+        let ssa = Ssa::from_str(src).unwrap();
+        let ssa = ssa.remove_unreachable_instructions();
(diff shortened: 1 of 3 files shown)


asterite (approved):
Looks good!

I do wonder if the `context.dfg.runtime().is_acir()` check should be done depending on the side_effects_var value, but if all tests passes and there are no big regressions it probably means that gating it in all cases is probably fine.

vezenovm:
> I do wonder if the `context.dfg.runtime().is_acir()` check should be done depending on the side_effects_var value, but if all tests passes and there are no big regressions it probably means that gating it in all cases is probably fine.

Do you think it may be cleaner to just run this optimization on ACIR entirely? It looks like this pass was originally made for that runtime either way? I guess it does provide benefits for Brillig compilation where we potentially don't have to pass over many unreachable instructions.

asterite:
> Do you think it may be cleaner to just run this optimization on ACIR entirely? It looks like this pass was originally made for that runtime either way?

The pass was originally made to solve this issue: https://github.com/noir-lang/noir/issues/8774

In there there's an always-failing constraint and later there's a load from a value that can't exist becase it's from an `array_get` from an empty array. There was an alternative fix at that time (https://github.com/noir-lang/noir/pull/8943) but I thought it would be more correct to remove any instructions after an always-failing one and end the block with `unreachable`.

I don't know if that same code would fail in brillig, I'll check it now, but if not then we could gate it just to ACIR, though I feel that this is a good thing to do in general as it removes redundant instructions.

vezenovm:
> though I feel that this is a good thing to do in general as it removes redundant instructions.

Yes I agree. I'll try to keep what we can for Brillig then.

EDIT: We no longer get the Aztec end to end tests failure when gating just ArrayGet/Set

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -102,6 +102,10 @@ impl Function {
                 return;
             }
 
+            // Certain instructions can only be marked unreachable if we have determined they failed under a constant predicate.
+            // It is not accurate to use `side_effects_condition` for marking Brillig predicates (which are determined by the CFG).
+            // Currently we only account for ACIR predicates, thus, any instruction that can result in marking
+            // `current_block_reachability` as `Reachability::UnreachableUnderPredicate` should be gated for only ACIR runtimes.
```
vezenovm:
```suggestion
```

PR #9377, opened 2025-08-01
chore: add a mem2reg test for when all references need to be invalidated

asterite:
# Description

## Problem

No issue, just part of the mem2reg audit.

## Summary

While browsing the code I noticed this case where aliases is empty, in which case we have to invalidate all references. I wondered what piece of code could produce a reference for which we don't have at least one alias (itself) so I added a debug print in that case. The debug print was hit, but then I checked if, when it's hit, there were references and last stores to invalidate at all. At least when running all tests I couldn't find such case so this PR turns that into an assertion.

Well, when creating this PR it turns out the fuzzer found some code that triggered that case. I reduced that code, got an SSA for it and added a test for that scenario, trying to explain the reasoning behind it.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
@@ -1417,4 +1417,33 @@ mod tests {
         let ssa = ssa.mem2reg();
         assert_normalized_ssa_equals(ssa, src);
     }
+
+    #[test]
+    fn unknown_aliases() {
+        // This is just a test to ensure the case where `aliases.is_unknown()` is tested.
+        // Here, `v8 = load v0 -> Field` cannot be removed and replaced with `Field 10`
+        // because `v6` is passed to another function and we don't know what that reference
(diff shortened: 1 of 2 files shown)


compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -1417,4 +1417,33 @@ mod tests {
         let ssa = ssa.mem2reg();
         assert_normalized_ssa_equals(ssa, src);
     }
+
+    #[test]
+    fn unknown_aliases() {
+        // This is just a test to ensure the case where `aliases.is_unknown()` is tested.
+        // Here, `v8 = load v0 -> Field` cannot be removed and replaced with `Field 10`
+        // because `v6` is passed to another function and we don't know what that reference
+        // points it, and it could potentially change the value of `v0`.
```
asterite:
```suggestion
        // points to, and it could potentially change the value of `v0`.
```

jfecher (approved):

PR #9381, opened 2025-08-04
fix: disable comptime printing when requesting json output

### LINKED ISSUE #9380: Comptime printing breaks brillig bytecode benchmarks
Jake and I noticed that the `regression_cancelling` contains some comptime prints which pollute the output of stdout. [This then prevents `jq` from properly parsing the output](https://github.com/noir-lang/noir/actions/runs/16724041622/job/47335279780#step:4:20).

We already disable the runtime prints (by directing them into `std::io::empty()`) but we don't have any equivalent for comptime printing so these all go into stdout. We should allow passing a custom writer for the comptime interpreter to use as well.

TomAFrench:
# Description

## Problem\*

Resolves #9380 

## Summary\*

We now specify that we expect zero comptime printing when requesting a machine readable output.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/nargo_cli/src/cli/info_cmd.rs
+++ b/tooling/nargo_cli/src/cli/info_cmd.rs
@@ -62,7 +62,7 @@ impl WorkspaceCommand for InfoCommand {
 
 pub(crate) fn run(mut args: InfoCommand, workspace: Workspace) -> Result<(), CliError> {
     if args.json {
-        // In order to have a machine readable output, we cannot allow the program print arbitrary data to stdout.
+        // In order to have a machine readable output, we cannot allow the program to print arbitrary data to stdout.
         args.compile_options.disable_comptime_printing = true;
     }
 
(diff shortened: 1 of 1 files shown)


tooling/nargo_cli/src/cli/info_cmd.rs
```diff
@@ -61,6 +61,11 @@ impl WorkspaceCommand for InfoCommand {
 }
 
 pub(crate) fn run(mut args: InfoCommand, workspace: Workspace) -> Result<(), CliError> {
+    if args.json {
+        // In order to have a machine readable output, we cannot allow the program print arbitrary data to stdout.
```
TomAFrench:
```suggestion
        // In order to have a machine readable output, we cannot allow the program to print arbitrary data to stdout.
```

jfecher (approved):

PR #9387, opened 2025-08-04
chore: some mem2reg refactors

asterite:
# Description

## Problem

For the mem2reg internal audit.

## Summary

There are several commits here which should be self-contained so each can be reviewed independently.

## Documentation

Check one:

# PR Checklist


asterite:
> Blocking this for now since this touches the Expression enum which may change semantics in the code

I brought back ArrayElement in case not having it breaks something. That said, I still kept it as holding a `ValueId` instead of `Box<Expression>` because at least in the current mem2reg form it always holds another `Expression::Other`.

asterite:
I'll create issues for the things I find.

PR #9388, opened 2025-08-04
chore(inlining): Various unit tests

vezenovm:
# Description

## Problem\*

No issue, just noticed we were missing various unit tests for various logic in our inlining pass. 

## Summary\*

All unit tests added in this PR are passing as expected. 
Additional tests in inlining and inline info computation:
- Preventing inlining Brillig into ACIR
- Inlining functions containing static assertions
- Inlining no predicates with and without the config flag active 
- `#[inline_always]` attribute

CallGraph tests:
- Multiple recursive cycles
- Additional self recursion checks
- Dead code tracked in call graph

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/inlining.rs
```diff
@@ -1030,4 +1050,98 @@ mod test {
...
+              return
+        }
+
+        brillig(inline_always) fn always_inline f1 {
+            b0():
+              return
+        }
+        ";
+        let ssa = Ssa::from_str(src).unwrap();
+        let ssa = ssa.inline_functions(i64::MAX).unwrap();
```
TomAFrench:
```suggestion
        let ssa = ssa.inline_functions(i64::MIN).unwrap();
```

Probably more useful to use the minimum aggressiveness here as otherwise this `f1` would be inlined anyway. It would also be nice to have this test also check that if the function doesn't have `inline_always` then it won't be inlined (to ensure that this behaviour is from the `inline_always` status)

TomAFrench (approved):

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
vezenovm: Updated the test to also show the function not inlined without `inline_always`

PR #9400, opened 2025-08-05
fix: remember dereference in mem2reg ArrayGet

### LINKED ISSUE #9398: mem2reg: store to nested parameter reference is incorrectly removed
### Aim

This code:

```noir
fn main() {
    let string = &mut "OOPS";
    let param = [string];
    println(*param[0]);
    foo(param);
    println(*param[0]);
}

fn foo(param: [&mut str<4>; 1]) {
    let s = param[0];
    *s = "GOOD";
}
```

correctly prints "OOPS" first, then "GOOD".

However, if we change it to this:

```noir
fn main() {
    let string = &mut "OOPS";
    let param = [string];
    println(*param[0]);
    foo(param);
    println(*param[0]);
}

fn foo(param: [&mut str<4>; 1]) {
    let s = param[0];
    *s = "GOOD";
    for _ in 0..1 {}
}
```

it prints "OOPS" twice.

What's happening is that `*s = GOOD;` is incorrectly removed. The reason is that in mem2reg we check if a store is to a block or function parameter and don't remove it in that case, but I think we are not checking if those parameters have nested references.

Why did the first code not fail? I think it's because some inlining happens that prevents mem2reg to reach that bad state.

We can write this test:

```noir
    #[test]
    fn does_not_remove_store_to_nested_function_parameter() {
        // The last store can't be removed as it stores a value in a function parameter
        let src = "
        acir(inline) fn main f0 {
          b0(v0: [&mut Field; 1]):
            v1 = array_get v0, index u32 0 -> &mut Field
            jmp b1()
          b1():
            store Field 4 at v1
            return
        }
        ";

        let ssa = ssa.mem2reg();
        // We expect the program to be unchanged
    }
```

and it fails as the store is incorrectly removed.

### Expected Behavior

The second snippet should print "OOPS" then "GOOD".

### Bug

The second snippet should prints "OOPS" twice.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

The second snippet should print "OOPS" then "GOOD".

### Additional Context

It might be good to have the fuzzer create programs like the above so maybe we could find more bugs.

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9398

## Summary

Also adds a couple of tests for code that wasn't hit in these unit tests.

I'm not entirely sure the fix is correct though, but: it seems to fix the issue, and `array_get` that returns a reference does seem very similar to loading a reference value from an address (if we consider the array value to be an address).

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
@@ -1466,4 +1466,25 @@ mod tests {
         // We expect the program to be unchanged
         assert_normalized_ssa_equals(ssa, src);
     }
+
+    #[test]
+    fn does_not_remove_store_to_return_value() {
+        // The last store can't be removed as it stores a value in a function parameter
+        let src = "
+        acir(inline) fn main f0 {
+          b0():
+            v0 = allocate -> &mut Field
+            store Field 4 at v0
+            jmp b1()
+          b1():
+            return v0
+        }
+        ";
+
+        let ssa = Ssa::from_str(src).unwrap();
+
+        let ssa = ssa.mem2reg();
+        // We expect the program to be unchanged
+        assert_normalized_ssa_equals(ssa, src);
+    }
 }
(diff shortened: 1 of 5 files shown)


asterite:
Actually, I'll split this into two PRs: a small one with the fix, another bigger one with tests (these two and some others).

Issue #9404, opened 2025-08-05
Inlining has two distinct checks for whether a function should be inlined


vezenovm:
### Problem

We precompute [InlineInfos](https://github.com/noir-lang/noir/blob/b2713a99fca969a888ecf14ebee7db62499adef0/compiler/noirc_evaluator/src/ssa/opt/inlining/inline_info.rs#L48) to aid the inliner on whether a function should be inlined. 

We also have a separate function [PerFunctionContext::should_inline_call](https://github.com/noir-lang/noir/blob/b2713a99fca969a888ecf14ebee7db62499adef0/compiler/noirc_evaluator/src/ssa/opt/inlining.rs#L538) which we call https://github.com/noir-lang/noir/blob/b2713a99fca969a888ecf14ebee7db62499adef0/compiler/noirc_evaluator/src/ssa/opt/inlining.rs#L538 before another call into the `should_inline_call` closure set elsewhere https://github.com/noir-lang/noir/blob/b2713a99fca969a888ecf14ebee7db62499adef0/compiler/noirc_evaluator/src/ssa/opt/inlining.rs#L104

We are duplicating checks across two different filters. These have potential to diverge if one is updated but not the other. For example, if `PerFunctionContext::should_inline_call` returns `None` for a function while the `should_inline_call` returns true that function will not be inlined. Right now `PerFunctionContext::should_inline_call` looks to be returning the correct values, but this may not always stay the same. 

### Happy Case

We should have a single filter for deciding whether calls should be inlined. This makes maintaining inlining much simpler as there is one source of truth about whether a function should be inlined. 

### Workaround

None

### Project Impact

Nice-to-have

### Would you like to submit a PR for this Issue?

None

PR #9412, opened 2025-08-06
chore: do not inline acir calls in brillig

### LINKED ISSUE #9390: Inlining ACIR into Brillig can trigger incorrect semantics
### Aim

Calls from ACIR into Brillig are marked as Brillig entry points and prevented from being inlined. However, during inlining we do allow inlining ACIR into Brillig runtimes. It is assumed that the frontend has appropriately resolved all runtimes and thus any calls from a Brillig runtime must be into a Brillig runtime. This is not a safe assumption as the compiler currently has bugs for mislabeled runtimes https://github.com/noir-lang/noir/issues/7289 and even once #7289 is resolved we cannot guarantee a new bug will not come up in the future. 

### Expected Behavior

ACIR should not be inlined into Brillig runtimes. ACIR being called from a Brillig runtime should most likely fail at validation while also being prevented during inlining as an extra security measure (in case validation has a bug or is changed). 

### Bug

This test passes right now:
```rust

    #[test]
    fn inlining_acir_into_brillig_function() {
        let src = "
        brillig(inline) fn main f0 {
          b0(v0: u32):
            call f1(v0)
            return
        }
        acir(inline) fn foo f1 {
          b0(v0: u32):
            v4 = make_array [Field 1, Field 2, Field 3] : [Field; 3]
            v5 = array_get v4, index v0 -> Field
            return
        }
        ";
        let ssa = ssa.inline_functions(i64::MAX).unwrap();
        assert_ssa_snapshot!(ssa, @r"
        brillig(inline) fn main f0 {
          b0(v0: u32):
            v4 = make_array [Field 1, Field 2, Field 3] : [Field; 3]
            v5 = array_get v4, index v0 -> Field
            return
        }
        ");
    }
```
This will cause incorrect semantics. If the input to main causes an OOB array access this SSA will succeed for Brillig when we would expect a failure (preceding array OOB checks are expected to be laid down in Brillig post https://github.com/noir-lang/noir/pull/9200 while in ACIR array ops are side effectual). 

### To Reproduce

The bug can be shown in a few ways:
1. Copy the unit test in the section above to the tests module under `opt/inlining`
2. Demonstrate the ACIR vs. Brillig execution difference.
Copy the following test to the test module under `ssa_executor`:
```rust

    #[test]
    fn array_get_oob_in_acir_called_from_brillig() {
        let ssa_without_runtime = "
        (inline) fn main f0 {
          b0(v0: u32):
            v1 = call f1(v0) -> Field
            return
        }
        acir(inline) fn foo f1 {
          b0(v0: u32):
            v4 = make_array [Field 1, Field 2, Field 3] : [Field; 3]
            v5 = array_get v4, index v0 -> Field
            return v5
        }
        ";
        let acir_ssa = "acir".to_string() + ssa_without_runtime;
        // let _ = Ssa::from_str(&acir_ssa).unwrap();
        let brillig_ssa = "brillig".to_string() + ssa_without_runtime;
        let mut witness_map = WitnessMap::new();
        witness_map.insert(Witness(0), FieldElement::from(10_u32));
        let acir_result =
            execute_ssa(acir_ssa.to_string(), witness_map.clone(), CompileOptions::default());
        let mut compile_options = CompileOptions::default();
        // Uncomment the line below to see this test pass
        // compile_options.skip_ssa_pass = vec!["Inlining".to_string(), "Preprocessing".to_string()];
        let brillig_result =
            execute_ssa(brillig_ssa.to_string(), witness_map, compile_options);
        match (acir_result, brillig_result) {
            (Err(acir), Ok(_brillig)) => panic!("Acir failed with: {acir}, brillig succeeded"),
            (Ok(_acir), Err(brillig)) => panic!("Acir succeeded, brillig failed: {brillig}"),
            _ => {}
        }
    }
```
You can replace the SSA above with an unused array get to see the same result:
```
        let ssa_without_runtime = "
        (inline) fn main f0 {
          b0(v0: u32):
            call f1(v0)
            return
        }
        acir(inline) fn foo f1 {
          b0(v0: u32):
            v4 = make_array [Field 1, Field 2, Field 3] : [Field; 3]
            v5 = array_get v4, index v0 -> Field
            return
        }
        ";
```
Take note that skipping any passes that perform inlining causes the tests above to pass. You can uncomment the line setting `skip_ssa_pass` above to see this effect. 

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Blocker

### Blocker Context

This is a security issue and should be resolved. 

### Nargo Version

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #9390
## Summary\*
ACIR calls from Brillig are not inlined and furthermore they are forbidden during ssa validation.

## Additional Context
The defunctionalisation pass has also been updated, because of these problems that become visible with this PR:
- the signatures when computing the 'variants' (i.e the set of functions matching a signature) are not checked against their runtime.
- generated code for apply function crash if there is a runtime error, due to not having a location.
- empty variants get a dummy function.

It turns out that calling acir from brillig can happen in the apply function of the defunctionalisation pass: this happens for instance with the zeroed functions generated by the frontend in the lambda_from_dynamic_if test, with force-brillig.
In that case, generating a runtime error when calling acir from brillig generates a crash due to missing location.
Filtering the bad function in the 'variants' works in this case but fail in other cases due to the dummy function.

As a result, I ensured that the apply function never calls acir from brillig by adding an assert if it happens, which provides an error message to the user.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/inlining.rs
```diff
@@ -552,7 +564,8 @@ impl<'function> PerFunctionContext<'function> {
         match callee.runtime() {
             RuntimeType::Acir(inline_type) => {
                 // If the called function is acir, we inline if it's not an entry point
-                if inline_type.is_entry_point() {
+                // If it is called from brillig, it cannot be inlined because runtimes do not share the same semantic
+                if inline_type.is_entry_point() || self.entry_function.runtime().is_brillig() {
```
vezenovm: One of the audit issues https://github.com/noir-lang/noir/issues/9404 is about moving away from both this `should_inline_call` and the separate closure `should_inline_call` that checks the inline infos. Could we try to prevent inline ACIR into Brillig using that?

compiler/noirc_evaluator/src/ssa/ir/call_graph.rs
```diff
@@ -63,10 +63,17 @@ impl CallGraph {
         let mut ids_to_indices = HashMap::default();
         let mut indices_to_ids = HashMap::default();
 
-        for function in dependencies.keys() {
-            let index = graph.add_node(*function);
-            ids_to_indices.insert(*function, index);
-            indices_to_ids.insert(index, *function);
+        for dep in &dependencies {
```
vezenovm: It is unclear to me why this loop was changed can you elaborate? Were certain callees not part of the graph? If so, this surprises me as the `dependencies` should include every function in the SSA.

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
vezenovm: I guess as we are already looking at the entry function here this is fine for now and we can converge the should inline check in a separate PR.

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
```diff
@@ -122,9 +122,13 @@ impl Ssa {
         let mut new_functions = std::collections::BTreeMap::new();
         for entry_point in inline_targets {
             let function = &self.functions[&entry_point];
-            let inlined = function.inlined(&self, &should_inline_call)?;
+            let inlined = function.inlined(&self, &should_inline_call, &mut callees)?;
             new_functions.insert(entry_point, inlined);
         }
+        for callee in callees {
+            // If we have callees that were not inlined, we need to keep them in the SSA.
```
vezenovm: It feels unideal that we have to start threading a callers set just so that we can not remove incorrectly marked ACIR functions called from Brillig. Could we look at avoiding this `callers` set? 

We could try marking ACIR called from Brillig as an entry point within the inline infos or we just error out during inlining if we ever have an ACIR function called from Brillig. We already have Brillig functions we do not inline that remain in place as they are marked as should not inline by the inline infos and they do not require this tracking. Could we use a similar strategy and just mark ACIR called from Brillig as should not inline thus triggering them as an inline target? The other option would be to just error during inlining which I think is reasonable as it is a bad formed SSA.

guipublic: yes, in fact this is not enough in some more complex cases and I did not manage to make it through without inlining.
so I will remove this and simply return an error.

compiler/noirc_evaluator/src/ssa/ir/call_graph.rs
guipublic: This function assumes that all the function ids in dependencies.values() are also in dependencies.keys().
My changes do not require this assumption anymore.

I did it because not inlining acir calls from brillig break this assumption. Since I am going to undo this (and return an error instead, see below), I can also revert this, if you want (I don't have a strong opinion here).

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
guipublic: Done

TomAFrench:
I remember adding some assertions for this case in the past and from what I remember, I ran into issues with closures. I was working on the understanding that we'd need #7289 before we could enforce this.

guipublic:
> I remember adding some assertions for this case in the past and from what I remember, I ran into issues with closures. I was working on the understanding that we'd need #7289 before we could enforce this.

Yes this is what is happening here, most tests with lambdas have the issue (most of the case with sorting arrays). At least it is now explicit.

compiler/noirc_evaluator/src/ssa/ir/call_graph.rs
vezenovm: > This function assumes that all the function ids in dependencies.values() are also in dependencies.keys().

I think it is a reasonable assumption because then it means something is off with the SSA form. As this case will not matter any more I'd rather revert this change.

compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
```diff
@@ -337,7 +337,14 @@ fn find_variants(ssa: &Ssa) -> Variants {
     for (dispatch_signature, caller_runtime) in dynamic_dispatches {
         let target_fns =
             signature_to_functions_as_value.get(&dispatch_signature).cloned().unwrap_or_default();
-        variants.insert((dispatch_signature, caller_runtime), target_fns);
+        variants.insert(
+            (dispatch_signature, caller_runtime),
+            target_fns
+                .iter()
+                .filter(|f| ssa.functions[f].runtime() == caller_runtime)
```
vezenovm: Could we have a unit test for this new filter? Were you hitting validation failures without this check?

compiler/noirc_evaluator/src/ssa/opt/inline_simple_functions.rs
```diff
@@ -67,16 +69,24 @@ impl Ssa {
...
 
+        let mut result: Option<RuntimeError> = None;
         self.functions = btree_map(&self.functions, |(id, function)| {
+            let inlined = function.inlined(&self, &should_inline_call);
             (
                 *id,
-                function
-                    .inlined(&self, &should_inline_call)
-                    .expect("simple function should not be recursive"),
+                if let Ok(new_function) = inlined {
```
vezenovm: We should be able to use `try_btree_map` here as we return the resulting error after this mapping either way.

guipublic: Done

compiler/noirc_evaluator/src/ssa/ir/call_graph.rs
guipublic: Done

compiler/noirc_evaluator/src/ssa/opt/defunctionalize.rs
guipublic: I finally removed the filter, and added a unit test that show what happens in case of a acir and brillig mix.
I updated the PR description for more info.

TomAFrench:
Shall we change the base of this PR to go into @jfecher's branch?

vezenovm:
> Shall we change the base of this PR to go into @jfecher's branch?

This PR would be useful for testing out #9484.

vezenovm:
@guipublic #7289 is resolved so this PR should be ready again once merge conflicts are resolved.

vezenovm (approved):

PR #9425, opened 2025-08-06
chore(remove_paired_rc: Add various unit tests

vezenovm:
# Description

## Problem\*

Part of the internal audit

## Summary\*

We did not have many tests in `remove_paired_rc` confirming its intended functionality. I have added several here.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/rc.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/rc.rs
@@ -154,9 +154,13 @@ fn remove_instructions(to_remove: HashSet<InstructionId>, function: &mut Functio
 #[cfg(test)]
 mod test {
 
-    use crate::ssa::{
-        ir::{basic_block::BasicBlockId, dfg::DataFlowGraph, instruction::Instruction},
-        ssa_gen::Ssa,
+    use crate::{
+        assert_ssa_snapshot,
+        ssa::{
+            ir::{basic_block::BasicBlockId, dfg::DataFlowGraph, instruction::Instruction},
+            opt::assert_normalized_ssa_equals,
+            ssa_gen::Ssa,
+        },
     };
 
     fn count_inc_rcs(block: BasicBlockId, dfg: &DataFlowGraph) -> usize {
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/rc.rs
```diff
@@ -265,4 +269,178 @@ mod test {
...
+            jmp b2()
+          b2():
+            v1 = make_array [v0] : [[Field; 2]; 1]
+            return v1  
+        }
+        ";
+        let ssa = Ssa::from_str(src).unwrap();
+        let ssa = ssa.remove_paired_rc();
+        assert_normalized_ssa_equals(ssa, src);
+    }
```
TomAFrench: hmm, seems like we should be able to remove these.

vezenovm: Yes we should but this is how the pass currently operates. It is very conservative and only looks for inc_rc's in the entry block and dec_rc's in the exit block. We previously only ever expected to generate dec_rc instructions in the last block of a function (we currently do not generate them at all). We would need a more involved pass to safely remove dec_rc instructions in non exit blocks.

compiler/noirc_evaluator/src/ssa/opt/rc.rs
```diff
@@ -265,4 +269,178 @@ mod test {
...
+            dec_rc v0
+            dec_rc v1
+            return
+        }
+        ";
+
+        let ssa = Ssa::from_str(src).unwrap();
+        let ssa = ssa.remove_paired_rc();
+        // We expect the paired RC on v0 to remain, but we expect the paired RC on v1 to be removed
+        // as they operate over different types ([Field; 2] and [Field; 5]) respectively.
```
asterite: This comment seems to be wrong (copied from the above test)

compiler/noirc_evaluator/src/ssa/opt/rc.rs
```diff
@@ -265,4 +269,178 @@ mod test {
...
+            return v1  
+        }
+        ";
+        let ssa = Ssa::from_str(src).unwrap();
+        let ssa = ssa.remove_paired_rc();
+        assert_normalized_ssa_equals(ssa, src);
+    }
+
+    #[test]
+    fn remove_pair_across_blocks() {
```
asterite, line 419: Maybe explain that these are only removed if they happen in the entry and exit blocks? At first I didn't know why this was different than the previous test until I read your answer above.

compiler/noirc_evaluator/src/ssa/opt/rc.rs
vezenovm:
```suggestion
        // We expect the paired RCs on v0 and v1 to remain as they operate over the same type ([Field; 5])
```

vezenovm: fixed

compiler/noirc_evaluator/src/ssa/opt/rc.rs
vezenovm: Added comments to both tests

asterite (approved):

Issue #9432, opened 2025-08-07
SSA Parser: Unreachable block causes unexpected output


vezenovm:
### Aim

Take the following test:
```rust
    #[test]
    fn infinite_recursion_jmp_loop() {
        let src = r#"
        brillig(inline) predicate_pure fn main f0 {
          b0():
            jmp b1()
          b1():
            jmp b2()
          b2():
            jmp b3()
          b3():
            jmp b1()
          b4():
            return
        }
        "#;

        let ssa = Ssa::from_str(src).unwrap();
        println!("{}", ssa.print_with(None));
    }
```
We need to comment out normalization in the parser's translator `finish` step due to https://github.com/noir-lang/noir/issues/9433. https://github.com/noir-lang/noir/pull/9435 is also needed as the printer currently does not print all blocks, only reachable ones.

You will get the following output:
```
brillig(inline) predicate_pure fn main f0 {
  b0():
    jmp b1()
  b1():
    jmp b2()
  b2():
    jmp b3()
  b3():
    jmp b1()
  b4():
    (no terminator instruction)
}
```
Now let's switch to a reachable return block and making a regular block the unreachable one:
```
        let src = r#"
        brillig(inline) predicate_pure fn main f0 {
          b0():
            jmp b1()
          b1():
            jmp b2()
          b2():
            jmp b4()
          b3():
            jmp b1()
          b4():
            return
        }
        "#;
```
We get the following output:
```
brillig(inline) predicate_pure fn main f0 {
  b0():
    jmp b1()
  b1():
    jmp b2()
  b2():
    jmp b4()
  b3():
    (no terminator instruction)
  b4():
    return
}
```

### Expected Behavior

I would expect the SSA to be parsed as-is. In the first case I expect the b4 block to be parsed. In the second case I don't expect any simplifications on the blocks to be run. 

### Bug

Unreachable blocks are missing from the parsed SSA. 

### To Reproduce

1. Run the tests above in the SSA parser tests module

### Workaround

None

### Project Impact

Blocker

### Blocker Context

This blocks being able to use the SSA parser for accurate unreachable blocks tests in simplify_cfg. 

### Nargo Version

nargo version = 1.0.0-beta.9 noirc version = 1.0.0-beta.9+cceeec0d4f379632eae2f3228d2b8cefb07af7ee (git version hash: e71436bad81b93da3f8ca3fb3d271bcb14ed0823, is dirty: true)

### Would you like to submit a PR for this Issue?

None

[...]

PR #9435, opened 2025-08-07
fix(ssa): Print all blocks including unreachable ones

vezenovm:
# Description

## Problem\*

Part of resolving issues with testing unreachable blocks (https://github.com/noir-lang/noir/issues/9432)

## Summary\*

I switched to `basic_blocks_iter` when displaying a function instead of `reachable_blocks`.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/printer.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/printer.rs
@@ -15,7 +15,7 @@ use crate::ssa::{
 };
 
 use super::{
-    basic_block::BasicBlockId,
+    basic_block::{BasicBlock, BasicBlockId},
     dfg::DataFlowGraph,
     function::Function,
     instruction::{ConstrainError, Instruction, InstructionId, TerminatorInstruction},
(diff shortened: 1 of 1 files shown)


asterite (approved):

compiler/noirc_evaluator/src/ssa/ir/printer.rs
```diff
@@ -93,7 +93,7 @@ fn display_function(
         writeln!(f, "{} fn {} {} {{", function.runtime(), function.name(), function.id())?;
     }
 
-    for block_id in function.reachable_blocks() {
+    for (block_id, _) in function.dfg.basic_blocks_iter() {
         display_block(&function.dfg, block_id, files, f)?;
```
asterite: Minor: we could change `display_block` to take a block instead of looking it up again.

vezenovm:
Not sure if we want to do this by default. We may want the option to explicitly print all blocks, but otherwise it will cause confusion while testing.

vezenovm:
Going to close for now. If we decide to test with unreachable blocks explicitly we can do it in a separate PR.

PR #9436, opened 2025-08-08
chore: restore some optimizations

TomAFrench:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify/binary.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify/binary.rs
@@ -297,18 +297,19 @@ pub(super) fn simplify_binary(binary: &Binary, dfg: &mut DataFlowGraph) -> Simpl
         BinaryOp::Shr => {
             // Bit shifts by constants on unsigned types can be treated as divisions.
             if lhs_type.is_unsigned() {
-                if let Some(rhs_const) = rhs_value {
-                    if rhs_const >= FieldElement::from(lhs_type.bit_size() as u128) {
-                        // Shifting by the full width of the operand type, any unsigned `lhs` goes to zero.
+                // We restrict this to shifts by constants less than 128, as larger shifts are expected to trigger
+                // a overflow error.
+                if let Some(rhs_const) =
+                    rhs_value.filter(|rhs_const| *rhs_const < FieldElement::from(128u128))
+                {
+                    let two_pow_rhs = 1u128 << rhs_const.to_u128();
+                    let two_pow_rhs = dfg.make_constant(two_pow_rhs.into(), lhs_type);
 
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/ir/dfg/simplify/binary.rs
```diff
@@ -292,8 +292,25 @@ pub(super) fn simplify_binary(binary: &Binary, dfg: &mut DataFlowGraph) -> Simpl
...
-        BinaryOp::Shl | BinaryOp::Shr => {
-            return SimplifyResult::SimplifiedToInstruction(simplified);
+        BinaryOp::Shl => return SimplifyResult::SimplifiedToInstruction(simplified),
+
+        BinaryOp::Shr => {
+            // Bit shifts by constants on unsigned types can be treated as divisions.
+            if lhs_type.is_unsigned() {
+                if let Some(rhs_const) = rhs_value {
+                    if rhs_const >= FieldElement::from(lhs_type.bit_size() as u128) {
+                        // Shifting by the full width of the operand type, any unsigned `lhs` goes to zero.
```
guipublic: Shifting by the full width of the operand type is now an overflow. There is no need to do this check here, if rhs is constant, the overflow check should be constant folded on its own.

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -346,6 +347,20 @@ impl<'a> FunctionContext<'a> {
                         result
                     }
                     BinaryOpKind::ShiftLeft | BinaryOpKind::ShiftRight => {
+                        let dfg = &self.builder.current_function.dfg;
+                        let max_lhs_bits = dfg.get_value_max_num_bits(lhs);
+
+                        if let Some(rhs_const) = dfg.get_numeric_constant(rhs) {
+                            let bit_shift_size = rhs_const.to_u128() as u32;
+
+                            if max_lhs_bits + bit_shift_size <= bit_size {
```
guipublic, line 356: If lhs is not zero: max_lhs_bits + bit_shift_size <= bit_size implies that bit_shift_size < bit_size, which implies that the overflow check will pass at compile time, so there is no need for this.
if lhs is zero, then this test is incorrect (not sure if we can get max_lhs_bits to be zero in this case though).

compiler/noirc_evaluator/src/ssa/ir/dfg/simplify/binary.rs
TomAFrench: Ah yeah, I added this back in based on the comment which is inaccurate compared to the code. We must have this optimisation (converting `x>>2` into `x/4`) somewhere else then I guess.

PR #9440, opened 2025-08-08
feat(opt): Don't clone on array_len

jfecher:
# Description

## Problem\*

Experiment: `array_len` takes its array by value which means we always clone (inc_rc) the array

## Summary\*

Adds a hardcoded to check to avoid cloning array_len's argument

## Additional Context

The "Clone" terminology is a bit confusing since we could mean:
- A `Expression::Clone` node in the hir (which translates to an `inc_rc` in ssa)
- An actual array copy in brillig

Using clone for the former and copy for the later would also be a bit confusing coming from rust so going forward I'm going to use the following terms:
- `Clone`: always a `Expression::Clone` or a place where a variable would be copied/cloned in rust because it is used more than once. Currently this always translates to an `inc_rc` on arrays for us, or nothing for other types of values.
- `memcopy`: always refers to the actual runtime copying of an array in brillig.

It is expected that `Clone_count >= memcpy_count`

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench (approved):
Nice!

PR #9442, opened 2025-08-08
chore!: predicate for ecdsa

guipublic:
# Description

## Problem\*

backends may have various expectations on blackbox inputs when the blackbox is behind a non-taken branch.
E.g BB often requires the inputs to be valid (points on the curve..).
Using a predicate to indicate whether the computation is disabled or not makes the behaviour more explicit and not dependent on specific backend implementations.

## Summary\*
Add a predicate to ecdsa blackboxes

## Documentation\*

Check one:

# PR Checklist\*


noir_stdlib/src/ecdsa_secp256k1.nr
```diff
@@ -22,15 +22,39 @@ pub fn verify_signature<let N: u32>(
...
 
 #[foreign(ecdsa_secp256k1)]
+fn _verify_signature<let N: u32>(
+    public_key_x: [u8; 32],
+    public_key_y: [u8; 32],
+    signature: [u8; 64],
+    message_hash: [u8; N],
+    predicate: bool,
+) -> bool
+{}
```
TomAFrench: Is this necessary? I'd expect that we can just default the predicate to true when doing ssa codegen.

guipublic: I thought about this but then figured out it's a bit hacky to have just one blackbox with non-matching signature and hardcoded behaviour in the compiler.
But yes I think it can be done, I'll give it a try.

TomAFrench: Yeah, I looked at this in a bit more detail and it's more hacky than I thought. Agreed that modifying the stdlib is the best solution here.

PR #9449, opened 2025-08-08
fix: allow calling private impl method defined on another module from…

### LINKED ISSUE #9447: Private impl method incorrectly errors when impl is defined in a separate module
### Aim

This code:

```noir
pub mod bar {
    pub struct Foo {}
}

impl bar::Foo {
    fn foo(self) {
        let _ = self;
    }

    fn bar(self) {
        self.foo();
    }
}

fn main() {}
```

gives this error:

```
error: foo is private and not visible from the current module
   ┌─ src/main.nr:11:14
   │
11 │         self.foo();
   │              --- foo is private
```

but no error should happen.

### Expected Behavior

No error should happen.

### Bug

An incorrect visibility error happens.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
… same impl

# Description

## Problem

Resolves #9447

## Summary

I made the behavior the same as in Rust (as far as I can tell). At least private methods on impls with different generic types can be called between the different impls.

## Documentation

Check one:

# PR Checklist


TomAFrench (approved):
LGTM

PR #9450, opened 2025-08-08
fix: unify `type * var` against `type` setting `var` to 1

### LINKED ISSUE #9430: Explicit type annotation in trait method call doesn't match
### Aim

This code fails to compile:

```noir
trait Serialize {
    let N: u32;

    fn serialize(self) -> [u32; N];
}

impl Serialize for u32 {
    let N: u32 = 1;

    fn serialize(self) -> [u32; Self::N] {
        [self; 1]
    }
}

impl<let N: u32, T: Serialize> Serialize for [T; N] {
    let N: u32 = <T as Serialize>::N * N;

    fn serialize(self) -> [u32; Self::N] {
        [0; Self::N]
    }
}

pub struct CompressedString<let N: u32> {
    value: [u32; N],
}

impl<let N: u32> Serialize for CompressedString<N> {
    let N: u32 = N;

    fn serialize(self) -> [u32; Self::N] {
        let _: [u32; N] = self.value.serialize();
        [0; Self::N]
    }
}

fn main() {}
```

The error is:

```
error: Expected type [u32; N], found type [u32; (N * _)]
   ┌─ src/main.nr:31:27
   │
31 │         let _: [u32; N] = self.value.serialize();
   │                           ----------------------
```

However, if the type annotation is removed from that line the code compiles fine. The same type shows up in the LSP inlay hint:

Image

### Expected Behavior

The code should compile.

### Bug

The code doesn't compile.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9430

## Summary

I think the main issue here is that when we get to this line:

```noir
let _: [u32; N] = self.value.serialize();
```

we don't fully solve the trait implementation (that's done at the end of the function) and trying to unify `N * var` against `N` eventually fails.

I added a new unification rule which I think should be correct.

## Documentation

Check one:

# PR Checklist


jfecher:
> `let _: [u32; N] = self.value.serialize();`

It looks like we're not solving the `Serialize<Self = [u32; N]>` constraint if it's not able to see that `Serialize<Self = u32>::N == 1`. Maybe we also need to ensure this type information is propagated downward more generally.

Let's save further work here for after the audit though!

TomAFrench:
I threw claudebox at this to see if it's still an issue in master and it came up with an alternative fix. Posting to link these two PRs

 https://github.com/noir-lang/noir/pull/12397

PR #9453, opened 2025-08-10
fix: guard `Eq` for `[T]` to avoid OOB on length mismatch

Galoretka:
# Description

Add short-circuit in Eq for slices to stop element access when lengths differ

## Problem\*

Prevents out-of-bounds access and constraint failures

## Additional Context

Aligns Eq behavior with Ord for slices


TomAFrench:
Oooh nice catch, thanks for reporting this! I've made a couple of changes which should reduce the gate counts a bit as well as adding a test.

TomAFrench (approved):

PR #9458, opened 2025-08-11
feat: mark slice variants of ecdsa verification as deprecated

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This is just fundamentally broken as it cannot handle dynamic length slices properly so we should remove these.

## Documentation\*

Check one:

# PR Checklist\*


Savio-Sou:
@TomAFrench cannot handle dynamic length slices how?

TomAFrench:
Slices can have a dynamic length whereas the black box opcode has a fixed length. If the slice that your passing to this function then as a length which is less than its capacity then the backend will see additional elements on the end of the slice and it won't be able to work out that these are dummy elements.

PR #9471, opened 2025-08-11
fix(acir-gen): Use current side effect predicate in memory ops for slices

### LINKED ISSUE #9467: Incorrect ACIR: `pop_front` fails when should not execute
### Aim

The fuzzer found a case where ACIR fails with "Index out of bounds" while Brillig fails with "division by zero". The two turned out to be unrelated errors. Here's a minimised program:
```rust
global G_A: [u32] = &[];
fn main(b: bool) -> pub u32 {
    if b {
        G_A.pop_front().0
    } else {
        1
    }
}
```

`Prover.toml`:
```toml
b = false
```

### Expected Behavior

It should return 1. 

### Bug

```console
  ┌─ src/main.nr:4:9
  │
4 │         G_A.pop_front().0
  │         -------------
  │
  = Call stack:
    1. src/main.nr:4:9

```

In brillig it works:
```
[fuzz_testing] Circuit output: 1
```

### To Reproduce

```
NOIR_AST_FUZZER_SEED=0xc52a931f00100000 cargo test -q -p noir_ast_fuzzer_fuzz acir_vs_brillig
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.10 noirc version = 1.0.0-beta.10+501988e746da0240d7af5d8026e5dd17a68e166b (git version hash: b7273d7a31f8e0cdc2d086e30ac3fede1f0ed46c, is dirty: true)

### NoirJS Version

_No response_

_No response_

Yes

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9467 

## Summary\*

Changes the SSA-to-ACIR codegen handling of `Intrinsics::SlicePopFront` and other slice operations to pass `current_side_effects_var` when creating the ACIR `Opcode::MemoryOp`, so that if side effects are disabled, then nothing happens. This way trying to pop from an empty slice on a disabled branch doesn't cause circuit failure.

### Failing circuit size calculations in BB
There is some strange [failure](https://github.com/noir-lang/noir/actions/runs/16892811187/job/47856600976) somewhere [in bb](https://github.com/AztecProtocol/aztec-packages/blob/5d4b716b373cec51a1944bb8c1e2b206fe90f662/barretenberg/cpp/src/barretenberg/stdlib_circuit_builders/rom_ram_logic.cpp#L85).

## Additional Context

I'm not sure why the same predicate wasn't used before for array operations. I only enabled it for slices, because for arrays the predicate seems to become part of the indexing itself.

Another approach would be to insert constraints into SSA like in https://github.com/noir-lang/noir/pull/9323 but in that PR we discussed that these should only be required for Brillig, so I thought ACIR should be able to handle this without extra constraints. 

Yet another alternative would be for `remove_unreachable_instructions` to look for 0 size, but I figured that would i) fragment the knowledge about these intrinsics even further, ii) it wouldn't be much different from inserting the constraints up front and iii) wouldn't cover popping from slices where the compiler doesn't statically know that the size is 0.

With these changes the ACIR for the program in the bug ticket looks like this:
```
func 0
current witness index : _3
private parameters indices : [_0]
return value indices : [_1]
BLACKBOX::RANGE [(_0, 1)] []
INIT (id: 0, len: 0, witnesses: [])
EXPR [ (-1, _2) 0 ]
MEM PREDICATE: EXPR [ (1, _0) 0 ]
(id: 0, read at: EXPR [ (1, _2) 0 ], value: EXPR [ (1, _3) 0 ]) 
EXPR [ (-1, _0, _3) (1, _0) (1, _1) -1 ]
```

## Documentation\*

Check one:

# PR Checklist\*


aakoshh:
@TomAFrench thanks for the heads up. Should I insert constraints into the SSA the way we do for Brillig then?

TomAFrench:
Yeah, I think we're going to need to explicitly guard in cases where the slice length can potentially go to zero.

aakoshh:
Okay, I'll close this and open an alternative PR. 

A quick test with adding the same intrinsic call check as we did in Brillig revealed that it won't work on its own for ACIR. It resuls in an SSA like this:
```
    enable_side_effects v0
    constrain u1 0 == v0, "Index out of bounds"   	// src/main.nr:4:9
    v7, v8, v9 = call slice_pop_front(u32 0, v1) -> (u32, u32, [u32])	// src/main.nr:4:9
```
So it requires that we don't go into the `if`, because that would result in IOOB, but then the `slice_pop_front` fails anyway because it doesn't care about the side effect variable (which is also removed, because the constraint doesn't care about it either). 

I'll see if we can somehow reason about being under a predicate in `remove_unreachable_instructions` to replace the values with defaults, but I'm not sure how this can be handled without the ACVM being told not to pop.

aakoshh:
Closing in favour of https://github.com/noir-lang/noir/pull/9489

PR #9484, opened 2025-08-13
fix: Monomorphize function values as pairs of `(constrained, unconstrained)`

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/7289

## Summary\*

Changes the monomorphizer to represent function values as a pair of a constrained version and unconstrained version of the same function. In the case of closures, the entire closure tuple is duplicated in each element (so you'd have `(constrained_closure_tuple, unconstrained_closure_tuple)`. Then, when the function is later called we index the tuple to extract the correct version based on the current runtime.

When calling a function, if the function called is a name referring to a known function we avoid monomorphizing both versions and monomorphize only the necessary version instead, although this optimization is not necessary for correctness.

## Additional Context

This is the monomorphizer-only variant of this change with no corresponding `?unconstrained` syntax. If we're alright with the approach here we could merge this first and add `?unconstrained` later (if desired) to limit the functions we compile as both constrained & unconstrained to only the necessary ones.

All lambdas are compiled as both constrained and unconstrained which can add a bunch of `"lambda", "lambda", "lambda",..." entries to the compiled artifact.

## Documentation\*

Check one:

# PR Checklist\*


jfecher:
Looks like arb_ast_roundtrip isn't true anymore. It asserts that the Ast is the same before/after monomorphization but after monomorphization we now turn all functions into tuples of (constrained, unconstrained).

I added a hack to avoid showing the second element of this tuple for tuples and for tuple types of functions but am still seeing some differences. Notably some method calls look like function calls now for some reason (unsure where the code for displaying that is) and because my hack arbitrarily displays the first element of the tuple, unconstrained function types are showing as being changed to constrained.

I'm also not sure if there is value to keeping this test though? It may also be more clear if my hack to not display these function tuples is only optional since without them we're not seeing the code that is actually sent to SSA and it is unclear why we see e.g. lambdas compiled as both constrained and unconstrained but one version looks unused.

aakoshh:
> I'm also not sure if there is value to keeping this test though? 

This test did uncover discrepancies between the fuzzer and the actual monomorphizer, so if possible I'd try to keep it. Arguably if the monomorphizer creates tuples for lambdas and the fuzzer does not, then that's a pretty significant difference. 

If that in itself isn't a problem, then instead of hacking the display, you could just add `avoid_lambdas: true` to the config in `mono.rs`, with an explanation of why we are not testing this.

jfecher:
@aakoshh thanks - I think I was misunderstanding the purpose of the test then. I'll add the `avoid_lambdas` switch but I'm not sure that is sufficient - would that avoid first-class functions and function types entirely?

aakoshh:
I just had a look: `avoid_lambdas` basically says that no function should have a function-type parameter, so there won't be function passed in calls. 

I may have misunderstood something when I added function parameters to the fuzzer, but I thought there are no actual lambdas in the monomorphized AST, only first class functions, it's just that some of them have been created from Noir lambdas. So what I did was that as I create the signature of functions that I'll generate the body for, they have a chance to get a function parameter that matches one of the already generated first class function signatures. Then, when we need to generate a call, we can look for values in the global function list or the local parameter list. But there are no lambda "literals" if that makes sense.

aakoshh:
What the fuzzer relies heavily on though, is the ability to print the monomorphized AST as Noir with the `DisplayAstAsNoir` construct, so if you have a hack that can display the tuples as they would be written, that would be used to 1) render the code we have to copy into an integration test if it fails and 2) for the comptime tests. 

Maybe the lambda signature to display (in types) could be based on the context of the function being printed? 


> Notably some method calls look like function calls now for some reason (unsure where the code for displaying that is)

Could that be related to this? https://github.com/noir-lang/noir/blob/12465f3ae45cdbd96531889e9420ad02254e17f9/compiler/noirc_frontend/src/monomorphization/printer.rs#L528

aakoshh:
I guess it can also be a problem that I choose a signature of specific first class function for a function parameter, say it's unconstrained, there won't be a constraint counterpart to be passed along it. Maybe the fuzzer only implements the spirit of `?unconstrained` scenario, where we only have 1 version.

jfecher:
> All lambdas are compiled as both constrained and unconstrained which can add a bunch of `"lambda", "lambda", "lambda",..." entries to the compiled artifact.

@vezenovm do you know if this'd be an issue? E.g. if users may be expecting only certain names in the artifact and now there are a bunch of lambda entries.

vezenovm:
> @vezenovm do you know if this'd be an issue? E.g. if users may be expecting only certain names in the artifact and now there are a bunch of lambda entries.

The compiled artifact should only contain entry point names.

jfecher:
And don't we consider each unconstrained function to be an entry point?
Since each lambda is both constrained & unconstrained, this could change things.

vezenovm:
Ah, I see your concern now. 

> And don't we consider each unconstrained function to be an entry point? Since each lambda is both constrained & unconstrained, this could change things.

Yes, each unconstrained function called from ACIR. In the case of an unconstrained lambda called from ACIR I would expect it to be included in the debug artifacts, which is the desired effect. However, having that name us just "lambda" would be confusing as it does not provide much information. This would be the case with any lambda entry point though. I think this may be something that should be handled in follow-up work. I will confirm whether or not we are adding extra debug names and the output for something like `nargo info`. 

Either way this shouldn't be too much a burden on users as debug artifacts are mostly used by our tools internally for displaying error or benchmark information. While for benchmarking we should try to push users to use the profiler where this shouldn't be much of an issue as information is displayed based off of call stacks and we already attach a unique index to account for repeated names. The main place this may cause confusion for users is `nargo info` which does utilize the function names on the compiled artifact.

compiler/noirc_frontend/src/monomorphization/mod.rs
```diff
@@ -1584,12 +1636,25 @@ impl<'interner> Monomorphizer<'interner> {
...
+            matches!(env.as_ref(), ast::Type::Tuple(_captures))
+        // All functions should be a pair of (constrained, unconstrained), with
+        // closures possibly represented in both. So we unpack the tuple and see
+        // if `constrained` is also a tuple.
         } else if let ast::Type::Tuple(elements) = t {
             if elements.len() == 2 {
-                matches!(elements[1], ast::Type::Function(_, _, _, _))
+                match &elements[0] {
+                    ast::Type::Tuple(inner_elements) if inner_elements.len() == 2 => {
+                        matches!(inner_elements[1], ast::Type::Function(..))
```
asterite, line 1654: Is it needed to check here that the env of this function is a tuple?

compiler/noirc_frontend/src/monomorphization/mod.rs
```diff
@@ -1050,11 +1062,16 @@ impl<'interner> Monomorphizer<'interner> {
         ident: HirIdent,
         expr_id: node_interner::ExprId,
         generics: Option<Vec<HirType>>,
+        // If set and this is a function value, only monomorphize the function for the
+        // constrainedness given by `self.in_unconstrained_function` rather than returning a tuple
```
vezenovm, line 1066: nit: cspell or `constrained-ness`

compiler/noirc_frontend/src/monomorphization/mod.rs
vezenovm: Wouldn't hurt to have a more robust check

compiler/noirc_frontend/src/monomorphization/mod.rs
```diff
@@ -2423,6 +2535,37 @@ impl<'interner> Monomorphizer<'interner> {
         self.in_unconstrained_function
             || self.interner.function_modifiers(&func_id).is_unconstrained
     }
+
+    // Functions are represented as pairs of (constrained, unconstrained) versions of the same
+    // function. When calling them we select the version that correspondsd to the current
```
vezenovm:
```suggestion
    // function. When calling them we select the version that corresponds to the current
```

compiler/noirc_frontend/src/monomorphization/printer.rs
```diff
@@ -499,29 +499,49 @@ impl AstPrinter {
         write!(f, ")")
     }
 
+    fn get_called_function(expr: &Expression) -> Option<(bool, &Definition, &String)> {
+        let is_unconstrained = |typ: &Type| match typ {
+            Type::Function(_, _, _, unconstrained) => *unconstrained,
+            Type::Tuple(elems) => match elems.first() {
```
vezenovm:
```suggestion
            Type::Tuple(elements) => match elements.first() {
```
or cspell

vezenovm (commented):
In the new `dual_constrained_lambdas` test I added `println(force_brillig);` right after the `force_brillig` declaration in `foo`. 

I get this failure:
```
error: Assertion failed: 'Bit size for lhs 1 does not match op bit size 254'
   ┌─ src/main.nr:28:5
   │
28 │     f()
   │     ---
   │
   = Call stack:
     1. src/main.nr:2:5
     2. src/main.nr:11:5
     3. std/lib.nr:40:9
     4. src/main.nr:28:5

Failed assertion
```
But only when I execute with `nargo execute --force --show-ssa`. Without `--show-ssa` the test passes and prints `false` appropriately.

compiler/noirc_frontend/src/monomorphization/mod.rs
jfecher: We shouldn't need to check if we already have a tuple containing `(env, function)` we know that must be a closure.
That being said it was easier to read previously since now a closure would be `((env, constrained_fn), (env, unconstrained_fn))`

vezenovm:
> Yes, each unconstrained function called from ACIR. In the case of an unconstrained lambda called from ACIR I would expect it to be included in the debug artifacts, which is the desired effect. However, having that name us just "lambda" would be confusing as it does not provide much information. This would be the case with any lambda entry point though.

So yes this is what happens, however, I think this was always the case for entry point lambdas and can be made more user friendly in follow-up work (perhaps a start would be just adding unique index suffixes to matching names like in the profiler).

In testing this I think I found something awry though. For example, I can write the following:
```noir
fn main(x: u32) {
    let functions: [unconstrained fn(u32) -> u32; 1] = [x_times_2];
    let f = functions[x];
    unsafe {
        assert_eq(f(x), x * 2);
    }
}
unconstrained fn x_times_2(x: u32) -> u32 {
    x * 2
}
```
This will give me the appropriate function name of `x_times_2` and that will be reflected in the compilation artifacts. Here is the output of `nargo info`:
```
+--------------------------+----------------------------+----------------------+--------------+-----------------+
| Package                  | Function                   | Expression Width     | ACIR Opcodes | Brillig Opcodes |
+--------------------------+----------------------------+----------------------+--------------+-----------------+
| dual_constrained_lambdas | main                       | Bounded { width: 4 } | 10           | 41              |
+--------------------------+----------------------------+----------------------+--------------+-----------------+
| dual_constrained_lambdas | x_times_2                  | N/A                  | N/A          | 33              |
+--------------------------+----------------------------+----------------------+--------------+-----------------+
| dual_constrained_lambdas | directive_integer_quotient | N/A                  | N/A          | 8               |
+--------------------------+----------------------------+----------------------+--------------+-----------------+
```
This is expected as `x_times_2` is explicitly specified as an unconstrained entry point.

I can then do the following:
```noir
fn main(x: u32) {
    let functions = [|x: u32| x * 2];
    let f = functions[x];
    unsafe {
        assert_eq(f(x), x * 2);
    }
}
```
This will give `lambda` as a Brillig entry point:
```
+--------------------------+------------------+----------------------+--------------+-----------------+
| Package                  | Function         | Expression Width     | ACIR Opcodes | Brillig Opcodes |
+--------------------------+------------------+----------------------+--------------+-----------------+
| dual_constrained_lambdas | main             | Bounded { width: 4 } | 17           | 42              |
+--------------------------+------------------+----------------------+--------------+-----------------+
| dual_constrained_lambdas | lambda           | N/A                  | N/A          | 33              |
+--------------------------+------------------+----------------------+--------------+-----------------+
| dual_constrained_lambdas | directive_invert | N/A                  | N/A          | 9               |
+--------------------------+------------------+----------------------+--------------+-----------------+
```
Even though the lambda is called in an unsafe block there is nothing that indicates this lambda should be Brillig as it was specified in an ACIR runtime and called from an ACIR runtime. Combined with the bug in https://github.com/noir-lang/noir/pull/9484#pullrequestreview-3138011209 the above leads me to think that we are incorrectly marking runtimes.

vezenovm:
> ```nr
> fn main(x: u32) {
>     let functions = [|x: u32| x * 2];
>     let f = functions[x];
>     unsafe {
>         assert_eq(f(x), x * 2);
>     }
> }
> ```
> 
> This will give `lambda` as a Brillig entry point:
> 
> [info table]
> 
> Even though the lambda is called in an unsafe block there is nothing that indicates this lambda should be Brillig as it was specified in an ACIR runtime and called from an ACIR runtime. 

To add onto this statement, when compiling the above with nightly I get a single `main` ACIR entry point as expected:
```
+--------------------------+----------+----------------------+--------------+-----------------+
| Package                  | Function | Expression Width     | ACIR Opcodes | Brillig Opcodes |
+--------------------------+----------+----------------------+--------------+-----------------+
| dual_constrained_lambdas | main     | Bounded { width: 4 } | 1            | 0               |
+--------------------------+----------+----------------------+--------------+-----------------+
```

jfecher:
> I can then do the following:
> 
> ```nr
> fn main(x: u32) {
>     let functions = [|x: u32| x * 2];
>     let f = functions[x];
>     unsafe {
>         assert_eq(f(x), x * 2);
>     }
> }
> ```
> 
> This will give `lambda` as a Brillig entry point:
> 
> ```
> +--------------------------+------------------+----------------------+--------------+-----------------+
> | Package                  | Function         | Expression Width     | ACIR Opcodes | Brillig Opcodes |
> +--------------------------+------------------+----------------------+--------------+-----------------+
> | dual_constrained_lambdas | main             | Bounded { width: 4 } | 17           | 42              |
> +--------------------------+------------------+----------------------+--------------+-----------------+
> | dual_constrained_lambdas | lambda           | N/A                  | N/A          | 33              |
> +--------------------------+------------------+----------------------+--------------+-----------------+
> | dual_constrained_lambdas | directive_invert | N/A                  | N/A          | 9               |
> +--------------------------+------------------+----------------------+--------------+-----------------+
> ```
> 
> Even though the lambda is called in an unsafe block there is nothing that indicates this lambda should be Brillig as it was specified in an ACIR runtime and called from an ACIR runtime. Combined with the bug in [#9484 (review)](https://github.com/noir-lang/noir/pull/9484#pullrequestreview-3138011209) the above leads me to think that we are incorrectly marking runtimes.

I think this is happening since functions are now compiled as a tuple of `(constrained, unconstrained)` so when you store them in an array, both versions are preserved - hence the regressions specifically in `lambda_from_array`. Do you think this needs to be addressed before this fix is merged?

vezenovm:
> Do you think this needs to be addressed before this fix is merged?

I was thinking that this is most likely a similar bug to the one that triggered https://github.com/noir-lang/noir/pull/9484#pullrequestreview-3138011209. Yes, we are storing both versions in an array but it looks like we are triggering the one with the Brillig runtime when we should be triggering the one with ACIR runtime. 

If the two bugs have different root causes though I don't think we need to wait on the latter bug to get this PR in.

jfecher:
> In the new `dual_constrained_lambdas` test I added `println(force_brillig);` right after the `force_brillig` declaration in `foo`.
> 
> I get this failure:
> 
> ```
> error: Assertion failed: 'Bit size for lhs 1 does not match op bit size 254'
>    ┌─ src/main.nr:28:5
>    │
> 28 │     f()
>    │     ---
>    │
>    = Call stack:
>      1. src/main.nr:2:5
>      2. src/main.nr:11:5
>      3. std/lib.nr:40:9
>      4. src/main.nr:28:5
> 
> Failed assertion
> ```
> 
> But only when I execute with `nargo execute --force --show-ssa`. Without `--show-ssa` the test passes and prints `false` appropriately.

So your program was the following?:

```rust
fn main() {
    foo(|| std::runtime::is_unconstrained());
}

// #7289 originates from a lambda being used in both acir & brillig
fn foo(is_brillig: fn() -> bool) {
    // If this is true the `--force-brillig` flag was passed
    let force_brillig = std::runtime::is_unconstrained();
    println(force_brillig);

    assert(!acir_function(is_brillig) | force_brillig);

    // safety:
    unsafe {
        assert(brillig_function(is_brillig));
        assert(brillig_function2(is_brillig));
    };
}

fn acir_function(f: fn() -> bool) -> bool {
    f()
}

unconstrained fn brillig_function(f: fn() -> bool) -> bool {
    f()
}

// acir_function should become unconstrained when called in
// an unconstrained context - and this should apply to the lambda
// it takes as an argument as well.
unconstrained fn brillig_function2(f: fn() -> bool) -> bool {
    acir_function(f)
}
```
I get no error for this program

vezenovm:
> I get no error for this program

Hmm yes. That is what I had. Let me reproduce once more.

vezenovm:
> Hmm yes. That is what I had. Let me reproduce once more.

Ah you need `--show-ssa`. That program will fail with the error I sent above. In fact, any print will fail. Even a `println` before the call to `foo` will fail when executed with `--show-ssa`.

vezenovm:
> Ah you need `--show-ssa`. That program will fail with the error I sent above. In fact, any print will fail. Even a `println` before the call to `foo` will fail when executed with `--show-ssa`.

The bug looks to be the same as the cause of #9097 which is resolved by #9365. You can look at the PR description of #9365 as to why this bug occurs. As this bug is already captured (with a potential solution) we can merge this PR.

PR #9487, opened 2025-08-13
chore: Check an inline always weight threshold during inline info computation

### LINKED ISSUE #9486: inlining: Introduce a weight threshold at which we always inline a function
### Problem

We determine whether or not we should inline based upon a "net cost" calculation (cost to retain a call vs. the cost to inline a function) that is checked against a user specified inliner aggressiveness: https://github.com/noir-lang/noir/blob/b51b6164e5e49243bb65adb64638a821131568dd/compiler/noirc_evaluator/src/ssa/opt/inlining/inline_info.rs#L201

This is sufficient for most use cases, however, this is heavily dependent upon the "interface cost" (e.g., the number of parameters and return values). This means that generally heavily used but small utility functions will be prevented from being inlined, even though these types of utility functions being inlined usually leads to a large amount of knock down optimizations. We should have some kind of maximum instruction threshold at which we set always set a function to be inlined. 

We worked around this by creating the [inline_simple_functions](https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/inline_simple_functions.rs) pass. The `inline_simple_functions` essentially does what I stated above, but in a separate pass. 

### Happy Case

We should be able to determine all inlining from a singular inline info computation. We can then remove the `inline_simple_functions` pass.

### Workaround

Yes

### Workaround Description

It exists in our code base: [inline_simple_functions](https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/inline_simple_functions.rs)

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #9486 

## Summary\*

I was just looking to show how we could improve our inline infos computation to remove the `inline_simple_functions` pass entirely.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/inlining.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/inlining.rs
@@ -990,12 +990,23 @@ mod test {
         let src = "
         brillig(inline) fn foo f0 {
           b0():
-            v1 = call f1() -> Field
+            v1 = call f1(Field 10, Field 10) -> Field
             return v1
         }
 
+        // We add a bunch of nonsense instructions here as to go over the weight threshold at which
+        // functions are always inlined. 
         brillig(inline) fn bar f1 {
-          b0():
+          b0(v0: Field, v1: Field):
+            v2  = add v0, v1
+            v3  = add v0, v1
+            v4  = add v0, v1
+            v5  = add v0, v1
+            v6  = add v0, v1
+            v7  = add v0, v1
+            v8  = add v0, v1
+            v9  = add v0, v1
+            v10 = add v0, v1
             return Field 72
         }
         ";
@@ -1025,6 +1036,14 @@ mod test {
           b2():
             jmp b3(Field 2)
           b3(v3: Field):
+            // We add a bunch of nonsense instructions here as to go over the weight threshold at which
+            // functions are always inlined. 
+            v4 = add v3, v3
+            v5 = add v3, v3
+            v6 = add v3, v3
+            v7 = add v3, v3
+            v8 = add v3, v3
+            v9 = add v3, v3
             return v3
         }
         ";
@@ -1046,6 +1065,12 @@ mod test {
           b1():
             jmp b2(Field 1)
           b2(v0: Field):
+            v2 = add v0, v0
+            v3 = add v0, v0
+            v4 = add v0, v0
+            v5 = add v0, v0
+            v6 = add v0, v0
+            v7 = add v0, v0
             return v0
         }
         ");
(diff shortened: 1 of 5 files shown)


compiler/noirc_evaluator/src/ssa/opt/inlining.rs
```diff
@@ -930,12 +943,23 @@ mod test {
...
+          b0(v0: Field, v1: Field):
+            v2  = add v0, v1
+            v3  = add v0, v1
+            v4  = add v0, v1
+            v5  = add v0, v1
+            v6  = add v0, v1
+            v7  = add v0, v1
+            v8  = add v0, v1
+            v9  = add v0, v1
+            v10 = add v0, v1
```
TomAFrench: Seems like we'd benefit from making `MAX_INSTRUCTIONS` an `small_function_minimum_instructions` argument rather than a global and a boolean argument.

vezenovm:
I would like to get https://github.com/noir-lang/noir/pull/9738 in before this PR.

test_programs/gates_report_brillig.sh
```diff
@@ -36,6 +36,6 @@ done
 
 echo "]" >> Nargo.toml
 
-nargo info --silence-warnings --force-brillig --json --inliner-aggressiveness $1 | jq -r ".programs[].functions = []"  > gates_report_brillig.json
+nargo info --silence-warnings --force-brillig --json --inliner-aggressiveness $1 --small-function-max-instructions $2 | jq -r ".programs[].functions = []"  > gates_report_brillig.json
```
vezenovm:
```suggestion
nargo info --silence-warnings --force-brillig --json --inliner-aggressiveness $1 | jq -r ".programs[].functions = []"  > gates_report_brillig.json
```

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
```diff
@@ -1385,3 +1393,267 @@ mod test {
...
+        assert_ssa_snapshot!(ssa, @r"
+        brillig(inline) fn main f0 {
+          b0(v0: Field):
+            v1 = add v0, v0
+            return v1
+        }
+        ");
+    }
+
+    /// This test is here to make clear that this SSA pass does not attempt multiple passes.
```
asterite: However, in the end everything ends up being inlined. Is this correct?

asterite (approved):
Looks great!

compiler/noirc_evaluator/src/ssa/opt/inlining.rs
vezenovm: Doh, yes this test was naively copied from `inline_simple_functions` but we run our actual inlining in a fixed point loop. I will update the test to show us not inlining if we are over the threshold.

vezenovm: I added a test `does_not_inline_function_with_multiple_instructions` to show a case when we prevent inlining and kept this test as is but updated the name and comments as this being fully inlined is the expected functionality.

PR #9489, opened 2025-08-14
fix(ssa): Replace pop from 0-length slice with constraint and defaults

### LINKED ISSUE #9467: Incorrect ACIR: `pop_front` fails when should not execute

aakoshh:
# Description

## Problem\*

Resolves #9467 
Replaces https://github.com/noir-lang/noir/pull/9471

## Summary\*

If we see a `v1, v2, v3 = call slice_pop_front(u32 0, v0)` which is statically known to fail, we replace it with:
1. `constrain 0 == <side-effect-variable>, "Index out of bounds"` which will fails _iff_ the _pop_ was enabled, and
2. `v3 = make_array []` which creates a default value for the result, so that we can still compile and execute the SSA, without executing the _pop_, which would fail, because it ignores the side effect variable. 

## Additional Context

In https://github.com/noir-lang/noir/pull/9471 my approach was to add the side effect variable as a predicate to the ACIR memory operation, turning it into a no-op if the branch was not enabled. However, Barretenberg doesn't handle memory operation predicates, so this was a no go. 

In this PR we instead look for a constant zero length, and replace it with a constraint that fails when _pop_ would have failed, but does _not_ fail if the side effects are disabled. In both cases we can remove the _pop_. 

We can only handle the case when we statically know that the slice is empty. If the length is a variable, and it depends on a condition, then it will still fail at runtime if it's empty, even if the side effects are disabled. In my testing I was able to create a situation where the length was a variable by merging slices of unequal length, but in these cases the _storage_ was not empty, and it successfully popped even if the length was 0. For example `let s: [u32] = if b { &[] } else { &[10] }`) results in `v0 = b * 1; v1 = make_array [b * 10]`, and despite length being 0, popping from `[0]` caused no "Index out of bounds". 

## Documentation\*

Check one:

# PR Checklist\*


jfecher (approved):
> We can only handle the case when we statically know that the slice is empty. If the length is a variable, and it depends on a condition, then it will still fail at runtime if it's empty, even if the side effects are disabled. 

Can you make sure we have an issue for the runtime case?

aakoshh:
https://github.com/noir-lang/noir/pull/9489/commits/67adec63fbf92e71cab7d0635fbb350f8a21ee7c was required because of a fuzzing failure:
```
NOIR_AST_FUZZER_SEED=0x5624a20000100000 cargo test -q -p noir_ast_fuzzer_fuzz min_vs_full
```

PR #9492, opened 2025-08-14
feat: keep last loads from predecessors in mem2reg

### LINKED ISSUE #9493: mem2reg: last loads from a single predecessor could be reused
When we build a `Block` (references) for a given block, we start by clearing out its last loads and stores:

https://github.com/noir-lang/noir/blob/ee2ac1a04e85134079fbe557d4029d3f64aff7b6/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs#L314-L316

I guess the rationale was that if a block has multiple predecessors, because `last_loads` maps a `ValueId` to `InstructionId` it will never be the case that two predecessors will map a same `ValueId` to a same `InstructionId`.

However, if a block has a single predecessor there's no reason why the `last_loads` from a previous block couldn't be reused. This can lead to removing duplicate loads across blocks. #9492 is a proof of this, bringing some optimizations (and a few regressions which I didn't spend time looking into).

In theory the same logic could be applied to `last_stores` though a quick trying doing this leads to some failing tests which, again, I didn't spend time looking into.

asterite:
# Description

## Problem

Resolves #9493

## Summary

There's one regression among all of the performance improvements but it happens only with inliner=max. Definitely something to improve but maybe the improvements are worth it.

## Documentation

Check one:

# PR Checklist


asterite:
Closing because I just wanted to be sure this could be a potential optimization before capturing an audit issue. See #9493

vezenovm:
> There's one regression among all of the performance improvements but it happens only with inliner=max.

Yeah if the only regression is with a max inliner setting I think this is a good optimization. It is unlikely projects in production will want a maximally aggressive inliner.

vezenovm:
> `test_report_noir-lang_noir_bigcurve_`	`380` s	`287` s	`1.32`

I added bench-show as mem2reg is one of the more memory heavy passes so I want to see the effect on our benchmarks. bigcurve looks to be the only library that caused an alert, but it would be good to see all the results.

vezenovm:
> `test_report_noir-lang_noir-bignum_`	`519` s	`383` s	`1.36`
> `test_report_noir-lang_noir_bigcurve_`	`358` s	`336` s	`1.07`

The test suite timings seem to jump around quite a bit, but the other compilation time and memory benchmarks all look good.

vezenovm (approved):

jfecher (approved):

PR #9507, opened 2025-08-15
fix(brillig): Pass slice length to push operation

### LINKED ISSUE #9496: ACIR != Brillig: push to merged slice prints different
### Aim

Discovered running the fuzzer. 

The following program prints different output in ACIR and Brillig:

```rust
fn main(b: bool) -> pub u32 {
    let mut ctx_limit: u32 = 3_u32;
    unsafe {
        func_2_proxy(if { b } { &[10] } else { &[20, 30] }, ctx_limit)
    }
}
unconstrained fn func_2(a: [u32], ctx_limit: &mut u32) -> u32 {
    if *ctx_limit == 0 {
        a[1 % a.len()]
    } else {
        let i = 3 % a.len();
        let b = a[i];
        let ab = a.push_back(b);
        println((a, i, b, ab));
        func_2(ab, ctx_limit)
    }
}
unconstrained fn func_2_proxy(a: [u32], mut ctx_limit: u32) -> u32 {
    func_2(a, (&mut ctx_limit))
}
```

`Prover.toml`:
```toml
b = true
```

### Expected Behavior

The output and result should be the same.

### Bug

```console
❯ cargo run -q -p nargo_cli -- execute --silence-warnings --force; cargo run -q -p nargo_cli -- execute --silence-warnings --force-brillig
(&[10], 0, 10, &[10, 0])
(&[10, 0], 1, 0, &[10, 0, 10])
(&[10, 0, 10], 0, 10, &[10, 0, 10, 0])
(&[10], 0, 10, &[10, 10])
(&[10, 10], 1, 10, &[10, 10, 10])
(&[10, 10, 10], 0, 10, &[10, 10, 10, 10])
[fuzz_testing] Circuit output: 10

```
1st (ACIR) is wrong, 2nd (Brillig) is correct. 

To give us a clue, look at the state of `a` printed in each round:
1. `(&[10], 0, 10, &[10, 0])` means it took the 0th element from the slice and pushed it to the back, but the result is `[10, 0]`. 2. `(&[10, 0], 1, 0, &[10, 0, 10])`, means we then took the 1st element, which is 0, pushed it, and got `[10,0,10]`, which actually revealed the 10 pushed in the previous round

Since the slice `a` depends on the input `b`,  and `b` is `true`, `a` becomes a slice with length 1, and a storage of `[10, 0]`, because each element is multiplied by the predicate `b` (`[b * 10 + (1-b)*20, b*0 + (1-b)*30]`). It looks like instead of assigning to the first item after the length, it always pushes to the end of the storage.

### To Reproduce

```
NOIR_AST_FUZZER_SEED=0x34fe042500100000 cargo test -q -p noir_ast_fuzzer_fuzz acir_vs_brillig
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.10 noirc version = 1.0.0-beta.10+ee2ac1a04e85134079fbe557d4029d3f64aff7b6 (git version hash: 4f8dbbc20bb40115d87c499154655ebe7e962099, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9496 

## Summary\*

The Brillig codegen ignored the separate semantic length variable in push operations.

## Additional Context

The Brillig in-memory representation of a vector is `[rc, size, capacity, items...]`. The push operation used the `size` in the vector to initialise the target vector, increased with the number of items being pushed. 

However, this `size` was different when the data came from ACIR _and_ the slice was a result of two slices of different length merged together. For example if ACIR merges `if b { &[10] } else { &[20, 30] }` then the result is an array of `[b*10+(1-b)*20, (1-b)*30]`, plus a separate length variable of `b*1+(1-b)*2`. 

When this is passed to Brillig, it looks something like this:
```
BRILLIG CALL func 0: inputs: [EXPR [ (-1, _0) 2 ], [EXPR [ (-10, _0) 20 ], EXPR [ (-30, _0) 30 ]], EXPR [ 3 ]], outputs: [_2]
```
Where:
* `_0` is `b`
* `EXPR [ (-1, _0) 2 ]` is `2-b`, which is 1 when `b` is `true`, and `2` when `b` is `false`
* `[EXPR [ (-10, _0) 20 ], EXPR [ (-30, _0) 30 ]]` is `[20-10*b, 30-30*b]`, which is `[10, 0]` when `b` is `true`, and `[20, 30]` when `b` is `false`.

The problem was that the Brillig bytecode did not consider the first parameter, only the second, which became in memory something like `[1, 2, 2, 10, 0]`. When we pushed a new item (10) onto it, it thought that based on the size of 2 it already reached the capacity, doubled it, and appended to the end; the target vector was created as `[1, 3, 4, 10, 0, 10, 0]`. 

The codegen independently resulting length as source length + 1, so it still returned a vector with length 2, but it looked like `[10, 0]`.

## Documentation\*

Check one:

# PR Checklist\*


aakoshh:
The following test triggers the memory ballooning to over 20GB; I killed it after 40s:
```shell
cargo nextest run -p nargo_cli --test execute tests::execution_success::test_array_sort::forcebrillig_true_inliner_0_expects
```

Looking at the memory profile with _Instruments_ didn't reveal anything. Printing the opcodes and matching them up to the code suggests that Brillig gets into an infinite loop in `vector_pop_back` when executing the bytecode generated by `codegen_mem_copy`. The loop seems to never exit.

Ah okay, got it. The problem is with vectors of tuples: the semantic length is in terms of number of tuple items, but the bytecode arithmetic assumes flattened size. In this case we have a vector of size 1, but we are trying to pop 2 items from it, which results in an overflow, and a loop going from 0 to u32::MAX.

PR #9535, opened 2025-08-18
chore: greenlight `make_constrain_not_equal` for audits

### LINKED ISSUE #9524: Greenlight `make_constrain_not_equal` for audits
Prepare https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/make_constrain_not_equal.rs for audits:

  - [ ] Purpose of the code
  - [ ] Design decisions made

**Target Date:** 12 Sep 2025

asterite:
# Description

## Problem

Resolves #9524

## Summary

I copied an explanation of why this pass exists from #7032

## Additional Context

I introduced a cSpell regression so to fix it I ran `cSpell` locally and noticed a couple more issues so I decided to fix them here as well as they are very minor changes.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/make_constrain_not_equal.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/make_constrain_not_equal.rs
@@ -24,8 +24,9 @@
 //! unnecessary witnesses - one which is unconstrained and the other constrained to
 //! zero. This is unnecessary as we want the circuit to just fail in this case.
 //!
-//! Note that this pass must be placed after [`CFG flattening`](super::flatten_cfg)
-//! as the flattening pass cannot handle this instruction.
+//! ## Preconditions:
+//! - this pass must be placed after [`CFG flattening`](super::flatten_cfg)
+//!   as the flattening pass cannot handle this instruction.
 use acvm::AcirField;
 
 use crate::ssa::{


compiler/noirc_evaluator/src/ssa/opt/make_constrain_not_equal.rs
```diff
@@ -1,3 +1,31 @@
...
+//! constrain v0 != v1
+//! ```
+//!
+//! When constraining with an equality in ACIR generation we need all the handling for the
+//! case where the two values ARE equal. Rather than just asserting that an inverse
+//! exists for the difference between these two values we need to create two
+//! unnecessary witnesses - one which is unconstrained and the other constrained to
+//! zero. This is unnecessary as we want the circuit to just fail in this case.
+//!
+//! Note that this pass must be placed after [`CFG flattening`](super::flatten_cfg)
```
vezenovm: Could mark this more explicitly with `# Preconditions`

compiler/noirc_evaluator/src/ssa/opt/make_constrain_not_equal.rs
```diff
@@ -82,4 +123,21 @@ mod tests {
...
+            v2 = eq v0, v1
+            constrain v2 == u1 0
+            return
+        }
+        ";
+        let ssa = Ssa::from_str(src).unwrap();
+        let ssa = ssa.make_constrain_not_equal();
+
+        // Expect no changes
+        assert_normalized_ssa_equals(ssa, src);
```
vezenovm:
```suggestion
        assert_ssa_does_not_change(src, Ssa:: make_constrain_not_equal);
```
We have this utility now. You will need to import it though.

PR #9540, opened 2025-08-18
chore: greenlight `brillig_array_get_and_set` for audits

### LINKED ISSUE #9523: Greenlight `brillig_array_get_and_set` for audits
Prepare https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/brillig_array_get_and_set.rs for audits:

  - [ ] Purpose of the code
  - [ ] Design decisions made

**Target Date:** 12 Sep 2025

asterite:
# Description

## Problem

Resolves #9523

## Summary

I couldn't find much to do here though I did introduce a few tiny refactors, mainly for consistency with other passes and to avoid an `expect`. I hope it's fine to do such refactors at this point.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/brillig_array_get_and_set.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/brillig_array_get_and_set.rs
@@ -119,7 +119,7 @@ impl Function {
     }
 }
 
-/// Given an array or slice value and a constant index, returns an offseted index
+/// Given an array or slice value and a constant index, returns an offset (shifted) index
 /// together with which type of [`ArrayOffset`] was used to shift it.
 fn compute_index_and_offset(
     context: &mut SimpleOptimizationContext,


aakoshh:
As a noob, I think the docs of this pass would benefit from an illustrative example at the top that explains what happens with some SSA and plain English. At least I never found the offsetting intuitive, and while there are unit tests, understanding those seems harder than it could be with some extra explanation.

asterite:
@aakoshh You are totally right! I updated the docs to explain things a bit more. Let me know if it's okay now.

compiler/noirc_evaluator/src/ssa/opt/brillig_array_get_and_set.rs
```diff
@@ -1,14 +1,37 @@
...
 //!
-//! For array operations with constant indices adding an instruction to offset the pointer
-//! is unnecessary as we already know the index. This pass looks for such array operations
-//! with constant indices and replaces their index with the appropriate offset.
+//! Now the index to retrieve is 4 and there's no need to offset it in Brillig,
+//! avoiding one addition.
+//! The "minus 1" part is just there so that readers can understand that the index
+//! was offset and that the actual element index is 3. On the Brillig side,
+//! array operations with constant indexes are always assumed to have already been
+//! shifted.
```
aakoshh, line 29: Ah, interesting; just curious if when this was added whether we also considered `array_get v0, index u32 3 plus 1` instead? I assume it was just less code change to do it this way, as the use of the (shifted) `index` was already in place?

compiler/noirc_evaluator/src/ssa/opt/brillig_array_get_and_set.rs
```diff
@@ -1,14 +1,37 @@
...
+//!
+//! In the case of slices, these are represented as Brillig vectors, where the items
+//! pointer instead starts at three rather than one.
+//! A Brillig vector is represented as [RC, Size, Capacity, ...items].
```
aakoshh: In this case, would we get `array_get v0, index u32 6 minus 3`?

aakoshh (approved):
Brilliant explanation, thanks!

compiler/noirc_evaluator/src/ssa/opt/brillig_array_get_and_set.rs
asterite: I think so. I added the `offset` field of `ArrayGet` and `ArraySet` though when I did that the logic to shift the index was already in place. The problem was that the interpreter was fetching the incorrect index as it didn't know it was shifted. So now the interpreter takes that index and subtracts the offset to get the final value.

We could change it to `3 plus 1`, it'll be the same "effort" for the interpreter, but I think for brillig it might not work. I'm not exactly sure how brillig works but I think it collects all used constant values and then it will take advantage of "4" being a known constant value. If it's "3 plus 1" it'll have to do the math again, which is what we want to avoid. BUT this is just my guess because of how it's done, maybe @vezenovm knows more about the reason.

compiler/noirc_evaluator/src/ssa/opt/brillig_array_get_and_set.rs
asterite: Indeed. I just added a new example below this, more docs never hurt :-)

PR #9547, opened 2025-08-18
fix: `assert_constant` refactors and fixes from audit

### LINKED ISSUE #9406: assert_constant: possible refactors
## Earlier check for presence of intrinsics

The algorithm of the pass is as follows:
- for each function:
  - compute empty loops
  - for each block
    - for each instruction
      - check the instruction

Checking the instruction first grabs the intrinsics `assert_constant` and `static_assert` from the current function. If they don't exist it means they aren't used in the function. In that case `check_instruction` does an early return.

However, this check could be done right at the beginning of each function check: if there's no such intrinsics (highly likely!) then we can skip computing empty loops and traversing all blocks and instructions.

The speed boost will likely not be noticeable (in `token_contract` it goes from 56ms to 0.02ms) but it's an easy change to do, and it might also make the code a bit easier to understand as it slightly simplifies `check_instruction`.

### LINKED ISSUE #9414: assert_constant: static_assert is simplified even when the message is dynamic
### Aim

This code:

```noir
use std::static_assert;

fn main(x: Field) {
    let string = f"Assertion failed: {x}";
    static_assert(false, string);
}
```

correctly fails with this error:

```
error: The static_assert message is not constant
  ┌─ src/main.nr:5:5
  │
5 │     static_assert(false, string);
  │     ----------------------------
```

That is, it's not valid to pass a dynamic message here.

However, if the static assertion holds, in some cases it won't produce this error:

```noir
use std::static_assert;

fn main(x: Field) {
    let string = f"Assertion failed: {x}";
    static_assert(true, string); // No "The static_assert message is not constant" error
}
```

The reason is that `static_assert` is simplified if the condition is known to be true, but this simplification doesn't check whether the message is dynamic or not. It might be better to remove this simplification altogether.

### Expected Behavior

The second snippet should give an error that the message is not constant.

### Bug

The second snippet compiles without error.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9414
Resolves #9406

## Summary

Each commit addresses one refactor or bug from the linked issues, though there are a few commits that don't relate to any of those:
- moving docs to the module position
- changing a function from `pub(crate)` to private
- extracting a `get_blocks_within_empty_loop` helper to keep each function short

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify/call.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify/call.rs
@@ -293,7 +293,8 @@ pub(super) fn simplify_call(
                 panic!("ICE: static_assert called with wrong number of arguments")
             }
 
-            if !dfg.is_constant(arguments[1]) {
+            // Arguments at positions `1..` form the message and they must all be constant.
+            if arguments.iter().skip(1).any(|argument| !dfg.is_constant(*argument)) {
                 return SimplifyResult::None;
             }
 
--- a/test_programs/compile_failure/static_assert_dynamic_message/Nargo.toml
+++ b/test_programs/compile_failure/static_assert_dynamic_message/Nargo.toml
@@ -1,0 +1,7 @@
+[package]
+name = "static_assert_dynamic_message"
+type = "bin"
+authors = [""]
+compiler_version = ">=0.31.0"
+
+[dependencies]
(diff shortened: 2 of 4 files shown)


TomAFrench (approved):
LGTM

PR #9550, opened 2025-08-18
chore: remove redundant `enforce_bitshift_rhs_lt_bit_size`

asterite:
# Description

## Problem

No issue.

## Summary

While greenlighting the `remove_bit_shifts` pass I noticed `enforce_bitshift_rhs_lt_bit_size` existed twice and was almost duplicated but not quite (one uses a FunctionBuilder, the other one directly writes to the DataFlowGraph). I thought about extracting a common function for this but then realized that the call in `remove_bit_shifts` is not needed because the other one is inserted during SSA gen for any bitshift.

For example this Noir code:

```noir
fn main(x: u128) -> pub u128 {
    x << 4294967297
}
```

starts with this SSA:

```
acir(inline) fn main f0 {
  b0(v0: u128):
    constrain u1 0 == u1 1, "attempt to bit-shift with overflow"
    v4 = shl v0, u128 4294967297
    return v4
}
```

and after removing bitshifts it becomes:

```
  b0(v0: u128):
    constrain u1 0 == u1 1, "attempt to bit-shift with overflow"
    constrain u1 0 == u1 1, "attempt to bit-shift with overflow"
    v3 = cast v0 as Field
    v5 = mul v3, Field 8440413193081015954237343881579943487679465165173364518877740359351874239936
    v6 = truncate v5 to 128 bits, max_bit_size: 129
    v7 = cast v6 as u128
    return v7
}
```

This doesn't lead to incorrect code but there's no need for this redundant code either.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
@@ -5,9 +5,7 @@ use acvm::{FieldElement, acir::AcirField};
 use crate::ssa::{
     ir::{
         function::Function,
-        instruction::{
-            ArrayOffset, Binary, BinaryOp, ConstrainError, Endian, Instruction, Intrinsic,
-        },
+        instruction::{ArrayOffset, Binary, BinaryOp, Endian, Instruction, Intrinsic},
         types::{NumericType, Type},
         value::ValueId,
     },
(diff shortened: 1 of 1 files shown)


TomAFrench:
Yep, this is a bit of duplication I added while stripping down the amount of bitshift logic which is being left in SSA codegen.

I think I remember getting some test failures when I tried removing this but it should be able to be done (as the brillig bitshift opcode will implicitly constrains this)

asterite:
> I think I remember getting some test failures when I tried removing this but it should be able to be done (as the brillig bitshift opcode will implicitly constrains this)

Sounds good. Now tests seem to pass, let me know if you'd prefer me to close this (it's not very important, but I didn't know this was known).

TomAFrench:
Ugh, I just realised this removes the wrong copy of this function.

TomAFrench:
@asterite can you remove the copy in ssa_codegen? We want this check to be scoped more closely to bitshifts - in general we don't want ssa_codegen to have to inject multiple instruction to do a single operation.

PR #9560, opened 2025-08-19
chore: greenlight `array_set_optimization`

### LINKED ISSUE #9520: Greenlight `array_set` for audits
Prepare https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/array_set.rs for audits:

- [x] Code shall be clearly documented, covering:
  - [x] Purpose of the code
  - [x] Design decisions made
  - [x] Interfaces with other code (e.g. dependencies, downstream integrations)
  - [x] Expected (or the lack of) pre- and post-conditions
  - [x] If applicable, expected ACIR/Brillig differences or implications
- [x] Code shall be logically unit tested for every feature of the code

**Target Date:** 12 Sep 2025

aakoshh:
# Description

## Problem\*

Resolves #9520 

## Summary\*

* Added module documentation
* Some refactoring to reduce duplication
* Added unit tests
* Added a new pre-check condition to reject `IfElse`

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/array_set.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/array_set.rs
@@ -23,9 +23,9 @@ use crate::ssa::{
 use fxhash::{FxHashMap as HashMap, FxHashSet as HashSet};
 
 impl Ssa {
-    /// Map arrays with the last instruction that uses it
-    /// For this we simply process all the instructions in execution order
-    /// and update the map whenever there is a match
+    /// Finds the last instruction that writes to an array and modifies it
+    /// to do an in-place mutation instead of making a copy if there are
+    /// no potential shared references to it.
     #[tracing::instrument(level = "trace", skip(self))]
     pub(crate) fn array_set_optimization(mut self) -> Self {
         for func in self.functions.values_mut() {
@@ -44,11 +44,22 @@ impl Ssa {
 /// Pre-check condition for [Function::array_set_optimization].
 ///
 /// Panics if:
-///   - there already exists a mutable array set instruction.
+///   - An ACIR function contains more than 1 block, i.e. it hasn't been flattened yet.
+///   - There already exists a mutable array set instruction.
 #[cfg(debug_assertions)]
 fn array_set_optimization_pre_check(func: &Function) {
+    let reachable_blocks = func.reachable_blocks();
+
+    if func.runtime().is_acir() && !func.runtime().is_entry_point() {
+        assert_eq!(
+            reachable_blocks.len(),
+            1,
+            "Expected there to be 1 block remaining in Acir function for array_set optimization"
+        );
+    }
+
     // There should be no mutable array sets.
-    for block_id in func.reachable_blocks() {
+    for block_id in reachable_blocks {
         let instruction_ids = func.dfg[block_id].instructions();
         for instruction_id in instruction_ids {
             if matches!(func.dfg[*instruction_id], Instruction::ArraySet { mutable: true, .. }) {
@@ -61,7 +72,7 @@ fn array_set_optimization_pre_check(func: &Function) {
 /// Post-check condition for [Function::array_set_optimization].
 ///
 /// Panics if:
-///   - Mutable array_set optimization has been applied to Brillig function
+///   - Mutable array_set optimization has been applied to Brillig function.
 #[cfg(debug_assertions)]
 fn array_set_optimization_post_check(func: &Function) {
     // Brillig functions should be not have any mutable array sets.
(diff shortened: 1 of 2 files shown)


compiler/noirc_evaluator/src/ssa/opt/array_set.rs
```diff
@@ -171,34 +213,28 @@ impl<'f> Context<'f> {
...
+                        if self.dfg.type_of_value(*argument).is_array() {
+                            self.set_last_use(*argument, *instruction_id);
                         }
                     }
                 }
+                // Arrays loaded from input references might be shared with the caller.
                 Instruction::Load { address } => {
                     let result = self.dfg.instruction_results(*instruction_id)[0];
-                    if matches!(self.dfg.type_of_value(result), Array { .. } | Slice { .. }) {
+                    if self.dfg.type_of_value(result).is_array() {
```
asterite, line 228: I wonder if it would be better to call this method `is_array_or_slice`

compiler/noirc_evaluator/src/ssa/opt/array_set.rs
```diff
@@ -171,34 +213,28 @@ impl<'f> Context<'f> {
...
                     }
                 }
+                // Arrays nested in other arrays are a use.
                 Instruction::MakeArray { elements, .. } => {
                     for element in elements {
-                        if let Some(existing) =
-                            self.array_to_last_use.insert(*element, *instruction_id)
-                        {
-                            self.instructions_that_can_be_made_mutable.remove(&existing);
+                        if self.dfg.type_of_value(*element).is_array() {
```
asterite, line 237: Why was this `if` added?

compiler/noirc_evaluator/src/ssa/opt/array_set.rs
```diff
@@ -300,4 +336,120 @@ mod tests {
...
+                constrain u1 0 == u1 1
+                unreachable
+            }
+            ";
+        assert_ssa_does_not_change(src, Ssa::array_set_optimization);
+    }
+
+    #[test]
+    #[should_panic = "instruction exists before"]
+    fn does_not_mutate_array_with_if_else() {
```
asterite: Should this name say that it should panic on "if else"? Also maybe just calling `Ssa::array_set_optimization` at the end instead of using `assert_ssa_does_not_change`.

asterite (approved):
Looks great! I left a couple of optional comments.

compiler/noirc_evaluator/src/ssa/opt/array_set.rs
aakoshh: I think I agree, it's something I have to check every time, sort of background knowledge. Do you want me to do it here, or in another PR?

compiler/noirc_evaluator/src/ssa/opt/array_set.rs
aakoshh: It just made sense to me, the comment talks about nested arrays. I thought it was just an omission. The other instructions check it before looking at `array_to_last_use`.

compiler/noirc_evaluator/src/ssa/opt/array_set.rs
aakoshh: Can do. I thought this is what I want: that if there was this `IfElse`, where `v4` might actually be `v1`, then `v4` should not accidentally mutate `v1`, which is supposed to be returned as-is. Instead of handling this use case explicitly, we say that the pre-condition should panic if we have an `IfElse`. I wasn't sure how we treat pre-conditions: they are debug features after all. Maybe I should raise the panic inside the instruction handling logic instead as a non-debug feature. Or leave it on debug, but the test makes it clear that it's an assumption.

Anyway, I'll change it as you say if it's confusing like this.

compiler/noirc_evaluator/src/ssa/opt/array_set.rs
asterite: Either way is fine.

compiler/noirc_evaluator/src/ssa/opt/array_set.rs
asterite: I think either way is fine :-)

aakoshh: Renamed.

compiler/noirc_evaluator/src/ssa/opt/array_set.rs
aakoshh: There is actually quite a few methods that follow this pattern:

```rust
    /// True if this type is an array (or slice)
    pub(crate) fn is_array(&self) -> bool

    pub(crate) fn is_nested_slice(&self) -> bool // this should say `is_nested_array` to fit in

    /// True if this type is an array (or slice) or internally contains an array (or slice)
    pub(crate) fn contains_an_array(&self) -> bool

    /// Retrieves the array or slice type within this type, or panics if there is none.
    pub(crate) fn get_contained_array(&self) -> &Type 
```

I suggest we don't try to tackle this in this PR.

asterite: Wow, yeah, it's a bit more complex than I thought.

PR #9561, opened 2025-08-19
feat: remove redundant `enforce_bitshift_rhs_lt_bit_size`

asterite:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*


asterite:
@TomAFrench In the end a fuzzing test started to fail with this change. It turns out that the SSA interpreter had the old behavior for shifting so I updated here in this PR. If I understood things correctly, shift overflow will always lead to an overflow error now, right? Instead of sometimes returning zero or minus one. Or, well, I think in some cases it can return zero but if we know for sure it'll overflow then it will produce an error (and that's always the case in the SSA interpreter as we always know the values).

TomAFrench:
>  If I understood things correctly, shift overflow will always lead to an overflow error now, right? Instead of sometimes returning zero or minus one. Or, well, I think in some cases it can return zero but if we know for sure it'll overflow then it will produce an error (and that's always the case in the SSA interpreter as we always know the values).

Yep, the current situation is if you ever end up bitshifting by a number of bits equal or greater to the size of the type then you get an assertion error.

PR #9566, opened 2025-08-19
fix: bind trait impl func generics to trait func generics during monomorphization

### LINKED ISSUE #9504: Monomorphisation fails to determine numeric generics passed through generic trait methods.
Reproduction pulled and minimised from https://github.com/noir-lang/noir_json_parser/pull/90.

Here's a case where we're passing a numeric generic through turbofish into a function where we use it to call a generic method on a trait. This trait then uses the numeric generic as a value, at which point the compiler fails to determine its value during monomorphisation.

```noir
trait Bar {
    fn baz<let N: u32>(self);
}

impl Bar for Field {
    fn baz<let N: u32>(self) {
        assert(N != 0);
    }
}

fn foo<let N: u32, K: Bar>(key: K) {
    key.baz::<N>()
}

fn main() {
    foo::<2, _>(1);
}
```

```
error: Could not determine array length `_`, encountered error: `Expected a constant, but found `_``
  ┌─ temp/src/main.nr:7:16
  │
7 │         assert(N != 0);
  │                -
  │
```

asterite:
# Description

## Problem

Resolves #9504

## Summary

See the comments on `bind_trait_impl_func_generics_to_trait_func_generics` for an explanation.

## Additional Context

There are a few other things in this PR that I found while fixing this:
1. `Type::Constant` couldn't be displayed via `to_display_ast`. I think when the code was written type expressions didn't exist so now that they do I decided to handle that case (I tried to debug something while coding this and couldn't).
2. While coding this and debugging things I noticed trait implementation were resolved twice during monomorphization. It turns out that `patch_debug_instrumentation_call` called `expr`, repeating a call that's done right before calling that method. I changed this to pass the result of `expr` to avoid this duplication (makes things easier to debug and also faster to compile, I guess)
3. A couple of frontend tests had `should_panic` on them. When running these tests locally they would generate test programs but I think they weren't required to run because of the `should_panic`. I changed these to use the usual `check_errors`.

## Documentation

Check one:

# PR Checklist


jfecher:
> While coding this and debugging things I noticed trait implementation were resolved twice during monomorphization. It turns out that patch_debug_instrumentation_call called expr, repeating a call that's done right before calling that method. I changed this to pass the result of expr to avoid this duplication (makes things easier to debug and also faster to compile, I guess)

I ran into this in https://github.com/noir-lang/noir/pull/9484 as well. There's really no need for that function to re-monomorphize the call expression so I'm unsure why it was ever written that way.

compiler/noirc_frontend/src/monomorphization/debug.rs
```diff
@@ -39,10 +39,10 @@ impl Monomorphizer<'_> {
     /// different types for each instance at runtime.
     pub(super) fn patch_debug_instrumentation_call(
         &mut self,
+        original_func: Box<Expression>,
```
jfecher: Nitpick: we don't need to move `original_func` into this function either. Passing a reference should eliminate the clone when calling this function. See https://github.com/noir-lang/noir/pull/9484/files#diff-cf8aebeffeb8bd0425789bfa0bfcb832252a60340de52377b8bbc1a2ee240c1dR43

compiler/noirc_frontend/src/monomorphization/mod.rs
```diff
@@ -2548,6 +2548,14 @@ fn resolve_trait_item_impl(
                     // This is similar to what's done in `verify_trait_constraint` in the frontend.
                     let mut bindings = interner.get_instantiation_bindings(expr_id).clone();
                     bindings.extend(instantiation_bindings);
+
+                    bind_trait_impl_func_generics_to_trait_func_generics(
```
jfecher, line 2552: Not blocking but I wonder if the impl -> trait bindings should be in the instantiation bindings to begin with.

Also - have you seen `NodeInterner::trait_to_impl_bindings`? This seems to be the reverse of that.

jfecher (approved):
:+1:

compiler/noirc_frontend/src/monomorphization/mod.rs
asterite: > Not blocking but I wonder if the impl -> trait bindings should be in the instantiation bindings to begin with.

Indeed! That's what I wondered too. But I think before monomorphization we know the call resolves to the trait method but we don't know the trait impl so we can't do that mapping... I think!

Another thing I thought is maybe the trait method and the corresponding trait impl method should have the same type variables for the same generics? That would avoid having to do this mapping. 

> Also - have you seen NodeInterner::trait_to_impl_bindings? This seems to be the reverse of that.

I did, and I first tried applying those bindings but then I realized they are the reverse of what we need, as you mention.

As I wrote above, maybe we should try to use the same type variables for the same generics but I didn't know if that would cause any issues so I ended up doing the mapping during monomorphization. I'm still not sure that's the best way to solve this, though...

compiler/noirc_frontend/src/monomorphization/debug.rs
asterite: Done! Now it's written the same way as in your PR.

PR #9305, opened 2025-07-24
fix(mem2reg): Register aliases when the `IfElse` result in a reference

aakoshh:
# Description

## Problem\*

Resolves #9303 

## Summary\*

Added handling for `Instruction::IfElse` in `analyze_instruction` of the `mem2reg` pass, so that if the result contains a reference, then both the `then_value` and `else_value` are considered aliases. This way if the result is used later, its aliases keep alive the stores to those addresses, if any. 

## Additional Context

This was the outline of the SSA:
```rust
brillig(inline) impure fn func_5 f1 {
  b0(v2: &mut [Field; 1], v3: &mut u1, v4: Field, v5: u1, v6: u1, v7: &mut u32):
    ...
    v11 = allocate -> &mut u1                     	
    store v5 at v11                              // store a value to v5; this store got removed
    ...                         	
    jmpif v15 then: b1, else: b2
  b1():
    ...
    v27 = load v3 -> u1                          // in a recursive call, this comes from v21
  b2():
    ...        	
    v21 = if v19 then v3 else (if v20) v11       // v21 is a reference that is either v3 or v11
    v22 = load v11 -> u1                         // v22 is loaded from v11
    v24, v25 = call f1(..., v21, ..., v22, ...) -> (Field, u1)	// passing both v21 and v22 to the call
    jmp b3(v24, v25)
  ...
}
```

During `mem2reg`, the `call f1(..., v21, ..., v22, ...)` instruction was turned into `call f1(..., v21, ..., v5, ...)` because it recognised that the last stored value to `v11` was `v5`, but otherwise it considered `v11` unused, which resulted in `store v5 at v11` being removed. When `v11` was passed to the recursive call as `v21` and subsequently loaded, it was found empty. 

By registering `v11` and `v3` as aliases of `v21`, the existing handling for `Instruction::Call` marks `v11` as being used in an instruction, which results in the `store` being kept.

## Documentation\*

Check one:
- [x] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [x] I have tested the changes locally.
- [ ] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #9567, opened 2025-08-19
fix(mem2reg): Consider aliases of a loaded address to be loaded from as well

### LINKED ISSUE #9538: Execution crash: `value is not typed as brillig usize: 0: field`
### Aim

Reproduce a nightly failure from https://github.com/noir-lang/noir/actions/runs/16955085712/job/48055401197

The following program crashes during execution:
```rust
unconstrained fn main(a: bool, mut b: (bool, (str<4>, bool), (str<4>, bool))) {
    if a {
        func_2((if a { (&mut b.2) } else { (&mut b.1) }))
    } else {
        func_2((if a { (&mut b.2) } else { (&mut b.1) }))
    }
}
unconstrained fn func_2(a: &mut (str<4>, bool)) {
    println(*a)
}

```

`Prover.toml`:
```toml
a = true
b = [true, ["abcd", false], ["efgh", false]]
```

### Expected Behavior

Shouldn't crash, it should print `(efgh, false)`. 

### Bug

```console
Message:  value is not typed as brillig usize: 0: field
Location: /Users/aakoshh/Work/aztec/noir/acvm-repo/brillig_vm/src/memory.rs:75
```

### To Reproduce

```
RUST_LOG=debug NOIR_AST_FUZZER_SHOW_AST=1 NOIR_AST_FUZZER_SEED=0x8b77b79600100000 cargo test -q -p noir_ast_fuzzer_fuzz orig_vs_morph
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

### NoirJS Version

_No response_

_No response_

Maybe

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9538 

## Summary\*

Fixes `mem2reg` to add the aliases of the `address` in a `Load` instruction to the `last_loads` of the `PerFunctionContext`. 

## Additional Context

In https://github.com/noir-lang/noir/pull/9305 I added handling of `IfElse` instructions to `mem2reg` where if the result of the `IfElse` was a reference, then the `then_value` and the `else_value` were considered to be aliases of that result address. For example consider this SSA in the new integration test:

```rust
After Simplifying (4) (step 31):
  b0(v0: u1, v1: [u8; 3], v2: [u8; 3]):
    v4 = allocate -> &mut [u8; 3]
    store v1 at v4
    v5 = allocate -> &mut [u8; 3]
    store v2 at v5
    v6 = not v0
    jmpif v0 then: b1, else: b2
  b1():
    v9 = not v0
    v10 = if v0 then v5 else (if v9) v4
    v11 = load v10 -> [u8; 3]                     	// src/main.nr:5:9
    inc_rc v11                                    	// src/main.nr:5:9
    jmp b3(v11)
  b2():
    v7 = if v0 then v5 else (if v6) v4            	// src/main.nr:3:9
    v8 = load v7 -> [u8; 3]                       	// src/main.nr:5:9
    inc_rc v8                                     	// src/main.nr:5:9
    jmp b3(v8)
  b3(v3: [u8; 3]):
    return v3
}
```

In `v10 = if v0 then v5 else (if v9) v4`, `v5` and `v4` are registered as aliases of `v10`. 

In #9305 this had the effect in the handling of `Call`: for example if `v10` was passed to a call, then all of its aliases would be added to `instruction_references`. In this bug, however, that was not enough. 

The problem was we removed the stores to `v5` and `v4`. The final decision whether to keep them looks like this:
```rust
                if !self.last_loads.contains_key(store_address)
                    && !store_alias_used
                    && !is_dereference
                {
                    self.instructions_to_remove.insert(*store_instruction);
                }
```
I focused a lot on `store_alias_used`, but that doesn't apply here: when we consider whether to remove the store to `v4`, it doesn't have an alias that is used, instead, `v4` is an alias of `v10`. I'm not sure why this relationship isn't bidirectional, but it looked to me like `last_loads` is the key here: normally we would be loading `v4` or `v5`, but instead we are loading them indirectly due to the `IfElse`. I thought it makes sense that if what we load has aliases, we might consider them being loaded as well, which directly prevents stores to them from being removed.

Conveniently `last_loads` acts across blocks, while `Block::last_store` does not, so even though `Load` marks the address as used, which marks all of its aliases as used, which removes the `last_store` from being considered for removal, it doesn't affect the previous `Block`, where it will remain a candidate and be removed anyway.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
@@ -458,15 +458,20 @@ impl<'f> PerFunctionContext<'f> {
                 let result = self.inserter.function.dfg.instruction_results(instruction)[0];
                 references.remember_dereference(self.inserter.function, address, result);
 
-                // If the load is known, replace it with the known value and remove the load
+                // If the load is known, replace it with the known value and remove the load.
                 if let Some(value) = references.get_known_value(address) {
                     let result = self.inserter.function.dfg.instruction_results(instruction)[0];
                     self.inserter.map_value(result, value);
                     self.instructions_to_remove.insert(instruction);
                 } else {
+                    // We don't know the exact value of the address, so we must keep the stores to it.
                     references.mark_value_used(address, self.inserter.function);
-
+                    // Remember that this address has been loaded, so stores to it should not be removed.
                     self.last_loads.insert(address, (instruction, block_id));
+                    // Stores to any of its aliases should also be considered loaded.
+                    references.for_each_alias_of(address, |_, alias| {
+                        self.last_loads.insert(alias, (instruction, block_id));
+                    });
                 }
 
                 // Check whether the block has a repeat load from the same address (w/ no calls or stores in between the loads).
--- a/compiler/noirc_evaluator/src/ssa/opt/mem2reg/block.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/mem2reg/block.rs
@@ -186,11 +186,21 @@ impl Block {
         }
     }
 
+    /// Forget the last store to an address and all of its aliases, to eliminate them
+    /// from the candidates for removal at the end.
+    ///
+    /// Note that this only affects this block: when we merge blocks we clear the
+    /// last stores anyway, we don't inherit them from predecessors, so if one
+    /// block stores to an address and a descendant block loads it, this mechanism
+    /// does not affect the candidacy of the last store in the predecessor block.
     fn keep_last_stores_for(&mut self, address: ValueId, function: &Function) {
         self.keep_last_store(address, function);
         self.for_each_alias_of(address, |t, alias| t.keep_last_store(alias, function));
     }
 
+    /// Forget the last store to an address, to remove it from the set of instructions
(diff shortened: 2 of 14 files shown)


aakoshh:
To replicate the test failure in https://github.com/noir-lang/noir/actions/runs/17078806369/job/48427962743?pr=9567 : 
```console
❯ cargo run -q -p nargo_cli -- debug --force-brillig --expression-width 3
[regression_9538] Starting debugger
Debugger ready to receive messages..
At opcode 0:0 :: BRILLIG CALL func 0: inputs: [EXPR [ (1, _0) 0 ], [EXPR [ (1, _1) 0 ], EXPR [ (1, _2) 0 ], EXPR [ (1, _3) 0 ]], [EXPR [ (1, _4) 0 ], EXPR [ (1, _5) 0 ], EXPR [ (1, _6) 0 ]]], outputs: [[_7, _8, _9]]
> continue
(Continuing execution...)
Message:  assertion failed: field_as_bytes.into_iter().all(|b| b == 0)
Location: /Users/aakoshh/Work/aztec/noir/compiler/noirc_printable_type/src/lib.rs:403

```

PR #9574, opened 2025-08-19
chore(ssa): Update comments on `loop_invariant` for audit and some missing unit tests

### LINKED ISSUE #9513: Greenlight `loop_invariant` for audits
Prepare https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs for audits:

- [ ] Code shall be clearly documented, covering:
  - [ ] Purpose of the code
  - [ ] Design decisions made
- [ ] Code shall be logically unit tested for every feature of the code

**Target Date:** 22 Aug 2025

vezenovm:
# Description

## Problem\*

Working towards #9513. I am going to add more use of the SSA interpreter to check that side effects have been preserved in more unit tests.

## Summary\*

I did the following:
- Updated the doc comments to account for loop bound simplifications and ACIR/Brillig differences
- Added a couple unit tests for functionality I felt was missing

The loop_invariant module is actually quite well documented and thoroughly tested. I will probably take another look before the week is over to see if there are any more unit tests I could add. Only then will I resolve #9513.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
@@ -1,10 +1,34 @@
-//! The loop invariant code motion (LICM) pass moves code from inside a loop to before the loop
-//! if that code will always have the same result on every iteration of the loop.
+//! The loop invariant code motion (LICM) pass performs two related optimizations:
+//! 1. Loop-Invariant Code Motion: Moves computations that produce the same result
+//!    on every iteration out of the loop and into its pre-header.
+//! 2. Loop-Bounds Simplification: Simplifies computations inside the loop body
+//!    using information derived from induction variables and loop bounds. This optimization
+//!    is essentially constant folding using induction variables.
 //!
+//! ## Preconditions
+//! - Input program is in appropriate SSA form. Meaning that loop headers expose induction variables via block parameters.
+//! - Each loop has a single header block, and at most one exit (except `break`).
+//! - Loops are well-formed with a pre-header; otherwise the pass skips them.
+//!
+//! ## Post-conditions
+//! - All loop-invariant instructions that are safe to hoist are moved to the loop pre-header.
+//! - Instructions inside loops may be simplified if loop bounds or induction variable
+//!   constraints allow (e.g. replacing comparisons with constants).
+//! - Program semantics are preserved:
+//!   - Values computed remain identical to the unoptimized program.
+//!   - Side effects occur in the same order as before.
+//!   - Potential runtime errors (e.g. out-of-bounds, division by zero) occur only if they would have in the original program.
+//! - Control dependence is respected: instructions whose effects depend on runtime conditions
+//!   remain in the loop unless proven safe.
+//!
+//! ## Design
 //! To identify a loop invariant, check whether all of an instruction's values are:
 //! - Outside of the loop
 //! - Constant
 //! - Already marked as loop invariants
+//!
+//! If we know that an invariant will always be executed (a loop's bounds are not dynamic and the upper bound is greater than its lower bounds)
+//! we then hoist that invariant into the loop pre-header.
 //!
 //! We also check that we are not hoisting instructions with side effects.
 //! However, there are certain instructions whose side effects are only activated
@@ -44,6 +68,21 @@
 //!
 //! We then can store the PDFs for every block as part of the context of this pass, and use it for checking control dependence.
 //! Using PDFs gets us from a worst case n^2 complexity to a worst case n.
+//!
+//! ## ACIR vs Brillig
+//! - On ACIR, LICM operates only on pure value computations.
+//! - On Brillig, additional reference-counting rules apply. For example, hoisting [Instruction::MakeArray]
+//!   requires inserting an [Instruction::IncrementRc] to preserve reference semantics if the array
+//!   may later be mutated.
+//!
+//! ### Simplification from Loop Bounds
+//! We analyze induction variables and loop bounds to simplify instructions.
+//! - Replacing conditions with constants when bounds make the condition always true/false.
+//! - Simplifying arithmetic expressions involving induction variables (e.g., comparisons against loop bounds)
+//!
+//! Simplification is attempted before hoisting. This maximizes opportunities for
+//! eliminating redundant computations entirely (by replacing them with constants),
+//! and reduces the amount of code even considered for hoisting.
 use crate::ssa::{
     Ssa,
     ir::{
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -1,11 +1,35 @@
...
+//! 1. Loop-Invariant Code Motion: Moves computations that produce the same result
+//!    on every iteration out of the loop and into its pre-header.
+//! 2. Loop-Bounds Simplification: Simplifies computations inside the loop body
+//!    using information derived from induction variables and loop bounds. This optimization
+//!    is essentially constant folding using induction variables.
 //!
+//! ## Preconditions
+//! - Input program is in appropriate SSA form. Meaning that loop headers expose induction variables via block parameters.
+//! - Each loop has a single header block, and at most one exit (except `break`).
+//! - Loops are well-formed with a pre-header; otherwise the pass skips them.
```
jfecher: These sound like general ssa well-formedness properties. Do we want to list them on the Ssa struct instead maybe? Otherwise we should probably add these preconditions to at least the unrolling pass as well.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -1,11 +1,35 @@
...
+//!
+//! ## Post-conditions
+//! - All loop-invariant instructions that are safe to hoist are moved to the loop pre-header.
```
jfecher: Do we want to guarantee that we hoist _all_ safe instructions?
Moreover, what does safe mean in this context? I could guess but I think it'd be useful to specify.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -1,11 +1,35 @@
...
+//! - Instructions inside loops may be simplified if loop bounds or induction variable
+//!   constraints allow (e.g. replacing comparisons with constants).
```
jfecher: This sounds more like it should go in the description of the pass rather than a post-condition

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -1,11 +1,35 @@
...
+//! - Program semantics are preserved:
```
jfecher: Do we have any passes where program semantics are not expected to be preserved?

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -1,11 +1,35 @@
...
+//!   - Values computed remain identical to the unoptimized program.
```
jfecher: Same here - I'd expect this for all SSA passes

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -1,11 +1,35 @@
...
+//!   - Side effects occur in the same order as before.
```
jfecher: This one is also expected for all passes but may be more useful to state as something this pass needs to take care to preserve

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -1,11 +1,35 @@
...
+//!   - Potential runtime errors (e.g. out-of-bounds, division by zero) occur only if they would have in the original program.
```
jfecher: Same here - I'd expect this for all SSA passes

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -1,11 +1,35 @@
...
+//! - Control dependence is respected: instructions whose effects depend on runtime conditions
+//!   remain in the loop unless proven safe.
```
jfecher: > unless proven safe

to hoist out of the loop, I assume

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -238,6 +278,8 @@ impl<'f> LoopInvariantContext<'f> {
         self.current_pre_header.expect("ICE: Pre-header block should have been set")
     }
 
+    /// Perform loop invariant code motion and induction variable analysis.
+    /// See [`loop_invariant`][self] module for more information.
```
jfecher, line 277: Will `[self]` here actually link to the module and not the `LoopInvariantContext` type?

vezenovm: It will link to the module

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: Yes I can add that

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: I guess that is fair yes that is expected of all passes.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: I think I just may leave the control dependence comment to discuss we expect execution ordering to be preserved

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: It is mentioned in the description of the pass. Perhaps these pre post conditions should be mentioned after the full description?

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: Like in https://github.com/noir-lang/noir/pull/9574#discussion_r2288813998, "safe" is mentioned during the design of the pass. I will move the pre and post conditions after the pass design as to be clear.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: Yes this is essentially a restatement of well formed loops. I agree perhaps these should just be on the Ssa itself and we link to that here. With regards to the pre-header, I have kept that line and updated the comment. Sometimes we may not be able to determine the pre-header and that should still be a well formed loop.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: I removed this stuff as it is implicit as you mentioned. I have left the line about respecting control dependence.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
vezenovm: I ended up moving these conditions to occur after the pass' design description as to avoid re-defining terms.

jfecher (approved):

PR #9585, opened 2025-08-20
fix: incorrect max bit size in `remove_bit_shifts`

### LINKED ISSUE #9541: ACIR != Brillig: `Field failed to decompose into specified 4 limbs`
### Aim

Reproduce nightly failure from https://github.com/noir-lang/noir/actions/runs/17030224759/job/48271592149

The following program fails in ACIR and returns 0 in Brillig:
```rust
global G_A: u8 = 245_u8;
fn main(a: u8) -> pub u8 {
    (((a as u32) << ((G_A - a) as u32)) as u8)
}
```

`Prover.toml`:
```
a = "0x00000000000000000000000000000000000000000000000000000000000000db"
```

### Expected Behavior

ACIR and Brillig should agree. Instead ACIR fails, while Brillig returns 0.

### Bug

```console
error: Assertion failed: Field failed to decompose into specified 4 limbs
  ┌─ src/main.nr:3:7
  │
3 │     (((a as u32) << ((G_A - a) as u32)) as u8)
  │       --------------------------------
  │
  = Call stack:
    1. src/main.nr:3:7

Failed assertion

❯ cargo run -q -p nargo_cli -- execute --silence-warnings --force-brillig

```

### To Reproduce

```shell
NOIR_AST_FUZZER_SEED=0x0eff72dc00100000 cargo test -q -p noir_ast_fuzzer_fuzz acir_vs_brillig
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9541

## Summary

When decomposing an exponent into bits we always need to take into account the maximum number of bits the exponent type can hold. Previously if the exponent was casted from a smaller type, for example from `u8` to `u32`, we'd use the maximum number of bits of values in `u8`, which is 8, then take `ilog2` of that, which is 3, then add 1 to get 4. But values that fit in u8 that are less than 32, like 26, need more than 4 bits to be represented. Given that the maximum value to shift an u32 is 32, it makes sense that the maximum bits is the bits to represent 32.

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -189,7 +189,7 @@ impl Context<'_, '_, '_> {
         }
 
         let to_bits = self.context.dfg.import_intrinsic(Intrinsic::ToBits(Endian::Little));
-        let max_exponent_bits = self.context.dfg.get_value_max_num_bits(exponent).ilog2() + 1;
+        let max_exponent_bits = self.context.dfg.type_of_value(exponent).bit_size().ilog2();
```
asterite: Because `ilog2` returns log2 truncated, my guess is that if `get_value_max_num_bits` returned something like 26 then log2 of that is ~4.7, so ilog2 returns 4. But we actually need 5 bits for that so `+ 1` was added.

Now that we always rely on the type's bit size, which is always a power of 2, ilog2 and the actual log2 will always be the same.

Note: for `u1` this will be zero, but shifting `u1` requires the exponent to always be zero which is kind of useless anyway...

vezenovm: > Note: for `u1` this will be zero, but shifting `u1` requires the exponent to always be zero which is kind of useless anyway...

Yes it is useless, but it be good to confirm an identity operation works as expected and does not break. We could do this as part of the internal audit or a separate PR though as this is a separate bug fix.

vezenovm: Maybe we could add a comment here about the thought process with why `ilog2` gives us the exponents bit size. Would make it clearer for external contributors and auditors.

vezenovm (approved):

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
asterite: Until I figured shifting by anything other than zero fails I wrote a test program that checks how bitshifts works for `u1`. I can include such a test program, it could check that shifting 0 or 1 with 0 always gives 0. What do you think?

vezenovm: > I can include such a test program, it could check that shifting 0 or 1 with 0 always gives 0.

1 << 0 should should always give 1. Shifting by zero is an identity operation. 

Yeah sure you can add it here if you like. I was thinking we could also have some generated unit tests that {value}_{typ} << 0 (or >> 0) always gives {value}. Those tests being more involved is why I thought it can come later so we can get this fix in. Ideally we would test that the interpreter gives the same identity result before and after this optimization pass.

asterite: Done! While documenting this I realized there's an optimization opportunity when the exponent was upcasted from `u1` so I handled that too. It's likely a very rare case, but why not :-)

asterite: Sounds good, they can probably come in when auditing this pass.

PR #9586, opened 2025-08-20
chore: `--no-ssa-locations` for `nargo interpret` and show negative values when printing SSA

asterite:
# Description

## Problem

No issue, just a couple of things I noticed.

## Summary

1. `--no-ssa-locations` didn't work for `nargo interpret`
2. A value like `let x: i8 = -53` was displayed as `i8 203` which was a bit confusing, so now negative values are printed as such

## Additional Context

I wonder if `--no-ssa-locations` should be the default and instead have `--with-ssa-locations`. I find that I usually don't want locations and remember too late when I already pasted an SSA in an issue or PR.

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/ssa/ir/printer.rs
```diff
@@ -140,6 +140,21 @@ fn value(dfg: &DataFlowGraph, id: ValueId) -> String {
     }
 }
 
+fn number(number: FieldElement, typ: &NumericType) -> String {
```
aakoshh, line 147: I think this function could use some docstrings and some basic unit tests to serve as examples.

compiler/noirc_evaluator/src/ssa/ir/printer.rs
```diff
@@ -140,6 +140,21 @@ fn value(dfg: &DataFlowGraph, id: ValueId) -> String {
...
+    if let NumericType::Signed { bit_size } = typ {
+        let bit_size = *bit_size;
+        let max = if bit_size == 128 { i128::MAX as u128 } else { (1 << (bit_size - 1)) - 1 };
+        if number.num_bits() > 128 || number.to_u128() > max {
```
aakoshh: Can you add a comment on the `number.num_bits() > 128` part?

tooling/ast_fuzzer/src/program/tests.rs
```diff
@@ -86,9 +86,9 @@ fn test_modulo_of_negative_literals_in_range() {
     assert_ssa_snapshot!(ssa, @r"
     acir(inline) fn main f0 {
       b0():
-        jmp b1(i64 18446744073709551612)
+        jmp b1(i64 -4)
       b1(v0: i64):
-        v3 = lt v0, i64 18446744073709551615
+        v3 = lt v0, i64 -1
```
aakoshh, line 91: I remember we discussed that multiple Field values can be mapped to e.g. `i8` value of `0`. We don't have any ambiguity when it comes to negatives, right?

compiler/noirc_evaluator/src/ssa/ir/printer.rs
asterite: Done!

I actually copied this code from another place where I needed this logic. Now I decided to make it a `FieldElement` method to remove this duplication.

I also moved another unrelated method to `FieldElement` (`to_short_hex_string`) because I also had to use it from different places and whenever I needed it I had a hard time remembering where it was (it was in noirc_printable_type, a bit unexpected).

tooling/ast_fuzzer/src/program/tests.rs
asterite: I think no. At least in the range of the bit_size (`0..2^bit_size - 1`) there's no ambiguity.

It could happen that the field value falls outside of that range. In that case I decided to format the number as-is instead of erroring because otherwise a test called `does_not_crash_on_truncated_division_with_large_denominators` fails, but I don't understand why that test exists or how it could happen (in my mind we should never produce such code).

compiler/noirc_evaluator/src/ssa/ir/printer.rs
asterite: In the end that was a bit wrong... or, well, I assumed that if it has more than 128 bits then it must be some negative value... for some unsigned integer? That doesn't make sense. So I changed it to format the number as-is in that case (well, almost: when the value doesn't fall in the expected range).

acvm-repo/acir_field/src/field_element.rs
```diff
@@ -255,6 +278,18 @@ impl<F: PrimeField> AcirField for FieldElement<F> {
...
+
+    fn to_short_hex(self) -> String {
+        if self.is_zero() {
+            return "0x00".to_owned();
+        }
+        let mut trimmed_field = self.to_hex().trim_start_matches('0').to_owned();
+        if trimmed_field.len() % 2 != 0 {
+            trimmed_field = "0".to_owned() + &trimmed_field;
+        }
+        "0x".to_owned() + &trimmed_field
```
aakoshh: It might avoid some allocations to do it like this:
```
let mut hex = "0x".to_string();
let trimmed_field = self.to_hex().trim_start_matches('0');
if trimmed_field.len() % 2 != 0 {
  hex.push('0');
}
hex.extend(&trimmed_field);
hex
```
not sure about the method names but you see what I mean, it saves 2 `.to_owned()` calls.

acvm-repo/acir_field/src/generic_ark.rs
```diff
@@ -70,6 +70,10 @@ pub trait AcirField:
 
     fn to_hex(self) -> String;
 
+    /// Returns the value of this field as a hex string with leading zeroes removed.
+    /// A singular '0' will be prepended as well if the trimmed string has an odd length.
+    fn to_short_hex(self) -> String;
```
aakoshh: Maybe it would be worth highlighting the difference between `to_hex` and `to_short_hex` here?

compiler/noirc_evaluator/src/ssa/ir/printer.rs
aakoshh: I'll probably never feel comfortable with bit operations, but I'll try to get comfortable asking seemingly dumb questions 😛

acvm-repo/acir_field/src/generic_ark.rs
asterite: Done! It's now a bit strange that one of them has the `0x` prefix and one doesn't, but at least it's documented. This should all be implemented with `Display` to `LowerHex`/`UpperHex` and formatting flags, but it's probably not worth it right now.

acvm-repo/acir_field/src/field_element.rs
asterite: Done!

vezenovm:
> I wonder if `--no-ssa-locations` should be the default and instead have `--with-ssa-locations`. I find that I usually don't want locations and remember too late when I already pasted an SSA in an issue or PR.

I agree this would be nice. I usually do not want locations.

Issue #9512, opened 2025-08-15
Greenlight `flatten_cfg` for audits


Savio-Sou:
Prepare https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs for audits:

- [x] Code shall be clearly documented, covering:
  - [x] Purpose of the code
  - [x] Design decisions made
  - [x] Interfaces with other code (e.g. dependencies, downstream integrations)
  - [x] Expected (or the lack of) pre- and post-conditions
  - [x] If applicable, expected ACIR/Brillig differences or implications
- [x] Code shall be logically unit tested for every feature of the code

**Target Date:** 22 Aug 2025

Close this Issue when the requirements above are met. That is, the code piece is ready for audits.

PR #9588, opened 2025-08-20
chore: Update flattening docs

jfecher:
# Description

## Problem\*

Working towards https://github.com/noir-lang/noir/issues/9512

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
@@ -1,3 +1,27 @@
+//! This file contains the SSA flattening pass - a required pass for ACIR to remove any remaining
+//! control-flow in the singular program-function, resulting in a single block containing the
+//! program logic.
+//!
+//! ACIR/Brillig differences within this pass:
+//!   - This pass is strictly ACIR-only and never mutates brillig functions.
+//!
+//! Conditions:
+//!   - Precondition: There is at most one constrained function (inlining has been performed).
+//!   - Precondition: The 0-1 constrained functions have no loops (unrolling has been performed).
+//!   - Postcondition: The 0-1 constrained functions now consist of only one block where the
+//!     terminator instruction is always a return.
+//!
+//! Relevance to other passes:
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -1,3 +1,27 @@
+//! This file contains the SSA flattening pass - a required pass for ACIR to remove any remaining
+//! control-flow in the singular program-function, resulting in a single block containing the
+//! program logic.
+//!
+//! ACIR/Brillig differences within this pass:
+//!   - This pass is strictly ACIR-only and never mutates brillig functions.
+//!
+//! Conditions:
+//!   - Precondition: There is at most one constrained function (inlining has been performed).
```
TomAFrench: This isn't strictly true as we have ACIR calls for use with folding, it's then either 0 ACIR functions if main entrypoint is a brillig function or >=1 if main entrypoint is ACIR function.

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -1,3 +1,27 @@
...
+//!   - Precondition: The 0-1 constrained functions have no loops (unrolling has been performed).
+//!   - Postcondition: The 0-1 constrained functions now consist of only one block where the
```
TomAFrench: Can we add this to the post-check function in this pass? We can enforce this pretty easily.

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
jfecher: Even more precisely it'd be:
- 0 acir functions if entry point is unconstrained
- 1 main acir function + 1 additional function for each acir function marked `#[fold]`, right?

TomAFrench: Yes, that's the most correct version of this.

PR #9596, opened 2025-08-21
fix(ssa): Do not simplify on lhs being zero for shifts

### LINKED ISSUE #9595: Min != Full: `Attempted to shift by 3073358952 bits on a type of bit size 8`
### Aim

General fuzzing.

The following program fails with `--minimal-ssa` and `interpret`, but passes with the full pipeline:

```rust
global G_B: u32 = 3073358952_u32;
fn main() -> pub u32 {
    let mut a = G_B;
    ((a - a) >> if a == 1494723815 { a } else { a })
}
```

It also fails with the full workflow if we replace `if a == 1494723815 { a } else { a }` with just `a`, or if we remove the `mut` and just use `let a = G_B;`.

### Expected Behavior

The minimal and the full SSA pipeline should return the same results.

### Bug

```console
$ cargo run -q -p nargo_cli -- execute --silence-warnings --force-brillig --minimal-ssa
error: Assertion failed: 'Attempted to shift by 3073358952 bits on a type of bit size 8'
  ┌─ src/main.nr:4:6
  │
4 │     ((a - a) >> if a == 1494723815 { a } else { a })
  │      ----------------------------------------------
  │
  = Call stack:
    1. src/main.nr:4:6

Failed assertion

$ cargo run -q -p nargo_cli -- execute --silence-warnings --force
```

### To Reproduce

```
NOIR_AST_FUZZER_SEED=0xed7ac18100100000 cargo test -q -p noir_ast_fuzzer_fuzz min_vs_full
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.11 noirc version = 1.0.0-beta.11+414fc8b5d7c3b3943a691cfad9879150d1fc2089 (git version hash: 1f164984dea1b53b982e719929589a5a96cb8b18, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #9595
Resolves https://github.com/noir-lang/noir/pull/9587#issuecomment-3210560032

Even if a bit shift of zero will always result in zero we want to maintain consistency that bit shifting by an amount larger than the type's size will result in an error.

## Summary\*

Remove the lhs_is_zero simplification from Shl/Shr

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify/binary.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify/binary.rs
@@ -293,10 +293,6 @@ pub(super) fn simplify_binary(binary: &Binary, dfg: &mut DataFlowGraph) -> Simpl
             }
         }
         BinaryOp::Shl | BinaryOp::Shr => {
-            if lhs_is_zero {
-                let zero = dfg.make_constant(FieldElement::zero(), lhs_type);
-                return SimplifyResult::SimplifiedTo(zero);
-            }
             if rhs_is_zero {
                 return SimplifyResult::SimplifiedTo(lhs);
             }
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/ir/dfg/simplify/binary.rs
```diff
@@ -345,40 +341,4 @@ mod tests {
         }
         ");
     }
-
-    #[test]
-    fn replaces_shl_on_zero_lhs_with_zero() {
```
asterite, line 350: We could maybe keep these tests but asserting they don't change anything, explaining why we don't do this.

vezenovm: I debated that but we could do that for any value so it felt strange to have this case up. But I could keep the test and add a comment.

asterite: Makes sense 👍

Issue #9393, opened 2025-08-04
mem2reg: `PerFunctionContext` `last_loads` could be changed to `HashSet`


asterite:
The definition of `PerFunctionContext`'s `last_loads` field is:

```noir
    last_loads: HashMap<ValueId, (InstructionId, BasicBlockId)>,
```

However, the only read instruction that is done on it is `last_loads.contains_key` which means that the value is not used and likely not needed.

We could then change `last_loads` to be `HashSet<ValueId>`.

jfecher:
Relevant: comment on Expression -> Container refactor:
https://github.com/noir-lang/noir/issues/9482#issuecomment-3215450705

jfecher:
This refers to the old mem2reg pass, its obsolete now that the old pass has been replaced

PR #9599, opened 2025-08-21
chore: simplify Expression in mem2reg

asterite:
# Description

## Problem

Resolves one refactor in #9393

## Documentation

Check one:

# PR Checklist


jfecher (approved):
I'm not 100% sure I agree with this refactoring since like I mentioned in a prior comment it was the intention of the Expression enum to express nested expressions which should come up from nested dereferences/array accesses like `****foo` or `bar[0][0][0]`. I think the fact that we don't ever use the recursiveness of the enum to be indicative of bugs we likely have handling aliases of these nested cases.

That being said, it is true that this better represents the code as we have it now and it's not _too_ much more effort to re-add in the future (maybe during this same audit) if we get such a bug whose fix involves properly using this enum as intended. If that doesn't happen though, it'd be simpler for external auditors not to worry about this unhandled case.

PR #9603, opened 2025-08-21
fix(ssa): Put some default in `Value::uninitialized` for references in the SSA interpreter

### LINKED ISSUE #9594: SSA Interpreter vs Full: `UninitializedReferenceValueLoaded`
### Aim

General fuzzing. 

The following program fails with the SSA interpreter after Flattening, (but succeeds on later passes after DIE):

```rust
fn main(a: i32, mut b: bool, c: i32) -> pub i32 {
    match a {
        10 => {
            let mut e: [&mut bool] = &[];
            b = *e[(1132210792_u32 % e.len())];
            c
        },
        _ => 2,
    }
}

```

`Prover.toml`:
```toml
a = 0
b = true
c = 1
return = 2
```

### Expected Behavior

All SSA passes are expected to be successful. 

### Bug

```console
$ cargo run -q -p nargo_cli -- interpret --silence-warnings --force --show-ssa

...
--- Interpreter result after Simplifying (step 18):
Ok(i32 2)
---
After Flattening (step 19):
  b0(v0: i32, v1: u1, v2: i32):
    v3 = allocate -> &mut u1
    v5 = eq v0, i32 10
    enable_side_effects v5
    v6 = make_array [] : [&mut u1]
    v7 = allocate -> &mut u32
    v8 = allocate -> &mut [&mut u1]
    v11 = mod u32 1132210792, u32 0               	// src/main.nr:5:21
    constrain u1 0 == v5, "Index out of bounds"   	// src/main.nr:5:21
    v13 = array_get v6, index v11 -> &mut u1      	// src/main.nr:5:18
    v14 = load v13 -> u1                          	// src/main.nr:5:18
    v15 = not v5
    enable_side_effects u1 1
    v17 = cast v5 as i32
    v18 = cast v15 as i32
    v19 = unchecked_mul v17, v2
    v21 = unchecked_mul v18, i32 2
    v22 = unchecked_add v19, v21
    return v22
}

--- Interpreter result after Flattening (step 19):
Err(Reference value `*v13 = None` loaded before it was first stored to)
---

```

It works with `execute`:
```console
[fuzz_testing] Circuit output: 2

```

### To Reproduce

```
RUST_LOG=debug NOIR_AST_FUZZER_SEED=0xffff29a400100000 cargo test -p noir_ast_fuzzer_fuzz pass_vs_prev
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.11 noirc version = 1.0.0-beta.11+414fc8b5d7c3b3943a691cfad9879150d1fc2089 (git version hash: 6e503316e9d4af08a5354300e70f0442526cf6ef, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9594

## Summary\*

1. Changes `Value::uninitialized` in the SSA interpreter to store `Some` default value for `Reference` types. This way a subsequent `Load` will not fail, which it currently does even if side effects are disabled, if the value has never been stored to.
2. Run an extra `dead_instruction_elimination` after the first `mem2reg` pass to get rid of any `Load` that isn't actually needed. This isn't needed to fix the problem, but it was left in because it has some positive effects on the bytecode size.
3. Fix `remove_unreachable_instructions` to insert a `store` after an `allocate` in `zeroed_value` when replacing an instruction that returned a reference. This way a `Load` in the interpreter won't fail if executed on the resulting SSA.

## Additional Context

In the integration test we have this code:
```rust
fn main(a: bool, mut b: bool, c: i32) -> pub i32 {
    if a {
        let mut e: [&mut bool] = &[];
        b = *e[(1132210792_u32 % e.len())];
        c
    } else {
        2
    }
}
```
`e` is an empty slice, so indexing it with anything should fail. However `a` will be `false`, so we should not hit this problem.

The SSA before the first `mem2reg` reflects the state where a constraint exists on the index, but it's not an always-fail one, and we never hit that block during evaluation:
```rust
After Inlining (step 7):
acir(inline) fn main f0 {
  b0(v0: u1, v1: u1, v2: i32):
    v4 = allocate -> &mut u1
    store v1 at v4
    jmpif v0 then: b1, else: b2
  b1():
    v6 = make_array [] : [&mut u1]
    v7 = allocate -> &mut u32
    store u32 0 at v7
    v9 = allocate -> &mut [&mut u1]
    store v6 at v9
    v10 = load v7 -> u32                          	// src/main.nr:4:17
    v12 = mod u32 1132210792, v10                 	// src/main.nr:4:17
    v13 = load v7 -> u32                          	// src/main.nr:4:17
    v14 = load v9 -> [&mut u1]                    	// src/main.nr:4:17
    v15 = lt v12, v13                             	// src/main.nr:4:17
    constrain v15 == u1 1, "Index out of bounds"  	// src/main.nr:4:17
    v17 = array_get v14, index v12 -> &mut u1     	// src/main.nr:4:14
    v18 = load v17 -> u1                          	// src/main.nr:4:14
    store v18 at v4                               	// src/main.nr:4:14
    jmp b3(v2)
  b2():
    jmp b3(i32 2)
  b3(v3: i32):
    return v3
}

--- Interpreter result after Inlining (step 7):
Ok(i32 2)
---
```

Note that `v17` gets from the empty slice, `v18` loads the value of the reference, and then it's `store`d at `v4`. 

After `mem2reg` the constraint changes into an always-fail one, however this has no effect on the rest of the instructions in the block. `mem2reg` also removes the `store v18 at v4`, but it cannot remove the `v18 = load v17` instruction.
```rust
After Mem2Reg (step 8):
acir(inline) fn main f0 {
  b0(v0: u1, v1: u1, v2: i32):
    v4 = allocate -> &mut u1
    jmpif v0 then: b1, else: b2
  b1():
    v6 = make_array [] : [&mut u1]
    v7 = allocate -> &mut u32
    v8 = allocate -> &mut [&mut u1]
    v11 = mod u32 1132210792, u32 0               	// src/main.nr:4:17
    constrain u1 0 == u1 1, "Index out of bounds" 	// src/main.nr:4:17
    v14 = array_get v6, index v11 -> &mut u1      	// src/main.nr:4:14
    v15 = load v14 -> u1                          	// src/main.nr:4:14
    jmp b3(v2)
  b2():
    jmp b3(i32 2)
  b3(v3: i32):
    return v3
}

--- Interpreter result after Mem2Reg (step 8):
Ok(i32 2)
```

When we reach the flattening step, an `enable_side_effects v0` appears before the `array_get`:
```rust
After Flattening (1) (step 19):
  b0(v0: u1, v1: u1, v2: i32):
    v3 = allocate -> &mut u1
    enable_side_effects v0
    v4 = make_array [] : [&mut u1]
    v5 = allocate -> &mut u32
    v6 = allocate -> &mut [&mut u1]
    v9 = mod u32 1132210792, u32 0                	// src/main.nr:4:17
    constrain u1 0 == v0, "Index out of bounds"   	// src/main.nr:4:17
    v11 = array_get v4, index v9 -> &mut u1       	// src/main.nr:4:14
    v12 = load v11 -> u1                          	// src/main.nr:4:14
    v13 = not v0
    enable_side_effects u1 1
    v15 = cast v0 as i32
    v16 = cast v13 as i32
    v17 = unchecked_mul v15, v2
    v19 = unchecked_mul v16, i32 2
    v20 = unchecked_add v17, v19
    return v20
}
```

With the changes to the SSA Interpreter, this now returns a `0` value in the reference, and `load v11` can succeed.

### DIE

Eventually the DIE pass removes this as it's not used. My observation was even though `mem2reg` could not remove the `load`, since it removed the `store`, if we run DIE now, it's also eliminated at an early stage (before flattening), when we are still in a position of evaluating only the blocks which are activated by the jumps. 

After the extra DIE pass, we have the following state:
```rust
After Dead Instruction Elimination (step 9):
acir(inline) fn main f0 {
  b0(v0: u1, v1: u1, v2: i32):
    jmpif v0 then: b1, else: b2
  b1():
    v7 = mod u32 1132210792, u32 0                	// src/main.nr:4:17
    constrain u1 0 == u1 1, "Index out of bounds" 	// src/main.nr:4:17
    jmp b3(v2)
  b2():
    jmp b3(i32 2)
  b3(v3: i32):
    return v3
}

--- Interpreter result after Dead Instruction Elimination (step 9):
Ok(i32 2)
```

This doesn't handle the case of dangling loads for values which are _used_; for those we still get a compilation error. It only brings forward the "healing" of the SSA by the DIE, so we don't get a period of SSA states which cannot be interpreted.

### Remove Unreachable Instructions

If we disable the new DIE pass, then the _Remove Unreachable Instructions_ pass handles the case in the following way:

```rust
After Remove Unreachable Instructions (1) (step 33):
  b0(v0: u1, v1: u1, v2: i32):
    v3 = allocate -> &mut u1
    v4 = make_array [] : [&mut u1]
    v5 = allocate -> &mut u32
    v6 = allocate -> &mut [&mut u1]
    enable_side_effects v0
    v9 = mod u32 1132210792, u32 0                	// src/main.nr:4:17
    constrain u1 0 == v0, "attempt to calculate the remainder with a divisor of zero"	// src/main.nr:4:17
    constrain u1 0 == v0, "Index out of bounds"   	// src/main.nr:4:17
    v11 = allocate -> &mut u1
    store u1 0 at v11
    v12 = load v11 -> u1                          	// src/main.nr:4:14
    enable_side_effects u1 1
    v14 = cast v0 as i32
    v15 = unchecked_mul v14, v2
    v17 = unchecked_add v15, i32 2
    return v17
}
```

Notice that `v11 = allocate -> &mut u1; store u1 0 at v11` replaced the `array_get` because of the conditional-fail constraints preceding it, and together the allows `load v11` to be interpreted. 

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/mod.rs
```diff
@@ -126,6 +126,9 @@ pub fn primary_passes(options: &SsaEvaluatorOptions) -> Vec<SsaPass> {
         ),
         // Run mem2reg with the CFG separated into blocks
         SsaPass::new(Ssa::mem2reg, "Mem2Reg"),
+        // Run DIE in case mem2reg left some Load it could not eliminate, but it's also unused
+        // and could cause issues after flattening, such as loading from empty slices.
+        SsaPass::new(Ssa::dead_instruction_elimination, "Dead Instruction Elimination"),
```
vezenovm: Won't this have the `flattened` flag set while this occurs before flattening? The only thing that seems to be saving us is that we are also checking whether an ACIR block has a single block, but this check seems very brittle.

aakoshh: Thanks, I completely forgot about that. Added a `die_pre_check` that calls the flattening post-check to keep this from happening by accident.

aakoshh:
@jfecher I changed `Value::uninitialized` to create reference types with `Some` default element instead of `None`. I see the interpreter uses `Value::uninitialized` in `interpret_array_get` when side effects are disabled. 

However, `remove_unreachable_instructions` replaces instructions with side effects with default values after always-fail instructions. This was something I added in https://github.com/noir-lang/noir/pull/9462 to get around a similar situation.

In the case of this integration test, similar to the description of that PR, we will have the `array_get` replaced with `allocate`, and in the case of the interpreter `allocate` doesn't use `uninitialized`, so it will still fail on `load`:

```
After Remove Unreachable Instructions (step 33):
  b0(v0: u1, v1: u1, v2: i32):
    v3 = allocate -> &mut u1
    v4 = make_array [] : [&mut u1]
    v5 = allocate -> &mut u32
    v6 = allocate -> &mut [&mut u1]
    enable_side_effects v0
    v9 = mod u32 1132210792, u32 0                	// src/main.nr:4:17
    constrain u1 0 == v0, "attempt to calculate the remainder with a divisor of zero"	// src/main.nr:4:17
    constrain u1 0 == v0, "Index out of bounds"   	// src/main.nr:4:17
    v11 = allocate -> &mut u1
    v12 = load v11 -> u1                          	// src/main.nr:4:14
    enable_side_effects u1 1
    v14 = cast v0 as i32
    v15 = unchecked_mul v14, v2
    v17 = unchecked_add v15, i32 2
    return v17
}
```

`load` failing when `store` is missing was a great way to detect problems with `mem2reg`. Maybe we can return default from `interpret_load` or insert a default in `interpret_allocate`, when side effects are disabled?

Although technically that would be wrong, since `requires_acir_gen_predicate` returns `false` for `Load` and `Allocate` as well 😕

jfecher:
@aakoshh it sounds like remove_unreachable_instructions produces invalid SSA then. Regardless of the value for `enable_side_effects`, `allocate` instructions are still always expected to be stored to

aakoshh:
Okay, I'll have a look at undoing that change then.

jfecher:
@aakoshh I'm not sure if it needs to be undone entirely - maybe it could be changed to also store a default value to the element?

aakoshh:
Yeah that's what I was thinking as well: insert a `store` after the `allocate`. Reverting would not have solved the problem that `array_get` on an empty array fails even with side effects disabled.

compiler/noirc_evaluator/src/ssa/opt/.remove_unreachable_instructions.rs.pending-snap
```diff
@@ -0,0 +1,69 @@
+{"run_id":"1756222193-51099000","line":629,"new":{"module_name":"noirc_evaluator__ssa__opt__remove_unreachable_instructions__test","snapshot_name":"replaces_references_following_conditional_constraint_with_allocate_and_store_of_default","metadata":{"source":"compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs","assertion_line":629,"expression":"ssa_string"},"snapshot":"acir(inline) predicate_pure fn main f0 {\n  b0(v0: u1, v1: u1, v2: i32):\n    v3 = make_array [] : [&mut u1]\n    enable_side_effects v0\n    constrain u1 0 == v0, \"Index out of bounds\"\n    v6 = array_get v3, index u32 0 -> &mut u1\n    v7 = load v6 -> u1\n    enable_side_effects u1 1\n    return v7\n}"},"old":{"module_name":"noirc_evaluator__ssa__opt__remove_unreachable_instructions__test","metadata":{},"snapshot":"acir(inline) predicate_pure fn main f0 {\n  b0():\n    enable_side_effects u1 0\n    return u32 0\n}"}}
```
jfecher: Looks like a `.pending-snap` file was accidentally committed

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -206,8 +209,11 @@ fn flatten_cfg_pre_check(function: &Function) {
 #[cfg(debug_assertions)]
 #[allow(dead_code)]
 pub(super) fn flatten_cfg_post_check(function: &Function) {
+    if !function.runtime().is_acir() {
```
jfecher, line 210: Can we remove `#[allow(dead_code)]` now?

compiler/noirc_evaluator/src/ssa/opt/die.rs
```diff
@@ -965,6 +968,15 @@ fn should_remove_store(func: &Function, flattened: bool) -> bool {
     flattened && func.runtime().is_acir() && func.reachable_blocks().len() == 1
 }
 
+/// Check pre-execution properties:
+/// * If `flattened` is set to `true` then CFG should have already been executed.
```
jfecher:
```suggestion
/// * Passing `flattened = true` will confirm the CFG has already been flattened into a single block for ACIR functions
```

PR #9608, opened 2025-08-21
feat: attempt to inline successors in `simplify_cfg`

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR adds a forwards looking check to whether we can inline the blocks successor within `simplify_current_block`.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/.github/workflows/test-rust-workspace.yml
+++ b/.github/workflows/test-rust-workspace.yml
@@ -109,45 +109,6 @@ jobs:
       - name: Check for pending snapshots
         run: just check-pending-snapshots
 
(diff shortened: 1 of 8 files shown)


asterite:
I think this looks good but it looks like the newly added test also passes in master so I'm not sure something new is done here.

TomAFrench:
@asterite That's true but I think we end up getting a small performance boost due to us inlining chains of blocks into one other in a tighter loop compared to before.

asterite (approved):

PR #9610, opened 2025-08-21
chore: some mem2reg refactors regarding expressions and aliases

asterite:
# Description

## Problem

Part of #9393

## Summary

This PR tries to make the code more meaningful by removing some duplicated code and trying to avoid leaking some implementation details (although mem2reg is not very bug and it mostly knowns how everything is actually implemented). But the goal is to make the code easier to read, with less `if`s.

Then I also found out that `is_store_alias_used` could be slightly refactored as well. Now it doesn't call the `any` method on `AliasSet`, which had checks like `value == Some(true)`, and I also think it might be slightly faster as we need to traverse the alias set just once.

## Additional Context

Now `AliasSet` has an `iter()` method. I wanted to have `impl IntoIterator for &AliasSet` so it can be iterated without having to call `.iter()` but I couldn't find a way to do it without having `IntoIterator` use `Box<dyn Iterator...>`, resulting in a heap allocation.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/mem2reg/block.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/mem2reg/block.rs
@@ -170,15 +170,11 @@ impl Block {
     }
 
     /// Iterate through each known alias of the given address and apply the function `f` to each.
-    pub(super) fn for_each_alias_of<T>(
-        &mut self,
-        address: ValueId,
-        mut f: impl FnMut(&mut Self, ValueId) -> T,
-    ) {
+    pub(super) fn for_each_alias_of<T>(&self, address: ValueId, mut f: impl FnMut(ValueId) -> T) {
         if let Some(expr) = self.expressions.get(&address) {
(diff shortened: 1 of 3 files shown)


compiler/noirc_evaluator/src/ssa/opt/mem2reg/block.rs
```diff
@@ -256,6 +240,13 @@ impl Block {
 
     pub(super) fn keep_last_load_for(&mut self, address: ValueId) {
         self.last_loads.remove(&address);
-        self.for_each_alias_of(address, |block, alias| block.last_loads.remove(&alias));
+
+        if let Some(expr) = self.expressions.get(&address) {
+            if let Some(aliases) = self.aliases.get(expr) {
```
asterite, line 245: Here I didn't use `get_aliases_for_value` because we'd have to clone the return value to change `self.last_loads`. So this change avoids one clone.

Issue #9464, opened 2025-08-11
mem2reg: store lost after returning mutable reference from if


asterite:
### Aim

This code:

```noir
unconstrained fn main(predicate: bool) {
    let num1 = &mut 0;
    let num2 = &mut 1;
    let num3 = if predicate { num1 } else { num2 };
    *num3 = 9;
    println(*num1);
    println(*num2);
    println(*num3);
}
```

prints this, regardless of the value of `predicate`:

```
0x00
0x01
0x09
```

which shouldn't be right as modifying `num3`'s value should modify one of the others.

### Expected Behavior

The program should print either `0x09 0x01 0x09` or `0x00 0x09 0x09`.

### Bug

It prints a value that is impossible to happen.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

None

### Would you like to submit a PR for this Issue?

None

asterite:
We get this before mem2reg (not the first mem2reg but one of them):

```noir
brillig(inline) predicate_pure fn main f0 {
  b0(v0: u1):
    v1 = allocate -> &mut Field
    store Field 0 at v1
    v3 = allocate -> &mut Field
    store Field 1 at v3
    v5 = not v0
    v6 = if v0 then v1 else (if v5) v3
    store Field 9 at v6
    v8 = load v1 -> Field
    constrain v8 == Field 9
    v9 = load v3 -> Field
    constrain v9 == Field 1
    return
}
```

After mem2reg we get this:

```noir
brillig(inline) predicate_pure fn main f0 {
  b0(v0: u1):
    v1 = allocate -> &mut Field
    store Field 0 at v1
    v3 = allocate -> &mut Field
    store Field 1 at v3
    v5 = not v0
    v6 = if v0 then v1 else (if v5) v3
    v7 = load v1 -> Field
    constrain v7 == Field 9
    v9 = load v3 -> Field
    constrain v9 == Field 1
    return
}
```

So, effectively, the store to `v6`, which results from an `if then else`, is lost.

Issue #9468, opened 2025-08-11
mem2reg: Trivial store/load to reference returned from array not resolved


vezenovm:
### Problem

Take this test:
```rust
    #[test]
    fn store_load_from_array_get() {
        let src = "
        brillig(inline) fn main f0 {
          b0():
            v1 = allocate -> &mut u1
            store u1 0 at v1
            v3 = make_array [v1] : [&mut u1]
            jmpif u1 1 then: b1, else: b2
          b1():
            jmp b3(u32 0, u32 1)
          b2():
            jmp b3(u32 0, u32 1)
          b3(v0: u32, v8: u32):
            constrain v0 == u32 0
            v6 = array_get v3, index v0 -> &mut u1
            store u1 0 at v6
            v7 = load v6 -> u1
            return v7
        }
        ";

        let ssa = Ssa::from_str(src).unwrap();
        let ssa = ssa.mem2reg();
        assert_normalized_ssa_equals(ssa, src);
    }
```

### Happy Case

I would expect the store and load in the return block to be resolved and v7 to be replaced with u1 0. It looks to be that because v6 is treated as having an unknown alias set that the store to v6 does not set a known value. However, there are no stores before load from v6 so we should be able to safely resolve the reference in this case.

### Workaround

None

### Project Impact

Nice-to-have

### Would you like to submit a PR for this Issue?

None

asterite:
Opening as the PR that fixed this was reverted.

PR #9611, opened 2025-08-21
fix: Fix if-else alias in mem2reg

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/9464

## Summary\*

After an if-else instruction, `then_value` and `else_value` should both be aliased to the resulting value but this wasn't handled at all previously.

## Documentation\*

Check one:

# PR Checklist\*


aakoshh:
So previously I set the result to have the then_value and the end_value as aliases, and now it goes the other way as well. I asked a couple of times why aliasing wasn’t bidirectional. Or it is, but we have to remember to manually code it every time?

asterite:
Ah, I mentioned the above because I wonder if there's a different way to represent these alias sets...

aakoshh:
It's possible that with this, https://github.com/noir-lang/noir/pull/9567 would not be required on the load side.

jfecher:
> So previously I set the result to have the then_value and the end_value as aliases, and now it goes the other way as well. I asked a couple of times why aliasing wasn’t bidirectional. Or it is, but we have to remember to manually code it every time?

It's because the way aliasing is handled it isn't quite bidirectional.

Using this as an example, if we have `v3 = if v0 then v1 else v2` where `v0: u1` and all others are references, then we know:
- `v3` aliases `v1` or `v2` (conservatively, we say it aliases both)
- `v1` is (conservatively) aliased by `v3`
- `v2` is (conservatively) aliased by `v3`

So if we see:
```
store Field 0 in v3
v5 = load v1 -> Field
```
We can't simplify out the load in case `v3` refers to it.
However, we can simplify it out if the load refers to the same reference:
```
store Field 0 in v3
v5 = load v3 -> Field
```
In this case `v3` is still aliased to both `v1` and `v2` yet `v3` (and only `v3`) has a known value.

I am doubting our representation of this all though. E.g. maybe the aliases list should be more separate from whether a value is known? Such that the alias list is shared but we still track known values per-reference? The Expression enum is also maybe underused and seems questionable (can we always identify which Expression a value id should refer to?) and we also track aliases outside of the `Block` structure in `aliased_references`.

asterite:
> Such that the alias list is shared but we still track known values per-reference?

That's a good point. It seems we do track known values per-reference, in the `references` field of a `Block`. However, when we go to fetch an address' known value, if that references has aliases then we say we don't know what that value is. Also when we set the value of an address to point to a `ReferenceValue` we might either clear out all references or clear out all aliases, but we never remember that for this particular address we know the value. See https://github.com/noir-lang/noir/issues/9468

I'll try a PR changing that to see if it works. It would at least decouple `references` from `expressions` and `aliases` a bit.

jfecher:
Ah - one more thing. This is maybe obvious but I'll mention it anyway.

Even if we want setting an alias to be bidirectional (so that when we alias `v3` with `v1` and `v2` above, `v1` and `v2` are also aliased with `v3`), this does not mean they actually share the same alias list. Since `v3` may alias either, but `v1` is still known not to alias `v2` and vice-versa.

PR #9613, opened 2025-08-22
feat(mem2reg): address last known value is independent of its aliases

### LINKED ISSUE #9468: mem2reg: Trivial store/load to reference returned from array not resolved

asterite:
# Description

## Problem

Resolves #9468

## Summary

This PR should be an optimization (we'll see once CI runs) but the main purpose is to simplify the logic in mem2reg:
1. When we set an address to a known or unknown value we might also say that we don't know what the value of each of its aliases is. However, we definitely know the value in that particular address. That's the first commit and what fixes #9468. In here, when we want to determine if we know the value behind an address there's no need to check if the address has aliases, we can directly check if the value is in `references`. The reason is that if it had aliases we should have already cleared them in `set_value`.
2. The second commit is another refactor that I thought might make sense: we keep track of references with a `ReferenceValue` enum that's `Unknown` or `Known(ValueId)`. That's the same as having a value in a map or not having it, though maybe "unknown" and "known" convey a better semantic? But I think in the code before this PR we'd have maps with `Unknown` values in it which consume a bit more memory for no real reason. And removing `ReferenceValue` leads to less code which is usually good (unless we are golfing 😄).

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/ssa/opt/mem2reg/block.rs
```diff
@@ -93,16 +73,25 @@ impl Block {
             // Now we have to invalidate every reference we know of
             self.invalidate_all_references();
         } else if let Some(alias) = aliases.single_alias() {
-            self.references.insert(alias, value);
+            self.set_reference_value(alias, value);
```
jfecher, line 76: Nitpick: is this needed when we set the value below on line 86?

We can consider removing this case and changing the else to `else if aliases.len() > 1`

jfecher (approved):
1. Can you update the module-level mem2reg doc comment?
2. I'm surprised we don't see more circuit changes with this. But maybe the memory difference will be more significant since we've had issues with mem2reg's memory usage in the past.

compiler/noirc_evaluator/src/ssa/opt/mem2reg/block.rs
asterite: I wanted to do exactly what you mention here. I thought "if the address has a single alias then it must be itself". I added `assert_eq!(address, alias)` and... it fails! Which means that an address has a single alias that's not itself... but I didn't investigate why that happens. I guess I'll do that now as I think that shouldn't happen, right?

asterite: It seems that `MakeArray` is creating those aliases.

I'm debugging this code:

```
brillig(inline) fn foo f0 {
  b0(v0: &mut u1):
    v5 = call f1() -> &mut u1
    store u1 1 at v5
    v6 = make_array [v5] : [&mut u1; 1]
    return v6
}
```

Right before the `make_array` we have in aliases:
- Other(0): [0]
- Other(5): unknown

After `make_array` we end up with:
- Other(0): [0]
- Other(5): unknown
- ArrayElement(6): [5]

So v6 ends up having `v5` as an alias, but it doesn't have itself as an alias... is that right? 🤔

asterite: ~I guess that also means that here we remember that both `v5` and `v6` have a given value... but `v6` isn't a reference! It doesn't cause an issue because we can't load from `v6`, but it does feel like something's off here...~ In the end at least that's not what happens.

vezenovm: In `MakeArray`, if the array contains references we do insert every element as an alias to the array itself and this is intended, but definitely a little confusing.

asterite: Makes sense. I guess we can merge this PR and revisit this "address has one alias but it's not itself" thing in a follow-up issue or PR.

vezenovm:
Need to update some snaps

compiler/noirc_evaluator/src/ssa/opt/mem2reg/block.rs
jfecher: Yeah, going back to the "let's rework Expression" discussion this is definitely related to that. Maybe the more correct approach is to use `Expression` to represent values or collections that contain references, but aren't necessarily references themselves. Then when we `v1 = array_get v0, ...` where `expression_of(v0) = Expression::Array(all_references_that_may_be_in_the_array)` we create `v1` as a possible alias to all those references. This seems fairly in-line with the original intent of `Expression` I think but is maybe a more rigorous usage of it in practice than we currently have.

Issue #9482, opened 2025-08-13
mem2reg: expressions -> aliases indirection might not be needed?


asterite:
In mem2reg, in each block we track what Expression a value maps to, and then what aliases that Expression has:

https://github.com/noir-lang/noir/blob/b51b6164e5e49243bb65adb64638a821131568dd/compiler/noirc_evaluator/src/ssa/opt/mem2reg/block.rs#L19-L28

However, it's not clearly explained why this indirection is needed. Would mapping each value to an AliasSet be enough? My guess is "no" because if we detect that a value's related AliasSet needs to be modified, we'd need to modify it too for every value in that AliasSet. That's when the Expression indirection comes into play: we map all the aliased values to a single expression, then when we modify that expression's AliasSet it's automatically modified for all values that map to that expression.

Or maybe it's done to avoid having multiple identical AliasSets for a list of values, that is, for compilation performance reasons? It might be good to document this design decision. (though I checked the time and memory to compile `token_contract` and `ram_blowup_regression` and there was no noticeable change)

What's interesting is that if we remove this indirection all tests still pass (and no bytecode changes are produced). That means that either the indirection is not needed (at least not for correctness) or that a test is missing proving that this indirection is in fact needed. If the indirection is not needed we could consider removing it as it slightly simplifies the mem2reg logic.

Here's the patch to remove the indirection in case you want to try running all tests:

<details>
  <summary>Patch to remove indirection</summary>

```diff
diff --git a/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs b/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
index 6dd27aa282..b5af9ecd0a 100644
--- a/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
@@ -97,7 +97,7 @@ use crate::ssa::{
};

use self::alias_set::AliasSet;
-use self::block::{Block, Expression};
+use self::block::Block;

impl Ssa {
    /// Attempts to remove any load instructions that recover values that are already available in
@@ -192,12 +192,10 @@ impl<'f> PerFunctionContext<'f> {
                // then that reference gets to keep its last store.
                let typ = self.inserter.function.dfg.type_of_value(value);
                if Self::contains_references(&typ) {
-                    if let Some(expression) = references.expressions.get(&value) {
-                        if let Some(aliases) = references.aliases.get(expression) {
[...]

PR #9615, opened 2025-08-22
chore(mem2reg): add a few regression tests

### LINKED ISSUE #9398: mem2reg: store to nested parameter reference is incorrectly removed

### LINKED ISSUE #9415: mem2reg: store to reference inside array is lost
### Aim

This code:

```noir
// Prover.toml:
// index = 0
unconstrained fn main(index: u32) -> pub Field {
    let v0 = &mut 1;
    let v1 = &mut 2;
    let mut v3 = [v0];
    v3[index] = v1;
    *v3[index] = 3;
    *v3[index]
}
```

when executed, says the output if `0x02`. However, it should be `0x03`. At least in a similar Rust program the result is 3.

This doesn't happen if the `index` is not dynamic.

Also, this can only happen in unconstrained code because in constrained code the assignment `v3[index] = v1` is prohibited (it gives an error during monomorphization).

Similarly, this code incorrectly returns `0x02` instead of `0x03`:

```noir
// Prover.toml:
// index = 0
unconstrained fn main(index: u32) -> pub Field {
    let v0 = &mut 1;
    let v1 = &mut 2;
    let mut v3 = [v0];
    v3[index] = v1;
    *v3[index] = 3;
    *v1
}
```

### Expected Behavior

The code should return `0x03`.

### Bug

It incorrectly returns `0x02`.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #9439: mem2reg: mutable references in dynamic slices lose stores
This code:

```noir
use std::test::OracleMock;

unconstrained fn main() {}

#[oracle(get_slice)]
unconstrained fn get_slice() -> [Field] {}

#[test]
unconstrained fn test_get_slice() {
    let _ = OracleMock::mock("get_slice").returns((0, [0; 0])).times(1);
    let num1 = &mut 0;
    let slice1 = get_slice();
    let slice2 = slice1.map(|_| num1);
    let slice3 = slice2.push_back(num1);
    *slice3[0] = 9;
    assert_eq(*slice3[0], 9);
    assert_eq(*num1, 9);
}
```

fails when running `nargo test`.

It's very strange because the store to `*slice3[0]` seems to have no effect at all.

Note that this is a contrived example. 

The mem2reg doc comments say this:

```rust
//! - On `Instruction::Call { arguments }`:
//!   - If any argument of the call is a reference, set the value of each alias of that
//!     reference to `Unknown`
//!   - Any builtin functions that may return aliases if their input also contains a
//!     reference should be tracked. Examples: `slice_push_back`, `slice_insert`, `slice_remove`, etc.
//!   - Remove the instance of the last load instruction for any reference arguments and their aliases
```

However, there doesn't seem to be code to handle builtin functions or references returned from them. That made me wonder if there could be bugs around that. It was hard to find because these built-ins are immediately simplified to `make_array` if they are of known length, and the compiler forbids creating dynamic slices with mutable references (or at least I couldn't find a way to do so). But one way this can be done is via mocks, as shown above.

Edit: ah, it's actually simpler and doesn't involve mocks. I didn't know you could loop a dynamic number of times in unconstrained code:

```noir
// Prover.toml:
// n = 1
unconstrained fn main(n: u32) {
    let num1 = &mut 0;
    let mut slice = &[];
    for _ in 0..n {
        slice = slice.push_back(num1);
    }
    *slice[0] = 9;
    assert_eq(*slice[0], 9);
    assert_eq(*num1, 9);
}
```

That makes it a bit more concerning if mutable references are used in brillig code, in slices.

### LINKED ISSUE #9455: mem2reg: store lost after returning array with a mutable reference from `if`
### Aim

This code fails to execute when it should:

```noir
unconstrained fn main(predicate: bool) {
    let num1 = &mut 0;
    let array = if predicate { [num1] } else { [num1] };
    *array[0] = 9;
    assert_eq(*array[0], 9);
}
```

`predicate` can be true or false, it doesn't matter.

### Expected Behavior

The snippet should execute without errors.

### Bug

What's described above.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9398
Resolves #9415
Resolves #9455
Resolves #9439

## Summary

It seems all of the issues were fixed by #9632 so this PR just adds regression tests.

## Documentation

Check one:

# PR Checklist


vezenovm:
> (it's also interesting that this doesn't introduce any regressions in test programs...)

This isn't totally surprising as otherwise we would have been triggering this bug.

vezenovm:
> * that method's name is maybe misleading

Agreed it is misleading

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -1678,8 +1681,39 @@ mod tests {
...
+            jmpif v0 then: b1, else: b2
+          b1():
+            v5 = make_array [v2] : [&mut Field; 1]
+            jmp b3(v5)
+          b2():
+            v4 = make_array [v2] : [&mut Field; 1]
+            jmp b3(v4)
+          b3(v1: [&mut Field; 1]):
+            v7 = array_get v1, index u32 0 -> &mut Field
+            store Field 9 at v7
```
vezenovm, line 1743: What is the result of the SSA without the `remember_dereference`? It feels like we may not be appropriately marking v9 as an alias of v7. We then load from v9 and use it so the store should remain.

asterite: This single store line disappears without `remember_dereference`.

vezenovm: Ok yeah so that makes me think that it would be better to make v7 see v9 as an alias. Using `remember_dereference` will cause us to keep some stores that we do not have to keep.

asterite: How can we know we need to make v7 and v9 aliases?

vezenovm: Well when we have `v9 = array_get v1, index u32 0 -> &mut Field` we mark v9 as an alias of v1. It looks like we need to make this bidirectional where all pre-existing aliases of v1 now also are aliased by v9.

jfecher:
> I wonder if this was meant to track a reference being linked to a parent reference...

That is the case. Looking at the comment on the function:
>     /// Remember that `result` is the result of dereferencing `address`. This is important to
>     /// track aliasing when references are stored within other references.

It's to track that a reference result comes from dereferencing another reference. You can say its for references in general but since mem2reg only cares about values that contain references we don't care about results which don't contain references. So we don't care about loading from a `&mut Field` but we do want to remember loading from a `&mut &mut Field`.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
asterite: I see... though it doesn't seem like we are making `v9` an alias of `v1`. There's this code:

```rust
let expression = Expression::ArrayElement(array);
if let Some(aliases) = references.aliases.get_mut(&expression) {
    aliases.insert(result);
}
```

We only do so if there are already some aliases (this might be another bug).

vezenovm: > I see... though it doesn't seem like we are making `v9` an alias of `v1`

This would be another bug. I was going to say I would expect `v1` to be aliased by its elements already, however, `v1` comes from a block parameter so we may not be appropriately accounting for that. I would at least expect a known empty alias set so that v9 would be inserted as an alias of v1 here.

asterite:
@jfecher @vezenovm I added a small comment but I'm not sure it helps much.

I wonder if we could merge this PR as it is right now, mainly because it fixes 4 bugs and contains regression tests for them, making it easier to refactor to this code (like [this refactor](https://github.com/noir-lang/noir/issues/9482#issuecomment-3215450705)).

vezenovm:
> @jfecher @vezenovm I added a small comment but I'm not sure it helps much.
> 
> I wonder if we could merge this PR as it is right now, mainly because it fixes 4 bugs and contains regression tests for them, making it easier to refactor to this code (like [this refactor](https://github.com/noir-lang/noir/issues/9482#issuecomment-3215450705)).

I still would prefer we appropriately marked the result of two different ArrayGet's from the same array containing references as aliases as I do think this is the correct solution. I can take a closer look shortly to see if this fix can be done easily.

asterite:
In the end all of this was already working in master, https://github.com/noir-lang/noir/pull/9632 probably fixed all the issues. So this PR now just adds regression tests.

test_programs/execution_success/regression_9398/src/main.nr
```diff
@@ -0,0 +1,13 @@
+fn main() {
```
vezenovm: Can this be moved to `compile_success_empty`?

asterite: Done!

PR #9616, opened 2025-08-22
chore(mem2reg): removing seemingly unnecessary code

### LINKED ISSUE #9394: mem2reg: untested code
## `keep_last_store` logic not tested

In the definition of `Block::keep_last_store`:

https://github.com/noir-lang/noir/blob/b61f5eb6f371ee8e51c2ac42de38ed821b084b1b/compiler/noirc_evaluator/src/ssa/opt/mem2reg/block.rs#L194-L207

there's special handling for when the instruction is a `Store`. However, if that chunk of code is commented (lines 198-205) then all tests pass meaning that either that logic is not needed or that there's just no test for it yet.

We should find a test that hits that code path, otherwise remove it.

## `ArraySet` logic if the element type has references

If this code is commented out:

```noir
                    let array = *array;

                    let expression = Expression::ArrayElement(Box::new(Expression::Other(array)));

                    let mut aliases = if let Some(aliases) = references.aliases.get_mut(&expression)
                    {
                        aliases.clone()
                    } else if let Some((elements, _)) =
                        self.inserter.function.dfg.get_array_constant(array)
                    {
                        let aliases = references.collect_all_aliases(elements);
                        self.set_aliases(references, array, aliases.clone());
                        aliases
                    } else {
                        AliasSet::unknown()
                    };

                    aliases.unify(&references.get_aliases_for_value(*value));

                    references.expressions.insert(result, expression.clone());
                    references.aliases.insert(expression, aliases);
```

Then all tests (unit tests and the ones in test_programs) still pass, meaning that the logic there is either redundant or there's a missing test for it.

## `Terminator::Return` handling

This is how we handle the return terminator:

```rust
TerminatorInstruction::Return { return_values, .. } => {
    // We need to appropriately mark each alias of a reference as being used as a return terminator argument.
    // This prevents us potentially removing a last store from a preceding block or is altered within another function.
    for return_value in return_values {
        references.for_each_alias_of(*return_value, |_, alias| {
            self.instruction_input_references.insert(alias);
        });
    }
    // Removing all `last_stores` for each returned reference is more important here
    // than setting them all to ReferenceValue::Unknown since no other block should
    // have a block with a Return terminator as a predecessor anyway.
    self.mark_all_unknown(return_values, references);
}
```

However, if the entire `for` loop is commented, all tests still pass. It might be that `mark_all_unknown` is essentially doing what's needed, but I'm not sure.

asterite:
# Description

## Problem

Resolves #9394

## Summary

I don't fully intend this to get merged but it's interesting that there seems to be a lot of code that, when removed, has all tests still pass. Maybe there's some code that fails but isn't tested here? If CI passes I'll try to see if syncing this with Aztec-Packages causes no failures. If failures happen maybe we can craft some tests to prove why the code was needed.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
@@ -180,10 +180,7 @@ impl<'f> PerFunctionContext<'f> {
         }
 
         let mut all_terminator_values = HashSet::default();
-        let mut per_func_block_params: HashSet<ValueId> = HashSet::default();
         for (block_id, references) in self.blocks.iter_mut() {
-            let block_params = self.inserter.function.dfg.block_parameters(*block_id);
-            per_func_block_params.extend(block_params.iter());
             let terminator = self.inserter.function.dfg[*block_id].unwrap_terminator();
             terminator.for_each_value(|value| {
                 all_terminator_values.insert(value);
@@ -203,12 +200,8 @@ impl<'f> PerFunctionContext<'f> {
         // This rule does not apply to reference parameters, which we must also check for before removing these stores.
         for (_, block) in self.blocks.iter() {
             for (store_address, store_instruction) in block.last_stores.iter() {
-                let store_alias_used = self.is_store_alias_used(
-                    store_address,
-                    block,
-                    &all_terminator_values,
-                    &per_func_block_params,
-                );
+                let store_alias_used =
+                    self.is_store_alias_used(store_address, block, &all_terminator_values);
 
                 let is_dereference = block
                     .expressions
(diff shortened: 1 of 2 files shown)


compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -495,50 +483,19 @@ impl<'f> PerFunctionContext<'f> {
...
-                    {
-                        aliases.clone()
-                    } else if let Some((elements, _)) =
-                        self.inserter.function.dfg.get_array_constant(array)
-                    {
-                        let aliases = references.collect_all_aliases(elements);
-                        self.set_aliases(references, array, aliases.clone());
-                        aliases
-                    } else {
-                        AliasSet::unknown()
```
jfecher, line 534: Removing this whole section for aliases of arrayset instructions scares me haha.
I suppose it's indicative we don't have tests for it at the very least - I'd hesitate to say it's completely unnecessary but it does have me doubting.

asterite: I agree. At least some part of this PR must not be right because Aztec-Packages failed: http://ci.aztec-labs.com/bf55284bb1e49ea4
Now I'm checking if it also fails in the currently open PR for the noir sync (it's failing for another reason, but maybe the failure is before this PR).

asterite:
This breaks Aztec-Packages so I'm closing this. I'll try to see if I can find a small failing test for this.

asterite:
Huh, it seems that the latest Noir sync bumps into the same failure this PR bumped into: https://github.com/AztecProtocol/aztec-packages/pull/16548

That likely means that the changes here aren't the failure cause. I'll try it again once the noir sync is fixed because I'm curious to see if this breaks anything.

(in any case if we go with the `Container` refactor we might be able to see if some code here indeed had no consequences)

Issue #9451, opened 2025-08-08
purity_analysis: IncrementRc and DecrementRc are not pure


vezenovm:
### Aim

IncrementRc and DecrementRc mutate array pointers. The array values in this instructions are not reference types but they can still mutate information that is not local to the function itself.

```rust
    #[test]
    fn regression_array_pointer_mutation() {
        // This test ensures that a function which mutates an array pointer is marked impure.
        // This protects against future deduplication passes incorrectly assuming purity.
        let src = r#"
        brillig(inline) fn main f0 {
          b0(v0: u32):
            v3 = make_array [Field 1, Field 2] : [Field; 2]
            v5 = call array_refcount(v3) -> u32
            constrain v5 == u32 1
            v8 = call f1(v3) -> [Field; 2]
            v9 = call array_refcount(v3) -> u32
            constrain v9 == u32 2
            v11 = call f1(v3) -> [Field; 2]
            v12 = call array_refcount(v3) -> u32
            constrain v12 == u32 3
            inc_rc v3
            v15 = array_set v3, index v0, value Field 9
            return v3, v15
        }
        brillig(inline) fn mutator f1 {
          b0(v0: [Field; 2]):
            inc_rc v0
            v3 = array_set v0, index u32 0, value Field 5
            return v3
        }
        "#;

        let ssa = Ssa::from_str(src).unwrap();
        // Can uncomment to see that we do not fail when running constant folding before purity analysis
        // let ssa = ssa.fold_constants_using_constraints();        
        let _ = ssa.interpret(vec![Value::from_constant(1_u32.into(), NumericType::unsigned(32)).unwrap()]).unwrap();
        
        let ssa = ssa.purity_analysis();
        let _ = ssa.interpret(vec![Value::from_constant(1_u32.into(), NumericType::unsigned(32)).unwrap()]).unwrap();

        let ssa = ssa.fold_constants_using_constraints();        
        let _ = ssa.interpret(vec![Value::from_constant(1_u32.into(), NumericType::unsigned(32)).unwrap()]).unwrap();
    }
```
The following test will fail at `constrain v12 == u32 3` while interpreting the SSA after constant folding. This is because constant folding will deduplicate the `f1` call.

### Expected Behavior

I expect the interpreter to be not fail after constant folding is run with purity information. 

### Bug

Fails with the following:
```
[...]

PR #9617, opened 2025-08-22
fix: Make inc/dec_rc impure

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/9451

## Summary\*

Marks inc_rc and dec_rc as always impure.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
@@ -2174,20 +2174,15 @@ mod test {
         "#;
 
         let ssa = Ssa::from_str(src).unwrap();
-        // Can uncomment to see that we do not fail when running constant folding before purity analysis
-        // let ssa = ssa.fold_constants_using_constraints();
-        let _ = ssa
-            .interpret(vec![Value::from_constant(1_u32.into(), NumericType::unsigned(32)).unwrap()])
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -2142,4 +2143,51 @@ mod test {
...
+          b0(v0: [Field; 2]):
+            inc_rc v0
+            v3 = array_set v0, index u32 0, value Field 5
+            return v3
+        }
+        "#;
+
+        let ssa = Ssa::from_str(src).unwrap();
+        // Can uncomment to see that we do not fail when running constant folding before purity analysis
+        // let ssa = ssa.fold_constants_using_constraints();
```
vezenovm:
```suggestion
        let ssa = ssa.fold_constants_using_constraints();
```
nit: Should we just interpret here? Not sure why I didn't just have it this way in the original test. Might be pointless to interpret after purity analysis but before folding constants later as well. This is minor though as the semantics are still tested.

vezenovm (approved):

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
jfecher: I think I'll just remove it since it isn't needed

PR #9622, opened 2025-08-25
chore: some inlining refactors

### LINKED ISSUE #9397: Inlining: Panics when inlining a function with multiple return blocks
### Aim

We have a validation check that a function only has a single return block, however, we do have handling for multiple return blocks in the inlining pass. If we were ever wanted our SSA to support multiple returns blocks (LLVM and cranelift both support an arbitrary number of `return` instructions) this code becomes relevant again. 

### Expected Behavior

```rust
    #[test]
    fn multiple_return_blocks() {
        let src = "
        brillig(inline) fn main f0 {
          b0(v0: u1):
            v1 = call f1(v0) -> Field
            return v1
        }
        brillig(inline) fn foo f1 {
          b0(v0: u1):
            jmpif v0 then: b1, else: b2
          b1():
            return Field 1
          b2():
            return Field 2
        }
        ";
        let ssa = Ssa::from_str_no_validation(src).unwrap();
        let _ = ssa.inline_functions(i64::MAX).unwrap();
    }
```
The following program, should at minimum not panic and complete the inlining pass. Ideally f1 would be inlined into f0. 

### Bug

The snippet above gives the following panic:
```
thread 'ssa::opt::inlining::test::multiple_return_blocks' panicked at compiler/noirc_evaluator/src/ssa/opt/inlining.rs:641:9:
assertion `left == right` failed
  left: 1
 right: 0
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/panicking.rs:692:5
   1: core::panicking::panic_fmt
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:75:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:364:5
   4: noirc_evaluator::ssa::opt::inlining::PerFunctionContext::insert_new_instruction_results
             at ./src/ssa/opt/inlining.rs:641:9
   5: noirc_evaluator::ssa::opt::inlining::PerFunctionContext::inline_function
             at ./src/ssa/opt/inlining.rs:604:9
   6: noirc_evaluator::ssa::opt::inlining::PerFunctionContext::inline_block_instructions
             at ./src/ssa/opt/inlining.rs:500:33
   7: noirc_evaluator::ssa::opt::inlining::PerFunctionContext::inline_blocks
             at ./src/ssa/opt/inlining.rs:441:13
```
The panic comes from this assertion https://github.com/noir-lang/noir/blob/b61f5eb6f371ee8e51c2ac42de38ed821b084b1b/compiler/noirc_evaluator/src/ssa/opt/inlining.rs#L641

It looks like the panic is a result of returning the wrong values from `handle_function_returns`: https://github.com/noir-lang/noir/blob/b61f5eb6f371ee8e51c2ac42de38ed821b084b1b/compiler/noirc_evaluator/src/ssa/opt/inlining.rs#L478

The block parameters of the `return_block` are going to be zero (we just inserted the block). It looks like we are attempting to account for that our SSA requires a single return point by building a join block to have a unified return. 

A couple options:
- Explore a separate pass or utility function that modifies the CFG to have a single return point that is separate from inlining. This could be used by inlining, simplfiy cfg, etc. We can then also remove a restriction from our IR. 
- Modify inlining to only expect a single return block. Building a unified return block can come later if that restriction is relaxed

### To Reproduce

1. Copy the test above into the inlining test module
2. Run the test

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #9460: inlining: possible refactors
## Asserting that the inlined function ID is the same as the old ID

When looking at the code I noticed a new function with the inlined content is created, then that function replaces the original one. At first I thought that a purity might be lost here but it turns out that the new function is created with the same ID as the old one. It would be nice to add an assertion that this is true, right before replacing the old function with the new one.

## Duplicate code for copying block parameters

A code like this happens twice:

```rust
let original_parameters = self.source_function.dfg.block_parameters(source_block);

for parameter in original_parameters {
    let typ = self.source_function.dfg.type_of_value(*parameter);
    let new_parameter = self.context.builder.add_block_parameter(new_block, typ);
    self.values.insert(*parameter, new_parameter);
}
```

A helper function could be extracted, leading to less lines of code and a bit more clarity.

## `insert_new_instruction_results` could be simpler

The definition of `insert_new_instruction_results` is this:

```rust
fn insert_new_instruction_results(
    values: &mut HashMap<ValueId, ValueId>,
    old_results: &[ValueId],
    new_results: InsertInstructionResult,
) {
    assert_eq!(old_results.len(), new_results.len());

    match new_results {
        InsertInstructionResult::SimplifiedTo(new_result) => {
            values.insert(old_results[0], new_result);
        }
        InsertInstructionResult::SimplifiedToMultiple(new_results) => {
            for (old_result, new_result) in old_results.iter().zip(new_results) {
                values.insert(*old_result, new_result);
            }
        }
        InsertInstructionResult::Results(_, new_results) => {
            for (old_result, new_result) in old_results.iter().zip(new_results) {
                values.insert(*old_result, *new_result);
            }
        }
        InsertInstructionResult::InstructionRemoved => (),
    }
}
```

Essentially, it maps each old result to a new result by putting it into `values`.

The type `InsertInstructionResult` has a method `results()` that returns `Cow<'dfg, [ValueId]>` so that could be zipped with the old results. However, that method allocates a one-element vec so if that allocation is to be avoided a simple loop using an index could also work:

```rust
fn insert_new_instruction_results(
    values: &mut HashMap<ValueId, ValueId>,
    old_results: &[ValueId],
    new_results: InsertInstructionResult,
) {
    assert_eq!(old_results.len(), new_results.len());
    for i in 0..old_results.len() {
        values.insert(old_results[i], new_results[i]);
    }
}
```

## `handle_terminator_instruction` duplicates the logic to copying the call stack

This logic:

```rust
let call_stack = self.source_function.dfg.get_call_stack(call_stack);

let new_call_stack = self
    .context
    .builder
    .current_function
    .dfg
    .call_stack_data
    .extend_call_stack(self.context.call_stack, &call_stack);

self.context.builder.set_call_stack(new_call_stack);
```

is repeated three times. The `Terminator` type has a `call_stack` method, so we could call that in the beginning of the function, copy the call stack (only if it's not a return and we are inlining the entry function) to repeat less and make the code a bit more clear.

Alternative, if that's not possible, all that chunk of code could be moved to a helper function.

## Checking if a function has static_assert or assert_constant could be slightly faster

Right now it iterates all blocks and instructions trying to find a call to any of those intrinsics. However, the search could be skipped if neither intrinsic is registered in the function (this is really a very minor optimization).

## Dead code in `handle_function_returns`

The function handles the case of functions having multiple return blocks. However, that isn't possible (we have an SSA validation to prevent this) so that logic could be replaced with `unreachable!("...")`.

The logic could be kept if we'll eventually allow more than one return: in that case we should have a test for this scenario as there currently isn't one. The test will likely fail validation, once we support multiple returns it will fail for the wrong expected reason.

asterite:
# Description

## Problem

Resolves #9460
Resolves #9397

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/inlining/inline_info.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/inlining/inline_info.rs
@@ -163,16 +163,21 @@ fn compute_function_should_be_inlined(
     let function = &ssa.functions[&func_id];
     let assert_constant_id = function.dfg.get_intrinsic(Intrinsic::AssertConstant).copied();
     let static_assert_id = function.dfg.get_intrinsic(Intrinsic::StaticAssert).copied();
-    let contains_static_assertion = function.reachable_blocks().iter().any(|block| {
-        function.dfg[*block].instructions().iter().any(|instruction| {
-            match &function.dfg[*instruction] {
-                Instruction::Call { func, .. } => {
-                    Some(*func) == assert_constant_id || Some(*func) == static_assert_id
+
+    let contains_static_assertion = if assert_constant_id.is_none() && static_assert_id.is_none() {
+        false
+    } else {
+        function.reachable_blocks().iter().any(|block| {
+            function.dfg[*block].instructions().iter().any(|instruction| {
+                match &function.dfg[*instruction] {
+                    Instruction::Call { func, .. } => {
+                        Some(*func) == assert_constant_id || Some(*func) == static_assert_id
+                    }
+                    _ => false,
                 }
-                _ => false,
-            }
+            })
         })
-    });
+    };
 
     let neighbors = call_graph.graph().neighbors(index);
     let mut total_weight = compute_function_own_weight(function) as i64;


jfecher (approved):
Neat

compiler/noirc_evaluator/src/ssa/opt/inlining/inline_info.rs
```diff
@@ -163,16 +163,21 @@ fn compute_function_should_be_inlined(
...
-    let contains_static_assertion = function.reachable_blocks().iter().any(|block| {
-        function.dfg[*block].instructions().iter().any(|instruction| {
-            match &function.dfg[*instruction] {
-                Instruction::Call { func, .. } => {
-                    Some(*func) == assert_constant_id || Some(*func) == static_assert_id
+
+    let contains_static_assertion = if assert_constant_id.is_none() && static_assert_id.is_none() {
+        false
+    } else {
+        function.reachable_blocks().iter().any(|block| {
```
vezenovm, line 170: Can we not get rid of this logic entirely? Because if there are IDs for either intrinsic then we do contain a static assertion?

vezenovm (approved):

compiler/noirc_evaluator/src/ssa/opt/inlining/inline_info.rs
asterite: I thought the same. But I was wondering if the intrinsic was used in the function, then a simplification removed that call, I think the intrinsic would still be registered but unused, right?

vezenovm: Yes, good call. We never remove from the `intrinsics` map.

PR #9626, opened 2025-08-25
fix(inlining): Do not inline globals and lower them during ACIR gen

### LINKED ISSUE #9408: Inlining: Potential colliding value IDs when inlining globals
### Aim

In ACIR runtimes we do not want to keep global IDs and instead we want to inline that global constant. You can see the globals check [here](https://github.com/noir-lang/noir/blob/b2713a99fca969a888ecf14ebee7db62499adef0/compiler/noirc_evaluator/src/ssa/opt/inlining.rs#L348) during [translate_value](https://github.com/noir-lang/noir/blob/b2713a99fca969a888ecf14ebee7db62499adef0/compiler/noirc_evaluator/src/ssa/opt/inlining.rs#L313). 

### Expected Behavior

We should have a numeric constant that is local to the function when we inline ACIR functions containing globals. 

### Bug

The following test will fail:
```rust

    #[test]
    fn acir_global_constants_are_inlined_with_new_value_ids() {
        let src = "
        g0 = Field 1

        acir(inline) fn main f0 {
          b0():
            v0 = call f1() -> Field
            return v0
        }
        acir(inline) fn get_constant f1 {
          b0():
            return g0
        }
        ";

        let ssa = ssa.inline_functions(i64::MAX).unwrap();

        // The string output of globals resolves to their inner values, so we need to check whether they are globals explicitly.
        let main = ssa.main();
        let entry_block = main.entry_block();
        let terminator = main.dfg[entry_block].unwrap_terminator();
        let TerminatorInstruction::Return { return_values, .. } = terminator else {
            panic!("Expected return");
        };
        assert_eq!(return_values.len(), 1);
        assert!(!main.dfg.is_global(return_values[0]));

        assert_ssa_snapshot!(ssa, @r"
        g0 = Field 1

        acir(inline) fn main f0 {
          b0():
            return Field 1
        }
        ");
    }
```
This is most likely due to re-using constants, including globals, when making constants in the DFG: https://github.com/noir-lang/noir/blob/b2713a99fca969a888ecf14ebee7db62499adef0/compiler/noirc_evaluator/src/ssa/ir/dfg.rs#L433

For some additional context the following SSA program passes a similar test:
```rust
    #[test]
    fn acir_global_arrays_are_inlined_with_new_value_ids() {
        let src = "
        g0 = Field 1
        g1 = Field 2
        g2 = make_array [Field 1, Field 2] : [Field; 2]

        acir(inline) fn main f0 {
          b0():
            v0 = call f1() -> [Field; 2]
            // v1 = array_get g2, index u32 1 -> Field
            return v0
        }
        acir(inline) fn create_array f1 {
          b0():
            return g2
        }
        ";
        let ssa = ssa.inline_functions(i64::MAX).unwrap();
        // The string output of globals resolves to their inner values, so we need to check whether they are globals explicitly.
        let main = ssa.main();
        let entry_block = main.entry_block();
        let terminator = main.dfg[entry_block].unwrap_terminator();
        let TerminatorInstruction::Return { return_values, .. } = terminator else {
            panic!("Expected return");
        };
        assert_eq!(return_values.len(), 1);
        assert!(!main.dfg.is_global(return_values[0]));
        assert_ssa_snapshot!(ssa, @r"
        g0 = Field 1
        g1 = Field 2
        g2 = make_array [Field 1, Field 2] : [Field; 2]

        acir(inline) fn main f0 {
          b0():
            v6 = make_array [Field 1, Field 2] : [Field; 2]
            return v6
        }
        ");
    }
```
However, this version where we return a value from the formerly global array fails:
```rust
    #[test]
    fn acir_global_arrays_are_inlined_with_new_value_ids() {
        let src = "
        g0 = Field 1
        g1 = Field 2
        g2 = make_array [Field 1, Field 2] : [Field; 2]

        acir(inline) fn main f0 {
          b0():
            v0 = call f1() -> [Field; 2]
            v1 = array_get g2, index u32 1 -> Field
            return v1
        }
        acir(inline) fn create_array f1 {
          b0():
            return g2
        }
        ";

        let ssa = ssa.inline_functions(i64::MAX).unwrap();
        // The string output of globals resolves to their inner values, so we need to check whether they are globals explicitly.
        let main = ssa.main();
        let entry_block = main.entry_block();
        let terminator = main.dfg[entry_block].unwrap_terminator();
        let TerminatorInstruction::Return { return_values, .. } = terminator else {
            panic!("Expected return");
        };
        assert_eq!(return_values.len(), 1);
        assert!(!main.dfg.is_global(return_values[0]));

        assert_ssa_snapshot!(ssa, @r"
        g0 = Field 1
        g1 = Field 2
        g2 = make_array [Field 1, Field 2] : [Field; 2]
        acir(inline) fn main f0 {
          b0():
            v3 = make_array [Field 1, Field 2] : [Field; 2]
            v4 = make_array [Field 1, Field 2] : [Field; 2]
            return Field 2
        }
        ");
    }
```

### To Reproduce

1. Run the tests above

### Workaround

Yes

### Workaround Description

Right now neither program causes failures when continuing compilation. The SSA above doesn't cause a bad ACIR to be generated or any issues with execution but it is something to note as the behavior is different than expected. It may be better to inline all globals into ACIR in the frontend (thus banning them in SSA), or simply treating them as local values during SSA gen. The latter looks to already be happening and is why this hasn't triggered any bugs. So it seems a workaround already exists. It looks like we may be able to possibly avoid this inlining of globals for ACIR entirely. 

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #9409: Inlining: Global array gets recreated on every reference to it
### Aim

Take a look at this test (this test is the same as one of the examples from https://github.com/noir-lang/noir/issues/9408 just without the failing assertion):
```rust
    #[test]
    fn acir_global_arrays_are_inlined_with_new_value_ids() {
        let src = "
        g0 = Field 1
        g1 = Field 2
        g2 = make_array [Field 1, Field 2] : [Field; 2]

        acir(inline) fn main f0 {
          b0():
            v0 = call f1() -> [Field; 2]
            v1 = array_get g2, index u32 1 -> Field
            return v1
        }
        acir(inline) fn create_array f1 {
          b0():
            return g2
        }
        ";

        let ssa = ssa.inline_functions(i64::MAX).unwrap();
        // The string output of globals resolves to their inner values, so we need to check whether they are globals explicitly.
        let main = ssa.main();
        let entry_block = main.entry_block();
        let terminator = main.dfg[entry_block].unwrap_terminator();
        let TerminatorInstruction::Return { return_values, .. } = terminator else {
            panic!("Expected return");
        };
        assert_eq!(return_values.len(), 1);

        assert_ssa_snapshot!(ssa, @r"
        g0 = Field 1
        g1 = Field 2
        g2 = make_array [Field 1, Field 2] : [Field; 2]
        acir(inline) fn main f0 {
          b0():
            v3 = make_array [Field 1, Field 2] : [Field; 2]
            v4 = make_array [Field 1, Field 2] : [Field; 2]
            return Field 2
        }
        ");
    }
```

### Expected Behavior

I would expect either a single unused make array instruction or no make array at all. 

This is the result from a Brillig runtime which I would expect in this case as well:
```
        g0 = Field 1
        g1 = Field 2
        g2 = make_array [Field 1, Field 2] : [Field; 2]

        brillig(inline) fn main f0 {
          b0():
            return Field 2
        }
```

### Bug

It looks like we when inlining a global array in ACIR that the value is returned directly and never inserted into the `PerFunctionContext` values map https://github.com/noir-lang/noir/blob/b2713a99fca969a888ecf14ebee7db62499adef0/compiler/noirc_evaluator/src/ssa/opt/inlining.rs#L330

This means that whenever we reference the array we are going to generate a new make array instruction. 

### To Reproduce

1. Run the test above 

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

### Nargo Version

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #9408 
Resolves #9409 

## Summary\*

Before globals existed in SSA all globals were inlined in the frontend. Globals were then added, but treated as a Brillig exclusive concept. The inlining of globals was then moved to inlining. As per #9408 and #9409 this can cause some confusion in the IR output. I decided to move to that we keep globals throughout our SSA IR for any runtime. It is then up to the code gen of that runtime to handle globals lowering. ACIR then gets the same compilation benefits that Brillig does with globals (e.g., a more succinct IR in the case of large global arrays, reduced duplication). This compilation benefit can be seen in the `rollup-block-root` [benchmarks](https://github.com/noir-lang/noir/pull/9626#pullrequestreview-3152390126).

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/acir/mod.rs
+++ b/compiler/noirc_evaluator/src/acir/mod.rs
@@ -6,9 +6,9 @@
 //! ACIR generation is performed by calling the [Ssa::into_acir] method, providing any necessary brillig bytecode.
 //! The compiled program will be returned as an [`Artifacts`] type.
 
-use fxhash::FxHashMap as HashMap;
+use fxhash::{FxHashMap as HashMap, FxHashSet as HashSet};
 use noirc_errors::call_stack::CallStack;
-use std::collections::{BTreeMap, HashSet};
+use std::collections::BTreeMap;
 use types::{AcirDynamicArray, AcirValue};
 
 use acvm::acir::{
@@ -85,6 +85,12 @@ struct SharedContext<F: AcirField> {
     /// Keeps track of Brillig std lib calls per function that need to still be resolved
     /// with the correct function pointer from the `brillig_stdlib_func_pointer` map.
     brillig_stdlib_calls_to_resolve: HashMap<FunctionId, Vec<(OpcodeLocation, BrilligFunctionId)>>,
+
+    /// `used_globals` needs to be built from a function call graph.
+    ///
+    /// Maps an ACIR function to the globals used in that function.
+    /// This includes all globals used in functions called internally.
+    used_globals: HashMap<FunctionId, HashSet<ValueId>>,
 }
 
 impl<F: AcirField> SharedContext<F> {
@@ -239,7 +245,7 @@ impl<'a> Context<'a> {
             ssa_values: HashMap::default(),
             current_side_effects_enabled_var,
             acir_context,
-            initialized_arrays: HashSet::new(),
+            initialized_arrays: HashSet::default(),
             memory_blocks: HashMap::default(),
             internal_memory_blocks: HashMap::default(),
             internal_mem_block_lengths: HashMap::default(),
@@ -315,8 +321,37 @@ impl<'a> Context<'a> {
             expr.to_witness().expect("return vars should be witnesses")
         });
 
-        self.data_bus = dfg.data_bus.to_owned();
         let mut warnings = Vec::new();
+
+        let used_globals =
+            self.shared_context.used_globals.remove(&main_func.id()).unwrap_or_default();
+
+        let globals_dfg = (*main_func.dfg.globals).clone();
+        let globals_dfg = DataFlowGraph::from(globals_dfg);
+        for (id, value) in globals_dfg.values_iter() {
+            if !used_globals.contains(&id) {
+                continue;
+            }
+            match value {
+                Value::NumericConstant { .. } => {
+                    self.convert_value(id, dfg);
+                }
+                Value::Instruction { instruction, .. } => {
+                    warnings.extend(self.convert_ssa_instruction(
+                        *instruction,
+                        &globals_dfg,
+                        ssa,
+                    )?);
+                }
+                _ => {
+                    panic!(
+                        "Expected either an instruction or a numeric constant for a global value"
+                    )
+                }
+            }
+        }
+
+        self.data_bus = dfg.data_bus.to_owned();
         for instruction_id in entry_block.instructions() {
             warnings.extend(self.convert_ssa_instruction(*instruction_id, dfg, ssa)?);
         }
--- a/compiler/noirc_evaluator/src/acir/ssa.rs
+++ b/compiler/noirc_evaluator/src/acir/ssa.rs
@@ -42,7 +42,7 @@ pub(super) fn codegen_acir(
 ) -> Result<Artifacts, RuntimeError> {
     let mut acirs = Vec::new();
 
-    let used_globals = ssa.used_globals_in_brillig_functions();
+    let used_globals = ssa.used_globals_in_functions();
 
     // TODO: can we parallelize this?
     let mut shared_context = SharedContext {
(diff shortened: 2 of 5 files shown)


vezenovm:
> `rollup-block-root`	`191` s	`223` s	`0.86`

Looks like this fix gives a nice compilation benefit in the case of large global arrays.

PR #9628, opened 2025-08-25
Revert "feat(mem2reg): address last known value is independent of its…

asterite:
… aliases (#9613)"

This reverts commit 92aa75d2fc665a83c8c1b7f9596d2ec09ffdb01c.

# Description

## Problem

Unblocks the Noir sync on Aztec-Packages

## Documentation

Check one:

# PR Checklist


jfecher:
@asterite if you aren't already working on it, can you make sure we have a regression for this so it is caught in the future?

asterite:
@jfecher I'm working on it. I see that the PR caused a difference in SSA but it's a bit SSA (luckily not that big, around 90K and only a few changes here and there) so I need to check if those differences are okay or if there's something wrong, then I could create a regression test.

jfecher:
@asterite sounds good - just wanted to make sure this isn't lost particularly since we're continuing to work on more mem2reg changes

PR #9633, opened 2025-08-25
feat(mem2reg): address last known value is independent of its aliases (take three)

### LINKED ISSUE #9468: mem2reg: Trivial store/load to reference returned from array not resolved

asterite:
# Description

## Problem

Fixes #9468

## Summary

Trying to apply #9613 once more, but fixing an issue.

### First issue

I think the refactor in #9613 was fine except: `get_known_value` was changed to get an address' value regardless of whether the address has aliases.

This `get_known_value` method is only used in one place: when analyzing Load instructions. There, if `get_known_value` returned None we did this:

```
// We don't know the exact value of the address, so we must keep the stores to it.
references.mark_value_used(address, self.inserter.function);
```

That was actually "because the address might be aliased we better keep the stores to it". That is, because `get_known_value` implied both that the address didn't have aliases while having a known value, it returning `None` potentially meant that it was aliased. Now that `get_known_value` is not related to aliases it's not okay to do that if the value is unknown.

Instead, we must preserve stores to the address if the address is potentially aliased, which is what this PR does (so it's further separating the two concepts).

### Second issue

`get_known_value` retrieving a known value even if an address is aliases, combined with issue #9638, triggered a bug. That was then fixed by https://github.com/noir-lang/noir/pull/9640

### Third issue

The fix above worked because `get_known_value` returned nothing if an address has an alias. So in this code:

```
acir(inline) fn create_note f0 {
  b0(v0: &mut Field, v1: &mut Field):
    jmp b1(v0, v1)
  b1(v2: &mut Field, v3: &mut Field):
    store Field 2 at v2
    store Field 3 at v3
    v4 = load v2 -> Field
    return v4
}
```

with the second issue fix we end up with v2 being aliased to v0, and v3 being aliased to v1.
Then `load v2` cannot be simplified because it doesn't have a known value (well, it has, but the code in master returns `None` if it has aliases).

Now that `get_known_value` does return a value even if it has aliases, the code above breaks.
The issue is that we set v0 as an alias of v2, but we don't also set the aliases of v0 to be aliases of v2. That is, v2 should be aliased to v0 and v1. That's what the last commit does. It doesn't seem to introduce regressions, and in fact this PR is an optimization.

## Additional Context

With these three fixes, CI on Aztec-Packages is green: https://github.com/AztecProtocol/aztec-packages/pull/16576 🎉 

## Documentation

Check one:

# PR Checklist


asterite:
It still fails on Aztec Packages... I'll maybe try again tomorrow.

asterite:
Reopening because I'll try to fix the issue in this PR

asterite:
This is now ready for review. CI is green and this PR in Aztec-Packages is green too (https://github.com/AztecProtocol/aztec-packages/pull/16576).

jfecher (approved):

PR #9642, opened 2025-08-26
chore: LICM refactors

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This branch is going to accumulate small refactorings of the LICM while I'm working on the audit.

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
Would still be good to see the compilation effect as we did change some of the logic of the pass here.

TomAFrench:
> Would still be good to see the compilation effect as we did change some of the logic of the pass here.

I don't think there's been any significant changes to the main logic of the pass, mostly encapsulating logic and data a bit more.

vezenovm:
> I don't think there's been any significant changes to the main logic of the pass, mostly encapsulating logic and data a bit more.

Yes agreed, I noticed the alert so I wanted to see if it affected benchmarks more generally. I was mainly thinking of the `current_induction_variables` which have a new type signature. Agreed it looks like there are no other logic changes and the benchmarks are more in line with what I expected now. Most likely we were just out of date with master.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -318,30 +399,50 @@ impl<'f> LoopInvariantContext<'f> {
...
+                    .map(|value| self.inserter.resolve(*value))
+                    .collect::<Vec<_>>();
+
+                loop_context
+                    .extend_values_defined_in_loop_and_invariants(&results, hoist_invariant);
             }
         }
 
-        self.set_induction_var_bounds(loop_, Some(false));
+        // We're now done with this loop so it's no safe to insert its bounds into `outer_induction_variables`.
```
```suggestion
        // We're now done with this loop so it's now safe to insert its bounds into `outer_induction_variables`.
```

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -230,68 +223,156 @@ struct LoopInvariantContext<'f> {
...
+    /// - The current block is not impure
+    fn can_hoist_control_dependent_instruction(&self) -> bool {
+        !self.current_block_control_dependent
+            && self.current_block_executes
+            && !self.current_block_impure
+    }
+
+    /// A control dependent instruction (e.g. constrain or division) has more strict conditions for simplifying.
+    /// This function matches [Self::can_hoist_control_dependent_instruction] except
+    /// that simplification does not require that current block is pure to be simplified.
```
```suggestion
    /// that simplification does not require that the current block is pure to be simplified.
```
nit: looks like old comment was written incorrectly

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -1067,6 +719,29 @@ impl<'f> LoopInvariantContext<'f> {
     }
 }
 
+/// Get and resolve the induction variable of a loop.
+fn get_induction_variable(inserter: &FunctionInserter, loop_: &Loop) -> Option<ValueId> {
```
vezenovm, line 723: Nit: these could also live on `Loop`. Non-blocking ofc

vezenovm (approved):

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
TomAFrench: Yeah, I think we can clear these up a bit in future. I think I'm going to shelf this for now though.

PR #9643, opened 2025-08-26
fix: don't thread-bomb unnecessarily

ludamad:
Previously, setting RAYON_NUM_THREADS was the workaround for running many nargos in parallel. However, to be able to run on a very large number of core machine like aztec's shared "mainframe" instance, there was easy low-hanging fruit.

This removes all the cases where threads allocated could have been >> threads used.


TomAFrench (approved):
LGTM

PR #9647, opened 2025-08-26
chore(mem2reg): Only add to per function last_loads if load is not removed

### LINKED ISSUE #9469: mem2reg: Last loads per function not accounting for last loads removed per block
### Aim

As per the doc comments on `last_loads`:
https://github.com/noir-lang/noir/blob/e54057dc34c7b675e329f23425ff25855e049358/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs#L139

We want to track a value's last load across the entire function. Thus, we only insert when there is not a known value for a load address:
https://github.com/noir-lang/noir/blob/e54057dc34c7b675e329f23425ff25855e049358/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs#L469

### Expected Behavior

We also track last loads to an address per block. I would expect that if a load did not survive it would not be in the `last_loads` set in the `PerFunctionContext`. 

### Bug

`references.last_loads` is always checked when analyzing an `Instruction::Load`: https://github.com/noir-lang/noir/blob/e54057dc34c7b675e329f23425ff25855e049358/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs#L474

This means if we can potentially have values in `self.last_loads` for loads that did not survive. Take the following steps:
- We load from a known value
- Nothing is inserted into `self.last_loads`
- We set `references.last_loads`
- The value becomes unknown 
- On a repeat load we insert into `self.last_loads`
- That last load is removed based upon `references.last_loads` 

This means if we have a repeat load from the same unknown value that load will still be inserted into `PerFunctionContext::last_loads`. We also always set last loads per block https://github.com/noir-lang/noir/blob/e54057dc34c7b675e329f23425ff25855e049358/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs#L492

The situation above should be rare, but it is possible when looking at the `Instruction::Load` analysis in isolation.

### To Reproduce

1. I have been unable to create the situation above as it looks like other logic in the pass has protected against it thus far.  However, it would still be good to have the logic match the intention.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #9469

## Summary\*

As per the issue description, this case should be unlikely but we should be clear about the intention of our logic. `PerFunctionContext::last_loads` should only containing loads that have survived. Previously we could potentially have a load that returns None from `get_known_value` thus inserting that load address into `PerFunctionContext::last_loads` even though it could potentially be later removed by looking at `references.last_loads`.

## Documentation\*

Check one:

# PR Checklist\*


asterite:
> this case should be unlikely

In the PR where I'm trying to reapply the PR that we reverted I had to make some changes around there. It might be that this PR will trivially fix the issues I was having, but we'll see.

https://github.com/noir-lang/noir/pull/9633/files#diff-9cbe261da2bf371c7c0127b7c8780f5dcdc6a5782f64951bf75953feb7115f6bL448-L449

PR #9653, opened 2025-08-27
chore(ssa): Refactor `unrolling`

aakoshh:
# Description

## Problem\*

Some refactoring opportunities I saw to make the code easier to audit.

## Summary\*

A few changes to `unrolling` that shouldn't alter its behaviour, but hopefully make the code cleaner:
* Remove the elimination of `inc_rc` and `dec_rc` from ACIR functions: these have nothing to do with unrolling (we only count instructions for Brillig to see if a loop is small), and when we insert instructions into the block we unroll into, the simplification inevitably calls `is_handled_by_runtime`, which returns `false` for these, thus they will not be inserted anyway.
* Switch to `std::collections::HashSet` from `im::HashSet` as we don't need deterministic data structures for unrolling, and hopefully _O(1)_ is faster than _O(log(n))_.
* Get rid of `unroll_loop_block_helper`: it was only called from `unroll_loop_block` and didn't seem to increase clarity. 
* Get rid of the recursion in `unroll_each`: previously when `Loops::unroll_each` encountered a nested loop, it stopped consuming `yet_to_unroll`, created a new `Loops`, made a recursive call to itself, then merged the results. We would probably not have loops nested deep enough to cause a stack overflow, however:
  * I found it a difficult to grok; now we have a simpler `loop` construct with a succession of `continue` conditions.
  * It never processed the rest of `yet_to_unroll`, so all the unrelated "sibling" loops we discovered in `find_all` were wasted. Now we finish going through them before refreshing the context. 
  * It required carefully passing the `failed_to_unroll` to the next `Loops` to avoid redoing them again.
* Move the `failed_to_unroll` and `modified_blocks` fields from `Loops` into the `unroll_each` method: `Loops` are used from multiple modules now (e.g. LICM), and these were only relevant to unrolling. As mentioned they also required passing from one context to the next. By keeping them local there is less reasoning to be done about their scope. 
* Removed `Loops::unroll_each` and moved all functionality into `Function::try_unroll_loops`. The latter only forwarded to the former, and now that it's no longer recursive, I think it's cleaner to keep all unrolling in there, rather than `Loops`, because during the iteration we do call `Loops::find_all` to refresh the context, which makes it look a bit odd to call it on `self`.
* Only iterating reachable blocks. 
* Added some extra assertions where empty values were returned.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench (approved):
Looks sensible to me.

compiler/noirc_evaluator/src/ssa/opt/unrolling.rs
```diff
@@ -206,23 +270,22 @@ impl Loops {
...
-            // These reachable checks wouldn't be needed if we only iterated over reachable blocks
-            if dom_tree.is_reachable(block) {
-                for predecessor in cfg.predecessors(block) {
-                    // In the above example, we're looking for when `block` is `loop_entry` and `predecessor` is `loop_body`.
-                    if dom_tree.is_reachable(predecessor) && dom_tree.dominates(block, predecessor)
-                    {
-                        // predecessor -> block is the back-edge of a loop
-                        loops.push(Loop::find_blocks_in_loop(block, predecessor, &cfg));
-                    }
+        for block in function.reachable_blocks() {
```
vezenovm: We could just use the reverse post order here. We already have the `post_order` in this function for computing the cfg

aakoshh: Would that be better in some way?

aakoshh: Or just to avoid going through all the blocks again? I thought about that, but thought in the end that this is the clearest way to express what we want, even though the CFG already used it before, so we could indeed find these blocks in something.

vezenovm: For one it allows us to avoid going through all the blocks again. This would be a pretty minor optimization though. It is more that using `reachable_blocks` has bitten us before with hidden bugs (e.g., https://github.com/noir-lang/noir/pull/9495). It is usually only an issue with traversal order so I think we are safe here, but in general I think it would be good to move away from `reachable_blocks` in anything but testing code.

vezenovm: I won't block on this though because looking at it again I think we are safe.

aakoshh: Thanks, I updated it 👍

PR #9655, opened 2025-08-27
fix(ssa): Constant fold Brillig calls using the SSA interpreter

### LINKED ISSUE #9015: Consider replacing usage of the brillig VM in `fold_constants_with_brillig` with the SSA interpreter

### LINKED ISSUE #9097: Bug: value is not typed as brillig usize when using `--show-ssa`

vezenovm:
# Description

## Problem\*

Alternative to #9365

Resolves #9015 
Resolves #9097 
Resolves https://github.com/noir-lang/noir/pull/9484#issuecomment-3225247265 (same error as #9097)
Resolves a panic from a [Noir sync](https://github.com/AztecProtocol/aztec-packages/pull/15899) when attempting to debug log an array value (the same error as #9097)

Let's take the snippet from #9097:
```noir
fn main() {
    // Safety:
    let result = unsafe { foo([0]) };
    println(result);
}

unconstrained fn foo(bytes: [u8; 1]) -> Field {
    for _ in 0..bytes.len() {
        let _ = bytes[0];
    }
    0
}
```

We can see the following SSA when we fold constants with Brillig: 
<p>

```
After Inlining Brillig Calls (1) (step 42):
  b0():
    v1 = make_array [u8 0] : [u8; 1]                    // src/main.nr:3:32
    call f2(u1 1, Field 0)                              // std/lib.nr:40:9
    return
}
  b0(v0: [u8; 1]):
    inc_rc v0                                           // src/main.nr:8:17
    return Field 0
}
brillig(inline) impure fn print_unconstrained f2 {
  b0(v0: u1, v1: Field):
    v13 = make_array b"{\"kind\":\"field\"}"            // std/lib.nr:40:9
    call print(v0, v1, v13, u1 0)                       // std/lib.nr:34:5
    return
}

Remove Unreachable Instructions (2) (step 43): 0 ms
After Remove Unreachable Instructions (2) (step 43):
  b0():
    v1 = make_array [u8 0] : [u8; 1]                    // src/main.nr:3:32
    call f1(u1 1, Field 0)                              // std/lib.nr:40:9
    return
}
brillig(inline) predicate_pure fn print_unconstrained f1 {
  b0(v0: u1, v1: Field):
    v13 = make_array b"{\"kind\":\"field\"}"            // std/lib.nr:40:9
    call print(v0, v1, v13, u1 0)                       // std/lib.nr:34:5
    return
}
```

We can see that one Brillig function was folding into `f0`. However, the print remains. However, the Brillig context structure contains all the artifacts already has a set [mapping for function ids to Brillig artifacts](https://github.com/noir-lang/noir/blob/7d48355fa729a7e3d862e7afb193494c6a9d5aed/compiler/noirc_evaluator/src/brillig/mod.rs#L51). Thus, when we remove functions after Brillig gen and change their IDs (e.g., like when normalizing) we end up fetching an artifact for the wrong function when it comes times to generate an entry point. 

</p>
</details> 

## Summary\*

Alternative to #9365. I realized we were already cloning all Brillig functions in `fold_constants_with_brillig`. In this version I simply changed the interface of the SSA interpreter to allow passing a mapping of functions instead of the entire SSA. This enables us to fold Brillig calls as part of `constant_folding` which avoids some of the regressions (both in bytecode size and compilation time) from #9365. Folding Brillig calls as part of `constant_folding` allows us to access new constants that get exposed during the execution of the pass.

This issue could have been fixed for #9097 by simply updating the IDs point to the Brillig artifacts as part of normalization. However, this felt brittle and I felt the problem laid out above still feels like it could come up elsewhere. 

Aside the benefit of avoiding extra Brillig compilation, switching to using the SSA interpreter for constant folding Brillig entry points also avoids these kinds of situations where Brillig context needs to be updated due to secondary SSA passes post Brillig gen. 

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -88,15 +80,8 @@ impl Ssa {
             };
         }
 
-        let brillig_info = Some(BrilligInfo { brillig, brillig_functions: &brillig_functions });
-
         for function in self.functions.values_mut() {
-            // We have already performed our final Brillig generation, so constant folding
-            // Brillig functions is unnecessary work.
-            if function.dfg.runtime().is_brillig() {
```
vezenovm, line 96: I decided to run this over both runtimes as the compilation time benchmarks don't look to show many regressions. This does diverge from the original purpose of `fold_constants_with_brillig` (folding Brillig entry points), but I figured less runtime specific logic simplifies the compiler.

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -228,15 +208,34 @@ struct ResultCache {
...
+        let interpreter = brillig_functions.and_then(|brillig_functions| {
+            if brillig_functions.is_empty() {
+                None
+            } else {
+                let mut interpreter = Interpreter::new_from_functions(
+                    brillig_functions,
+                    InterpreterOptions { no_foreign_calls: true, ..Default::default() },
+                    std::io::empty(),
+                );
+                // Interpret globals once so that we do not have to repeat this computation on every Brillig call.
```
vezenovm: I experimented with interpreting globals only once across all functions. However, the interpreter currently requires a mutable reference so we cannot simply have it live on a `Context` with just lifetimes, we would most likely need a RefCell. As the compilation benchmarks are fine with the interpreter as it is now, I think this is ok.

If we still want to avoid this repeat computation (which may be worth it for cases with large globals) we can also thread the interpreter as a mutable reference argument through our constant folding methods. I have done that in [this PR](https://github.com/noir-lang/noir/pull/9658).

compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
```diff
@@ -104,7 +106,20 @@ impl Ssa {
 impl<'ssa, W: Write> Interpreter<'ssa, W> {
     fn new(ssa: &'ssa Ssa, options: InterpreterOptions, output: W) -> Self {
         let call_stack = vec![CallContext::global_context()];
-        Self { ssa, call_stack, side_effects_enabled: true, options, output }
+        Self { functions: &ssa.functions, call_stack, side_effects_enabled: true, options, output }
```
asterite: This could call `new_from_functions`

vezenovm: done

PR #9659, opened 2025-08-27
feat: brillig functions can be pure if they are not entry points

### LINKED ISSUE #9444: purity_analysis: Overly conservative default purity for Brillig functions
### Problem

When checking if a function is pure the default purity for all Brillig functions is `PureWithPredicate`:
https://github.com/noir-lang/noir/blob/a24cfcc3822da1b9db40b63f88e967ead2fabd68/compiler/noirc_evaluator/src/ssa/opt/pure.rs#L129

As per the comment this is because:
```
            // Because we return bogus values when a brillig function is called from acir
            // in a disabled predicate, brillig functions can never be truly pure unfortunately.
```
However, not all Brillig functions are called from ACIR. Brillig can be callled from Brillig.

### Happy Case

We should appropriately mark Brillig entry points as being `PureWithPredicate` by default, but regular Brillig functions should be able to be marked `Pure` as default. Having more Brillig functions accurately marked as pure should have an impact on optimizations the compiler can perform.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9444

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/ssa/opt/pure.rs
```diff
@@ -271,6 +280,32 @@ fn analyze_call_graph(
 
     finished_purities
 }
+
+/// Returns the set of Brillig functions that are entrypoints:
+/// main (if it's Brillig) and any Brillig functions that are called from ACIR.
+fn compute_brillig_entry_points(ssa: &Ssa) -> HashSet<FunctionId> {
```
vezenovm: We actually have a similar method in the `brillig_entry_points` pass. It makes a mapping of entry points -> internal calls, but we could just call that method and fetch the keys of the map.

vezenovm: Or we add a parameter that we should skip looking for internal calls.

asterite: Done! I reused that method as-is. Given that this is called just twice it probably doesn't have a big impact on performance, but I added `bench-show` so we can see.

asterite:
I had to make a change because a recursive brillig function was now considered `pure` which meant `DIE` could remove it. However, recursive functions can loop forever so they cannot be pure, so I marked those as impure (only in brillig). Maybe recursive functions (either directly or mutually) must always be impure for that reason, not sure.

TomAFrench (approved):

PR #9663, opened 2025-08-28
chore(ssa): Refactor flattening

aakoshh:
# Description

## Problem\*

Trying to improve clarity for the audit.

## Summary\*

* Some of the methods returned "next" blocks which were unused. Removed these to avoid the confusion about what drives the algorithm. 
* Added a `WorkList` which maintains a separate stack and set of blocks, to avoid _O(n)_ inclusion checks for deeply nested branching.
* Format docs. 
* Replace `arguments_stack: Vec<_>` with `next_arguments: Option<_>` to get rid of a bug that did not manifest itself in practice, but didn't make any sense either (see [below](https://github.com/noir-lang/noir/pull/9663#issuecomment-3233007574)).
* Break up `handle_instruction_side_effects` to make it easier to see what it does with calls/intrinsics/blackbox functions. 
* Make the matches in `handle_*_side_effects` exhaustive so we don't miss handling any new instruction/intrinsic. 
* Add the absence `ConstrainNotEqual` a pre-condition. 
* Fix the passing on of `local_allocations` from the `then_branch` to the `else_branch` and then back to the previous context so that we don't pass out the `else_branch` allocations, but rather restore what was before. The effect is that some loading of previous reference values can be saved, where we are storing to references allocated on a branch before a nested branch made it forget its history. Added test to [demonstrate](https://github.com/noir-lang/noir/pull/9663/commits/3b978802c9ab58f47b09aaad85f7fd33a15c3590).

## Documentation\*

Check one:

# PR Checklist\*


aakoshh:
There seems to be a bug in the handling of `arguments_stack`

* In `inline_block` we pop it unconditionally.
* In `handle_terminator` we push 1 item: empty for `JmpIf` and the actual args for `Jmp`
* `handle_terminator` enqueues [`then_block`, `else_block`, `exit_block`]
* `handle_terminator` calls `then_stop` or `else_stop` , and the latter calls `inline_branch_end`  to prepare the args for the exit block
* `inline_branch_end` pops 2 items from `arguments_stack` and replaces them with the merged arguments, which seem to be assumed to already be in the inserters cache.

The following is a rundown of the `basic_jmpif` test:
```rust
b0(v0: u1):                     // flatten:    args_stack = []
  jmpif v0 then: b1, else: b2   // hndl_trm: push(args_stack = [], args = [])  -> [[]]
b1():                           // inln_blk: pop (args_stack = [[]])           -> ([], Some([]))
  jmp b3(Field 3)               // hndl_trm: push(args_stack = [], args = [3]) -> [[3]]
b2():                           // inln_blk: pop (args_stack = [[3]])          -> ([], Some([3]))
  jmp b3(Field 4)               // hndl_trm: push(args_stack = [], args = [4]) -> [[4]]
                                // brch_end: pop (args_stack = [[4]])          -> ([], Some([4]))
                                // brch_end: pop (args_stack = [])             -> ([], None))
                                // brch_end: push(args_stack = [v0*3+!v0*4])   -> [[v0*3+!v0*4]]
b3(v1: Field):                  // inln_blk: pop (args_stack = [[v0*3+!v0*4]]) -> ([], Some([[v0*3+!v0*4]]))
  return v1
```
So the _else_ block consumes the arguments pushed by the _then_ branch and treats it as its own, which is surely not correct. However, it doesn't have parameters, so this is not causing any issues, and in the end it actually looks up the arguments of the _then_ terminator, to do the merge, at which point those values are already in the inserter. 

The _else_ handler pops two items from the arguments stack, which doesn't seem correct either, but because it always pushes one argument, the _exit_ block will always see something.

I simplified this by replacing the `arguments_stack: Vec<_>` with a `next_arguments: Option<_>` and only pushing/consuming in blocks which have non-empty blocks, so there should only ever be 1 item in there.

compiler/noirc_evaluator/src/ssa/ir/dfg.rs
```diff
@@ -250,7 +250,7 @@ impl DataFlowGraph {
         block: BasicBlockId,
         ctrl_typevars: Option<Vec<Type>>,
         call_stack: CallStackId,
-    ) -> InsertInstructionResult {
+    ) -> InsertInstructionResult<'_> {
```
TomAFrench, line 253: Are you getting these new clippy warnings as well? I've been holding off on fixing these as we'll pick them up again pretty quickly if we don't enforce it globally.

aakoshh: Yes, but I only seem to get them in the editor, not from running clippy. I started fixing them in the open files as they are a bit distracting, then I saw there were 30+ other places as well.

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -11,6 +11,7 @@
 //!     This also means the only acir functions in the program should be `main` (if main is
 //!     constrained), or any constrained `InlineType::Fold` functions.
 //!   - Precondition: Each constrained function should have no loops (unrolling has been performed).
+//!   - Precondition: "Equal" constraints have not been turned into "NotEqual".
```
jfecher, line 14: I'm curious - what breaks in this pass if there are already not equal constraints?

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -300,22 +325,58 @@ fn flatten_function_cfg(function: &mut Function, no_predicates: &HashMap<Functio
...
 
+/// Blocks enqueued for processing.
+/// It contains a block at most once.
+#[derive(Debug, Default)]
+pub(crate) struct WorkList {
+    /// Stack of blocks to process, always popping from the back.
+    stack: Vec<BasicBlockId>,
+    /// Blocks currently on the stack, for quick inclusion checks.
+    added: HashSet<BasicBlockId>,
+}
```
jfecher: This reminds me of https://docs.rs/indexmap/2.11.0/indexmap/set/struct.IndexSet.html

Looks like we use it in acvm-repo but not elsewhere. (not saying we need to change this, just making a note)

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
TomAFrench: We don't have a good way of disabling these constraints if they're under a predicate like we can for equal constraints.

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
aakoshh: Thanks, I'm using that now.

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -234,59 +253,66 @@ pub(crate) struct Context<'f> {
     /// the most recent condition combined with all previous conditions via `And` instructions.
     condition_stack: Vec<ConditionalContext>,
 
-    /// Stack of block arguments
-    /// When processing a block, we pop this stack to get its arguments
-    /// and at the end we push the arguments for his successor
-    arguments_stack: Vec<Vec<ValueId>>,
+    /// Arguments prepared by the last inlined block for the for the next block we are about to process.
```
aakoshh:
```suggestion
    /// Arguments prepared by the last inlined block for the next block we are about to process.
```

compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs
```diff
@@ -1107,15 +1242,17 @@ impl<'f> Context<'f> {
...
+    /// in order to 'nullify' side-effects when the 'condition' is false (in 'handle_instruction_side_effects' function).
+    ///
     /// Since the condition is a boolean, it can be safely casted to any other type.
     fn cast_condition_to_value_type(
         &mut self,
         condition: ValueId,
         value: ValueId,
         call_stack: CallStackId,
     ) -> ValueId {
+        // TODO: Can we early return in the value is a boolean?
```
aakoshh:
```suggestion
```

aakoshh:
On thing I forgot to touch on in the end was that `inline_block` mentions "track slice capacities" but I didn't see any evidence of this.

PR #9664, opened 2025-08-28
fix(mem2reg): reuse existing expression and add missing alias

### LINKED ISSUE #9657: mem2reg reuses incorrect load from array element
### Aim

While trying to exercise some mem2reg code related to ArraySet I stumbled upon another bug.

The SSA for it is this:

```
brillig(inline) fn main f0 {
  b0(v0: &mut Field, v1: &mut Field, v2: u32):
    v3 = make_array [v0] : [&mut Field; 1]
    v4 = array_set v3, index v2, value v1
    v5 = load v0 -> Field
    v6 = array_get v4, index v2 -> &mut Field
    store Field 0 at v6
    v7 = load v0 -> Field
    return v7
}
```

When running mem2reg we get this SSA:

```
brillig(inline) fn main f0 {
  b0(v0: &mut Field, v1: &mut Field, v2: u32):
    v3 = make_array [v0] : [&mut Field; 1]
    v4 = array_set v3, index v2, value v1
    v5 = load v0 -> Field
    constrain v2 == u32 0, "Index out of bounds"
    v7 = array_get v4, index u32 0 -> &mut Field
    store Field 0 at v7
    return v5
}
```

Note how the last `load v0` was replaced with `v5`. This is incorrect as v0 and v1 are potentially aliases of each other. Then the previous load from v0 cannot be reused.

This Noir code also reproduces it:

```noir
// Prover.toml
// v2 = 0
unconstrained fn main(v2: u32) {
    let v0 = &mut 0;
    let v7 = foo(v0, v0, v2);
    assert_eq(v7, 100);
}

unconstrained fn foo(v0: &mut Field, v1: &mut Field, v2: u32) -> Field {
    let mut v3 = [v0];
    v3[v2] = v1;
    let _v5 = *v0;
    let v6 = v3[v2];
    *v6 = 100;
    *v0
}
```

The assertion fails because `v7` ends up being `0`. If you comment out the `let _v5 = *v0;` line then it works fine. Same reason as above: the value loaded at `_v5` is incorrectly reused in the later `*v0`.

### Expected Behavior

The code above should execute fine.

### Bug

It fails to execute, giving an assertion failure.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

I thought this might have been introduced by #9633 but I checked and it happened before that was merged.

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9657

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -513,18 +513,26 @@ impl<'f> PerFunctionContext<'f> {
...
+                if self.inserter.function.dfg.type_of_value(result).contains_reference() {
+                    let array = *array;
                     self.instruction_input_references
                         .extend(references.get_aliases_for_value(array).iter());
                     references.mark_value_used(array, self.inserter.function);
 
-                    let expression = Expression::ArrayElement(array);
+                    // An expression for the array might already exist, so try to fetch it first
+                    let expression = references.expressions.get(&array).copied();
+                    let expression = expression.unwrap_or(Expression::ArrayElement(array));
```
jfecher: I think this is not correct - ArrayElement would imply we're getting the element of an array e.g. if we were looking up `result` instead. We're getting the array itself here though so it should probably just be Other.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -535,7 +543,8 @@ impl<'f> PerFunctionContext<'f> {
                     let result = self.inserter.function.dfg.instruction_results(instruction)[0];
                     let array = *array;
 
-                    let expression = Expression::ArrayElement(array);
+                    let expression = references.expressions.get(&array).copied();
+                    let expression = expression.unwrap_or(Expression::ArrayElement(array));
```
jfecher: Same here. If you want we can keep these though since they're removed in the container refactor anyway.

jfecher (approved):

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
asterite: Good point. I wasn't sure ArrayElement was the right choice here so I was about to change it to Other. If we do that, ArrayElement is never used. Then I thought maybe Dereference wasn't used either (we could track dereferences separately) but I wasn't sure whether to do that entire refactor in this PR. But I'll do it (I'll also check that it doesn't break Aztec-Packages).

asterite: Ah, no, it's still used in MakeArray...

asterite: Done!

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
asterite: Done!

asterite:
I forgot to add the Noir program as a regression test. When I added it... it failed! There was another issue. I fixed it in one way... it might introduce a small regression but doing it differently might be more complex and I don't know if it's worth it if the Container refactor will come.

PR #9672, opened 2025-08-28
feat: Propagate purities using SCCs

### LINKED ISSUE #9446: purity_analysis: mutually recursive functions are always impure
In the case of mutually recursive functions we end up assuming they are impure:

https://github.com/noir-lang/noir/blob/a24cfcc3822da1b9db40b63f88e967ead2fabd68/compiler/noirc_evaluator/src/ssa/opt/pure.rs#L246-L251

However, we already have `starting_purities` as a base purity so we could change `Purity::Impure` to be `starting_purities[&neighbor]`. Doing that change, in [this test](https://github.com/noir-lang/noir/pull/9445/files#diff-d920aec91ca4e3b5396636827ecabd909be429f58e9ca8d271d9e554446254c6R484) all functions end up being pure, which is correct and likely better.

vezenovm:
# Description

## Problem\*

Resolves #9446

## Summary\*

After updating `get_brillig_entry_points` to use the centralized call graph structure I realized I was already computing the set of recursive functions using SCCs. So I figured we may as well switch to propagating purities using SCCs as that is the better standard for detecting mutually recursive cycles in modern compilers.  

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/pure.rs
```diff
@@ -228,63 +234,68 @@ fn analyze_call_graph(
...
+    let mut finished = HashMap::default();
+
+    // Map FunctionId -> SCC index for quick lookup
+    let mut func_to_scc = HashMap::default();
+    for (i, scc) in sccs.iter().enumerate() {
+        for &func in scc {
+            // Each function belongs to exactly one SCC by definition of SCCs.
+            // Therefore inserting into func_to_scc here is safe, and there will
+            // be no overwrites.
+            func_to_scc.insert(func, i);
```
asterite: We could assert that `insert` returns `None`.

compiler/noirc_evaluator/src/ssa/opt/pure.rs
```diff
@@ -228,63 +234,68 @@ fn analyze_call_graph(
...
+                    let neighbor = call_graph.indices_to_ids()[&neighbor_idx];
+                    let neighbor_scc = func_to_scc[&neighbor];
+                    if neighbor_scc != i {
+                        combined_purity = combined_purity.unify(scc_purities[neighbor_scc]);
                     }
-                });
-                purity = purity.unify(neighbor_purity);
+                }
+
+                // If this function is recursive AND a Brillig function, mark as PureWithPredicate
```
asterite: We could explain why we do this (is it because it could run forever?)

compiler/noirc_evaluator/src/ssa/opt/pure.rs
```diff
@@ -593,10 +604,140 @@ mod test {
...
+        let purities = &ssa.main().dfg.function_purities;
+        assert_eq!(purities[&FunctionId::test_new(0)], Purity::PureWithPredicate);
+        assert_eq!(purities[&FunctionId::test_new(1)], Purity::PureWithPredicate);
+        assert_eq!(purities[&FunctionId::test_new(2)], Purity::PureWithPredicate);
+    }
+
+    #[test]
+    fn mutual_recursion_marks_functions_impure() {
+        // f1 -> f2 -> f3 -> f1 (a cycle of three functions)
+        // Only f3 is locally impure (it does an allocate).
```
asterite: I think `allocate` is pure but it's impure because it returns a reference.

compiler/noirc_evaluator/src/ssa/opt/pure.rs
asterite: Never mind, I was looking it commit by commit but now I see it's done in a next commit.

asterite (approved):
Looks good!

compiler/noirc_evaluator/src/ssa/opt/pure.rs
vezenovm: This felt like overkill as it is essentially verifying the `petgraph` algorithm and the definition of SCCs. If we want to be safe I could maybe add a debug assert?

compiler/noirc_evaluator/src/ssa/opt/pure.rs
vezenovm: Ah yes this is incorrect I'll update

compiler/noirc_evaluator/src/ssa/opt/pure.rs
asterite: Not having an assertion is fine 👍

vezenovm: I added a `debug_assert` just in case the graph upstream is broken. This is unlikely but is zero cost in release builds.

PR #9674, opened 2025-08-28
chore: break `NodeInterner` into chunks

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

I ended up looking at the node interner file and it made me very sad. We should avoid having huge files like this as it's very hard to maintain.

I've taken the first few steps of splitting this functionality into submodules so we can start encapsulating logic more.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/node_interner/mod.rs
```diff
@@ -1949,13 +1033,6 @@ impl NodeInterner {
...
 
-    /// Returns what the next trait impl id is expected to be.
-    pub fn next_trait_impl_id(&mut self) -> TraitImplId {
-        let next_id = self.next_trait_implementation_id;
-        self.next_trait_implementation_id += 1;
-        TraitImplId(next_id)
-    }
-
     /// Removes all TraitImplKind::Assumed from the list of known impls for the given trait
     pub fn remove_assumed_trait_implementations_for_trait(&mut self, trait_id: TraitId) {
```
asterite: This one could also be moved to `trait_impl.rs`

PR #9681, opened 2025-08-29
chore: remove unnecessary truncation when removing shift-right

asterite:
# Description

## Problem

No issue, just something I noticed while writing docs for remove_shift_right.

## Summary

At the end of replacing shift-right we truncate, for example, an i8 to 8 bits. That should always be a no-op so in this PR that's removed.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
@@ -158,12 +158,11 @@ impl Context<'_, '_, '_> {
                 // - ones_complement(lhs) / (2^rhs) == 0
                 // As the upper bit is set for the ones complement of negative numbers we'd need 2^rhs
                 // to be larger than the lhs bitsize for this to overflow.
-                let shifted = self.insert_binary(
+                self.insert_binary(
                     shifted_complement,
                     BinaryOp::Sub { unchecked: true },
                     lhs_sign_as_int,
-                );
-                self.insert_truncate(shifted, bit_size, bit_size + 1)
+                )
             }
 
             NumericType::NativeField => unreachable!("Bit shifts are disallowed on `Field` type"),
@@ -721,8 +720,7 @@ mod tests {
                 v9 = div v7, i32 4
                 v10 = cast v2 as i32
                 v11 = unchecked_sub v9, v10
-                v12 = truncate v11 to 32 bits, max_bit_size: 33
-                return v12
+                return v11
             }
             ");
         }
@@ -799,8 +797,7 @@ mod tests {
                 v65 = div v64, v57
                 v66 = cast v59 as i32
                 v67 = unchecked_sub v65, v66
-                v68 = truncate v67 to 32 bits, max_bit_size: 33
-                return v68
+                return v67
             }
             "#);
         }


asterite:
I'm working on something else and now I'm not sure this is right, even though all tests pass. Just a sec...

asterite:
In the end I think this is fine.

vezenovm:
> In the end I think this is fine.

I was trying to produce a bad result with the interpreter as I also thought this would be unsafe but I was unable to do so. I can take a closer look later.

asterite:
It's very strange because a simple shift right also ends up with this SSA:

```
acir(inline) fn main f0 {
  b0(v0: i8):
    v2 = shr v0, i8 1
    v3 = truncate v2 to 8 bits, max_bit_size: 9
    ...
}
```

Shouldn't truncating an integer of bit size N to N always be a no-op?

I'll make that change in this PR too to check if something breaks...

(maybe this code is not needed after all the changes in bit-shifting)

vezenovm:
> Shouldn't truncating an integer of bit size N to N always be a no-op?

Yes that makes sense to me, especially in the SSA you posted there. However, I don't think is the case if we have an unchecked operation and we do transform shifts into different operations in ACIR. 

> (maybe this code is not needed after all the changes in bit-shifting)

Perhaps this is just the case 🤷

compiler/noirc_evaluator/src/ssa/ssa_gen/context.rs
```diff
@@ -332,15 +332,6 @@ impl<'a> FunctionContext<'a> {
 
         let mut result = self.builder.insert_binary(lhs, op, rhs);
 
-        // Check for integer overflow
-        if matches!(operator, |BinaryOpKind::ShiftLeft| BinaryOpKind::ShiftRight) {
-            let result_type =
-                self.builder.current_function.dfg.type_of_value(result).unwrap_numeric();
-            if let NumericType::Signed { bit_size } = result_type {
-                result = self.builder.insert_truncate(result, bit_size, bit_size + 1);
```
vezenovm, line 340: I do think we still want to truncate here. What would happen if we have an overflowing shift left in ACIR?

asterite: Wouldn't that be handled when transforming the shift left by `remove_bit_shifts`? That computes 2^rhs and if that overflows it'll give a constrain failure.

But it's true that this breaks things so this must be wrong but I don't understand why... it might also be why https://github.com/noir-lang/noir/pull/9682 doesn't work either (though maybe it would work if we keep the truncations? I'll try it)

asterite: Hm, #9682 doesn't work if it just simplifies the math operation but keeps the existing truncation... but I also don't understand why. The SSA looks fine to me. Do you happen to know why?

vezenovm: > That computes 2^rhs and if that overflows it'll give a constrain failure.

Where do we do this? I see that we only insert a constrain in `remove_bit_shifts` for checking that the rhs is less than the bit size?

vezenovm: > Hm, #9682 doesn't work if it just simplifies the math operation but keeps the existing truncation... but I also don't understand why. The SSA looks fine to me. Do you happen to know why?

I would need to investigate further. But I see in #9682 we no longer truncate the lhs input to `shifted_complement`.

asterite: > Where do we do this?

Ah, it's true. I guess the truncations are needed then, even though we are operating with non-fields, because in the end they are all fields? It's a bit unintuitive but maybe it's done for performance reasons. For now I'll abandon this PR and the other one (I was trying to simplify the doc comments I was writing if we were able to simplify the code).

Thanks!

PR #9682, opened 2025-08-29
feat: simpler remove_bit_shifts shift-right

asterite:
# Description

## Problem

No issue, just something else I noticed while trying to document how remove_bit_shifts work.

## Summary

When we shift-right a signed integer we first check if it's positive or negative. If it's negative we do a bunch of math to slightly adjust that number before doing the division. When trying to understand how that works I noticed that the result is always the original value plus one (in the negative case). That's what this PR does.

For example here's a Rust program showing that:

```rust
fn main() {
    for x in -128_i8..=127 {
        for y in 1..7 {
            let pow = 2_i8.pow(y);
            if x >= 0 {
                assert_eq!(x >> y, x / pow, "Failed for x: {}, y: {}", x, y);
            } else {
                assert_eq!(x >> y, ((x + 1) / pow) - 1, "Failed for x: {}, y: {}", x, y);
            }
        }
    }
}
```

(I didn't check negative `y` because the transformation done in this PR only affects `x`)

However... one test fails (`bit_shifts_runtime`) and I don't understand why.

In particular, this program:

```noir
// Prover.toml
// x = -1
fn main(x: i8) {
    assert_eq(x >> 1, -1);
}
```

when executed, gives:

```
error: Failed constraint
  ┌─ src/main.nr:4:15
  │
4 │     assert_eq(x >> 1, -1);
  │               ------
  │
  = Call stack:
    1. src/main.nr:4:15

```

The SSA is:

```
  b0(v0: i8):
    v2 = lt v0, i8 0
    v3 = cast v2 as i8
    v4 = unchecked_add v0, v3
    v6 = div v4, i8 2
    v7 = unchecked_sub v6, v3
    constrain v7 == i8 -1
    return
}
```

The interpreter runs fine through that SSA, here's the trace:

```
v0 = i8 -1
v2 = u1 1
v3 = i8 1
v4 = i8 0
v6 = i8 0
v7 = i8 -1
```

So I don't understand why it doesn't work well in ACIR... maybe it's an ACIR bug?

## Documentation

Check one:

# PR Checklist


asterite:
Probably even though these are all signed integers, in the end they are fields and overflows are not automatically caught... not sure!

PR #9683, opened 2025-08-29
chore(licm): Break things up further in LICM

aakoshh:
# Description

## Problem\*

`LoopContext` had a bunch of `current_block_*` fields which had to be reset for each block we iterated over in `hoist_loop_invariants`, and assigned new values in various methods that also mutated other fields on `LoopContext`. Some were unrelated, like `is_control_dependent_post_pre_header` also set the block purity, seemingly just because `all_predecessors` was available there. This made it hard to find where various flags on `LoopContext` came from.

## Summary\*

* Breaks out the `current_block_*` fields from `LoopContext` into a `BlockContext` and introduces an `LoopInvariantContext::init_block_context` method to create it, so we cannot forget to set the values, and there is no risk of reusing something from the previous iteration. 
* Changes all query methods for the `BlockContext` initialisation to avoid mutating the `LoopContext`, so it's a bit more obvious where things are set, all clear in the `init_block_context` method.
* Renames `get_values_defined_in_loop` to `init_loop_context` as it did more than what it said on the tin.
* Makes `LoopContext::pre_header` non-optional.
* Copied some extra explanation for control dependence from the paper to have an intuitive definition.
* Avoid checking repeatedly whether the bounds in the `LoopContext` guarantee execution by storing the result in a field.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
@@ -612,7 +612,7 @@ impl<'f> LoopInvariantContext<'f> {
 
         // When hoisting a control dependent instruction, if a side effectual instruction comes in the predecessor block
         // of that instruction we can no longer hoist the control dependent instruction.
-        // This is important for maintaining ordering semantic correctness of the code.
+        // This is important for maintaining the execution order and semantic correctness of the code.
         // If the predecessors are all pure, the block might turn impure as an when we encounter
         // a side-effectful instruction in it later.
         let is_impure = all_predecessors.iter().any(|block| {
--- a/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
@@ -421,7 +421,7 @@ impl<'f> LoopInvariantContext<'f> {
                     }
                 } else {
                     let dfg = &self.inserter.function.dfg;
-                    // If the block has already been labelled as impure, we do need to check the current
+                    // If the block has already been labelled as impure, we don't need to check the current
                     // instruction's side effects.
                     if !block_context.is_impure {
                         block_context.is_impure = dfg[instruction_id].has_side_effects(dfg);
--- a/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
@@ -332,16 +332,6 @@ impl PostDominanceFrontiers {
     }
 }
 
-/// Result for checking whether a block is control dependent.
-enum ControlDependenceResult<'a> {
-    No,
-    Yes,
-    /// The block was already marked as control dependent as part of a previously found loop.
-    InSomeNestedLoop,
-    /// We found a loop which is control dependent and this block is part of it.
-    InNestedLoop(&'a Loop),
-}
-
 impl<'f> LoopInvariantContext<'f> {
     fn new(function: &'f mut Function, loops: &[Loop]) -> Self {
         let cfg = ControlFlowGraph::with_function(function);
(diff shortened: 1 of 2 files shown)


compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -210,46 +224,38 @@ struct LoopContext {
...
-    }
+struct BlockContext {
+    /// Indicate that this block is the header of the current loop.
+    is_header: bool,
+    /// Stores whether the current block being processed is control dependent.
+    is_control_dependent: bool,
+    /// Tracks whether the current block has a side-effectual instruction.
+    /// This is maintained per instruction for hoisting control dependent instructions
+    is_impure: bool,
+    /// Stores whether the current loop has known fix upper and lower bounds that
```
```suggestion
    /// Stores whether the current loop has known fixed upper and lower bounds that
```
nit: leftover from old comment

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -527,16 +518,12 @@ impl<'f> LoopInvariantContext<'f> {
...
                         self.post_dom_frontiers.is_control_dependent(*predecessor, nested.header)
                     });
+
             if nested_loop_is_control_dep {
-                loop_context.current_block_control_dependent = true;
-                // Mark all blocks in the nested loop as control dependent to avoid redundant checks
-                // for each of these blocks when they are later visited during hoisting.
-                // This is valid because control dependence of the loop header implies dependence
-                // for the entire loop body.
-                loop_context.nested_loop_control_dependent_blocks.extend(nested.blocks.iter());
```
vezenovm, line 536: I feel it would be nicer to still accept a loop_context and extend the `nested_loop_control_dependent_blocks` set here. We can then just return a `ControlDependenceResult`, remove the `InNestedLoop(&'a Loop)` variant, and remove the lifetime. 

It seems cleaner that `find_control_dependent_nested_loop` then returns `ControlDependenceResult::InSomeNestedLoop` or `ControlDependenceResult::No`. The result of a control dependent nested loop then matches this check:
```
        if loop_context.nested_loop_control_dependent_blocks.contains(&block) {
            return ControlDependenceResult::InSomeNestedLoop;
        }
```

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -617,15 +593,70 @@ impl<'f> LoopInvariantContext<'f> {
...
+    ) -> BlockContext {
+        let dfg = &self.inserter.function.dfg;
+
+        // Find all blocks between the current block and the loop header, exclusive of the current block and loop header themselves.
+        let mut all_predecessors = Loop::find_blocks_in_loop(loop_.header, block, &self.cfg).blocks;
+        all_predecessors.retain(|&predecessor| predecessor != block && predecessor != loop_.header);
+
+        // When hoisting a control dependent instruction, if a side effectual instruction comes in the predecessor block
+        // of that instruction we can no longer hoist the control dependent instruction.
+        // This is important for maintaining ordering semantic correctness of the code.
```
```suggestion
        // This is important for maintaining the execution order and semantic correctness of the code.
```
nit: looks like leftover grammar error

vezenovm:
Overall looks good! Just have that one request for removing the `ControlDependenceResult` variant with the lifetime.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
aakoshh: Hm, the reason I added this ControlDependenceResult was pretty much to avoid having to have to mutate the LoopContext here, which was a difficult to see side effect, not a typical caching pattern 🤔 

If we did that then I would see little reason to differentiate between No and InSomeNestedLoop, a book would be fine. 

I wanted to avoid mutation at all, then compromised to encapsulate it in init, but maybe you’re right and if it happens in in the query method it’s even better, as it would make the read and write appear next to each other. 

I’ll have another look 👍

vezenovm: > Hm, the reason I added this ControlDependenceResult was pretty much to avoid having to have to mutate the LoopContext here, which was a difficult to see side effect, not a typical caching pattern 🤔

I see. I just disliked the extra threading of the lifetime for this purpose. Maybe we could just returns the blocks? 

> If we did that then I would see little reason to differentiate between No and InSomeNestedLoop, a book would be fine.

Yes fair point. 

> it would make the read and write appear next to each other.

This would be nice as well.

aakoshh: Tried to find a middle ground by passing `&mut loop_context.nested_loop_control_dependent_blocks` to `is_control_dependent_post_pre_header` and got rid of `ControlDependencyResult`. This way it's clear we are passing a simple cache, not using any logic.

PR #9686, opened 2025-08-29
fix: greenlight `remove_bit_shifts`

### LINKED ISSUE #9525: Greenlight `remove_bit_shifts` for audits
Prepare https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs for audits:

  - [ ] No big brain needed to read the code
  - [ ] Purpose of the code
  - [ ] Design decisions made

**Target Date:** 12 Sep 2025

asterite:
# Description

## Problem

Resolves #9525

## Summary

This PR also fixes a bug: when simplifying shr/shl we were not checking if the result overflowed. It worked by change because we always truncated the result to fit the target bit size, which explains why just removing that truncation led to a failure.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
@@ -55,7 +55,7 @@
 //!
 //! This case is similar to unsigned shift-right except that for negative numbers we
 //! slightly adjust the value to shift, then adjust it again after performing the division,
-//! so the results are the exepcted ones.
+//! so the results are the expected ones.
 //!
 //! ## Signed shift-left
 //!


compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -1,3 +1,65 @@
...
+//! for i in 1 .. bit_size + 1 {
+//!     let r_squared = r * r;
+//!     let b = exponent_bits[bit_size - i];
+//!     r = if b { 2 * r_squared } else { r_squared };
+//! }
+//! ```
+//!
+//! ## Unsigned shift-left
+//!
+//! Shifting an unsigned integer to the right by N is the same as multipling by 2^N.
```
asterite:
```suggestion
//! Shifting an unsigned integer to the right by N is the same as multiplying by 2^N.
```

TomAFrench (changes requested):
blocking on above

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -1,3 +1,65 @@
...
+//! ```
+//!
+//! Like in the previous case, if the shift amount is not a constant it's computed
+//! via square&multiply.
+//!
+//! ## Signed shift-right
+//!
+//! This case is similar to unsigned shift-right except that for negative numbers we
+//! slightly adjust the value to shift, then adjust it again after performing the division,
+//! so the results are the exepcted ones.
```
TomAFrench:
```suggestion
//! so the results are the expected ones.
```

TomAFrench (approved):

PR #9687, opened 2025-08-29
fix(docs): moves docs to correct path for sharing domain with landing

alejoamiras:
# Description

## Problem\*
Currently docusaurus solves the assets paths as /[assets]/ and with the new landing coming, we need them to be under /docs/[assets]. 

## Summary\*
- We are removing the current index, to move towards the new landing.
- We are making docusaurus look for assets in /docs/[assets] when we are in production
- Letting docusaurus know our base url will be /docs/ when we are in production
- Redirecting assets being searched under /docs/ to be find in / (check redirects)

## Additional Context
Since this will be served under `noir-lang.org/docs/` on our production environment, before annoying our devops team I've worked on a way to fully test it's working (now and in the future): 
1. Closest to production environment, Docker: create a reverse-proxy through `nginx` and expose via port `3000`
2. A "patchy" solution: move everything to a temp `/docs` and serve from there

Maybe in the future we decide the setup works and we remove these files, but it was for the sake of trying to test as much as possible before annoying devops / going into prod.

## Side-effects
As a side-effect this currently breaks our link-previews since: Netlify build is serving it as if it where production and docusaurus expects us to be under `/docs/` on production.

**Proposed solution:** Let's make netlify deploy as a "staging" environment so everything is "as production" but docusaurus serves everything on the index `/` instead of `/docs/`.  

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Looks like the dockerfile, etc. can be removed now that the new marketing site is in production.

docs/README.md
```diff
@@ -67,8 +67,18 @@ yarn workspace docs build
...
+
+## Production Testing
+
+The site will be deployed at `noir-lang.org/docs/`. Test production configuration locally:
+
+### Simple Test
+```sh
+yarn workspace docs production:serve
+```
+Access at: `http://localhost:3003/docs/`
```
critesjosh:
```suggestion
Access at: `http://localhost:3000/docs/`
```

when i run the command it serves on port 3000

critesjosh (approved):
minor nit but looks good

PR #9378, opened 2025-08-01
fix(ssa): Replace unreachable array access with a constrain in Brillig

vezenovm:
# Description

## Problem\*

Testing an alternative to #9376 

## Documentation\*

Check one:
- [X] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [X] I have tested the changes locally.
- [X] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #9689, opened 2025-08-29
chore: greenglight `remove_ureachable_instructions`

### LINKED ISSUE #9518: Greenlight `remove_unreachable_instructions` for audits
Prepare https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs for audits:

  - [ ] Purpose of the code
  - [ ] Design decisions made

**Target Date:** 12 Sep 2025

asterite:
# Description

## Problem

Resolves #9518

## Summary

This just add docs.

I documented how it works but I'm not sure why:
- for array operations we don't put a constrain failure instead of creating a dummy array and fetching from it
- for slice operations we create a zeroed slice and return it when we already have a guaranteed empty slice (the one that produces the failure) so we could continue with that slice instead of create a new, empty one

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
@@ -6,6 +6,119 @@
 //!
 //! This pass might also add constrain checks after existing instructions,
 //! for example binary operations that are guaranteed to overflow.
+//!
+//! ## Handling of `constrain`
+//!
+//! Given an SSA like this:
+//!
+//! ```ssa
+//! constrain u1 0 == u1 1
+//! v1 = load v0 -> Field
+//! return v1
+//! ```
+//!
+//! because the constrain is guaranteed to fail, every instruction after it is removed
+//! and the terminator is replaced with `unreachable`:
+//!
+//! ```ssa
+//! constrain u1 0 == u1 1
+//! unreachable
+//! ```
+//!
+//! Similarly, `constrain u1 0 != u1 0` will have the same treatment.
+//!
+//! ## Handling of binary operations
+//!
+//! If a binary operation is guaranteed to overflow or fail:
+//!
+//! ```ssa
+//! v4 = add u8 254, u8 127 // guaranteed to overflow
+//! v5 = add v4, v6
+//! return v5
+//! ```
+//!
+//! the operation is left in the SSA but a constrain failure is added after it,
+//! and every subsequent instruction is removed:
+//!
+//! //! ```ssa
+//! v4 = add u8 254, u8 127
+//! constrain u1 0 == u1 1, "attempt to add with overflow"
+//! unreachable
+//! ```
+//!
+//! An operation that is guaranteed to fail but doesn't overflow is division by zero.
+//!
+//! Because binary operations can depend on a side-effects variable (`enable_side_effects`),
+//! if there's an active non-constant side-effects in place the constrain will check
+//! the value of this side-effects variable, and every subsequent instruction will be replaced
+//! with a zeroed-value. So for example this SSA:
+//!
+//! ```ssa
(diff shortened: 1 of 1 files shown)


vezenovm:
> * for array operations we don't put a constrain failure instead of creating a dummy array and fetching from it

This is because we gate the check only for ACIR. ACIR array operations are side effectual and check for array OOBs, while in Brillig they do not so we would have to insert a constrain in the Brillig runtime https://github.com/noir-lang/noir/pull/9378. However, this was breaking due to #7289 still being open at the time. I want to hold off on merging #9378 though as it looks like we are getting failures in aztec-packages from #9412 still.

> * for slice operations we create a zeroed slice and return it when we already have a guaranteed empty slice (the one that produces the failure) so we could continue with that slice instead of create a new, empty one

This is a good point. We could change this logic and slightly simplify that case.

asterite:
> This is a good point. We could change this logic and slightly simplify that case.

In the end I'm not sure this is worth it. The slice operations return the slice but also the popped element so we'd have to zero that anyway.

vezenovm:
> The slice operations return the slice but also the popped element so we'd have to zero that anyway.

We could just add that as an inline comment as an extra justification.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -6,6 +6,119 @@
...
+//!
+//! Given an SSA like this:
+//!
+//! ```ssa
+//! constrain u1 0 == u1 1
+//! v1 = load v0 -> Field
+//! return v1
+//! ```
+//!
+//! because the constrain is guaranteed to fail, every instruction after it is removed
```
vezenovm: this is meant to be lower case?

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -6,6 +6,119 @@
...
+//! the operation is left in the SSA but a constrain failure is added after it,
+//! and every subsequent instruction is removed:
+//!
+//! ```ssa
+//! v4 = add u8 254, u8 127
+//! constrain u1 0 == u1 1, "attempt to add with overflow"
+//! unreachable
+//! ```
+//!
+//! An operation that is guaranteed to fail but doesn't overflow is division by zero.
```
vezenovm:
```suggestion
//! Division by zero is an operation that is guaranteed to fail, but it does not overflow.
```
I think this reads better.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -6,6 +6,119 @@
...
+//!
+//! Because binary operations can depend on a side-effects variable (`enable_side_effects`),
+//! if there's an active non-constant side-effects in place the constrain will check
+//! the value of this side-effects variable, and every subsequent instruction will be replaced
+//! with a zeroed-value. So for example this SSA:
```
vezenovm:
```suggestion
//! Because binary operations can depend on a side-effects variable (`enable_side_effects`),
//! the constrain will check the value of this variable whenever a non-constant side effect 
//! is active. Every subsequent instruction is then replaced with a zeroed-value. 
//! So for example this SSA:
```
nit: just a proposed re-wording

vezenovm (approved):

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
asterite:
```suggestion
//! Because the constrain is guaranteed to fail, every instruction after it is removed
```

PR #9690, opened 2025-08-29
chore: have `--with-ssa-locations` instead of `--no-ssa-locations`

asterite:
# Description

## Problem

No issue, just something I noticed and other might agree with.

## Summary

By default locations in SSA are printed but it's more often that we don't want them than we do.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_driver/src/lib.rs
+++ b/compiler/noirc_driver/src/lib.rs
@@ -80,10 +80,10 @@ pub struct CompileOptions {
     #[arg(long, hide = true)]
     pub show_ssa_pass: Vec<String>,
 
-    /// Do not emit source file locations when emitting debug information for the SSA IR to stdout.
-    /// By default, source file locations will be shown.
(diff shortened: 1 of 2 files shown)


jfecher (commented):
I don't really have any preferences on this - I'll wait for someone else to approve.

I do think that if you do want locations it can be helpful to have them already be there when you're going through the ssa instead of figuring out you need them and need to re-generate everything. Compared to if you don't want locations but do have them you can just ignore them.

vezenovm:
> I do think that if you do want locations it can be helpful to have them already be there when you're going through the ssa instead of figuring out you need them and need to re-generate everything. Compared to if you don't want locations but do have them you can just ignore them.

I personally find myself not needing locations the majority of the time.

vezenovm (approved):
I am going to approve as no SSA locations is my default the vast majority of the time I use the SSA display. This is not difficult to revert if some one feels strongly the other way.

PR #9691, opened 2025-08-29
chore: replace usage of shell scripts with justfile

TomAFrench:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
@jfecher thoughts on adding a justfile to the repo here? Thinking being that it simplifies and goes some way towards documenting some of the commands which external developers may need to run. i.e. if there's something we expect an external developer to do as part of their PR then it can have an associated command in here which will install any necessary dependencies, etc.

jfecher:
@TomAFrench I haven't heard of it before and had to look it up but it does sound helpful to both external and internal developers. @aakoshh in the past for example has documented a few commands and options he has found useful for testing. It'd be nice to have those in the justfile as well.

I have a bunch of aliases in my .zshrc currently but it'd be nice for all devs to have a baseline of commands.

jfecher (approved):

PR #9692, opened 2025-08-29
feat(ssa): Inlining cost model refinement

### LINKED ISSUE #9410: Inlining: More accurate function weight calculation
### Problem

We have differing runtimes that have different costs for a given SSA instruction. In the case of inlining we are mostly concerned with the Brillig runtime as we expect all ACIR functions to be inlined. 

However, when computing the weights of different functions we use the number of SSA instructions as our heuristic https://github.com/noir-lang/noir/blob/b2713a99fca969a888ecf14ebee7db62499adef0/compiler/noirc_evaluator/src/ssa/opt/inlining/inline_info.rs#L301

This is imperfect as some SSA operations could translate to many different Brillig instructions while other could be closer to a one-to-one translation. This could lead to marking function weights incorrectly and an incorrect cost calculation.

### Happy Case

We should have cost estimates for how many Brillig instructions are generated from a given SSA instruction. This cost estimate should be used in computing the weights of functions in Brillig. 

We actually do something similar in the `basic_conditional` optimization: https://github.com/noir-lang/noir/blob/b2713a99fca969a888ecf14ebee7db62499adef0/compiler/noirc_evaluator/src/ssa/opt/basic_conditional.rs#L167

We could have a `brillig_cost` function on `Instruction` that returns this estimate and is used across passes that may need this estimate. 

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Nice-to-have

### Blocker Context

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #9410

## Summary\*

See the issue for more information as to why the function weight calculation is being updated.

- Introduces target_cost.rs, a consolidated module with Brillig opcode cost estimates for BinaryOp, Instruction, TerminatorInstruction, and Function. We could still improve some of the cost assessments. I have left TODOs within the match cases that need refinement.
- Accounts for call overhead (5 + N_args + M_returns call-site opcodes + CheckMaxStackDepth entry cost) and return-cost discount (return terminator eliminated when inlined) in the inline/retain cost comparison.
- Adds MAX_SIMPLE_FUNCTION_WEIGHT (80 Brillig opcodes) as the cost-model counterpart to the raw MAX_INSTRUCTIONS (10 SSA instructions) threshold for "simple" functions. I went with 80 cause 8 instructions is the max Brillig opcodes for an SSA instruction.

## Additional Context

Future work:
- Loop unrolling counts raw SSA instruction to decide whether unrolling is profitable. This new cost model will provide more accurate size estimates
- Basic conditional flattening has its own hardcoded cost table. We should consolidate that table into this one.

## Documentation\*

Check one:

# PR Checklist\*


jfecher (approved):
I see this is still a draft but LGTM.
I haven't verified all the exact cost estimates used but given that they lead to notably better optimizations I'd say they are good.

PR #9702, opened 2025-09-01
fix: wrong error message in brillig bit shift overflow

### LINKED ISSUE #9696: Strange shifting errors and results: `Attempted to shift by 0 bits on a type of bit size 8`
### Aim

Reproduce a nightly error from https://github.com/noir-lang/noir/actions/runs/17338660641/job/49229352879

The following program behaves differently in ACIR, Brillig and comptime.

```rust 
fn main() -> pub u64 {
    (0_u64 << (584130945418788864_u64))
}
```
alternatively:
```rust 
fn main() -> pub u64 {
    (10_u64 << (584130945418788864_u64))
}
```

### Expected Behavior

They should agree. 

### Bug

ACIR is good:
```
error: Assertion failed: attempt to bit-shift with overflow
  ┌─ src/main.nr:2:6
  │
2 │     (0_u64 << (584130945418788864_u64))
  │      ---------------------------------
  │
  = Call stack:
    1. src/main.nr:2:6

Failed assertion
```

Brillig is strange:
```
❯ cargo run -q -p nargo_cli -- execute --silence-warnings --force-brillig
error: Assertion failed: 'Attempted to shift by 0 bits on a type of bit size 8'
  ┌─ src/main.nr:2:6
  │
2 │     (0_u64 << (584130945418788864_u64))
  │      ---------------------------------
  │
  = Call stack:
    1. src/main.nr:2:6

Failed assertion
```

Why does it say "type of bit size 8" and "shift by 0 bits"? Does the same if we shift `10_u64 <<`. 

And `comptime` doesn't fail (possible duplicate of https://github.com/noir-lang/noir/issues/9542):
```rust
fn main() -> pub u64 {
    comptime {
        (10_u64 << (584130945418788864_u64))
    }
}
```

```
[fuzz_testing] Circuit output: 10
```

### To Reproduce

```shell
NOIR_AST_FUZZER_SEED=0xfb5b8eed00100000 cargo test -q -p noir_ast_fuzzer_fuzz comptime_vs_brillig_direct
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.11 noirc version = 1.0.0-beta.11+3b6b3592bd3058c4325637f4b13556b2ed704124 (git version hash: 3b6b3592bd3058c4325637f4b13556b2ed704124, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #9696 

## Summary\*
Brillig and comptime interpreter check for bit shift overflows

## Documentation\*

Check one:

# PR Checklist\*


acvm-repo/brillig_vm/src/arithmetic.rs
```diff
@@ -17,6 +17,8 @@ pub(crate) enum BrilligArithmeticError {
     MismatchedRhsBitSize { rhs_bit_size: u32, op_bit_size: u32 },
     #[error("Attempted to shift by {shift_size} bits on a type of bit size {bit_size}")]
     BitshiftOverflow { bit_size: u32, shift_size: u32 },
+    #[error("Attempted to shift by more than {bit_size} bits on a type of bit size {bit_size}")]
+    BitshiftOverflowBitSize { bit_size: u32 },
```
TomAFrench: these look redundant.

guipublic: Do you have another suggestion? 
May be just: "Attempted to shift by more than {bit_size} bits",  to avoid the repetition?

acvm-repo/brillig_vm/src/arithmetic.rs
```diff
@@ -228,6 +226,14 @@ pub(crate) fn evaluate_binary_int_op<F: AcirField>(
     }
 }
 
+fn bit_shift_overflow(shift: Option<u32>, bit_size: u32) -> BrilligArithmeticError {
+    if let Some(shift_size) = shift {
+        BrilligArithmeticError::BitshiftOverflow { bit_size: 8, shift_size }
```
aakoshh: Why is this 8?

guipublic: it should not be, nice spot.

aakoshh:
Should I create a separate issue for why `comptime` returns the `lhs` instead of an overflow in the examples of the ticket?

acvm-repo/brillig_vm/src/arithmetic.rs
guipublic: Done

aakoshh: Ok, I thought man, I'll never understand the nitty gritty of bit shifting 😅 

I'd add a quick docstring to this function to make it obvious why it exists (ie. shifting by larger than u32::MAX)

guipublic:
> Should I create a separate issue for why `comptime` returns the `lhs` instead of an overflow in the examples of the ticket?

No need, I can look at the comptime case as well within the same issue.

acvm-repo/brillig_vm/src/arithmetic.rs
aakoshh: I see that some tests are failing because they expected `BitshiftOverflow` and got `BitshiftOverflowBitSize`. 

Since it might be a niche error, maybe you could just use `BitshiftOverflow { shift_size: u32::MAX, ...`. That would indicate to people that they are dealing with some huge number they probably didn't intend, and it could be even more obvious than saying "attempted to shift by more than 32 bits", because they might think, hm, was it 64, or 33?

TomAFrench: I'd be in favour of just standardising on a single error variant for this. I haven't looked in detail on why this was added however.

aakoshh: My understanding was that if the `shift_size` is over `u32::MAX` we get the new error which doesn't have the `shift_size` field, otherwise the old error which does.

guipublic: I finally fixed this by using a u128 in the original error, so it can display properly all the cases.

TomAFrench (approved):

PR #9703, opened 2025-09-01
chore: Refactor `compute_dominance_frontiers`

aakoshh:
# Description

## Problem\*

`compute_dominance_frontiers` contained a confusing `while` condition.

## Summary\*

Changes `compute_dominance_frontiers` to compute `immediate_dominator` once and use `loop`, instead of computing it repeatedly with the same arguments in a `while` condition. Added some extra comments and examples as well.

## Additional Context

On an earlier bug fix I found this `while` suspicious: I wasn't sure if we should have been calling it with the `runner` instead of the `block_id`, since the latter doesn't change during the iteration.

For the record, this is the pseudo code from the paper; it might be where the `while` was coming from.

Screenshot 2025-09-01 at 16 40 39

## Documentation\*
Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/dom.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/dom.rs
@@ -787,6 +787,34 @@ mod tests {
     }
 
     #[test]
+    fn dom_frontiers_not_include_self() {
+        // In this example b1 is its own successor, by definition dominates itself,
+        // but not strictly (because it equals itself), so it fits the definition of
+        // the blocks in its own Dominance Frontier. But its dominance does not end
+        // there, so we don't consider it part of the DF.
+        let src = "
+        brillig(inline) fn main f0 {
+          b0(v0: u1):
+            jmp b1()
+          b1():
+            jmpif v0 then: b1, else: b2
+          b2():
+            return
+        }
+        ";
+        let ssa = Ssa::from_str(src).unwrap();
+        let main = ssa.main();
+
+        let cfg = ControlFlowGraph::with_function(main);
+        let post_order = PostOrder::with_cfg(&cfg);
+
+        let mut dt = DominatorTree::with_cfg_and_post_order(&cfg, &post_order);
+        let dom_frontiers = dt.compute_dominance_frontiers(&cfg);
+
+        assert!(dom_frontiers.is_empty());
+    }
+
+    #[test]
     fn post_dom_frontiers() {
         let ssa = loop_with_cond();
         let main = ssa.main();
--- a/compiler/noirc_evaluator/src/ssa/ir/dom.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/dom.rs
@@ -279,6 +279,22 @@ impl DominatorTree {
 
     /// Computes the dominance frontier for all blocks in the dominator tree.
     ///
+    /// The Dominance Frontier of a basic block X is the set of all blocks that are immediate
+    /// successors to blocks dominated by X, but which aren’t themselves strictly dominated by X.
+    /// It is the set of blocks that are not dominated X, and which are “first reached” on paths from X.
+    ///
+    /// For example in the following CFG, the DF of `B` is `{E}`, because while `B` dominates `C`,
+    /// there is another path to `E` through `D`.
+    /// ```text
+    ///    A
+    ///   / \
+    ///  B   D
+    ///  |   |
+    ///  C   |
+    ///   \ /
+    ///    E
+    /// ```
+    ///
     /// This method uses the algorithm specified in Cooper, Keith D. et al. “A Simple, Fast Dominance Algorithm.” (1999).
     /// As referenced in the paper a dominance frontier is the set of all CFG nodes, y, such that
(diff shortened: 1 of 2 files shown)


compiler/noirc_evaluator/src/ssa/ir/dom.rs
```diff
@@ -279,6 +279,22 @@ impl DominatorTree {
 
     /// Computes the dominance frontier for all blocks in the dominator tree.
     ///
+    /// The Dominance Frontier of a basic block X is the set of all blocks that are immediate
+    /// successors to blocks dominated by X, but which aren’t themselves strictly dominated by X.
+    /// It is the set of blocks that are not dominated X, and which are “first reached” on paths from X.
+    ///
+    /// For example in the following CFG, the DF of `B` is `{E}`, because while `B` dominates `C`,
+    /// there is another path to `E` through `D`.
```
asterite: Maybe say it like "because while B dominates C and E, only E is not strictly dominated by B because there is another path to E through D"?

asterite (approved):
Nice!

compiler/noirc_evaluator/src/ssa/ir/dom.rs
aakoshh: I'm not sure whether the word "strictly" means something significant, but in my understanding `B` does not dominate `E`, exactly because you can reach `E` by going `A -> D -> E`, which does not include `B`. By contrast to get to `C` you _have_ to go through `B`. 

So the Dominance Frontier is a set of nodes that the block _just_ doesn't dominate.

aakoshh: Apparently "strictly dominates" mean it dominates it and isn't equal to it, so strictly no self dominance.

~It feels redundant then, because if block Y is an immediate successor of a block dominated by X, then even if the block dominated by X is X itself, it's successor Y cannot be non-strictly dominated by it~ 🤷 

That definition is from https://pages.cs.wisc.edu/~fischer/cs701.f14/lectures/L6.4up.pdf btw.

aakoshh: Ah, but if X points at itself, then it dominates itself and itself is its own successor, but it doesn't _strictly_ dominate :D

aakoshh: But then if X dominates itself and has itself as its successor, then it would fit the definition of "not strictly dominated by itself? 🙃

aakoshh: So yeah a node can be in its own dominance frontier 🎉 
It's interesting though that the second sentence says "not dominated by X", even though it can dominate itself non-strictly 🤔 
Maybe I should add a unit test to see if the algorithm gets stuck.
Although it shouldn't because it has that dominance check, and the definition of _immediate dominator_ does include strictness, so it should move on.

aakoshh: So our algorithm will _not_ consider a self-link to be part of the dominance frontier, because it has a `break` there, which may make it incorrect in the strictest sense.

aakoshh: OTOH we use the post-dominance frontiers to decide control dependency, which roughly means that X is control dependent on Y if in Y we can make a decision not to travel towards X, which means Y is in the post dominance frontier of X, ie. we can go through Y without going through X. If X is in its own DF, then we can decide in X whether to go into X _again_ or not, but we are already in X, so it's a moot point from a control perspective, we won't ever _not_ reach X after reaching X.

aakoshh: Added a test to show that the DF does not include self. 

Apparently this is sort of a contradiction, and the most common interpretation of DF excludes self-links:


> Why a node is not in its own dominance frontier Let's consider if a node \(d\) could be in its own dominance frontier, based on the two conditions above: 
Condition 1: For \(d\) to be in its own dominance frontier, it would need to dominate one of its own predecessors. A node always dominates itself, so it dominates any of its predecessors that it also happens to be.
Condition 2: The second condition is where it fails. A node \(d\) must not strictly dominate \(d\) itself. By definition, "strict dominance" excludes the node itself (\(d\ne n\)). A node can never strictly dominate itself, so this condition (\(d\) does not strictly dominate \(d\)) is always met. 
However, the definition of the dominance frontier is specifically constructed to capture the nodes where dominance ends. If a node could be in its own dominance frontier, that would imply its dominance over itself was ending—a contradiction. 
A notable exception: Self-looping edge The nuance where a node can meet both conditions for itself relates to loops. A node with a self-loop, dominates itself and dominates its own predecessor (itself). It also does not strictly dominate itself. Some definitions have been interpreted to include a node in its own dominance frontier in this case. 
However, the canonical definition from Cytron et al., used in most modern compilers, excludes the node itself to maintain the property that dominance frontiers capture the first nodes reached on paths where dominance is lost. 


I also rephrased the comment, hopefully it's clearer that E isn't dominated by B.

PR #9708, opened 2025-09-02
chore: Add `PostOrder::into_vec_reverse`

aakoshh:
# Description

## Problem\*

Maxim said we should move away from `reachable_blocks` towards using reverse-post-order, which is used in multiple places already, but always in an ad-hoc fashion.

## Summary\*

Adds a `PostOrder::into_vec_reverse` method to encapsulate reversing the post-order vector to get the forward order. Makes it easier to discover this option and reduces the need for copy pasting.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench (approved):
LGTM

PR #9709, opened 2025-09-02
fix(ssa): Simplify instructions during `remove_unreachable_instructions`

### LINKED ISSUE #9695: Compiler crash: `Truncates are only ever applied to the result of a binary op or a param`
### Aim

Reproduce nightly failure from https://github.com/noir-lang/noir/actions/runs/17338660641/job/49229352879

The following program crashes during compilation:

```rust
fn main(mut a: pub Field) {
    if a == -1 {
        for _ in 0..2 {
            a = 1 / (unsafe { func_3_proxy(ctx_limit) } as Field);
        }
        for _ in 0..1 {
            assert(((a as u128) >= 1_u128), "OCS");
        }
    }
}
unconstrained fn func_3(ctx_limit: &mut u32) -> bool {
    if ((*ctx_limit) == 0_u32) {
        false
    } else {
        if func_3(ctx_limit) {
            true
        } else {
            false
        }
    }
}
unconstrained fn func_3_proxy(mut ctx_limit: u32) -> bool {
    func_3((&mut ctx_limit))
}
```

### Expected Behavior

Shouldn't crash.

### Bug

```console
❯ cargo run -q -p nargo_cli -- compile --silence-warnings --force
Message:  internal error: entered unreachable code: ICE: Truncates are only ever applied to the result of a binary op or a param
Location: compiler/noirc_evaluator/src/acir/mod.rs:1226
```

Looking at the SSA reveals the problem:
```rust
After Inlining Brillig Calls (1) (step 40):
  b0(v0: Field):
    v2 = eq v0, Field -1                          	// src/main.nr:3:8
    enable_side_effects v2                        	// src/main.nr:3:8
    v5 = div Field 1, Field 0                     	// src/main.nr:5:17
    v6 = div Field 1, Field 0                     	// src/main.nr:5:17
    v7 = truncate v6 to 128 bits, max_bit_size: 254	// src/main.nr:8:22
    v8 = cast v7 as u128                          	// src/main.nr:8:22
    v10 = eq v8, u128 0                           	// src/main.nr:8:21
    v11 = unchecked_mul v10, v2                   	// src/main.nr:8:21
    constrain v11 == u1 0, "OCS"                  	// src/main.nr:8:21
    enable_side_effects u1 1                      	// src/main.nr:3:8
    return
}
brillig(inline) impure fn func_3_proxy f1 {
  b0(v0: u32):
    v1 = allocate -> &mut u32                     	// src/main.nr:8:21
    store v0 at v1                                	// src/main.nr:8:21
    v3 = call f2(v1) -> u1                        	// src/main.nr:25:5
    return v3
}
brillig(inline) impure fn func_3 f2 {
  b0(v0: &mut u32):
    v2 = load v0 -> u32                           	// src/main.nr:25:5
    v4 = eq v2, u32 0                             	// src/main.nr:13:9
    jmpif v4 then: b1, else: b2
  b1():
    jmp b3(u1 0)
  b2():
    v6 = sub v2, u32 1                            	// src/main.nr:16:23
    store v6 at v0                                	// src/main.nr:16:23
    v8 = call f2(v0) -> u1                        	// src/main.nr:17:12
    jmp b3(v8)
  b3(v1: u1):
    return v1
}

After Remove Unreachable Instructions (2) (step 41):
  b0(v0: Field):
    v2 = eq v0, Field -1                          	// src/main.nr:3:8
    enable_side_effects v2                        	// src/main.nr:3:8
    v5 = div Field 1, Field 0                     	// src/main.nr:5:17
    constrain u1 0 == v2, "attempt to divide by zero"	// src/main.nr:5:17
    v7 = truncate Field 0 to 128 bits, max_bit_size: 254	// src/main.nr:8:22
    v8 = cast v7 as u128                          	// src/main.nr:8:22
    v10 = eq v8, u128 0                           	// src/main.nr:8:21
    v11 = unchecked_mul v10, v2                   	// src/main.nr:8:21
    constrain v11 == u1 0, "OCS"                  	// src/main.nr:8:21
    enable_side_effects u1 1                      	// src/main.nr:3:8
    return
}
```

So the _Constant folding with Brillig_ figures out that we will be dividing by zero, and then the _Remove Unreachable Instructions_ inserts and conditional-fail constraint (only fail if v2 is true), and replaces the results of subsequent instructions with defaults. At that point the `truncate v6` becomes `truncate Field 0`, which is unexpected.

### To Reproduce

```shell
NOIR_AST_FUZZER_SHOW_AST=1 NOIR_AST_FUZZER_SEED=0x0a6d55ab00100000 cargo test -q -p noir_ast_fuzzer_fuzz acir_vs_brillig
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.11 noirc version = 1.0.0-beta.11+3b6b3592bd3058c4325637f4b13556b2ed704124 (git version hash: 3b6b3592bd3058c4325637f4b13556b2ed704124, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9695 

## Summary\*

Changed `SimpleOptimizationContext` to check the hash of the instruction before and after the optimisation pass, and if changes are detected then work like `FunctionInserter::push_instruction_value` and use `DataFlowGraph::insert_instruction_and_results_if_simplified`, instead of just pushing the instruction ID to the end of the block, which results in the instruction potentially be simplified out of the SSA. 

This is relied on by `remove_unreachable_instructions`: when we encounter a conditional failure, which triggers the `UnreachableUnderPredicate` mode that replaces side-effecting instructions with default values, those values can be constants that can now enable simplifications for subsequent instructions. 

## Additional Context

In the bug ticket we had this SSA:
```rust
    enable_side_effects v2
    v5 = div Field 1, Field 0                     	
    v6 = div Field 1, Field 0                     	
    v7 = truncate v6 to 128 bits, max_bit_size: 254	
```
The pass recognised that `v5` will always fail if `v2` is `true`, and inserted a conditional constraint. That caused `v6` to be replaced by a default 0 (which doesn't appear in the SSA), and subsequently caused the constant `u64 0` to appear in `v7`:
```rust
    v5 = div Field 1, Field 0                     	
    constrain u1 0 == v2, "attempt to divide by zero"	
    v7 = truncate Field 0 to 128 bits, max_bit_size: 254	
```
By the time we get to ACIR gen, we don't expect `truncate` to operate on constants, however the [simplify](https://github.com/noir-lang/noir/blob/63f1528a210015029f1647285aaf084222051a63/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify.rs#L152-L156) module does handle it. 

Instead of adding special casing during ACIR gen, I thought it would be nicer to make `remove_unreachable_instruction` work akin to `flatten_cfg`, which [simplifies during insertion](https://github.com/noir-lang/noir/blob/63f1528a210015029f1647285aaf084222051a63/compiler/noirc_evaluator/src/ssa/opt/flatten_cfg.rs#L2294).

With this change the SSA simplifies as follows:
```rust
After Remove Unreachable Instructions (2) (step 41):
  b0(v0: Field):
    v2 = eq v0, Field -1                          	
    enable_side_effects v2                        	
    v5 = div Field 1, Field 0                     	
    constrain u1 0 == v2, "attempt to divide by zero"
    constrain v2 == u1 0                          	
    enable_side_effects u1 1                      	
    return
}
```
## Documentation\*

Check one:

# PR Checklist\*


asterite:
When I was auditing mem2reg I realized `simple_optimization` didn't simplify instructions like FunctionInserter. I tried to change `simple_optimization` to always simplify instructions if possible, that is, unconditionally, but it led to some failures. I wonder if it would be possible to always do it, it might be simpler than having to decide to do it or not.

Another thing I tried is to only simplify if an instruction would change because some of its values changed. That is, `Instruction::replace_values` returned a bool that said whether any value changed. If not, we can insert it without simplifying because if it didn't change then there's no way it will be simplified (I think!).

aakoshh:
> I tried to change simple_optimization to always simplify instructions if possible, that is, unconditionally, but it led to some failures.

Interesting, do you remember what failed? I assume you tracked the new IDs. 

I didn't think it was necessary to unconditionally simplify instructions, it seems like a lot of work relative to what is happening now, and that's work that might not even have to be done. I thought this would be the lightest touch, but according to the reports it might already have increased the memory requirements. 

That reminds me that there is an option to keep the existing instruction if it wasn't simplified; pushed a commit to take advantage of that.

> Another thing I tried is to only simplify if an instruction would change because some of its values changed. That is, Instruction::replace_values returned a bool that said whether any value changed.

That seems reasonable. The only thing I'm not sure about is that `simple_optimization_result` calls `replace_values`, so yes, if there was a change we know we can try to simplify, but then the lambda can also modify the instruction in place, before it's inserted at the end, and those modifications might mean that something that didn't have to be simplified to begin with can become eligible by the end.

aakoshh:
I have an idea: we can hash the values in the instruction before and after the call to see if we need to simplify. I'll try that.

aakoshh:
@asterite I tried to make simplification automatic, but it's not entirely safe to do so; I ended up putting it behind a flag, but now it uses the hash to check if it needs to do it for the given instruction. The flag is set to `true` by `remove_unreachable_instructions`, but it's done unconditionally, so there is less thinking about where it has to be set, reset, etc.

The reason it couldn't be done automatically is that the `static_assert_empty_loop` integration test failed. Consider this Noir code:
```rust
    let x: [Field; 0] = [];
    let _ = x.as_slice().map(|_| {
        std::static_assert(false, "test");
        0
    });
```
So we know that the array is empty, and therefore `map` will not run, and the `static_assert` should not fail. But with simplification it did! This was the SSA before it:
```rust
After `as_slice` optimization (1) (step 11):
acir(inline) fn main f0 {
  b0():
 	  ...
    v6 = make_array [] : [Field]                  	
    v7 = allocate -> &mut u32                     	
    store u32 0 at v7                             	
    v9 = allocate -> &mut [Field]                 	
    store v6 at v9                                	
    ...
    jmp b1(u32 0)
  b1(v0: u32):
    v30 = lt v0, u32 0                            	
    jmpif v30 then: b2, else: b3
  b2():
    ...                    	
    constrain u1 0 == u1 1, "test"                	
    ...  	
    call static_assert(u1 0, v14, v29, u1 0)      	
    ...             	
    v43 = unchecked_add v0, u32 1                 	
    jmp b1(v43)
  b3():
    return
}
```
And this is with the simplification:
```rust
After `as_slice` optimization (1) (step 11):
acir(inline) fn main f0 {
  b0():
    ...
    jmp b1(u32 0)
  b1(v0: u32):
    jmpif u1 0 then: b2, else: b3
  b2():
    ...
    v41 = unchecked_add v0, u32 1                 	
    jmp b1(v41)
  b3():
    return
}
```
Notice how the terminator of `b1` had `v30` replaced by `u1 0`, so it always go to `b3`. Why does it fail, then?
That's because the static assertion pass [looks at the loop boundaries](https://github.com/noir-lang/noir/blob/56325c40cf0a3986152c9b554cbab6ab97b4af4b/compiler/noirc_evaluator/src/ssa/opt/evaluate_static_assert_and_assert_constant.rs#L84-L86) to find blocks which are in empty loops, and that assumes having an induction variable there, not a constant, so it appears as a non-empty loop. 

I'll have a look to see if we can enhance loop detection for this, but I had trouble with it before during unrolling.

compiler/noirc_evaluator/src/ssa/opt/simple_optimization.rs
```diff
@@ -124,8 +136,39 @@ impl SimpleOptimizationContext<'_, '_> {
...
+    /// If `simplify_instruction_if_changed` is `true` and the instruction or its values have been updated,
+    /// then it will attempt to simplify it before re-insertion. This should only be enabled after the CFG
+    /// itself has been flattened, as it has a chance to replace induction variables in `jmpif` terminators
+    /// with constants, which hinders the recognition of loops.
     pub(crate) fn insert_current_instruction(&mut self) {
-        self.dfg[self.block_id].insert_instruction(self.instruction_id);
+        // If the instruction changed, then there is a chance that we can (or have to)
+        // simplify it before we insert it back into the block.
+        let instruction_hash = fxhash::hash64(self.instruction());
+        let instruction_changed = self.orig_instruction_hash != instruction_hash;
```
asterite: We could avoid computing this hash if `simplify_instruction_if_changed` is false, though I know it will make the code slightly uglier, so this is just an optional suggestion.

asterite (approved):

compiler/noirc_evaluator/src/ssa/opt/simple_optimization.rs
aakoshh: Yeah I thought the same thing, it might not be worth having a field that is 0 in esoteric ways. I

'm still trying to fix unrolling in the face of simplified jumps, if it works, we might not need this flag.

aakoshh: Removed now, hopefully nothing breaks 🤞

aakoshh:
Okay so now the simple optimisation loop simplifies any instruction that changed, and unrolling does some extra work to restore loop bounds which were simplified out. 

A potential downside is that e.g. the `remove_if_else` pass, which is anyway a bit complicated, now has an output which is yet another step removed from the input. We considered disabling the simplification for this pass, or just its test, but decided to keep the number of free variables simple and just accept that the SSA changes more. After all, flattening and mem2reg, and whatever else uses the `FunctionInserter`, also simplify stuff.

PR #9712, opened 2025-09-02
fix(ssa): Mark whether an ArrayGet requires a predicate based upon the runtime

vezenovm:
# Description

## Problem\*

No issue just something I noticed while looking at the purity pass once more. 

## Summary\*

I simply gated whether an `ArrayGet` requires a predicate based upon the runtime. This check technically more accurately reflects the functionality of an ArrayGet in isolation in an ACIR vs. Brillig runtime. I would expect this change to only really have an effect on loop hoisting and deduplication. With regards to function purity, ArrayGet's in Brillig are expected to be preceded with `constrain` statements for the OOB checks which would already be marking the function at least as `PureWithPredicate`. 

## Additional Context

I was thinking we should rename `requires_acir_gen_predicate` to just `requires_predicate` with some extra documentation about what constitutes a "predicate" for our different backend targets. We use `requires_acir_gen_predicate` to indicate a predicate is required in both runtimes throughout the compiler already so we may as well accurately name the function. 

## Documentation\*

Check one:

# PR Checklist\*


asterite:
Would it make sense to always check if the runtime is ACIR? Like, here we check if it's ACIR in ArrayGet but why isn't the same done for Binary? (I'm just trying to understand). But this looks good (not sure if the regressions are desirable though?)

vezenovm:
> Would it make sense to always check if the runtime is ACIR? Like, here we check if it's ACIR in ArrayGet but why isn't the same done for Binary? (I'm just trying to understand). 

It would for cases where they differ. For `Binary` we inject the same overflow checks in both ACIR/Brillig. It could also be clearer to make this method per backend target even if there is redundant logic. I think that is better for a separate PR though. 

> But this looks good (not sure if the regressions are desirable though?)

Yeah I originally thought we were seeing us hoist an instruction under a predicate that is not taken. This is a general trade-off that exists for hoisting non control dependent loop invariants under predicates. However, if we look at a simplfiication of `loop_invariant_regression` and run `nargo info --force-brillig --profile-execution --inliner-aggressiveness 0`:
```noir
fn main(x: u32, y: u32) {
    array_read_loop(4, x);
}
fn array_read_loop(upper_bound: u32, x: u32) {
    let arr = [2; 5];
    for i in 0..upper_bound {
        for j in 0..upper_bound {
            for _ in 0..upper_bound {
                assert_eq(arr[i], x);
                assert_eq(arr[j], x);
            }
        }
    }
}
```
`master`: 266 Brillig opcodes executed
This branch: 290 opcodes executed

This result surprises me as I would not expect anything new to be hoisted after this PR. I think I am missing also updating `Instruction::has_side_effects` which is used by LICM for checking whether previous instructions are impure in a block for a control dependent invariant. 

EDIT: Now that is more like what I was expecting :) https://github.com/noir-lang/noir/pull/9712#issuecomment-3246706362

asterite (approved):
Looks good!

Now I'm noticing that `is_safe_index` is always preceded by `is_acir`, so maybe `is_safe_index` could check if the runtime is acir. There are two more calls to `is_safe_index`:
- one in DIE, but it's only called if the runtime is ACIR
- one in ACIR gen, where the runtime is obviously ACIR 😄 

We could maybe rename `is_safe_index` to something like `index_might_be_out_of_bounds` explaining why in Brillig it can't be OOB, but that name is long so maybe it could be something else...

vezenovm:
> We could maybe rename `is_safe_index` to something like `index_might_be_out_of_bounds` explaining why in Brillig it can't be OOB, but that name is long so maybe it could be something else...

This feels like it may be a foot-gun on its own. Technically the index is also still "out of bounds" in Brillig. The array get just happens to be a pure operation even when out of bounds.

aakoshh:
This change allows an unsafe index access to be hoisted out of a loop which never executes in https://github.com/noir-lang/noir/issues/9804  leaving the corresponding constraint unhoisted. 

I'm still investigating whether this is due to some refactoring I made, or maybe the loop shouldn't be touched because it has a `break`, but just wanted to ask whether this is what you expected would happen.

EDIT: It looks like the test case fails on this commit already, so it's not something I messed up.

PR #9717, opened 2025-09-03
chore(test): Run `ram_blowup_regression` only on CI

aakoshh:
# Description

## Problem\*

Executing integration can be slow, in particular `ram_blowup_regression` takes up a disproportionate amount of time. 

## Summary\*

Changes the `nargo_cli` integration tests to exclude `ram_blowup_regression` unless the `CI` env var is set.

## Additional Context

Say I want to have an idea whether commenting out some code in an SSA pass breaks any of the integration or regression tests. I can run the following command, which runs almost 700 tests; way less than the total (over 5000). With this PR, this takes just over 15 seconds on my laptop. 

```shell
cargo test -p nargo_cli --test execute forcebrillig_false_inliner_i64_min
```

However if I include the `ram_blowup_regressions`, this goes up to almost 2 minutes, which means I'm less likely to run the tests:

```console
❯ CI=1 cargo test -p nargo_cli --test execute ram_blowup_regression::forcebrillig_false_inliner_i64_min
    Finished `test` profile [optimized + debuginfo] target(s) in 0.61s
     Running tests/execute.rs (target/debug/deps/execute-44c3ac57c912e029)

running 2 tests
test tests::execution_success::test_ram_blowup_regression::forcebrillig_false_inliner_i64_min_expects has been running for over 60 seconds
test tests::interpret_execution_success::test_ram_blowup_regression::forcebrillig_false_inliner_i64_min_expects has been running for over 60 seconds
test tests::interpret_execution_success::test_ram_blowup_regression::forcebrillig_false_inliner_i64_min_expects ... ok
test tests::execution_success::test_ram_blowup_regression::forcebrillig_false_inliner_i64_min_expects ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 5381 filtered out; finished in 108.74s
```

(It is not exactly clear why it's so slow, because if I `cd` into the directory and use `cargo run -p nargo_cli -- execute --force` then it's fast).

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/tooling/nargo_cli/build.rs
+++ b/tooling/nargo_cli/build.rs
@@ -28,6 +28,7 @@ fn main() -> Result<(), String> {
     // Rebuild if the tests have changed
     println!("cargo:rerun-if-changed=tests");
     println!("cargo:rerun-if-changed={}", test_dir.as_os_str().to_str().unwrap());
+    println!("cargo:rerun-if-env-changed=CI");
 
     generate_execution_success_tests(&mut test_file, &test_dir);
     generate_execution_failure_tests(&mut test_file, &test_dir);
@@ -68,6 +69,9 @@ fn main() -> Result<(), String> {
 
     Ok(())
 }
+
+/// Some tests take a very long time to run, which discourages local testing.
+const CI_ONLY_TESTS: [&str; 1] = ["ram_blowup_regression"];
 
 /// Some tests are explicitly ignored in brillig due to them failing.
 /// These should be fixed and removed from this list.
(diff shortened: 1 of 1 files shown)


asterite:
I think this means this test will never run, as it currently doesn't run in CI: https://github.com/noir-lang/noir/blob/910cd9349e0b9382c1512935bc3e1d791891bd21/.config/nextest.toml#L1-L6

asterite:
Maybe it could only run in nightly? Not sure how that runs but I know there's something that runs nightly from which we get fuzz failures?

TomAFrench:
We run this test in CI on the master branch. We just disable it for PRs to make them quicker to merge.

I'd like to lean into just using nextest profiles if we want to vary how tests are run under certain circumstances.

aakoshh:
Okay thanks, I'll use nextest to pick a profile locally 👍 

```
cargo nextest run -p nargo_cli --test execute forcebrillig_false_inliner_i64_min --filter-expr "not test(ram_blowup)"
```

asterite:
I do agree that `ram_blowup_regression` takes a lot of time, and lately I run the tests locally very frequently, so having something as a default for everyone (except master) would be great.

PR #9718, opened 2025-09-03
chore(licm): Identify untested code; refactoring; minor fixes

aakoshh:
# Description

## Problem\*

Trying to find pieces of logic not covered by unit or integration tests.

## Summary\*

Add more unit test coverage to LICM.

Refactor and fix some stuff:
* Remove `Constrain` and `RangeCheck` from `can_be_hoisted_from_loop_bounds`: they don't look at the bounds, and the conditions they check are redundant with `can_be_hoisted`.
* Remove `Call` from `can_be_hoisted_from_loop_bounds`: same reason as above.
* Remove `BlockContext::can_be_hoisted_with_control_dependence`: I thought the extra jumping around to call `can_be_hoisted` again with `true` was confusing, and it was clearer to see the calls to `can_be_hoisted` with `false` and `true` next to each other, with the latter being restricted by `block_context.can_hoist_control_dependent_instruction()`.
* Changed `can_be_hoisted` to return `Yes | No | WithPredicate`, rather than having to call it twice with `false` and `true`; this way if it returns `No` we don't have to retry in futility, and I think it reads clearer.
* Return `false` for loops without an exit block from the header in `is_fully_executed`. It used return `true`, but two examples I found were 1) an actual infinite `loop` in `regression_7103` and 2) a `while true` loop with a `break` created by the fuzzer to avoid infinite loops. The latter is clearly not in the spirit of the method, as it has a `break`.
* Handle loops with headers that have no invariant parameter in `get_const_upper_bound`, e.g. `while <bool-variable>`.
* Consider the loop header as an ancestor of the blocks in the loop for calculating `is_impure`. 
* Fix the simplification of `ConstrainNotEqual`

### Untested code

The following changes don't cause any test failure:
  - `can_be_hoisted_from_loop_bounds`: 
    - [x] `ArrayGet` can be disabled 
    - [x] `ArrayGet` can `return true` instead of `false`
    - [x] `Constrain(..) | ConstrainNotEqual(..) | RangeCheck { .. }` can be disabled 
    - [x] `Call { func, .. } ` can be disabled
  - `can_be_hoisted`:
    - [x] `Call` / `Purity::Pure`: return `false` instead of `true`
    - [x] `Call` / `Purity::PureWithPredicate`: return `false` instead of `hoist_with_predicate`
    - [x] `Call` / purity `None`: return `true` instead of `false`
    - [x] `MakeArray`: return `true` or `false` instead of `function.runtime().is_acir()`
  - `can_evaluate_binary_op`: 
    - [x] `Add | Mul | Sub` when `unchecked: true`: return `false` instead of `true`
    - [x] `Div | Mod`: return `false` instead of `true` when `!value.is_zero()`
  - `simplify_from_loop_bounds`:
    - [x] `SimplifyResult::Remove`: return `false` instead of `true`
  - `simplify_induction_variable`:
    - [x] `Constrain`: remove `loop_context.no_break` from condition
    - `ConstrainNotEqual`: 
      - [ ] remove `loop_context.no_break` from condition
      - [ ] remove `block_context.can_simplify_control_dependent_instruction()` from condition
      - [x] disable altogether
  - `simplify_induction_variable_in_binary`:
    - [x] `false if upper_bound <= constant.inc()` if we remove `.inc()`
    - [x] `false if lower_bound > constant` if we return `false_value` instead of `true_value`
    - [x] `true if lower_bound >= constant` if we return `true_value` instead of `false_value`
    - [x] `false if upper_bound <= constant.inc()` if we return `true_value` instead of `false_value`

This is not an exhaustive list, just what I found by tinkering with conditions.

I'll open follow up PRs to add tests, but wanted to get this PR reviewed because of the refactoring it contains.

## Additional Context

I used the following commands to look for failing tests:
```
cargo nextest run --cargo-quiet -p noirc_evaluator loop_invariant
cargo nextest run --cargo-quiet -p nargo_cli --test execute forcebrillig_false_inliner_i64_min --filter-expr "not test(ram_blowup)"
```

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench (approved):
LGTM

PR #9719, opened 2025-09-03
chore: simplify signed shift right

asterite:
# Description

## Problem

No issue, just something I didn't fully understand while greenlighting remove_bit_shifts.

## Summary

When simplifying a signed shift right the SSA we produced was:
- cast lhs to Field
- add one to it, if it's negative
- truncate it to the original type
- cast it back to the original type
- perform the division
- subtract one to it, if it's negative
- truncate it

I noticed two things:
1. I didn't understand why going to Field to perform the first operation was necessary
2. I didn't understand why the final truncation was needed, if it's truncating to N bits a value that has a maximum value of N bits

I finally understood why the final truncation is needed (and also the intermediate one). I documented this in the `Truncate` instruction. I'm still not sure it's intuitive (I feel there's a leaky abstraction here) but at least now it's documented.

Then, going to Field isn't necessary because truncating already takes care of doing it the right way. This doesn't result in any ACIR improvement but at least it's less Rust code and less SSA instructions.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
@@ -197,23 +197,20 @@ impl Context<'_, '_, '_> {
                 let zero =
                     self.numeric_constant(FieldElement::zero(), NumericType::signed(bit_size));
                 let lhs_sign = self.insert_binary(lhs, BinaryOp::Lt, zero);
-                let lhs_sign_as_field = self.insert_cast(lhs_sign, NumericType::NativeField);
-                let lhs_as_field = self.insert_cast(lhs, NumericType::NativeField);
-                // For negative numbers, convert to 1-complement using wrapping addition of a + 1
-                // Unchecked add as these are fields
-                let one_complement = self.insert_binary(
-                    lhs_sign_as_field,
-                    BinaryOp::Add { unchecked: true },
-                    lhs_as_field,
-                );
-                let one_complement = self.insert_truncate(one_complement, bit_size, bit_size + 1);
-                let one_complement =
-                    self.insert_cast(one_complement, NumericType::signed(bit_size));
-                // Performs the division on the 1-complement (or the operand if positive)
-                let shifted_complement = self.insert_binary(one_complement, BinaryOp::Div, pow);
-                // Convert back to 2-complement representation if operand is negative
                 let lhs_sign_as_int = self.insert_cast(lhs_sign, lhs_typ);
 
+                // For negative numbers, convert to 1-complement using wrapping addition of a + 1.
+                // We achieve wrapping addition with an unchecked add followed by truncation. For example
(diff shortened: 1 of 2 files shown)


asterite:
I'm thinking the alternative would be to still use Field for this operations. What was missing is also doing the subtraction using Field so it would wrap around, then truncating the Field value. That way we could also try simplifying `truncate v0, bit_size: 8` when `v0` has 8 bits or less.

Thoughts?

asterite:
Or maybe nevermind the above, I noticed we do this tricky in a few places already.

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -224,6 +221,10 @@ impl Context<'_, '_, '_> {
                     BinaryOp::Sub { unchecked: true },
                     lhs_sign_as_int,
                 );
+
+                // Similar to the above wrapping addition, this is a wrapping subtraction because if
+                // `shifted_complement` is `i8 0`, so `Field 0` which, so the subtraction would give
+                // the maximum Field value which, when truncated, equals `i8 -1`.
```
TomAFrench, line 227: I don't think this is correct. It's commented above that this cannot overflow so we're just using an unchecked operation to avoid an unnecessary range check.

 If `shifted_complement == 0` and `lhs_sign_as_int == 1` as required for this to overflow then we'd get `0x30644e72e131a029b85045b68181585d2833e84879b9709143e1f593f0000000` as the result which would not truncate into a signed type correctly.

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -197,23 +197,20 @@ impl Context<'_, '_, '_> {
...
-                let lhs_sign_as_field = self.insert_cast(lhs_sign, NumericType::NativeField);
-                let lhs_as_field = self.insert_cast(lhs, NumericType::NativeField);
-                // For negative numbers, convert to 1-complement using wrapping addition of a + 1
-                // Unchecked add as these are fields
-                let one_complement = self.insert_binary(
-                    lhs_sign_as_field,
-                    BinaryOp::Add { unchecked: true },
-                    lhs_as_field,
-                );
-                let one_complement = self.insert_truncate(one_complement, bit_size, bit_size + 1);
```
TomAFrench, line 209: I'm not a fan of us allowing values to escape the bounds implied by their type as we could have logic elsewhere in the compiler making assumptions based on type information which this would then break.

Casting to `Field` in this case ensures that this does not happen and is safer.

asterite: In `expand_signed_checks` we also perform some math in signed types, then truncate to the max bit size (which in my mind should be a no-op). I tried casting to Field there too before doing the math but it doesn't seem to work or to be a straight-forward change, so for now I'll drop this PR.

PR #9720, opened 2025-09-03
chore: add `LoopContext::new`

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR simply moves `init_loop_context` into a method on `LoopContext` called `new` in order to be more idiomatic.

I've left `init_block_context` alone as it pulls in a lot of `LoopInvariantContext` fields and mutates the `LoopContext`.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
@@ -897,7 +897,7 @@ mod test {
     use crate::ssa::ir::instruction::{Instruction, Intrinsic, TerminatorInstruction};
     use crate::ssa::ir::types::Type;
     use crate::ssa::opt::loop_invariant::{
-        CanBeHoistedResult, LoopInvariantContext, can_be_hoisted,
+        CanBeHoistedResult, LoopContext, LoopInvariantContext, can_be_hoisted,
     };
     use crate::ssa::opt::pure::Purity;
     use crate::ssa::opt::unrolling::Loops;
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -274,6 +274,42 @@ impl BlockContext {
 }
 
 impl LoopContext {
+    /// Create a `LoopContext`:
+    /// * Gather the variables declared within the loop
+    /// * Determine the induction variable bounds
+    fn new(
+        inserter: &FunctionInserter,
```
aakoshh, line 282: Maybe it can take `&Function`, I think it doesn't use the inserter itself.

TomAFrench: It does in the call to `get_induction_var_bounds`

aakoshh:
> I've left init_block_context alone as it pulls in a lot of LoopInvariantContext fields and mutates the LoopContext.

I'd consider creating `BlockContext::new` just to match the other two even if it's a bit needy. Originally I thought the two `init_` next to each other give an easy overview of how these variables evolve, but if we have 2 `new` contexts already, a 3rd could be an easier pattern to look for.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
aakoshh: Sorry, I meant all it does it get the `function` from it:
```rust
fn get_induction_variable(inserter: &FunctionInserter, loop_: &Loop) -> Option<ValueId> {
    loop_.get_induction_variable(inserter.function).map(|v| inserter.resolve(v))
}

fn get_induction_var_bounds(
    inserter: &FunctionInserter,
...
) -> Option<(ValueId, (IntegerConstant, IntegerConstant))> {
    let bounds = loop_.get_const_bounds(&inserter.function.dfg, pre_header)?;
    let induction_variable = get_induction_variable(inserter, loop_)?;
...
```

aakoshh: Ah, no, you're right, it resolves 🙈

TomAFrench:
Yeah, I think we could do something with more refactoring of this code but I don't think I have the time to unpick everything to get it into this PR.

asterite (approved):

PR #9723, opened 2025-09-03
fix: left bit shift u128 would overflow Field

### LINKED ISSUE #9716: add tests for bitshifting u128s in acir
I think this is a situation where we are under testing currently. I wouldn't be surprised if there was scope for field overflow when shifting a large u128 by 127 bits.

We should add tests along with any necessary overflow protection.

asterite:
# Description

## Problem

Resolves #9716

## Summary

Effectively, shifting a large u128 to the left by a large number would overflow the maximum Field element, ultimately giving an incorrect result.

I thought of one way to do this. Let me know if there's a better way!

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -166,15 +164,50 @@ impl Context<'_, '_, '_> {
...
+            let pow = self.two_pow(rhs);
             let lhs_field = self.insert_cast(lhs, NumericType::NativeField);
             // Unchecked mul as this is a wrapping operation that we later truncate
             let result = self.insert_binary(lhs_field, BinaryOp::Mul { unchecked: true }, pow);
             let result = self.insert_truncate(result, typ.bit_size(), max_bit);
             self.insert_cast(result, typ)
+        } else {
+            // Otherwise, the result might not bit in a FieldElement.
+            // For this, if we have to do `lhs << rhs` we can first shift by half of `rhs`, truncate,
+            // then shift by `rhs - half_of_rhs` and trunate again.
```
asterite:
```suggestion
            // then shift by `rhs - half_of_rhs` and truncate again.
```

TomAFrench:
> I thought of one way to do this. Let me know if there's a better way!

I'm not sure if this is better or worse but one alternative would be to apply a mask before we perform the shift so x << i becomes `x%(2^(128-i)) * 2^i`

This avoids us doing two unknown powers (as we can calculate 2^i once and reuse it)

TomAFrench (approved):
LGTM, we could explore the affects of the above alternative method but not blocking

asterite:
> would be to apply a mask before we perform the shift so x << i becomes x%(2^(128-i)) * 2^i

Nice idea!

However, I'm trying it now and I'm not sure how to do the operations. I guess to reuse `2^i`, for `2^(128 - i)` we can do `2^128 / 2^i` where `2^128` is a known constant. But to do that division we can't use Field, we have to use u128 and `2^128` doesn't fit in `u128`.

Another thing I tried is truncating by `128 - i` but the truncate operation always requires a constant so that doesn't work.

Alternatively we can compute `2^(128 - i)` by doing `128 - i`, then `2^(128 - i)` but that will require two unknown pows like what's in the current PR, so I don't know if it'll improve things.

I'll still capture an issue for this in case I'm missing something!

PR #9736, opened 2025-09-04
fix(ssa): Accurate purities of ArraySet and RC instructions in Brillig

### LINKED ISSUE #9735: purity_analysis: ArraySet marked incorrectly and IncrementRc/DecrementRc overly conservative in Brillig
### Problem

Right now we mark all IncRc/DecRc instructions as `Impure`. However, if we never have an RC instructions that operates on a parameter all mutations should be local to that function. This is the same case for ArraySet. We can potentially mutate an array that is external to the function if we perform the set on a parameter. However, if all array sets are local to the function the operation is still pure. 

### Happy Case

Accurately mark the purity of Brillig functions containing array reference counts and array sets to reflect their actual functionality.

1. If an inc_rc/dec_rc occurs on a parameter then the function is Impure as we are operating on a pointer that is external to the function
2. If an inc_rc/dec_rc never occurs on a parameter the function is pure
3. (potential bug) If an array set occurs on a parameter then the function is impure as we can potentially mutate that array. I will try to create a regression test here to show this failure (similar to #9451). I believe our inc_rc/dec_rc has protected us here. I may split this out into a separate issue.  
4. If an array set never occurs on a parameter then the function is pure

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #9746: purity_analysis: ArraySet in Brillig is not pure
### Aim

In Brillig, ArraySet can potentially mutate an array pointer (if the RC is equal to one).

The following test will fail on the final interpretation:
```rust
    #[test]
    fn do_not_deduplicate_call_with_array_set_brillig() {
        let src = "
        brillig(inline) fn main f0 {
          b0(v0: u32):
            v2 = make_array [Field 1, Field 2] : [Field; 2]
            call f1(v2, Field 9)
            v7 = array_set v2, index v0, value Field 7
            call f1(v2, Field 9)
            v9 = array_get v2, index v0 -> Field
            constrain v9 == Field 9
            return
        }
        brillig(inline) fn mutator f1 {
          b0(v0: [Field; 2], v1: Field):
            v3 = array_set v0, index u32 0, value v1
            return
        }
        ";
        println!("{}", ssa.print_with(None));
        ssa.interpret(vec![Value::from_constant(0_u32.into(), NumericType::unsigned(32)).unwrap()])
            .unwrap();

        let ssa = ssa.purity_analysis();
        ssa.interpret(vec![Value::from_constant(0_u32.into(), NumericType::unsigned(32)).unwrap()])
            .unwrap();

        let ssa = ssa.fold_constants_using_constraints();
        println!("{}", ssa.print_with(None));
        ssa.interpret(vec![Value::from_constant(0_u32.into(), NumericType::unsigned(32)).unwrap()])
            .unwrap();
    }
```
Where this is the failing SSA after constant folding:
```
  b0(v0: u32):
    v3 = make_array [Field 1, Field 2] : [Field; 2]
    call f1(v3, Field 9)
    v7 = array_set v3, index v0, value Field 7
    v8 = array_get v3, index v0 -> Field
    constrain v8 == Field 9
    return
}
brillig(inline) predicate_pure fn mutator f1 {
  b0(v0: [Field; 2], v1: Field):
    v3 = array_set v0, index u32 0, value v1
    return
}
```

The example is slightly contrived as the compiler should not be producing this type of SSA (we will have preceding inc_rc/dec_rc instructions). However, this still shows that we are marking array sets improperly and if there were ever a bug in the frontend it would be harder to catch. 

### Expected Behavior

I expect constant folding to not fail after constant folding is run with purity information. 

### Bug

Fails with the following on the final interpreter run:
```
called `Result::unwrap()` on an `Err` value: ConstrainEqFailed { lhs: "Field 7", lhs_id: Id(8), rhs: "Field 9", rhs_id: Id(5), msg: None }
stack backtrace:
```

### To Reproduce

1. Place the test above into the `constant_folding` test module and run it

### Workaround

None

### Workaround Description

We have been kind of saved by our frontend correctly outputting inc_rc instructions, but we should still mark these correctly. 

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #9735
Resolves #9746 

## Summary\*

This PR only changes purity in the Brillig runtime. `ArraySet`, `IncrementRc`, and `DecrementRc` are marked as impure if they operate on a function parameter or a global. Otherwise they are pure instructions. 

## Documentation\*

Check one:

# PR Checklist\*


asterite (approved):

TomAFrench:
I've bumped this from the merge queue so that https://github.com/noir-lang/noir/pull/9752 can go in. We need that to go in as it will conflict with basically anything which affects a snapshot.

PR #9738, opened 2025-09-04
chore(ssa): Consolidate should inline check into a single filter

### LINKED ISSUE #9404: Inlining has two distinct checks for whether a function should be inlined

vezenovm:
# Description

## Problem\*

Resolves #9404 

## Summary\*

We perform an inline info computation that sets whether or not we should inline a function. That inline info computation is accessed through a closure we pass along to the inlining context. We also have a second `PerFunctionContext::should_inline_call`. In fact, something the issue missed is that when we have extra inline logic when we set the `should_inline_call` closure to determine whether or not we should preserve a no_predicates function.

This PR moves all determination of the inlining policy into the inline info computation. I left the runtime validation check inside of inlining itself as this still felt like the correct place to error check. The new method `validate_callee` is essentially an  extra validation check on whether we have well-formed SSA input to the inlining pass. 

## Additional Context

I had to update `inline_simple_functions` to account for Brillig entry points. This logic is redundant for what we do during the inline info computation but is necessary for correctness without #9486. This adds another motivation for #9486 as `inline_simple_functions` is essentially another ad-hoc inlining filter. I find it easier to reason about changes to the inlining policy when it is all contained within a single location.

Looks like we have some good compilation time benefits as well. In this PR we started skipping the weight computation for ACIR and Brillig entry points which I imagine is where most of the benefit comes from. 

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
> `sha512-100-bytes`	`17405` opcodes	`13173` opcodes	`1.32`

Hmm this is surprising. Something must be off still. I'm adding the bench label to get a more full picture.

EDIT: This was due to me not accounting for the ad-hoc inlining policy in `inline_simple_functions`.

vezenovm:

TomAFrench (approved):
LGTM

PR #9739, opened 2025-09-04
chore(inlining): Skip weight calc for ACIR functions and recursive Brillig functions

vezenovm:
# Description

## Problem\*

No issue just something I noticed while wokring on #9738. We have ACIR functions set to always be inlined unless they are an entry point. Thus, computing the weight of them is wasted computation.

## Summary\*

I decided to branch a draft off of #9738 as this is slightly separate from the goal of that PR. We can also see the benchmarks in isolation here. Even if it does not have a large effect on compilation times I think this would still be worthy of merging.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/inlining/inline_info.rs
```diff
@@ -166,6 +166,14 @@ fn compute_function_should_be_inlined(
     }
 
     let function = &ssa.functions[&func_id];
+    let runtime = function.runtime();
+
+    if runtime.is_acir() {
+        let info = inline_infos.entry(func_id).or_default();
+        info.should_inline = true;
+        return;
+    }
```
TomAFrench, line 175: This doesn't take into account of `#[fold]` (and maybe no_predicates)

vezenovm: In the parent PR #9738 we moved to skipping the processing of these entry points so they should all be marked to not be inlined.

vezenovm:
> `test_report_AztecProtocol_aztec-packages_noir-projects_noir-protocol-circuits_crates_blob`	`199` s	`266` s	`0.75`

Most meaningful benefit seen here.

PR #9742, opened 2025-09-05
chore(acir): improve error message for static regex compilation

strmfos:
# Description

Replace generic "failed to create regex" with more descriptive error message that clarifies this is a static regex pattern that should always compile successfully. This provides better context for developers if compilation unexpectedly fails.
## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
I don't think this really adds much value as this is in purely test code and any panic will have a callstack pointing to this line.

PR #9749, opened 2025-09-08
chore(evaluator): fix BigIntToLeBytes to always output 32 bytes

Fibonacci747:
# Description

Replace modulus-derived output length with a fixed output_count = 32 and remove the assert. This change is necessary because the BigIntToLeBytes contract across the stack (frontend interpreter docs/logic, ACVM pedantic checks, and Brillig VM implementation) mandates a 32-byte little-endian output with zero-padding. The previous behavior could panic or desynchronize sizes for non-standard moduli and conflicted with the global expectation of a fixed 32-byte output.

# PR Checklist\*


TomAFrench:
This isn't necessary as these are being removed.

PR #9761, opened 2025-09-08
chore(ssa): Flatten basic Brillig conditionals before inlining of simple functions

vezenovm:
# Description

## Problem\*

No issue just something I noticed when working on #9487. There are sometimes functions that would have a lot of knockdown simplifications but get muddled as being complex due to having multiple blocks. 

## Summary\*

I am pushing this as a draft to see if it has any large effect on compilation times. 

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
The results are not wide reaching enough to make me want to merge this as is so I am closing the PR. The main takeaway is that the `trait_impl_base_type_inliner_zero` results indicate our inliner still has some room for improvement when determining "simple" functions.

PR #9766, opened 2025-09-08
chore: add extra constraint folding pass

TomAFrench:
# Description

## Problem\*

This is a quick fix for #9767 

## Summary\*

I noticed that we don't ever deduplicate any fallible instructions which appear in the same block after loop unrolling so this PR adds a quick solution to this by running another pass of the constant folding pass with constraint info (for some reason this is necessary for us to deduplicate like this?)

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/constant_folding/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/constant_folding/mod.rs
@@ -814,7 +814,7 @@ mod test {
         }";
 
         let ssa = Ssa::from_str(src).unwrap();
-        let ssa = ssa.fold_constants();
+        let ssa = ssa.fold_constants(MIN_ITER);
 
         assert_ssa_snapshot!(ssa, @r"
         brillig(inline) predicate_pure fn main f0 {


compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -1116,6 +1116,32 @@ mod test {
...
+          b0(v0: Field):
+            v1 = div Field 1, v0
+            v2 = div Field 1, v0
+            return
+        }";
+
+        let ssa = Ssa::from_str(src).unwrap();
+        let ssa = ssa.fold_constants();
+
+        // The terminators of b1 and b2 should now have constant arguments
```
aakoshh: This comment seems to be a remnant of copy pasting from another test.

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -1116,6 +1116,32 @@ mod test {
...
+        assert_ssa_snapshot!(ssa, @r"
+        brillig(inline) predicate_pure fn main f0 {
+          b0(v0: Field):
+            v2 = div Field 1, v0
+            v3 = div Field 1, v0
```
aakoshh, line 823: Shouldn't this be deduplicated now?

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
```diff
@@ -1116,6 +1116,32 @@ mod test {
         assert_normalized_ssa_equals(ssa, expected);
     }
 
+    // TODO: https://github.com/noir-lang/noir/issues/9767
```
aakoshh, line 804: This PR says it's a quick fix for this exact ticket. What's the TODO about?

TomAFrench: The issue is that we don't deduplicate these instruction unless we're using constraint information while constant folding. I've opened that issue so that we get a proper fix for this but this PR is just to add another round of constant folding with constraints later in the SSA pipeline to have a round of deduplication after we do loop unrolling.

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
TomAFrench: It is if we do `constant_fold_with_constraints`, this test is to check that we do it with regular constant folding as well.

compiler/noirc_evaluator/src/ssa/opt/constant_folding.rs
TomAFrench:
```suggestion
```

TomAFrench: Ah yes, thanks

vezenovm:
> `rollup-block-root-empty`	`1130` MB	`998.53` MB	`1.13`
> `rollup-block-root-single-tx`	`10660` MB	`9680` MB	`1.10`
> `rollup-block-root`	`10660` MB	`9690` MB	`1.10`

If we are ok with the slight compilation memory degradation I am good with merging.

PR #9769, opened 2025-09-08
chore: always interpret brillig calls

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR switches us over to always use the interpreter when constant folding.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/constant_folding/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/constant_folding/mod.rs
@@ -2965,7 +2965,7 @@ mod test {
         }
         "#;
         let ssa = Ssa::from_str(src).unwrap();
-        let ssa = ssa.fold_constants_with_brillig(DEFAULT_MAX_ITER);
+        let ssa = ssa.fold_constants(DEFAULT_MAX_ITER);
         // The `make_array` should be hoisted into b3, not b4.
         assert_ssa_snapshot!(ssa, @r#"
         brillig(inline) impure fn main f0 {
@@ -3029,7 +3029,7 @@ mod test {
         }
         "#;
         let ssa = Ssa::from_str(src).unwrap();
-        let ssa = ssa.fold_constants_with_brillig(DEFAULT_MAX_ITER);
+        let ssa = ssa.fold_constants(DEFAULT_MAX_ITER);
 
         assert_ssa_snapshot!(ssa, @r"
         brillig(inline) impure fn main f0 {
@@ -3085,7 +3085,7 @@ mod test {
       }
       ";
         let ssa = Ssa::from_str(src).unwrap();
-        let ssa = ssa.fold_constants_with_brillig(DEFAULT_MAX_ITER);
+        let ssa = ssa.fold_constants(DEFAULT_MAX_ITER);
         let elements = vec![Value::field((-2i128).into()), Value::field((-1i128).into())];
         let inputs = Value::array(elements, vec![Type::field()]);
         let results = ssa.interpret(vec![inputs]).unwrap();


TomAFrench:
Interesting that this causes regressions. @vezenovm you switched us over to the interpreter so I imagine you thought about this - do you know why this causes regressions?

I guess we could be affecting the sizes of brillig functions which call other brillig functions here which causes problems in the inliner?

vezenovm:
> Interesting that this causes regressions. @vezenovm you switched us over to the interpreter so I imagine you thought about this - do you know why this causes regressions?
> 
> I guess we could be affecting the sizes of brillig functions which call other brillig functions here which causes problems in the inliner?

All the constant folding passes are run after all the inlining passes so I don't think this is the cause of the regressions.

compiler/noirc_evaluator/src/ssa/mod.rs
```diff
@@ -185,7 +185,6 @@ pub fn primary_passes(options: &SsaEvaluatorOptions) -> Vec<SsaPass<'_>> {
         SsaPass::new(Ssa::remove_truncate_after_range_check, "Removing Truncate after RangeCheck"),
         SsaPass::new(Ssa::checked_to_unchecked, "Checked to unchecked"),
         SsaPass::new(Ssa::fold_constants_using_constraints, "Constant Folding using constraints"),
-        SsaPass::new(Ssa::fold_constants_with_brillig, "Inlining Brillig Calls"),
```
vezenovm: This looks to me like the cause of the regressions. What will the benchmarks be if we call `fold_constants` here?

compiler/noirc_evaluator/src/ssa/mod.rs
```diff
@@ -185,7 +185,6 @@ pub fn primary_passes(options: &SsaEvaluatorOptions) -> Vec<SsaPass<'_>> {
         SsaPass::new(Ssa::remove_truncate_after_range_check, "Removing Truncate after RangeCheck"),
         SsaPass::new(Ssa::checked_to_unchecked, "Checked to unchecked"),
         SsaPass::new(Ssa::fold_constants_using_constraints, "Constant Folding using constraints"),
```
TomAFrench:
```suggestion
        SsaPass::new(Ssa::fold_constants_using_constraints, "Constant Folding using constraints"),
        SsaPass::new(Ssa::fold_constants, "Constant Folding"),
```

compiler/noirc_evaluator/src/ssa/mod.rs
TomAFrench: Hmm, I was thinking that this duplicate would be unnecessary if we're going to be handling brillig calls in `fold_constants_using_constraints`. Maybe something is being hoisted up to another block so the second fold is necessary to do followup optimizations.

asterite:
Is the regression acceptable?

TomAFrench:
The performance alerts above are outdated. I've merged master into here so we can get an up-to-date view of the performance.

TomAFrench:
@jfecher can you take a look?

jfecher:
Confirmed the compilation time regressions are good now

compiler/noirc_evaluator/src/ssa/mod.rs
```diff
@@ -237,7 +237,7 @@ pub fn primary_passes(options: &SsaEvaluatorOptions) -> Vec<SsaPass<'_>> {
             "Constant Folding using constraints",
         ),
         SsaPass::new(
-            |ssa| ssa.fold_constants_with_brillig(options.constant_folding_max_iter),
+            |ssa| ssa.fold_constants(options.constant_folding_max_iter),
```
jfecher, line 262: Is the "Inlining Brillig Calls" string after this accurate? It seems like a secondary effect of constant folding.

compiler/noirc_evaluator/src/ssa/opt/constant_folding/mod.rs
```diff
@@ -57,35 +57,49 @@ const DEFAULT_INTERPRETER_STEP_LIMIT: usize = 10_000_000;
...
+        // Collect all brillig functions so that later we can find them when processing a call instruction
+        let mut brillig_functions: BTreeMap<FunctionId, Function> = BTreeMap::new();
+        for (func_id, func) in &self.functions {
+            if func.runtime().is_brillig() {
+                let cloned_function = Function::clone_with_id(*func_id, func);
+                brillig_functions.insert(*func_id, cloned_function);
+            };
+        }
+        let mut interpreter = Interpreter::new_from_functions(
+            &brillig_functions,
```
jfecher, line 72: Why do we need to deeply clone all brillig functions? Seems wasteful to do this every time we constant fold. Can they be references instead?

compiler/noirc_evaluator/src/ssa/opt/constant_folding/mod.rs
```diff
@@ -94,27 +108,22 @@ impl Ssa {
...
-                std::io::empty(),
-            );
-            // Interpret globals once so that we do not have to repeat this computation on every Brillig call.
-            interpreter.interpret_globals().expect("ICE: Interpreter failed to interpret globals");
-            Some(interpreter)
-        };
+
+        let mut interpreter = Interpreter::new_from_functions(
+            &brillig_functions,
+            InterpreterOptions {
```
jfecher, line 103: Seems like the same setup is repeated here - can you refactor this into a common function? Looks a bit non-trivial

TomAFrench: I looked into this originally but the issue is that the `Interpreter` doesn't own the brillig functions so when we return that back into the calling function the brillig functions go out of scope.

compiler/noirc_evaluator/src/ssa/opt/constant_folding/mod.rs
TomAFrench: The constant folding pass acts on mutable references to all of the functions in SSA so we'd be violating borrow checking rules.

compiler/noirc_evaluator/src/ssa/opt/constant_folding/mod.rs
```diff
@@ -57,76 +57,88 @@ const DEFAULT_INTERPRETER_STEP_LIMIT: usize = 10_000_000;
...
+                no_foreign_calls: true,
+                step_limit: Some(DEFAULT_INTERPRETER_STEP_LIMIT),
+                ..Default::default()
+            },
+            std::io::empty(),
+        );
+        // Interpret globals once so that we do not have to repeat this computation on every Brillig call.
+        interpreter.interpret_globals().expect("ICE: Interpreter failed to interpret globals");
```
vezenovm, line 81: nit: Could probably move this to a helper too as we do it in `fold_constants_using_constraints` as well

PR #9770, opened 2025-09-08
fix: correct max_bit_size when left-shifting

### LINKED ISSUE #9765: ACIR != Brillig: `shl` with overflow on `u64`
### Aim

Replicate nightly failure from https://github.com/noir-lang/noir/actions/runs/17554730948/job/49855802906

The following program fails to execute ACIR but succeeds in Brillig

```rust
fn main(b: u8) -> pub u64 {
    (15877946327378367777 << (b as u64))
}
```

`Prover.toml`:
```toml
b = 10
```

### Expected Behavior

They should agree. 

### Bug

```console
For more information about this error, try `rustc --explain E0602`.
error: Failed constraint
  ┌─ src/main.nr:2:6
  │
2 │     (15877946327378367777 << (b as u64))
  │      ----------------------------------
  │
  = Call stack:
    1. src/main.nr:2:6

```

```console

For more information about this error, try `rustc --explain E0602`.
[fuzz_testing] Circuit output: 7435510297333629952

```

### To Reproduce

```
NOIR_AST_FUZZER_SEED=0x60dbd0af00100000 cargo test -q -p noir_ast_fuzzer_fuzz acir_vs_brillig
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.12 noirc version = 1.0.0-beta.12+55afd6f787eb29daaec0236db80a8279b4df3e89 (git version hash: 55afd6f787eb29daaec0236db80a8279b4df3e89, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9765

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
@@ -148,11 +148,15 @@ impl Context<'_, '_, '_> {
             || {
                 // If we don't know `rhs`'s value then it could be anything up to the number
                 // of bits in the type, e.g. u32 means shifting by up to 32 bits as otherwise we get overflow.
-                let type_max = self.context.dfg.type_of_value(rhs).bit_size();
                 // get_value_max_num_bits might indicate that the underlying type actually has less
                 // bits than the RHS; e.g. because it was upcast from a u1 to a u32.
-                let value_max = 2u32.pow(self.context.dfg.get_value_max_num_bits(rhs));
-                value_max.min(type_max)
+                // The maximum value we can get would be `2^bits - 1`, but `u1` is the only interesting case,
+                // because even for u8 the max value is 255, larger than anything we get based on type.
+                if self.context.dfg.get_value_max_num_bits(rhs) == 1 {
+                    1
+                } else {
+                    self.context.dfg.type_of_value(rhs).bit_size()
+                }
             },
             |rhs_constant| {
                 // Happy case is that we know precisely by how many bits we're shifting by.
@@ -566,6 +570,8 @@ mod tests {
             let ssa = Ssa::from_str(src).unwrap();
             let ssa = ssa.remove_bit_shifts();
 
+            // `max_bit_size` in `truncate` has to be 64, because even though `u8` is just 8 bits,
+            // the maximum value can be 255, which would clearly overflow, as anything over 32 would.
             assert_ssa_snapshot!(ssa, @r#"
             acir(inline) fn main f0 {
               b0(v0: u32, v1: u8):
@@ -622,6 +628,50 @@ mod tests {
                 v60 = truncate v59 to 32 bits, max_bit_size: 64
                 v61 = cast v60 as u32
                 return v61
+            }
+            "#);
+        }
+
+        #[test]
+        fn removes_shl_with_non_constant_rhs_casted_from_u1() {
+            let src = "
+            acir(inline) fn main f0 {
+              b0(v0: u64, v1: u1):
+                v2 = cast v1 as u8
+                v3 = cast v2 as u32
+                v4 = cast v3 as u64
+                v5 = shl v0, v4
+                return v5
+            }
+            ";
+            let ssa = Ssa::from_str(src).unwrap();
+            let ssa = ssa.remove_bit_shifts();
+
+            // What we are casting with can originally be only 1 bit,
+            // so in the truncate we expect to shift with less than
+            // if only considered the 64 bits based on the type.
+            assert_ssa_snapshot!(ssa, @r#"
+            acir(inline) fn main f0 {
+              b0(v0: u64, v1: u1):
+                v2 = cast v1 as u8
+                v3 = cast v2 as u32
+                v4 = cast v3 as u64
+                v5 = cast v3 as u64
+                v7 = lt v5, u64 64
+                constrain v7 == u1 1, "attempt to bit-shift with overflow"
+                v9 = cast v3 as Field
+                v11 = call to_le_bits(v9) -> [u1; 1]
+                v13 = array_get v11, index u32 0 -> u1
+                v14 = not v13
+                v15 = cast v13 as Field
+                v16 = cast v14 as Field
+                v18 = mul Field 2, v15
+                v19 = add v16, v18
+                v20 = cast v0 as Field
+                v21 = mul v20, v19
+                v22 = truncate v21 to 64 bits, max_bit_size: 65
+                v23 = cast v22 as u64
+                return v23
             }
             "#);
         }


compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -573,6 +574,79 @@ mod tests {
...
+                v51 = cast v49 as Field
+                v52 = cast v50 as Field
+                v53 = mul v47, v47
+                v54 = mul v53, v52
+                v55 = mul v53, Field 2
+                v56 = mul v55, v51
+                v57 = add v54, v56
+                v58 = cast v0 as Field
+                v59 = mul v58, v57
+                v60 = truncate v59 to 32 bits, max_bit_size: 64
```
asterite, line 628: The bug was that previously `max_bit_size` was 40 here: 32 + 8. The 32 comes from the left-hand side max bits, while the 8 came (incorrectly) from the rhs bits. Because rhs was casted from u8 and the max value is 256, the max bits for 256 is 8... but this is wrong (there was a similar bug somewhere else in the past).

test_programs/execution_success/shift_right_value_casted_from_smaller_type/src/main.nr
TomAFrench, line 1: This is a shift left.

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
TomAFrench: I'm not clear on why exactly this is wrong though?

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -148,7 +148,7 @@ impl Context<'_, '_, '_> {
             || {
                 // If we don't know `rhs`'s value then it could be anything up to the number
                 // of bits in the type, e.g. u32 means shifting by up to 32 bits as otherwise we get overflow.
```
aakoshh, line 150: This comments suggests there is something to do if we do know the value, but it looks like that's no longer the case?

test_programs/execution_success/shift_right_value_casted_from_smaller_type/src/main.nr
TomAFrench: ohhh it's `(shift)_(right_value_casted_from_smaller_type)` rather than (shift_right)_(value_casted_from_smaller_type).

TomAFrench: can we rename this to be less ambiguous?

asterite: Ah, good point! I just renamed it.

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
asterite: I thought so, but in the end the comment is kind of right but the implementation was wrong. If we don't know what `rhs` is then the number of bits we shift will be at most the value's bit size, so for u32 that's 32 (or, well, it could be up to 2^^32 - 1 but after 32 it's an overflow).

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
asterite: Yeah, to be honest it's a bit confusing...

I'm not sure what `max_bit_size` is used for, but I assume if we do `truncate bit_size: 32, max_bit_size: 40` then for a value with 64 bits the 24 most significant bits will be unchanged?

Now for why 40 is wrong here. If we assume `max_bit_size` is how many non-zero bits will remain after shifting, if we are doing `lhs << rhs` then we'll end up with a value that is `bit_size(lhs) + rhs`. For example if `lhs = 8` the bit size is 3. If we shift it by `rhs = 2` then we'll end up with a bit size of 5 (3 + 2). But note that the formula is `bit_size(lhs) + rhs`, it's not `bit_size(lhs) + bit_size(rhs)`. In the "else" case of the line I changed we use `rhs`, but on the "then" case we incorrectly used `bit_size(rhs)`.

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
aakoshh: Hm, I think I got confused by the comment because I didn't see that this code was the `default` case of `map_or_else`, so "If we don't know `rhs`'s value then" is exactly what this code handles, and the `|rhs_constant|` lambda is for when we know it. 

What I don't understand is why it was wrong. If the value was not a constant, then `get_value_max_num_bits` had special handling for `Instruction` by looking up whether it was a `Cast` from a lower bit size value, in which case it might have returned less than `type_of_value(value).bit_size()`, why was that incorrect?

asterite: Yes, this is super confusing. Every time I want to explain it I need to think it again 😅 

Let's take the new test program as an example. We are left-shifting `some_u64 << (some_u8 as u64)`.

First, let's say it's actually different and want we know exactly by how much we are shifting, so it's `some_u64 << 3`. The final number of bits we'll need for the result is 64 + 3 = 67, where 64 comes from lhs and 3 comes from the constant 3.

Now, if we don't know the constant value of rhs and we have `some_u64 << another_u64`  we end up needing 64 + 64 bits: 64 from lhs and 64 max from rhs (well, `another_u64` could have a value like 1234, but that would overflow so at most we can shift by 64).

Finally, if we don't know the constant value of rhs and we have `some_u64 << (some_u8 as u64)`, what's the maximum bit size we can get? The previous math was doing this: we have a `some_u8` casted as a `u64`. The maximum bit size of a `u8` value is 8, so the maximum bit size is 64 + 8. But if we know the value of `some_u8`, for example we know it's 63, then the result is `some_u64 << 63` so we need 64 + 63 bits for the result... so that 8 there is definitely not correct.

The new way to compute it is to say that the type of rhs is u64, so the maximum we can shift by is 64 (past that it's an overflow).

The comment above this line says that too: if we are shifting by a u32, the maximum we can shift by is 32 regardless of whether the value was upcasted from a smaller value. Well, I guess there's an exception: if we upcast from a u1 then we can only be shifting by 0 or 1, and that case is not handled in this new code... we could try it as a separate optimization though I don't know how frequently that will happen.

aakoshh: Ah, yes, I don't know why it's so confusing either, I just "got it" 10 minutes ago, then wanted to comment and by that time I lost it again 😖 

Trying out the integration tests makes it easy to grasp though: we have `(15877946327378367777 << (b as u64))` and we know that the dynamic `b` will have a value of 10, and yet we end up with `v69 = truncate v68 to 64 bits, max_bit_size: 72`, because it comes from a `u8`. 72-64=8 is not the right value; we know it's more than that!

Okay, so the maximum can be `u8::MAX`, but we have a constraint to limit this to 64 already. I'm not sure if `truncate` would accept such a big `max_bit_size`, as I just saw it inspecting it and say "wait, that doesn't even fit into a Field", and that's just u8. 

Because we know we have a constraint, we can use `max` though.

aakoshh: For the recursive case, could we try this?

```rust
let max_bit_shift_size = self.context.dfg.get_numeric_constant(rhs).map_or_else(
            || {
                let type_max = self.context.dfg.type_of_value(rhs).bit_size();
                let value_max = 2.pow(self.context.dfg.get_value_max_num_bits(rhs));
                value_max.min(type_max)
            },
            |rhs_constant| {
                rhs_constant.to_u128() as u32
            },
        );
```

aakoshh: I tried it in https://github.com/noir-lang/noir/pull/9965 and it's okay in principle. I'll try it on this PR and push to it if that's okay.

aakoshh: hm, maybe this isn't the right formula, and we need a separate function that only works on types.

aakoshh: I pushed something that adds handling for when `get_value_max_num_bits` returns 1, because it originates from a `u1`, which it now recursively detects. 

The formula above was wrong, it had to be `2u32.pow(self.context.dfg.get_value_max_num_bits(rhs)) - 1`, but it's easy to overflow and can only beat the type based value for `u1`, so it's easier to just make that the exceptional case.

aakoshh: Hah, I ended up basically with the same thing as https://github.com/noir-lang/noir/blob/04a79de8777e90aab251a298f0b5cf5f36a63a74/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs#L303-L307

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -148,7 +148,11 @@ impl Context<'_, '_, '_> {
...
-                self.context.dfg.type_of_value(rhs).bit_size()
+                let type_max = self.context.dfg.type_of_value(rhs).bit_size();
+                // get_value_max_num_bits might indicate that the underlying type actually has less
+                // bits than the RHS; e.g. because it was upcast from a u1 to a u32.
+                let value_max = 2u32.pow(self.context.dfg.get_value_max_num_bits(rhs));
```
asterite: Would this overflow for u128, etc.?

aakoshh: I only checked `u32` because that's what the result needs to be, and e.g. `2u32^32` does overflow, so it's only worth trying if the value is not greater than `type_max.ilog2()`, but since we know that it will only ever beat it with `u1`, I scrapped this idea.

TomAFrench (approved):

compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
asterite: Aaaah... I started looking commit by commit, had to take my kid to school so I didn't see there were more commits. This looks good 👌

PR #9772, opened 2025-09-08
feat(ownership): Do not clone nested Index expressions

vezenovm:
# Description

## Problem\*

No issue just something I noticed while looking at the necessity of the RC tracker in DIE now that we have more robust ownership semantics. 

The following program:
```noir
    let EXPONENTIATE: [[Field; 2]; 2] = [[1, 1], [0, 0]];

    let mut acc: Field = 0;
    for i in 0..2 {
        for j in 0..2 {
            acc += EXPONENTIATE[i][j];
        }
    }
    assert(acc != 0);
    assert(x != y);
}
```
Will produce the follow monomorphized result:
```
unconstrained fn main$f0(x$l0: Field, y$l1: pub Field) -> () {
    let EXPONENTIATE$l2 = [[1, 1], [0, 0]];
    let mut acc$l3 = 0;
    for i$l4 in 0 .. 2 {
        for j$l5 in 0 .. 2 {
            acc$l3 = (acc$l3 + EXPONENTIATE$l2[i$l4].clone()[j$l5])
        }
    };
    assert((acc$l3 != 0));;
    assert((x$l0 != y$l1));
}
```

## Summary\*

The nested access to `EXPONENTIATE` should not need to be cloned as we are simply borrowing the expression. 

## Additional Context

This cloning was the main cause of regressions after removing the RC tracker. Hopefully after this update we may be able to remove the RC tracker entirely. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_frontend/src/ownership/mod.rs
+++ b/compiler/noirc_frontend/src/ownership/mod.rs
@@ -153,7 +153,7 @@ impl Context {
                 self.handle_reference_expression(&mut index.collection);
                 self.handle_expression(&mut index.index);
             }
-            
+
             // If we have something like `f(arg)` then we want to treat those variables normally
             // rather than avoid cloning them. So we shouldn't recur in `handle_reference_expression`.
             other => self.handle_expression(other),


compiler/noirc_frontend/src/ownership/mod.rs
```diff
@@ -149,6 +149,11 @@ impl Context {
             }
             Expression::ExtractTupleField(tuple, _index) => self.handle_reference_expression(tuple),
 
+            Expression::Index(index) => {
```
jfecher, line 152: I swear we used to have this same snippet but it was removed due to a bug in aztec-packages, but I can't seem to find it in the git history (and if there was a bug fixed we evidently don't have a regression for it). So maybe there was no bug.

TomAFrench: Let's mirror this PR in aztec-packages to check in that case.

TomAFrench: CI's passing in https://github.com/AztecProtocol/aztec-packages/pull/16877 so we're good to go.

PR #9773, opened 2025-09-08
chore: encapsulate constant folding logic better

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR breaks up the constant folding pass into a couple of submodules so that we can place the different caches and their associated logic into separate files to build a proper interface between them and the rest of the pass.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/constant_folding/interpret.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/constant_folding/interpret.rs
@@ -14,15 +14,15 @@ use crate::ssa::{
     },
 };
 
-/// Checks if the given instruction is a call to a brillig function with all constant arguments.
+/// Checks if the given instruction is a call to a function with all constant arguments.
 /// If so, we can try to evaluate that function and replace the results with the evaluation results.
-pub(super) fn try_interpret_brillig_call(
+pub(super) fn try_interpret_call(
     instruction: &Instruction,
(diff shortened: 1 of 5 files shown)


compiler/noirc_evaluator/src/ssa/opt/constant_folding/interpret.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// This can occur in the situation where the brillig function reaches a "trap" or a foreign call opcode.
+    CannotEvaluate,
+    /// The instruction was a call to a brillig function and we were able to evaluate it,
+    /// returning [SSA interpreter][Interpreter] [values][InterpreterValue].
+    Evaluated(Vec<InterpreterValue>),
+}
+
+/// Tries to evaluate an instruction if it's a call that points to a brillig function,
+/// and all its arguments are constant.
+/// We do this by directly executing the function with a brillig VM.
```
asterite: This comment is outdated.

asterite: I thought about replacing just the last line in the comment (we don't use a brillig VM), but in this function we don't check if the function is brillig.

What's happening is that the interpreter is built with a list of all the brillig functions... I guess this could be clarified in this comment too.

asterite (approved):

PR #9776, opened 2025-09-09
fix(ssa): Replace failing array access with constraint and defaults

### LINKED ISSUE #9764: Compiler crash: Load after DIE in ACIR
### Aim

Replicate nightly failure from https://github.com/noir-lang/noir/actions/runs/17554730948/job/49855802906

The following program fails to compile:
```rust
fn main(a: Field) {
    let _ = if a == 1 {
        0
    } else {
        let mut d = [];
        assert((0 < (*d[0])));
        a
    };
}
```

### Expected Behavior

It would be nice if it didn't crash.

### Bug

```console
❯ cargo run -q -p nargo_cli -- compile --force --silence-warnings 

Message:  not expected to have Load or Store instruction after DIE in an ACIR function: main f0 / b0 / 5: Load { address: Id(6) }
Location: compiler/noirc_evaluator/src/ssa/opt/die.rs:968
```

We know that `d` is an empty array, so for example with `--force-brillig` the `remove_unreachable_instructions` pass makes that block terminate with `unreachable`, and it will fail with _Index out of bounds_, however in ACIR due to flattening that's not happening:

```
❯ cargo run -q -p nargo_cli -- compile --force --silence-warnings --show-ssa-pass "step 34" --show-ssa-pass "step 35"

After Mem2Reg (5) (step 34):
  b0(v0: Field):
    v2 = eq v0, Field 1
    v3 = not v2
    v4 = make_array [] : [&mut u32; 0]
    enable_side_effects v3
    v6 = array_get v4, index u32 0 -> &mut u32
    v7 = load v6 -> u32
    v8 = lt u32 0, v7
    v9 = unchecked_mul v8, v3
    constrain v9 == v3
    enable_side_effects u1 1
    return
}

After Remove Unreachable Instructions (1) (step 35):
  b0(v0: Field):
    v2 = eq v0, Field 1
    v3 = not v2
    v4 = make_array [] : [&mut u32; 0]
    enable_side_effects v3
    v6 = array_get v4, index u32 0 -> &mut u32
    v7 = load v6 -> u32
    v8 = lt u32 0, v7
    v9 = unchecked_mul v8, v3
    constrain v9 == v3
    enable_side_effects u1 1
    return
}
...
```

Even though we know that `v6 = array_get v4, index u32 0 -> &mut u32` will fail we don't insert a constraint, because the predicate is not constant 1. For `Binary` instructions, however, we do insert a _guarded_ constraint that only fails if the current side effect variable is `true`. 

In both cases, we set the reachability of the rest of the block to `UnreachableUnderPredicate`, however this has no effect on the `load` because it only replaces instructions with their default result where `requires_acir_gen_predicate` is `true`, which for `load` is not the case.

Maybe `remove_unreachable_instructions` could make an exception for `load` and replace it with its default once we know it will never be reached? Alternatively, we could replace the `array_get` with a conditionally failing constraint and the default value value for its result, because the `zeroed_value` for references now involves inserting a `store` as well.

### To Reproduce

```
NOIR_AST_FUZZER_SHOW_AST=1 NOIR_AST_FUZZER_SEED=0xd75e62fd00100000 cargo test -q -p noir_ast_fuzzer_fuzz orig_vs_morph
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.12 noirc version = 1.0.0-beta.12+55afd6f787eb29daaec0236db80a8279b4df3e89 (git version hash: 55afd6f787eb29daaec0236db80a8279b4df3e89, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9764 

## Summary\*

Changes the way `remove_unreachable_instructions` handles an always-fail `array_get` from the current way of creating a dummy array and getting from that, to inserting an "Index out of bounds" constraint which will fail if enabled, and replacing the results of the array operation with default values (which, if the array contained references, make sure the returned reference is allocated and initialised to a default value). This is what we do for certain slice operations as well.

Moves the `remove_unreachable_instructions` in the SSA pipeline to appear _before_ `mem2reg`, so that if the above mechanism inserts a `store` of a default, then `mem2reg` can use that information to remove the subsequent `load`. 

## Additional Context

Up to now `remove_unreachable_instructions` was done just after `mem2reg` to remove any unused `Load`, but then it turned out that this wasn't as useful as letting them survive to the end, so that the `verify_no_dynamic_indices_to_references` pass can return better error messages. There is another `remove_unreachable_instructions` pass after `mem2reg` anyway, but if we do `mem2reg -> remove_unreachable_instructions -> dead_instruction_elimination`, then DIE removes the `Store`, leaving the `Load` hanging there, which is where this ticket started.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/mod.rs
```diff
@@ -169,14 +169,15 @@ pub fn primary_passes(options: &SsaEvaluatorOptions) -> Vec<SsaPass<'_>> {
...
         // yet-to-be-visited predecessor we forget known values; less blocks mean less unification.
         SsaPass::new(Ssa::simplify_cfg, "Simplifying"),
-        // We cannot run mem2reg after DIE, because it removes Store instructions.
-        // We have to run it before, to give it a chance to turn Store+Load into known values.
-        SsaPass::new(Ssa::mem2reg, "Mem2Reg"),
-        // Removing unreachable instructions before DIE, so it gets rid of loads that mem2reg couldn't,
-        // if they are unreachable and would cause the DIE post-checks to fail.
+        // Removing unreachable instructions before mem2reg, which may result in some default Store
+        // instructions being added, which it can pair up with Loads. If we ran it after it,
+        // then DIE would just remove the Stores, leaving the Loads dangling.
```
vezenovm, line 174: By dangling loads do you mean unused? I would expect these to be removed in DIE so this should be fine. Did you see some regressions caused by having these passes switched?

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -303,47 +303,26 @@ impl Function {
...
+                        );
 
+                        current_block_reachability = if always_fail {
+                            // If the block fails unconditionally, we don't even need a default for the results.
+                            context.remove_current_instruction();
                             Reachability::Unreachable
                         } else {
+                            // We will never use the results (the constraint fails if the side effects are enabled),
+                            // but we need them to make the rest of the SSA valid even if the side effects are off.
+                            remove_and_replace_with_defaults(context, func_id, block_id);
```
vezenovm, line 330: I'm surprised we do not have to replace with defaults here as previously I was getting failures with array containing references when `is_predicate_constant_one` returns true. Is this related to the re-ordering of the passes above? Would this break if we switched to not have a `mem2reg` after this pass? Could we just always remove and replace with defaults?

compiler/noirc_evaluator/src/ssa/mod.rs
aakoshh: By dangling I mean there is a `Load` from a reference which doesn't ever get `Store`d to.

So DIE only removes the `Store`, but leaves the `Load` there. I can't remember why now, but that's how it works. If all is well there should not be any `Load` left, which is why we have the `dead_instruction_elimination_post_check` assert this and panic if there is.

aakoshh: If I leave these switched, then what happens is that the new logic added in this PR replaces this:
```
v1 <- make_array [] -> [&mut u8; 0]
v2 <- array_get v1, index u32 0 -> &mut u8
v3 <- load v2 -> u8
```
with this:
```
v1 <- make_array [] -> [&mut u8; 0]
v2 <- allocate -> &mut u8
store u8 0 at v2
v3 <- load v2 -> u8
```
but then the DIE removes the store:
```
v1 <- make_array [] -> [&mut u8; 0]
v2 <- allocate -> &mut u8
v3 <- load v2 -> u8
```

One would think that hey, `load v2` uses `v2`, so why are we getting rid of the `store` to it. But then I read the comments that `mem2reg` is supposed to have already replaced the `load v2` with the value in the last `store`, so `store` should not be required any more, and there should naturally be no `load` left either, but DIE's job is limited to `Store` apparently.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
aakoshh: Do you mean you are surprised we don't have to replace with defaults when `is_predicate_constant_one` is `true`, only when it's a variable? 

When the predicate is `true`, then we make the rest of the block `unreachable`, so we don't need the results, as nothing will ever use them. But that's generally not the case after flattening, only in Brillig where blocks keep reflecting control flow. 

So far when the array contained references and the side effects was `true`, it did something with a dummy array to fill it up, but I'm not exactly sure how that behaved. Now, when the result is a reference, we make sure that it's not just allocated but we stored a default value to it, so a subsequent load does not fail. And that's why we need `mem2reg` to run after this, to eliminate those store+load pairs.

aakoshh:
> It would be good to match up the Brillig functionality here as well, but that can be done in a follow-up just leaving it as a note.

I wasn't sure why this logic only applied to ACIR actually.

compiler/noirc_evaluator/src/ssa/mod.rs
aakoshh: Clarification: DIE removes `Store` if the CFG has been flattened, and also removes `Load` _iff_ it's unused, which we assume it will be if `mem2reg` was able to do its job. In our case the result of the loads are used, so it stays, but the store goes. That's why we need to let `mem2reg` have a go, or replace loads with defaults as well, but so far I only replaced with defaults those which have no side effects (or don't require a predicate).

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
aakoshh: One potential reason you might have got problems with the dummy array before is that the `zeroed_value` it fills the array with only got enhanced to `store` default values to references since https://github.com/noir-lang/noir/pull/9603 

I assume before that it only allocated the reference, so if a load was attempted from it, it still failed.

compiler/noirc_evaluator/src/ssa/mod.rs
vezenovm: Ok got it. Can we add needing mem2reg to be run after this pass as a post condition? This can just be a module comment.

vezenovm: > replace loads with defaults as well, but so far I only replaced with defaults those which have no side effects (or don't require a predicate).

If we know we always have a preceding failing instruction we should be able to always replace with defaults no? This can be experimented with in a follow-up though.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
vezenovm: > One potential reason you might have got problems with the dummy array before is that the `zeroed_value` it fills the array with only got enhanced to `store` default values to references since #9603

Ah this is most likely the reason.

vezenovm (approved):

compiler/noirc_evaluator/src/ssa/mod.rs
aakoshh: Yes, I think it's safe to always replace with defaults. At the time I thought nothing else should fail, but empty arrays are still problematic. I'll make a follow up.

aakoshh: https://github.com/noir-lang/noir/issues/9783

Updated the docs too.

PR #9782, opened 2025-09-09
fix(mem2reg): handle instruction simplified to multiple

### LINKED ISSUE #9758: Full != Min: `&mut bool` array loses store
### Aim

Capture nightly failure from https://github.com/noir-lang/noir/actions/runs/17554730948/job/49855802906

The following program fails with the default pipeline but passes with `--minimal-ssa`:

```rust
    let e = func_5(&[(&mut false), (&mut false)]);
    let e = e.push_back((&mut false));
    let e = func_5(e);
    !(*e[0])
}
unconstrained fn func_5(a: [&mut bool]) -> [&mut bool] {
    if (*a[1]) {
        for idx_d in 0_u128..2_u128 {
            assert((if (1_u128 <= idx_d) { idx_d } else { idx_d } <= 0_u128), "FTN");
        }
        a
    } else {
        a
    }
}
```

### Expected Behavior

It should return `true`. 

### Bug

```console
$ cargo run -q -p nargo_cli -- execute --force --silence-warnings
...
error: Assertion failed: 'Bit size for value 254 does not match the expected bit size 1'
  ┌─ src/main.nr:5:8
  │
5 │     !(*e[0])
  │        ----
  │
  = Call stack:
    1. src/main.nr:5:8

Failed assertion
```

```console
$ cargo run -q -p nargo_cli -- interpret --force --silence-warnings
...
--- Interpreter result after Simplifying (step 17):
Ok(u1 1)
---
--- Interpreter result after Mem2Reg (step 18):
Err(Reference value `*v82 = None` loaded before it was first stored to)
...
```

### To Reproduce

```shell
NOIR_AST_FUZZER_SEED=0xa61a5d9400100000 cargo test -q -p noir_ast_fuzzer_fuzz min_vs_full
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.12 noirc version = 1.0.0-beta.12+6ab715a1dc9cc0bb77558e7f5cd15c998d1f68f7 (git version hash: 6ab715a1dc9cc0bb77558e7f5cd15c998d1f68f7, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9758

## Summary

The explanation of the fix is in the code comments.

In addition to this, I added handling of `SimplifiedToMultiple`. We analyze instructions that result from simplifying an instruction and `slice_push_back` might be simplified to a `make_array` and we weren't analyzing that. I _think_ we never end up with multiple instructions but I preferred to be correct here.

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -410,28 +410,63 @@ impl<'f> PerFunctionContext<'f> {
...
+                InsertInstructionResult::InstructionRemoved => return,
             }
         };
 
+        // A simplified call might be SlicePushBack or SlicePushFront.
+        // In that case we still need to analyze the original call arguments as we don't
+        // want to lose stores to references passed into the call.
+        if simplified && is_call {
+            self.analyze_instruction_without_simplifying(references, instruction_id, false);
+        }
```
jfecher: I'm confused why we need this in addition to the loop below.

If the instruction was simplified out, why do we need it at all? In the case of SlicePushBack that'd result in a constant make_array which we should be able to analyze instead to see that it contains the reference in question.

asterite: It's a good point. It's what I first tried but for some reason that doesn't fix the issue.

When we analyze a `Call` we put all arguments in `instruction_input_references` and also mark all references as unknown. When we analyze a `MakeArray` we just relate the result to the array elements. But in the final code the results of `make_array` are not used. However, I'm seeing that `v3` is used (we load from it) so I don't know why the store is removed... I'll take a look.

asterite: Aaah... you were right. When I didn't have that extra code I saw the `store`s were gone, but I didn't realize the return value was simplified by what was stored in `v3` (`u1 0`).

In the end that test was wrong. I started from a bigger test and tried to reduce it but it ended up not failing without the fix.

I replaced the test with a new one, and also removed this extra code.

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
```diff
@@ -410,28 +410,55 @@ impl<'f> PerFunctionContext<'f> {
...
+        let (instruction, loc) = self.inserter.map_instruction(instruction_id);
+
+        let (instructions, simplified) = {
+            match self.inserter.push_instruction_value(instruction, instruction_id, block_id, loc) {
+                InsertInstructionResult::Results(id, _) => (vec![id], false),
                 InsertInstructionResult::SimplifiedTo(value) => {
                     let value = &self.inserter.function.dfg[value];
-                    let Value::Instruction { instruction, .. } = value else {
+                    if let Value::Instruction { instruction, .. } = value {
+                        (vec![*instruction], true)
```
jfecher: Nit: it'd be nice if we could avoid allocating a vec for the common case of handling 1 instruction

jfecher (approved):

compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs
asterite: I'll push a commit for this tomorrow. I guess we can call the helper function directly from each match case.

asterite: Done!

PR #9789, opened 2025-09-09
chore(ownership): Add tests for cloning nested arrays returned from indexing

### LINKED ISSUE #9785: Ownership: Can avoid cloning array access in function call args
### Problem

Look at `execution_success/regression_9206`:
```noir
global G_S: [[u128; 0]] = [[], []];
fn main(i: u32) -> pub u32 {
    let _a = unsafe { func_3(G_A[i])[1] };
    let _s = unsafe { func_3(G_S[i])[1] };
    i
}
unconstrained fn func_3(_a: [u128; 0]) -> [[u128; 0]; 2] {
    G_A
}
```
This gives us the following result with `nargo info --force-brillig --show-monomorphized`:
```
global G_A$g0: [[u128; 0]; 2] = [[], []];
global G_S$g1: [[u128; 0]] = &[[], []];
unconstrained fn main$f0(i$l0: u32) -> pub u32 {
    let _a$l1 = {
        func_3$f1(G_A$g0[i$l0].clone())[1].clone()
    };
    let _s$l2 = {
        func_3$f1(G_S$g1[i$l0].clone())[1].clone()
    };
    i$l0
}
unconstrained fn func_3$f1(_a$l3: [u128; 0]) -> [[u128; 0]; 2] {
    G_A$g0.clone()
}
```
The clones of `G_A[i]` and `G_S[i]` are entirely unnecessary. 

### Happy Case

We should be able to prevent cloning array indexing in a call argument and the indexing of a call's result. 

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #9785 

## Summary\*

This PR adds some tests to guarantee that we are appropriately cloning nested arrays.
 
After no longer cloning nested arrays returned from indexing, `regression_9206` has the following MAST:
```
global G_A$g0: [[u128; 0]; 2] = [[], []];
global G_S$g1: [[u128; 0]] = &[[], []];
unconstrained fn main$f0(i$l0: u32) -> pub u32 {
    let _a$l1 = {
        func_3$f1(G_A$g0[i$l0])[1].clone()
    };
    let _s$l2 = {
        func_3$f1(G_S$g1[i$l0])[1].clone()
    };
    i$l0
}
unconstrained fn func_3$f1(_a$l3: [u128; 0]) -> [[u128; 0]; 2] {
    G_A$g0.clone()
}
```
You will note that we do not clone the index argument. This caused no failures in our test suite which is unexpected. 

I have added a couple tests that fail our test suite if we were to not clone nested arrays. 

## Documentation\*

Check one:

# PR Checklist\*


vezenovm:
Hmm realizing this may actually not be valid for cases where we pass a nested array to a call and then use that nested array afterwards. Interesting that none of our tests fail with this change. I will try to create a case that fails here.

jfecher:
> Hmm realizing this may actually be not be valid for cases when we pass a nested array to a call and then use that nested array afterwards. Interesting that none of our tests fail with this change. I will try to create a case that fails here.

Yep, was just about to mention that is why we clone in this position in general

vezenovm:
> Interesting that none of our tests fail with this change. I will try to create a case that fails here.

This PR instead adds tests that would fail if we avoided cloning nested arrays returned from indexing. I also added some tests to ownership.

jfecher (approved):

PR #9791, opened 2025-09-09
feat(ownership): Do not clone indexed call results containing arrays

### LINKED ISSUE #9787: Ownership: Can avoid cloning the array result of a function call
### Problem

Similar to https://github.com/noir-lang/noir/issues/9785 we look at the same snippet:

Look at `execution_success/regression_9206`:
```noir
global G_S: [[u128; 0]] = [[], []];
fn main(i: u32) -> pub u32 {
    let _a = unsafe { func_3(G_A[i])[1] };
    let _s = unsafe { func_3(G_S[i])[1] };
    i
}
unconstrained fn func_3(_a: [u128; 0]) -> [[u128; 0]; 2] {
    G_A
}
```
This gives us the following result with `nargo info --force-brillig --show-monomorphized`:
```
global G_A$g0: [[u128; 0]; 2] = [[], []];
global G_S$g1: [[u128; 0]] = &[[], []];
unconstrained fn main$f0(i$l0: u32) -> pub u32 {
    let _a$l1 = {
        func_3$f1(G_A$g0[i$l0].clone())[1].clone()
    };
    let _s$l2 = {
        func_3$f1(G_S$g1[i$l0].clone())[1].clone()
    };
    i$l0
}
unconstrained fn func_3$f1(_a$l3: [u128; 0]) -> [[u128; 0]; 2] {
    G_A$g0.clone()
}
```
The clones to `func_3(G_A[i])[1]` and `func_3(G_S[i])[1]` are unnecessary. 

### Happy Case

We should not clone the result of a function call. 

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #9787 

## Summary\*

Reference the issue for the monomorphized AST (MAST) output we see on master.

This is the new MAST output for regression_9206:
```
global G_A$g0: [[u128; 0]; 2] = [[], []];
global G_S$g1: [[u128; 0]] = &[[], []];
unconstrained fn main$f0(i$l0: u32) -> pub u32 {
    let _a$l1 = {
        func_3$f1(G_A$g0[i$l0].clone())[1]
    };
    let _s$l2 = {
        func_3$f1(G_S$g1[i$l0].clone())[1]
    };
    i$l0
}
unconstrained fn func_3$f1(_a$l3: [u128; 0]) -> [[u128; 0]; 2] {
    G_A$g0.clone()
}
```
When combined with #9789 we will now have no clones in `main`. 

## Additional Context

I don't expect many bytecode changes as we still have the RC tracker during DIE which should be catching that we never mutate any arrays in this program and remove the IncrementRc instructions.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_frontend/src/ownership/mod.rs
```diff
@@ -280,7 +280,10 @@ impl Context {
         self.handle_reference_expression(&mut index.collection);
         self.handle_expression(&mut index.index);
 
-        if contains_array_or_str_type(&index.element_type) {
+        // We do not need to clone the result of a function call.
+        // The appropriate cloning should be handled by the function internally or the caller of the function.
+        let is_call = matches!(*index.collection, Expression::Call(_));
+        if !is_call && contains_array_or_str_type(&index.element_type) {
```
jfecher: We should avoid specific checks for a certain expression variant with possible to keep this more general.

It looks like what we're looking for is:
- If `index.collection` was referred to by reference, we may need to clone our resulting index
- otherwise, if it was moved, we don't need to clone the resulting index.

So maybe `handle_reference_expression` should return a boolean indicating whether it moved or not? We could probably forward this to the nested index case to ensure we compile `nested[0][1][2]` as `nested[0][1][2].clone()` without intermediate clones, and `foo()[0][1][2]` as `foo()[0][1][2]` rather than `foo()[0][1].clone()[2].clone()` which it looks like we'd get currently.

jfecher (changes requested):

compiler/noirc_frontend/src/ownership/mod.rs
vezenovm: Will update to avoid the specific check. But I do think this is fine for the nested index case.

On master, for this program:
```noir
    let EXPONENTIATE: [[[Field; 2]; 2]; 2] = [[[1, 1], [0, 0]], [[1, 1], [0, 0]]];
    let mut acc: Field = 0;
    for i in 0..2 {
        for j in 0..2 {
            acc += EXPONENTIATE[i][j][i];
        }
    }
    assert(acc != 0);
    assert(x != y);
}
```
We get the expected MAST:
```
unconstrained fn main$f0(x$l0: Field, y$l1: pub Field) -> () {
    let EXPONENTIATE$l2 = [[[1, 1], [0, 0]], [[1, 1], [0, 0]]];
    let mut acc$l3 = 0;
    for i$l4 in 0 .. 2 {
        for j$l5 in 0 .. 2 {
            acc$l3 = (acc$l3 + EXPONENTIATE$l2[i$l4][j$l5][i$l4])
        }
    };
    assert((acc$l3 != 0));;
    assert((x$l0 != y$l1));
}
```
And in the case of `foo()[0][1][2]` we already give `foo()[0][1][2].clone()`. This is using the `regression_9206` example from this PR.

vezenovm: For clarity, `foo()[0][1][2]` turning into `foo()[0].clone()[1].clone()[2].clone()` is what occurs on nightly which does not yet include https://github.com/noir-lang/noir/pull/9772

compiler/noirc_frontend/src/ownership/mod.rs
```diff
@@ -131,32 +131,44 @@ impl Context {
     ///
     /// Note that this also matches on dereference operations to exempt their LHS from clones,
     /// but their LHS is always exempt from clones so this is unchanged.
-    fn handle_reference_expression(&mut self, expr: &mut Expression) {
+    ///
+    /// # Returns
+    /// A boolean representing whether or not the expression was borrowed by reference or moved.
+    /// This helps the caller determine whether we need to clone the result of a borrowed expression.
+    /// `true` represents an expression that was moved
+    /// `false` represents an expression borrowed by reference
```
jfecher:
```suggestion
    /// A boolean representing whether or not the expression was borrowed by reference (false) or moved (true).
```

PR #9803, opened 2025-09-10
feat(fuzz): Allow index OOB with a small probability

aakoshh:
# Description

## Problem\*

Follow up for https://github.com/noir-lang/noir/pull/9786

## Summary\*

With a 1/10 chance leave out the modulo operation that the fuzzer uses to make sure that an arbitrary index will fall into the bounds of an array or slice. This will most certainly result in "Index out of bounds", and the goal is to make sure all backends handle it the same way. 

It is only used if `allow_overflow` is `true`, which we randomise in fuzz targets so that half the time we don't fall into trivial errors, but give the circuit a chance to complete.

## Additional Context

TODO: Some bug came out because of this change, trying to minimise now.

## Documentation\*

Check one:

# PR Checklist\*


aakoshh:
Thanks for pointing that out @rkarabut , added a new flag and a `default_config` function to make it easier to control this initialization.

PR #9806, opened 2025-09-10
fix: do not simplify constraints with induction variable

### LINKED ISSUE #9727: Comptime allows side effects in blocks where an assertion is guaranteed to fail
### Aim

The following program:

```
fn main() {
    comptime {
        for i in 1..3 {
            assert((i < 2), "fail");
            println(i);
        }
    }
}
```

prints `1` and then errors out with `error: fail`. Removing `comptime` though leads to the following output:

```
bug: Assertion is always false: fail
  ┌─ src/main.nr:4:21
  │
4 │             assert((i < 2), "fail");
  │                     ----- As a result, the compiled circuit is ensured to fail. Other assertions may also fail during execution
  │
  = Call stack:
    1. src/main.nr:4:21

error: Assertion failed: fail
  ┌─ src/main.nr:4:21
  │
4 │             assert((i < 2), "fail");
  │                     -----
  │
  = Call stack:
    1. src/main.nr:4:21

Failed assertion
```

As a result, nothing is printed with `println`.

### Expected Behavior

Comptime should probably fail in a similar way to ACIR, disallowing side effects in case the block is guaranteed to fail.

### Bug

With `comptime`:
```
$ nargo execute --force
1
error: fail
  ┌─ src/main.nr:4:21
  │
4 │             assert((i < 2), "fail");
  │                     ----- Assertion failed
  │

```

Without `comptime`:
```
bug: Assertion is always false: fail
  ┌─ src/main.nr:4:21
  │
4 │             assert((i < 2), "fail");
  │                     ----- As a result, the compiled circuit is ensured to fail. Other assertions may also fail during execution
  │
  = Call stack:
    1. src/main.nr:4:21

error: Assertion failed: fail
  ┌─ src/main.nr:4:21
  │
4 │             assert((i < 2), "fail");
  │                     -----
  │
  = Call stack:
    1. src/main.nr:4:21

Failed assertion
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #9727

## Summary\*
Simplifying constraints in a loop can lead to side-effect code to not be executed. I removed the simplification which uses induction bounds.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
@@ -2761,7 +2761,7 @@ mod control_dependence {
 
     #[test]
     fn simplify_constraint() {
-        // This test shows the simplification of the constraint constrain v17 == u1 1 which is converted into constrain u1 0 == u1 1 in b5
+        // This test shows the constraint constrain v17 == u1 1 is not simplified into constrain u1 0 == u1 1
         let src = "
         brillig(inline) fn main f0 {
           entry(v0: u32, v1: u32, v2: u32):
@@ -2793,11 +2793,10 @@ mod control_dependence {
         ";
 
         let ssa = Ssa::from_str(src).unwrap();
-
         let ssa = ssa.loop_invariant_code_motion();
-        // The loop is guaranteed to fully execute, so we expect the constrain to be simplified into constrain u1 0 == u1 1
-        // However, even though the constrain is not a loop invariant we expect it to remain in place
-        // as it is control dependent upon its predecessor blocks which are not pure.
+
+        // Despite the loop is guaranteed to fully execute, which implies that the constrain will fail at some iteration,
+        // the constraint is not simplified in case some side-effect instruction would run in the previous iterations.
         assert_ssa_snapshot!(ssa, @r"
         brillig(inline) fn main f0 {
           b0(v0: u32, v1: u32, v2: u32):
@@ -2821,9 +2820,9 @@ mod control_dependence {
             jmp b5()
           b5():
             v15 = lt v3, u32 4
-            constrain u1 0 == u1 1
-            v17 = unchecked_add v3, u32 1
-            jmp b1(v17)
+            constrain v15 == u1 1
+            v16 = unchecked_add v3, u32 1
+            jmp b1(v16)
         }
         ");
     }


vezenovm (approved):
I guess we could check whether there are side effectual instructions which follow this constrain. This is what we do for hoisting constrains just in the reverse https://github.com/noir-lang/noir/blob/574550bbc3640b78545985e013706ffcfe637e8e/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs#L457

However, I think we should just get this fix in as it doesn't look to have much of an effect on our benchmarks and it simplifies this pass before audit. Could we capture an issue with this possible improvement though?

guipublic:
> I guess we could check whether there are side effectual instructions which follow this constrain. This is what we do for hoisting constrains just in the reverse
> 
> https://github.com/noir-lang/noir/blob/574550bbc3640b78545985e013706ffcfe637e8e/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs#L457
> 
> However, I think we should just get this fix in as it doesn't look to have much of an effect on our benchmarks and it simplifies this pass before audit. Could we capture an issue with this possible improvement though?

Issue #9820

asterite:
@guipublic I think this fixed the wrong issue

vezenovm:
> @guipublic I think this fixed the wrong issue

Which issue? This looks like the correct fix to the linked issue to me?

asterite:
Aaaah... you are right. I thought the fix was going to be to not print stuff during comptime, but the actual fix was to print stuff in ACIR even on failure? I'm not sure I understand how it got fixed if it's just changing how loops work.

vezenovm:
> Aaaah... you are right. I thought the fix was going to be to not print stuff during comptime, but the actual fix was to print stuff in ACIR even on failure? I'm not sure I understand how it got fixed if it's just changing how loops work.

We were previously hoisting a constrain that we always knew would fail outside of the loop. So yes we want to print on failure as otherwise we will potentially change the ordering of side effects. If side effectual instructions come after that hoisted constrain they will never be executed. Imagine instead of printing the foreign function was instead modifying some database. Doing this simplification without checking the purity of the block could lead to soundness errors.

asterite:
Thanks!

I got confused because the issue said "Comptime should probably fail in a similar way to ACIR, disallowing side effects in case the block is guaranteed to fail." so when I tried it comptime still fails and printed to stdout, so I thought it wasn't fixed.

vezenovm:
> I got confused because the issue said "Comptime should probably fail in a similar way to ACIR, disallowing side effects in case the block is guaranteed to fail." so when I tried it comptime still fails and printed to stdout, so I thought it wasn't fixed.

Ah yes, it is the issue that is written incorrectly. It is the reverse, `comptime` is correct and ACIR needed to fail in the same way with the same side effects ordering.

PR #9805, opened 2025-09-10
fix(licm): Do not hoist unsafe `array_get` in Brillig

aakoshh:
# Description

## Problem\*

Resolves #9804

## Summary\*

_DRAFT: Opening for discussion._

## Documentation\*

Check one:
- [ ] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist\*

- [ ] I have tested the changes locally.
- [ ] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #9807, opened 2025-09-10
fix(ssa): Use `can_be_hoisted` in `constant_folding`

aakoshh:
# Description

## Problem\*

Alternative fix to https://github.com/noir-lang/noir/pull/9805

## Summary\*

Experiment with keeping https://github.com/noir-lang/noir/pull/9712 which considers `ArrayGet` in `Brillig` to return `false` from `requires_acir_gen_predicate` and `has_side_effects`, and instead change `constant_folding` to base its decision on `loop_invariant::can_be_hoisted` to decide whether to deduplicate instructions into the common denominator, instead of `has_side_effects`.

Change `can_be_hoisted` to have special handling for `ArrayGet` where we just look whether the index is safe, ignoring the runtime.

This required using `has_side_effects` from `can_be_hoisted` for `Binary` instructions, to pass the unit tests in `constant_folding`; `has_side_effects` has some extra sophistication around constants in certain operations (see below).

## Additional Context

`has_side_effect` and `requires_acir_gen_predicate` are not equivalent for `Binary`:
* `has_side_effects`:
   * considers [unchecked arithmetic](https://github.com/noir-lang/noir/blob/8cfbdb9d2cbe21faa883271d3259c726db275ad5/compiler/noirc_evaluator/src/ssa/ir/instruction.rs#L532-L533) to have side effects unless they operate on `Field`
   * inspects whether the [number of bits](https://github.com/noir-lang/noir/blob/8cfbdb9d2cbe21faa883271d3259c726db275ad5/compiler/noirc_evaluator/src/ssa/ir/instruction.rs#L538-L544) is known in `shl` and `shr`
   * inspects whether the denominator in `div` and `mod` are known constants
* `requires_acir_gen_predicate`:
  * considers [unchecked arithmetic](https://github.com/noir-lang/noir/blob/8cfbdb9d2cbe21faa883271d3259c726db275ad5/compiler/noirc_evaluator/src/ssa/ir/instruction.rs#L846-L851) to require predicate for unsigned only
  * [does not inpect](https://github.com/noir-lang/noir/blob/8cfbdb9d2cbe21faa883271d3259c726db275ad5/compiler/noirc_evaluator/src/ssa/ir/instruction.rs#L853) `div`, `mod`, `shl` and `shr` for known constants.

Could these be reconciled?

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
@@ -883,11 +883,9 @@ pub(super) fn can_be_hoisted(instruction: &Instruction, dfg: &DataFlowGraph) -> 
 
         // These can have different behavior depending on the predicate.
         Binary(_) => {
-            if !instruction.has_side_effects(dfg) {
-                Yes
-            } else {
-                WithPredicate
-            }
+            // `requires_acir_gen_predicate` and `has_side_effects` make slightly different
+            // decisions for binary ops, e.g. the latter looks for safe constants.
+            if !instruction.has_side_effects(dfg) { Yes } else { WithPredicate }
         }
 
         // ArrayGet is considered not to need a predicate and be "pure" in Brillig,
@@ -1681,7 +1679,8 @@ mod test {
     #[test]
     fn hoist_safe_div() {
         // This test is identical to `do_not_hoist_unsafe_div`, except the loop
-        // in this test starts with a lower bound of `1`.
+        // in this test starts with a lower bound of `1`, so it will never divide
+        // by a zero induction variable.
         let src = "
         brillig(inline) fn main f0 {
           b0(v0: u32):
@@ -1744,6 +1743,51 @@ mod test {
             jmp b4(v14)
         }
         ");
+    }
+
+    /// Test whether a division that doesn't depend on the induction variable gets hoisted out of a loop,
+    /// when it's in a control dependent block, which makes it depend on a predicate.
+    /// With a decision based on `requires_acir_gen_predicate` it would not be hoisted, but based on `has_side_effects` it is.
+    #[test_case(true, 10, true)]
+    #[test_case(false, 10, true)]
+    #[test_case(true, 0, false)]
+    #[test_case(false, 0, false)]
+    fn hoist_invariant_div(loop_executes: bool, denominator: u32, should_hoist: bool) {
+        let (lower, upper) = if loop_executes { (0, 5) } else { (0, 0) };
+        let src = format!(
+            "
+        brillig(inline) fn main f0 {{
+          b0(v0: u32, v1: u1):
+            jmp b1(u32 {lower})
+          b1(v2: u32):
+            v3 = lt v2, u32 {upper}
+            jmpif v3 then: b2, else: b3
+          b2():
+            jmpif v1 then: b4, else: b5
+          b3():
+            return
+          b4():
+            v4 = div v0, u32 {denominator}
+            jmp b5()
+          b5():
+            v5 = unchecked_add v2, u32 1
+            jmp b1(v5)
+        }}
+        "
+        );
+
+        let ssa = Ssa::from_str(&src).unwrap();
+        let ssa = ssa.loop_invariant_code_motion();
+
+        println!("{ssa}");
+
+        // The pre-header of the loop b1 is b0
+        let pre_header = &ssa.main().dfg[BasicBlockId::new(0)];
+        if !should_hoist {
+            assert!(pre_header.instructions().is_empty(), "should not hoist");
+        } else {
+            assert_eq!(pre_header.instructions().len(), 1, "should hoist");
+        }
     }
 
     #[test]


aakoshh:
It looks like using `has_side_effects` instead of `requires_acir_gen_predicate` for `Binary` in `can_be_hoisted` causes some big size regressions. Perhaps because `has_side_effects` returns `false` for `Cast`, whereas `can_be_hoisted` only returns true if it can prove it's safe, so some `Cast`s won't be deduplicated in constant folding.

Is it worth trying to bring some of the logic to inspect constants into `requires_acir_gen_predicate`, even if we don't adopt this approach of considering `ArrayGet` pure in Brillig?

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -2095,6 +2108,8 @@ mod test {
         "#
         );
 
+        println!("{src}");
```
aakoshh:
```suggestion
```

aakoshh:
It looks like in `lambda_from_array` the difference in the bytecode size came from not being able to hoist `make_array` instructions into the common dominator, because in Brillig we considered it unsafe to be hoisted, and the `can_hoist_invariant` method had an override for it. 

Changed it so that `can_be_hoisted` returns `WithRefCount` for this in Brillig, which now `constant_folding` interprets as `true`.

vezenovm:
> Is it worth trying to bring some of the logic to inspect constants into `requires_acir_gen_predicate`, even if we don't adopt this approach of considering `ArrayGet` pure in Brillig?

We would have to look at how the instructions are implemented in ACIR. For certain operations we insert values based upon the predicate even if they would not fail. For example we [use the current predicate for div](https://github.com/noir-lang/noir/blob/8cfbdb9d2cbe21faa883271d3259c726db275ad5/compiler/noirc_evaluator/src/acir/mod.rs#L1088) even if we know the operation is technically a "safe" div.

compiler/noirc_evaluator/src/ssa/opt/loop_invariant.rs
```diff
@@ -793,10 +792,11 @@ fn get_induction_var_bounds(
 
 /// Indicate whether an instruction can be hoisted.
 #[derive(Debug, PartialEq, Eq)]
-enum CanBeHoistedResult {
+pub(super) enum CanBeHoistedResult {
     Yes,
     No,
     WithPredicate,
+    WithRefCount,
```
vezenovm, line 799: I like this additional enum variant instead of the early return on `MakeArray` before `can_be_hoisted`. However, I'm not sure if the rest of the PR is necessary with https://github.com/noir-lang/noir/pull/9805 now being merged. It makes sense to me now to say that an OOB Brillig array get does in fact have side effects as it can potentially return an unexpected type.

aakoshh: Okay, then I’ll undo those and keep this. Do you see value in trying to use can_be _hoisted from constant_folding, or was that good based on has_side_effects?

vezenovm: I think it is good based off of `has_side_effects`. I think what they are trying to tell us is different. Deduplicating vs. hoisting a single instruction can lead to different semantics. For example, `Cast` does not have side effects on its own and can be deduplicated if we find a matching `Cast`. However, it may depend on a range check and cannot be hoisted when in isolation.

aakoshh: Will the deduplication across sibling blocks in the the common dominator not detach the Cast from its RangeCheck though?

vezenovm: As it is a deduplication we can assume that the other cast has an appropriate range check.

aakoshh: What I mean is, can we not end up in a situation where we have 2 identical range checks and casts in 2 blocks which are not going to be executed, but because those instructions are duplicated, the casts, but not the range checks, and perhaps something that uses them, are all hoisted into the common dominator that _does_ get executed? 

Something like:
```
if a {
  range_check b to 32 bits
  let c = cast b as u32
  let d = ... c ...;
  foo(d);
}
if e { 
  range_check b to 32 bits
  let c = cast b as u32
  let d = ... c ...;
  bar(d);
}
```
becoming:
```
let c = cast b as u32;
let d = ... c ...; // oops, fails because the cast was invalid somehow?
if a {
  range_check b to 32 bits; // can't be hoisted
  foo(d);
}
if e { 
  range_check b to 32 bits; // can't be hoisted
  bar(d);
}
```

vezenovm: Good point. I wonder if we could create a failing test here. We can most likely use the Cast failure from LICM as inspiration.

My main concern was the discrepancies we have between `has_side_effects` and `can_be_hoisted`. Reviewing again I do see that you changed the `Binary` case in `can_be_hoisted` to use `has_side_effects` as to mitigate those differences. For some reason though I thought I remember not using the correct predicate potentially breaking div and having to use `requires_acir_gen_predicate`, but it looks like all tests are passing (perhaps something was fixed, maybe #8934). However, we do [use the current predicate for div](https://github.com/noir-lang/noir/blob/8cfbdb9d2cbe21faa883271d3259c726db275ad5/compiler/noirc_evaluator/src/acir/mod.rs#L1088) no matter if the divisor is a non-zero constant. If we had something like `v0 / 10` under a predicate in a loop it would technically be hoisted now although it still relies on the current side effects var. Perhaps this is fine, but I am not sure right now. Could you make a unit test to show this new hoisting functionality? I want to take a wack at trying to create a failing test tomorrow before merging.

aakoshh: I'll try to create some tests for this. A quick check with this program shows the casts are not hoisted when we use `has_side_effects` as the criteria:
```rust
fn main(a: bool, c: i8) {
    if a {
        let d = c * 127;
    }
    if a {
        let d = c * 127;
    }
}
```
But that I think is because the cache is keyed by `Instruction`, which act on different IDs, which makes `cast v34 as u8` different from `cast v19 as u8`, even though the series of instructions leading up to them are the same.
```rust
After Constant Folding (1) (step 26):
  b0(v0: u1, v1: u1, v2: i8):
    v3 = cast v2 as u8
    v4 = cast v2 as Field
    jmpif v0 then: b1, else: b2
  b1():
    v5 = cast v2 as u8
    v7 = lt v5, u8 128
    v8 = not v7
    v9 = cast v2 as Field
...
    v18 = mul v16, Field 127
    range_check v18 to 8 bits, "attempt to multiply with overflow"
    v19 = truncate v18 to 8 bits, max_bit_size: 254
    v20 = cast v19 as u8
    v21 = cast v8 as u8
    v22 = unchecked_add u8 128, v21
    v23 = lt v20, v22
    constrain v23 == u1 1, "attempt to multiply with overflow"
    jmp b2()
  b2():
    jmpif v0 then: b3, else: b4
  b3():
    v25 = lt v3, u8 128
...
    v33 = mul v32, Field 127
    range_check v33 to 8 bits, "attempt to multiply with overflow"
    v34 = truncate v33 to 8 bits, max_bit_size: 254
    v35 = cast v34 as u8
...
    jmp b4()
  b4():
    return
}
```
I'll try to see if I can get some loop into the mix, to match the IDs.

aakoshh: This test definitely fails with the definition based on `has_side_effects`:
```rust
    #[test]
    fn does_not_hoist_range_checked_cast_to_common_ancestor() {
        let src = "
            brillig(inline) fn main f0 {
              b0(v0: Field, v1: u1):
                jmpif v1 then: b1, else: b2
              b1():
                range_check v0 to 8 bits
                v2 = cast v0 as u8
                jmp b2()
              b2():
                jmpif v1 then: b3, else: b4
              b3():
                range_check v0 to 8 bits
                v3 = cast v0 as u8
                jmp b4()
              b4():
                return
            }
            ";
        assert_ssa_does_not_change(src, Ssa::fold_constants_using_constraints);
    }

```

aakoshh: Here's a way I found to make the `cast` detach from the `range_check` and bubble up into the common dominator. 

There is this comment in constant folding:
> // This will only move the current instance of the instruction right now.
> // When constant folding is run a second time later on, it'll catch
> // that the previous instance can be deduplicated to this instance.

In our example we have 2 `if` blocks with a series of instructions where the first few instructions refer to variables in the common ancestor blocks, then they do the same operations, but use different IDs. Only the first instructions get hoisted, where the IDs match. The results of the hoisted (moved) instructions will be different, and they appear next to each other in the common ancestor, however, as the comment says, in the next run they will be deduplicated. At that point, the _next_ instruction in the `if` blocks will again refer the same variables, and _they_ can be moved. 

So, what if we repeatedly apply constant folding, as the comment suggests? The effect is that one by one the instructions bubble up in the common dominator.

Take the SSA of the following program just before constant folding:
```rust
fn main(a: bool, c: i8) -> pub i8 {
    let mut s = 0;
    if a {
        let d = c * 127;
        s += d;
    }
    if a {
        let d = c * 127;
        s += d;
    }
    s
}
```

It looks like this:
<details>
  <summary>After Purity Analysis (2) (step 25):</summary>
  
```rust
  b0(v0: u1, v1: i8):
    v2 = allocate -> &mut i8
    store i8 0 at v2
    jmpif v0 then: b1, else: b2
  b1():
    v5 = unchecked_mul v1, i8 127
    v6 = cast v5 as u16
    v7 = truncate v6 to 8 bits, max_bit_size: 16
    v8 = cast v1 as u8
    v10 = lt v8, u8 128
    v11 = not v10
    v12 = cast v1 as Field
    v13 = cast v10 as Field
    v14 = mul v13, v12
    v16 = sub Field 256, v12
    v17 = cast v11 as Field
    v18 = mul v17, v16
    v19 = add v14, v18
    v21 = mul v19, Field 127
    range_check v21 to 8 bits, "attempt to multiply with overflow"
    v22 = truncate v21 to 8 bits, max_bit_size: 254
    v23 = cast v22 as u8
    v24 = not v10
    v25 = cast v24 as u8
    v26 = unchecked_add u8 128, v25
    v27 = lt v23, v26
    constrain v27 == u1 1, "attempt to multiply with overflow"
    v29 = cast v7 as i8
    v30 = truncate v29 to 8 bits, max_bit_size: 9
    v31 = cast v30 as u8
    v32 = cast v7 as u8
    v33 = lt v32, u8 128
    v34 = lt v31, u8 128
    v35 = unchecked_mul v34, v33
    constrain v35 == v33, "attempt to add with overflow"
    store v30 at v2
    jmp b2()
  b2():
    jmpif v0 then: b3, else: b4
  b3():
    v36 = unchecked_mul v1, i8 127
    v37 = cast v36 as u16
    v38 = truncate v37 to 8 bits, max_bit_size: 16
    v39 = cast v1 as u8
    v40 = lt v39, u8 128
    v41 = not v40
    v42 = cast v1 as Field
    v43 = cast v40 as Field
    v44 = mul v43, v42
    v45 = sub Field 256, v42
    v46 = cast v41 as Field
    v47 = mul v46, v45
    v48 = add v44, v47
    v49 = mul v48, Field 127
    range_check v49 to 8 bits, "attempt to multiply with overflow"
    v50 = truncate v49 to 8 bits, max_bit_size: 254
    v51 = cast v50 as u8
    v52 = not v40
    v53 = cast v52 as u8
    v54 = unchecked_add u8 128, v53
    v55 = lt v51, v54
    constrain v55 == u1 1, "attempt to multiply with overflow"
    v56 = cast v38 as i8
    v57 = load v2 -> i8
    v58 = unchecked_add v57, v56
    v59 = truncate v58 to 8 bits, max_bit_size: 9
    v60 = cast v59 as u8
    v61 = cast v57 as u8
    v62 = cast v38 as u8
    v63 = lt v61, u8 128
    v64 = lt v62, u8 128
    v65 = eq v63, v64
    v66 = lt v60, u8 128
    v67 = eq v66, v63
    v68 = unchecked_mul v67, v65
    constrain v68 == v65, "attempt to add with overflow"
    store v59 at v2
    jmp b4()
  b4():
    v69 = load v2 -> i8
    return v69
}
```
</details>

Now apply constant folding repeatedly, using the CLI in https://github.com/noir-lang/noir/pull/9826

```shell
cat ./pre-constant-folding.ssa \
        | cargo run -q -p noir_ssa_cli -- transform --ssa-pass "constant folding" \
        | cargo run -q -p noir_ssa_cli -- transform --ssa-pass "constant folding" \
        | cargo run -q -p noir_ssa_cli -- transform --ssa-pass "constant folding" \
        | cargo run -q -p noir_ssa_cli -- transform --ssa-pass "constant folding" \
        | cargo run -q -p noir_ssa_cli -- transform --ssa-pass "constant folding" \
        | cargo run -q -p noir_ssa_cli -- transform --ssa-pass "constant folding" \
        | cargo run -q -p noir_ssa_cli -- transform --ssa-pass "constant folding" \
        | cargo run -q -p noir_ssa_cli -- transform --ssa-pass "constant folding" \
        | cargo run -q -p noir_ssa_cli -- transform --ssa-pass "constant folding"
```

The result is that the cast after the `range_check` is now in `b0`:
<details>
  <summary>After Constant Folding (step 26):</summary>
  
```rust
  b0(v0: u1, v1: i8):
    v2 = allocate -> &mut i8
    store i8 0 at v2
    v49 = unchecked_mul v1, i8 127
    v6 = cast v1 as u8
    v7 = cast v1 as Field
    v8 = cast v49 as u16
    v50 = lt v6, u8 128
    v51 = sub Field 256, v7
    v13 = truncate v8 to 8 bits, max_bit_size: 16
    v14 = not v50
    v15 = cast v50 as Field
    v52 = mul v15, v7
    v17 = cast v14 as Field
    v18 = cast v14 as u8
    v19 = cast v13 as i8
    v20 = cast v13 as u8
    v53 = mul v17, v51
    v54 = unchecked_add u8 128, v18
    v55 = lt v20, u8 128
    v56 = add v52, v53
    v57 = mul v56, Field 127
    v27 = truncate v57 to 8 bits, max_bit_size: 254
    v36 = cast v27 as u8
    jmpif v0 then: b1, else: b2
  b1():
    range_check v57 to 8 bits, "attempt to multiply with overflow"
    v29 = cast v27 as u8
    v58 = lt v29, v54
    constrain v58 == u1 1, "attempt to multiply with overflow"
    v32 = truncate v19 to 8 bits, max_bit_size: 9
    v33 = cast v32 as u8
    v59 = lt v33, u8 128
    v60 = unchecked_mul v59, v55
    constrain v60 == v55, "attempt to add with overflow"
    store v32 at v2
    jmp b2()
  b2():
    jmpif v0 then: b3, else: b4
  b3():
    range_check v57 to 8 bits, "attempt to multiply with overflow"
    v61 = lt v36, v54
    constrain v61 == u1 1, "attempt to multiply with overflow"
    v38 = load v2 -> i8
    v62 = unchecked_add v38, v19
    v40 = truncate v62 to 8 bits, max_bit_size: 9
    v41 = cast v40 as u8
    v42 = cast v38 as u8
    v63 = lt v42, u8 128
    v64 = eq v63, v55
    v65 = lt v41, u8 128
    v66 = eq v65, v63
    v67 = unchecked_mul v66, v64
    constrain v67 == v64, "attempt to add with overflow"
    store v40 at v2
    jmp b4()
  b4():
    v48 = load v2 -> i8
    return v48
}
```
</details>

TomAFrench: > We can most likely use the Cast failure from LICM as inspiration.

I think if this is a recurring issue that we're running into then we should just lean into creating a safe and unsafe cast similar to checked and unchecked arithmetic.

aakoshh: Try as I might, I couldn't find an example where hoisting the unsafe cast during constant folding could cause an issue, because (currently) the instructions which the variable would be used in would also have to come from the block the cast was hoisted from, and they are all marked as having side effects, so they are not hoisted outside their predicated blocks. For example I can't get an `array_get` or a checked multiplication to go along and cause problem.

Also, whenever there is a range constraint, there also seems to be a truncate, so while the data might not be intentional, it's also not outside the range of the type we cast into.

How exactly do we define an unsafe cast?

TomAFrench: I define an unsafe cast as one which doesn't include the necessary truncation to ensure that the resulting value fits into the output type. Basically all casts in the ACIR runtime today are unsafe casts.

A safe cast by comparison implicitly truncates the input value to fit into the result type. All casts in the brillig runtime today are safe casts.

vezenovm: @aakoshh How about with something similar to the `do_not_hoist_control_dependent_cast` test in `loop_invariant`?

aakoshh: Right, so in our case it looks like as long as a truncate was emitted, the cast will stay safe, because with the logic based on `has_side_effects` both can be hoisted. `can_be_hoisted` hoists all `truncate`, but not all `cast`, so it's even more conservative. 

Furthermore, not sure if it's by design or coincidence, but constant folding happens after the flattening of the CFG in ACIR, so nothing is hoisted into any common ancestor, as there is only one block.

vezenovm: I'm realizing we also already hoist safe div/mod instructions when we know the induction variable is the divisor and will never be zero. So it is most likely fine to use `has_side_effects` inside of `can_be_hoisted` over `requires_acir_gen_predicate`.

aakoshh: > How about with something similar to the do_not_hoist_control_dependent_cast test in loop_invariant?

@vezenovm that's exactly what I used as a starter for https://github.com/noir-lang/noir/pull/9807#discussion_r2343456029

As I demonstrated in https://github.com/noir-lang/noir/pull/9807#discussion_r2343784633 constant folding, when applied a number of times, will hoist the `cast` into the common ancestor (based on `has_side_effects`), but then I couldn't come up with anything to exploit this and turn it into a runtime failure, because in practice it was protected by the `truncate` which got hoisted with it. I used Brillig to even have a block to hoist into, and tried the SSA CLI to remove the truncate, but the interpreter did not die even then. I also tried `shl <casted-i8>, 7` but that just returns 0 even on an overflowed `i8` 🤷 Anything that would fail doesn't get hoisted.

vezenovm: > Furthermore, not sure if it's by design or coincidence, but constant folding happens after the flattening of the CFG in ACIR, so nothing is hoisted into any common ancestor, as there is only one block.

The side effects var is only used during `fold_constants_using_constraints` so I do believe in this case it was intentional. After adding dominance based deduplication though I think this should be fine to place before flattening. However, I don't know if we would desire that as constant folding stores a lot of stuff in memory and performing it early in the pipeline may not lead to desired results.

vezenovm: > @vezenovm that's exactly what I used as a starter for [#9807 (comment)](https://github.com/noir-lang/noir/pull/9807#discussion_r2343456029)

Ah sorry I was reading too quickly and skimmed over the actual Noir snippet. 

> because in practice it was protected by the `truncate` which got hoisted with it.

I guess this is where the difference between hoisting in isolation vs. deduplication came into play. As we only hoist into the common denominator if the instructions are matching we don't end up making a safe cast unsafe as its preceding instructions are expected to also be hoisted. This does still feel slightly brittle but it seems to be fine based off your testing.

As I see no regressions I would be good with moving to using `can_be_hoisted` in `constant_folding` but curious if @TomAFrench has an opinion here. The main reason I would advocate for still just using `Instruction::has_side_effects` would be to maintain the simplicity of `constant_folding`.

PR #9818, opened 2025-09-10
fix: Fix post-order and DomTree block ordering

### LINKED ISSUE #9771: mem2reg is performed in an incorrect block order
### Aim

I was running into this in the alias graph refactor: for a given function:
```
acir(inline) fn main f0 {
  b0():
    v1 = allocate -> &mut Field
    store Field 0 at v1
    v3 = allocate -> &mut &mut Field
    store v1 at v3
    jmp b1(Field 0)
  b1(v0: Field):
    v4 = eq v0, Field 0
    jmpif v4 then: b2, else: b3
  b2():
    v12 = load v3 -> &mut Field
    store Field 2 at v12
    v14 = add v0, Field 1
    jmp b1(v14)
  b3():
    v5 = load v1 -> Field
    v7 = eq v5, Field 2
    constrain v5 == Field 2
    v8 = load v3 -> &mut Field
    v9 = load v3 -> &mut Field
    v10 = load v9 -> Field
    v11 = eq v10, Field 2
    constrain v10 == Field 2
    return
}
```
We iterate in reverse post order which is:
```
block order = [Id(0), Id(1), Id(3), Id(2)]
```
Notably, this has us analyzing b3 before its predecessor b2 is finished.

### Expected Behavior

Barring loops, we should analyze all blocks only after their predecessor has finished being analyzed

### Bug

We do not do this.

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

jfecher:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/9771

## Summary\*

Changes our post order to a reverse topological ordering... except there are a few difficulties:
- Topological ordering is only defined for DAGs. For any program with a loop it is not defined so I had to define what it meant for blocks within a loop.
- Existing algorithms (notably kosaraju_scc and DfsPostOrder) give a general post ordering but leave blocks within loops in an undefined order. It turns out this undefined order breaks several of our optimization passes so it is unusable.
- A rough, mostly correct definition of what we want for topological ordering is to order all blocks by the _maximum_ cost to reach them from the entry block, using each edge at most once. Using the minimum cost would skip over loops. Unfortunately, it turns out this is the maximum-path problem which is NP-hard.
- The current solution in this PR is based on a semi-arbitrary ordering of blocks within loops by cutting off edges from the loop header to treat it as a new root to rerun the algorithm recursively on. This is rather slow though - we already stall on one unit test for a bit so I imagine running it on a larger real-world program would be prohibitive.

## Documentation\*

Check one:

# PR Checklist\*


aakoshh:
I wonder if the `branch_analysis` results could be used to help with the limits for finding regions to be ordered. For example if we encounter `b1` and it branches into `b2` and `b3`, and we know that it rejoins at `b10`, then we can recursively process the `b2` and then the `b3` branch up until `b10`, then resume from there.

aakoshh:
Ah, never mind, I forgot the branch analysis is on the unrolled CFG without loops 🤦

aakoshh:
Another idea that came to me was that perhaps when we encounter a "back edge" then we could replace it with an edge to the "other" node from where we branched off. For example:
```
   b2 - b3 - b4
  /          |
b1 <---------+
  \ 
   b5 --------------b6
```
When we encounter the `b4 -> b1` back-edge, we would know we went down the route of `b1 -> b2` to reach `b4`, so we replace `b4 -> b1` with `b4 -> b5`. Once all back-edges have been removed, we topo-sort the resulting graph.

jfecher:
@aakoshh how do we identify a back-edge in that case?

If we're traversing a graph in general then a "branch to an already visited node" could also just be a if-else case where we took the path of `entry -> then -> end_if` first then later went back to `entry -> else -> end_if` and we would already have visited `end_if` so it'd look like a back-edge. We'd maybe need a reachability analysis like an SCC algorithm to determine if end_if is reachable from itself I think, which gets us closer to the algorithm in this PR currently.

I'm going to try an approach where I order blocks using information on how we actually lay out `jmp_if`s in practice in ssa-gen. For example, when we have a `jmp_if v0 then: b1, else: b2` these could only be:
- a if-then-else: `b1` is the then block, `b2` is the else block
- a if-then: `b1` is the then block, **`b2` is the end block**
- a loop: `b1` is the jump into the loop, **`b2` is the jump out of the loop
- a break/continue in brillig

In all cases `b2` leads out of the loop/then construct so maybe I can use how we construct blocks in practice to just always take the then block first and order it that way. The main thing this would need is a way to handle if-then-else specifically so we stop at the end block and take the else branch before it since that is the only case of the above where the end block is unknown.

aakoshh:
> If we're traversing a graph in general then a "branch to an already visited node" could also just be a if-else case where we took the path of entry -> then -> end_if first then later went back to entry -> else -> end_if and we would already have visited end_if so it'd look like a back-edge. 

I think what you mean is this:
```
   b1 
  /  \
b0    b3 - b4
^ \  /      |
|  b2       |
+-----------+
```
We visit `b0` (entry), `b1` (then), `b3` (end_if), `b4`, and then go back to traverse `b2` (else) and see `b3` (end_if) again and think it's a back edge because we have seen it already? 

Sorry, I failed to mention that I thought about a DFS traversal, where we would only consider "visited" as in the path we reached the current node from (e.g. some `im` data structure), so for example if we go over the successors of `b1`, and call traverse on `b1 -> b3 -> b4`, then we see that the `b4 -> b0` edge is a back-edge because `b0` is on the path. Later when we go down the `b2 -> b3` route, `b3` is not considered visited on the current path, nor is `b4`, but at this point the `b4 -> b0` edge would have been replaced by a `b4 -> b2` edge. 

This example shows that this idea doesn't really work for this case, because the `b0` junction didn't provide an _else_ branch that didn't end up back at `b4`, and by rerouting to `b2` we would now be unable to apply a similar trick, as there is not even an _else_ to consider. 

Not sure if this actually comes up in our CFG. But I agree with you that we could find something specific to the CFG nature of our graph that could help, that was the intuition behind this idea as well.

jfecher:
@aakoshh I was considering that graph but without b4 and was just wondering how you'd identify loops in general. Keeping track of currently visited nodes for the same path is a good idea, I'll see if that can get me anywhere. There's a neat data structure I heard about called a `VList` which sounds like it'd be good for this.

I'm not sure if your exact scenario could be possible in practice but the main case I'm aware of where it is possible we can have looping like that with no exits is with `loop {}` in brillig. We should have errors that prevent these from being used with no `break` whatsoever but it's also possible this topological sort is core enough that we need it to be working even for those invalid examples. That being said, for such examples an arbitrary sort for the loop blocks may also be fine.

jfecher:
It looks like a decent chunk of the overhead I'm getting with some algorithms I'm using is the fact I'm often converting from CFG to a petgraph (to use that library's algorithms) then back to a CFG. For tests like `test_large_unroll_stack_overflow` with ~70k blocks this already becomes an issue, and some real world noir programs have more than this. Any efficient algorithm will need to avoid the petgraph conversion entirely I think.

PR #9823, opened 2025-09-11
fix: allow overwrites at load factor threshold in HashMap::insert

Forostovec:
Move the load factor assertion inside the branch that performs a true insertion (when the slot is available) and remove the unconditional check at the start of insert. This ensures that overwriting an existing key’s value is allowed even when the map is at or above the α_max threshold, since overwrites do not increase length or degrade the load factor. The change preserves the existing behavior of asserting on new unique inserts beyond the threshold, aligning with the load factor test and the intent to warn only when capacity-related performance would degrade.


TomAFrench:
Looks to be an LLM account with spurious PRs in other repos.

PR #9824, opened 2025-09-11
chore: use enum for instruction deduplication safety

TomAFrench:
# Description

## Problem\*

Resolves 

## Summary\*

This PR updates `can_be_duplicated` to return a `CanBeDeduplicated` enum to make the three options more explicit.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/constant_folding/mod.rs
```diff
@@ -415,11 +427,12 @@ impl Context {
         // If the instruction doesn't have side-effects and if it won't interact with enable_side_effects during acir_gen,
         // we cache the results so we can reuse them if the same instruction appears again later in the block.
         // Others have side effects representing failure, which are implicit in the ACIR code and can also be deduplicated.
-        let can_be_deduplicated = can_be_deduplicated(&instruction, dfg, self.use_constraint_info);
+        let can_be_deduplicated = can_be_deduplicated(&instruction, dfg);
+
+        let use_constraint_info = self.use_constraint_info;
+        let is_make_array = matches!(instruction, Instruction::MakeArray { .. });
```
asterite, line 425: Why `is_make_array` is handled separately? I'm seeing that `can_be_deduplicated` is only called once so maybe that could return `CanBeDeduplicated::Always` for `MakeArray`?

TomAFrench: There's a comment in `can_be_deduplicated` here which touches on this. I was undecided on whether we want to make the change you suggest so I left it as is.

Issue #8707, opened 2025-05-28
Move SSA and ABI related code into common libraries


aakoshh:
Follow up https://github.com/noir-lang/noir/pull/8700 with moving code from the AST fuzzer to a place they can be shared without having to add the fuzzer as a dependency:
* Producing an ABI description from the monomorphized AST
* Converting ABI inputs into SSA interpreter arguments

Then remove the dependency on `noir_ast_fuzzer` from `nargo_cli`.

PR #9826, opened 2025-09-11
feat(ssa): SSA CLI

### LINKED ISSUE #8503: Make a binary for the SSA interpreter

aakoshh:
# Description

## Problem\*

Resolves #8503

## Summary\*

Adds a `noir-ssa` CLI tool. There are two main sub-commands:
* `transform` applies SSA pass(es) and prints the resulting SSA
* `interpret` takes a TOML formatted input file and runs it with the SSA interpreter

Both take SSA from a file or `stdin`, so it is possible to chain multiple commands with by piping the output, for example:
```shell
cat ./initial.ssa \
  | noir-ssa transform --ssa-pass "Remove Unreachable Instructions" \
  | noir-ssa transform --ssa-pass mem2reg --ssa-pass "Dead Instruction elimination" \
  | noir-ssa interpret --input-path ./inputs.toml 
```

We can specify multiple `--ssa-pass` options, which is a bit less verbose than piping, and saves re-parsing the SSA after each pass. The passes are executed in the order of appearance, unlike with `nargo interpret`, where they are just a filter to tell at which point the SSA should be interpreted, while it goes through the whole pipeline. 

We can also use the `--show-ssa` and `--show-ssa-pass` options to see the SSA along the way, but only the last state is printed to `stdout`, so that it can be piped to `interpret` for example; the interim states go to `stderr`.

The inputs for the circuit are expected to be in JSON or TOML format, with the same syntax as the ABI values in a typical `Prover.toml`, but using the SSA variable IDs for parameter names. For example if we have this Noir function:
```rust
fn main(a: [[u8; 2]; 2], b: bool, c: u32) { ... }
```
and it's turned into this SSA:
```rust
acir(inline) fn main f0 {
  b0(v0: [[u8; 2]; 2], v1: u1, v2: u32):
...
```
then we can `interpret` it with the following inputs:
```toml
v0 = [[1, 2], [3, 4]]
v1 = true
v2 = 0
```

The way it works is that we generate an artificial ABI from SSA, and use that ABI to parse the input file. (For this we are reusing some existing utilities that are a bit scattered around. There is already a ticket to move them: https://github.com/noir-lang/noir/issues/8707).

The inputs can come from a file with `--input-path`, or for quick and simple cases it can be specified on the CLI using `--input-toml` or `--input-json`. When using `--input-toml`, we can use `;` to define multiple keys, like `v0 = 1; v1 = [2,3]`. 

`interpret` prints the output to `stdout`, but not the SSA source. This could potentially be useful if we wanted to alternate transforming and interpreting, but it's probably an overkill. Easy to add later anyway.

## Additional Context

```console
For more information about this error, try `rustc --explain E0602`.
CLI tool to work with SSA independently of any Noir program

Usage: noir-ssa [OPTIONS] <COMMAND>

Commands:
  list       List the SSA passes we can apply
  check      Parse and (optionally) validate the SSA
  interpret  Parse the input SSA and it arguments, run the SSA interpreter, then write the return values to stdout
  transform  Parse the input SSA, run some SSA passes on it, then write the output SSA
  help       Print this message or the help of the given subcommand(s)

Options:
  -s, --source-path <SOURCE_PATH>
          Path to the source SSA.
          
          If empty, the SSA will be read from stdin.

      --no-validate
          Turn off validation of the source SSA.
          
          This can be used to test how invalid input behaves.

  -h, --help
          Print help (see a summary with '-h')

  -V, --version
          Print version
```

#### Interactive mode

For quickly trying out some SSA. Maybe it could ask for the SSA, then the TOML, which the user would copy-paste or type in. 

Currently this works for example like this:
```shell
pbpaste | cargo run -q -p noir_ssa_cli -- transform --ssa-pass mem2reg
pbpaste | cargo run -q -p noir_ssa_cli -- interpret --input-toml "v0 = false; v1=100"
```

#### Render the CFG

We often use pen and paper to draw out a CFG to better follow the jumps. It would be great to visualise it with Mermaid for example.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3577,6 +3577,7 @@ dependencies = [
  "noirc_driver",
  "noirc_errors",
  "noirc_evaluator",
+ "tempfile",
  "thiserror 1.0.69",
  "toml",
  "tracing-subscriber",
--- a/tooling/ssa_cli/Cargo.toml
+++ b/tooling/ssa_cli/Cargo.toml
@@ -21,6 +21,7 @@ color-eyre.workspace = true
 const_format.workspace = true
 thiserror.workspace = true
 toml.workspace = true
+tempfile.workspace = true
 tracing-subscriber.workspace = true
 
 # Noir repo dependencies
--- a/tooling/ssa_cli/src/cli/interpret_cmd.rs
+++ b/tooling/ssa_cli/src/cli/interpret_cmd.rs
@@ -3,14 +3,20 @@ use std::{collections::BTreeMap, path::PathBuf};
 use clap::Args;
 use color_eyre::eyre::{self, Context, bail};
 use iter_extended::vecmap;
-use noir_artifact_cli::commands::parse_and_normalize_path;
-use noirc_abi::{Abi, AbiParameter, AbiReturnType, AbiType, AbiVisibility, Sign};
+use noir_artifact_cli::{commands::parse_and_normalize_path, fs::artifact::write_to_file};
+use noirc_abi::{
+    Abi, AbiParameter, AbiReturnType, AbiType, AbiVisibility, InputMap, Sign,
+    input_parser::InputValue,
+};
 use noirc_errors::println_to_stdout;
 use noirc_evaluator::ssa::{
     interpreter::InterpreterOptions,
     ir::types::{NumericType, Type},
     ssa_gen::Ssa,
 };
+use tempfile::NamedTempFile;
+
+const TOML_LINE_SEP: char = ';';
 
 /// Parse the input SSA and it arguments, run the SSA interpreter,
 /// then write the return values to stdout.
@@ -18,11 +24,21 @@ use noirc_evaluator::ssa::{
 pub(super) struct InterpretCommand {
     /// Path to the input arguments to the SSA interpreter.
     ///
-    /// Expected to be in TOML format, similar to `Prover.toml`.
+    /// Expected to be in TOML format or JSON, similar to `Prover.toml`.
     ///
     /// If empty, we assume the SSA has no arguments.
-    #[clap(long, short, value_parser = parse_and_normalize_path)]
+    #[clap(long, short, value_parser = parse_and_normalize_path, conflicts_with = "input_json", conflicts_with = "input_toml")]
     pub input_path: Option<PathBuf>,
+
+    /// Verbatim inputs in JSON format.
+    #[clap(long, conflicts_with = "input_path", conflicts_with = "input_toml")]
+    pub input_json: Option<String>,
+
+    /// Verbatim inputs in TOML format.
+    ///
+    /// Use ';' to separate what would normally be multiple lines.
+    #[clap(long, conflicts_with = "input_path", conflicts_with = "input_json")]
+    pub input_toml: Option<String>,
 
     /// Turn on tracing in the SSA interpreter.
     #[clap(long, default_value_t = false)]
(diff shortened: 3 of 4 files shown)


tooling/ssa_cli/src/cli/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+    ///
+    /// If empty, the SSA will be read from stdin.
+    #[clap(long, short, global = true, value_parser = parse_and_normalize_path)]
+    source_path: Option<PathBuf>,
+
+    /// Turn off validation of the source SSA.
+    ///
+    /// This can be used to test how invalid input behaves.
+    #[clap(long, global = true, default_value_t = false)]
+    no_validate: bool,
```
vezenovm, line 41: Feels like we could get rid of this flag and just have `check`, but this is a preference nit.

aakoshh: You mean you would never not validate an SSA (makes sense), or that `check` should have a parse-only option?

vezenovm: > You mean you would never not validate an SSA (makes sense), or that check should have a parse-only option?

I was saying that `check` would just always validate and that the default would just be not running validation. It is then up to the user (in this case us) to validate by calling `check` and we can get rid of this flag. This is a minor preference though. I guess the other side of it is that just passing a `--validate` to other calls removes the need to make a separate call to `check` again which is QoL improvement. But I do think the default for `check` should just be to always validate.

vezenovm: I read the flags wrong lol

vezenovm: So the default is to always validate. I guess this is fine as this is a debug tool.

aakoshh: Ah, I see. We always validate, which I think is nice to do when we use the `transform` or `interpret` commands, as I in my experience trying to manually author SSA, I often end up trying to trying to use different types on the lhs and rhs of an operation. I suspect someone who wants to experiment with SSA directly would benefit from simple mistakes being caught early. At least I do it all the time when I try to insert a variable into a small example while keeping the IDs canonically increasing to keep it nice. 

I included `check` just so that we have something to see if an SSA parses without having to transform it or come up with inputs. 

The `--no-validate` was just a thought that _maybe_ someone would want to see what happens if we have a purposefully invalid SSA and try to run it.

aakoshh:
It just occurred to me that `check`should print the normalized SSA so it helps with the parameter name matching problem @asterite encountered

Savio-Sou:
Who could be relevant users of this new tool? (E.g. us internal engineers, auditors, tooling engineers, app developers...)

aakoshh:
Internal engineers and auditors I think.

PR #9827, opened 2025-09-11
chore: use the ACIR parser in redundant_ranges tests

### LINKED ISSUE #9577: Use the ACIR parser in unit tests
Follow up for https://github.com/noir-lang/noir/pull/9340#discussion_r2286092173 

Rewrite the unit tests in the `redundant_ranges` ACIR transformation to use the new ACIR parser.

asterite:
# Description

## Problem

Resolves #9577

## Summary

- I didn't end up extracting a new crate for the parser because I'm not sure that's needed, but let me know otherwise (for example the SSA parser isn't in its own crate)
- I did end up needing to extract a crate, `noirc_span`, because `noirc_errors` has a lot of dependencies
- This PR also fixes a bug in the ACIR parser where parsing a witness vector removed duplicate values (because it was first parsed using a BTreeSet)
- And comments in ACIR are now allowed

## Additional Context

- There are probably more tests that could use the ACIR parser but I wanted to start with just one file.
- I wonder if the sections "private parameters indice" (and public and return) could be optional when parsing. That would mean more concise tests. They could also be skipped when printing a circuit if they are empty. But maybe it's not a big deal...

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -72,6 +72,7 @@ dependencies = [
  "brillig_vm",
  "criterion",
  "indexmap 2.11.0",
+ "insta",
  "num-bigint",
  "pprof",
  "proptest",
--- a/acvm-repo/acvm/Cargo.toml
+++ b/acvm-repo/acvm/Cargo.toml
@@ -38,6 +38,7 @@ ark-bls12-381 = { version = "^0.5.0", default-features = false, features = [
 ark-ff.workspace = true
 ark-bn254.workspace = true
 bn254_blackbox_solver.workspace = true
+insta.workspace = true
 proptest.workspace = true
 num-bigint.workspace = true
 zkhash = { version = "^0.2.0", default-features = false }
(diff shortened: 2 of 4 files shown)


acvm-repo/acir/src/parser/mod.rs
```diff
@@ -176,8 +177,8 @@ impl<'a> Parser<'a> {
         Ok(witnesses)
     }
 
-    fn parse_witness_vector(&mut self) -> ParseResult<Vec<Witness>> {
-        self.parse_witness_ordered_set().map(|set| set.into_iter().collect::<Vec<_>>())
```
vezenovm, line 180: 🤦 nice catch

asterite: Haha, I found it by accident. One of the redundant_range tests had an INIT with duplicate witnesses and it produced a circuit that didn't match the original source.

PR #9828, opened 2025-09-11
chore(die): Encapsulate array access checks in separate module

vezenovm:
# Description

## Problem\*

Working towards #9517 

## Summary\*

We want to make it easier for reviewers to read code. Thus, I wanted to encapsulate the array access OOB logic to its own module in DIE. This essentially just moves this logic out of the main DIE pass to make the logic easier to follow.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/die/array_oob_checks.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/die/array_oob_checks.rs
@@ -103,7 +103,7 @@ impl Context {
                 );
                 let index = index.first();
                 let array_length =
-                    function.dfg.make_constant((array_length as u128).into(), length_type);
+                    function.dfg.make_constant(u128::from(array_length).into(), length_type);
 
                 let is_index_out_of_bounds = function.dfg.insert_instruction_and_results(
                     Instruction::binary(BinaryOp::Lt, index, array_length),


compiler/noirc_evaluator/src/ssa/opt/die/array_oob_checks.rs
```diff
@@ -0,0 +1,293 @@
...
+/// These checks are not laid down in the ACIR runtime as that runtime maps the SSA
+/// to a memory model where OOB accesses will be prevented. Essentially all array ops
+/// in ACIR will have a side effect where they check for the index being OOB.
+/// However, in order to maintain parity between the Brillig and ACIR runtimes,
+/// if we have an unused array operation we need insert an OOB check so that the
+/// side effects ordering remains correct.
+pub(super) fn should_insert_oob_check(function: &Function, instruction: &Instruction) -> bool {
+    use Instruction::*;
+    let might_be_out_of_bounds = match instruction {
+        ArrayGet { array, index, .. } | ArraySet { array, index, .. } => {
```
asterite: I know this PR just moves some code around, but we could handle `offset` here and assert it's `None`.

compiler/noirc_evaluator/src/ssa/opt/die/array_oob_checks.rs
```diff
@@ -0,0 +1,293 @@
...
+            // We only care about arrays here as slices are expected to have explicit checks laid down in the initial SSA.
+            function.dfg.try_get_array_length(*array).is_some()
+                && !function.dfg.is_safe_index(*index, *array)
+        }
+        _ => false,
+    };
+    function.runtime().is_acir() && might_be_out_of_bounds
```
asterite: We could check `function.runtime().is_acir()` as an early return.

compiler/noirc_evaluator/src/ssa/opt/die/array_oob_checks.rs
vezenovm: Good note, I added it here anyway

compiler/noirc_evaluator/src/ssa/opt/die/array_oob_checks.rs
vezenovm: I switched to early return if we do not have ACIR

PR #9831, opened 2025-09-11
fix(interpreter): Return fresh nested array on array get in ACIR

### LINKED ISSUE #9830: SSA interpreter: Different result after array set opt pass
### Aim

I added the following test as part of https://github.com/noir-lang/noir/pull/9789:
```noir
fn main(a: bool, zero: u32) -> pub bool {
    let G_C: [[bool; 3]; 1] = [[true, false, true]];
    let b = func_1(G_C[zero])[zero + 1];
    if a {
        G_C[zero][zero + 1]
    } else {
        b
    }
}
fn func_1(mut b: [bool; 3]) -> [bool; 3] {
    b[1] = true;
    b
}
```

### Expected Behavior

I would expect the actual result to be the same for every step of the interpreter. 

### Bug

This fails `nargo interpret` after the `array_set_optimization` pass.

Here is the SSA and nargo interpret output on failure:
```
After Array Set Optimizations (step 37):
  b0(v0: u1, v1: u32):
    v4 = make_array [u1 1, u1 0, u1 1] : [u1; 3]
    v5 = make_array [v4] : [[u1; 3]; 1]
    v7 = add v1, u32 1
    v8 = array_get v5, index v1 -> [u1; 3]
    v9 = array_set mut v8, index u32 1, value u1 1
    v10 = array_get v9, index v7 -> u1
    enable_side_effects v0
    v11 = add v1, u32 1
    v12 = array_get v5, index v1 -> [u1; 3]
    v13 = array_get v12, index v11 -> u1
    v14 = not v0
    enable_side_effects u1 1
    v15 = unchecked_mul v0, v13
    v16 = unchecked_mul v14, v10
    v17 = unchecked_add v15, v16
    return v17
}

--- Interpreter result after Array Set Optimizations (step 37):
Ok(u1 1)
---
Error: interpreter produced an unexpected result.
Expected result: u1 0
Actual result:   u1 1
```

The test passes with regular nargo execution. Running `nargo interpret --force-brillig` will also pass as expected.

### To Reproduce

1. Copy the test above 
2. Run `nargo interpret`. You can run with `--show-ssa` to see the exact output above

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

Blocker

### Blocker Context

Blocks these regression tests https://github.com/noir-lang/noir/pull/9789

### Nargo Version

_No response_

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #9830

This will unblock adding the tests in #9789.

## Summary\*

When fetching arrays in ACIR we treat them as entirely fresh objects. However, in the interpreter we implement arrays use shared mutable references. Thus, when we returned a nested array and then marked it mutable in the array set optimization we were in fact mutating the parent array we fetched from as well. This breaks the semantics of arrays in ACIR.

Looking at the SSA after the array set optimization:
```
  b0(v0: u1, v1: u32):
    v4 = make_array [u1 1, u1 0, u1 1] : [u1; 3]
    v5 = make_array [v4] : [[u1; 3]; 1]
    v7 = add v1, u32 1
    v8 = array_get v5, index v1 -> [u1; 3]
    v9 = array_set mut v8, index u32 1, value u1 1
    v10 = array_get v9, index v7 -> u1
    enable_side_effects v0
    v11 = add v1, u32 1
    v12 = array_get v5, index v1 -> [u1; 3]
    v13 = array_get v12, index v11 -> u1
    v14 = not v0
    enable_side_effects u1 1
    v15 = unchecked_mul v0, v13
    v16 = unchecked_mul v14, v10
    v17 = unchecked_add v15, v16
    return v17
}
```
We expect `v8` to be its own fresh array. I changed the interpreter to check if we are returned an array from an array get and if we are to clone its elements and construct a new shared mutable reference to its elements. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
@@ -928,7 +928,23 @@ impl<'ssa, W: Write> Interpreter<'ssa, W> {
             let element = elements.get(index as usize).ok_or_else(|| {
                 InterpreterError::IndexOutOfBounds { index, length: elements.len() as u32 }
             })?;
-            element.clone()
+
+            // Either return a fresh nested array (in constrained context) or just clone the element.
+            element
+                .as_array_or_slice()
+                .filter(|_| !self.in_unconstrained_context())
+                .map(|array| {
+                    // In the ACIR runtime we expect fresh arrays when accessing a nested array.
+                    // If we do not clone the elements here a mutable array set afterwards could mutate
+                    // not just this returned array but the array we are fetching from in this array get.
+                    Value::ArrayOrSlice(ArrayValue {
+                        elements: Shared::new(array.elements.borrow().to_vec()),
+                        rc: array.rc,
+                        element_types: array.element_types,
+                        is_slice: array.is_slice,
+                    })
+                })
+                .unwrap_or_else(|| element.clone())
         };
         self.define(result, element)?;
         Ok(())
--- a/compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
+++ b/compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
@@ -930,10 +930,8 @@ impl<'ssa, W: Write> Interpreter<'ssa, W> {
             })?;
 
             // Either return a fresh nested array (in constrained context) or just clone the element.
-            element
-                .as_array_or_slice()
-                .filter(|_| !self.in_unconstrained_context())
-                .map(|array| {
+            if !self.in_unconstrained_context() {
+                if let Some(array) = element.as_array_or_slice() {
                     // In the ACIR runtime we expect fresh arrays when accessing a nested array.
                     // If we do not clone the elements here a mutable array set afterwards could mutate
                     // not just this returned array but the array we are fetching from in this array get.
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/interpreter/mod.rs
```diff
@@ -928,7 +928,23 @@ impl<'ssa, W: Write> Interpreter<'ssa, W> {
             let element = elements.get(index as usize).ok_or_else(|| {
                 InterpreterError::IndexOutOfBounds { index, length: elements.len() as u32 }
             })?;
-            element.clone()
+
+            // Either return a fresh nested array (in constrained context) or just clone the element.
+            element
+                .as_array_or_slice()
+                .filter(|_| !self.in_unconstrained_context())
+                .map(|array| {
```
jfecher: I feel like using `.filter` here is being a bit clever in a bad way since the intent isn't actually filtering the element. Can you rewrite this with an if instead?

vezenovm: done

vezenovm: I was trying to avoid two `element.clone()` expressions. I still could, but it felt roundabout so I ultimately submitted to calling `element.clone()` twice.

jfecher (approved):

PR #9832, opened 2025-09-11
chore: Switch to im-rc where possible

jfecher:
# Description

## Summary\*

Just something I noticed. `im` provides a faster, single-threaded version `im-rc` so I switched to it where possible (most places) to hopefully get a speedup.

## Additional Context

The only remaining user of the thread-safe im is `Instruction::MakeArray` since we do `ssa.functions.par_iter()` in a few places, and functions hold instructions which thus need to be thread-safe.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/noir_stdlib/src/collections/vec.nr
+++ b/noir_stdlib/src/collections/vec.nr
@@ -64,34 +64,299 @@ impl<T> Vec<T> {
     }
 }
 
-mod tests {
-    use crate::collections::vec::Vec;
+mod vec_tests {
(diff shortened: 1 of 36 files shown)


jfecher:
test_report_zkpassport_noir-ecdsa_ looks to be the only speedup observed, the others were short-running tests all within a margin of error.

There were also some timings which took longer on this PR which I'm not sure the reason for given this should be a straightforward speed increase.

PR #9833, opened 2025-09-11
chore: greenlight remove enable side effects

### LINKED ISSUE #9527: Greenlight `remove_enable_side_effects` for audits
Prepare https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs for audits:

  - [ ] No big brain needed to read the code
  - [ ] Purpose of the code
  - [ ] Design decisions made

**Target Date:** 12 Sep 2025

asterite:
# Description

## Problem

Resolves #9527

## Summary

I added doc comments and tests.

One of the tests checks that if we have `enable_side_effects u1 1` then it's inserted right away as opposed to before an instruction affected by the side-effects vars... but I don't know why this is done. Wouldn't it be better to always insert these right before the instructions they affect?

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs
@@ -65,7 +65,6 @@ impl Function {
             }
             previous_block = Some(context.block_id);
 
-            let instruction_id = context.instruction_id;
             let instruction = context.instruction();
 
             // If we run into another `Instruction::EnableSideEffectsIf` before encountering any
@@ -89,8 +88,8 @@ impl Function {
                 if condition_is_one {
                     last_side_effects_enabled_instruction = None;
                 } else {
+                    last_side_effects_enabled_instruction = Some(context.instruction_id);
                     context.remove_current_instruction();
-                    last_side_effects_enabled_instruction = Some(instruction_id);
                 }
                 return;
             }


compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs
```diff
@@ -33,13 +42,13 @@ impl Ssa {
...
-    pub(crate) fn remove_enable_side_effects(&mut self) {
-        if matches!(self.runtime(), RuntimeType::Brillig(_)) {
+    fn remove_enable_side_effects(&mut self) {
+        if self.runtime().is_brillig() {
             // Brillig functions do not make use of the `EnableSideEffects` instruction so are unaffected by this pass.
             return;
         }
 
-        // Make sure this optimization runs when there's only one block
+        // Check the precondition thatthis optimization runs when there's only one block
```
TomAFrench:
```suggestion
        // Check the precondition that this optimization runs when there's only one block
```

TomAFrench:
> One of the tests checks that if we have enable_side_effects u1 1 then it's inserted right away as opposed to before an instruction affected by the side-effects vars... but I don't know why this is done. Wouldn't it be better to always insert these right before the instructions they affect?

:shrug: it shouldn't really make any difference but I chose to do this so that as much of the program runs in an unconditional context as possible (despite whether the instructions have side effects or not)

compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs
```diff
@@ -1,23 +1,32 @@
...
+//!       [Instruction::EnableSideEffectsIf] is encountered.
 //!
+//! The pass will also remove redundant [Instruction::EnableSideEffectsIf] instructions,
+//! for example if two consecutive [Instruction::EnableSideEffectsIf] instructions have the same
+//! condition.
+//!
+//! This pass doesn't run in Brillig functions as [Instruction::EnableSideEffectsIf] is not allowed
+//! in Brillig functions.
+//!
+//! //! ## Preconditions:
```
TomAFrench:
```suggestion
//! ## Preconditions:
```

compiler/noirc_evaluator/src/ssa/opt/remove_enable_side_effects.rs
```diff
@@ -80,8 +88,8 @@ impl Function {
                 if condition_is_one {
                     last_side_effects_enabled_instruction = None;
                 } else {
-                    last_side_effects_enabled_instruction = Some(instruction_id);
                     context.remove_current_instruction();
+                    last_side_effects_enabled_instruction = Some(context.instruction_id);
```
vezenovm, line 91: I'm seeing various regressions and this is the only place where it looks like that could have been caused.

asterite: I just reverted that. Maybe the benchmarks are off...

asterite: Or maybe it was this, not sure. The benchmarks looks good now (they are gone).

PR #9835, opened 2025-09-12
feat: better check_u128_mul_overflow logic when an operand is constant

### LINKED ISSUE #9760: check_u128_mul_overflow could avoid checks in more cases
In `check_u128_mul_overflow` we'll trigger an overflow if both values are greater or equal to 2^64.

If either operand is less than 2^64 then it's guaranteed that the value won't overflow the Field element range. An overflow will still be triggered if the value overflows u128::max because ACIR checks that in all cases.

Then, if either operand is a constant we could avoid the overflow check if the value is less than 2^125. The reason is that 2^125 multiplied by a u128 would give a value less than 2^(125 + 128) = 2^253, which still fits in the range of Field.

asterite:
# Description

## Problem

Resolves #9760

## Summary

See the linked issue explanation and also the comments in this PR.

## Additional Context

I thought we could also do better in more cases. For example if the LHS is a constant that, when multiplied, could overflow, then we'd only need to check that the bits of RHS are less than `bits(Field) - bits(LHS)` instead of checking that it's less than `2^64`. However, the only thing that would change here is the constant we divide by: the check will still be there so the performance will be the same.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/check_u128_mul_overflow.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/check_u128_mul_overflow.rs
@@ -72,7 +72,7 @@ fn check_u128_mul_overflow(
     let lhs_value = dfg.get_numeric_constant(lhs);
     let rhs_value = dfg.get_numeric_constant(rhs);
 
-    // If we multiply a constant value 2^n by an uknown u128 value we get at most `2^(n+128) - 2`.
+    // If we multiply a constant value 2^n by an unknown u128 value we get at most `2^(n+128) - 2`.
     // If `n+128` does not overflow the maximum Field element value, there's no need to check for overflow.
     let max_const_value_that_does_not_overflow = 1_u128 << (FieldElement::max_num_bits() - 128);
     if lhs_value.is_some_and(|value| value.to_u128() < max_const_value_that_does_not_overflow)


compiler/noirc_evaluator/src/ssa/opt/check_u128_mul_overflow.rs
```diff
@@ -72,23 +72,17 @@ fn check_u128_mul_overflow(
...
-            return;
-        }
-    }
-
-    // Same goes for rhs
-    if let Some(value) = rhs_value {
-        if value.to_u128() < two_pow_64 {
-            return;
-        }
+    // If we multiply a constant value 2^n by an uknown u128 value we get at most `2^(n+128) - 2`.
```
TomAFrench:
```suggestion
    // If we multiply a constant value 2^n by an unknown u128 value we get at most `2^(n+128) - 2`.
```

TomAFrench:
Can you add a new test to show that we don't inject the extra constraints with a small constant argument?

asterite:
There are probably no circuit changes because we don't have test programs for this, but I had to use larger numbers in the tests to avoid snapshots changing (which became smaller, so...)

TomAFrench:
My understanding of this PR is that previously the below program would have extra instruction added to it but now it wouldn't. Is that incorrect?

```
acir(inline) fn main f0 {
  b0(v0: u128):
    v2 = mul v0, u128 2
    return
}
```

We don't have a test to check this behaviour afaict.

---

Ah, to be clear I'm referring to an SSA test rather than generating an actual circuit.

asterite:
> My understanding of this PR is that previously the below program would have extra instruction added to it but now it wouldn't. Is that incorrect?

Yes, though when multiplying by `u128 2` there wasn't an overflow check before (and there isn't one now either). The difference is that we'd insert an overflow check for values larger than 2^64, but now we'd only insert a check for values larger than 2^126.

I adjusted a couple of tests to show that for values just below 2^126 there's no overflow check, and for a value at exactly 2^126 there is a check.

TomAFrench:
Ahhhh, ok sorry I'm just not reading closely enough.

PR #9837, opened 2025-09-12
chore(ssa): Run purity analysis before preprocessing of functions

vezenovm:
# Description

## Problem\*

No issue, I just noticed that during preprocessing we run LICM however we do not have any purity information. We run purity analysis before LICM in our standard SSA pass flow for this reason.

## Summary\*

Add a purity analysis pass before preprocessing. 

In order for this to be safe we also had to unify recursive ACIR functions with `PureWithPredicates`. Similarly to Brillig, if we marked recursive functions as pure we could accidentally remove an infinite loop. This only came up after this change as purity is now being run before inlining and as we inline everything in ACIR, indefinite recursion (or in this case a recursion limit) is handled during inlining. 

If we have this program:
```
acir(inline) fn pure main f0 {
  b0():
    call f1()
    return
}
acir(inline) fn pure main f1 {
  b0():
    call f1()
    return
}
```
Running DIE before inlining (which is what occurs during `preprocess_fns`) would remove the infinite loop.

The benchmarks don't show many improvements. As there are not any regressions either though I think this is fine to merge as LICM will now have purity information it did not before. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/pure.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/pure.rs
@@ -382,7 +382,7 @@ mod test {
         assert_eq!(purities[&FunctionId::test_new(4)], Purity::PureWithPredicate);
         assert_eq!(purities[&FunctionId::test_new(5)], Purity::PureWithPredicate);
         assert_eq!(purities[&FunctionId::test_new(6)], Purity::Pure);
-        assert_eq!(purities[&FunctionId::test_new(7)], Purity::Pure);
+        assert_eq!(purities[&FunctionId::test_new(7)], Purity::PureWithPredicate);
 
         assert_ssa_snapshot!(ssa, @r"
         acir(inline) impure fn main f0 {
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/pure.rs
```diff
@@ -272,10 +269,10 @@ fn analyze_call_graph(
                     }
                 }
 
-                // Recursive Brillig functions cannot be fully pure (may loop indefinitely),
+                // Recursive functions cannot be fully pure (may recurse indefinitely),
                 // but we still treat them as PureWithPredicate for deduplication purposes.
-                // If this function is recursive and a Brillig function, mark it as PureWithPredicate.
-                if recursive_functions.contains(&func) && functions[&func].runtime().is_brillig() {
+                // If we were to mark recursive functions pure we may entirely eliminate an infinite loop.
+                if recursive_functions.contains(&func) {
```
vezenovm, line 275: @TomAFrench FYI this had to be done. As per the OP:

> In order for this to be safe we also had to unify recursive ACIR functions with PureWithPredicates. Similarly to Brillig, if we marked recursive functions as pure we could accidentally remove an infinite loop. This only came up after this change as purity is now being run before inlining and as we inline everything in ACIR, indefinite recursion (or in this case a recursion limit) is handled during inlining.

PR #9839, opened 2025-09-12
chore: use `w` prefix for ACIR witnesses

asterite:
# Description

## Problem

No issue, just might be easier to read.

## Summary

I also removed references to "indexes" when printing and parsing ACIR as they previously referred to `_5`, now `w5`, but we can consider these witnesses instead of witness indexes.

## Additional Context

One more thing I noticed is that when we parse a memory INIT opcode there's a `len` that's parsed but discarded. Two things:
- should we check that the len matches the parsed list of witnesses? Or...
- should we just not print/parse the len? It might be redundant at this point (or maybe it's helpful for debugging?)

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/acvm-repo/acir/src/parser/tests.rs
+++ b/acvm-repo/acir/src/parser/tests.rs
@@ -706,21 +706,21 @@ fn array_dynamic() {
 fn fold_basic() {
     let src = "
     func 0
-    current witness index : _2
-    private parameters indices : [_0]
-    public parameters indices : [_1]
-    return value indices : []
+    current witness: w2
+    private parameters: [w0]
+    public parameters: [w1]
+    return values: []
     CALL func 1: PREDICATE: EXPR [ 1 ]
-    inputs: [_0, _1], outputs: [_2]
+    inputs: [w0, w1], outputs: [w2]
     
     func 1
-    current witness index : _3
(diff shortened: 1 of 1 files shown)


vezenovm:
> One more thing I noticed is that when we parse a memory INIT opcode there's a `len` that's parsed but discarded. Two things:

`len` is discarded? I see it here https://github.com/noir-lang/noir/blob/00f5ece64d32fb2c602efd7525c0a950bb33f7e7/acvm-repo/acir/src/parser/tests.rs#L477. Do we not print it? 

> * should we just not print/parse the len? It might be redundant at this point (or maybe it's helpful for debugging?)

It does seem redundant, but I guess it could be useful for debugging how many elements do we have in this array without having to count the actual number of witnesses in the init list. I think it would be good to just check this len matches the actual number of witnesses.

asterite:
> I think it would be good to just check this len matches the actual number of witnesses.

Done!

vezenovm:
@asterite I resolved the merge conflicts post #9859. I am switching over the ACIR gen tests to use the ACIR parser and I wanted to get this PR in first.

asterite:
@vezenovm Thank you!

PR #9867, opened 2025-09-16
feat(cli): Visualize the Control Flow Graph

aakoshh:
# Description

## Problem\*

Resolves a follow up for #9826 

## Summary\*

Adds a `visualize` command to the SSA CLI which renders the CFG as a [Mermaid flowchart](https://mermaid.js.org/syntax/flowchart.html). 

### Example

Take this Noir program:
```rust
fn main(a: bool) {
    if a {
        for i in 0..5 {
            foo(i);
        }
    }
}

fn foo(i: u32) {
    if i < 3 {
        println(f"cold {i}");
    } else {
        println(f"hot {i}");
    }
}
```

Render it with Mermaid using the `--markdown` option to get a file we can preview in VS Code or include on Github:
```shell
cargo run -q -p nargo_cli -- compile --silence-warnings --show-ssa-pass Initial \
  | tail -n +2 \
  | cargo run -q -p noir_ssa_cli -- visualize --markdown \
  > test.md
```

Open `test.md` in VSCode in Preview mode (requires the [Mermaid Markdown extension](https://marketplace.visualstudio.com/items?itemName=bierner.markdown-mermaid)):

Screenshot 2025-09-16 at 12 27 15

Alternatively use the `--url-encode` option to construct a URL that can be used with the [Mermaid Live](https://mermaid.live) editor:

```bash
open https://mermaid.live/view#$( \
  cargo run -q -p nargo_cli -- compile --silence-warnings --show-ssa-pass Initial \
    | tail -n +2 \
    | cargo run -q -p noir_ssa_cli -- visualize --url-encode)
```

(The `tail` gets rid of the `After Initial SSA:` message printed by `nargo`, which the SSA parser doesn't handle. We could change this to a comment, if nobody relies on it being this particular way). 

In this example it opens [this URL](https://mermaid.live/view#base64:eyJjb2RlIjoiZmxvd2NoYXJ0IFRCXG4gICAgc3ViZ3JhcGggZjAgW21haW4gZjBdXG4gICAgICAgIGYwLmIwKChiMCkpXG4gICAgICAgIGYwLmIwIC0tPiBmMC5iMVxuICAgICAgICBmMC5iMCAtLT4gZjAuYjJcbiAgICAgICAgZjAuYjEoKGIxKSlcbiAgICAgICAgZjAuYjEgLS0+IGYwLmIzXG4gICAgICAgIGYwLmIzKChiMykpXG4gICAgICAgIGYwLmIzIC0tPiBmMC5iNFxuICAgICAgICBmMC5iMyAtLT4gZjAuYjVcbiAgICAgICAgZjAuYjUoKGI1KSlcbiAgICAgICAgZjAuYjUgLS0+IGYwLmIyXG4gICAgICAgIGYwLmIyKCgoYjIpKSlcbiAgICAgICAgZjAuYjQoKGI0KSlcbiAgICAgICAgZjAuYjQgLS0+IGYwLmIzXG4gICAgZW5kXG5cbiAgICBzdWJncmFwaCBmMSBbZm9vIGYxXVxuICAgICAgICBmMS5iMCgoYjApKVxuICAgICAgICBmMS5iMCAtLT4gZjEuYjFcbiAgICAgICAgZjEuYjAgLS0+IGYxLmIyXG4gICAgICAgIGYxLmIyKChiMikpXG4gICAgICAgIGYxLmIyIC0tPiBmMS5iM1xuICAgICAgICBmMS5iMSgoYjEpKVxuICAgICAgICBmMS5iMSAtLT4gZjEuYjNcbiAgICAgICAgZjEuYjMoKChiMykpKVxuICAgIGVuZFxuXG4gICAgc3ViZ3JhcGggZjIgW3ByaW50bG4gZjJdXG4gICAgICAgIGYyLmIwKCgoYjApKSlcbiAgICBlbmRcblxuICAgIHN1YmdyYXBoIGYzIFtwcmludF91bmNvbnN0cmFpbmVkIGYzXVxuICAgICAgICBmMy5iMCgoKGIwKSkpXG4gICAgZW5kXG5cbiAgICBmMC5iNCAtLi4tPiBmMVxuICAgIGYxLmIyIC0uLi0+IGYyXG4gICAgZjEuYjEgLS4uLT4gZjJcbiAgICBmMi5iMCAtLi4tPiBmM1xuIiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQifX0=). 

When working in the context of a `nargo` project, we can visualize an SSA pass using the following helper command:

```
just visualize-ssa-cfg Initial
```

Note that the SSA pass name must be unique for this to work.

## Additional Context

I added a `FunctionView` wrapper for `Function` so that we don't have to pollute its API with a random selection of `pub` functions, because most things are `pub(crate)` only.

## Documentation\*

Check one:

# PR Checklist\*


TomAFrench:
Nice! this is really cool and something I've been wishing for when working with some of the CFG related passes!

aakoshh:
I thought it would be cool to have options to highlight the Post-Dominance-Frontier of a block for example.
It would probably awkward to do that in the CLI, but imagine if we had an LSP for the SSA to support hand editing it, with some right click menu to help with that 🤓

jfecher:
> cargo run -q -p nargo_cli -- compile --silence-warnings --show-ssa-pass Initial \
  | tail -n +2 \
  | cargo run -q -p noir_ssa_cli -- visualize --markdown \
  > test.md

What is the purpose of the first two lines here? We're compiling the program first, then using tail for something, then using the visualize command? Can we use the visualize command on its own?

If not, lets add this to the `justfile` so team members won't need to reconstruct this command each time.

tooling/ssa_cli/README.md
```diff
@@ -0,0 +1,100 @@
+# SSA CLI
+
+The SSA CLI facilitates experimenting with SSA snippets outside the context of a Noir program. For example we can print the Initial SSA (or any itermediate step) of a Noir program from `nargo [compile|execute|interpret]` using the `--show-ssa-pass` option, and then use the SSA CLI to further `transform`, `interpret` or `visualize` it.
```
TomAFrench:
```suggestion
The SSA CLI facilitates experimenting with SSA snippets outside the context of a Noir program. For example we can print the Initial SSA (or any intermediate step) of a Noir program from `nargo [compile|execute|interpret]` using the `--show-ssa-pass` option, and then use the SSA CLI to further `transform`, `interpret` or `visualize` it.
```

aakoshh:
> What is the purpose of the first two lines here? 

Basically just getting rid of the `After Initial SSA:` line. 

> We're compiling the program first, 

Yes, we're using `nargo compile` to get the SSA of a Noir program. We could be copying the SSA from a unit test, or from the terminal instead if that was the use case.

> then using tail for something, 

The SSA CLI treats input as something that can be parsed. The SSA parser handles comments, but doesn't expect the `After ...` messages that `nargo` uses to print stuff. When the SSA CLI itself prints SSA for logging purposes, it prints the message as `After` as well, but in the output itself it appears as `// After`, so we can parse it back, while still being reminded what it was.

I didn't want to touch `nargo`, but I suppose we could just print `//` there as well. (Or filter `After` in the CLI, but I thought that's ugly, whereas the scripting can be a sin-bin).

> then using the visualize command? 

Yes, it reads the SSA from stdin in this case.

> Can we use the visualize command on its own?

You can use it on its own with `--source-path` if you have an SSA in a file, or `pbpaste` if you copied from a unit test. 

There is no `visualize` command in `nargo`, so "on its own" doesn't work that way. Ostensibly we could add it, but I'm not sure it meshes well. 

> If not, lets add this to the justfile so team members won't need to reconstruct this command each time.

I thought people would come up with their own scripting depending on what they find most useful (e.g. local vs live editor), but I'll have a look where it could go 👍

aakoshh:
Added a `just visualize <pass>` command, only handling the blue sky scenario of no errors and a unique SSA pass matched by name.

jfecher:
> Added a just visualize <pass> command, only handling the blue sky scenario of no errors and a unique SSA pass matched by name.

Can you rename it to `just visualize_cfg <pass>` or even `just visualize_ssa_cfg <pass>`? In the context of `noir_ssa_cli` it is more clear it refers to something in the ssa but for the project as a whole `visualize` alone sounds very broad.

compiler/noirc_evaluator/src/ssa/ir/function.rs
```diff
@@ -223,10 +226,69 @@ impl Function {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            RuntimeType::Acir(inline_type) => write!(f, "acir({inline_type})"),
+            RuntimeType::Brillig(inline_type) => write!(f, "brillig({inline_type})"),
+        }
+    }
+}
+
+/// Provide public access to certain aspects of a `Function` without bloating its API.
+pub struct FunctionView<'a>(&'a Function);
```
jfecher, line 256: Can you elaborate on why this was added? I thought it may have been since Function was only pub(crate), but it is pub as well. If we want these functions to be public anyway I think gating them behind an additional `.view()` method may make them more difficult to find.

aakoshh: Over time I had to add a few methods already: the ones you see in `FunctionView` are mainly used by the fuzzer. Since I wanted to use these from a different crate, `pub(crate)` didn't work. Yes, `Function` is `pub`, but the methods we need to use are `pub(crate)`. Another example is `BasicBlock`: it's `pub`, but all its methods are `pub(crate)`. Then, to find `Call` instructions, I'd have needed the `DataFlowGraph` as well, which is not `pub`. 

Rather than go one by one and make e.g. `reachable_blocks` public, and then `successors` public, or make everything `pub`, I added methods which are `pub` but either not directly on `Function` or expose a bit more, like `return_types`, which returns the `Type` rather than the ID, since the `dfg` is not accessible to do the resolution from another crate. 

I thought this was getting messy: I need the DFG, but don't want to make it public just for visualisation, so I added this `FunctionView` to expose just what is needed for my purposes without going all-in on publicising everything. 

It's not actually meant to be easy to find, that's why I didn't add any docs to `.view()`. It's just a back-door for a friend.

aakoshh: > If we want these functions to be public anyway

I don't know if we want to make them public, I just wanted to make an exception to make things accessible enough to be tested or visualised. I think that discussion can be had separately.

aakoshh: I'd be in favour of making read-only methods public, but then what about `ControlFlowGraph`, `PostOrder`, etc. If we make them public, should we consider modifying their APIs breaking changes for example? By clearly marking `FunctionView` as made for testing, I think it's clear that there are no guarantees around it. 

Happy to follow up with another PR that makes read-only methods `pub` and adds getters to things like `Function::dfg` if we all agree there is no harm in it.

TomAFrench: > If we make them public, should we consider modifying their APIs breaking changes for example? By clearly marking FunctionView as made for testing, I think it's clear that there are no guarantees around it.

I think that even post 1.0 we can have the stability guarantee not extend to the compiler internals exposed by the rust crates.

tooling/ssa_cli/src/cli/mod.rs
```diff
@@ -46,9 +48,11 @@ enum SsaCommand {
     /// List the SSA passes we can apply.
     List,
     /// Parse and (optionally) validate the SSA.
+    /// Prints the normalized SSA, with canonic ID assignment.
```
TomAFrench:
```suggestion
    /// Prints the normalized SSA, with canonical ID assignment.
```

compiler/noirc_evaluator/src/ssa/ir/function.rs
aakoshh: Do you mean we should make things like `DataFlowGraph`, `ControlFlowGraph` and `PostOrder` public, but document that they are not considered part of a public API?

TomAFrench: I'm good with having this wrapper exposed in order to minimise the interface though. Easier for us to open things up later than restrict it.

TomAFrench: Ah, sorry I just saw your reply.

In an ideal world, it would be nice to avoid having more than we need marked `pub` as it makes it easier for us to nuke functionally dead code which wouldn't be found automatically if we made it `pub` - something like `FunctionView` helps with this. I was just saying that we shouldn't think of marking something `pub` as irreversible as until we explicitly say we're going to support an interface as stable then nothing is.

PR #9869, opened 2025-09-16
feat: LSP lightweight mode

asterite:
# Description

## Problem

To address the LSP slowness in big projects.

## Summary

There are a bunch of LSP features that are triggered as soon as the user starts typing, or when they stop typing:
- code lens
- code actions
- autocompletions
- inlay hints
- signature help

When they happen, the current way our LSP works is by type-checking the current project. This needs to be done because if you type `let x = 1;` and don't save, we's till want to show inlay hints on that `x`, and to know that we need type-checking. The same goes for all of the above requests. And these are the ones that make LSP slow in big projects.

Other features only happen when the user actively requests them: go to definition, hover. Well, there's also "document symbol" which shows the document's outline, but for that there's no need to type-check (the first commit addresses this).

This PR then adds some settings to disable these features, and also adds an `enableLightweightMode` setting to disable them all with just one setting.

The VSCode side is in https://github.com/noir-lang/vscode-noir/pull/111

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/tooling/lsp/src/requests/mod.rs
+++ b/tooling/lsp/src/requests/mod.rs
@@ -275,11 +275,8 @@ pub(crate) fn on_initialize(
                 text_document_sync: Some(TextDocumentSyncCapability::Kind(
                     TextDocumentSyncKind::FULL,
                 )),
-                code_lens_provider: if enable_code_lens {
-                    Some(CodeLensOptions { resolve_provider: Some(false) })
-                } else {
-                    None
-                },
+                code_lens_provider: enable_code_lens
+                    .then_some(CodeLensOptions { resolve_provider: Some(false) }),
                 document_formatting_provider: true,
                 nargo: Some(NargoCapability {
                     tests: Some(NargoTestsOptions {
@@ -307,16 +304,14 @@ pub(crate) fn on_initialize(
                         work_done_progress: None,
                     },
                 })),
-                inlay_hint_provider: if enable_inlay_hints {
-                    Some(lsp_types::OneOf::Right(lsp_types::InlayHintOptions {
+                inlay_hint_provider: enable_inlay_hints.then_some(lsp_types::OneOf::Right(
+                    lsp_types::InlayHintOptions {
                         work_done_progress_options: WorkDoneProgressOptions {
                             work_done_progress: None,
                         },
                         resolve_provider: None,
-                    }))
-                } else {
-                    None
-                },
+                    },
+                )),
                 document_symbol_provider: Some(lsp_types::OneOf::Right(
                     lsp_types::DocumentSymbolOptions {
                         work_done_progress_options: WorkDoneProgressOptions {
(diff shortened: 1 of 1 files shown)


tooling/lsp/src/requests/mod.rs
```diff
@@ -217,29 +258,36 @@ pub(crate) fn on_initialize(
...
-                text_document_sync: Some(text_document_sync),
-                code_lens_provider: code_lens,
+                text_document_sync: Some(TextDocumentSyncCapability::Kind(
+                    TextDocumentSyncKind::FULL,
+                )),
+                code_lens_provider: if enable_code_lens {
+                    Some(CodeLensOptions { resolve_provider: Some(false) })
+                } else {
+                    None
+                },
```
jfecher:
```suggestion
                code_lens_provider: enable_code_lens.then_some(
                    CodeLensOptions { resolve_provider: Some(false) }
                ),
```
:nerd_face:

tooling/lsp/src/requests/mod.rs
```diff
@@ -259,12 +307,16 @@ pub(crate) fn on_initialize(
...
+                inlay_hint_provider: if enable_inlay_hints {
+                    Some(lsp_types::OneOf::Right(lsp_types::InlayHintOptions {
+                        work_done_progress_options: WorkDoneProgressOptions {
+                            work_done_progress: None,
+                        },
+                        resolve_provider: None,
+                    }))
+                } else {
+                    None
+                },
```
jfecher:
```suggestion
                inlay_hint_provider: enable_inlay_hints.then_some(
                    lsp_types::OneOf::Right(lsp_types::InlayHintOptions {
                        work_done_progress_options: WorkDoneProgressOptions {
                            work_done_progress: None,
                        },
                        resolve_provider: None,
                    })
                ),
```

tooling/lsp/src/requests/mod.rs
```diff
@@ -273,35 +325,45 @@ pub(crate) fn on_initialize(
...
+                        ]),
+                        all_commit_characters: None,
+                        work_done_progress_options: WorkDoneProgressOptions {
+                            work_done_progress: None,
+                        },
+                        completion_item: None,
+                    }))
+                } else {
+                    None
+                },
```
jfecher:
```suggestion
                completion_provider: enable_completions.then(|| {
                    lsp_types::OneOf::Right(lsp_types::CompletionOptions {
                        resolve_provider: None,
                        trigger_characters: Some(vec![
                            ".".to_string(), // For method calls
                            ":".to_string(), // For paths
                            "$".to_string(), // For $var inside `quote { ... }`
                        ]),
                        all_commit_characters: None,
                        work_done_progress_options: WorkDoneProgressOptions {
                            work_done_progress: None,
                        },
                        completion_item: None,
                    })
                ),
```
etc

jfecher (approved):

tooling/lsp/src/requests/mod.rs
asterite: Aaah... I was sure there was a bool method for this but I thought it would end up looking bad, but it's actually better! I'll make this change.

asterite: Done!

tooling/lsp/src/requests/mod.rs
asterite: I ended up using `then_some` here too. It's just a one-time thing and it was running before so it's probably fine.

Issue #9861, opened 2025-09-15
`fuzz_with_arbtest` failing with `index out of bounds: the len is 28 but the index is 39`


michaeljklein:
### Aim

I ran the `fuzz_with_arbtest` test as follows:

```bash
❯ NOIR_AST_FUZZER_SEED=0x25e99f7d00100000 cargo test -p noir_ast_fuzzer_fuzz
```

at [this commit](https://github.com/noir-lang/noir/commit/d88b3564938fc7ce2679dbad15431f18f7cf3cdf)

### Expected Behavior

I expected this test to succeed

### Bug

These fuzz targets:

- `acir_vs_brillig`
- `pass_vs_prev`
- `min_vs_full`
- `orig_vs_morph`

failed with:

```bash
panicked at compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs:454:46:
index out of bounds: the len is 28 but the index is 39
```

### To Reproduce

1.
2.
3.
4.

### Workaround

None

### Project Impact

Nice-to-have

### Would you like to submit a PR for this Issue?

None

aakoshh:
On this one it looks like the way the AST is generated by the fuzzer causes the problem, but copying the Noir-like AST and running it through `nargo compile` makes it disappear. 

Using the following command I could print the SSA that fails in the test:

```
NOIR_AST_FUZZER_SHOW_SSA=1 NOIR_AST_FUZZER_SHOW_AST=1 NOIR_AST_FUZZER_SEED=0x25e99f7d00100000 cargo test -p noir_ast_fuzzer_fuzz acir_vs_brillig
```


This is the SSA before the panic in `mem2reg`:
```
g0 = u8 89
g1 = u8 85
g2 = make_array b"YU"
g3 = u8 79
g4 = u8 86
g5 = make_array b"OV"
g6 = u8 78
g7 = make_array b"ON"
g8 = u8 75
g9 = u8 67
g10 = make_array b"KC"
g11 = make_array [g2, g5, g7, g10] : [[u8; 2]; 4]
g12 = u8 71
g13 = u8 82
g14 = make_array b"GR"
g15 = u8 87
g16 = u8 76
g17 = make_array b"WL"
g18 = u8 84
g19 = u8 73
g20 = make_array b"TI"
g21 = u8 68
g22 = make_array b"VD"
g23 = make_array [g14, g17, g20, g22] : [[u8; 2]; 4]
g24 = u8 83
g25 = make_array b"TS"
[...]

PR #9882, opened 2025-09-16
fix(mem2reg): Do not attempt to analyze an instruction simplified to a global

### LINKED ISSUE #9880: mem2reg: Bad access of global instruction
### Aim

Run the following:
```
NOIR_AST_FUZZER_SEED=0xe6851ab700100000 NOIR_AST_FUZZER_SHOW_AST=1 RUST_BACKTRACE=1 cargo test -p noir_ast_fuzzer_fuzz pass_vs_prev
```

### Expected Behavior

For the test to succeed. 

### Bug

We will panic with `index out of bounds: the len is 23 but the index is 37` at https://github.com/noir-lang/noir/blob/0eead18ffe3373b74853cd06bacc12b4d9ff42f8/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs#L454 

This usually occurs due to a mismatch in the instruction IDs of the globals graph and the regular function graph. We most likely just want to skip analyzing when we see global values (which is fine as we do not allow references in globals).  

### To Reproduce

1. Run the command above

### Workaround

Yes

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.12 noirc version = 1.0.0-beta.12+1b4e25231fd0518d75e794889b4dc5a68e1ebf46 (git version hash: 1b4e25231fd0518d75e794889b4dc5a68e1ebf46, is dirty: false)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

vezenovm:
# Description

## Problem\*

Resolves #9880 
Resolves https://github.com/noir-lang/noir/issues/9861

## Summary\*

We were panicking here [noir/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs](https://github.com/noir-lang/noir/blob/0eead18ffe3373b74853cd06bacc12b4d9ff42f8/compiler/noirc_evaluator/src/ssa/opt/mem2reg.rs#L454)

An index OOB when accessing an instruction on the DFG directly is usually caused by attempting to a mismatch between the global and function graphs and using the ID from one in the other.

Instead of using `get_local_or_global_instruction` I chose to instead just check whether we have simplified to a global value. We know globals cannot have referencing so we do not need to go through this analysis. We also use the `InstructionId` to fetch results when analyzing instructions so I felt it was easier to simply skip global handling rather than try and make `analyze_possibly_simplified_instruction` check for whether we have a global.

## Additional Context

For some reason I am not getting an AST from the fuzzer. I ran the following:
```
NOIR_AST_FUZZER_SEED=0xe6851ab700100000 NOIR_AST_FUZZER_SHOW_AST=1 RUST_BACKTRACE=1 cargo test -p noir_ast_fuzzer_fuzz pass_vs_prev 
``` 
But I am not getting an AST, so I was unable to produce a simplified regression. Running locally, the seed above passes on this PR but not on master. 

## Documentation\*

Check one:

# PR Checklist\*


aakoshh:
> For some reason I am not getting an AST from the fuzzer.

Sorry about that, the printing of the AST could use some overhaul. The `NOIR_AST_FUZZER_SHOW_AST` is used by `create_ssa_or_die` to print the AST just before running the SSA pipeline on it, which happens during the generation of arbitrary test inputs. However that's not used for `pass_vs_prev`, where we generate the AST, then the initial SSA, then run some passes, but not the whole pipeline. For no apparent reason, the logging of the AST and the SSA happens during `exec` rather than the arbitrary generation, probably because this way it stayed together and in the generic part of the tester. Because SSA generation panicked, it never got to `exec`. Unlike the other cases, it also uses `log::debug!`, so to it needs to be enabled by `RUST_LOG=debug`. It would probably be better if all of them used that, rather than `SHOW_AST` variants.

aakoshh (approved):
LGTM!

vezenovm:
Now that we have the AST output I wanted to add a reduced regression test. I added a small unit test that would fail without the checks included in this PR.

PR #9888, opened 2025-09-17
fix(ssa): Start with checked operations in index calculations

### LINKED ISSUE #9856: SSA Interpreter: indexing with overflow succeeds
### Aim

Testing https://github.com/noir-lang/noir/issues/9853

The following program succeeds in the SSA interpreter, but fails with ACIR and Brillig
```rust
fn main() -> pub bool {
    let mut e: [(Field, bool); 1] = [((-1), true)];
    e[2147483648_u32].1
}
```

### Expected Behavior

The interpreter should fail with Index OOB.

### Bug

```console
❯ cargo run -q -p nargo_cli -- interpret --force --silence-warnings --show-ssa
After Initial SSA:
acir(inline) fn main f0 {
  b0():
    v2 = make_array [Field -1, u1 1] : [(Field, u1); 1]
    v3 = allocate -> &mut [(Field, u1); 1]
    store v2 at v3
    v4 = load v3 -> [(Field, u1); 1]
    v7 = unchecked_mul u32 2147483648, u32 2
    v8 = array_get v4, index v7 -> Field
    v10 = unchecked_add v7, u32 1
    v11 = array_get v4, index v10 -> u1
    return v11
}

Ok(u1 1)
---
...
--- Interpreter result after Dead Instruction Elimination (step 44):
Ok(u1 1)
---
After Verifying no dynamic array indices to reference value elements (step 45):
  b0():
    v2 = make_array [Field -1, u1 1] : [(Field, u1); 1]
    v5 = unchecked_mul u32 2147483648, u32 2
    v7 = unchecked_add v5, u32 1
    v8 = array_get v2, index v7 -> u1
    return v8
}

--- Interpreter result after Verifying no dynamic array indices to reference value elements (step 45):
Ok(u1 1)
---
```

By contrast in ACIR we get the following error:

```console
error: Index out of bounds, array has size 2, but index was 4294967297
  ┌─ src/main.nr:3:5
  │
3 │     e[2147483648_u32].1
  │     -----------------
  │
  = Call stack:
    1. src/main.nr:3:5

Failed to solve program: 'Index out of bounds, array has size 2, but index was 4294967297'
```

I assume the way ACIR implements `unchecked_add` is by using a `Field`, because e.g. if we try to index by `u32::MAX_VALUE` it complains that `array has size 2, but index was 8589934591`, whereas the SSA interpreter wraps around.

I'm not sure if `unchecked_add` in ACIR should not wrap around instead as well. It looks like [Brillig does wrap around](https://github.com/noir-lang/noir/blob/158470fdd93b21dedcebbb9ab4ceec11f0adb087/acvm-repo/brillig_vm/src/arithmetic.rs#L324).

### To Reproduce

See above. 

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

### Nargo Version

nargo version = 1.0.0-beta.12 noirc version = 1.0.0-beta.12+affa9727ac36f708b4e59db5c7c18c817446acb2 (git version hash: affa972, is dirty: true)

### NoirJS Version

_No response_

_No response_

None

### Support Needs

_No response_

aakoshh:
# Description

## Problem\*

Resolves #9856

## Summary\*

Implements @vezenovm 's idea from https://github.com/noir-lang/noir/issues/9856#issuecomment-3293206729

When generating SSA for indexing arrays, use `unchecked: false` by default and let the `simplify` machinery promote it to `unchecked: true` if we know that it's safe. This prevents such SSA from being generated that 1) SSA interpreter wraps around, 2) ACIR overflows into u64 and beyond and 3) Brillig wraps around but it's protected by constraints.

This had some knock-on effects: now that we use `mul` instead of `unchecked_mul` to calculate an index, it can fail, and when we see that it will, `remove_unreachable_indexes` replaces the failing instruction with a default 0. It also does this if e.g. the `add <prev-var>, u32 1` follows a failing `mul`: since the `add` requires a predicate, it is replaced by 0. 

This has the unfortunate potential for causing type-errors with `array_get` where the array contains heterogenous data: if e.g. every `index n * 4 + 3` contains a `u1`, but we replace that with `index 0`, that could be some different type. The result is somehow the SSA value that is assigned to the result will have the actual type from `index 0`, and then fail when it's used e.g. in a subsequent `binary` operation, as it doesn't match the other side.

I solved this by introducing a `should_replace_instruction_with_defaults` function, which verifies that a `result = array_get arr, index 0 -> type` checks out, ie. that `typeof(result) == typeof(array)[0]`. 

Another change I made to `replace_unreachable_instructions` was that it is now removing a failing binary instruction, leaving only the constraint. This is so that it's in line with the handling of `array_get` and slice operations. (This wasn't the full motivation, I wanted to remove and replace the binary with a default as well, but it turned out that doing so still did not allow me to detect where the result was used later on, because the `simple_optimization` already substitutes values, so what I get is just a value ID of the constant zero, not where it came from).

## Additional Context

I added a test which expects that the SSA interpreter should fail on it, but there is no such integration test group. I'll create a separate issue for that: https://github.com/noir-lang/noir/issues/9908

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/ssa_gen/mod.rs
```diff
@@ -631,7 +634,7 @@ impl FunctionContext<'_> {
 
         let result = self.codegen_expression(&for_expr.block);
         self.codegen_unless_break_or_continue(result, |this, _| {
-            let new_loop_index = this.make_offset(loop_index, 1);
+            let new_loop_index = this.make_offset(loop_index, 1, true);
```
aakoshh, line 638: Using unchecked here because there should be a legit upper bound.

aakoshh:
Here's a minimised version of on of the failing tests:
```rust
fn main() {
    let b = unsafe { func_3() };
    if b {
        let c = unsafe { func_4() }[1532016929_u32].3;
        if c {
            let d = unsafe { func_3() };
        }
    }
}

unconstrained fn func_3() -> bool {
    true
}
unconstrained fn func_4() -> [(str<3>, str<3>, bool, bool); 4] {
    [
        ("ABC", "CDE", false, true),
        ("FGH", "IJK", false, true),
        ("LMN", "OPQ", false, true),
        ("RST", "UVW", false, true),
    ]
}
```

```
❯ cargo run -q -p nargo_cli -- compile --force --silence-warnings --show-ssa-pass "step 41" --show-ssa-pass "step 42"

After Checked to unchecked (1) (step 41):
  b0():
    v1 = call f1() -> u1
    enable_side_effects v1
    v3 = call f2() -> [([u8; 3], [u8; 3], u1, u1); 4]
    v6 = mul u32 1532016929, u32 4
    constrain u1 0 == v1, "attempt to multiply with overflow"
    v9 = array_get v3, index u32 0 -> u1
    v10 = unchecked_mul v1, v9
    enable_side_effects v10
    v11 = call f1() -> u1
    enable_side_effects u1 1
    return
}
brillig(inline) predicate_pure fn func_3 f1 {
  b0():
    return u1 1
}
brillig(inline) predicate_pure fn func_4 f2 {
  b0():
    v3 = make_array b"ABC"
    v6 = make_array b"CDE"
    v10 = make_array b"FGH"
    v14 = make_array b"IJK"
    v18 = make_array b"LMN"
    v22 = make_array b"OPQ"
    v26 = make_array b"RST"
    v30 = make_array b"UVW"
    v33 = make_array [v3, v6, u1 0, u1 1, v10, v14, u1 0, u1 1, v18, v22, u1 0, u1 1, v26, v30, u1 0, u1 1] : [([u8; 3], [u8; 3], u1, u1); 4]
    return v33
}

Message:  Expected NumericType, found [u8; 3]
Location: compiler/noirc_evaluator/src/ssa/ir/types.rs:198
```
It seems to fail trying to simplify the `unchecked_mul`. 

The problem seems to be that `v9 = array_get v3, index u32 0 -> u1` is supposed to use `v9 = array_get v3, index v6 -> u1`, because at `index u32 0` there certainly is a `[u8; 3]` (a.k.a. `str<3>`), not a `u1`. 

Original failure came from 
```
NOIR_AST_FUZZER_SEED=0x05de72c100031cc9 cargo test -q -p noir_ast_fuzzer_fuzz orig_vs_morph
```

aakoshh:
Here's what goes down. Before `remove_unreachable_instructions` we have this:
```rust
...
    enable_side_effects v1
    v3 = call f2() -> [([u8; 3], [u8; 3], u1, u1); 4]
    v6 = mul u32 1532016929, u32 4
    v8 = add v6, u32 1
    v10 = add v6, u32 2
    v12 = add v6, u32 3
    v13 = array_get v3, index v12 -> u1
    v14 = unchecked_mul v1, v13
...
```
The `remove_unreachable_instructions` pass recognises that `v6` will always fail and replaces it with a constraint. Then it goes into `UnreachableUnderPredicate` mode and replaces subsequent side-effecting instructions with their default, which in this case means `v0`, `v10` and `v12` all become `0`. 

When it comes to inspect `array_get v3, index v12` it looks like `array_get v3, index u32 0` because of having replaced `v12` with 0, which is now a nonsensical instruction, but it looks safe and `requires_acir_gen_predicate` returns `false`, so the instruction is left in.

guipublic:
I think it will be simpler to always replace array-get with default in `should_replace_instruction_with_defaults()`, whatever the index is.

I am not convinced by the use of checked multiplication in SSA in order to solve an issue on the interpreter, but it seems that there are plans to properly fix the interpreter later on, so I'm fine with this.

aakoshh:
> I think it will be simpler to always replace array-get with default in should_replace_instruction_with_defaults(), whatever the index is.

I wanted to be specific just to know that I captured the essence of it, but I can't see any downside to widening the replacement to all array-gets. I would certainly make the code simpler.

vezenovm:
> I think it will be simpler to always replace array-get with default in `should_replace_instruction_with_defaults()`, whatever the index is.
> 
> I am not convinced by the use of checked multiplication in SSA in order to solve an issue on the interpreter, but it seems that there are plans to properly fix the interpreter later on, so I'm fine with this.

Agreed on both points. Let's just update `should_replace_instruction_with_defaults` and capture an issue for updating the interpreter before merging.

aakoshh:
Done. Here's the follow up ticket: https://github.com/noir-lang/noir/issues/9930

vezenovm:
> Done. Here's the follow up ticket: #9930

Interesting, generalizing the check caused some failures. Looking at https://github.com/noir-lang/noir/actions/runs/17862409749/job/50796809412?pr=9888 these look to all be related to different inline types (e.g., calls ACIR/Brillig entry points and no predicates).

aakoshh:
Okay, this is great, it gives me examples to demonstrate why it's not safe to replace everything with defaults when we go under `UnreachableUnderPredicate`. 

One of the failures is `regression_8236`, where the interpreter changes its mind after `remove_unreachable_instructions`. 

Here's is the SSA at various stages. 

After flattening:
```rust
  b0(v8: [u16; 3], v9: [u1; 1]):
    v11 = call f1() -> [u1; 1]
    v13 = array_get v11, index u32 0 -> u1
    enable_side_effects v13
    v14 = call f1() -> [u1; 1]
    v15 = array_get v9, index u32 0 -> u1
    v16 = not v13
    enable_side_effects v16
    v17 = array_get v8, index u32 0 -> u16
    v18 = cast v17 as u32
    v20 = mod v18, u32 3
    v21 = array_get v8, index v20 -> u16
    v22 = cast v21 as u32
    v24 = lt u32 24993, v22
    enable_side_effects u1 1
    v26 = unchecked_mul v13, v15
    v27 = unchecked_mul v16, v24
    v28 = unchecked_add v26, v27
    return v28
}
```

After EnableSideEffectsIf removal:
```rust
  b0(v8: [u16; 3], v9: [u1; 1]):
    v11 = call f1() -> [u1; 1]
    v13 = array_get v11, index u32 0 -> u1
    enable_side_effects v13
    v14 = call f1() -> [u1; 1]
    v15 = array_get v9, index u32 0 -> u1
    v16 = not v13
    v17 = array_get v8, index u32 0 -> u16
    v18 = cast v17 as u32
    enable_side_effects v16
    v20 = mod v18, u32 3
    v21 = array_get v8, index v20 -> u16
    v22 = cast v21 as u32
    v24 = lt u32 24993, v22
    enable_side_effects u1 1
    v26 = unchecked_mul v13, v15
    v27 = unchecked_mul v16, v24
    v28 = unchecked_add v26, v27
    return v28
}
```
Notice how the `enable_side_effects v16` sunk down: it jumped over `v17 = array_get v8, index u32 0 -> u16` because that did not need a predicate because the index is safe, and now `v17` is under `enable_side_effects v13` instead. 

After inlining brillig calls:
```rust
  b0(v8: [u16; 3], v9: [u1; 1]):
    v11 = make_array [u1 0] : [u1; 1]
    enable_side_effects u1 0
    v12 = make_array [u1 0] : [u1; 1]
    v14 = array_get v9, index u32 0 -> u1
    v15 = array_get v8, index u32 0 -> u16
    v16 = cast v15 as u32
    enable_side_effects u1 1
    v19 = mod v16, u32 3
    v20 = array_get v8, index v19 -> u16
    v21 = cast v20 as u32
    v23 = lt u32 24993, v21
    enable_side_effects u1 1
    return v23
}
```
Now we have constants for `enable_side_effects`, namely `enable_side_effects v13` has been replaced by `enable_side_effects u1 0`. 

At some point I added logic to `remove_unreachable_instructions` to say that if we are under `enable_side_effects u1 0` then it means we cannot have side effects, so we can go into `UnreachableUnderPredicate` mode, which means we can replace any instruction which could fail with its default value. This helped me get rid of some leftover `Load` on branches that can never happen. 

However if we expand this to replace all instructions, then it includes the `v15 = array_get v8, index u32 0 -> u16` as well, which would not be considered to have a side effect as it is safe, and the types check out as well. 

But if we remove that, then `v15` is 0 instead of whatever was in the `v0` array, and then `v16` is zero, and then even though we have `enable_side_effects u1 1`, it doesn't give the expected result because `v19 = mod v16, u32 3` now operates on 0. 

So: something that should not have been under `v13` bubbled up (or `v16` sunk down), then `v13` got disabled, `v16` enabled, but that does not mean that the instructions which were under `v13` were all meant to be disabled, only the ones that had potential side effects, which would not have migrated between side effect variables.

vezenovm:
> But if we remove that, then `v15` is 0 instead of whatever was in the `v0` array, and then `v16` is zero, and then even though we have `enable_side_effects u1 1`, it doesn't give the expected result because `v19 = mod v16, u32 3` now operates on 0.
> 
> So: something that should not have been under `v13` bubbled up (or `v16` sunk down), then `v13` got disabled, `v16` enabled, but that does not mean that the instructions which were under `v13` were all meant to be disabled, only the ones that had potential side effects, which would not have migrated between side effect variables.

Could we not potentially run into the same issue even when gating for a zero index? Is it fine because we assume we do not have mismatched SSA types in our SSA input? I guess this is a fine assumption if we validate our SSA for this. 

Makes sense, thank you for the detailed explanation. We handle something similar in ACIR gen. When we have an instruction which is not dependent on the predicate we use a predicate of one. This guarantees we get the value that was meant to be there rather than a default value https://github.com/noir-lang/noir/blob/44b6be785e16d20a4f3e3c472d37a50d83e5aee1/compiler/noirc_evaluator/src/acir/mod.rs#L422 
I see we have the same fall back inside of `should_replace_instruction_with_defaults`.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
```diff
@@ -190,29 +195,35 @@ impl Function {
...
-                        // We can replace side effecting instructions with defaults until the next predicate.
-                        Reachability::UnreachableUnderPredicate
-                    }
-                    _ => Reachability::Reachable,
-                };
+                current_block_reachability =
+                    if let Some(predicate) = context.dfg.get_numeric_constant(*condition) {
+                        // If side effects are turned off, we can replace side effecting instructions with defaults until the next predicate
+                        if predicate.is_zero() {
+                            Reachability::UnreachableUnderPredicate
```
vezenovm, line 202: I wonder if we could differentiate between a failure and a zero predicate (e.g., `UnreachableUnderPredicateFailure` and `UnreachableUnderZeroPredicate`). In the case where we have a failure like in your unit test:
```
          b0(v0: u1):
            enable_side_effects v0
            v1 = make_array [u8 1, u8 2] : [u8; 2]
            v2 = make_array [v1, u1 0] : [([u8; 2], u1); 1]
            v3 = mul u32 4294967295, u32 2          // overflow
            v4 = add v3, u32 1                      // after overflow, replaced by default
            enable_side_effects u1 1                // end of side effects mode
            enable_side_effects v0                  // restore side effects to what we know will fail
            v5 = array_get v1, index v4 -> u1       // if v4 is replaced by default, the item at 0 is not a u1
            v6 = unchecked_mul v0, v5               // v5 is no longer a u1, but [u8; 2]
            enable_side_effects u1 1
            return
        }
``` 
In this case we know we are `UnreachableUnderPredicate` but we have a preceding failure under this predicate. 

While in `regression_8236` case we have the following SSA input to this pass:
```
  b0(v8: [u16; 3], v9: [u1; 1]):
    v11 = make_array [u1 0] : [u1; 1]
    enable_side_effects u1 0
    v12 = make_array [u1 0] : [u1; 1]
    v14 = array_get v9, index u32 0 -> u1
    v15 = array_get v8, index u32 0 -> u16
    v16 = cast v15 as u32
    enable_side_effects u1 1
    v19 = mod v16, u32 3
    v20 = array_get v8, index v19 -> u16
    v21 = cast v20 as u32
    v23 = lt u32 24993, v21
    enable_side_effects u1 1
    return v23
}
```
We are also in `UnreachableUnderPredicate` state, but we are not guaranteed to have a preceding failure. Instead we simply have side effects turned off. So as you outlined in https://github.com/noir-lang/noir/pull/9888#issuecomment-3312895653 it is unsafe to replace instructions not dependent on the predicate with defaults in these situations and thus we should not replace them. But in the case of knowing we have a preceding failure under that predicate it should be safe.

vezenovm: We could then have a generalized "yes replace all array gets with defaults". This could potentially even be extended to all instructions as we know we have a preceding failure under that predicate.

aakoshh:
> Could we not potentially run into the same issue even when gating for a zero index? 

I was thinking about this yesterday: when we see an ArrayGet with index 0, can we be sure that it returns the right data, considering that it could have migrated between side effect variables? I think the answer is yes, because if we had a non-empty array, then index 0 was safe to begin with and we had no reason to tamper with it, so there should not be any inconsistency in the types. If it turns out that we did have to set the index to a default value, but the types still check out, then it's okay to leave it as it is, because it will read data that fits the shape, but most likely the circuit will fail before this comes to pass.

> Is it fine because we assume we do not have mismatched SSA types in our SSA input?

Yes, the assumption is that this type inconsistency can only be the product of this pass, and that it should be a temporary phenomenon that we must eliminate by the end of it. 

One thing I wondered about was that the elimination is gated by whether we are in the `UnreachableUnderPredicate` mode, and effect variables stack up, but the reset mechanism treats them as independent. So, could we be in a situation like the following:
```rust
enable_side_effects v1
constraint v1 == 0                 // This activates the UnreachableUnderPreciate mode
v3 = <load-index-from-memory>      // Some side effecting operation, replaced by default, so v3 will be 0
enable_side_effects (v1*v2)        // If `v1` was off, this will be off, yet we are back to being `Reachable`
v4 = array_get v0, index v3 -> u1  // the data at 0 might not be a u1 any more, and we are not removing because we are Reachable
```

The reason I thought this will not pose a problem in practice, is that where we get in trouble is with composite types, such as `[(T1, T2); 3]`, indexed as `a[i].j`, where the `index` comes `size * i + j`, and the problem is where `j` disappears, and one could think that we could have zeroed out the index under the influence of v1 and then executed the read under a different effect, but based on my experiments it looks like the index+offset operation never detaches from getting the array, since if the index was assigned separately to a variable, then the array would not be safely indexed, and it's actually the array-indexing plus tuple access operation together that triggers the index+offset code generation, so, as long as we use `mul`, the side effect variables will not jump over them.

aakoshh:
>  When we have an instruction which is not dependent on the predicate we use a predicate of one.

Something similar to the `one` trick happens in [simple_optimization](https://github.com/noir-lang/noir/blob/44b6be785e16d20a4f3e3c472d37a50d83e5aee1/compiler/noirc_evaluator/src/ssa/opt/simple_optimization.rs#L65C17-L66) as well.

compiler/noirc_evaluator/src/ssa/opt/remove_unreachable_instructions.rs
aakoshh: This sounds reasonable to me. Should we try it in this PR, or a follow up?

vezenovm: At this point we have diverged quite a bit from the goal of this PR which was to "Start with checked operations in index calculations" 😅. As the proposed change alters `remove_unreachable_instructions` quite significantly (and that pass is currently being internally audited) let's make it a separate PR.

aakoshh: https://github.com/noir-lang/noir/issues/9945

PR #9894, opened 2025-09-17
fix(acir_gen): Do not call into AcirValue::flatten during AcirValue::flat_numeric_types

vezenovm:
# Description

## Problem\*

Alternative to #9887

In https://github.com/noir-lang/noir/issues/9860 we look to be panicking due to attempting flatten an `AcirValue::Array` containing an `AcirValue::DynamicArray`. This panic is ultimately triggered by a call to [flat_numeric_types](https://github.com/noir-lang/noir/blob/a2ec2e5ac84afbb95f8ddf9ca6695d83eb2f574a/compiler/noirc_evaluator/src/acir/arrays.rs#L595). We should be able to safely write an array value that contains dynamic arrays (in fact this call to `flat_numeric_types` occurs after `array_set_value`). 

## Summary\*

I switched to iterating through each array element in `flat_numeric_types` rather than calling `AcirValue::flatten` directly. 

## Additional Context

These issues could be avoid with https://github.com/noir-lang/noir/issues/6231 as well as arrays will not be able to contain one another. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/acir/types.rs
+++ b/compiler/noirc_evaluator/src/acir/types.rs
@@ -147,6 +147,13 @@ impl AcirValue {
         }
     }
 
+    /// Fetch a flat list of ([AcirVar], [AcirType]).
+    ///
+    /// # Panics
+    /// If [AcirValue::DynamicArray] is supplied or an inner element of an [AcirValue::Array].
+    /// This is because an [AcirValue::DynamicArray] is simply a pointer to an array
+    /// and fetching its internal [AcirValue::Var] would require laying down opcodes to read its content.
+    /// This method should only be used where dynamic arrays are not a possible type.
     pub(super) fn flatten(self) -> Vec<(AcirVar, AcirType)> {
         match self {
             AcirValue::Var(var, typ) => vec![(var, typ)],
@@ -155,6 +162,9 @@ impl AcirValue {
         }
     }
 
+    /// Fetch a flat list of the [NumericType] contained within an array
+    /// An [AcirValue::DynamicArray] should already have a field representing
+    /// its types and should be supported here unlike [AcirValue::flatten]
     pub(super) fn flat_numeric_types(self) -> Vec<NumericType> {
         match self {
             AcirValue::Array(array) => {


compiler/noirc_evaluator/src/acir/types.rs
```diff
@@ -156,11 +157,11 @@ impl AcirValue {
 
     pub(super) fn flat_numeric_types(self) -> Vec<NumericType> {
         match self {
-            AcirValue::Array(_) => {
-                self.flatten().into_iter().map(|(_, typ)| typ.to_numeric_type()).collect()
+            AcirValue::Array(array) => {
+                array.into_iter().flat_map(|elem| elem.flat_numeric_types()).collect()
             }
             AcirValue::DynamicArray(AcirDynamicArray { value_types, .. }) => value_types,
-            _ => unreachable!("An AcirValue::Var cannot be used as an array value"),
```
aakoshh, line 163: @TomAFrench comments such as these is why I think it would be great to get a walkthrough of how ACIR arrays are supposed to work in general. cc @asterite

compiler/noirc_evaluator/src/acir/types.rs
```diff
@@ -156,11 +157,11 @@ impl AcirValue {
 
     pub(super) fn flat_numeric_types(self) -> Vec<NumericType> {
```
aakoshh, line 168: It would be nice to add some docs to this method, in particular why this can work for all `AcirValue` variants, but `flatten` does not, while they conceptually sound similar.

compiler/noirc_evaluator/src/acir/arrays.rs
```diff
@@ -226,13 +226,6 @@ impl Context<'_> {
         }
 
         if let Some(store_value) = store_value {
-            // We should avoid storing a DynamicArray directly inside an Array. Instead we must allow
-            // the regular `array_set_value` to take place which reads the item from the block the
-            // dynamic array refers to, and writes items one by one. Otherwise subsequent operations
-            // with non-constant indexes will fail.
-            if matches!(store_value, AcirValue::DynamicArray(_)) {
-                return Ok(false);
-            }
```
aakoshh, line 235: Just to confirm: the bailing out on the _read_ side cannot be solved in a similar way, right?

vezenovm: No not for the read. The read was failing due to attempting to re-initialize a dynamic array. This one is panicking due to a bad method for fetching the types of an array.

compiler/noirc_evaluator/src/acir/types.rs
vezenovm: Yeah they could use a revamp and improved documentation in general. I'll be going over the arrays soon for audit prep work.

compiler/noirc_evaluator/src/acir/types.rs
vezenovm: Added some comments

compiler/noirc_evaluator/src/acir/types.rs
TomAFrench: Agreed, I can maybe do some docs writing next week while doing the ignition review.

vezenovm:
Merging into the parent @aakoshh

PR #9895, opened 2025-09-17
chore(acir_gen): New shared_context module and some additional unit tests

vezenovm:
# Description

## Problem\*

Working towards #9795 

## Summary\*

`acir/mod.rs` is over 1800 lines. This can be difficult to navigate especially for newcomers. I have started the process of splitting up this file and documenting/testing as I go.

To start I have pulled out the `SharedContext` structure into its own module. I added some new module docs and docs to the methods.

I also included a couple unit tests to have some more sanity surrounding `SharedContext` ID resolution. This led me to add an extra sanity assertion in `SharedContext::insert_generated_brillig`. After writing the `override_brillig_function_pointer` I realized it was possible to override Brillig function pointers. We should not be overriding any Brillig pointers in ACIR gen as we have a [safety check for whether a pointer already exists for a given function ](https://github.com/noir-lang/noir/blob/a2ec2e5ac84afbb95f8ddf9ca6695d83eb2f574a/compiler/noirc_evaluator/src/acir/mod.rs#L775). However, it felt good to add this sanity check either way. I can switch to a debug assertion if we see that as better. 

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/acir/shared_context.rs
```diff
@@ -0,0 +1,331 @@
...
+        code: GeneratedBrillig<F>,
+    ) {
+        let key = (func_id, arguments);
+        if self.brillig_generated_func_pointers.contains_key(&key) {
+            panic!(
+                "Attempting to override Brillig pointer for function {} with arguments {:?} already exists",
+                func_id, key.1
+            );
+        }
+        self.brillig_generated_func_pointers.insert(key, generated_pointer);
```
asterite: I know this is just moving things around, but we could check the return value of `insert` instead of checking `contains_key` (just a minor thing, really)

compiler/noirc_evaluator/src/acir/shared_context.rs
```diff
@@ -0,0 +1,331 @@
...
+        } else {
+            let code = self.brillig_stdlib.get_code(*brillig_stdlib_func);
+            let generated_pointer = self.new_generated_pointer();
+            self.insert_generated_brillig_stdlib(
+                *brillig_stdlib_func,
+                generated_pointer,
+                func_id,
+                opcode_location,
+                code.clone(),
+            );
```
asterite: I wonder if this call could be inlined. It seems to be only done here. The reason is that when I look at `insert_generated_brillig_stdlib` in isolation there's no check that `brillig_stdlib_func_pointer` doesn't contain that entry. That check is done here, but it's hard to immediately tell. I also thought a call to `add_call_to_resolve` was missing here, but it's done in the helper function. Plus it's 3 lines of code so it might also end up shorter.

compiler/noirc_evaluator/src/acir/shared_context.rs
```diff
@@ -0,0 +1,331 @@
...
+        func_id: FunctionId,
+    ) -> Option<&Vec<(OpcodeLocation, BrilligFunctionId)>> {
+        self.brillig_stdlib_calls_to_resolve.get(&func_id)
+    }
+
+    /// Remove and return the set of globals used by the given function,
+    /// or an empty set if the function had no globals recorded.
+    ///
+    /// We remove as an entry point should only go through ACIR generation a single time.
+    pub(super) fn get_used_globals_set(&mut self, func_id: FunctionId) -> HashSet<ValueId> {
```
asterite: Maybe call this `get_and_remove_used_globals_set` for clarity?

vezenovm: renamed

compiler/noirc_evaluator/src/acir/shared_context.rs
vezenovm: This was done to avoid cloning the arguments to include them in the message. But to be honest I don't know if the arguments are super necessary in this panic message as it is indicative of an internal error that a team member will have to investigate further either way. I moved to checking the insert's return value.

compiler/noirc_evaluator/src/acir/shared_context.rs
vezenovm: I inlined the call. Like you said it is in fact less code and a bit clearer.

PR #9899, opened 2025-09-18
feat: allow initializing dynamic arrays

TomAFrench:
# Description

## Problem\*

Resolves 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/acir/acir_context/mod.rs
+++ b/compiler/noirc_evaluator/src/acir/acir_context/mod.rs
@@ -1575,9 +1575,6 @@ impl<F: AcirField> AcirContext<F> {
             }
             Some(optional_value) => {
                 let mut values = Vec::new();
-                if let AcirValue::DynamicArray(_) = optional_value {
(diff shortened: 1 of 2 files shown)


TomAFrench:
Looking at that ACIR, it's still suboptimal however as we're just reading all the elements out of 2 to create 3 and then writing the same value at the same index as has been done in 2.

compiler/noirc_evaluator/src/acir/acir_context/mod.rs
```diff
@@ -1575,9 +1575,6 @@ impl<F: AcirField> AcirContext<F> {
             }
             Some(optional_value) => {
                 let mut values = Vec::new();
-                if let AcirValue::DynamicArray(_) = optional_value {
```
vezenovm, line 1578: Hmm I guess this restriction could always have been removed. It was moved here https://github.com/noir-lang/noir/pull/5810, but after `initialize_array_inner` started supporting dynamic arrays we basically have the same `arrays::copy_dynamic_array` method.

PR #9906, opened 2025-09-18
chore: add more to/from le/be bits/bytes edge case tests

michaeljklein:
# Description

## Summary\*

Adding tests for:
- `from_be_bytes`, `to_be_bytes`
- `from_le_bytes`, `to_le_bytes`
- `from_be_bits`, `to_be_bits`
- `from_le_bits`, `to_le_bits`

involving:
- `modulus_be_bits`
-  `modulus_le_bits`
-  `modulus_be_bytes`
-  `modulus_le_bytes`

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/noir_stdlib/src/field/mod.nr
+++ b/noir_stdlib/src/field/mod.nr
@@ -339,7 +339,9 @@ fn lt_fallback(x: Field, y: Field) -> bool {
 
 mod tests {
     use crate::{panic::panic, runtime, static_assert};
-    use super::{field_less_than, modulus_be_bits, modulus_le_bits, modulus_be_bytes, modulus_le_bytes};
+    use super::{
+        field_less_than, modulus_be_bits, modulus_be_bytes, modulus_le_bits, modulus_le_bytes,
+    };
 
     #[test]
     // docs:start:to_be_bits_example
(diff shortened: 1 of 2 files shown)


noir_stdlib/src/field/mod.nr
```diff
@@ -573,4 +575,174 @@ mod tests {
...
+
+            let p_bytes: [u8; 32] = p.to_le_bytes();
+            assert_eq(p_bytes, [0; 32]);
+            assert(p_bytes.as_slice() != modulus_le_bytes());
+        }
+    }
+
+    /// Convert a little endian bit array to a field element.
+    /// If the provided bit array overflows the field modulus then the Field will silently wrap around.
+    pub fn from_le_bits<let N: u32>(bits: [u1; N]) -> Field {
```
asterite: Does this need to be `pub`?

noir_stdlib/src/field/mod.nr
```diff
@@ -573,4 +575,174 @@ mod tests {
...
+        for i in 0..N {
+            result += (bits[i] as Field) * v;
+            v = v * 2;
+        }
+        result
+    }
+
+    /// Convert a big endian bit array to a field element.
+    /// If the provided bit array overflows the field modulus then the Field will silently wrap around.
+    pub fn from_be_bits<let N: u32>(bits: [u1; N]) -> Field {
```
asterite: Does this need to be `pub`?

noir_stdlib/src/field/mod.nr
```diff
@@ -573,4 +575,174 @@ mod tests {
...
+
+            assert(q_bytes != p_plus_1_bytes);
+
+            assert_eq(modulus_le_bytes().len(), 32);
+            let p = Field::from_le_bytes::<32>(modulus_le_bytes().as_array());
+            assert_eq(p, 0);
+
+            let p_bytes: [u8; 32] = p.to_le_bytes();
+            assert_eq(p_bytes, [0; 32]);
+            assert(p_bytes.as_slice() != modulus_le_bytes());
```
asterite: It might be good to comment why this equality doesn't hold. At first glance I see `p` comes from `modules_le_bytes`, then `p` is converted back to bytes using `to_le_bytes` so I don't know why they aren't equal.

It might also be good to put comments on each chunk of this test explaining what we are testing as it seems multiple things are tested. I can kind of guess based on the var names, but comments would make it a bit clearer, maybe.

PR #9916, opened 2025-09-19
added pub to to_le_radix and to_be_radix

porco-rosso-j:
# Description

Proposing to add `pub` to `to_le_radix()` and `to_be_radix()` in Field impl

## Problem\*

These methods are useful but inaccessible from outside due to compile error `to_be_radix is private`. Let me know if this change doesn't make sense. 

details: https://github.com/AztecProtocol/aztec-packages/issues/17153

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/noir_stdlib/src/field/mod.nr
+++ b/noir_stdlib/src/field/mod.nr
@@ -155,7 +155,7 @@ impl Field {
         bytes
     }
 
-    fn to_le_radix<let N: u32>(self: Self, radix: u32) -> [u8; N] {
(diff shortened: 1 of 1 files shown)


TomAFrench:
Hey @porco-rosso-j, I'm afraid this was purposefully restricted as these methods are currently underconstrained (potential for field overflows) outside of the usage in byte/bit decompositions and also to limit the scope of the stdlib.

If you need an alternative decomposition then I'd recommend constructing it in an unconstrained function and then constraining this appropriately.

PR #9929, opened 2025-09-19
chore: remove_if_else docs and refactors

### LINKED ISSUE #9913: remove_if_else: only merging changed indices no longer happens
### Problem

In the `remove_if_else` pass there's a `try_merge_only_changed_indices` test that suggest that only changed indices are updated. However, all indices are currently updated as that logic was removed in https://github.com/noir-lang/noir/pull/8142

### Happy Case

The `try_merge_only_changed_indices` is either removed or renamed so that it's clear that all indices are changed.

Alternatively, we could try to [bring back this optimization](https://github.com/noir-lang/noir/issues/8145) although removing it seems to have led to performance improvements.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

### LINKED ISSUE #9914: remove_if_else: untested code, and could use some more docs
### Problem

The docs are really great! However, once you dive into the implementation a few more instructions, other than IfElse, are handled: `Call` and `ArraySet`. They seem to track some slice sizes, but this isn't documented at all. Then if we remove the `ArraySet` handling, all unit tests in the file pass.

### Happy Case

Tracking of slice sizes, how it works and why it's needed, should be documented. Then unit tests around this handling should be added.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9913
Resolves #9914

## Summary

As this isn't changing behavior and instead just adding some more docs and tests I wonder if this could be merge during the audit, mainly because some things weren't clear to me and now they are a bit clearer.

Each commit should be self-contained.

## Documentation

Check one:

# PR Checklist


compiler/noirc_evaluator/src/ssa/opt/remove_if_else.rs
```diff
@@ -1,5 +1,5 @@
-//! This file contains the SSA `remove_if_else` pass - a required pass for ACIR to remove any remaining
-//! `Instruction::IfElse` in the singular program-function, and replace them with
+//! This file contains the SSA `remove_if_else` pass - a required pass for ACIR to remove any
+//! remaining`Instruction::IfElse` in the singular program-function, and replace them with
```
aakoshh:
```suggestion
//! remaining `Instruction::IfElse` in the singular program-function, and replace them with
```

compiler/noirc_evaluator/src/ssa/opt/remove_if_else.rs
```diff
@@ -283,6 +298,11 @@ fn slice_capacity_change(
         }
 
         Intrinsic::SlicePopBack | Intrinsic::SliceRemove => {
+            // fn pop_back(self) -> (Self, T)
+            // fn remove(self, index: u32) -> (Self, T)
+            //
+            // These functions return the slice as the first result `(len, slice)`,
```
aakoshh:
```suggestion
            // These functions return the slice as the result `(len, slice, item)`,
```
It took me a long time to understand how `(Self, T)` becomes `(len, slice)`, it looked like they just swapped places.

TomAFrench (approved):
LGTM

PR #9940, opened 2025-09-22
chore: assumes no load-store in array_set

### LINKED ISSUE #9921: Remove Load handling in array_set optimisation
### Problem

The handling of Load instructions is incorrect.
The can_mutate does not check for the terminator if the array is from a Load
Furthermore, Load instructions that return arrays do not count as use, meaning we can mutate an array and Load the same one later on.
We can see this with the following example:
```
acir(inline) fn main f0 {
  b0():
    v1 = make_array [Field 0] : [Field; 1]
    v2 = allocate -> &mut [Field; 1]
    store v1 at v2
    v3 = load v2 -> [Field; 1]
    v6 = array_set mut v3, index u32 0, value Field 1
    v7 = load v2 -> [Field; 1]
    return v7
}
```

It returns 1, which means that it has been mutated, although it is returned and loaded after the array-set.

### Happy Case

Remove any logic related to Load instructions, and add in the pre-condition that Load and Store have been removed.
We should also consider moving this pass after the `verify_no_dynamic_indices_to_references` pass because of this new pre-condition.

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

guipublic:
# Description

## Problem\*

Resolves #9921

## Summary\*
Remove handling of Load instructions in array_set optimisation, and rather assumes that there is no such instructions.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/opt/array_set.rs
```diff
@@ -222,14 +218,11 @@ impl<'f> Context<'f> {
...
-                // Arrays loaded from input references might be shared with the caller.
-                Instruction::Load { address } => {
-                    let result = self.dfg.instruction_results(*instruction_id)[0];
-                    if self.dfg.type_of_value(result).is_array() {
-                        let is_reference_param =
-                            self.dfg.block_parameters(block_id).contains(address);
-                        self.arrays_from_load.insert(result, is_reference_param);
-                    }
+                // Arrays loaded from memory might reference an existing array
+                // For instance if the array comes from a load we may potentially being mutating an array
```
vezenovm:
```suggestion
                // For instance if the array comes from a load we may potentially be mutating an array
```

PR #9944, opened 2025-09-22
feat: inverted make_constrain_not_equal condition

### LINKED ISSUE #9943: make_constrain_not_equal could consider a flipped condition
### Problem

The `make_constrain_not_equal` will change this SSA:

```
v2 = eq v0, v1
constrain v2 == u1 0
```

to this:

```
v2 = eq v0, v1
constrain v0 != v1
```

That's because when you write `assert(x != y)` in Noir, the first SSA is generated.

However, the pass could also consider an SSA where the condition is flipped, that is, `constrain u1 0 == v2`. That doesn't directly result from Noir code but it could happen as a result of other optimizations.

### Happy Case

This SSA:

```
v2 = eq v0, v1
constrain u1 0 == v2
```

is also changed to this:

```
v2 = eq v0, v1
constrain v0 != v1
```

### Workaround

None

### Workaround Description

_No response_

### Additional Context

_No response_

### Project Impact

None

### Blocker Context

_No response_

None

### Support Needs

_No response_

asterite:
# Description

## Problem

Resolves #9943

## Additional Context

As with all optimizations, I get curious about whether they actually happen in our test programs.

## Documentation

Check one:

# PR Checklist

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/make_constrain_not_equal.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/make_constrain_not_equal.rs
@@ -58,6 +58,12 @@ impl Function {
         }
 
         self.simple_optimization(|context| {
+            let instruction = context.instruction();
+
+            let Instruction::Constrain(lhs, rhs, msg) = instruction else {
+                return;
+            };
+
             // This Noir code:
             //
             // ```noir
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/make_constrain_not_equal.rs
```diff
@@ -70,28 +76,43 @@ impl Function {
             // v0 = eq x, y
             // constrain v0 == u1 0
             // ```
-            let instruction = context.instruction();
+            if context.dfg.get_numeric_constant(*lhs).is_some_and(|constant| constant.is_zero()) {
```
aakoshh, line 79: This looks to be the opposite of what the SSA in the comment shows: `lhs` and `rhs` are swapped.

asterite: Ah, you are right.

In any case I think I'll close this PR as it doesn't seem to provide a big benefit...

asterite:
Closing as it's not a big win.

Issue #9218, opened 2025-07-16
Check whether we need to make updates to account for `ArrayOffset` in various SSA passes


TomAFrench:
We added offsets for indices for array accesses in https://github.com/noir-lang/noir/pull/8536 and  found it necessary to account for these in https://github.com/noir-lang/noir/pull/9203.

There's a number of locations in the compiler where we just completely ignore these however with no reasoning. Only a subset of the passes in the compiler run after `Ssa::brillig_array_get_and_set`:

https://github.com/noir-lang/noir/blob/d867e35628c265daf18bd8ceaa4e777dac51d67e/compiler/noirc_evaluator/src/ssa/mod.rs#L189-L218

This means that we have basically no testing that feeding in array operations with offsets won't break other passes. We should:

- Add tests which test the behaviour of the passes with array offsets set.
- add assertions that the offset is zero to any passes which are expected to run before `Ssa::brillig_array_get_and_set` only (and the above is infeasible).

cc @defkit, this might be something useful to add into the SSA fuzzer.

vezenovm:
> * add assertions that the offset is zero to any passes which are expected to run before `Ssa::brillig_array_get_and_set` only (and the above is infeasible).

This is one option and we could have some shared validation utility. However, this would require us to manually add it to all of our passes. This feels a bit hacky, but may be fine for some temporary protection. Perhaps we are ok with some pre and post checks just on `brillig_array_get_and_set`?

In the longer term, it may be better to simply always handle the offset field.

TomAFrench:
Yeah, it's a little hacky to inject something into every pass. We do already have a check on `brillig_array_get_and_set` that checks to make sure that no offsets are set already.

> In the longer term, it may be better to simply always handle the offset field.

I think that we need to have handling for this in instruction simplification + DIE at the very least as I think the only reason that these are correct atm is due to instructions which would cause problems already being optimized out at this point.

asterite:
Another option could be to change the representation of `ArrayGet` and `ArraySet`.

Right now we have this:

```rust
pub enum Instruction {
    ...
    ArrayGet { array: ValueId, index: ValueId, offset: ArrayOffset },
    ...
}

pub enum ArrayOffset {
    None,
    Array,
    Slice,
}
```

Instead we could have this:

```rust
pub enum Instruction {
    ...
    ArrayGet { array: ValueId, index: ArrayIndex },
    ...
}

// I didn't think about these names too much
pub enum ArrayIndex {
[...]

PR #8532, opened 2025-05-15
fix: do "Brillig Array Gets" optimization during brillig gen

asterite:
# Description

## Problem

Resolves #8462

## Summary

If this optimization is done as an SSA optimization then there's no way the SSA interpreter can run on the code with indexes having shifted.

## Additional Context

For this program:

```noir
unconstrained fn main(x: [Field; 3]) -> pub Field {
    x[0]
}
```

The old brillig was:

```
GlobalInit(Id(0)):
  RETURN
Function(Id(0), None):
  CALL Procedure(CheckMaxStackDepth)
Function(Id(0), Some(Id(0))):
  CONST S2 = 1
  S4 = S1 + S2
  LOAD S3 = *S4
  MOV S1, S3
  RETURN
```

The new brillig is:

```
GlobalInit(Id(0)):
  RETURN
Function(Id(0), None):
  CALL Procedure(CheckMaxStackDepth)
Function(Id(0), Some(Id(0))):
  CONST S2 = 0
  CONST S4 = 1
  S5 = S1 + S4
  LOAD S3 = *S5
  MOV S1, S3
  RETURN
```

I understand that now we need a new constant for the offset (`S4` now, previously `S2`) though I don't understand where the `CONST S2 = 0` comes from 🤔 

## Documentation

Check one:
- [x] No documentation needed.
- [ ] Documentation included in this PR.
- [ ] **[For Experimental Features]** Documentation to be submitted in a separate PR.

# PR Checklist

- [x] I have tested the changes locally.
- [x] I have formatted the changes with [Prettier](https://prettier.io/) and/or `cargo fmt` on default settings.

PR #9948, opened 2025-09-22
chore(ssa): Assert that array offset are not set

aakoshh:
# Description

## Problem\*

Part of auditing `brillig_array_get_and_set`. 

## Summary\*

Add assertions that `ArrayGet::offset` and `ArraySet::offset` are `None` in some operations where they weren't handled.

For example `Instruction::has_side_effects` uses `DataFlowGraph::is_safe_index` to test that a constant index is less than the length of the array, however it did not take into account that the `offset` could have made the index equal to the logical size. 

We could fix this by accounting for the offset and subtracting its value. This PR instead just panics, as this pass should not be followed up by anything that accesses these functions at the moment.

Removes `offset` handling from `remove_unreachable_instructions`: since it appeared under `runtime.is_acir()`, it should never be set anyway. 

## Additional Context

I am not exactly sure why this pass exists if I'm honest. I understand what it's doing, thanks to @asterite having added docs during the green lighting, however during Brillig codegen we [check if the instruction has offset](https://github.com/noir-lang/noir/blob/74b44a7a683ee8447cf00d39b864f1ad9c8a05f8/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs#L840-L846) and if the index is dynamic then we [have to generate offsets](https://github.com/noir-lang/noir/blob/74b44a7a683ee8447cf00d39b864f1ad9c8a05f8/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs#L1166-L1169) anyway, so why don't we insert the necessary constants into the DFG during codegen, and then avoid having to have the `offset: _` pattern throughout the codebase, and having to reason about whether using `index` without considering `offset` is correct or not? 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/dfg/simplify.rs
@@ -119,9 +119,10 @@ pub(crate) fn simplify(
             if matches!(array_or_slice_type, Type::Array(_, 1))
                 && array_or_slice_type.flattened_size() == 1
             {
+                assert_eq!(*offset, ArrayOffset::None);
                 // If the array is of length 1 then we know the only value which can be potentially read out of it.
                 // We can then simply assert that the index is equal to zero and return the array's contained value.
-                optimize_length_one_array_read(dfg, block, call_stack, *array, *index, *offset)
+                optimize_length_one_array_read(dfg, block, call_stack, *array, *index)
             } else {
                 None
             }
@@ -342,7 +343,6 @@ fn optimize_length_one_array_read(
     call_stack: CallStackId,
     array: ValueId,
     index: ValueId,
-    offset: ArrayOffset,
 ) -> SimplifyResult {
     let zero = dfg.make_constant(FieldElement::zero(), NumericType::length_type());
     let index_constraint = Instruction::Constrain(
@@ -357,7 +357,7 @@ fn optimize_length_one_array_read(
         SimplifyResult::SimplifiedToInstruction(Instruction::ArrayGet {
             array,
             index: zero,
-            offset,
+            offset: ArrayOffset::None,
         })
     } else {
         result
--- a/compiler/noirc_evaluator/src/ssa/ir/instruction.rs
+++ b/compiler/noirc_evaluator/src/ssa/ir/instruction.rs
@@ -453,7 +453,7 @@ impl Instruction {
         match self {
             Instruction::Binary(binary) => binary.requires_acir_gen_predicate(dfg),
 
-            Instruction::ArrayGet { array, index, offset: _ } => {
+            Instruction::ArrayGet { array, index, offset } => {
                 // `ArrayGet`s which read from "known good" indices from an array should not need a predicate.
                 // This extra out of bounds (OOB) check is only inserted in the ACIR runtime.
                 // Thus, in Brillig an `ArrayGet` is always a pure operation in isolation and
@@ -461,6 +461,11 @@ impl Instruction {
                 // not be safe to separate the `ArrayGet` from the OOB constraints that precede it,
                 // because while it could read an array index, the returned data could be invalid,
                 // and fail at runtime if we tried using it in the wrong context.
+                assert_eq!(
+                    *offset,
+                    ArrayOffset::None,
+                    "Array offsets are not supposed to be set yet"
+                );
                 !dfg.is_safe_index(*index, *array)
             }
 
@@ -567,7 +572,14 @@ impl Instruction {
             // used in the wrong context. Since we use this information to decide whether to hoist
             // instructions during deduplication, we consider unsafe values as potentially having
             // indirect side effects.
-            ArrayGet { array, index, offset: _ } => !dfg.is_safe_index(*index, *array),
+            ArrayGet { array, index, offset } => {
+                assert_eq!(
+                    *offset,
+                    ArrayOffset::None,
+                    "Array offsets are not supposed to be set yet."
+                );
+                !dfg.is_safe_index(*index, *array)
+            }
 
             // ArraySet has side effects
             ArraySet { .. } => true,
--- a/compiler/noirc_evaluator/src/ssa/opt/brillig_array_get_and_set.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/brillig_array_get_and_set.rs
@@ -23,14 +23,14 @@
 //!
 //! Now the index to retrieve is 4 and there's no need to offset it in Brillig,
 //! avoiding one addition.
+//!
 //! The "minus 1" part is just there so that readers can understand that the index
 //! was offset and that the actual element index is 3. On the Brillig side,
 //! array operations with constant indexes are always assumed to have already been
 //! shifted.
 //!
-//! In the case of slices, these are represented as Brillig vectors, where the items
-//! pointer instead starts at three rather than one.
-//! A Brillig vector is represented as [RC, Size, Capacity, ...items]. So for a slice
+//! In the case of slices, they are represented as Brillig vectors as [RC, Size, Capacity, ...items],
+//! thus the items pointer instead starts at three rather than one. So for a slice
 //! this pass will transform this:
 //!
 //! ```ssa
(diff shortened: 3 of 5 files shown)


TomAFrench:
Looks like this is for https://github.com/noir-lang/noir/issues/9218

asterite:
> however during Brillig codegen we [check if the instruction has offset](https://github.com/noir-lang/noir/blob/74b44a7a683ee8447cf00d39b864f1ad9c8a05f8/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs#L840-L846) and if the index is dynamic then we [have to generate offsets](https://github.com/noir-lang/noir/blob/74b44a7a683ee8447cf00d39b864f1ad9c8a05f8/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs#L1166-L1169) anyway, so why don't we insert the necessary constants into the DFG during codegen (not Brillig opcodes to calculate them, just a new constant based on the Array/Slice type), and then avoid having to have the offset: _ pattern throughout the codebase, and having to reason about whether using index without considering offset is correct or not?

I think this could be done, though I didn't check where constants are initialized in brillig (or, well, it doesn't seem there's a single place where this happens).

I think the reason is that if you have this:

```
v0 = array_get v1, index u32 10 -> Field
```

brillig will generate a CONST op for 10, then refer to that memory address for the array get.

That `u32 10` comes from a `ValueId`. Then a numeric constant is transformed into a const here: https://github.com/noir-lang/noir/blob/ca32861f78f9668371ac49e47edbbb7c9aae0984/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs#L1861

I guess what you are proposing is, when in Brillig we need to generate code for the above array_get, if we see a constant index and, say, it's an array and not a slice, transform that `10` into `11`. But then it's not guaranteed that that `11` has a `ValueId` so we'd need to either generate a new `ValueId` (impossible at this point) or just to brillig math with opcodes to reach that final value. Or, alternatively, which I guess is also what you propose here, is to do a pre-pass in Brillig to already have that `11` as a CONST somewhere, or as a `ValueId`, etc. But I think this is tricky to do, though it might be very well worth it to simplify things.

(I'm not entirely sure what I said above is correct so someone please correct me)

vezenovm:
> if the index is dynamic then we [have to generate offsets](https://github.com/noir-lang/noir/blob/74b44a7a683ee8447cf00d39b864f1ad9c8a05f8/compiler/noirc_evaluator/src/brillig/brillig_gen/brillig_block.rs#L1166-L1169) anyway

In theory we could. I originally did the optimization this way, but I cannot remember why I did not keep it that way. 

> But then it's not guaranteed that that `11` has a `ValueId` so we'd need to either generate a new `ValueId` (impossible at this point) or just to brillig math with opcodes to reach that final value

Yes we cannot simply have the constants as part of the DFG and use them in Brillig. They still needed to be allocated to a register using the CONST opcode. The block in which they are allocated can have a significant affect on the byte code executed (e.g., in a loop vs. in the pre-header). 

> Or, alternatively, which I guess is also what you propose here, is to do a pre-pass in Brillig to already have that `11` as a CONST somewhere, or as a `ValueId`, etc. But I think this is tricky to do, though it might be very well worth it to simplify things.

So we do already have a pre-pass on Brillig where we [declare the locations to allocate constants](https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/brillig/brillig_gen/constant_allocation.rs). We can see some of the regressions when we attempted to move this logic back to Brillig gen https://github.com/noir-lang/noir/pull/8532. We could just move this pass to `Ssa::to_brillig` as a way to guarantee it is only a pre-Brillig pass and separate from our normal SSA pipeline. This was always meant to be the last pass to precede Brillig's constant allocation pass. I would rather keep the pass and move where it is run. It will have the same effect as the pre-pass @asterite mentioned.

vezenovm:
I agree that the idea of the "offset" is confusing. But I would say this is largely due to ACIR/Brillig sharing instructions to perform array operations while their array semantics differ. In ACIR arrays are truly arrays representing a block of memory. In Brillig they are basically pointers. Thus, those pointers have their own layout and they must be appropriately offset. All this pass does is essentially pre-compute any offsets it can. So yes, this is unique to Brillig, but it also needs to occur before [constant allocation](https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/brillig/brillig_gen/constant_allocation.rs). I would not want to clutter constant allocation with this logic as it is better to maintain separation of concerns.

aakoshh:
> Or, alternatively, which I guess is also what you propose here, is to do a pre-pass in Brillig to already have that 11 as a CONST somewhere, or as a ValueId, etc. But I think this is tricky to do, though it might be very well worth it to simplify things.

I wonder if we could do a bit of a slight of hand:
1. In `brillig_array_get_set` we visit all the `ArrayGet` and `ArraySet`, we call `compute_index_and_offset` to lay the constant in the `&mut DataFlowGraph` while we can mutate it with `make_constant`, but _do not_ update the `index` in the instruction. We just made some groundwork. 
2. In `BrilligBlock::convert_ssa_instruction` we recalculate the offset and call a new `DataFlowGraph::get_constant` method which looks up the previously prepared constants. It's ICE if they are not there. 
3. Since these constants don't appear in the SSA, we call `brillig_array_get_set` as part of Brillig gen at an appropriate place. 

This way the internal memory layout of Brillig remains its own concern, and we don't need the `offset` field any more. Since we are using `make_constant`, as long as they are not forgotten, they get deduplicated with any other use of the same value ID.

aakoshh:
Although I do realise this doesn't mesh well with functions like `variables_used_in_instruction`, so maybe modifying the `index` is the best course of action, but perhaps if this pass was "internal" to the Brillig generation, then at least we wouldn't need to keep the `offset` on the instructions.

vezenovm:
> * In `BrilligBlock::convert_ssa_instruction` we recalculate the offset and call a new `DataFlowGraph::get_constant` method which looks up the previously prepared constants. It's ICE if they are not there.
> * Since these constants don't appear in the SSA, we call `brillig_array_get_set` as part of Brillig gen at an appropriate place.

This feels like added complexity for little gain imo (all we get is that removed the `offset` field). A little nit-pick is also that if we are already inside of `BrilligBlock` we have skipped [constant allocation](https://github.com/noir-lang/noir/blob/master/compiler/noirc_evaluator/src/brillig/brillig_gen/constant_allocation.rs) and will see performance regressions (this mostly will affect array accesses in loops). Doing something like what you propose would require modifying the constant allocation analysis. 

> This way the internal memory layout of Brillig remains its own concern, and we don't need the `offset` field any more. Since we are using `make_constant`, as long as they are not forgotten, they get deduplicated with any other use of the same value ID.

The key thing is that they will only be deduplicated in the DFG. Not in Brillig memory. As you said Brillig is responsible for its own memory allocation and must determine how to allocate registers for constants as this is not a concept in our SSA.  I contend it would be cleaner to declare constants while still in SSA form and lower them with a single strategy in Brillig gen. I agree though that this logic could be more clearly made Brillig only. 

> Although I do realise this doesn't mesh well with functions like `variables_used_in_instruction`, so maybe modifying the `index` is the best course of action, but perhaps if this pass was "internal" to the Brillig generation, then at least we wouldn't need to keep the `offset` on the instructions.

I think it may be cleanest to still just keep the `offset` while the two runtimes still share array instructions. We can add a restriction on SSA that ACIR array instructions must always have `ArrayOffset::None`. Another option would be to introduce a `BrilligArrayGet` and `BrilligArraySet` and that has an offset and is only introduced after this pass. I don't really like this option though. I think the best middle ground would be moving this to be an internal Brillig SSA pass. In general, separate compilation pipelines for our runtime semantics that differ significantly would be ideal so I think doing this would be a move in that direction.

aakoshh:
Thanks, yeah, I see that it's all about the DFG. I'll open a PR soon that removes the `offset` from the instructions, while keeping this logic in the pass, and does its best to even present it to humans. 

Here: https://github.com/noir-lang/noir/pull/9956

aakoshh:
Closing in favour of https://github.com/noir-lang/noir/pull/9956

PR #9955, opened 2025-09-22
chore(acir): Do not copy element type sizes array when initializing

vezenovm:
# Description

## Problem\*

Working towards green-lighting of ACIR gen

## Summary\*

Still a draft as I want to see its effect on CI

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/acir/arrays.rs
+++ b/compiler/noirc_evaluator/src/acir/arrays.rs
@@ -836,19 +836,8 @@ impl Context<'_> {
                     unreachable!("ICE: element type size arrays are expected to be initialized");
                 }
 
-                let type_sizes_array_len = *self.internal_mem_block_lengths.get(inner_elem_type_sizes).ok_or_else(||
-                                            InternalError::General {
-                                                message: format!("Array {array_id}'s inner element type sizes array does not have a tracked length"),
-                                                call_stack: self.acir_context.get_call_stack(),
-                                            }
-                                        )?;
-                self.copy_dynamic_array(
-                    *inner_elem_type_sizes,
-                    element_type_sizes,
-                    type_sizes_array_len,
-                )?;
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/acir/arrays.rs
```diff
@@ -836,19 +836,8 @@ impl Context<'_> {
...
-                                            }
-                                        )?;
-                self.copy_dynamic_array(
-                    *inner_elem_type_sizes,
-                    element_type_sizes,
-                    type_sizes_array_len,
-                )?;
-                self.internal_mem_block_lengths.insert(element_type_sizes, type_sizes_array_len);
-                Ok(element_type_sizes)
+                self.internal_memory_blocks.insert(array_id, *inner_elem_type_sizes);
```
vezenovm:
```suggestion
                // We can safely overwrite the memory block from the initial call to `self.internal_block_id(&array_id)` here.  
                // The type sizes array is never mutated so we can re-use it. 
                self.internal_memory_blocks.insert(array_id, *inner_elem_type_sizes);
```

TomAFrench:
Spoke to Maxim about this and we're adding a check to enforce that these arrays are immutable.

compiler/noirc_evaluator/src/acir/mod.rs
```diff
@@ -256,6 +248,9 @@ impl<'a> Context<'a> {
         warnings.extend(return_warnings);
         warnings.extend(self.acir_context.warnings.clone());
 
+        #[cfg(debug_assertions)]
+        acir_post_check(&self, self.acir_context.acir_ir.opcodes());
```
vezenovm: This felt like the best place for the post check as we need access to the `Context`.

TomAFrench (approved):
LGTM

PR #9956, opened 2025-09-22
Af/array idx offset during brillig gen

aakoshh:
# Description

## Problem\*

Resolves https://github.com/noir-lang/noir/issues/9218

## Summary\*

Removes the `offset` field from `ArraySet` and `ArrayGet` so there is no confusion during the other SSA passes whether we should care about the `offset` field.

Moves the `brillig_array_get_and_set` SSA pass to be performed internally in the `optimize_builder_into_acir` function, not as part of the configurable SSA pipelines. This way we can make sure it's run exactly once, and that no other passes are performed after it, where the indexes could have in unexpected values.

Keeps the support for parsing and displaying the `minus 1` (array) and `minus 3` (slice) offsets, but the effect is based on a single `brillig_arrays_offset` flag added to the `DataFlowGraph`. 

## Additional Context

The `offset` field should only ever be set to anything but `None` in ACIR functions, and only at the end, just before Brillig generation, but it is often ignored, to the extent that if it _was_ set, it could silently change the behaviour of the compiler, or cause panics. 

See the discussion in https://github.com/noir-lang/noir/pull/9948 for what has been attempted before, and why it is performed as an SSA pass. The TL;DR is that we need these constants to be part of the DFG so they can flow into the constant allocation algorithm.

## Documentation\*

Check one:

# PR Checklist\*


compiler/noirc_evaluator/src/ssa/parser/tests.rs
```diff
@@ -7,6 +7,7 @@ use crate::{
 
 fn assert_ssa_roundtrip(src: &str) {
     let ssa = Ssa::from_str(src).unwrap();
+    println!("offset: {}", ssa.main().dfg.brillig_arrays_offset);
```
aakoshh, line 10: Oops, I'll remove this in the next PR.

PR #9962, opened 2025-09-23
chore(acir): More arrays refactors 

vezenovm:
# Description

## Problem\*

Working towards green-lighting of ACIR

## Summary\*

As array get/set started having more requirements we kind of just shoved things into the main methods. This refactors our main `array_set` and `array_get` method to be more readable. 

I also moved some of the type assertions such as that we do not have nested arrays into the validation module so that our SSA invariants are enforced in one central location.

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/acir/arrays.rs
+++ b/compiler/noirc_evaluator/src/acir/arrays.rs
@@ -251,7 +251,7 @@ impl Context<'_> {
         if !self.has_zero_length(array, dfg) {
             return Ok(false);
         }
-        
+
         // Zero result.
         let result_ids = dfg.instruction_results(instruction);
         for result_id in result_ids {


compiler/noirc_evaluator/src/acir/arrays.rs
```diff
@@ -279,19 +228,59 @@ impl Context<'_> {
...
+    /// # Returns
+    /// `true` if we have a zero length array
+    /// `false` if we do not have a zero length array
+    fn handle_zero_length_array(
+        &mut self,
+        array: ValueId,
+        dfg: &DataFlowGraph,
+        instruction: InstructionId,
+    ) -> Result<bool, RuntimeError> {
+        if self.has_zero_length(array, dfg) {
```
asterite: Just a minor thing, but if this condition, inverted, is used to early return `Ok(false)` then the `assert_eq_var` call fits in one line, making things a bit shorter.

vezenovm: I went with the early escape.

asterite (approved):

PR #9965, opened 2025-09-23
chore(ssa): Nits in `remove_bit_shift` and `remove_if_else`

aakoshh:
# Description

## Problem\*

Part of auditing.

## Summary\*

* Fixes some typos in `remove_bit_shift` and `remove_if_else`
* Some refactoring in `remove_bit_shift`:
  * Makes `get_value_max_num_bits` recursive, which has also been added to https://github.com/noir-lang/noir/pull/9770
  * Seek clarification on why the 1-complement and 2-complement logic seems to do the opposite of how the literature describes we should do conversions, and add it to the docs. This was introduced in 2247814f951f5d33257cd123a3bdcba857c9b167 .
  * Got the source from @guipublic 🙏 https://dspace.mit.edu/bitstream/handle/1721.1/6090/AIM-378.pdf 
  * Added an extended explanation, thanks for the hints Guillaume 🍻 
* Some refactoring in `remove_if_else`:
  * Make all slice capacity calculations into `remove_if_else`, leaving the `ValueMerger` free of it.
  * Only store the capacity of slices, not arrays.
  * Unify the `SizeChange` to always return just IDs. 
  * Move assertions into the pre-check

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_if_else.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_if_else.rs
@@ -249,14 +249,17 @@ impl Context {
         new: ValueId,
         f: impl Fn(u32) -> u32,
     ) {
+        // No need to store the capacity of arrays, only slices.
+        if !matches!(dfg.type_of_value(new), Type::Slice(_)) {
+            return;
+        }
         let capacity = self.get_or_find_capacity(dfg, old);
         self.slice_sizes.insert(new, f(capacity));
     }
 
-    /// Make sure the capacity is recorded.
+    /// Make sure the slice capacity is recorded.
     fn ensure_capacity(&mut self, dfg: &DataFlowGraph, slice: ValueId) {
-        let capacity = self.get_or_find_capacity(dfg, slice);
-        self.slice_sizes.insert(slice, capacity);
+        self.set_capacity(dfg, slice, slice, |c| c);
     }
 
     /// Get the tracked size of array/slices, or retrieve (and track) it for arrays.
@@ -264,14 +267,14 @@ impl Context {
         match self.slice_sizes.entry(value) {
             Entry::Occupied(entry) => *entry.get(),
             Entry::Vacant(entry) => {
+                // For arrays we know the size statically, and we don't need to store it.
+                if let Type::Array(_, length) = dfg.type_of_value(value) {
+                    return length;
+                }
                 // Check if the item was made by a MakeArray instruction, which can create slices as well.
                 if let Some((array, typ)) = dfg.get_array_constant(value) {
                     let length = array.len() / typ.element_types().len();
                     return *entry.insert(length as u32);
-                }
-                // For arrays we know the size statically.
-                if let Type::Array(_, length) = dfg.type_of_value(value) {
-                    return *entry.insert(length);
                 }
                 // For non-constant slices we can't tell the size, which would mean we can't merge it.
                 let dbg_value = &dfg[value];


asterite:
We should run benchmarks here because the last time I tried making this recursive it considerably slowed things down. That's why there's memoization in the other place where this happens.

aakoshh:
I think that’s because you handled Binary, whereas I just added support for the unlikely case of a chain of casts, potentially originating from a constant. 

I did it more just because it seems like an obvious place where some sort of recursion is needed, but I don’t expect it to be common

TomAFrench:
>  I just added support for the unlikely case of a chain of casts

I think that the dfg flattens these whenever we encounter chains for casts anyway so I'd expect this to not fire.

aakoshh:
> I think that the dfg flattens these whenever we encounter chains for casts anyway so I'd expect this to not fire.

Fair enough. I think another case recursion handles is casting a constant to another type, for example:
```
v1 = u8 7
v2 = cast v1 as u32
```
For this with recursion it will now return 3 instead of 8. 

I can imagine that the simplification does some magic on this too, but I reckon it's easier to see that the recursive version would work for constants across casts (which are the only two cases it handles, other than types), without having to take into account all the other machinery.

Put differently, I'd expect an auditor to highlight the that this method is incomplete, when looked at in isolation, so I thought better handle something that might never come up, rather than have to explain why we think this is fine.

compiler/noirc_evaluator/src/ssa/opt/remove_if_else.rs
```diff
@@ -218,65 +215,89 @@ impl Context {
...
                 }
                 // Track slice sizes through array set instructions
                 Instruction::ArraySet { array, .. } => {
                     let results = context.dfg.instruction_results(instruction_id);
-                    let result = if results.len() == 2 { results[1] } else { results[0] };
-
-                    let old_capacity = self.get_or_find_capacity(context.dfg, *array);
-                    self.slice_sizes.insert(result, old_capacity);
+                    let result = results[0];
+                    self.set_capacity(context.dfg, *array, result, |c| c);
```
asterite, line 236: I think `ArraySet` for an array returns one result, but for a slice it returns two: the length, and the slice. Maybe we shouldn't do anything here for the array case, just handle the slice case?

aakoshh: Thanks for commenting on this. I thought that the case of 2 return values might be for slices, but then I didn't understand how that could be: setting an item in the slice does not modify the length, so why should it be returned?  I thought that it must look exactly the same as it does for an array. 

I tried to find any evidence of `ArraySet` returning 2 values anywhere, and thought surely [make_instruction_results](https://github.com/noir-lang/noir/blob/03d9dac082c2d11cb055327fc75f5c67df0cbd68/compiler/noirc_evaluator/src/ssa/ir/dfg.rs#L479-L492) would know. It calls [Instruction::result_type](https://github.com/noir-lang/noir/blob/03d9dac082c2d11cb055327fc75f5c67df0cbd68/compiler/noirc_evaluator/src/ssa/ir/instruction.rs#L424-L426) which, in my reading, simply returns the type of the `array` field for an `ArraySet`, a single value.

Looking for "= array_set" in the codebase only shows single result assignments.

aakoshh: > Maybe we shouldn't do anything here for the array case, just handle the slice case?

You are right that storing the length of arrays is not doing anything super useful. I saw that `get_or_find_capacity` does it anyway, and had a similar conundrum when I added `ensure_capacity` to the handling of `IfElse`: if it's an `Array`, then `merge_values` won't read it, it will get it from the type. 

I suppose I can change `set_capacity` to skip doing it for `Array`, to save some memory.

aakoshh: Done in https://github.com/noir-lang/noir/pull/9965/commits/f5f411a416d1b111696c5dd1e067e1e16914b31d

aakoshh: On the topic of number of results: maybe we could add a helper method for this pattern:
```rust
let results = context.dfg.instruction_results(instruction_id);
let result = results[0];
```
For example:
```rust
    pub(crate) fn instruction_result(&self, instruction_id: InstructionId) -> &ValueId {
        let results = self.instruction_results(instruction_id);
        assert_eq!(results.len(), 1, "expected {instruction_id} to have 1 result");
        results[0]
    }
```
And then we would at least be notified if this assumption is wrong.

asterite: That would be nice! I ended up doing something like that in the SSA validations module.

aakoshh: Or `instruction_result<const N: usize>(...) -> [ValueId; N]`, which I used somewhere in the blackbox function interface.

PR #9966, opened 2025-09-23
chore: add a proptest to ensure that converting between complements restores input

TomAFrench:
…

# Description

## Problem\*

Resolves 

## Summary\*

This PR adds a quick test to sanity check that the conversion between two's and one's complement is correct in the `remove_bit_shifts` pass.

As part of this I've started creating an `InstructionBuilder` which basically acts as a wrapper around a `DataflowGraph` which makes it easier for us to generate reusable sets of instructions. 

## Documentation\*

Check one:

# PR Checklist\*

### DIFF
--- a/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
+++ b/compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
@@ -261,13 +261,14 @@ impl Context<'_, '_, '_> {
                 // Get the sign of the operand; positive signed operand will just do a division as well
                 let unsigned_typ = NumericType::unsigned(bit_size);
                 let lhs_as_unsigned = instruction_builder.insert_cast(lhs, unsigned_typ);
-                
+
                 // The sign will be 0 for positive numbers and 1 for negatives, so it covers both cases.
                 // To compute this we check if the value, as a Field, is greater or equal than the maximum
                 // value that is considered positive, that is, 2^(bit_size-1)-1: 2^(bit_size-1)-1 < lhs_as_field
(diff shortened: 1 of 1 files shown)


compiler/noirc_evaluator/src/ssa/opt/remove_bit_shifts.rs
```diff
@@ -222,39 +223,96 @@ impl Context<'_, '_, '_> {
...
-                let lhs_sign_as_field = self.insert_cast(lhs_sign, NumericType::NativeField);
-                let lhs_as_field = self.insert_cast(lhs, NumericType::NativeField);
-                // For negative numbers, convert to 1-complement using wrapping addition of a + 1
-                // Unchecked add as these are fields
-                let add = BinaryOp::Add { unchecked: true };
-                let one_complement = self.insert_binary(lhs_sign_as_field, add, lhs_as_field);
-                let one_complement = self.insert_truncate(one_complement, bit_size, bit_size + 1);
                 let one_complement =
-                    self.insert_cast(one_complement, NumericType::signed(bit_size));
+                    instruction_builder.convert_twos_complement_to_ones_complement(lhs, lhs_sign);
```
aakoshh: FYI I updated https://github.com/noir-lang/noir/pull/9965/files#diff-6a664951cab26f1ffa0e8854826acb88fe48bc294100187283ebd81cd6a7c874 with more comments after discussing with Guillaume: the comments that said we are converting from 2-complement to 1-complement were misleading, that's not what it did. I don't think there's any particular name for the converted value.

TomAFrench: Thanks for flagging this up! I'll take a look and update this PR

